PR #451, opened 2014-11-11
GA Gazetteer Search

Layoric:
Hi Team!

I'm currently tasked with building additional features required by GA for national map. I've expanded the search widget to include a way for users to search GA's place name Gazetteer as requested.
#### Main changes
1. Introduction of jQuery UI
2. Fixes for z-index to allow more levels between overlay and map.
3. Knockout binding for jQuery UI Autocomplete

I've noticed none of the more common UI frameworks are being used in this project, so please let me know if this introduction of a dependency on jQuery UI is a problem.

I've tried to follow conventions I've found in the project as much I could whilst building this feature, but would be very appreciative of any feedback.

Cheers!


kring:
@Layoric is the Gazetteer service still up?  I'm finally getting around to rolling it into the new UI, but I'm getting 404s from URLs like this:
http://www.ga.gov.au/gazetteer-search/select/?q=name:*garden*

Layoric:
@kring I have learnt this has been changed to:

http://www.ga.gov.au/gazetteer-search/gazetteer2012/select/

Apparently due to Solr config issues here locally.

Layoric:
@kring looks like more than just the URL has change, response now in JSON where as it was previously an XML response. I'll update the PR if that helps.

Layoric:
Updated handling new response.

kring:
That's a huge help, thanks @Layoric!

Layoric:
This looks like it's incorporated into new UI. Closing.

PR #481, opened 2015-01-12
Fix scale bar for 2D view

Layoric:
The scale bar doesn't update when in 2D mode.

Added support using some leaflet functions.


kring:
Thanks @Layoric!  One thing problem I noticed is that you're only updating the scale when the zoom changes, but you also need to update it when the user pans the map.  The reason is that the map has different scales at different latitudes.  That can be a separate pull request, though, if you're up for it.  Merging.

Layoric:
Ahh right, I'll have a look at that. Thanks!

PR #483, opened 2015-01-12
Update 2D scale on pan.

Layoric:
As suggested. Update on pan. Scale is based on the middle of the map position.


kring:
Thanks @Layoric!

PR #485, opened 2015-01-14
CKAN support for base NM Catalog

kgrochow:
This builds on the CKAN catalog support for vic and tries to make CKAN catalogs and our JSON init file as compatible as possible so that either can be used for initialization.
- A CKAN exporter that exports the current open/selected NM catalog items to ckan via their api.
- An init_ckan file for loading our datasets from a ckan catalog.
- A CKAN instance up at http://54.79.117.201 for now (still waiting for the domain from IT).
- Cleanups to the init_nm file to make it more consistent with ckan naming, etc.
- Allowing multiple and more complex searches in ckan (groups, wms + esri rest)
- Esri REST support for ckan catalog items and in our exporter


src/ViewModels/ArcGisMapServerItemViewModel.js
```diff
@@ -43,7 +43,7 @@ defineProperties(ArcGisMapServerItemViewModel.prototype, {
      */
     type : {
         get : function() {
-            return 'esri-mapServer';
+            return 'esri-rest';
```
kring: Why this change?  Esri has lots of REST APIs.  `esri-mapServer` is a much more precise description of what this catalog item talks to.

src/ViewModels/ArcGisMapServerItemViewModel.js
```diff
@@ -66,6 +66,8 @@ ArcGisMapServerItemViewModel.prototype._enableInCesium = function() {
 
     var scene = this.application.cesium.scene;
 
+    this.legendUrl = this.url + '/legend';
```
kring: This shouldn't be done here.  See how `WebMapServiceItemViewModel` handles the `legendUrl`.

src/ViewModels/ArcGisMapServerItemViewModel.js
```diff
@@ -99,6 +101,8 @@ ArcGisMapServerItemViewModel.prototype._enableInLeaflet = function() {
         throw new DeveloperError('This data source is already enabled.');
     }
 
+    this.legendUrl = this.url + '/legend';
```
kring: Or here.

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -76,6 +77,14 @@ var CkanGroupViewModel = function(application) {
      */
     this.minimumMaxScaleDenominator = undefined;
 
+    /**
+     * Gets or sets any extra wms parameters that should be added to the wms query urls in this CKAN group.
+     * If this property is undefined or if {@link CkanGroupViewModel#filterByWmsGetCapabilities} then no parameters are added
```
kring: I'm pretty sure this sentence isn't true.  `wmsParameters` is passed to the created WMS catalog items as their `parameters` property, regardless of the `filterByWmsGetCapabilities` property, correct?

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -159,40 +168,57 @@ CkanGroupViewModel.prototype._load = function() {
...
-and ask us to add this server to the list of non-CORS-supporting servers that may be proxied by \
-National Map itself.</p>\
-<p>If you did not enter this link manually, this error may indicate that the group you opened is temporarily unavailable or there is a \
-problem with your internet connection.  Try opening the group again, and if the problem persists, please report it by \
-sending an email to <a href="mailto:nationalmap@lists.nicta.com.au">nationalmap@lists.nicta.com.au</a>.</p>'
+        };
+
+    //FYI: to filter by group the filterQuery param should be set to something like 
+    //  'q=groups%3dSurface%20Water&fq=res_format%3aWMS' in init_xx.json
+    var promises = [];
```
kring: This note should go on the documentation for the `filterQuery` property so that people building apps on NM can see it without studying our source code.

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -159,40 +168,57 @@ CkanGroupViewModel.prototype._load = function() {
...
-Map team by emailing <a href="mailto:nationalmap@lists.nicta.com.au">nationalmap@lists.nicta.com.au</a> \
-and ask us to add this server to the list of non-CORS-supporting servers that may be proxied by \
-National Map itself.</p>\
-<p>If you did not enter this link manually, this error may indicate that the group you opened is temporarily unavailable or there is a \
-problem with your internet connection.  Try opening the group again, and if the problem persists, please report it by \
-sending an email to <a href="mailto:nationalmap@lists.nicta.com.au">nationalmap@lists.nicta.com.au</a>.</p>'
+        };
+
+    //FYI: to filter by group the filterQuery param should be set to something like 
+    //  'q=groups%3dSurface%20Water&fq=res_format%3aWMS' in init_xx.json
```
kring: This note should go on the documentation for the filterQuery property so that people building apps on NM can see it without studying our source code.

public/init_vic.json
```diff
@@ -11,7 +11,7 @@
             "name": "Victorian Government",
             "type": "ckan",
             "url": "http://www.data.vic.gov.au/data",
-            "filterQuery": "res_format:wms",
+            "filterQuery": ["fq=res_format%3awms"],
```
kring, line 14: I think it would be a lot cleaner to introduce a separate `query` property to `CkanGroupViewModel`, in addition to `filterQuery`, and handle the URL encoding automatically.  Both properties could still optionally be arrays if that's helpful.

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -159,40 +168,57 @@ CkanGroupViewModel.prototype._load = function() {
...
-
     var that = this;
 
-    return loadJson(url).then(function(json) {
-        if (that.filterByWmsGetCapabilities) {
-            return when(filterResultsByGetCapabilities(that, json), function() {
-                populateGroupFromResults(that, json);
+    var allResults = {result: {results: []}};
+    var handleResults = function(json) {
+            allResults.result.results = allResults.result.results.concat(json.result.results);
```
kring: `when.all` resolves to an array, where each element in the array corresponds to the resolved value of the promise in the same position in the array passed to `when.all`.  So you can do this concatenation in the `.then` at the bottom of this `_load` function, and it would be a lot easier to understand what's going on, IMO.  Plus you'd get deterministic ordering.  Currently results are concatenated in whatever order the promises happen to resolve in.

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -76,6 +77,14 @@ var CkanGroupViewModel = function(application) {
...
 
+    /**
+     * Gets or sets any extra wms parameters that should be added to the wms query urls in this CKAN group.
+     * If this property is undefined or if {@link CkanGroupViewModel#filterByWmsGetCapabilities} then no parameters are added
+     * This property is observable.
+     * @type {Object}
+     */
+    this.wmsParameters = undefined;
+
     knockout.track(this, ['url', 'dataCustodian', 'filterQuery', 'blacklist']);
```
kring: `wmsParameters` should be added to the array here so that that property is actually observable, as its documentation claims.

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -307,12 +333,21 @@ function populateGroupFromResults(viewModel, json) {
             textDescription = item.notes.replace(/\n/g, '<br/>');
         }
 
-        if (item.license_url) {
+        if (item.license_url && (item.notes === null || item.notes.indexOf('[Licence]') === -1)) {
             textDescription += '<br/>[Licence](' + item.license_url + ')';
         }
 
+        var extras = {};
+        if (item.extras !== undefined) {
```
kring: Use `defined(item.extras)`

src/ViewModels/WebMapServiceItemViewModel.js
```diff
@@ -108,8 +108,8 @@ var WebMapServiceItemViewModel = function(application) {
     delete this.__knockoutObservables.dataUrlType;
     knockout.defineProperty(this, 'dataUrlType', {
         get : function() {
-            if (defined(this._dataUrlType)) {
-                return this._dataUrlType;
+            if (defined(this._dataUrl)) {
+                return 'direct';
```
kring: No, not all dataUrls should be considered direct links.  Why was this change necessary?

src/ViewModels/CkanGroupViewModel.js
```diff
@@ -339,23 +374,31 @@ function populateGroupFromResults(viewModel, json) {
             // Extract the layer name from the WMS URL.
             var uri = new URI(wmsUrl);
```
kring: It's confusing that this variable is called `wmsUrl` when it's no longer necessarily a WMS URL.

src/ViewModels/CkanGroupViewModel.js
kgrochow: This is so we can pass the tiled parameter on source we detect have tiled.  This is detected in getCapabilities version 1.1.1

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: This is in line with ckan definitions from what I could find and most esri sites.

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: Except that this is not really a settable/trackable item. This is where there url is defined to be.  Setting it would only make it wrong.

src/ViewModels/CkanGroupViewModel.js
kgrochow: This is basically your original code in a different order.  I haven't been looking for things to document from that pass.

src/ViewModels/CkanGroupViewModel.js
kring: Ok, so if `filterByWmsGetCapabilities` is true, we're going to do a `GetCapabilities` anyway, so we'll ignore `wmsParameters` and instead deduce the correct parameters based on what we see in `GetCapabilities`.  That's reasonable, but I don't see the code actually doing that.  Am I missing something?

src/ViewModels/ArcGisMapServerItemViewModel.js
kring: Not sure about CKAN, but Esri REST API docs lists a whole bunch of APIs.  We're only using the one called "Map Service":
http://resources.arcgis.com/en/help/rest/apiref/

src/ViewModels/WebMapServiceItemViewModel.js
kgrochow: Because dataUrls are now direct and wms.  The one outlier we had was removed and it now points to the ckan page rather some broken url in the json file.

src/ViewModels/ArcGisMapServerItemViewModel.js
kring: That's the only correct URL for the legend produced by the ArcGIS Server; I agree with that.  But you could point NM to a separate, hand-generated legend URL.  We used to do this with the Broadband layers.

src/ViewModels/CkanGroupViewModel.js
kring: It doesn't look like a comment I'd write, but ok.

public/init_vic.json
kgrochow, line 14: Ckan queries can be quite extensive including urlencoded JSON objects.  It's seems all the different methods have strengths and weaknesses so I left it the most flexible.  Also there is a 'wonderful' search feature which won't allow us to search for BOTH esri resources and wms resources, so it needs to be an array to carry out multiple searches if necessary.

src/ViewModels/CkanGroupViewModel.js
kgrochow: I was looking at the wrong section.  The documention will need to refer them to the ckan search documentation.  It is fiddly as all get out.

src/ViewModels/WebMapServiceItemViewModel.js
kring: Let's say I have a WMS and a WFS, but it's not a Geoserver serving both roles, it's two separate servers.  My `url` property might be `http://wms.made.up.nicta.com.au/fancyServer` and my `dataUrl` would be `http://wfs.made.up.nicta.com.au/someOtherServer`.  But that's not a direct URL.  It's a WFS URL that needs to have the layer name and such added.  This scenario would be impossible after your change.

Sure, it might not be an important use-case, but it's also easy to support (and was supported until this change).  In general, properties that are settable in a `CatalogItemViewModel` (such as `dataUrlType`) should also be settable in derived classes like this one, even if there is a reasonable default.

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: I've dropped legendUrl from the init file and instead figured out what the 'right' url should be in all our random cases so we can generate it based on the service.  I want to hopefully stay out of that business and let the providers set their servers up properly.  We can always add it back later if someone gives us a sensible reason.

src/ViewModels/CkanGroupViewModel.js
kring: What you have is fine, no need for it to be comprehensive.  Just having that example is great.  I'm just suggesting you move what you already have to a place where users of our API will be able to see it.

src/ViewModels/ArcGisMapServerItemViewModel.js
kring: But you haven't dropped it from the init file.  It's still there, and still used.  But then when the layer is enabled, the init-specified legend gets clobbered and replaced with a different one.  If we ever put the legend on the info popup (or anywhere else you could see without enabling the layer), the user would see the init file legend before enabling it, and the server's legend afterward.

Also, I think the ability to specify a legend other than the one the server suggests is a handy bit of functionality, and it's already implemented.  I don't see any good reason to get rid of it.

src/ViewModels/CkanGroupViewModel.js
kgrochow: I was thinking of a different file.  Here it was meant to pass in tiled=true for an entire ckan group which is appended at line 395.  It was a quick solution for our base NM data.  It is probably not necessary when we make tiled the wms default in a later checkin.

kring: Yes, I understand that.  The doc comment is just wrong.

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: It's not there.  It's been removed in all instances.  It was a hack then and it hasn't improved with age.

kring: I'm sorry, we must be talking about two different things, because I have no idea what you mean. :)

kring: Here's where the `legendUrl` property is defined in the base class:
https://github.com/NICTA/nationalmap/blob/ckanNMinit/src/ViewModels/CatalogItemViewModel.js#L62

Then this code in `CatalogMemberViewModel` will set that property from the value in the init file:
https://github.com/NICTA/nationalmap/blob/ckanNMinit/src/ViewModels/CatalogMemberViewModel.js#L188

Until `_enableInCesium` or `_enableInLeaflet` is invoked, the `legendUrl` property will have whatever value was specified in the init file.  After one of those functions is invoked, it will be set to `this.url + '/legend'`.

On the other hand, if you do what I'm proposing and just copy/paste the `legendUrl` property code from `WebMapServiceItemViewModel`, it will work exactly as it should (at least according to me).  If `legendUrl` is not specified in the init file, its value will be `this.url + '/legend'`.  If it is specified in the init file, the value will be whatever was specified in the init file.  I don't really see how any of this is a hack.

kring: I'm happy to make this change if what I'm suggesting is unclear.

src/ViewModels/WebMapServiceItemViewModel.js
kgrochow: We had only one of each case in the initialization file that were resolved away pretty quickly so it is currently not necessary.  We can discuss if we need the extra functionality.  The best way to do this going forward is probably to get out of this business altogether and point all the data set links to ckan pages and let people do what they want there.  That has a lot more flexibility than we will ever probably want to have.

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: Doesn't matter much to me.  arcgis/esri - rest/mapServer/mapservice seems to get used interchangeably.  data.gov.au has 2 formats and neither are named a combination of these.  Data.gov has Esri REST being most popular.  esri-mapService or esri-mapServer?

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: There was a reference back to the init file which kind of took us away from the actual issue.  I think the init file legendUrl field is something we should avoid.  How it's done in the arcgis....js file can be whatever is consistent and easy.

src/ViewModels/CkanGroupViewModel.js
kgrochow: We must be reading it differently. I think it's fine.  It tracks a property called wmsParameters (from the init file most likely) and appends it the wms service items that it creates from the ckan search.

src/ViewModels/CkanGroupViewModel.js
kgrochow: OK

src/ViewModels/CkanGroupViewModel.js
kgrochow: Sounds cleaner.

src/ViewModels/ArcGisMapServerItemViewModel.js
kring: Well I'm reviewing the code, so I was just suggesting changes to the code (to make it consistent with the WMS item, specifically). :)  No argument that we should get out of the legendUrl business, just like we should get out of the init file business entirely.

src/ViewModels/ArcGisMapServerItemViewModel.js
kring: Perhaps I'm being pedantic, but I think it's important to use an accurate name because it's pretty likely that NM will support very different Esri REST services in the future.  For example, their Feature Service returns vector features, so it's a bit like WFS, as opposed to a Map Service which is similar to WMS.  We'd handle that with a different view-model type, even though that one could legitimately be called `esri-rest` as well.

I had called it `esri-mapServer`, because the URL pretty much always contains "MapServer".  But it seems like `esri-mapService` would be more correct.

src/ViewModels/CkanGroupViewModel.js
kring: "If this property is undefined or if {@link CkanGroupViewModel#filterByWmsGetCapabilities} then no parameters are added"

That's the part that is not true.  `filterByWmsGetCapabilities` has no impact on whether or not parameters are added.

src/ViewModels/WebMapServiceItemViewModel.js
kring: Roger.  Again, I'm ok with removing functionality we don't currently need, but we should be consistent about it.  ArcGIS catalog items should basically work the same way as WMS catalog items.

src/ViewModels/ArcGisMapServerItemViewModel.js
kgrochow: Support is in and hopefully we'll never need it.

src/ViewModels/CkanGroupViewModel.js
kgrochow: Actually here it is, since we should only be doing filterResultsByGetCapabilities for wms resources.

kgrochow:
OK that should address the issues mentioned.
- dataUrlType is there and can have a value of "none" now to fix things like the Earthquake data url display.
- it's esri-mapService
- legendUrl is supported for esri but never used in init_nm
- simpler logic on multiple query loads
- using new domain name for ckan server site  http://ckan-nm.nicta.com.au

kgrochow:
Updated to integrate the major UI changes.

PR #487, opened 2015-01-21
Adjusted typography and colour to improve visual consistency, accessibility

philipgrimmett:


src/Styles/MenuBar.less
```diff
@@ -3,17 +3,24 @@
     right: 25px;
     top: 20px;
     height: 30px;
+    border: 1px solid rgba(255,255,255,0.1);
```
kring, line 6: Are you sure you want `rgba(255,255,255,0.1)` rather than `rgba(25,25,25,1.0)`?  On a black background they're the same, but I think the latter will look better when the borders are over a multicolored background like the Earth.

kring, line 6: Sorry, they're not the same at all.  Something more like `rgba(60,60,60,0.8)` would be good I think.

kring, line 6: You know, nevermind.  I have no idea what I'm talking about. :)

philipgrimmett, line 6: It's a fairly subtle difference, but I'd consider it noticeable:

screen shot 2015-01-21 at 4 25 30 pm
screen shot 2015-01-21 at 4 27 16 pm

The border itself isn't essential – that said, i'm inclined to keep it matching the internal menu bar dividers while it's there.

PR #491, opened 2015-01-27
Fix state issue with viewerMode causing problems for devices without 3d

Layoric:
Viewer was switching to leaflet when device didn't support WebGL, but viewerMode was left with default value causing issues if changeViewer was called.


kring:
Thanks @Layoric.  Coincidentally, I fixed this in the new UI (`phil-ui`) branch today.  You can try it out here:
http://nationalmap.research.nicta.com.au/demo

If it's alright with you, I probably won't merge your pull request because the new UI is imminent and this will just create a merge conflict with that.

Layoric:
Ahh no worries. Sounds good.

PR #492, opened 2015-01-27
new state data files

kgrochow:
connects to all their wms service at 'thelist'.  base maps next.


kgrochow:
This was tested yesterday before and during the demo and it's not part of the core stuff, so I'm merging so it's available on the next release.

PR #500, opened 2015-01-31
Add the compass

kring:
image

Features:
- Drag the outer ring or the N to rotate the camera around the point in the center of the screen.  If the globe isn't in the center of the screen the camera is rotated in place instead.
- Click and drag the inner gyroscope to free orbit around the point at the center of the screen.  Again, if there's no point, the camera orbits in place.
- Double click anywhere to re-orient the camera to look straight down at the globe with North up.

(Apologies to Phil for totally messing up his design, but I wanted to get this in quickly so we can roll it out on Monday.  We can improve it later.)


kring:
You can try this out here:
http://nationalmap.research.nicta.com.au/compass

kring:
I'm going to merge this, deploy it to staging, and do a thorough test in a bunch of different browsers before we deploy it to production this afternoon.

Issue #519, opened 2015-02-06
Add support for querying feature info from Esri Map/Image servers


kring:
Esri ImageServers, at least at the 10.04 version currently used by GA, do not do a good job of implementing the WMS GetFeatureInfo service.  For example, this query returns an internal server error:
http://www.ga.gov.au/gisimg/services/topography/dem_s_1s/ImageServer/WMSServer?service=WMS&version=1.1.1&request=GetFeatureInfo&layers=0&query_layers=0&srs=EPSG%3A3857&bbox=10018754.17139462%2C-10018754.17139462%2C15028131.257091936%2C-5009377.085697312&x=248&y=-173&width=256&height=256&info_format=application%2Fjson

However, Esri's native `identify` service returns good results, and quickly:
http://www.ga.gov.au/gisimg/rest/services/topography/dem_s_1s/ImageServer/identify?geometry=149.575%2C-33.43&geometryType=esriGeometryPoint&mosaicRule=&pixelSize=&f=pjson

National Map should be able to use the Esri `identify` service when users click on a point on the map.

PR #526, opened 2015-02-09
fix population/nexis server

### LINKED ISSUE #472: Update to new NEXIS server
@philipgrimmett noticed that the National Exposure Informatiom System 2012 has a layer called "Warning".  When enabled, you get this:
image
As far as I can tell, the map referenced in the URL in the screenshot doesn't actually exist.  Here's the ArcGIS services directory for the server, though:
http://www.ga.gov.au/gis/rest/services/hazards

We should figure out what we should actually be using and switch over to it.  Assuming the service will actually be shut down in December 2014 like the image says, this needs to happen soon.

kgrochow:
Fixes #472 

The single nexis wms server is gone and has been replaced with several esri rest servers and the occasional wms server at http://www.ga.gov.au/gis/rest/services/hazards
- I fixed population under Social and Economic as well as I could and earthquake was fixed a while ago
- I removed the nexis wms provider since it no longer exists.
- Of the remaining nexis data sets
  - We have a newer catchments one from data.gov.au
  - Agricultural commodities and buildings require esri featureInfo and more esri map selection support to be usable

### DIFF
--- a/public/init_nm.json
+++ b/public/init_nm.json
@@ -1185,8 +1185,8 @@
                                 "-11.05104999964455"
                             ],
                             "type": "wms",
-                            "url": "http://www.ga.gov.au/gis/services/hazards/NEXIS_2012_WMS/MapServer/WMSServer",
-                            "layers": "7",
+                            "url": "http://www.ga.gov.au/gis/services/hazards/NEXIS_National_Exposure_Information_System_Population_Density_Exposure/MapServer/WMSServer",
+                            "layers": "0,1,2,3,4",
                             "getFeatureInfoAsGeoJson": false
                         },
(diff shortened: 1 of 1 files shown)


kring:
CC #519

kring:
I'm reopening this to see if we can add the Agricultural Commodities and Buildings layers now that we have better ArcGIS MapServer support (see #587).

stevage:
Where are these layers exactly? http://services.ga.gov.au/gis/rest/services/NEXIS_Agricultural_Exposure/MapServer and 

http://services.ga.gov.au/gis/rest/services/NEXIS_Building_Exposure/MapServer ?

PR #528, opened 2015-02-10
Use new topographic basemap from GA

kring:
This is a simple National Map change and some more complicated Cesium changes.

In National Map, it is now optional to treat a 404 as an error when requesting a tile.  The default value is false - ignore 404s.  This not only lets us use the new topo map without popping up a "dataset failed" message to the user, but it also eliminates some spurious messages like this when using other Esri datasets (for example, several in the Land category).

In Cesium, pass the exception, if any, through to the `ImageryProvider` `errorEvent`.  This allows us to determine the HTTP status code of the failure in event handlers (as described above).

Also in Cesium, fix a problem with imagery loading that would cause half the globe to be missing (the important half too!) when the new topo map was used as the default basemap  or when switching from Leaflet to Cesium while the topo map was active.


kring:
@philipgrimmett it may be worth resnapping the Australian Topography, Australian Hydrography, and Natural Earth II thumbnails, because they all look a bit different after this and pull request and other recent changes.

PR #529, opened 2015-02-11
Tools panel

kgrochow:
Added a tools panel that is invoked by #tools on the url which holds our non-enduser system/dev tools.

Currently holds our old tile caching and ckan exporting code as well as a new tool to export a serialized version of the current data menu.

### DIFF
--- a/src/Models/Application.js
+++ b/src/Models/Application.js
@@ -242,6 +242,7 @@ Application.prototype.addInitSource = function(initSource) {
 };
 
 var initSourceNameRegex = /\b(\w+)\b/i;
+var reservedProperties = ['tools'];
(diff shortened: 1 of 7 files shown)


kring:
This looks good.  I made some tweaks:
- Show the Tools menu item on the left rather than the right, so the callout on the Maps panel still points at the Maps button.
- Turn on/off the tools menu item via `userProperties` rather than custom handling.  This does require that we use `tools=1` (or similar) in the URL now.
- Make sure that when the `updateApplicationUrl` promise rejects, the user is informed.  Previously (before this pull request, even), bad values in the hash after startup would silently fail.

Issue #55, opened 2014-06-23
Add more to share url - camera angle, mode, base layer, etc.


kgrochow:
The stored camera params don't take tilt and view direction into account

stevage:
So it looks like camera angle is now supported, but base layer isn't. (What's "mode"? 2d/3d?)

Not sure whether to close this issue and open new one, or edit the title.

PR #533, opened 2015-02-12
Extent-related fixes

### LINKED ISSUE #282: Poor computation of extent from 3D view
When switching between 2D/3D modes, we determine the extent in the mode we're leaving in order to try to view the same extent in the new mode.  This works pretty well going from 2D to 3D, but poorly when going from 3D to 2D.  It used to work pretty well; I think it broke with the Cesium FOV changes.

kring:
- Improve the computation of view extent in Cesium.
- Tell Leaflet to animate zooming (though it ignores it when zooming far away)
- Make Leaflet zooming work when the extent crosses the international date line.
- Remove some redundant code in AusGlobeViewer.
- When sharing, share the actual Cesium camera parameters in addition to the estimated bounding box.  This way the view is recreated exactly when possible.

Fixes #282
Fixes part of #55.

### DIFF
--- a/src/Models/Application.js
+++ b/src/Models/Application.js
@@ -225,7 +225,7 @@ Application.prototype.addInitSource = function(initSource) {
                 position: initSource.camera.position,
                 direction: initSource.camera.direction,
                 up: initSource.camera.up
-            }
+            };
         } else {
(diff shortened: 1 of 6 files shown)


kring:
Fixes #282

PR #544, opened 2015-02-18
Add support for Markdown image references to layer description sanitiser...

axman6:
We may need to embed custodian logos in the descriptions of some layers so thought we'd add support. Code is heavily based on the link replacement code so I'm not sure if there's a better way or not, but this seems to work.


kring:
Looks good, thanks, Alex!

PR #548, opened 2015-02-25
Show KML datasets from data.gov.au with #dgakml.

kring:
This is just a prototype and maybe some demo fodder for now, but it works reasonably well.

In addition to a new init file and changes to our CKAN support to support KML datasets in CKAN, this pull request also updates Cesium to include the latest changes from the `kml` branch.

### DIFF
--- a/public/config.json
+++ b/public/config.json
@@ -3,7 +3,9 @@
         "nicta.com.au",
         "gov.au",
         "csiro.au",
-        "arcgis.com"
+        "arcgis.com",
+        "argo.jcommops.org",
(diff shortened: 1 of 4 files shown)


kring:
Try it here:
http://nationalmap.research.nicta.com.au/ckan/#dgakml

Open up the "Data.gov.au KML" group at the bottom of the catalogue.  It takes a few seconds because there are a lot of KML datasets.

kring:
I should mention the KMLs in the first few groups don't actually work, because they're loaded from domains that are not in our proxy whitelist.

PR #593, opened 2015-03-24
Improvements to performance check tool.

kring:
- Add support for testing ArcGIS servers.
- Clear messages about tile request success/failure.
- When a tile request returns an error, inform the user of the HTTP status code and the problematic URL.

image


kring:
Ok this is ready now.

PR #600, opened 2015-03-25
Performance test tweaks

stevage:
I tried to make the results of the performance test a bit cleaner and more relevant. Ditched minimum tile response time (not useful), added number of slow tile requests (quite interesting...often there are only 2-3 slow tiles out of hundreds of requests.)


kring:
Thanks @stevage!

Issue #607, opened 2015-03-27
Show WMS names, not just titles


stevage:
Probably the real cause of issue #605 is that we only show the `title` of each WMS layer when fetching a getCapabilities request, not the `name`:

![screenshot 2015-03-27 12 14 50](https://cloud.githubusercontent.com/assets/678344/6860362/1fd94a18-d47b-11e4-9790-a301a0d67cd6.png)

Here we see dozens of indistinguishable layers "mass_concentration_of_chlorophyll_in_sea_...", because that's the title of each. Yet they each have a different `name`, like the `dR_MIM_n_obs` seen here, and a meaningful `abstract`.

Perhaps we need an option to indicate that a WMS catalogue's `name` properties should be shown instead of its `title` (`WebMapServiceCatalogGroup.js.createWmsDataSource`).

kring:
Fixed by #608.

PR #608, opened 2015-03-27
Allow showing names, abstracts of WMS layers as per #607.

stevage:
This didn't turn out to be super useful for NEII but perhaps some other WMS catalog in the future.

### DIFF
--- a/src/Models/WebMapServiceCatalogGroup.js
+++ b/src/Models/WebMapServiceCatalogGroup.js
@@ -59,6 +59,12 @@ var WebMapServiceCatalogGroup = function(application) {
      */
     this.blacklist = undefined;
 
+    /**
+     * Specifies whether to use each WMS layer's "title" (default), "name", or "abstract" as the primary title.
+     * @type {String}
+     */
+    this.titlefield = 'title';
+
     knockout.track(this, ['url', 'dataCustodian', 'parameters', 'blacklist']);
 };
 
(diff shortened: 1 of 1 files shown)


src/Models/WebMapServiceCatalogGroup.js
```diff
@@ -59,6 +59,12 @@ var WebMapServiceCatalogGroup = function(application) {
      */
     this.blacklist = undefined;
 
+    /**
+     * Specifies whether to use each WMS layer's "title" (default), "name", or "abstract" as the primary title.
+     * @type {String}
+     */
+    this.titlefield = 'title';
+
     knockout.track(this, ['url', 'dataCustodian', 'parameters', 'blacklist']);
```
kring: 'titleField` should be added her as well so that it becomes a knockout observable and we can subscribe to its change notification.

It should also be added to the list of properties returned by `_getValuesThatInfluenceLoad` because it does.

src/Models/WebMapServiceCatalogGroup.js
```diff
@@ -59,6 +59,12 @@ var WebMapServiceCatalogGroup = function(application) {
      */
     this.blacklist = undefined;
 
+    /**
+     * Specifies whether to use each WMS layer's "title" (default), "name", or "abstract" as the primary title.
+     * @type {String}
+     */
+    this.titlefield = 'title';
```
kring: Sorry for the nitpick, but can you make this camelCase: `titleField`?  Also, start the doc with "Gets or sets" for a read/write property like this, or just "Gets" if it's read-only.

src/Models/WebMapServiceCatalogGroup.js
stevage: thanks, I was hoping someone would tell me whether or not I needed to do that. should all fields be observable? any reason to not?

src/Models/WebMapServiceCatalogGroup.js
stevage: thanks, you're right.

src/Models/WebMapServiceCatalogGroup.js
kring: Yes, all public, writable properties should be observable, generally.  Sometimes you can make the case that it's not really useful.  The `preserveOrder` property in your other pull request, for example, only has an effect at a very specific time, so being notified when it changes doesn't seem very useful.  It doesn't really ever hurt to make something observable though.

kring:
Thanks @stevage.  For future reference, you should add a note to the pull request when you've made changes after a review, because GitHub doesn't notify of new commits.  Merging.

PR #610, opened 2015-03-30
Add "preserveOrder" flag to CatalogGroup.

stevage:
Allows a group to be unsorted, which gives better results when you're hand-crafting the entries.


kring:
I removed the `sortItemsOnLoad` property because I have no idea what I was thinking when I added that.  Your new `preserveOrder` property is better.  Thanks!

PR #613, opened 2015-03-31
Perf tool tweaks

stevage:
I made zoom level a range, rather than 0-X, and capped the number of tiles requested per level at 50. (I'm not sure that's the best thing to do in every situation but it's better than requesting thousands of tiles...)

Also a progress bar.

### DIFF
--- a/src/Styles/ToolsPanel.less
+++ b/src/Styles/ToolsPanel.less
@@ -162,3 +162,11 @@ p.tools-content-new-section {
 .tools-tip strong {
     font-weight: bold;
 }
+.tools-loading-message {
+  color: gray;
+}
+.tools-loading-message-bar {
+  background: gray;
+  display: block;
+  height: 10px;
+}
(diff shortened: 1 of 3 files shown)


kring:
This looks good to me.  @kgrochow do you want to review and merge?

kgrochow:
Why would we limit the requests to 50?  Part of the testing is to see if they fall over at scale.

kring:
Yes, Steve and I talked about that and he removed it.  You should join us on HipChat to be part of these conversations. :)

stevage:
Yeah, we had a debate. For me (as someone working directly with data
custodians), the value of the performance tool is not a pass/fail bludgeon,
but a specific instrument with precise information about performance
different zoom levels. Requesting every tile at every zoom level means that
testing zoom level 10, for instance, is impractical. And more generally,
testing a large number of datasets is much easier if you can cap the total
number of tiles that will be gathered for any individual dataset.

As you say, _part_ of the testing is bulk download. Another part is average
performance at specific zoom levels.

I'll disable for now until I get time to add a GUI element.

PR #615, opened 2015-04-01
Add support for time-dynamic WMS datasets

### LINKED ISSUE #426: Time-dynamic data items don't show timeline until they're clicked
Previously, if you dragged in a time-dynamic CSV file, for example, the camera would immediately zoom to it and the timeline would appear showing the relevant time period.  In the refactor branch, the timeline doesn't appear and the camera doesn't zoom until you click on the item's label.  Also, the timeline doesn't disappear when we disable the time-dynamic data item.

kring:
To play with this, add `#test` to your URL and enable Test Data -> Time Dynamic -> Chlorophyll-a concentration (monthly).  This is an eReefs layer.

In the init file, you can now specify a list of intervals and the `time` parameter to send to the WMS server during each of the intervals.  When such a dataset is enabled, National Map displays the time controls and adjusts the displayed interval to match the dataset's intervals.  The tiles for the next interval (whether you're animating forward or backward) are pre-loaded so the switch is smooth if the tiles download reasonably fast and you're not jumping around in time.

Still to do in separate pull requests: improve the timeline/animation controls, automatically populate the `intervals` property from WMS GetCapabilities (instead of needing to hard code it into the init file), improve handling of infinite-ish intervals.

Fixes #426


kring:
This pull request also refactors `ImageryLayerCatalogItem` to eliminate some code duplication and to simplify writing catalog items for Cesium `ImageryProviders` (and having them work in both Cesium and Leaflet).

PR #618, opened 2015-04-02
Fix WFS on Safari and IE10.

### LINKED ISSUE #584: WFS appears not to work in Safari
We are having trouble using WFS layers in Safari on OSX (both Mavericks and Yosemite). Using the following WFS:

http://www.ga.gov.au/gis/services/topography/Australian_Topography/MapServer/WFSServer?service=WFS&version=1.1.0&request=GetCapabilities

When you enable any of the WFS layers you get error messages like the following (the first one only seems to happen when you enable Aircraft_Facilities):

```
TypeError: null is not an object (evaluating 'xml.documentElement') 
TypeError: undefined is not an object (evaluating 'gmlNode.children.length')
TypeError: undefined is not an object (evaluating 'e.children.length')
```

kring:
Fixes #584.

### DIFF
--- a/src/Map/gmlToGeoJson.js
+++ b/src/Map/gmlToGeoJson.js
@@ -61,8 +61,8 @@ var gmlSimpleFeatureNames = [
 function getGmlPropertiesRecursively(gmlNode, properties) {
     var isSingleValue = true;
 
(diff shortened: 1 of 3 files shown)


kring:
This pull request also makes feature picking work again in IE9.

PR #619, opened 2015-04-03
Trim CKAN catalog error message, add exception string to output.

stevage:


kring:
I tweaked this a bit to handle HTTP errors as well.  Previously it would just print "Exception: undefined" when the server returned a 404, for example.  And I made the styling consistent between this and render loop errors.

@kgrochow since Steve and I have both touched this now, can you review and merge?

PR #626, opened 2015-04-13
Small UI improvements

### LINKED ISSUE #501: Search input box scrolls with results
The search input box on the Search tab scrolls with the results, so if you scroll down it becomes invisible.

image

### LINKED ISSUE #605: Very long layer names block access to info button
nm long layer names

A very long layer name means you can't see the info button, which means you also can't see the full layer name (to be able to distinguish it from others).

A couple of possible solutions:
1. CSS style changes. Set a `max-width: 260px; word-wrap: break-word;` on `.data-catalog-item-label`. (Probably something smarter than hardcoding 260px...)

nm-fix-maybe

`2`. Change the info button so it's either left-justified, or otherwise doesn't get pushed off to the right.
3. Tweak UX to remove the info button altogether. IMVHO, clicking on the layer should show the info dialogue, with some other mechanism to zoom to it (possibly within the info dialogue).

### LINKED ISSUE #622: Search field loses focus when 'clear' button is clicked
Proposed solution: retain focus after clearing the search field

### LINKED ISSUE #625: Input events aren't passed through to the underlying map while 'maps' menu is open
**Suggested solution**

While 'Maps' dropdown is open, accept the following input events:
- mouse scroll (zoom)
- click+drag (pan)

Close 'Maps' dropdown on:
- single/double mouse click outside of dropdown

kring:
- Restore focus  to the search box after clicking the (x) to clear the search  text.
- Improve how the Maps panel closes when clicking outside.  It no longer eats attempts to interact with the map.

Fixes #622 
Fixes #625 
Fixes #605 
Fixes #501

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@ Change Log
 * Made numerous improvements to the server performance check tool, accessed by appending `#tools=1` to the URL and clicking the Tools button.
 * Added `preserveOrder` property to catalogue groups.  When set, the group's items will not be sorted by name.
 * Added `titleField` property to WMS catalogue items to specify whether the WMS layer's title (default), name, or abstract is displayed in the catalogue.
-
+* Clicking the clear (x) button on the search panel now returns focus to the search box.
 
 ### 2015-03-26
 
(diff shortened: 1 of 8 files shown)


kring:
One more:
- Very long labels in the Data Catalogue and Now Viewing tab are now wrapped much more gracefully.

Fixes #605

kring:
- Don't let search box scroll with results.
  Fixes #501

PR #631, opened 2015-04-15
Add support for WMS MinScaleDenominator and Esri maxScale

kring:
This prevents layers from disappearing when you zoom in too close.  Instead, we continue to use the lower-detail tiles and the layer just gets blurry.


kring:
Oops, don't merge this yet.  Leaflet problems.

Issue #569, opened 2015-03-03
Get clicked point feature info for all enabled layers


meh9:
When several layers are enabled and a point on the map is clicked, get the feature info for that point for all the enabled layers instead of just one of them. Then display these in the feature info popup as different tabs, or initially minimised expandable stack (like the Now Viewing tab), or something like that.

Probably a good idea to expand the info in real-time as they are returned, so we don't have to wait for all the info requests to finish before showing results.

chloeleichen:
Anyone working on this issue? Are we going ahead with @meh9 's recommendation?

stevage:
Yeah, this is a really important feature that is required for the new Northern Australia map project. The use case there is that a user will click on a point and we want to show them all native title, indigenous land use agreements etc that affect it. Each is present as a WMS layer.

I have a vague recollection that implementing this isn't trivial, but unfortunately I don't remember why. It may be that the feature picking happens in Cesium, so would require changes at that level. But maybe I'm misremembering - so please do dive in and see if you can make it work.

The final result will need to be fairly flexible - some projects will want to immediately display properties stacked, others will want to concertina them etc. The UX that Mitch and I discussed for Northern Australia would be something like:
- "Native Title"
  - Name: ...
  - _Click for more info..._
- "Indigenous Land Use Agreement"
  - Name: ...
  - _Click for more info..._

Where clicking expands that section (and collapses any other expanded sections).

chloeleichen:
@kring would you please have a look at this? just dump all the html into the template and I will update the template and styles based on @philipgrimmett 's design.

PR #635, opened 2015-04-16
Region mapping updates

kgrochow:
To make the abs merge a bit more manageable, this is the updates to the CsvCatalogItem and everything below it.  New stuff
- region mapping based on string region ids (example in countries.csv)
- time based region mapping (example in droughts.csv)
- table styling in init file (examples in test.json and in DataTable docs)
- general refactor of TableDataSource, DataTable, and DataVariable
- Improved legend visuals to go with the now viewing panel
- added country regions
- added country region test files
- cleaned up existing test csv files (date formats mainly)


src/Models/CsvCatalogItem.js
```diff
@@ -59,20 +65,43 @@ var CsvCatalogItem = function(application, url) {
...
+    this.opacity = 0.6;
 
-    knockout.track(this, ['url', 'data', 'dataSourceUrl', 'colorByValue']);
+    knockout.track(this, ['url', 'data', 'tableStyle', 'opacity', '_clock']);
+
+    knockout.getObservable(this, 'opacity').subscribe(function(newValue) {
+        updateOpacity(this);
+    }, this);
+
+    delete this.__knockoutObservables.clock;
```
kring: Maybe I'm missing something, but I don't see why the base class `clock` property needs to be replaced.  Its replacement below looks functionally identical.

kring: I guess this was copied from `ImageryLayerCatalogItem`, but the clock property's getter there has special logic that is absent here.

src/Models/CsvCatalogItem.js
```diff
@@ -239,14 +301,14 @@ CsvCatalogItem.prototype._showInCesium = function() {
             return when(featurePromise, function(results) {
                 if (defined(results)) {
```
kring: Should add `&& results.length > 0` here, because returning an empty array is allowed (it'll happen when a layer is pickable but there are no features until the picked position).

kring: Actually, wouldn't a loop be better?  Currently NM only shows the first picked feature, so updating the others is arguably unnecessary, but this is likely to change soon.  See #569.

src/Models/CsvCatalogItem.js
kgrochow: The time based datasets need to create their own clock settings that are used to update the main clock just like in imagery layer.  The actual creation of the clock is done in createRegionMappingClock which is called after the data is loaded and the setting can be determined.

kgrochow: I'll look at whether how it's being done in imageryLayer now can be used, but there were problems with the previous iteration of the clock in that.

---

From: Keith Grochow
Sent: Friday, 17 April 2015 11:02 AM
To: NICTA/nationalmap; NICTA/nationalmap
Subject: RE: [nationalmap] Region mapping updates (#635)

The time based datasets need to create their own clock settings that are used to update the main clock just like in imagery layer.  The actual creation of the clock is done in createRegionMappingClock which is called after the data is loaded and the setting can be determined.

src/Models/CsvCatalogItem.js
kgrochow: OK, I've addressed the comments and everything seems to be working fine.

kgrochow: I skipped the adding the loop until we have the multi-feature picking since it's all pulled from in memory so no real overhead.

kring:
> I skipped the adding the loop until we have the multi-feature picking since it's all pulled from in memory so no real overhead.

I added the loop myself yesterday.

PR #639, opened 2015-04-17
Support disclaimers, licenses and updates in catalog item info pane

axman6:
We need some more fine grain control of what information is available in the info pane for layers in the AREMI project. These changes meet our needs so we don't have to stuff everything into the description field. It would be good to have the UX team take a look and give their opinions on this.

### DIFF
--- a/src/Models/CatalogMember.js
+++ b/src/Models/CatalogMember.js
@@ -48,7 +48,7 @@ var CatalogMember = function(application) {
      */
     this.isUserSupplied = true;
 
-    knockout.track(this, ['name', 'description', 'isUserSupplied']);
+    knockout.track(this, ['name', 'description', 'disclaimer', 'license', 'updates', 'isUserSupplied']);
 };
 
 defineProperties(CatalogMember.prototype, {
--- a/src/Models/CatalogMember.js
+++ b/src/Models/CatalogMember.js
@@ -40,6 +40,15 @@ var CatalogMember = function(application) {
     this.description = '';
 
     /**
+    * An array of section titles and contents for display in the layer info panel.
+    * In future this may replace 'description' above. Content will be rendered using sanitizeHtml.
+    * This property is observable.
+    * @type {{name:string, content:string}}
+    * @default []
+    */
+    this.info = [];
+
+    /**
      * Gets or sets a value indicating whether this member was supplied by the user rather than loaded from one of the
      * {@link Application#initSources}.  User-supplied members must be serialized completely when, for example,
      * serializing enabled members for sharing.  This property is observable.
--- a/src/Models/CatalogMember.js
+++ b/src/Models/CatalogMember.js
@@ -48,7 +48,7 @@ var CatalogMember = function(application) {
      */
     this.isUserSupplied = true;
 
-    knockout.track(this, ['name', 'description', 'disclaimer', 'license', 'updates', 'isUserSupplied']);
+    knockout.track(this, ['name', 'info', 'description', 'isUserSupplied']);
 };
 
 defineProperties(CatalogMember.prototype, {
--- a/src/Models/CatalogMember.js
+++ b/src/Models/CatalogMember.js
@@ -40,10 +40,13 @@ var CatalogMember = function(application) {
     this.description = '';
 
     /**
-    * An array of section titles and contents for display in the layer info panel.
-    * In future this may replace 'description' above. Content will be rendered using sanitizeHtml.
+    * Gets or sets the array of section titles and contents for display in the layer info panel.
+    * In future this may replace 'description' above - this list should not contain
+    * sections named 'description' or 'Description' if the 'description' property
+    * is also set as both will be displayed.
+    * Content will be rendered using sanitizedHtml.
     * This property is observable.
-    * @type {{name:string, content:string}}
+    * @type [{name:string, content:string}]
     * @default []
     */
     this.info = [];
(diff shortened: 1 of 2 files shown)


kring:
What if we add a property like this to the init file:

```
"info": [
    {
        "name": "Disclaimer",
        "content": "blah blah blah <strong>blah</strong>"
    },
    {
        "name": "Data Updates",
        "content": "This is so up-to-date you can't even handle it!"
    },
    {
        "name": "License",
        "content": "CCBY or whatever"
    }
]
```

Seems better than baking-in the ones AREMI needs, right?  And not really much harder to do.

axman6:
Seems like a good alternative to me (though the changes would be more difficult than I'd know how to do in JS). One thing to note is that in our change the disclaimer is shown in a `strong` tag with red text; we probably want to keep the ability to do that (the simplest would be to add an optional `style` field to the info objects so we can put in arbitrary CSS, but I'm sure there are really good reasons to not do that). @Meh9 thoughts?

kring:
> One thing to note is that in our change the disclaimer is shown in a strong tag with red text

If `santizedHtml` took arbitrary HTML instead of the hacked up markdowny thing, that would solve the problem, right?

axman6:
Yep, sounds reasonable to me.

meh9:
@kring your solution looks fine to me - very flexible. @axman6 has covered the other things.

The init file will grow, but if it was possible to include arbitrary JSON fragments at compile time (as we discussed regarding splitting out individual layer configs into separate files yesterday) then this could/would be part of that externally managed file, which would keep the size of the init file index down.

kring:
ok, I'll probably work on this next week (assuming you don't have an urgent need for it).

axman6:
It would be good to have it in before April 30 (preferably before then so we have time to get it deployed by GA if we decide to do so), but if it's not we'll go with my changes until your solution is in place.

kring:
Well you could always learn yourself from JavaScript.  Should be easy for a Haskeller, at least once you shake the dirty feeling. :)

I want to finish up the Terria split first.  If that goes well, April 30 shouldn't be too unrealistic.  The Terria split itself will cause you guys a bit of pain though, probably.

axman6:
Can this change be made in the nationalmap repo and then brought across into terria? That would help avoid pain at a pretty critical time in our development.

kring:
You know I'm only one person right? :)

axman6:
I do, I wish we had a whole team of Kevins to do all the things. I've been trying to do what I can, but JS is gross and magical and there's only so many showers I can take in a day. =)

kring:
Read JavaScript: The Good Parts.  It's short, you can probably finish it in a couple days, and then the magic will be gone.  It's still gross, but we can work with gross.

hilarycinis:
hey - sorry @axman6 only just saw this thread and it was closed before I could get into it. If you need ux or ui design eyes, can you @ me so I can catch it early? Is this on staging or production yet?

meh9:
Hi @hilarycinis, it's only on our local branches yet, we can probably set something up for you to have a look at.

However, at this point we are very pressed for time for the milestone, so currently we are simply hacking things in and will go back and tweak with your input later.

hilarycinis:
Ok. Avoid red text though? Looks like an error or warning. Ta :)

Hilary Cinis
Principal UX & Designer
UX & Design Manager
NICTA

On 20 Apr 2015 9:01 am, Mats H notifications@github.com wrote:

Hi @hilarycinishttps://github.com/hilarycinis, it's only on our local branches yet, we can probably set something up for you to have a look at.

However, at this point we are very pressed for time for the milestone, so currently we are simply hacking things in and will go back and tweak with your input later.

—
Reply to this email directly or view it on GitHubhttps://github.com/NICTA/nationalmap/pull/639#issuecomment-94321527.

---

The information in this e-mail may be confidential and subject to legal professional privilege and/or copyright. National ICT Australia Limited accepts no liability for any damage caused by this email or its attachments.

axman6:
The red text is a warning though - it's there to clearly point out vital information about the layer that the user needs to know tho to use it properly.

hilarycinis:
ok then :)

axman6:
@kring I'm going to have a go at implementing the changes you've suggested, it looks like it should be relatively easy using knockoutjs. My only issue is how to deal with both the old and new methods of specifying descriptions etc. It should't be too difficult to write a jq script that converts the old style into the new I guess.

axman6:
The new commit should implement all the changes we want, while support the old style config for descriptions. Let me know if there's anything more you'd like to see changed before it can be merged @kring.

kring:
Thanks @axman6, this looks good!  I'm going to tweak it a bit before merging, probably later today.

kring:
Actually I'll tell you what I have in mind and you can do it instead if you're up for it.

src/Models/CatalogMember.js
```diff
@@ -48,7 +48,7 @@ var CatalogMember = function(application) {
      */
     this.isUserSupplied = true;
 
-    knockout.track(this, ['name', 'description', 'isUserSupplied']);
+    knockout.track(this, ['name', 'info', 'description', 'isUserSupplied']);
```
kring, line 63: `this.info` should be defined above as well.  Doing this _may_ improve the browser's ability to optimize the code (not that performance is a concern for this code), but more importantly it gives us a place to put the property's documentation.

src/Views/CatalogItemInfo.html
```diff
@@ -38,18 +38,32 @@
...
+                <h2>Description</h2>
                 <div class="catalog-item-info-description" data-bind="sanitizedHtml: catalogItem.description"></div>
-                <!-- /ko -->
-                <!-- ko if: !catalogItem.description -->
+            </div>
+            <!-- /ko -->
+
+            <!-- ko if:
+            !catalogItem.description &&
+            (!catalogItem.info || !catalogItem.info.some(function(i){return (i.name === "Description" || i.name == "description")}))
```
kring: It might be good to move this complicated-ish logic to a computed property on the view-model instead of sticking it here in the view.  You can define a computed property with `knockout.defineProperty`.

kring:
I think that's it.  Thanks @axman6!

Issue #530, opened 2015-02-11
ckan previewer not working at data.gov.au


kgrochow:
I'm pretty sure it's out of date due to our url changes and they've pulled it for the time being.

kring:
This was fixed in #660.

PR #660, opened 2015-04-30
Update ckanext-cesiumpreview for CKAN 2.3

maxious:
Updates ckanext-cesiumpreview to use JSON config string.

Related to #530 

Default view extent is set to the dataset extent recorded in the
"spatial" field in CKAN if available.

WMS resources can use "wms_layer" field to specify a specific layer
to be included in the config and be loaded by default.

CKAN 2.3 includes Resource Views feature which allows adding of
a view including manually to CSV files when the data owner knows the
location columns are appropriate for use in spatial preview.


kring:
Thanks @maxious!  This looks good to me.  @kgrochow can you review and merge?

PR #663, opened 2015-05-01
Terriafy!

kring:
Make National Map use [TerriaJS/terriajs](https://github.com/TerriaJS/terriajs) via npm instead of embedding all of the code in his repository.


kgrochow:
Everything clones and builds fine from scratch up here.  Let's take the plunge.

PR #665, opened 2015-05-02
Allow loading init files in .hjson format.

stevage:
HJson (http://hjson.org/) is more user friendly (no quotes or commas),
allows inline comments, and is slightly smaller.

Quick explanation: the init/*.json files are our primary link with data sources, but they're relatively hard for us to maintain, very hard to data custodians to maintain, and don't allow comments, which makes it hard to document issues as we go.

Using hjson isn't a full solution, but it's a modest improvement, I think. There are other "JSON plus" formats which all work on the "JSON is not a great format for configuration files" problem, including [json5](json5.org). Hjson seems as good as any.

I opted for browser side processing so that you can modify the .hjson file and see the change straight away without rebuilding. It will also work better with a purely static deployment. The cost of the hjson library is more than made up for the reduction in size in the .json files themselves.

There's no lock-in risk as we can always trivially convert the .hjson back to .json.

### DIFF
--- a/lib/Models/Terria.js
+++ b/lib/Models/Terria.js
@@ -13,10 +13,11 @@ var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
 var FeatureDetection = require('terriajs-cesium/Source/Core/FeatureDetection');
 var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
 var loadJson = require('terriajs-cesium/Source/Core/loadJson');
+var loadText = require('terriajs-cesium/Source/Core/loadText');
 var queryToObject = require('terriajs-cesium/Source/Core/queryToObject');
 var Rectangle = require('terriajs-cesium/Source/Core/Rectangle');
 var when = require('terriajs-cesium/Source/ThirdParty/when');
-
+var Hjson = require('hjson');
 var CameraView = require('./CameraView');
 var Catalog = require('./Catalog');
 var corsProxy = require('../Core/corsProxy');
@@ -381,7 +382,7 @@ function interpretHash(hashProperties, userProperties, persistentInitSources, te
 }
 
 function generateInitializationUrl(url) {
-    if (url.toLowerCase().substring(url.length-5) !== '.json') {
+    if (!url.match(/\.h?json$/i)) {
         return 'init/' + url + '.json';
     }
     return url;
@@ -405,7 +406,17 @@ function loadInitSources(terria, initSources) {
 
 function loadInitSource(source) {
     if (typeof source === 'string') {
-        return loadJson(source).then(function(initSource) {
+        var p;
+        if (source.match(/\.json$/)) {
+            p = loadJson(source);
+        } else {
+            p = loadText(source, "Accept : 'application/json,application/hjson,*/*;q=0.01'")
+                .then(function (h) {
+                    return Hjson.parse(h);
+                });
+        }
+
+        return p.then(function(initSource) {
             initSource.isFromExternalFile = true;
             return initSource;
         }).otherwise(function() {
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "URIjs": "^1.15.0",
     "brfs": "^1.4.0",
+    "hjson": "^1.6",
     "html2canvas": "^0.5.0-alpha2",
     "javascript-natural-sort": "^0.7.1",
     "leaflet": "^0.7.3",


stevage:
Here's a sample of what .hjson looks like (wa.hjson): 

```
{
  homeCamera:
  {
    west: 112
    south: -34
    east: 128
    north: -25
  }
  services: []
  catalog:
  [
    {
      name: Western Australia (Landgate)
      type: group
      items:
      [
        {
          name: Locate Public
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-00913315481290556980-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Cultural and Society
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-03311849775732631692-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Environment
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-11920815285354965867-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Infrastructure and Utilities
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-14702451115530212229-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Property and Planning
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-17372459662889040700-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Resources
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-11493353092997567468-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Cultural and Society
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-03311849775732631692-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
        {
          name: Locate Pilbara
          dataCustodian: "[Landgate](http://slip.landgate.wa.gov.au/Pages/AboutUs.aspx)"
          url: https://mapsengine.google.com/09372590152434720789-09499575881840677463-4/wms/
          type: wms-getCapabilities
          clipToRectangle: true
          itemProperties:
          {
            ignoreUnknownTileErrors: true
            opacity: 0.8
            legendUrl: ""
          }
        }
      ]
    }
  ]
}
```

kring:
I have mixed feelings on this.

On the one hand, I also find it frustrating that JSON doesn't support comments and the fact that parsing breaks when there's a trailing comma is annoying as well.  The requirement for wrapping strings in quotes and separating properties and array elements with commas never bothered me as much, but ok, I can see some appeal.

On the other hand, I can think of some arguments against this:
- If we're worried about the size of the file, we'd gain more by using regular JSON and stripping out whitespace than by switching to HJson.  Minifying HJson in the same way is presumably harder because whitespace is significant.
- It's true there's no lock-in because we can easily convert to JSON, but it still might be a hassle in some scenarios.  For example, if @axman6 wants to write a Haskell tool to work with the init file, can he do that?  Probably not without writing a Haskell HJson library first, or transforming the file to regular JSON before running his program.
- Lots of tools (e.g. editors) that we or others might want to use with the init file don't know about HJson.  For example, telling Sublime Text that your example HJson file above is JSON results in a mess of red.  There's probably an HJson plugin for Sublime Text already (I didn't look), but since the format doesn't seem to be super popular yet there are going to be plenty of tools that don't know how to deal with it.

Any other opinions on this?  I guess my position is that I'm willing to merge it if you feel strongly that it will make the file easier to work with for Comms / data custodians / us.  But I'm not entirely convinced that will be the case.

meh9:
I'm inclined to agree with you @kring.

I'm all for improving our init file situation, but changing from a well-known format that has existing tools and good existing library support (`jq` etc, and all languages have a JSON lib) to something less universal doesn't sound appealing to me right now.

Specifically commenting on the no-comments-allowed of JSON, I haven't tried if NM is unhappy with it or not, but we could use the "comment" key trick - as long as NM isn't going to explode when there are keys present that it doesn't recognise then you can add a `"comment": "This is my comment about this thing."`. It's not fantastic, but it may alleviate not being able to make comments to some degree.

stevage:
> If we're worried about the size of the file, we'd gain more by using regular JSON and stripping out whitespace than by switching to HJson. 

The size argument is really a pre-emptive counterargument against the complaint of the size of the hjson library itself. :)

> Minifying HJson in the same way is presumably harder because whitespace is significant.

Fwiw (not much), only newlines are significant, so a minified HJson would almost always be smaller than a minified Json. (The latter saves 1 char per newline, but the former gains at least 2 for the quotes around the key, and probably another 2 around the value, and maybe 1 more for the comma.)

> For example, if @axman6 wants to write a Haskell tool to work with the init file, can he do that? Probably not without writing a Haskell HJson library first, or transforming the file to regular JSON before running his program.

If this is taking place in the command line with a full deployment, then that conversion is as simple as `node_modules/hjson/bin/hjson -c < file.hjson > file.json`. Or we could trigger that generation in gulp.

> Sublime Text that your example HJson file above is JSON results in a mess of red

Your point about lack of support is valid. For editing specifically, I find that YAML syntax highlighting is good enough. There is so little syntax left in hjson that it almost doesn't matter - well, for me anyway. Unhighlighted hjson is still much better than highlighted json (again, for me).

Would one of you be able to flesh out a bit more what the interoperability requirements of our init files are exactly? I was imagining that the only workflow is:
1. One of us edits init file
2. Terria loads it in browser and hence shows layers etc.

Where does the Haskell, jq etc fit into it?

And how would you feel about allowing .hjson files to be checked in, with .json generated at gulp time, or possibly checked in along side?

nahidakbar:
I think

1) I like normal json a lot more than what they describe at hjson.org. +1 vote for json.

2) Someone should talk to the hjson guys about xml.

3) People who would have difficulty with json would prefer excel to degraded json.

PR #668, opened 2015-05-04
Be smarter about finding the terriajs-cesium package.

kring:
Instead of assuming it's in `./node_modules/terriajs-cesium`, find the actual path using npm package resolution.  This should let allow it to be found correctly even when npm de-duping puts `terriajs-cesium` as a sibling of `terriajs` instead of a child.


kring:
This is ready now.  Unfortunately, due to some npm insanity (https://github.com/npm/npm/issues/6926), `npm install` of an app that uses a GitHub-version of TerriaJS may sometimes fail.  Running `npm install` again will fix it.  More details here:
https://github.com/TerriaJS/terriajs/wiki/Problems-and-Solutions

PR #677, opened 2015-05-08
Search all datasets, not just the loaded ones.

### LINKED ISSUE #658: Searching for layers in dynamic catalogs
Currently, when a user searches for a layer, it only finds layers that are either:
1. in a static catalog (a .json file); or
2. in a dynamic catalog that has been expanded (and hence, the list of layers pulled down)

I understand why this is the case. But I don't think it's really adequate for discovering layers. If a user wants to know if there is any dataset anywhere about, say, "health", they need to expand all the top level catalogs, then an unknown (and difficult to ascertain) number of second-level catalogs (potentially dozens). 

And of course they have no way of knowing that they need to do this, or what "Sorry, no catalogue items match your search query" really means.

Unfortunately, this is one of our most important use cases (user trying to find whether or not there exists a government geospatial dataset on a given topic), so we can't ignore it. Maybe it's even our most important use case.

I can think of a few solutions, but they're not exactly trivial:
1. Replace all dynamic catalogs with static ones that are frequently updated dynamically. (Downside: massive JSON download)
2. Have a separate search index (eg, a generated JSON file which is just layer names without most of the other metadata) which is frequently updated using all the dynamic catalogs (Downside: More work, could be tricky to implement so you can click on a result and then the main catalog finds its way to the layer you clicked on)
3. Create a web service with this pre-computed list (bonus: free analytics on what people are searching for)
4. Have a mechanism allowing the user to do a "super search", retrieving the contents of every single dynamic catalog prior to searching. (Downside: super slow, probably bad UX)

Number 2 or 3 seems the best at first glance.

kring:
Fixes #658 

This does the simplest possible thing: before searching a group, go load it.  It happens asynchronously, so a slow group won't keep the quick results from coming back, and the UI displays a "Searching..." message until all results are in.  There are smarter things we could do, but this is pretty good and really simple.

We currently have a broken server, the Tasmania "Planning Online" group, that takes about 25 seconds and then returns an error.  All the others take less than 5 seconds on my system.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### 1.0.7
 
 * `CatalogItemNameSearchProviderViewModel` now asynchronously loads groups so items in unloaded groups can be found, too.
+* Do not automatically fly to the first location when pressing Enter in the Search input box.
 
 ### 1.0.6
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### 1.0.7
+
+* `CatalogItemNameSearchProviderViewModel` now asynchronously loads groups so items in unloaded groups can be found, too.
+
 ### 1.0.6
 
 * Added support for region mapping based on region names instead of region numbers (example in `public/test/countries.csv`).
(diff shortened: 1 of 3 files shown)


kring:
Bonus change: don't automatically fly to a location when pressing enter in the Search box.  No one really expected this, and it did weird things sometimes.

PR #687, opened 2015-05-13
Use WMS layer hierarchy in GetCapabilities for the catalogue

### LINKED ISSUE #672: Support nested WMS layers
Here's an instance of a WMS server with deeply nested layers: http://srss-ows.landgate.wa.gov.au/wxs/firewatch.php?service=wms&request=getCapabilities

Loading it into Terria, it gets flattened out and looks bad - lots of layers with duplicate names like "Band 5". 

I'm guessing the behaviour change is around [here somewhere](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/WebMapServiceCatalogGroup.js#L255) - probably needs to be configurable, as I'm guessing the current flattening behaviour was intentional for existing data sources.

kring:
If desired, the old behavior can be restored by setting flatten=true.

When a layer is both a container for other layers and a toggle-able thing itself, the first item in the group will be "(All) Whatever Layer Name".  See the screenshot below.

Fixes #672 

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### 1.0.8
+
+* `WebMapServiceCatalogGroup` now populates the catalog using the hierarchy of layers returned by the WMS server in GetCapabilities.  To keep the previous behavior, set the `flatten` property to true.
+
 ### 1.0.7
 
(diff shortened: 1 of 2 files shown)


kring:
All of the existing WMS servers in National Map work well with this new feature.  Other Terria apps using WMS GetCapabilities querying will want to make sure it works well with their servers, too, and set `flatten: true` in the init file if necessary to get the old behavior.

PR #697, opened 2015-05-18
Allow Esri ArcGIS MapServers to be added via the Add Data panel.

kring:


stevage:
Awesome - ran into this last week with APHCRI. (But I think I forgot to report it) :)

PR #698, opened 2015-05-18
Share maps panel settings

### LINKED ISSUE #55: Add more to share url - camera angle, mode, base layer, etc.

kgrochow:
Fixes #55

Adds baseMapName and initialViewerMode fields to init files and share links.
baseMapName is any base map name in the map settings panel
initialViewerMode can be set to '2d' and is handled the same as if it is in the url.  It can also be set to '3d', but since that's the default it will have no effect.
The retaining 2d mode in links may still be contentious, but I've come to the view it's better to be consistent with the original sharer than try to upgrade it at the sharee.  It also seemed inconsistent to support 1 thing from the map panel and not the other.


lib/ViewModels/SettingsPanelViewModel.js
```diff
@@ -22,6 +22,15 @@ var SettingsPanelViewModel = function(options) {
     this.baseMaps = defaultValue(options.baseMaps, []).slice();
     this.mouseOverBaseMap = undefined;
 
+    var that = this;
+    if (defined(options.baseMapName)) {
+        this.baseMaps.map (function(item) {
+            if (options.baseMapName === item.catalogItem.name) {
+                that.terria.baseMap = item.catalogItem;
```
kring: This needs to go somewhere else, because the base map specification needs to work even in an application that doesn't use the `SettingsPanelViewModel`.  I'll see if I can find a better place.

kgrochow: I started with it in index.js, but then it's only available it you build that in a certain way.  This is the first chance to have the basemaps list to check against.

kring: > I started with it in index.js, but then it's only available it you build that in a certain way.

Yeah, but you're still requiring that index.js do something specific (i.e. use the settings panel).  If you're going to impose a requirement on applications, it's much better for it to be a non-UI requirement.

kring:
I'm happy with this now.  @kgrochow want to take a look and make sure you still are?

PR #710, opened 2015-05-25
Fix brandbar

axman6:
These changes were made in NationalMap but somehow went missing during the TerriaJS split.


axman6:
Hmm, seems my TerriaJS was out of date when I made this. I'll close it and make a new pull request.

PR #711, opened 2015-05-26
Allow BrandBar height to be set with @brand-bar-height

axman6:
These changes were made in NationalMap but somehow went missing during the TerriaJS split.


kring:
Please update `CHANGES.md`.  I'd do it myself, but I won't have write access to your branch.  This is what I was planning to add to the 1.0.13 section:

```
* Breaking changes:
  * Added a required `@brand-bar-height` property.
```

kring:
We should reorganize our less files so that this type of thing wouldn't be a breaking change.  Less uses weird (yet consistent with CSS) rules for variable overriding, so basically we'd want to restructure index.less to put all the variable declarations at the bottom instead of the top, and then define default values at the top of `BrandBar.less`.

PR #728, opened 2015-06-02
user message when catalog item initially opened

kgrochow:
allow a user message from the init file to be displayed to the user when a catalog item or group is initially opened.  To keep it simple it will appear only once per session, but currently doesn't use cookies to work across multiple sessions.  Needed to NA site and probably states.


lib/Models/CatalogMember.js
```diff
@@ -68,7 +68,14 @@ var CatalogMember = function(terria) {
      */
     this.isPromoted = false;
 
-    knockout.track(this, ['name', 'info', 'description', 'isUserSupplied', 'isPromoted']);
+    /**
+     * Gets or sets the URL from which this data item's raw data can be retrieved, or undefined if raw data for
```
kring: This doc is wrong (copy/pasted).

lib/ViewModels/DataCatalogTabViewModel.js
```diff
@@ -38,4 +40,32 @@ DataCatalogTabViewModel.prototype.showInfo = function(item) {
...
+    }
+
+    item.toggleEnabled();
+};
+
+DataCatalogTabViewModel.prototype.clickOpen = function(item) {
+    ga('send', 'event', 'dataSource', 'clickOpen', item.name);
+
+    if (!item.isOpen && defined(item.initialMessage) && item.initialMessage.length > 0) {
+        PopupMessageViewModel.open('ui', {
```
kring, line 61: Because this is done here (rather than being something initiated by the `CatalogItem` itself), it will only happen when the catalog item is enabled via the UI.  E.g. it won't show up when the catalog item is enabled via a share link or by any other programmatic mechanism.  Is this what we want?

lib/ViewModels/DataCatalogTabViewModel.js
```diff
@@ -38,4 +40,32 @@ DataCatalogTabViewModel.prototype.showInfo = function(item) {
...
+            title: 'Data Provider Information',
+            message: '<div>'+ item.initialMessage +'</div>'
+        });
+        item.initialMessage = undefined;
```
kring: I don't like the lossiness of this.  For one thing, if you enable a catalog item, then use the Tools menu to download the catalog as JSON, the JSON won't contain the `initialMessage`.  A separate property to track whether its already shown would be better.  The property should be made non-serialized by adding a do-nothing handler for it to `defaultSerializers`.

lib/ViewModels/DataCatalogTabViewModel.js
kgrochow, line 61: I think it's fine.  If you're sending it to someone via a link it's not clear what should happen.  What if there are 10 sections open with a message each?

lib/ViewModels/DataCatalogTabViewModel.js
kgrochow: The catalog doesn't download properly for lots of reasons right now.

kring: Ok, but that's just one example of why losing information like this isn't a good idea, especially since it's easy to fix.

PR #730, opened 2015-06-03
Region from file

kgrochow:
- Moved region mapping definition to wwwroot/data/RegionMapping.json.  Defining valid regions can now be set by editing this file.
- Change displayable variables to just be scalar variables (dropped enumerated ones for now).
- Info box displays csv variables greater than 99999 with commas.
- General cleanup of CsvCatalogItem and tableStyle in prep for abs integration


lib/Models/CsvCatalogItem.js
```diff
@@ -192,6 +193,20 @@ defineProperties(CsvCatalogItem.prototype, {
...
             return CsvCatalogItem.defaultPropertiesForSharing;
         }
+    },
+
+    /**
+     * Gets or sets the tableStyle object
+     * TODO: add definition for subfields
+     * @type {Object}
+     */
+    tableStyle :  {
```
kring: This isn't used because the `tableStyle` defined in the constructor (with the `knockout.track` line) overrides it.

kring:

lib/Models/CsvCatalogItem.js
kgrochow: It actually does get used and it's the only way I could get tableStyle to get handled properly across load, setting, and sharing.

kgrochow: Well it does get hit on loading.  I'll see if it works without it.

kgrochow: Yep, share doesn't work properly without it.

kgrochow:
I moved the getter/setter and removed tableStyle from knockout.track.

PR #732, opened 2015-06-04
Slow dataset related changes

kring:
- Upgraded to TerriajS Cesium 1.10.2 (with support for `loadWithXhr` timeouts, most notably).
- Added `ImageryLayerCatalogItem.isRequiredForRendering`.  This is set to false by default and to try for base maps.  Slow datasets with `isRequiredForRendering=false` are less likely to prevent other datasets from appearing in the 3D view.
- The "Dataset Testing" functionality (on the hidden Tools menu accessible by adding `#tools=1` to the URL) now gives up tile requests and considers them failed after two seconds.  It also outputs some JSON that can be used as the `blacklist` property to blacklist all of the datasets that timed out.
- Added a feature to count the total number of datasets from the hidden Tools menu.
- Fixed a bug that caused the 2D / 3D buttons the Maps menu to get out of sync with the actual state of the map after switching automatically to 2D due to a performance problem.


kring:
image

bsimyo:
Nice!

PR #733, opened 2015-06-04
Added Open Street Map imagery

reneenoble:
Add ability to add Open Street Map Imagery from Add Data Panel


kring:
Thanks @reneenoble!  I'm going to merge this and then tweak a couple of things.

PR #738, opened 2015-06-09
Add support for Web Map Tile Service (WMTS)

### LINKED ISSUE #473: work with geowebcache formats
Geowebcache is the default webcache that ships with Geoserver and supports 4 different ways of getting cached tiles - integrated, WMS-C, TMS, and WMTS.  Supporting them is not difficult and will make our data provider story better.

This is related to #413 in that both will require us to rebuild our tile cache from scratch, so we should do it all at once.

### LINKED ISSUE #737: Loading CZML files from init files broken
At some point after terria 1.0.15/terria-cesium 1.8.8 loading CZML files from an init file has broken. It still works fine to drag and drop the file.

There are CZML files in the `test.json` init file, Test Data -> Files -> Large feature info box (CZML).

The error message show is:

```
TypeError: e.loadUrl is not a function
```

kring:
Added `WebMapTileServiceCatalogGroup` and `WebMapTileServiceCatalogItem`.

Fixes #473 

Bonus change:
Fixes #737

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 
 * Added `OpenStreetMapCatalogItem` for connecting to tile servers using the OpenStreetMap tiling scheme.
 * Added `CkanCatalogGroup.allowEntireWmsServers` property.  When set and the group discovers a WMS resource without a layer parameter, it adds a catalog item for the entire server instead of ignoring the resource.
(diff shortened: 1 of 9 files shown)


kgrochow:
lint isn't happy.

kring:
ok this is fixed now.  I had tried to specify Cesium as a peer dependency of TerriaJS, and it worked ok locally, but Travis didn't like it (possibly because I don't know what I'm doing).  Back to a regular dependency for now.

PR #740, opened 2015-06-09
add featureInfoFields to tableStyle

kgrochow:
Added featureInfoFields to csvCatalogItem.tableStyle (see nm test.json for example).  For example to show the incident_date and incident_severity items only with better names use the following in the style in the init.
"featureInfoFields": {
                                    "incident_date": "Date", 
                                    "incident_severity" : "Severity"
                                }
Also right aligned data in value column of featureInfoPopup and moved click bind in leaflet for markers so click works on images.


kgrochow:
It was going to take some more thought to see how to apply this to WMS or other catalog types, so I've left that for a future project.

PR #746, opened 2015-06-11
Made navigation toolbar modular by abstracting out its controls

keattang:
Each control is abstracted out into a NavControl which is then added to list of controls on the NavigationViewModel. This is to allow easier addition of navigation controls (e.g. find my position). To make this work the NavigationViewModel should be instantiated with the desired nav controls. 

```
// Create the navigation controls.
    NavigationViewModel.create({
        container: ui,
        terria: terria,
        controls: [
            new ZoomInNavControl(terria),
            new ResetViewNavControl(terria),
            new ZoomOutNavControl(terria)
        ]
    });
```

### DIFF
--- a/lib/Models/NavControl.js
+++ b/lib/Models/NavControl.js
@@ -1,0 +1,168 @@
+'use strict';
+
+/*global require*/
+
+var defined = require('terriajs-cesium/Source/Core/defined');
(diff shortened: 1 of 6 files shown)


kring:
Thanks @keattang this is great!  I'm going to pull this into a separate branch to make a small change to it to preserve backward compatibility for apps already using the `NavigationViewModel`.

PR #748, opened 2015-06-11
Add an attention grabber for the Now Viewing panel.

kring:
image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### 1.0.19
+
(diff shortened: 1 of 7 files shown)


kring:
I just modified this to use `localStorage` (which surprisingly is supported on IE8+) to make sure that this popup is only shown once ever.  Local storage key/values are similar to cookies except they're not sent to the server on each request.

To reset the flag (for testing, etc.), use the Chrome developer tools, Resources tab, Local Storage section and delete the `NationalMap.NowViewingAttentionGrabberViewModel.shownBefore` key, or set it to false.

PR #763, opened 2015-06-16
Default terrain server to use Geoscience Australia's server

Layoric:
Hi Team,

I was asked to update the terrain server URL to use GA's own terrain server. I'd understand that you might not want to take a direct dependency on another terrain server out of your own control into TerriaJS framework, however I'm told this is something the NationalMap project needs.

Also note that although the URL updated below is also using a protocol relative URL as previously used, the HTTPS endpoint will redirect to HTTP for the Geoscience Australia server which may not be appropriate for your requirements.

Is this URL planned to be exposed via a form of configuration in the future?

Cheers!

### DIFF
--- a/lib/viewer/AusGlobeViewer.js
+++ b/lib/viewer/AusGlobeViewer.js
@@ -140,7 +140,7 @@ latest version of <a href="http://www.google.com/chrome" target="_blank">Google 
                     viewer.selectViewer(true);
                 } else {
                     terria.cesium.scene.globe.terrainProvider = new CesiumTerrainProvider({
-                        url : '//cesiumjs.org/stk-terrain/tilesets/world/tiles'
+                        url : '//www.ga.gov.au/terrain/terrain/'
                     });
                 }
             } else if (newMode === ViewerMode.CesiumEllipsoid) {


kgrochow:
We will need to do a set of performance and accuracy tests before feeling confident about switching over.  From just a quick scan there still appear to be problems with the GA data.  Here is a an image in Katoomba with Cesium data.

image

And this is the GA data with obvious problems with the cliff edge:

image

Layoric:
@kgrochow Thanks for checking it out so quickly. I've passed on the issue you've found, again completely understand not wanting the risk of taking a dependency on a server further from your control. I'm not aware of the hosting situation of the terrain server (server specs and performance) but would appreciate if you post any performance results you have here if/when you perform them.

Are you able to provide any advice regarding having the terrain server URL being configurable by the application consuming TerriaJS? It would make sense to allow an application using TerriaJS to customise the use of terrain data instead of always relying on Cesium public servers. Applications (like NM) might want to use alternate data sources for various other reasons.

Thanks again.

kring:
I just opened #769 to change TerriaJS so that the terrain provider can be specified externally.  Once that's merged, we can close this and open a similar pull request in [NICTA/nationalmap](https://github.com/NICTA/nationalmap).  We still need to resolve if this switchover is beneficial for National Map, of course.

Layoric:
Perfect! Thanks for the update.

PR #772, opened 2015-06-22
Add support for CSV with lat/lon but no value column.

kring:


kring:
Closing this #773 is a better way to do it.

PR #779, opened 2015-06-23
Custom Catalog item controls

keattang:
I've created the capability to include additional controls in the data catalog on a per item basis. These controls appear next to the Info control and are represented by SVGs. This gives catalog items greater capability and allows them to be used for more than just displaying things on the map. The following describes the key changes:
- Abstracted the key parts of NavigationControl into UserInterfaceControl
- Created CatalogItemControl to be displayed on the right hand side of the item in the catalog.
- Controls also show in the NowViewing panel as text, inline with the other controls (Info, Remove, etc).
- Made the checkbox and info controls optional through the CatalogItem properties isMappable and showsInfo
- Created a CatalogItemDownloadControl as an example which takes a url parameter and downloads the resource at that url when clicked.


kring:
Thanks for this @keattang, it's a very useful feature.  I hope it doesn't sound too pedantic, but there's some level violation going on here.  A `CatalogItem` shouldn't hold on to UI-related objects like the `rightSideControls`.

IMO a better design (though admittedly more complicated) is for the `DataCatalogTabViewModel` to have a list of potential right-side controls.  Those controls then choose whether or not to show themselves for each catalog item based on properties of that catalog item.  This way the UI knows about catalog items but the catalog items don't need to know about the UI.

keattang:
I had thought about doing it that way but it would potentially require adding a unique property to the `CatalogItem` for every control that you created and then you'd also have to have system for registering those controls both of which seem messy. 

For example if an item has a `url` property that it uses and you wanted to add a download control to that item that downloads from a different `url` you can't use that property. You won't necessarily want to have a download control for every item that has a `url` property.

However, if you're happy for the control to check for the existence of a unique property that it defines then I can rework it that way and add a `registerCatalogItemControls` method.

kring:
Yes, I think controls would have to check for a unique property.  In your example, the download control might check for a `downloadUrl` property, which may or may not be the same as the `url` property.

keattang:
Do you think `downloadUrl`  should be explicitly defined in the `CatalogItem` class or should the control just check for it either way?

kring:
That's a good question.  We already have a `dataUrl` property and a corresponding `dataUrlType`.  So maybe the download control just looks for a specific type of `dataUrl`, such as `download`?  Longer term I'd like to extend catalog items to have multiple data URLs, each with a different format, description, etc., but that's for later.

It might also be useful to make it easier to extend catalog items with custom properties.  In JavaScript you can just add any property to any object, of course, but custom properties won't be automatically read from init files.  Plus it's kind of ugly.  A slight improvement is to have a `customProperties` property on `CatalogItem` that is then a free-form property bag.

Or, we could get a bit fancier and have a mixin system.  So the download mixin adds properties for specifying how a catalog item can be downloaded.  Mixins would be named and registered so they can be specified in init files (e.g. `mixins: ['download']`) and then their properties can be read from the init file along with the built-in ones.  The download control would then check for the presence of the download mixin on the object.  Mixins would be useful in some other places in TerriaJS, too.

I'm mostly just thinking out loud; I certainly don't expect you to do all this.  If you're happy with a `dataUrl` of `dataUrlType='download'` I think that's a good way to go in the short term.

keattang:
To do that I will actually need multiple dataUrls as I have some items with multiple controls using different urls. So I can make a start on that if you like? 

Looking at it, the quickest way would be to make a `dataUrls` property that is an array of objects each with a url and type. Then create a `defaultUpdater` that handles the existing case of a single `dataUrl` and separate `dataUrlType`. And finally tweak `CatalogItemInfo` to run through the array. Does that sound about right?

Alternatively, I can just use the `customProperties` method for now. What do you think?

kring:
Adding a `dataUrls` property sounds ok to me.  You could take a look at the recently added `legendUrls` (in `CatalogItem.js`) as a model for how to do it in a knockout-y way.

If you would prefer to use the `customProperties` approach for now I'm ok with that, too.  I don't mean to create a lot of extra work for you.

keattang:
I think I might go with the customProperties option for now as I have a bit of a deadline. Thanks for the help! I'll let yo know when this is ready to be merged again

keattang:
Hi @kring,

I've just finished implementing this the right way around. In summary:
- Controls are created by subclassing `CatalogItemControl` and then registered with the property to look for in `registerCatalogItemControls` which needs to be called in an `index.js` like file.
- `CatalogItem`s have an additional property named `customProperties` which is an object.
- `CatalogItemControl`s look for the property they were registered with, first on the `CatalogItem` then in `customProperties`
- `CatalogItemControl`s appear in the catalog and in now viewing if the item has the required property.
- Controls can be designated to be either on the left or right side of the item text
- `CatalogItem`s have an additional property `isMappable` that defines whether or not the item can be displayed on the map. Consequently, if this is false, the checkbox doesn't appear on the item.

lib/Models/CatalogItem.js
```diff
@@ -334,6 +354,8 @@ CatalogItem.defaultPropertiesForSharing.push('isEnabled');
 CatalogItem.defaultPropertiesForSharing.push('isShown');
 CatalogItem.defaultPropertiesForSharing.push('isLegendVisible');
 CatalogItem.defaultPropertiesForSharing.push('nowViewingIndex');
+CatalogItem.defaultPropertiesForSharing.push('isMappable');
+CatalogItem.defaultPropertiesForSharing.push('showsInfo');
```
kring: These two properties don't need to be included in share URLs, since they're usually configured from a catalog (as opposed to a user-modified value like `isShown`).

keattang:
Cool fixed!

kring:
Thanks @keattang!

PR #780, opened 2015-06-23
Flexible catalog item zoomTo tolerance

keattang:
Made zoomTolerance a property of CatalogItem so that the zoomTo tolerance (epsilon) can be altered. This is useful for catalog items with small rectangles. Without changing it, zoomTo doesn't zoom in far enough.

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -463,7 +463,7 @@ var scratchRectangle = new Rectangle();
 
         ga('send', 'event', 'dataSource', 'zoomTo', that.name);
 
-        var epsilon = this.zoomTolerance;
+        var epsilon = that.zoomTolerance;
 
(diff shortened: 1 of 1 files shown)


kring:
@keattang what tolerance are you using?  It may be ok to just get rid of the tolerance entirely and allow zooming to any extent where the corners are not at the exact same location.  Changes made to Cesium since this code was added should prevent the camera from going under the terrain no matter how closely we zoom.

keattang:
@kring I'm using `CesiumMath.EPSILON6`. If that's the case then that's probably a better solution. On the other hand I suppose it can be used to prevent zooming to a level that's ridiculously close on really small extents such as a point. I don't mind either way.

kring:
Ok I opened pull request #815 and I'm going to close this one.

PR #785, opened 2015-06-25
Billboard item

keattang:
Created a catalog item that displays on the map as a billboard. Selecting the billboard also triggers an empty clicked function that can be overwritten in subclasses to perform an action. This is useful for displaying clickable pins on the map


kring:
@keattang why not use `CzmlCatalogItem` for this?  For example, here's a JSON catalog item definition that will display a point in the middle of Australia:

```
{
    "name": "Embedded CZML",
    "type": "czml",
    "data": [
        {
            "id": "document",
            "version": "1.0"
        },
        {
            "position": {
                "cartographicDegrees": [133.0, -25.0, 0.0]
            },
            "point": {
                "pixelSize": 5,
                "color": {
                    "rgba": [255, 0, 0, 255]
                }
            }
        }
    ]
}
```

I think the only thing this doesn't have is the `billboardClicked` bit.  We could either add that to `CzmlCatalogItem` in general, or you could just subscribe to the `Terria#pickedFeatures` property manually.

keattang:
Ok I'll have a play around with that and see if I can get the same outcome. My main purpose for the billboard item was to be used as a clickable pin that then performs some action.

I'm assuming that a czml item won't automatically update the location of the billboard so that it is on the terrain and that I'll have to do that manually too?

kring:
Cesium now has built-in support for automatically putting billboards and labels on terrain (or a specified height above it).  Unfortunately this support hasn't been extended to CZML yet.  I don't think there's any good reason for that, though, other than no one has gotten around to doing it yet.

keattang:
I've been trying to find that feature. How is it triggered?

kring:
You have to create a BillboardCollection, pass the `scene` to its constructor, and then set the `heightReference` property of each `Billboard` in the collection to `HeightReference.CLAMP_TO_GROUND` or `HeightReference.RELATIVE_TO_GROUND`.

keattang:
awesome thanks!

keattang:
I can't seem to create a billboard collection with or without the scene. I keep getting

`Uncaught TypeError: Cannot read property '_textureAtlas' of undefined`

kring:
Hmm I don't know.. what's your code look like, and where exactly is that error occurring (i.e. what line in the Cesium code)?  Maybe move this to email?

RacingTadpole:
Hi guys - where's this PR up to? Do you want me to take a look?

kring:
@keattang I'm closing this because it hasn't moved in months, and I _think_ `CzmlCatalogItem` is an effective alternative.  Feel free to reopen if I'm wrong.

keattang:
No worries.

PR #788, opened 2015-06-26
Abstracted out core methods of popup views

keattang:
Useful if more popup views need to be implemented down the track. (coming in another pull request)

The html view has also been made into a variable allowing multiple views to work with the same view model.


kring:
Thanks for the pull requests @keattang, all good stuff!  We're working hard to wrap up some things at the moment so I may not get a chance to merge these until late next week.

keattang:
No worries! Thanks!

PR #794, opened 2015-06-29
Custom catalog group sorting

keattang:
Added a property to CatalogGroup that defines the function used to sort its items when preserveOrder = true.


kring:
Thanks @keattang!

PR #802, opened 2015-07-01
Region mapping feedback

stevage:
Better UX when using the CSV drag-and-drop feature, following some feedback at Spatial Analytics Workshop.

### DIFF
--- a/lib/Models/CsvCatalogItem.js
+++ b/lib/Models/CsvCatalogItem.js
@@ -546,7 +546,7 @@ function loadTable(csvItem, text) {
     }
 
     csvItem._tableDataSource.loadText(text);
-
+    // If there is specifically a 'lat' and 'lon' column.
     if (!csvItem._tableDataSource.dataset.hasLocationData()) {
         console.log('No locaton date found in csv file - trying to match based on region');
         return loadJson(csvItem.terria.regionMappingDefinitionsUrl).then(function(obj) {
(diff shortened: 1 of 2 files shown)


stevage:
(I suspect it's not totally kosher to use PopupMessageViewModel from within CsvCatalogItem, so I'm open to hearing the right way to display the warning.)

stevage:
screenshot 2015-07-01 17 19 03

kring:
> I suspect it's not totally kosher to use PopupMessageViewModel from within CsvCatalogItem

You can use `terria.error.raiseEvent(new ModelError({ ... }))` where `...` is the same arguments you'd pass to `PopupMessageViewModel.

stevage:
I don't want to actually throw an error though (which would prevent the layer being added) - I just want to show a warning as well.

kring:
Yep, raising the error event like that won't abort anything that's currently happening (unlike if you threw that ModelError instead).

kgrochow:
This still needs to be thought out more based on what our future goals are here.  Also the error message should go by UX since it is user facing.

kgrochow:
As to raising an error if it can't find a region, if it looked for and can't find a region then it has no location data.  We can't show it on the map, so there's no reason to load it.

kring:
@hilarycinis @philipgrimmett take a look.  The message is an obvious improvement though, let's not put up unnecessary barriers.

I know there's some debate about how strictly we should match column names.  Maybe Keith and Steve could have a short Skype about it?  Just to throw out another idea (compromise?), maybe we could rig up the code to match `steCode` and `ste_code` and `ste-code` but not `stecode`?

Meeena:
This is definitely on the radar for UX. It's part of a larger UX issue of how to explain that region mapping exists, and how the user can do it effectively. 

I think the message is an improvement. Can we do some basic styling while we come up with a more comprehensive solution?
screen shot 2015-07-02 at 12 52 51 pm

Umm. Queensland should be on another line with a point.

kgrochow:
All the region mapping files in the test folder now fail or give spurious errors.  e.g, it doesn't appear to handle time based region mapping properly.

kgrochow:
There also appear to be problems with the ABS data.

stevage:
> As to raising an error if it can't find a region, if it looked for and can't find a region then it has no location data. We can't show it on the map, so there's no reason to load it.

Yeah I would definitely do that if no regions matched, but I don't think it's correct to fail to show a CSV just because a single region couldn't match, for instance. (Especially since there are likely datasets with the odd non-spatial "region" like "No fixed address" etc)

stevage:
> I think the message is an improvement. Can we do some basic styling while we come up with a more comprehensive solution?

Done.

> Umm. Queensland should be on another line with a point.

I'm unsure about this one. It could easily be dozens to hundreds of items, and that will look really bad. Happy to keep discussing though.

kring:
@stevage let us know when this is ready for another look.

stevage:
yes, it's ready for another look.

stevage:
Just a quick update - the name matching stuff works as intended, but it has two shortcomings:
- 'state' can only be numeric (ABS codes) or text. I think text makes more sense, but the test data has codes, and really the best solution is to allow either. (IMHO 'ste' should require numeric thuogh).
- 'lga_name' is completely ambiguous for two "City of Campbelltown"s, and ambiguous other than council type for half a dozen others: Kingston, Wellington, Latrobe, etc. I think I really need to add multi-column disambiguation.

Within what I've implemented it's possible to solve both the above problems through creative regexes in the regionmapping.json (eg, search/replace "NSW" with "1" or whatever), but it's fairly hacky.

kring:
This looks good to me.  The only problem I see is that in the time-dynamic example `#test -> Test Data -> Files -> Droughts by Country`, warnings pop up at every time step, which makes it pretty much unusable.  Is it possible to find and report all the problems on load rather than at each time step?

kgrochow:
Please do not merge this before I check in my region mapping changes.

lib/Models/CsvCatalogItem.js
```diff
@@ -672,25 +722,40 @@ function createRegionLookupFunc(csvItem) {
...
+            row = i;
+        }
+        totalrows++;
+        if (typeof codes[row] === "number") {
             index = ids.indexOf(codes[row]);
-            if (index >= 0 && colors[index][3] === 0) {
-                colors[index] = dataSource._mapValue2Color(vals[row]);
-            }
+        } else {
+            index = ids.indexOf(String(codes[row]).toLowerCase());
```
kring, line 741: I don't think this is new, but if performance is a concern it looks it should be easy to replace this linear search with a hash lookup.  I have to imagine that would make a big difference with e.g. SA1.

kring:
@kgrochow yep I'll let you merge it.

stevage:
Closing this as it's contained with region-mapping-advanced-feedback branch.

PR #805, opened 2015-07-03
Use correct template for search results.

arrayjam:
The data-catalog-item template was being used, which prevented some
interactions with the search results.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### 1.0.34
+
+* Fixed a bug that prevented catalog items inside groups on the Search tab from being enabled.
+
 ### 1.0.33
 
 * Added the `legendUrls` property to allow a catalog item to optionally have multiple legend images.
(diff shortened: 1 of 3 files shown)


kring:
@arrayjam what interactions was this preventing?  While I agree that previous code was wrong-ish, the change here has problems, too.  For example, if you search for "nat" and open National Data Sets, you'll see that all groups within that group are opened as well.  That's because search-catalog-item-template detects whether a group is open with `$parent.isOpen` rather than just `isOpen`.  The `$parent` makes sense at the root level because the parent of the catalog item is a search result, which has its own `isOpen` property.  We bind to that one so that "open in the search tab" and "open in the catalogue" are not the same property.

hilarycinis:
@kring Is there an impact on the presentation layer (GUI and interactions) with this?

kring:
@hilarycinis I don't think so, but it depends on what problem @arrayjam saw.

arrayjam:
Thanks for your help, Kevin.

The issue I was seeing was #820. I've fixed it so that it now works in both cases.

kring:
Great, thanks @arrayjam!  Just FYI, I added an entry to `CHANGES.md`, merged in master, and switched to consistently using single quotes for the `search_result` string.

PR #813, opened 2015-07-06
Geojson styles

stevage:
Support styling GeoJSON objects at both catalog and feature level.
    To style all objects in a GeoJSON file the same (overriding the
    defaults), add a .style {} property with SimpleStyle features.
    Or embed SimpleStyle properties directly on features in the file


kring:
@stevage there are jshint warnings here, at least some of which are real problems, too.  Visit `#test` and enable `Test Data -> Files -> Cemeteries`.  It crashes in `loadGeoJson` because `defaultProperties` is not defined.

stevage:
Thanks @kring. Fixed both.

kgrochow:
Is there a test file available in NM to to make sure it's working?  If not, we should add one there along with the other geojson flavours.

stevage:
Yes, there's [this](https://github.com/TerriaJS/terriajs/commit/dc84c6b458d3cd1a02c119983afe47f1192f7504), which is referred to in [this](https://github.com/TerriaJS/terriajs/commit/dc84c6b458d3cd1a02c119983afe47f1192f7504).

It contains styled and unstyled polygons, lines and points (markers).

I guess I should make explicit the connections between commits in these different repos. Haven't quite found a smooth workflow for that yet.

PR #824, opened 2015-07-09
Fine grained styling

keattang:
Added extra less variables to styles for more fine grained customisation. The changes revolve around text and background colours. The current values are set as default so nothing should change for anyone but there is the option to customise further in a `main.less` file.


kring:
Great, thanks @keattang!

PR #825, opened 2015-07-10
Tile Map Service Catalog Item

keattang:
Pretty self explanatory

### DIFF
--- a/lib/Models/TileMapServiceCatalogItem.js
+++ b/lib/Models/TileMapServiceCatalogItem.js
@@ -1,0 +1,97 @@
+'use strict';
+
+/*global require*/
+
+var TileMapServiceImageryProvider = require('terriajs-cesium/Source/Scene/TileMapServiceImageryProvider');
+var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
+var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
+
+var ImageryLayerCatalogItem = require('./ImageryLayerCatalogItem');
+var inherit = require('../Core/inherit');
+
+/**
+ * A {@link ImageryLayerCatalogItem} representing a layer from a tile map server.
+ *
+ * @alias TileMapServiceCatalogItem
+ * @constructor
+ * @extends ImageryLayerCatalogItem
+ *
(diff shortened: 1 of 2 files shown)


kring:
Thanks!  By the way, have you seen Cesium's new `UrlTemplateImageryProvider`?  It might be better to create a catalog item for that rather than TMS (and we can get rid of the `OpenStreetMapCatalogItem`, too).  The only thing we'd lose is the TMS provider's inspection of `tilemapresource.xml`.  Is that useful for you guys?

keattang:
We don't use the metadata anywhere. Does the imagery provider use it to optimise the tiles it calls by not calling tiles that are outside the bounds? Will UrlTemplateImageryProvider be able to do the same?

kring:
Yes it uses the metadata (if available) to get min/max levels, tiling scheme, valid extent, etc.  All of these things can be specified to `UrlTemplateImageryProvider` explicitly, it just won't read them automatically from `tilemapresource.xml`.  It sounds like `UrlTemplateImageryProvider` might be what you want.

lib/Models/TileMapServiceCatalogItem.js
```diff
@@ -0,0 +1,97 @@
...
+     * @type {String}
+     */
+    this.url = undefined;
+
+    /**
+     * Gets or sets the maximum zoom level.
+     * @type {Integer}
+     * @default 18
+     */
+    this.maxZoom = 18;
```
kring, line 37: Is this line the cause of #939?

kring:
@keattang sorry for sitting on this for so long.  Is it still relevant?  As mentioned above, I think `UrlTemplateCatalogItem` might be the better way to go.  I'm going to close it but feel free to reopen if you think it still makes sense.

keattang:
@kring I'm looking at implementing a `UrlTemplateCatalogItem` for TMS data but I still want to use the metadata included in `tilemapresource.xml` so as to avoid requests to tiles that don't exist. What would you recommend is the best way to do this? Should I inspect `tilemapresource.xml` separately and feed the info into the `UrlTemplateCatalogItem` or should I allow some provision in the catalog item to try go and find the metadata?

PR #831, opened 2015-07-12
Use the entire group name, not just the part before the first comma.

kring:
When grouping CKAN datasets by Group.


kgrochow:
I'm pretty sure this was there to fix some other dga oddity.  I'll try to figure out what it was.

kring:
Yeah I looked at the git blame and I learned that I added it (very early on) but there's no explanation there or in my memory as to why.  In any case, d.g.a no longer groups by Group so it may not matter, at least in the short term.  Vic looks fine as far as I can see.

kgrochow:
It was probably was to keep some group names manageable in length, but I don't see any obvious issues out there.

PR #841, opened 2015-07-15
Add support for loading topojson files.

arrayjam:
Detects whether a file is TopoJSON, converts it to GeoJSON, and carries on. 

[Here is a topojson file](http://bost.ocks.org/mike/map/uk.json) to try it out with.


kring:
@arrayjam TerriaJS / Cesium already have built in support for TopoJSON.  Drag your file into nationalmap.gov.au, it works fine.

arrayjam:
@kring: I see, it loads but doesn't pan to the extent. It also creates a mesh, rather than a feature polygon, but it also loads all the objects rather than just the first. I had a quick look for TopoJSON in TerriaJS and couldn't find an implementation, so I did this. I'll look into improving the current implementation. Thanks!

kring:
Yeah Cesium's TopoJSON support is built into `GeoJsonDataSource` (for some reason).  The automatic zoom-to doesn't work because the TerriaJS code that computes the extent isn't TopoJSON-aware.

By "mesh" do you mean it just has an outline, the polygons aren't filled?  If so, that's already fixed in the latest versions of TerriaJS (it's just not up on nationalmap.gov.au yet).

PR #844, opened 2015-07-16
Added ability to filter catalog search results by type: `is:wms`,

stevage:
`is:esri-mapserver`, `is:geojson` and so on.


kring:
This crashed when I search for `is:g`.  The problem is that `findItemsInLoadedGroup` calls `findMatchingItemsRecursively` and doesn't pass the new parameter.

stevage:
Hmm, I can't reproduce that. What kind of "crash"? I'm not seeing any exceptions (other than the failed Gazetteer lookups), and `is:group` behaves as expected for me.

But I've fixed it anyway.

kring:
Do you have your debugger set to stop on all exceptions?  If not, the exception will be ignored (and the search won't find any items in groups that need loading).  In any case, I confirmed that your fix works.

stevage:
I had 'stop on caught exceptions' turned on, fwiw.

PR #846, opened 2015-07-17
More search filtering

stevage:
Surprisingly useful. Yes, we should have a proper UX with UI elements, but this was pretty easy to knock up.

### DIFF
--- a/lib/ViewModels/CatalogItemNameSearchProviderViewModel.js
+++ b/lib/ViewModels/CatalogItemNameSearchProviderViewModel.js
@@ -24,6 +24,43 @@ var CatalogItemNameSearchProviderViewModel = function(options) {
 inherit(SearchProviderViewModel, CatalogItemNameSearchProviderViewModel);
 
 CatalogItemNameSearchProviderViewModel.prototype.search = function(searchText) {
+    function parseSearchFilters() {
+        /* Filter search by type, eg 'is:wms fish' or '-is:wfs house'. Multiple 'is' clauses are or-ed. */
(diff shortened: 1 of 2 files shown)


lib/ViewModels/CatalogItemNameSearchProviderViewModel.js
```diff
@@ -72,11 +104,52 @@ CatalogItemNameSearchProviderViewModel.prototype.search = function(searchText) {
...
+function itemMatchesFilters(item, searchFilters) {
+    if (searchFilters.typeIs.length > 0 && searchFilters.typeIs.indexOf(item.type.toLowerCase()) < 0) {
+        return false;
+    }
+    if (searchFilters.typeIsNot.length > 0 && searchFilters.typeIsNot.indexOf(item.type.toLowerCase()) >= 0) {
+        return false;
+    }
+    if (!item.url) {
+        // if no URL, it can't match any positive or negative filters.
+        return searchFilters.urlMatches.length >= 0;
```
kring: I think this logic is wrong.  It should return false when the item has no URL and there's a URL filter, not true, right?

kring:
This looks good other than the one comment.

stevage:
Thanks for catching that!

kring:
There seems to be a problem with queries not containing a `url` portion.  For example, the query `url:gov ADSL` finds several catalog items, as expected.  But `ADSL` finds no results.

kring:
@stevage for future reference, drop a note in here when you've committed some code that addresses review comments and it's ready for another review.  GitHub doesn't automatically notify of commits to pull requests, so a message is my reminder to come back and look again.

stevage:
Ta, I was wondering about that.

PR #853, opened 2015-07-20
Socrata support basically working.

stevage:
Adds basic support for Socrata, and hence data.melbourne.vic.gov.au and data.act.gov.au.

By "basic" I mean that it doesn't support:
- blacklists
- datasets which don't have a Map visualisation defined on them. Well, I'm not really sure exactly what subset is covered.
- accessing data other than via WMS. It's a bit quick and dirty.


kring:
I think you forgot to add `SocrataCatalogGroup` to git.

stevage:
Git is hard.

Issue #93, opened 2014-06-29
Make URLs and email addresses in info panel into links


bsimyo:
Not needed for POC but a nice-to-have

PR #857, opened 2015-07-21
Basemap cookie2

stevage:


kring:
@stevage this looks good.  The only thing I might change (it can be a separate pull request if you're up for it) is to improve the symmetry of saving/loading the basemap.  Right now the basemap choice is saved by `TerriaViewer` and then loaded by app code.  I think it would be easy to write a function that wires up both and then apps can choose to call that function or not depending on whether or not they want that functionality.  The function could subscribe to change notification on `Terria.basemap` to know when to save the new base map selection in local storage.

Whether you tackle the above or not, this still needs to be wired up in National Map, right?

stevage:
Whoops, forgot to open the PR. https://github.com/NICTA/nationalmap/pull/93

And yeah, that seems sensible. Maybe even just an option you set that makes it happen for you?

kring:
Yep, just an option sounds fine.  Another thought: can you make it only store the cookie when the user actually changes the basemap?  Currently it is stored on startup, which means that all of our users will have a cookie with "Bing Maps Aerial with Labels".  That means that if/when we do change the default, anyone who has used the app before will still get Bing.

kring:
I wrote a separate issue for this: #864.

PR #858, opened 2015-07-22
Refactor Ckan handling of esri/csv/geojson/kml/wms and fix tiny geojs…

stevage:
…on bug in the process

Signed-off-by: Steve Bennett stevage@gmail.com

### DIFF
--- a/lib/Models/CkanCatalogGroup.js
+++ b/lib/Models/CkanCatalogGroup.js
@@ -300,7 +300,7 @@ function regexSerializer (fieldName) {
         if(defined(ckanGroup[fieldName])) {
             json[fieldName] = ckanGroup[fieldName].source;
         }
-    }
+    };
 }
 
 CkanCatalogGroup.defaultSerializers.wmsResourceFormat = regexSerializer('wmsResourceFormat');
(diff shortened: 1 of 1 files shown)


stevage:
Ready for review @kring

kring:
Nice, thanks!

kring:
Can you update `CHANGES.md` (in master is ok) with a short blurb about the GeoJSON bug fix?

stevage:
To be honest I don't know if it was an actual bug - I never saw any symptoms. I just noticed that the `knockout.track(` line was missing a reference to `includeGeoJson`. Hard to imagine any scenario where that would produce any noticeable effect.

That line also doesn't include any of the `geoJsonResourceFormat` type properties either. I'm a bit fuzzy on the implications of adding a property to knockout, but since there's no UI for adjusting these properties of a CkanCatalogGroup, there's probably no need for them to be tracked?

kring:
> I don't know if it was an actual bug

Ok, nevermind then.

> Hard to imagine any scenario where that would produce any noticeable effect.

You're probably right that it doesn't.  The `knockout.track` is important for things that are reflected in the UI, especially if we want the UI to automatically update when they change.  For example, for WMS datasets, we get the Description from GetCapabilities, in some cases after the dataset info panel is up, and we want the updated Description to appear as soon as the GetCapabilities response comes back.

Also, it's possible for other parts of code to subscribe to change notification on any knockout.track'd field.  This is not commonly used, but it's nice to have the option.

Finally, there's the whole `_getValuesThatInfluenceLoad` thing.  The idea is that our code can call `load` on a catalog item as often as it wants, but nothing will actually happen unless a load is needed.  And a load is only need, by definition, when the "values that influence load" change.

If the properties returned by `_getValuesThatInfluenceLoad` are observable (passed to `knockout.track`) then it's possible (but we don't do it currently) to make TerriaJS watch them and automatically reload when they change.  For example, when the value of `includeGeoJson` is changed from false to true, the group would automatically reload and now show GeoJSON items as well.

TL;DR: forgetting to include something like `includeGeoJson` in `knockout.track` has almost no user-visible impact, but it's still good to put it there when we notice it's missing.

stevage:
Got it - thanks for explaining all that.

PR #861, opened 2015-07-23
Put Description near the top of dataset info page

kring:
Before:
image

After:
image


axman6:
Is this just for the auto generated sections in the info pane? Somehow the work I did to allow you to write your own sections in in the init.json seems to be broken - the sections should be rendered in the order specified in the file, and that's what my patch implemented.

kring:
> Somehow the work I did to allow you to write your own sections in in the init.json seems to be broken

Not broken, "fixed". :)

The trouble is that it's really hard to ensure consistent order of the info sections when some of them are defined in the init file and others are determined at runtime by querying GetCapabilities (for example).  So I changed `CatalogItemInfoViewModel` to have a preferred order for known sections (see `infoSectionOrder`).  Any other sections are after the known ones and sorted alphabetically.  We can tweak this logic (e.g. add to the list of known sections, or maintain order rather than sorting alphabetically for unknown sections), but I think something like it is necessary.

axman6:
Well, that is kind of broken, because being able to specify the exact order was the reason for doing the work in the first place - we have (maybe had, but I think we still have it) a need for specifying things in a specific order (sometimes we needed to place a disclaimer before anything else, followed by the description, then legal etc.)

kring:
Understood.  Can you solve this by adding more sections to that `infoSectionOrder`?  If not, do you have another suggestion?

kring:
FWIW, I was under the impression it was more about being able to create arbitrary sections, with a sensible order, rather than about the exact order of the sections.  I apologize if I misunderstood.

kring:
Solution discussed on Slack was to allow info section order to be optionally specified in the init file.

@axman6 suggested something like this:

``` javascript
"info": [
    "Description",
    {
        "name": "Legal",
        "content": "sdfkjdsfkjgh"
    },
    "Some other WMS provided thing"
]
```

Where strings in the info list specify section order without specifying the content of those sections (because the content comes from e.g. WMS GetCapabilities).

Thinking about this more, that particular syntax makes the common case (global section ordering across datasets) complicated.  Most datasets in an init file will want to specify a section or two, but use the default order.  How do we do that with the syntax above?  Assume that if the `info` array doesn't have any strings, it's only specifying section content and the default order should be used?  Seems confusing.

IMO a better approach, even though it's a tiny bit more verbose, is an explicit `infoSectionOrder` for those (hopefully rare) datasets that need to do something other than the default:

``` javascript
"infoSectionOrder": [
    "Description",
    "Legal",
    "Some other WMS provided thing"
]
```

Anyway this can all be considered separate from this pull request, which can be merged when ready.

kring:
I implemented the above in this pull request rather than creating a new one.

meh9:
Works perfectly, thanks @kring :)

PR #863, opened 2015-07-23
Remove right align of info popup value column for TableDataSource …

meh9:
…to be consistent with other info popups.

### DIFF
--- a/lib/Map/TableDataSource.js
+++ b/lib/Map/TableDataSource.js
@@ -306,12 +306,12 @@ TableDataSource.prototype.describe = function(properties) {
                     value = JulianDate.toDate(value).toDateString();
                 }
                 if (typeof value === 'object') {
-                    html += '<tr><td>' + name + '</td><td align="right">' + this.describe(value) + '</td></tr>';
+                    html += '<tr><td>' + name + '</td><td>' + this.describe(value) + '</td></tr>';
                 } else {
                     if (typeof value === 'number') {
                         value = numberWithCommas(value);
                     }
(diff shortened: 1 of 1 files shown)


kgrochow:
That was done for a specific reason.  To make the info popup more readable and usable, especially when viewing a lot of numeric data, which is what a lot of table data sources are.  The fix that has been discussed had been to update the cesium default to make it more readable, but it has been low priority.

meh9:
Ok, understood.

Since the table data sources already have extensive configuration options for presentation, how about we extend this pull request to add an option that allows left-align of text? We can leave the default as the current right align.

kgrochow:
That sounds like a good idea.  My initial thinking is a markdown template that would allow replacement of the data_name and data_value for the row.

axman6:
All that needs to be done is to remove the align="right" from the td tags, set a class, and allow this to be set in CSS instead of hardcoded.

kring:
Yeah I agree @axman6.  I _think_ left aligning might be ok even for CSV now that the overall formatting of that table is improved (there used to be no space between the columns).  But in any case, I can't think of any reason for the alignment to be different for different types of catalog items.  So the right thing to do is to add a class to the table (no need to add it to each cell) and let the app decide.

kring:
So who wants to volunteer to do it? :)

kring:
Err nevermind, this pull request basically already does this (the class is already there, even if it has a slightly-awkward `cesium` name)...  Merging away! :)

axman6:
Thanks Kevin!

PR #866, opened 2015-07-24
Moved Reorder.svg into a js file

keattang:
Now using the js file. 
Allows for setting the colour via the fill style. 
Also fixed a bug when e.dataTransfer is undefined


lib/ViewModels/NowViewingTabViewModel.js
```diff
@@ -118,7 +120,11 @@ NowViewingTabViewModel.prototype.dragStart = function(viewModel, e) {
         e.preventDefault();
     }, false);
 
-    e.dataTransfer.setData('text', 'Dragging a Now Viewing item.');
+    if (defined(e.dataTransfer)) {
+        e.dataTransfer.setData('text', 'Dragging a Now Viewing item.');
+    } else {
+        e.originalEvent.dataTransfer.setData('text', 'Dragging a Now Viewing item.');
```
kring, line 126: Is this because you guys are using jQuery?

keattang, line 126: I wasn't sure why it was occurring but we are using jQuery

kring, line 126: Yeah jQuery inserts itself into all drag events for some reason.  It's good stuff. :)

kring:
I tweaked this to maintain the previous color of the drag handle.  You can override it by setting `@now-viewing-drag-handle-color`.

PR #871, opened 2015-07-24
Tools updates for ga datasets

kgrochow:
Updated the request loader to wait for the dataset to load before enabling.
- note: this can make the caching take longer to start since it has to query layers info from all the arcgis servers

Also added 'more info' option in data testing to return server and layers for a request.

### DIFF
--- a/lib/ViewModels/ToolsPanelViewModel.js
+++ b/lib/ViewModels/ToolsPanelViewModel.js
@@ -246,7 +246,7 @@ function getAllRequests(types, mode, requests, group, promises, blacklistGroup) 
         }
         return false;
     }
-    function generateReguest(item, group, blacklistGroup) {
+    function generateRequest(item, group, blacklistGroup) {
         return function() {
             var enabledHere = false;
             if (!item.isEnabled) {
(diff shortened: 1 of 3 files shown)


kgrochow:
it's up at http://nationalmap.research.nicta.com.au/demo/

lib/ViewModels/ToolsPanelViewModel.js
```diff
@@ -257,25 +281,11 @@ function getAllRequests(types, mode, requests, group, promises, blacklistGroup)
...
-                }
-                requests.push({
-                    item : item,
-                    blacklistGroup : blacklistGroup,
-                    group : group.name,
-                    enabledHere : enabledHere,
-                    provider : imageryProvider
-                });
+                    //this provides closure for the for loop inside the promise chain
+                var promise = when(item._load()).then( ( generateReguest ) (item, group, blacklistGroup) );
```
kring: Did you try this with `load` instead of `_load`?  I think `load` is better (assuming there isn't some non-obvious problem with it) so that it won't load more than once and will set `isLoading` in the process.  Basically, use the public API unless there's a reason we can't.

kgrochow: It didn't work previously, but I'll try it again with the final code.  imo: We're already using _enable, so _load doesn't seem that problematic in tools.

kgrochow: load appears to work similarly to _load.

PR #875, opened 2015-07-28
url shortening

kgrochow:
A set of minor edits to support url shortening.
- add urlShortener to terria at start
- add google analytics tracking of launch and share url (since our launch method isn't recorded in goo.gl)
- handle null for urlKey to deal with being set from json
- return terria error event to show error to user if can't find url from key
- move shorten checkbox to bottom and on by default if available


kring:
I made a change in `SharePopupViewModel` to persist the state of the "shorten URL" flag across sessions.  Otherwise this looks good!

PR #877, opened 2015-07-28
Drastically improve the performance of polygons in 2D

kring:
The `LeafletVisualizer` was previously doing tons of work that involved touching every vertex ever time the clock ticked, even though the vertices almost never change.


arrayjam:
If I drag in a layer myself, I can no longer toggle its visibility. I can hide it, but not show it again.

kring:
You're right, polygons aren't being cleaned up correctly anymore.  I'll have it fixed shortly.

kring:
Fixed now, sorry about that.

PR #878, opened 2015-07-28
Avoid breaking change in Google Analytics refactoring

kring:
If the `ga` function is defined in `index.html` (as it is in current versions of TerriaJS apps), an instance of `GoogleAnalytics` is created and used, so the app will work just as it did before.

Apps without a global `ga` function will get the console event logger, unless they explicitly pass something else as the `analytics` parameter to `Terria`.


kgrochow:
I'm wondering if logToConsole should be false by default with maybe an intro message in the console on how to turn on logging.  All the messages make the console a bit hard to use without turning it off.

kring:
Good idea, I changed it to not log by default.  I don't think an intro message is really necessary.

PR #879, opened 2015-07-28
Allow an ARCGis server to be defined with a single-layer URL

stevage:
like https://gisservices.scc.qld.gov.au/arcgis/rest/services/Society/Society_SCRC/MapServer/16


kring:
Nice!  @stevage mind updating CHANGES.md and putting an init file with an example in `wwwroot/test/init`?

kring:
Nevermind @stevage I took care of it.

Issue #101, opened 2014-07-01
Terrain height bug


kgrochow:
I know we have some inherent problems with terrain height, but I think there may also be a bug.  If you roam with mouse point around the coastline you get some odd numbers.   If you go to Tuross Heads, the number get smaller as you go inland and higher as you go out to sea.  It's almost like the tile is being read backwards.

PR #883, opened 2015-07-31
Fix two bugs in AbsIttCatalogItem

kring:
- Fixed a race condition in `AbsIttCatalogItem` that could cause the legend and map to show different state than the Now Viewing UI suggested.
- Fixed a bug where an ABS concept with a comma in its name (e.g. "South Eastern Europe,nfd(c)" in Country of Birth) would cause values for concept that follow to be misappropriated to the wrong concepts.

To see the bug in the second bullet, enable Country of Birth, then enable "South Eastern Europe,nfd(c)" and another after it in the list (e.g. Greece).  Then click on a region and you'll see something like this;

image

Note that TerriaJS has treated "South Eastern Europe,nfd(c)" as two separate columns and so all of the values that follow are misaligned and there's an "undefined" at the end.

Fixes NICTA/nationalmap#101

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Change Log
 ==========
 
+### 1.0.39
+
+* Fixed a race condition in `AbsIttCatalogItem` that could cause the legend and map to show different state than the Now Viewing UI suggested.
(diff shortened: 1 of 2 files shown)


kring:
@kgrochow please review this, but I'm going to merge it now to get today's release moving.

PR #890, opened 2015-08-07
Added option on catalog members to force proxying

keattang:
Allows programmatic determination of whether or not a catalog member should be proxied


kring:
@keattang thanks for the PR, and sorry for ignoring it for so long!  Can you tell me briefly when this would be desirable?  Why wouldn't you just remove the server from `corsDomains` instead?  Or do you have a case where a single server has some datasets that support CORS and others that do not?

stevage:
One use case I can think of is if you have a caching proxy.

kring:
> One use case I can think of is if you have a caching proxy.

Maybe I'm not following you.  I understand you might want to use the proxy even with a dataset that supports CORS (to get caching).  But in that scenario, why not just remove it from `corsDomains`?

keattang:
Hey Kevin, 

All good! It's been a while since I implemented this so I'm a bit vague on the detail but off the top of my head I think it was because I have some catalog items that point to external servers and therefore need proxying to get around CORS issues and some that point to internal servers and therefore do not.

However, I don't know ahead of time what the addresses for the servers that need proxying will be but, I do know which catalog items are will have them so i built the option to always proxy those catalog items.

Does that make sense? If not I'll have a closer look at what I did and get back to you.

keattang:
Did that make sense? I've brought this up to speed with some of the latest changes to the proxying code.

PR #906, opened 2015-09-08
Make timeseries autoplay optional, preserving existing autoplay behaviour.

meh9:


kring:
@meh9 you said on slack that you need to fix a bug in this before we can merge it, right?

meh9:
The problem I noticed and was attempting to fix was "skipping timeseries", but it turns out this is an existing problem and nothing caused by this PR.

Have logged the issue separately: https://github.com/TerriaJS/terriajs/issues/917

This PR should now be ready to merge.

kring:
Ok, thanks.  One last question before merging: who is tsimon?

meh9:
tsimon is Travis here in Canberra, he's also on our Slack.

PR #918, opened 2015-09-16
Expand tab for layers added via URL hash

travissimon:
When users share a map (via the 'share' button), the resulting map
should have any pre-selected layers expanded in the Catalog
control. This will provide context to the user indicating which layers
are already activated.

Resolves: https://github.com/TerriaJS/terriajs/issues/902


kring:
Closing this in favor of #919.

PR #921, opened 2015-09-17
Prevent timeseries data from skipping after Cesium has been in paused…

meh9:
… state by always tick()'ing before resuming timeseries play.

Fixes TerriaJS/terriajs#917.


kring:
Thanks @meh9!

PR #924, opened 2015-09-21
Region mapping advanced matching

stevage:
Significant refactoring of the region mapping code. Major benefits:
- Region mapping code split out from CsvCatalogItem to RegionProvider, RegionProviderList, DataVariable, DataTable and TableDataSource
- added functionality: matching by regex (enabling matching LGA names), disambiguation using a second column
- dozens of new unit tests
- fixed many edge cases, particularly with drag-and-drop CSV files (which lack explicit styling).
- temporal datasets work much better: each datapoint by default is shown until the next time point.
- new architecture will make it much easier to add new region provider types


kring:
I haven't looked at the code yet, but I ran through the CSVs in `#test` and noticed some differences.  It's possible these are intentional but I wanted to note them regardless:
- `Miscellaneous Files -> Auto Incidents No Time` now uses the `id` to color the points.  Previously all the points were the same color.
  image
  image
- `Region Mapping (CSV) -> NEXIS NSW 2012 (LGA)` and `SEIFA 2011 (POA)` no longer load:
  image
- `Temporal -> Pacific Earthquakes` shows each earthquake for a very brief time, so it's flickery.  This might be a case where showing each point until the start of the next doesn't make sense?

wwwroot/test/csv/regionMapping.json
```diff
@@ -0,0 +1,197 @@
+{
+    "comments": "Matching takes place in the order defined in this file. Place code matches before name matches, and smaller regions before larger ones.",
```
kring, line 2: As you probably know, JSON keys are unordered.  So if this currently works in all browsers it's just a coincidence.  Unless you're talking about the order in the `aliases` array, in which case maybe clarify this in the comment?

kring, line 2: Actually this may not be a new problem, eh?  Probably worth fixing in any case.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    * URL of the WMS server
+    * @type {String}
+    */
+    this.server = properties.server;
+    /**
+     * List of aliases which will be matched against if found as column headings.
+     * @type {String[]}
+     */
+    this.aliases = defaultValue(properties.aliases, [this.regiontype]);
+    this.serverReplacements = properties.serverReplacements;
```
kring: It'd be helpful to have doc for these three properties.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+     * @type {Object[]}
+     */
+    this.regions = [];
+
+
+};
+
+/**
+ * Turn a WFS list of attribute values into a straight array
+ * @param {String} propertyName: optional, for setting an additional (disambiguating) property on a region
```
kring: Optional properties are indicated in jsdoc by enclosing them in square brackets.  E.g. `[propertyName]` or `[propertyName='foo']` (to state the default value when a value isn't supplied).

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    }
+    for (var i = 0; i < obj.member.length; i++) {
+        if (this.regions[i] === undefined) {
+            this.regions[i] = {};
+        }
+        // either { id: 123 } or { id: 'Bunbury', STE_NAME: 'VIC' }
+        if (!propertyName) {
+            this.regions[i].id = obj.member[i][this.regionProp];
+        } else {
+            this.regions[i][propertyName] = obj.member[i][propertyName];
```
kring: Not sure if I'm following this right, but isn't it necessary to set the value of the `id` property even when a `propertyName` is supplied?

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    var baseuri = URI(this.server).addQuery({ 
+        service: 'wfs',
+        version: '2.0',
+        request: 'getPropertyValue',
+        typenames: this.layerName});
+
+    // get the list of IDs that we will attempt to match against for this column
+    var p = [];
+    var url = baseuri.addSearch('valueReference', this.regionProp).toString();
+    p.push(loadText(corsProxy.getURL(url))
```
kring: It's unfortunate, but `corsProxy.getURL` always proxies the URL even if it is not necessary to do so.  That's why we usually call `corsProxy.shouldUseProxy` first.  A helper function would be great, along the lines of `proxyCatalogItemUrl`.

lib/Models/RegionProvider.js
kring: Nevermind, I see now that the `id` and `propertyName` are set on separate invocations of the function from `loadRegionIDs`.  But it does rely on the WFS `getPropertyValue` returning the same features in the same order in two successive calls to `getPropertyValue`.  Is there a better way to do it?

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+            console.log(err);
+            throw(err);
+        }));
+    // if this column might be ambiguous then fetch the disambiguating values for each column as well (usually State)
+    var dab = this.disambigColumns;
+    if (defined(dab)) {
+        
+        url = baseuri.addQuery('valueReference' + dab[0].regionProp).toString();
+        p.push(loadText(corsProxy.getURL(url)).then(function(xml) {
+            that.loadRegionsFromXML(xml, dab[0].regionProp);
```
kring: Only the _first_ disambiguation column is used here, and that appears to be true everywhere else `disambigColumns` is used in this file as well.  Why take an array and then only use the first element?

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+// replacements: [ [ regex, replacement], ... ] 
+// replacement can contain '$1' etc.
+    var r = s.trim().toLowerCase(); // just in case...
+
+    if (replacements === undefined || replacements.length === 0) {
+        return r;
+    }
+
+    var cachekey = s + '%' + replacements.reduce(function(key, item) { 
+        return key + item[0] + '%' + (item[1] ? item[1] : '') + '%';
```
kring: How about using `JSON.stringify(replacements)` as the cache key?

kring: Sorry, `s` is in there too.  So `JSON.stringify({s: s, replacements: replacements})`.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    return r;
+}
+applyReplacements.cache = {};
+
+
+/* Checks if a given region code matches a given ID after applying regexs and additional disambiguation lookups */
+RegionProvider.prototype.codeMatchesRegionID = function(id, code, disambigId, disambigCode) { //, row, idnum, csvItem ) {
+    code = String(code).toLowerCase(); //## should be happening upstream somewhere
+    id = String(id).toLowerCase(); 
+    if (applyReplacements(id, this.serverReplacements) !== applyReplacements(code, this.dataReplacements)) {
```
kring: Couldn't you do the `serverReplacements` once instead of doing it while testing each code?  Then your replacement cache, if you need it at all after turning this `n*n` operation into an `n` operation, could just be a map of strings (pre-replacement codes) to region indices.

kring: Also can you come up with clearer names than `id` and `code` so that it's clearer that `id` is the actual ID of a region on the server and `code` is a value in a column that hopefully matches one of the known region IDs?

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    if (defined(disambigId) && !defined(disambigCode)) {
+        // warn? we have a value to disambiguate against, but it's not provided
+        return true;
+    }
+
+    // Hooray, let's disambiguate. Our args look something like: Campbelltown, Campbelltown, New South Wales, NSW.
+
+    var dabMatched = this._getDisambigProvider().codeMatchesRegionID(disambigId, disambigCode);
```
kring: It seems a bit awkward for each region provider to keep a reference back to the list of providers that contains it just so that it can go find its disambiguation provider.

It would be better, I think, if the `regionProviderList` were passed down as parameters to the functions along the way, rather than given to the `RegionProvider` constructor and stored.

Also, I suspect that looking up the disambig provider with a a linear search on every attempted match is making the matching go quite a bit slower than it needs to.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+ *   @returns {Object} Object containing properties:
+ *      regionValues: Sparse array of values corresponding to the provided list.
+ *      ambiguousMatches: Object whose keys are IDs of regions that had more than one row match against them.
+ *      failedMatches: Object whose keys are unmatched region codes.
+ *      successes: Number of successful matches.
+ *      totalRows: Number of attempted matches.
+ */
+RegionProvider.prototype.getRegionValues = function (dataset, regionVariable, disambigVariable, rowList) {
+    if (this.regions.length < 1) {
+        // should this return a promise instead of throwing?
```
kring: I don't think so, it looks like a synchronous function to me.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    var disambigCodes = (defined(disambigVariable) ? dataset.getVariableEnums(disambigVariable) : []);
+    var vals = dataset.getVariableValues(dataset.getDataVariable());
+    if (!defined(vals)) {
+        console.log('Warning: no data variable for region-mapped dataset with region variable "' + regionVariable + '"');
+    }
+
+    var numericCodes = false;
+    if (!defined(codes)) {
+        // not an enum type? the IDs must be digit-based and have been misclassified as numbers
+        // ##TODO think about whether this matters for disambigs
```
kring: Let's try to avoid TODOs in the code.  This sounds like something that should be considered before we merge this.  In other cases, TODOs can be turned into GitHub issues instead.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+            return i;
+        }
+    }
+    return -1;
+};
+
+/**
+ *   Provides a list of values for each region, based on a provided dataset.
+ *
+ *   @param {Dataset} dataset
```
kring: I don't see a type called `Dataset`.  `DataTable` maybe?

lib/Models/RegionProvider.js
stevage: Agreed - in fact, as I progressed, I started backtracking on actually needing to store that reference. 

I think actually the DataTable should know what the RegionVariable and DisambigVariable are, and pass those references around as necessary.

stevage:
> Miscellaneous Files -> Auto Incidents No Time now uses the id to color the points. 

Previously lat-long CSVs didn't support enums - which I think they ought to. The question now is really, "for how many distinct values does colour coding not make sense for enums" or "are there fields which should not be colour coded on"?

So, yes, it's intentional, but there are cases where the new behaviour is undesirable.

> Temporal -> Pacific Earthquakes shows each earthquake for a very brief time, so it's flickery. This might be a case where showing each point until the start of the next doesn't make sense?

Yeah. I suspect that the default behaviour was probably crafted around this dataset, or ones very similar to it. Options:
- specify an explicit `displayDuration` value for this dataset
- have a minimum display duration for any point

> Region Mapping (CSV) -> NEXIS NSW 2012 (LGA) and SEIFA 2011 (POA) no longer load: 

That's because you're testing in NM, which has a `regionMapping.json` which is incompatible. Try copying over the one from `node_modules/terriajs/wwwroot/test/csv/regionMapping.json`.

lib/Models/RegionProvider.js
stevage: Fixed. There seems to be a convention of variables called `dataset` referring to `DataTable` objects.

lib/Models/RegionProvider.js
stevage: Fixed.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+ * Pre-generates a function which quickly turns a value into a colour.
+ * @param {Number[]} regionValues Array of values.
+ * @param {Function} colorFunc should be function(val) { return [r,g,b,a]; } 
+ * @returns {Function} Function of type f(regionIndex) { return [r,g,b,a]; }
+ */
+RegionProvider.prototype.getColorLookupFunc = function(regionValues, colorFunc) {
+    var colors = regionValues.map(colorFunc);
+    return function(regionIndex) {
+        // actually maybe it's ok to return undefined after all
+        return colors[regionIndex] !== undefined ? colors[regionIndex] : [0,0,0,0]; //# is it important to get rid of this if?
```
kring: These comments are a bit dodgy. :)

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+    for (var j = 0; j < this.aliases.length; j++) {
+        for (var i = 0; i < varNames.length; i++) {
+            //check that it is the right type of code
+            //if (typeof this.regionTextCode === "boolean" && this.regionTextCode !== variables[i].isTextCode) {
+            //    continue;
+            //}
+            // SB: I don't think that's a worthwhile test
+
+            //if (varNames[i].toLowerCase() === this.aliases[j]) {
+            if (new RegExp('^' + this.aliases[j] + '$', 'i').test(varNames[i])) {
```
kring: Hoist the construction of the regex out of the inner loop.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+ * @param varNAmes Array of variable names.
+ */
+RegionProvider.prototype.findRegionVariable = function(varNames) {
+    for (var j = 0; j < this.aliases.length; j++) {
+        for (var i = 0; i < varNames.length; i++) {
+            //check that it is the right type of code
+            //if (typeof this.regionTextCode === "boolean" && this.regionTextCode !== variables[i].isTextCode) {
+            //    continue;
+            //}
+            // SB: I don't think that's a worthwhile test
```
kring: I don't understand the question well enough to say if I agree with you, but if you're sure then just delete the code and comment.

lib/Models/RegionProvider.js
stevage: > Couldn't you do the serverReplacements once instead of doing it while testing each code?

Ah, yep. Logical next refinement. Basically just moving the cache up a level, so it's a simple lookup. Nice catch.

lib/Models/RegionProvider.js
```diff
@@ -0,0 +1,405 @@
...
+ */
+RegionProvider.prototype.findDisambigVariable = function(variables) {
+    var dabp = this._getDisambigProvider();
+    if (!defined(dabp)) {
+        return undefined;
+    }
+    var dv = variables[dabp.findRegionVariable(variables)];
+    return (defined(dv) ? dv.name : undefined);
+};
+// this probably doesn't belong here
```
kring: yeah, I agree.  It's ok for `RegionProvider` to offer up this information in an object literal or something like that, but it shouldn't be formatting it as HTML.

lib/Models/RegionProviderList.js
```diff
@@ -0,0 +1,135 @@
...
+
+
+/**
+  RegionProviderList encapsulates the regionMapping.json file and provides support for choosing the best region
+  provider for a given dataset.
+*/
+
+var RegionProviderList = function(terria) {
+    /** 
+     * URL pointing to a JSON file where the region provider definitions will be loaded from.
```
kring: Doc is wrong.

lib/Models/RegionProviderList.js
```diff
@@ -0,0 +1,135 @@
...
+  RegionProviderList encapsulates the regionMapping.json file and provides support for choosing the best region
+  provider for a given dataset.
+*/
+
+var RegionProviderList = function(terria) {
+    /** 
+     * URL pointing to a JSON file where the region provider definitions will be loaded from. 
+     * @type {String}
+     */
+    this.terria = terria;
```
kring: This isn't used for anything other than getting the `regionMappingDefinitionsUrl`.  `RegionProvider` doesn't really use `Terria` or any other models, either.

So I think it would make sense to move both `RegionProvider` and `RegionProviderList` to `Map`.

Then, change `RegionProviderList` to take the JSON region definition (like `initFromObject`), remove the `_initialized` logic entirely, and add a static `fromRegionMappingDefinitionsUrl` function that returns a promise for a `RegionProviderList`.  `fromRegionMappingDefinitionsUrl` can also cache the loaded instance so that multiple `CsvCatalogItem`s only need to load it once.

test/Models/CsvCatalogItemSpec.js
```diff
@@ -94,31 +111,318 @@ describe('CsvCatalogItem', function() {
...
+            expect(csvItem._tableDataSource.dataset).toBeDefined();
+            expect(csvItem._tableDataSource.dataset.getRowCount()).toEqual(2);
+        }).otherwise(fail).then(done);
     });
 
-    it('is correctly loading csv data from text', function() {
-        expect(csvItem instanceof CatalogItem).toBe(true);
+    it('throws an error on non-csv file', function(done) {
+        csvItem.url = 'test/GeoJSON/polygon.topojson';
+        expect(true).toBe(true);
```
kring: Remove this.

test/Models/CsvCatalogItemSpec.js
```diff
@@ -94,31 +111,318 @@ describe('CsvCatalogItem', function() {
...
     });
 
-    it('returns an error on non-csv text', function() {
-        expect(csvItem instanceof CatalogItem).toBe(true);
+    it('matches LGAs by code', function(done) {
+        csvItem.updateFromJson( { data: 'lga_code,value\n31000,1' });
+        csvItem.load().then(function() {
+            expect(csvItem._regionMapped).toBe(true);
+            expect(csvItem.colorFunc).toBeDefined();
+            // 242 is the shapefile index of LGA boundary 31000. What a crappy way to test...
```
kring, line 157: Can you look at the created entities maybe?  `expect(csvItem._tableDataSource.czmlDataSource.entites.values[0].point.color.getValue()).not.toEqual(new Color(0,0,0,0))`.  Still pretty ugly but at least there's no magic 242 involved.

test/Models/CsvCatalogItemSpec.js
```diff
@@ -94,31 +111,318 @@ describe('CsvCatalogItem', function() {
...
     });
+    it('matches LGAs by names in various formats', function(done) {
+        csvItem.updateFromJson( { data: 'lga_name,value\nCity of Melbourne,1\nGreater Geelong,2\nSydney (S),3' });
+        csvItem.load().then(function() {
+            expect(csvItem._regionMapped).toBe(true);
+            expect(csvItem.colorFunc).toBeDefined();
+            expect(csvItem.colorFunc(121)).not.toEqual([0,0,0,0]);
+            expect(csvItem.colorFunc(180)).not.toEqual([0,0,0,0]);
+            expect(csvItem.colorFunc(197)).not.toEqual([0,0,0,0]);
+        }).otherwise(fail).then(function() { done(); });
```
kring: No reason for the anonymous function, right?

test/Models/CsvCatalogItemSpec.js
```diff
@@ -94,31 +111,318 @@ describe('CsvCatalogItem', function() {
...
+
+    });
+
+    it('matches SA4s', function(done) {
+        csvItem.updateFromJson( { data: 'sa4,value\n209,correct' });
+        csvItem.load().then(function() {
+            expect(csvItem._regionMapped).toBe(true);
+            expect(csvItem.colorFunc).toBeDefined();
+            expect(csvItem.rowProperties(209).value).toBe('correct');
+        }).otherwise(fail).then(function() { done(); });
```
kring: Same here.

test/Models/CsvCatalogItemSpec.js
kring, line 157: Actually that can be slightly simplified:
`expect(csvItem._tableDataSource.entites.values[0].point.color.getValue()).not.toEqual(new Color(0,0,0,0))`

test/Models/CsvCatalogItemSpec.js
```diff
@@ -94,31 +111,318 @@ describe('CsvCatalogItem', function() {
...
+            expect(source.dataset.hasTimeData()).toBe(true);
+            expect(source.getDataPointList(j('2015-07-31')).length).toBe(0);
+            expect(source.getDataPointList(j('2015-08-01')).length).toBe(2);
+            expect(source.getDataPointList(j('2015-08-02')).length).toBe(3);
+            expect(source.getDataPointList(j('2015-08-06')).length).toBe(2);
+            expect(source.getDataPointList(j('2015-08-07')).length).toBe(0);
+        }).otherwise(fail).then(done);
+    });
+    /*
+    Nope - I don't know how feature picking on lat-longs works.
```
kring: Yeah, it's tricky.  Lat-long files create actual vector points (attached to Cesium entities), which are then rendered and picked differently in the two clients:
- In Leaflet, `LeafletVisualizer` turns the point into a circle marker, which is an actual browser DOM element.  The browser detects a click on the circle marker DOM element and raises an event on `LeafletScene` that causes `featurePicked` in `Leaflet` to be invoked.
- In Cesium, our `Cesium.js` handles left clicks by calling `drillPick` on Cesium's `Scene`.  `drillPick` does one or more actual renders of the scene to an off-screen buffer in a fancy way so that the color of the rendered pixel indicates the ID of the vector feature that was picked.

Neither of these can be tested without the actual Leaflet / Cesium machinery up and running.

So, instead of testing picking, I recommend that you simply check that entities with the correct properties were created (one of my comments above has an example of checking the color of an entity's point) and trust that correctly-created entities will be rendered and will be pickable.

test/Models/CsvCatalogItemSpec.js
```diff
@@ -180,7 +484,8 @@ describe('CsvCatalogItem', function() {
             done();
         });
     });
-
+    /*
+    // Removed: not clear that this is correct behaviour, and it's failing.
```
kring, line 453: I don't know if transparent black is the right behavior.... but it's certainly possible to have a CSV file where the value column for most rows has a value, but it's missing for a few.  So what should happen in that case?  Transparent black (i.e. don't show the point or region) seems reasonable, but maybe there are other reasonable options.  In any case, being clear about what happens and writing a test for it seems worthwhile.

test/Models/RegionProviderListSpec.js
```diff
@@ -0,0 +1,112 @@
...
+var DataTable = require('../../lib/Map/DataTable.js');
+var terria;
+var rpl;
+
+beforeEach(function() {
+    terria = new Terria({
+        baseUrl: './',
+        regionMappingDefinitionsUrl: 'test/csv/regionMapping.json',
+    });
+    terria.corsProxy.baseProxyUrl = ""; // there is no localhost:3002/proxy, so ...
```
kring: Is this necessary because we're calling `corsProxy.getURL` directly somewhere, so it's always trying to proxy even when it shouldn't?

test/Models/RegionProviderListSpec.js
```diff
@@ -0,0 +1,112 @@
...
+
+describe('RegionProviderList', function() {
+  it('loads some region providers', function(done) {
+    rpl.init().then(function(x) {
+      expect(rpl.regionProviders.length).toBeGreaterThan(2);
+    }).otherwise(fail).then(done);
+  }); 
+  /*
+  Test fails...and do we care anyway?
+  it('does not allow duplicate identifiers', function(done) {
```
kring: apparently `JSON.parse` doesn't fail if the JSON string contains duplicate keys.  But if we change to using an array instead of an object in order to preserve order, it could be worthwhile (but not essential) to check for duplicate IDs.

lib/Models/RegionProviderList.js
```diff
@@ -0,0 +1,135 @@
...
+*/
+RegionProviderList.prototype.getRegionProvider = function(regionType) {
+    if (!this._initialized) {
+        throw new DeveloperError('Region mapping list not initialized.');
+    }
+    var r = this.regionProviders.filter(function(p) { return p.regionType.toLowerCase() === regionType.toLowerCase(); });
+    if (r.length > 1)
+        throw new DeveloperError ('More than one definition of region provider: ' + regionType);
+    if (r.length === 0)
+        return null;
```
kring: We usually use `undefined` instead of `null`.  JS conveniently has two "there's no object here" values.  Using one consistently (undefined) means we don't have to check for both all over the place.

lib/Map/DataVariable.js
```diff
@@ -65,29 +80,16 @@ DataVariable.prototype.update = function () {
...
-        { hints: ['time', 'date'], type: VarType.TIME }];
-
-    for (var vt in hintSet) {
-        if (matchColumn(this.name, hintSet[vt].hints)) {
-            this.varType = hintSet[vt].type;
+        { hint: /^(lon|longitude|lng)$/i, type: VarType.LON },
+        { hint: /^(lat|latitude)$/i, type: VarType.LAT },
+        { hint: /^(.*[_ ])?(depth|height|elevation)$/i, type: VarType.ALT },
+        { hint: /^(.*[_ ])?(time|date)$/i, type: VarType.TIME },
+        { hint: /^postcode|poa|(.*_code)$/i, type: VarType.ENUM }]; //## This doesn't actually prevent misparsing.
```
kring: Not sure what the comment means.

lib/Map/TableDataSource.js
```diff
@@ -230,10 +230,81 @@ TableDataSource.prototype.loadUrl = function (url) {
...
+    var times = {};
+    for (var i = 0; i < pointList.length; i++) {
+        times[pointList[i].time.toString()] = true;
+    }
+    this._timeSlices = Object.keys(times).map(function(x) { return JulianDate.fromIso8601(x); } ); 
+
+    this._timeSlices = this._timeSlices.sort(function(a,b) {
+        if (a.equals(b))
+            return 0; // probably can't happen
+        return JulianDate.lessThan(a,b) ? -1 : 1;
```
kring: Just use `JulianDate.compare`.

kring: Well this is one way to remove duplicates, but it also round-trips every date through a string and allocates a bunch of memory and stuff..  How about something like this (not tested):

```
this._timeSlices = pointList.map(function(point) { return item.time; });
this._timeSlices.sort(JulianDate.compare);
this._timeSlices = this._timeSlices.filter(function(element, index, array) { return index === 0 || !JulianDate.equals(array[index-1], element); });
```

The `filter` creates an extra copy still (the elements could be removed in place) but it's still faster and clearer that the original implementation I think.

lib/Map/TableDataSource.js
```diff
@@ -230,10 +230,81 @@ TableDataSource.prototype.loadUrl = function (url) {
...
+    });
+    return this._timeSlices;
+};
+
+/**
+ * Returns a { start, finish } pair of JulianDates corresponding to some pair of rows around this time. 
+ */
+TableDataSource.prototype.getTimeSlice = function(time) {
```
kring, line 272: Have you seen Cesium's `TimeInterval` and `TimeIntervalCollection` types?  I think they may simplify this logic quite a bit.

lib/Models/CkanCatalogGroup.js
```diff
@@ -329,8 +329,7 @@ CkanCatalogGroup.prototype._load = function() {
 
     var promises = [];
     for (var i = 0; i < this.filterQuery.length; i++) {
-        var url = cleanAndProxyUrl(this, this.url) + '/api/3/action/package_search?rows=100000&' + this.filterQuery[i];
-
+        var url = cleanAndProxyUrl( this.terria, this.url) + 'api/3/action/package_search?rows=100000&' + this.filterQuery[i];
```
kring: Is this a bad merge?  I believe the original code is correct and the new code is wrong.

lib/Models/CsvCatalogItem.js
```diff
@@ -12,14 +12,13 @@ var freezeObject = require('terriajs-cesium/Source/Core/freezeObject');
...
 var loadText = require('terriajs-cesium/Source/Core/loadText');
-var loadJson = require('terriajs-cesium/Source/Core/loadJson');
 var WebMapServiceImageryProvider = require('terriajs-cesium/Source/Scene/WebMapServiceImageryProvider');
 var WebMapServiceCatalogItem = require('./WebMapServiceCatalogItem');
 var WebMercatorTilingScheme = require('terriajs-cesium/Source/Core/WebMercatorTilingScheme');
 var when = require('terriajs-cesium/Source/ThirdParty/when');
 
 var CatalogItem = require('./CatalogItem');
-var corsProxy = require('../Core/corsProxy');
+//var corsProxy = require('../Core/corsProxy'); // really not used?
```
kring: Yes, because it uses `proxyCatalogItemUrl` now.  Please remove this line.

lib/Models/CsvCatalogItem.js
```diff
@@ -53,14 +55,15 @@ var CsvCatalogItem = function(terria, url) {
...
-    this._regionWmsMap = undefined;
-    this._regionDescriptor = undefined;
     this._regionMapped = false;
 
     this._csvDataset = undefined;
 
     this._clockTickUnsubscribe = undefined;
 
+    // a function that returns a colour for a given index.
+    this.colorFunc = undefined;
```
kring: I don't think this should be a public property.  Public properties get serialized for sharing, for one thing.

lib/Models/RegionProvider.js
stevage: `JSON.stringify` turns regexes into `{}`, but good point.

lib/Models/RegionProvider.js
stevage: Yeah, initially I was thinking there would be multiple levels of disambiguation (suburb, state, country?), but I don't think any use cases warrant it at this stage. The disambiguation stuff is a bit more general than it needs to be.

lib/Models/RegionProvider.js
kring: Despite the name, the regexes are actually strings that will be turned into regexes, not actual `RegExp` instances, right?

lib/Models/RegionProvider.js
kring: Probably worthwhile to simplify it then, no?

kring:
> Previously lat-long CSVs didn't support enums - which I think they ought to. The question now is really, "for how many distinct values does colour coding not make sense for enums" or "are there fields which should not be colour coded on"?
> 
> So, yes, it's intentional, but there are cases where the new behaviour is undesirable.

Yeah I don't have any major problem with the new behavior, especially if it's possible to put something in the init file to make it draw all the points with a single color (is it?).

kring:
> Yeah. I suspect that the default behaviour was probably crafted around this dataset, or ones very similar to it. Options:
> - specify an explicit displayDuration value for this dataset
> - have a minimum display duration for any point

The second point sounds useful, but I'm ok with the first point being the answer for now.

kring:
> That's because you're testing in NM, which has a regionMapping.json which is incompatible. Try copying over the one from node_modules/terriajs/wwwroot/test/csv/regionMapping.json.

Thanks, it works now.  However, I notice that loading `SEIFA 2011` now locks up the UI for several seconds while it loads.  It used to load almost instantly.

kring:
That's all I have for now.  Thanks for doing this @stevage, it looks like a nice improvement!

lib/Models/RegionProviderList.js
stevage: Great ideas, done.

test/Models/CsvCatalogItemSpec.js
stevage: Right. I was [going a little crazy](http://stackoverflow.com/questions/32431544/idiomatically-testing-when-js-promises-with-jasmine/32433098?noredirect=1#comment52730976_32433098) trying to get Jasmine and (our version of) When to play together.

test/Models/CsvCatalogItemSpec.js
stevage, line 453: See https://github.com/TerriaJS/terriajs/issues/928

lib/Models/RegionProviderList.js
stevage: Agreed. Changed.

lib/Map/DataVariable.js
stevage: I was trying to figure out why numeric codes were being coerced into integers. Figured it out. Fixed it. Comment gone now.

lib/Map/TableDataSource.js
stevage: Yep, that's better, fixed.

lib/Models/CkanCatalogGroup.js
stevage: Thanks, yep. The change from `/api` to `api/` is intentional - totally unrelated to this branch, but necessary (vanilla CKANs don't like foo//api URLs).

lib/Models/CsvCatalogItem.js
stevage: Done.

lib/Models/CsvCatalogItem.js
stevage: Is renaming it to `_colorFunc` sufficient?

kring: Yep.  The serialization code ignores properties that start with an underscore.

stevage:
> Yeah I don't have any major problem with the new behavior, especially if it's possible to put something in the init file to make it draw all the points with a single color (is it?).

Yes - manually confirmed, will add a test case. Will also update the sample file.

> The second point sounds useful, but I'm ok with the first point being the answer for now.

Ok, will update this one too.
https://github.com/NICTA/nationalmap/pull/114

wwwroot/test/csv/regionMapping.json
stevage, line 2: > As you probably know, JSON keys are unordered

Yeah, good point. The truth seems to be a lot murkier than that simple statement. To fix it will require backwardly incompatible changes so I won't try to squeeze it in here. https://github.com/TerriaJS/terriajs/issues/930

lib/Models/RegionProvider.js
stevage: Done. Unfortunately that reminded me how silly the disambig data structure is.

lib/Models/RegionProvider.js
stevage: Thanks, done.

lib/Models/RegionProvider.js
stevage: Definitely unfortunate, I just stripped out the references to `terria` ;) My short term fix is to never proxy - our region mapping server(s) support CORS.

lib/Models/RegionProvider.js
stevage: Yep.

lib/Models/RegionProvider.js
stevage: Oh, good point. Done. (The cache is much less important now that server side matching happens up front. It was critical when I first implemented it.)

lib/Models/RegionProvider.js
kring: > My short term fix is to never proxy - our region mapping server(s) support CORS.

This will break region mapping in IE9, because IE9 doesn't support CORS at all.

lib/Models/RegionProvider.js
stevage: Done.

lib/Models/RegionProvider.js
stevage: Nice. Done.

lib/Models/RegionProvider.js
stevage: Done.

lib/Models/RegionProviderList.js
stevage: Err, which bit?

stevage:
Just keeping track of comments/issues not resolved yet.

### High priority to fix
- DONE the data structure for disambigColumns is unnecessarily convoluted.
- DONE broken test case for fuzzy-matched feature picking

### Should be fixed, but not as urgent
- SEMI-DONE https://github.com/TerriaJS/terriajs/pull/924#discussion-diff-40044160 "But it does rely on the WFS getPropertyValue returning the same features in the same order in two successive calls to getPropertyValue. Is there a better way to do it?"
  -- it makes that assumption much more explicit now. Any alternative way would generate more traffic.
- "Also can you come up with clearer names than id and code so that it's clearer that id is the actual ID of a region on the server and code is a value in a column that hopefully matches one of the known region IDs?"
- DONE RegionProvider shouldn't generate HTML. https://github.com/TerriaJS/terriajs/pull/924#discussion-diff-40047839
- DONE https://github.com/TerriaJS/terriajs/pull/924#discussion_r40051096 It seems a bit awkward for each region provider to keep a reference back to the list of providers that contains it just so that it can go find its disambiguation provider.
- DONE RegionProviderList doc https://github.com/TerriaJS/terriajs/pull/924#discussion_r40047951
- DONE "However, I notice that loading SEIFA 2011 now locks up the UI for several seconds while it loads. It used to load almost instantly."

### Nice to have
- https://github.com/TerriaJS/terriajs/pull/924#discussion_r40050546 >Have you seen Cesium's TimeInterval and TimeIntervalCollection types? I think they may simplify this logic quite a bit.
- DONE How to test lat-long feature picking: https://github.com/TerriaJS/terriajs/pull/924#discussion-diff-40049525
- Behaviour of no-data rows in region-mapped CSVs #928 
- DONE Simplify disambig stuff: https://github.com/TerriaJS/terriajs/pull/924#discussion_r40051096
- DONE disambig-related TODO https://github.com/TerriaJS/terriajs/pull/924#discussion_r40046629

test/Models/CsvCatalogItemSpec.js
stevage: Done.

test/Models/CsvCatalogItemSpec.js
stevage, line 157: Done.

lib/Models/RegionProviderList.js
kring: The doc says (said?) it's a URL, but it's actually a Terria instance.

test/Models/RegionProviderListSpec.js
stevage: Yep, removed.

stevage:
> However, I notice that loading SEIFA 2011 now locks up the UI for several seconds while it loads. It used to load almost instantly.

That's annoying. It's not bad on my system (just over 1 second). Probably there are opportunities to streamline the matching process.

kring:
> That's annoying. It's not bad on my system (just over 1 second). Probably there are opportunities to streamline the matching process.

It was closer to 5 seconds on my system, but I tried again just now and it's more like 2-3 seconds.  Maybe you optimized it a bit?  In any case, this is only the first time it's enabled after a fresh reload; after that it's fast.  Could that be why you're seeing only 1 second?  This machine isn't exactly slow...

stevage:
Oh, another possible reason is that the test `regionMapping.json` uses our `regionmap-dev.` server, which I think is less powerful than our `geoserver.` server, which Peter recently upgraded.

kring:
> Oh, another possible reason is that the test regionMapping.json uses our regionmap-dev. server, which I think is less powerful than our geoserver. server, which Peter recently upgraded.

I don't think so.  My browser is locked up during the 2-3 seconds, which means it's actively doing something CPU intensive, not waiting for a remote server.

kring:
Looks like sharing doesn't work with (some?) `CsvCatalogItem`s at the moment.  Add this file by URL using the Add Data panel: `test/NSW_LGA_NEXIS_201212.csv` and then hit the share button.

stevage:
> Looks like sharing doesn't work with (some?) CsvCatalogItems at the moment.

Caused by circular object, fortunately fixed by removing back reference to RegionProviderList from RegionProvider.

kring:
> Caused by circular object, fortunately fixed by removing back reference to RegionProviderList from RegionProvider.

Hmm I'd argue we shouldn't even be trying to serialize `RegionProvider` and `RegionProviderList` for sharing.  Is there a public property somewhere that shouldn't be public?

stevage:
> Hmm I'd argue we shouldn't even be trying to serialize RegionProvider and RegionProviderList for sharing. Is there a public property somewhere that shouldn't be public?

Fixed.

kring:
There are still performance problems here.  Enable the `Age` layer under ABS, and then switch to SA2.  On my system the app freezes for about 5 seconds.  On nationalmap.gov.au there is no freezing.

kring:
Here's where the time is spent:
image

kring:
Some ideas for making `codeMatchesRegionID` much faster:
- Don't do `trim`, `toLowerCase`, etc. any more than necessary.  Those operations add up because they make a copy of the string.
- When IDs are numeric (as they are in the SAx case), don't convert them to strings.  Comparing numbers is much faster than comparing strings.

kring:
And one more:
- Construct `RegExp` instances once.  Currently a new instance is constructed for each regex string for every attempted match.  Unless the browser is being clever, this should result in a _huge_ speedup.  Even if the browser is clever, it'll still be noticeably faster.

This one won't help with the ABS case, though, because there are no replacements.

stevage:
> There are still performance problems here. Enable the Age layer under ABS, and then switch to SA2. On my system the app freezes for about 5 seconds.

This is intriguing to me - I don't get this on my Macbook Pro. I enable the layer, and immediately grab the map and start panning around. It takes maybe 2-3 seconds for all the SA2s to display, but the app is responsive and panning during that time. There are two brief glitches (<0.5 seconds) when you could say it's "frozen", but nothing like what you're seeing.

I'll still try to fix it. :)

stevage:
Ok, the really inefficient bit was applying all the replacements to each region code every time it was checked against each server side ID. That was dumb.

Currently on my Macbook, the whole call to `updateRegionMapping` takes 150ms for the ABS SA2 case. And the whole block of `loadRegionsFromXML` + `updateRegionMapping` + `finishTableLoad` (everything under `loadWithXhr.load.xhr.onload`) takes 215ms. 

The equivalent running on a fresh master build on my machine is actually slower: 250ms. Significantly faster running on nationalmap.research.nicta.com.au, which suggests that the `gulp release` process is actually doing something good :)

stevage:
> Don't do trim, toLowerCase, etc. any more than necessary. Those operations add up because they make a copy of the string.

Removed a few of these.

> When IDs are numeric (as they are in the SAx case), don't convert them to strings. Comparing numbers is much faster than comparing strings.

"Numeric" IDs come out of xml2json as strings. It's possible to convert them there to integers, but there are messy edge cases (leading zeroes) to deal with. Leaving this out for now.

> Construct RegExp instances once.

Done.

Thanks very much for these tips btw, I'm definitely learning a lot about writing less sucky JavaScript :)

PR #929, opened 2015-09-22
Render JSON object data as tables in the feature info popup.

meh9:
If the supplied test data is not enough, then we can write more docs for this in the future.


meh9:
Leaving `jsonTreeview` branch in place as we will depend on that until we can remove our dependency on this functionality.

PR #931, opened 2015-09-23
Make handleInitialMessage() actually call the callback if no message …

meh9:
…is to be shown.

### DIFF
--- a/lib/ViewModels/DataCatalogTabViewModel.js
+++ b/lib/ViewModels/DataCatalogTabViewModel.js
@@ -101,7 +101,6 @@ function handleInitialMessage(viewModel, item, successCallback) {
     }
 
     if (keySpecified) {
-        item.initialMessageViewed = true;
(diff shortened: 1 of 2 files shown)


lib/ViewModels/DataCatalogTabViewModel.js
```diff
@@ -47,7 +47,7 @@ DataCatalogTabViewModel.prototype.showInfo = function(item) {
 DataCatalogTabViewModel.prototype.clickEnabled = function(item) {
     item.terria.analytics.logEvent('dataSource', 'clickItem', item.name);
 
-    if (defined(item.initialMessage) && !item.isEnabled && !item.initialMessageViewed) {
```
kring, line 50: Doesn't this `!item.initialMessageViewed` bit need to be moved somewhere else?  Or if it's really not needed anymore, the property should be removed from `CatalogMember`.

meh9:
`initialMessageViewed` has been removed.

PR #933, opened 2015-09-23
Add WfsFeaturesCatalogGroup

kring:
This new catalog group, when opened, queries a WFS server and creates a catalog item for each of the features returned by the server.  The features are optionally grouped by a given property of the features.  When you're set up with this branch, you can try it at this URL:
http://localhost:3001/#build/TerriaJS/test/init/wfs.json


stevage:
Looks fine to me (with added cleanAndProxyUrl, unless I misunderstood...)

Just a comment on the `type` of `wfs-features-group`. The range of types is getting a bit ugly and inconsistent:
- `wfs`
- `wfs-getCapabilities`
- `wfs-features-group`

Ignoring backward compatibility issues, these would be clearer:
- `wfs-feature-type`
- `wfs-feature-type-group`
- `wfs-feature-group`

And similarly renaming the classes:
- `WfsFeatureTypeCatalogItem`
- `WfsFeatureTypeCatalogGroup`
- `WfsFeatureCatalogGroup`

Anyway, consider your PR approved :) (but because I added commits, I guess you should review those and then merge.)

kring:
Those new names are sensible to me.  We need to avoid breaking backward compatibility when we rename, though.

kring:
Thanks for writing the tests!

kring:
Your changes look good.  I switched the new catalog group to use `proxyCatalogItemUrl` (which didn't exist when I originally wrote this), so back to you for final review and merge. :)

PR #935, opened 2015-09-24
Loading indicator

travissimon:
This commit uses css to animate the checkbox associated with data
catalog items. It provides three states - unchecked, loading and
loading. The checkbox itself is created with SVG.

The loading animation animates the dash-offset property to provide the
motion. The checked property uses an animation to draw the stroke path
and then fade in the fill. It's a less-effective animation at smaller
sizes and may (or may not) require tweaking.

### DIFF
--- a/lib/Styles/DataCatalogTab.less
+++ b/lib/Styles/DataCatalogTab.less
@@ -158,7 +158,13 @@
 	stroke-dasharray: 80;
 	stroke-dashoffset: 0;
 	animation: border-loading 1.25s linear;
+	-moz-animation: border-loading 1.25s linear;
+	-webkit-animation: border-loading 1.25s linear;
+	-o-animation: border-loading 1.25s linear;
 	animation-iteration-count: infinite;
+	-moz-animation-iteration-count: infinite;
+	-webkit-animation-iteration-count: infinite;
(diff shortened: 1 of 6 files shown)


kring:
Thanks @travissimon, this is pretty sweet.  Only problem I see is that it breaks the checkboxes on the Search tab (yes, there's a horrendous amount of duplication between `SearchTab.html` and `DataCatalogTab.html`, sorry).

image

travissimon:
Yeah, I just saw this myself. I believe it's on NowViewing as well. I added another commit for both cases (and IE9 xmlns support)

travissimon:
Ack, not closed?

kring:
Search looks good now, but we now get errors when we try to enable region-mapped datasets.  For example, try anything under `Australian Bureau of Statistics` or go to:
http://localhost:3001/#test
and try the datasets under `Test Data -> Region Mapping`.

The problem is that `CsvVariable` doesn't have an `isLoading` property.

travissimon:
@kring OK, thanks for the help - this all feels better and cleaner to me. Double check for sure, but I've tested the checkboxes in the now viewing, the csv items in the data catalog, the search results, as well as general clicking around.

kring:
Awesome, thanks @travissimon!

PR #944, opened 2015-09-30
Issue790 attribution

chloeleichen:
issue [790](https://github.com/TerriaJS/terriajs/issues/790):

Added ability to display attribution text if it is specified in the info file.

To test, go to http://127.0.0.1:3001/#build/TerriaJS/test/init/arcgis.json, and toggle on and off the arcgis data. For example:
screen shot 2015-09-30 at 3 44 56 pm

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -310,7 +310,6 @@ CatalogItem.defaultMetadata.isLoading = false;
 CatalogItem.defaultMetadata.dataSourceErrorMessage = 'This data item does not have any details available.';
 CatalogItem.defaultMetadata.serviceErrorMessage = 'This service does not have any details available.';
 
-
 freezeObject(CatalogItem.defaultMetadata);
 
 /**
@@ -336,8 +335,6 @@ CatalogItem.defaultUpdaters.attribution = function(catalogItem, json, prototypeN
         }
     }
 };
-
-
 
 freezeObject(CatalogItem.defaultUpdaters);
 
@@ -590,7 +587,6 @@ CatalogItem.prototype._enable = function() {
     }
 };
 
-
 /**
  * Disables this data item on the globe or map.  This method:
  * * Should not be called directly.  Instead, set the {@link CatalogItem#isEnabled} property to false.
@@ -779,8 +775,6 @@ function isEnabledChanged(catalogItem) {
 
         terria.analytics.logEvent('dataSource', 'added', catalogItem.name);
         catalogItem._enabledDate = Date.now();
-
-
     } else {
         catalogItem.isShown = false;
 
@@ -788,13 +782,10 @@ function isEnabledChanged(catalogItem) {
         // actually gone through.
         if (!defined(catalogItem._loadForEnablePromise)) {
             catalogItem._disable();
-
             catalogItem.terria.currentViewer.removeAttribution(catalogItem.attribution);
         }
 
-
         terria.nowViewing.remove(catalogItem);
-
 
         var duration;
         if (catalogItem._enabledDate) {
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -18,6 +18,7 @@ var CatalogMember = require('./CatalogMember');
 var inherit = require('../Core/inherit');
 var raiseErrorOnRejectedPromise = require('./raiseErrorOnRejectedPromise');
 var runLater = require('../Core/runLater');
+var Credit = require('terriajs-cesium/Source/Core/Credit');
 
 /**
  * A data item in a {@link CatalogGroup}.
@@ -30,7 +31,7 @@ var runLater = require('../Core/runLater');
  * @param {Terria} terria The Terria instance.
  */
 var CatalogItem = function(terria) {
-     CatalogMember.call(this, terria);
+    CatalogMember.call(this, terria);
 
     this._enabledDate = undefined;
     this._shownDate = undefined;
@@ -86,6 +87,13 @@ var CatalogItem = function(terria) {
      * @type {String}
      */
     this.dataCustodian = undefined;
+    /**
+     * Gets or sets an attribution text of this data item.
+     * This property is an HTML string that must be sanitized before display to the user.
+     * This property is observable.
+     * @type {String}
+     */
+    this.attribution = undefined;
 
     /**
      * Gets or sets the URL from which this data item's metadata description can be retrieved, or undefined if
@@ -157,7 +165,7 @@ var CatalogItem = function(terria) {
 
     this._legendUrls = [];
 
-    knockout.track(this, ['rectangle', 'legendUrl', 'dataUrlType', 'dataUrl', 'dataCustodian',
+    knockout.track(this, ['rectangle', 'legendUrl', 'dataUrlType', 'dataUrl', 'dataCustodian', 'attribution',
                           'metadataUrl', 'isEnabled', 'isShown', 'isLegendVisible', 'clock',
                           'isLoading', 'isMappable', 'showsInfo', 'customProperties',
 						  'nowViewingMessage', '_legendUrls']);
@@ -302,6 +310,7 @@ CatalogItem.defaultMetadata.isLoading = false;
 CatalogItem.defaultMetadata.dataSourceErrorMessage = 'This data item does not have any details available.';
 CatalogItem.defaultMetadata.serviceErrorMessage = 'This service does not have any details available.';
 
+
 freezeObject(CatalogItem.defaultMetadata);
 
 /**
@@ -317,6 +326,18 @@ CatalogItem.defaultUpdaters.rectangle = function(catalogItem, json, propertyName
         catalogItem.rectangle = Rectangle.MAX_VALUE;
     }
 };
+
+CatalogItem.defaultUpdaters.attribution = function(catalogItem, json, prototypeName) {
+    if(defined(json.attribution)){
+        if((typeof json.attribution === 'object' && json.attribution.text) && json.attribution.link ){
+            catalogItem.attribution = new Credit(json.attribution.text, undefined, json.attribution.link);
+        }else if(typeof json.attribution === 'string'){
+            catalogItem.attribution = new Credit(json.attribution, undefined, undefined);
+        }
+    }
+};
+
+
 
 freezeObject(CatalogItem.defaultUpdaters);
 
@@ -341,6 +362,7 @@ CatalogItem.defaultSerializers.rectangle = function(catalogItem, json, propertyN
 CatalogItem.defaultSerializers.legendUrls = function(catalogItem, json, propertyName) {
     json.legendUrls = catalogItem._legendUrls;
 };
+
 
 freezeObject(CatalogItem.defaultSerializers);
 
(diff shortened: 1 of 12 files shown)


kring:
Thanks @chloeleichen!  This looks good.  A few comments coming your way before I merge this...

lib/Models/CatalogItem.js
```diff
@@ -86,6 +87,13 @@ var CatalogItem = function(terria) {
      * @type {String}
      */
     this.dataCustodian = undefined;
+    /**
+     * Gets or sets an attribution text of this data item.
+     * This property is an HTML string that must be sanitized before display to the user.
+     * This property is observable.
+     * @type {String}
+     */
+    this.textAttribution = undefined;
```
kring: Should this just be `attribution`?  Currently it can be both text and a link.  In the future we might add support for an image, and we wouldn't want to change the name of this top-level property to do so.

lib/Models/CatalogItem.js
```diff
@@ -318,6 +327,20 @@ CatalogItem.defaultUpdaters.rectangle = function(catalogItem, json, propertyName
     }
 };
 
+CatalogItem.defaultUpdaters.textAttribution = function(catalogItem, json, prototypeName) {
+    if(defined(json.textAttribution)){
+        if((typeof json.textAttribution === 'object' && json.textAttribution[0].text) && json.textAttribution[0].link ){
```
kring: Why the array inside `textAttribution`?  Shouldn't it just be `textAttribution.text` (and then update the json file accordingly)?

lib/Models/CatalogItem.js
```diff
@@ -318,6 +327,20 @@ CatalogItem.defaultUpdaters.rectangle = function(catalogItem, json, propertyName
...
+            catalogItem.textAttribution = new Credit(json.textAttribution[0].text, undefined, json.textAttribution[0].link);
+        }else if(typeof json.textAttribution === 'string'){
+            catalogItem.textAttribution = new Credit(json.textAttribution, undefined, undefined);
+        }
+    } else{
+        return false;
```
kring: No need to return anything.  You can get rid of the entire `else` clause.

lib/Models/CatalogItem.js
```diff
@@ -562,15 +584,18 @@ CatalogItem.prototype._enable = function() {
     var terria =  this.terria;
 
     if (defined(terria.cesium)) {
+        console.log("called");
```
kring: I'm guessing you meant to remove this.

lib/Models/CatalogItem.js
```diff
@@ -562,15 +584,18 @@ CatalogItem.prototype._enable = function() {
...
         terria.cesium.stoppedRendering = true;
         this._enableInCesium();
     }
 
     if (defined(terria.leaflet)) {
+        console.log("called");
```
kring: and this

lib/Models/CatalogItem.js
```diff
@@ -820,4 +853,33 @@ function isShownChanged(catalogItem) {
...
 }
 
+/**
+ * Toggle enable and disable attribution text
+ */
+
+function toggleAttributionText(catalogItem){
+    var terria = catalogItem.terria;
+    function add(textAttribution){
+        if (defined(terria.cesium)){
```
kring: No need for the Cesium/Leaflet check here.  Just do `terria.currentViewer.addAttributionText(textAttribution)`.  With that, I think you can eliminate this entire `toggleAttributionText` function and just call the above (and the remove version) directly.

lib/Models/Cesium.js
```diff
@@ -403,6 +404,27 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
...
+/**
+ * Add and remove attribution text.
+ * @param  {textAttribution} the textAttribution object of each Catalog item.
+ */
+
+ Cesium.prototype.addAttributionText = function(textAttribution){
+    if(textAttribution){
+        this.scene.frameState.creditDisplay.addDefaultCredit(textAttribution);
+    }else{
+        return false;
```
kring: I don't think the `else` / `return` statement is necessary here or below.

lib/Models/GlobeOrMap.js
```diff
@@ -91,4 +91,14 @@ GlobeOrMap.prototype._createEntityFromImageryLayerFeature = function(feature) {
     return entity;
 };
 
+//set text attribution
```
kring: Use jsdoc style comments to document methods and properties.  The `DocBlockr` Sublime Text plugin helps you write them.

lib/Models/GlobeOrMap.js
```diff
@@ -91,4 +91,14 @@ GlobeOrMap.prototype._createEntityFromImageryLayerFeature = function(feature) {
...
+
+GlobeOrMap.prototype.addAttributionText = function(textAttribution){
+    return textAttribution;
```
kring: This should throw a `DeveloperError` as a reminder to implement it in the derived class.  See the other methods in this file, such as `captureScreenshot`.  Same below.

lib/Models/Leaflet.js
```diff
@@ -201,6 +201,36 @@ Leaflet.prototype.computePositionOnScreen = function(position, result) {
...
+ */
+
+Leaflet.prototype.addAttributionText = function(textAttribution){
+    if(textAttribution){
+        var a = textAttribution.link?("<a href='" + textAttribution.link +"'>" + textAttribution.text + "</a>") : textAttribution.text;
+        addAttributionControl(this.map);
+        this.map.addControl(this.map.attributionControl);
+        this.map.attributionControl.setPrefix(a);
+    }else{
+        return false;
```
kring: Unnecessary return (unless there's a reason that I'm missing!).

lib/Models/Leaflet.js
```diff
@@ -201,6 +201,36 @@ Leaflet.prototype.computePositionOnScreen = function(position, result) {
...
+    }
+};
+
+Leaflet.prototype.removeAttributionText = function(textAttribution){
+    if(textAttribution){
+        this.map.attributionControl.removeFrom(this.map);
```
kring: The add/remove logic isn't working quite right in Leaflet.  I think you want to update the existing `attributionControl` instead of creating a new one.  If you switch to 2D and then enable a layer with attribution, the text gets added to a separate control above the first one.  If you enable the layer while in 3D mode, and then switch to 2D, the attribution is instead added to the main attribution control (which is what we want, I'd say), but then when you disable the layer all attributions disappear, not just the ones for that layer.

lib/Models/NoViewer.js
```diff
@@ -64,4 +64,17 @@ NoViewer.prototype.computePositionOnScreen = function(position, result) {
     return undefined;
 };
 
+/**
+ * Add Attibution text.
+ * @param  {textAttribution} the attribution text from the data source
+ */
+
+NoViewer.prototype.addAttributionText = function(textAttribution){
+    return undefined;
```
kring: This can just do nothing; no return needed.

test/Models/WebMapServiceCatalogItemSpec.js
```diff
@@ -35,6 +41,18 @@ describe('WebMapServiceCatalogItemViewModel', function() {
         expect(wmsItem).toBeDefined();
     });
 
+    it('add addAttributionText methods called', function() {
```
kring: This test and the one below are both failing for me, because globeormap is never associated with the `Terria` instance, so enabling `wmsItem` doesn't add it to that map (it adds it to the `NoViewer` instance in `terria.currentViewer`).

I think you want to remove the code above that creates a `GlobeOrMap` instance, and change the `spyOn` below to spy on `terria.currentViewer`.

kring:
That's it for now!  Feel free to keep this PR open and keep committing to this branch.

By the way, it would be good if you could fix the problem mentioned by CLAassistant above, if possible.  If it doesn't make sense to add the email address to your github account, then maybe fix it by squashing the commits (I know I told you not to do that before, and generally it's not necessary, but it's a reasonable thing to do here to associate the commits with the right person).

chloeleichen:
@kring Thank you very much. Will let you know when I've fixed everything :D

chloeleichen:
Hi @kring I have fixed the issues above, please review, thank you very much :D

lib/Models/CatalogItem.js
```diff
@@ -86,6 +87,13 @@ var CatalogItem = function(terria) {
      * @type {String}
      */
     this.dataCustodian = undefined;
+    /**
+     * Gets or sets an attribution text of this data item.
+     * This property is an HTML string that must be sanitized before display to the user.
+     * This property is observable.
+     * @type {String}
```
kring: The type of `attribution` is actually `Credit`, not `String`, right?

chloeleichen: yes you are right

lib/Models/Cesium.js
```diff
@@ -403,6 +403,23 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
     return SceneTransforms.wgs84ToWindowCoordinates(this.scene, position, result);
 };
 
+/**
+ * Add and remove attribution text.
+ * @param  {attribution} the attribution object of each Catalog item.
+ */
```
kring, line 409: Two problems with this doc:
- One jsdoc block describes one method.  So you need a separate block for `removeAttribution` below.
- With `@param`, the first identifier in curlies is the type of the parameter, and the next identifier is the name of the parameter.  So jsdoc thinks you have a parameter named `the` of type `attribution`.

You can run `gulp docs` in the TerriaJS directory, then npm start, then navigate to `http://localhost:3002/doc` to look at the generated reference documentation.

lib/Models/Leaflet.js
```diff
@@ -201,6 +201,28 @@ Leaflet.prototype.computePositionOnScreen = function(position, result) {
...
+};
+
+Leaflet.prototype.removeAttribution = function(attribution){
+    this.map.attributionControl.removeAttribution(leafletCredit(attribution));
+};
+
+/**
+ * A convenient function for handling leaflet credit display
+ * @param {attribution object} for leaflet to display as text or link
+ * @return {[String] text }
```
kring: This is generating a jsdoc error.  It should be something like `@return {String} The HTML for the credit.`.

lib/Models/Leaflet.js
```diff
@@ -201,6 +201,28 @@ Leaflet.prototype.computePositionOnScreen = function(position, result) {
...
+    this.map.attributionControl.addAttribution(leafletCredit(attribution));
+};
+
+Leaflet.prototype.removeAttribution = function(attribution){
+    this.map.attributionControl.removeAttribution(leafletCredit(attribution));
+};
+
+/**
+ * A convenient function for handling leaflet credit display
+ * @param {attribution object} for leaflet to display as text or link
```
kring: Maybe `@param {Credit} attObj The credit to turn into HTML.`

kring: Renaming `attObj` to `credit` would be a nice improvement, too.  We try to avoid abbreviations in the public API.

lib/Models/Leaflet.js
```diff
@@ -201,6 +201,28 @@ Leaflet.prototype.computePositionOnScreen = function(position, result) {
...
+ * @return {[String] text }
+ */
+function leafletCredit(attObj){
```
kring, line 225: Functions should have verb-like names.  Maybe `createHtmlFromCredit`?

kring:
Sorry to nitpick the doc.  This looks great other than the doc things!

chloeleichen:
@kring not at all, thank you for showing me how to do the doc, that's amazing :D

kring:
Thanks @chloeleichen!

Issue #673, opened 2015-05-06
3D Buildings


keithamoss:
One of the things we miss from Google Maps Engine is its support for streaming 3D buildings into Google Earth. While support was patchy and tied to Google Earth, we still managed to put it to good use with data supplied by the City of Perth.

![screen shot 2015-05-06 at 15 53 22](https://cloud.githubusercontent.com/assets/1672960/7489484/c371ccd8-f409-11e4-95c5-2757272d3a99.png)

Cesium currently supports [converting Collada files to gITF](https://cesiumjs.org/convertmodel.html) and [loading those into a Cesium scene](https://cesiumjs.org/2014/03/03/Cesium-3D-Models-Tutorial/), but ideally users should be able to access either a single service or a pre-prepared catalogue of buildings grouped together in a single UI element.

AGI have just announced their intention to [support for streaming 3D buildings via CZML, gITF, and GeoJSON](http://cesiumjs.org/2015/04/27/3D-Buildings-in-Cesium/), however the technical details remain unclear.

Consider this a bit of a "nice to have, maybe let's monitor this space for a few months" enhancement request.

nahidakbar:
Nice.

Not sure if it is still around but another thing in this space that had a lot of hype (5-6 years ago) is CityGML. Saw some kool demos of modeled cities and different levels of detail etc back in the day.

Not sure if they ever got around to building standards for streaming their models.

stevage:
Yeah, it's still around, but does seem to have stagnated a bit. Will
McIntosh at City of Greater Geelong has been playing around with trying to
export their city's 3D model in it.

maxious:
There's an example of how to import CityGML into PostGIS and then serve buildings over WFS for Cesium https://github.com/Oslandia/cesium-buildings https://github.com/Oslandia/workshop-3d

(I tried to import KML via GDAL to PostGIS and Geoserver and it didn't work, possibly because there was one "model" https://data.gov.au/dataset/central-geelong-3d-massing-model )

kring:
I played around with that Geelong model in Cesium almost a year ago.  I renamed the .kmz to .zip, extracted the Collada (.dae) file and texture from there, used the Cesium tool to convert it to glTF, and then manually loaded and geolocated it in Cesium.  It worked reasonably well.  I just tried again now and the model is too big for the online glTF conversion tool, so we'd have to do the conversion using the offline tool.

It'd be really cool to get this sort of data into National Map, one way or another.

[...]

PR #949, opened 2015-10-02
WIP Add a CatalogItem for Cesium buildings

vmora:
Hi @kring, this is a draft version for the integration of tiled 3D WFS datasources in terriajs related to #673 

What I'm looking for at the moment are **comments on the names** (for the class and the catalog type "twfs") and **advices to avoid a hardcoded dependency on terriajs-cesium in cesium-building**.

You have a POC [here](http://37.187.164.233/nationalmap/) where you can zoom on Gold Coast city an see some "buildings", don't be bothered by the apparent motion of the buildings when the point of view changes, it's just due to the fact that the buildings are not at the appropriate altitude (they are not the appropriate height either, I just extruded some public dataset for testing).

To have this demo working, you will need to:
- clone [cesium-buildings terriajs branch](https://github.com/Oslandia/cesium-buildings/tree/terriajs)
- `cd  cesium-buildings` and `sudo npm link`
-  `ce terriajs` checkout this branch for terriajs and `npm link cesium-buildings`
- copy the included 00_Building_test.json into your nationalmap /datasource directory
- `cd nationalmap` and `npm link cesium-buildings` (I believe it's necessary, and I assume you have a linked terriajs in there too)
- `gulp` and lastly `mkdir wwwroot/js/; browserify node_modules/cesium-building/lib/createWfsGeometry.js > wwwroot/js/createWfsGeometry.js` (this is a web worker)

I'm working on the build process to have the last steps integrated. 

I also want to base cesium-buildings apps build process on gulp, such that the terriajs branch can be merged in master.

I'm not familiar with leaflet, so I'll need a hand if a leaflet support is required for integration in terriajs master. We can use the same web worker and worker pool for both interfaces, but I have no clue about the leaflet equivalent of Cesium.QuadTreePrimitive and the associated tile provider.

### DIFF
--- a/00_Building_test.json
+++ b/00_Building_test.json
@@ -1,0 +1,37 @@
+{
+  "catalog": [
+    {
+      "items": [
+        {
+          "items": [
+            {
+              "nowViewingMessage": "Look here for the *legend* and to *interactively filter* the data.",
+              "name": "Buildings",
+              "description": "Layer description goes here",
+              "dataCustodian": "oslandia",
+              "rectangle": [
+                "112.93529999999998",
+                "-38.208809999082796",
+                "152.06206",
+                "-11.05104999964455"
+              ],
+              "type": "twfs",
+              "url": "http://ks29236.kimsufi.com/cgi-bin/tinyows",
+              "typeNames": "tows:lyongeom"
+            }
+          ],
+          "isOpen": true,
+          "type": "group",
+          "name": "Cesium buildings demo (Selected Datasets)"
+        }
+      ],
+      "initialMessage": {
+        "key": "AbsDataProviderWarning",
+        "content": "This is an experimental section ",
+        "title": "Data Provider Information"
+      },
+      "type": "group",
+      "name": "Buildings demo"
+    }
+  ]
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -28,19 +28,19 @@ if (!fs.existsSync('wwwroot/build')) {
     fs.mkdirSync('wwwroot/build');
 }
 
-gulp.task('build-specs', ['prepare-cesium'], function() {
+gulp.task('build-specs', ['prepare-cesium', 'prepare-cesium-buildings'], function() {
     return build(specJSName, glob.sync(testGlob), false);
 });
 
 gulp.task('build', ['build-specs']);
 
-gulp.task('release-specs', ['prepare-cesium'], function() {
+gulp.task('release-specs', ['prepare-cesium', 'prepare-cesium-buildings'], function() {
     return build(specJSName, glob.sync(testGlob), true);
 });
 
 gulp.task('release', ['release-specs']);
 
-gulp.task('watch-specs', ['prepare-cesium'], function() {
+gulp.task('watch-specs', ['prepare-cesium', 'prepare-cesium-buildings'], function() {
     return watch(specJSName, glob.sync(testGlob), false);
 });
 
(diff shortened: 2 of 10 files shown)


vmora:
Hi @kring, I'm trying to build the cesium-buildings demos the same way nationalmap is build. the aim is to keep or demo running while providing a proper node module that can be used by terriajs (the second part works, but the first is also important for us). 

I installed cesium sources with `sudo npm link` (i.e. it's not the terriajs-cesium) and 'npm link cesium' in cesium-building, I cpied/modified the gulpfile from nationalmap for our needs. The problem I'm facing it that I can't figure out how the 'define' (AMD in cesium) is eventually defined in nationalmap/terriajs build: I can't find a reference to either requirejs, r.js or amdefine, is it something else that is used ?

kring:
Thanks @vmora!

re: Leaflet, yes, we do need Leaflet support before we can merge this.  It's ok if the buildings just show up as 2d building footprints on the map.  Also, TerriaJS needs to work on Internet Explorer 9, which means Web Workers cannot be required for the 2D view.  This is to merge the functionality into the main TerriaJS repo.  If you're making a companion npm module, and your users don't need 2D support, then you're of course free to skip 2D support if you want.

TerriaJS pulls in Cesium using the browserify deamdify plugin.  Basically, the Cesium AMD modules (which are defined using a call to the `define` function) get converted on-the-fly to CommonJS (node style) modules by deamdify.  If deamdify is not working, you'll see that `define is not defined` error.  Check your `package.json` and make sure you have lines like this:

```
  "browserify": {
    "transform": [
      "deamdify"
    ]
  }
```

vmora:
Thanks for the information. I found out about damdify on my own enventually. 

Another problem I have is that I'd like the catalog item to depend on a clean cesium-building node module, but I would also like to avoid a direct reference to terriajs-cesium such that the module be usable on it's own. I'm wondering how to achieve that (a sed/cesium/terriajs-cesium/ in the build maybe?).

Concerning leaflet support, I'm afraid I won't have the resources to develop that on my own. The IE9 support is especially painful, because the processing of wfs data on the scale we are targeting is pretty heavy and the interface may slow down to a crawl if we do that in the main thread.

If there is a developer who is interested in this feature, knows about leaflet and is willing to help on this part, I'd be glad to collaborate.

kring:
> Another problem I have is that I'd like the catalog item to depend on a clean cesium-building node module, but I would also like to avoid a direct reference to terriajs-cesium such that the module be usable on it's own. I'm wondering how to achieve that (a sed/cesium/terriajs-cesium/ in the build maybe?).

By "usable on its own" do you mean usable in an application that uses the combined/minified Cesium.js?  If so, yeah, this is going to be tricky!  The only way I can think of, off the top of my head, is to require that your module be invoked with a reference to the `Cesium` namespace.  Then, TerriaJS or another app that uses Cesium via individually-required modules will need to create a dummy Cesium namespace by requiring in all the modules that `cesium-building` needs.  It's ugly.

On the other hand, if you intend for it to be used with the Cesium AMD modules (not the built version) then you should be able to require Cesium modules using the normal AMD syntax, and this will work when building inside TerriaJS as well (because of the Browserify deamdify plugin).

> Concerning leaflet support, I'm afraid I won't have the resources to develop that on my own. The IE9 support is especially painful, because the processing of wfs data on the scale we are targeting is pretty heavy and the interface may slow down to a crawl if we do that in the main thread.

What sort of processing are you doing on the WFS data?  Perhaps it would be sensible to do the processing on a server - and cache it - rather than doing it on the client?

vmora:
> By "usable on its own" do you mean usable in an application that uses the combined/minified Cesium.js?

Something like that. Eventually we would like to PR/Cesium to integrate our devs which would make things a lot simpler, but for the moment the Cesium.QuadTreePrimitive is not in the public API, so we may have to maintain this as a "plugin" for some time.

>  What sort of processing are you doing on the WFS data? 

Json parsing and triangulation, not heavy in itself, it's just the number of features that can be big. But thinking about it, the triangulation is probably not needed in leaflet. For leaflet we may have to do proper reprojection on cpu too (in cesium-buildings we use an approximate transform for each tile that is processed on the GPU).

> Perhaps it would be sensible to do the processing on a server - and cache it - rather than doing it on the client?

Yes, some kind of cached vector tiles. We are considering that, but nothing ready for prod AFAIK.

kring:
@vmora I'm going to close this PR for now, but feel free to reopen if there's a need for further discussion or review.

Issue #783, opened 2015-06-25
No download link for locally hosted files?


stevage:
Locally hosted files (at least CZML?) don't seem to generate a download link:

![screenshot 2015-06-25 10 00 54](https://cloud.githubusercontent.com/assets/678344/8344178/2ab5b240-1b21-11e5-93f0-b43446b68211.png)

Adding the URL fragment to the base URL [works fine](http://nationalmap.gov.au/renewables/datasets/dance/DANCE-2012-AD.czml)

hilarycinis:
IxD note: replicate any other download patterns for other file types

stevage:
To reproduce:
1. http://nationalmap.gov.au#test
2. Select CZML files > Simple Test (CZML)
3. Click (i)

chloeleichen:
You mean it should have a link like this? @stevage 
![sc](https://cloud.githubusercontent.com/assets/7508939/10298074/e070b490-6c25-11e5-9b9a-0cd2ef78c030.png)

chloeleichen:
I see, you mean this one:
![dl](https://cloud.githubusercontent.com/assets/7508939/10298336/b92ee132-6c29-11e5-964e-bf8246b23c0a.png)

chloeleichen:
This happens to not only locally hosted file, for example:
![screen shot 2015-10-06 at 4 08 10 pm](https://cloud.githubusercontent.com/assets/7508939/10300610/90abf6f8-6c44-11e5-9767-aee8c879bcfd.png)
I think this is because for some data type(e.g: CZML) , the dataUrl and dataUrlType has not been given a fallback when they don't have any value, so I set them to the url of the catalogue url if they dont have a dataUrl defined, and dataUrlType is set to 'direct':#954

stevage:
Looks good. Doesn't work for GeoJSON yet? (eg, the GeoJSON files in the #test init file)

chloeleichen:
it should work with geojson, let me check again

stevage:
Ah, actually there's only one that doesn't work, the zipped airports.

chloeleichen:
yeah I don't know why that dataset doesn't work, I can't select that dataset on my local at all. When i try to select it I get:
`An error occurred while attempting to convert this file to GeoJson. This may indicate that the file is invalid or that it is not supported by the AREMI conversion service. If you would like assistance or further information, please email us at aremi@nicta.com.au.`

stevage:
[...]

PR #954, opened 2015-10-06
Issue#783 - download links for all data types

chloeleichen:
issue #783 : add download links to all data type

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -40,12 +40,6 @@ var CatalogItem = function(terria) {
     this._lastLoadInfluencingValues = undefined;
 
     /**
-     * The url of the data
-     * @type {String}
-     */
-    this.url = '';
-
-    /**
      * The index of the item in the Now Viewing list.  Setting this property does not automatically change the order.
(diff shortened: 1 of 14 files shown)


lib/Models/CatalogItem.js
```diff
@@ -373,6 +389,23 @@ CatalogItem.defaultSerializers.attribution = function(catalogItem, json, propert
     }
 };
 
+CatalogItem.defaultSerializers.url = function(catalogItem, json, prototypeName) {
```
kring, line 392: These three serializers shouldn't be necessary, because the default serialization will do the same thing.

However, it'd be useful to avoid serializing `dataUrl` when it is identical to `url`.

lib/Models/CatalogMember.js
```diff
@@ -34,6 +34,12 @@ var CatalogMember = function(terria) {
     this.name = 'Unnamed Item';
 
     /**
+     * Gets or sets the url of the item.  This property is observable.
+     * @type {String}
+     */
+    this.url = '';
```
kring, line 40: Is this needed?

Issue #496, opened 2015-01-29
Add ability to filter which WFS/WMS data to display in popup - templates for the Info boxes


meh9:
Many of the WFS layers have large numbers of key-value pairs that are not relevant to anybody but the custodian of the data, it would be great if we could filter which keys to display in the data popup.

For instance, turning on the GA Island layer and clicking on an island shows the following data:

```
OBJECTID    2970
FEATTYPE    Island
NAME    HINCHINBROOK ISLAND
STATE   QLD
FEATREL 2001-11-28T13:00:00Z
ATTRREL 2001-11-28T13:00:00Z
PLANACC 9999
SOURCE  GEOSCIENCE AUSTRALIA
CREATED 2006-05-08T14:00:00Z
RETIRED 
PID 3491633
SYMBOL  0
SHAPE_Leng  1.32347920603
SHAPE_Area  0.0338646391
```

In this case probably the only keys we are really interested in displaying to an end user might be FEATTYPE, NAME, and STATE. If we do not display all the keys maybe we should have a [...] symbol that could display them all when clicked.

We could control it something like this in the init_nm.json:

```
{
    "name": "Islands",
    ...
    "url": "http://geoserver-nm.nicta.com.au/geotopo_250k/ows",
    "layers": "Framework:islands",
    "displayKeys": "FEATTYPE,NAME,STATE"
}
```

kring:
CC #348 

Even nicer than the list of display keys: we could provide an HTML template to display in the feature popup with something like {FEATTYPE} for feature fields.

meh9:
This is actually just as relevant for WMS services as well.

meh9:
This is becoming a higher and higher priority for AREMI.

@kring having a template for the Info box is definitely the way we would like to take this, it opens up so many possibilities.

chloeleichen:
I can give this a go, do we have a rough wireframe for how it should work? Do we want to allow users to filter data via the UI? Something like the following? 

![info](https://cloud.githubusercontent.com/assets/7508939/10153064/e1de86a2-669b-11e5-89e0-c4728080f10c.png)

kring:
We don't need the user to be able to select the fields (at least not currently!).  I think what AREMI wants (and it's a great TerriaJS feature in general!) is to be able to specify an HTML template in the catalog item definition in the init file.  For example, something like:

```
[...]

PR #955, opened 2015-10-06
Issue496 use info template to display popup info

chloeleichen:
DON'T MERGE YET

Issue #496  This changes will be ready when https://github.com/AnalyticalGraphicsInc/cesium/pull/3054 is merged. 
Also I'm not quite sure about my tests, don't seem strong enough. Anyone can help?


lib/Core/readTemplate.js
```diff
@@ -0,0 +1,45 @@
...
+        };
+    }
+    //if supplied a string
+    function parse(template) {
+        var keys,
+            output ='',
+            currentKey,
+            currentValue,
+            i = 0;
+        keys = (keyFilter(template))(rg);
```
kring: I'm confused about the purpose of this function call.  Isn't this line equivalent to `keys = template.match(rg)`?

chloeleichen: You are right:D, thanks @kring

kring:
Thanks @chloeleichen, this is great!  I just updated this branch to use a new version of terriajs-cesium that includes your Cesium changes.

One question for you: did you consider (and reject?) using something like [mustache](https://github.com/janl/mustache.js) instead of your own templating system?

test/Core/readTemplateSpec.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+var readTemplate = require('../../lib/Core/readTemplate');
+var mockData, mockTemplate, output;
+
+describe("readTemplate", function() {
+  it("contains spec with an expectation", function() {
```
kring: Better names for this test and the one below?

chloeleichen:
@kring:

> One question for you: did you consider (and reject?) using something like mustache instead of your own templating system?

 Yes I thought about it, but I think what we want here is quite simple, while templating frameworks have pretty complicated functions that we won't need in this case. 

But I do think that using one of those library would be a more robust solution. So maybe we should use mustache? Would you like me to give it a try?

kring:
I think it could be worthwhile, to handle repeated templates over lists, elements that are treated as text versus HTML, etc.  If you (want to / have time to) try it out, that'd be awesome, but let's make it a separate pull request.

lib/Models/CatalogItem.js
```diff
@@ -163,6 +163,12 @@ var CatalogItem = function(terria) {
      */
     this.nowViewingMessage = undefined;
 
+    /**
+     * Gets or sets a template to display message in a info box.
+     * @type {String} Html template that can be used to display info
+     */
+    this.infoTemplate = undefined;
```
kring: Mind renaming this to `featureInfoTemplate`?  I should have realized it sooner, but `infoTemplate` is a bit ambiguous considering that `CatalogItem` has an `info` property that is completely separate from this.

chloeleichen: sure, what about featureInfoTemplate?

lib/ViewModels/FeatureInfoPanelViewModel.js
```diff
@@ -139,21 +139,25 @@ FeatureInfoPanelViewModel.prototype.showFeatures = function(features) {
...
         }
-
-         that.terria.selectedFeature = feature;
+        that.terria.selectedFeature = feature;
         that.name = feature.name ? feature.name : feature.id;
-        if (defined(feature.description)) {
-            that.html = feature.description.getValue( that.terria.clock.currentTime);
+
+        if (defined(feature.properties)) {
+            var data = feature.properties.getValue(feature),
```
kring: Should probably pass `that.terria.clock.currentTime` instead of `feature`, right?

lib/Models/CatalogItem.js
kring: Sounds good!

lib/ViewModels/FeatureInfoPanelViewModel.js
```diff
@@ -139,21 +139,25 @@ FeatureInfoPanelViewModel.prototype.showFeatures = function(features) {
...
+                template = feature.imageryLayer.infoTemplate;
+
+            that.html = template ? readTemplate(data, template) : feature.description.getValue();
```
kring: Need to pass `that.terria.clock.currentTime` to `getValue` here, too, just in case the description is time-varying.

lib/Models/CatalogItem.js
chloeleichen: done :smile: @kring

Issue #507, opened 2015-02-02
Data navigation - Should be able to use TAB key to navigate data/now viewing selections


Layoric:
A user wanting to turn data layers on and off should be able to do so using keyboard navigation. At the moment, no input/label or tabindexs are used for the left hand menu (except the search box).

![image](https://cloud.githubusercontent.com/assets/234642/5995667/795b2c4e-aaed-11e4-8733-68a5f08c68d8.png)

kring:
Thanks @Layoric, we will definitely do this as well sometime soon.  As you know, support for keyboard support in National Map is not great at the moment.

hilarycinis:
Agreed, thanks as this would be part of accessibility options too.

Layoric:
Just as an update to this, a user at GA discovered the following behaviour when trying to use tab to navigate the current version of NationalMap.

![image](https://cloud.githubusercontent.com/assets/234642/7804758/8a9b037e-03aa-11e5-95f7-177c52b1bb9c.png)

Screenshot above is behaviour in Chrome 42, but behaviour slightly differs depending on browser. Problem occurs when search field grabs focus and search tab having not been clicked.

kring:
Thanks @Layoric.  Keith submitted a fix for the immediate problem above in #717.  Proper keyboard navigation will have to wait a bit longer.

chloeleichen:
This issue is related to #959

Issue #506, opened 2015-02-02
Search Results – Should be able to use TAB key to toggle between the results


Layoric:
When the user enters a search, the application displays a list. In the screen shot example below the user entered “Maranboy” as the search string.  User should be able to use the TAB key to move between the results, or even use the up and down arrow.

![issue-search-results](https://cloud.githubusercontent.com/assets/234642/5995621/701c4aba-aaec-11e4-87f3-d61979715e93.png)

kring:
Thanks @Layoric, we definitely plan to do this.

kring:
This was done in #966.

PR #966, opened 2015-10-12
Keyboard nav issue507

chloeleichen:
Fixes issue #507 and #506

### DIFF
--- a/lib/Styles/FeatureInfoPanel.less
+++ b/lib/Styles/FeatureInfoPanel.less
@@ -31,12 +31,32 @@
 
 .feature-info-panel {
     .panel;
+    width: auto;
     background-color: @panel-background-color;
     top: 60px;
     right: 95px;
     max-width: 600px; // adjusted dynamically in FeatureInfoPanelViewModel.js
     position: absolute;
     box-shadow: @panel-box-shadow;
+}
+
+.feature-info-panel-section-content-wrapper{
+    max-height: 0;
+    max-width: 0;
+    overflow: hidden;
+    table{
+        width: 100%;
+    }
+    &.closed{
+        max-height: 0;
+        max-width: 100%;
+        overflow: hidden;
+    }
+    &.open{
+        width: auto;
+        max-height: 100%;
+        max-width: 100%;
+    }
 }
 
 .feature-info-panel-header {
@@ -51,6 +71,7 @@
 .feature-info-panel-body {
     .opaque-to-input;
 }
+//data-bind="style: { 'max-width': maxWidth + 'px' }
 
 .feature-info-panel-sections {
     overflow-y: auto;
@@ -82,8 +103,8 @@
 .feature-info-panel-section-content {
     color: @info-text-color;
     font-weight: lighter;
+    overflow: visible;
     padding: 15px;
-    overflow: auto;
     // max-height: 600px; // adjusted dynamically in FeatureInfoPanelViewModel.js
     text-align: left;
 }
--- a/lib/ViewModels/FeatureInfoPanelSectionViewModel.js
+++ b/lib/ViewModels/FeatureInfoPanelSectionViewModel.js
@@ -74,6 +74,7 @@ FeatureInfoPanelSectionViewModel.prototype.toggleOpen = function() {
     } else {
         this.terria.selectedFeature = this.feature;
     }
+
     // ensure the targeting cursor keeps updating (as it is hooked into the Cesium render loop)
     this.terria.currentViewer.notifyRepaintRequired();
 };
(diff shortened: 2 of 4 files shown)


kring:
This change makes the labels in the catalogue not fit as nicely as they did before.  For example, previously it looked like this:
image

Now it looks like this:
image

Note that the `Department of Regional Australia, Local Government...` label is wrapped too early, and that the checkbox next to `Regional Development Australia (RDA) 2014-15` is aligned with the second line instead of the first.

I think these changes are a result of removing `display: table-cell` from the label DOM element.  I know tables are not popular in this situation, but I couldn't figure out another way to get the desired layout.  A fixed width won't work because the width changes based on a bunch of factors, including the depth of the item in the tree and the number of extra buttons (like the (i) button) are active for the item.  NationalMap doesn't use any of the latter, but other apps do.

kring:
Two small (unintentional?) changes to the Navigation controls:
- They're now displayed with a brighter white color.  Probably fine.
- The mouse cursor no longer changes to a hand when over the navigation buttons.

kring:
More places where the cursor used to become a hand, but doesn't anymore:
- The tabs on the explorer panel.
- Top-level group labels (e.g. `National Data Sets`).

And on the Add Data panel, the Add button lost some styling.  Before:
image

After:
image

kring:
Sorry for the nitpicking; it's awesome to have real keyboard controls, so thanks for doing this!

lib/Styles/Core.less
```diff
@@ -167,3 +167,15 @@ select {
   font-size: 14px;
   padding-left: 5px;
 }
+
+
+.sr-only{
```
kring: can you come up with a more descriptive name for this?

kring:
It doesn't seem to be possible to navigate to the (i) button on the Data Catalogue tab, or to the Zoom To / Info / Remove buttons on the Now Viewing tab.

kring:
I can still interact with the underlying UI with the keyboard even while a modal dialog is up.

lib/Views/PopupMessage.html
```diff
@@ -1,7 +1,7 @@
 <div class="modal-background" data-bind="click: closeIfClickOnBackground">
     <div class="popup-message" data-bind="style: { width: width + 'px', height: height + 'px', 'max-width': maxWidth + 'px', 'max-height': maxHeight + 'px' }">
         <div class="modal-header">
-            <div class="modal-close-button" data-bind="click: close">&times;</div>
+            <div class="modal-close-button" data-bind="click: close, hasFocus: true">&times;</div>
```
kring: I don't think this is working.  I can't seem to control a modal dialog with keyboard controls.  To test this, open the Chrome debugger, Resources tab, Local Storage, localhost, and delete the `NationalMap.AbsDataProviderWarning` key.  Then open the `Australian Bureau of Statistics (BETA)` folder.

lib/Styles/Core.less
chloeleichen: hi @kring sr-only means screen reader only, i think it's a pretty standard name? it's used by bootstrap as well :)

lib/Views/PopupMessage.html
chloeleichen: Thanks Kevin!
1. My current approach is to not have keyboard access to modal window(so it's not possible to navigate to the (i) button on the Data Catalogue tab) , and if a modal window opens up because of an error, the close button is auto focused so user can close it(this seems to be broken will fix this). But I guess we should make the modal window Accessible by keyboard as well. i will do more work on this. 
2. On national map, i think when page load, the data catalog tap get activated immediately, and that causes the hidden input box below it get focused, that's why you can't go to info tab. I think if you try it on Aremi it should work. Or if you manually focus on the now viewing tab then you should be able to tap to remove. I will double check.

Thanks for the Comments Kevin. I think I need to do more testing on this, especially with all the modal windows.

lib/Styles/Core.less
kring: Oh ok, I hadn't run into that before.  I guess technically you're not using it here for content intended for screen readers, but more as a general "don't show this thing".  But it's fine in any case.

kring:
> But I guess we should make the modal window Accessible by keyboard as well.

Yeah, ideally we'd make the app completely usable with the keyboard, including map navigation!  But that can happen incrementally; we don't have to get all the way there in a single pull request.  So if modals are keyboard-inaccessible (except for closing them) for now, I'm ok with that!

chloeleichen:
Deleted all changes and made this branch the same with master, as i think the way i did it was not good enough, it's kinda semi fixing things. 

Giving another try

PR #977, opened 2015-10-19
allow composite catalog items (issue 956)

RacingTadpole:
I've upgraded CompositeCatalogItem so that:
- it can be added via json
- it functions as a single item in the UI
- its legends can be reordered
- it shows its own group info in the info box (we may want to enhance this one day to show its constituent items' info)

I also refactored the layer `raise` and `lower` code out of NowViewing into the Cesium and Leaflet models, and the `lowerToBottom` out of the various CatalogItems into the Cesium and Leaflet models too.


lib/Models/CatalogGroup.js
```diff
@@ -52,6 +52,15 @@ var CatalogGroup = function(terria) {
     this.isLoading = false;
 
     /**
+     * Gets or sets a value that tells the UI whether this is a group. 
+     * Groups, when clicked, expand to show their constituent items.
+     * This property is observable.
+     * @type {Boolean}
+     * @default false
+     */
+    this.isGroup = true;
```
kring: It would be better to define this property with the `defineProperties` section below, and make it gettable-only.  Defining it here means it is settable, and it will even be deserialized from JSON.

lib/Models/Cesium.js
```diff
@@ -423,6 +423,138 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
     }
 };
 
+/**
+ * Updates the order of layers on the Leaflet map to match the order in the Now Viewing pane.
+ */
+Cesium.prototype.updateLayerOrder = function() {
```
kring: The name of this is a bit misleading in the context of Cesium, right?  Because it doesn't update layer order in general, it only moves `keepOnTop` layers to the top.

kring: Also the comment above mentions Leaflet.

lib/Models/Cesium.js
```diff
@@ -423,6 +423,138 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
...
+
+Cesium.prototype.updateLayerOrderAfterReorder = function() {
+    // because this Cesium model does the reordering via raise and lower, no action needed.
+};
+
+// useful for counting the number of items in composite and non-composite items
+function numSubItems(item) {
+    var N = 1;
+    if (defined(item.numItems)) {
+        N = item.numItems();
```
kring: I think I would just use `item.items.length` here, instead of this `numItems` function.  Because:
- This code already needs to know about sub-items and how to manage them, so there's little reason it shouldn't also know how to count them.
- `numItems` is a bit general for the public API.  Based on the name, it sounds like it should also be implemented on `CatalogGroup`.

lib/Models/Cesium.js
```diff
@@ -423,6 +423,138 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
...
+    }
+};
+
+Cesium.prototype.updateLayerOrderAfterReorder = function() {
+    // because this Cesium model does the reordering via raise and lower, no action needed.
+};
+
+// useful for counting the number of items in composite and non-composite items
+function numSubItems(item) {
+    var N = 1;
```
kring: Total nitpick: use `camelCase` for local variables and function names.  Only use `PascalCase` for class names.

lib/Models/Cesium.js
```diff
@@ -423,6 +423,138 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
...
+        }
+    }
+};
+
+Cesium.prototype.updateLayerOrderAfterReorder = function() {
+    // because this Cesium model does the reordering via raise and lower, no action needed.
+};
+
+// useful for counting the number of items in composite and non-composite items
+function numSubItems(item) {
```
kring: Another total nitpick: avoid abbreviations, and name functions with verbs.  `countNumberOfSubItems` maybe.

lib/Models/Cesium.js
```diff
@@ -423,6 +423,138 @@ Cesium.prototype.computePositionOnScreen = function(position, result) {
...
+
+/**
+ * Lowers this imagery layer to the bottom, underneath all other layers.  If this item is not enabled or not shown,
+ * this method does nothing.
+ * @param {CatalogItem} item The item to lower to the bottom (usually a basemap)
+ */
+Cesium.prototype.lowerToBottom = function(item) {
+
+    if (defined(item.items)) {
+        // TODO: think about whether this loop is better fwd or backward
```
kring: Try to avoid leaving TODOs in the code.  If it can't be addressed now but you don't want to forget about it, write a GitHub issue instead.

lib/Models/CompositeCatalogItem.js
```diff
@@ -25,6 +35,20 @@ var CompositeCatalogItem = function(terria, items) {
 
 	this.items = defined(items) ? items.slice() : [];
 
+    overrideProperty(this, 'legendUrls', {
```
kring, line 40: In order for Knockout dependency tracking to work properly, `knockout.defineProperty` (which `overrideProperty` calls) must come _after_ `knockout.track`.   See the "Computed Properties" section here:
http://blog.stevensanderson.com/2013/05/20/knockout-es5-a-plugin-to-simplify-your-syntax/

kring:
All I have for now, looks good!

kring:
@RacingTadpole I've just assigned this issue to you.  Assign it back to me when it's ready for another look.  (@stevage's excellent idea)

lib/Models/CatalogMember.js
```diff
@@ -128,6 +128,18 @@ defineProperties(CatalogMember.prototype, {
     },
 
     /**
+     * Gets a value that tells the UI whether this is a group. 
+     * Groups, when clicked, expand to show their constituent items.
+     * @memberOf CatalogGroup.prototype
```
kring: Should be `CatalogMember.prototype`.

kring:
Just that one minor thing, so I fixed it myself.  Merging!

PR #994, opened 2015-10-23
569 multi feature

### LINKED ISSUE #569: Get clicked point feature info for all enabled layers

RacingTadpole:
Fixes #569 - shows point feature info for all enabled layers at the clicked position.


kring:
Awesome!  A couple of thoughts before I look at the code more closely:
- When expanding/collapsing sections, we should call `terria.currentViewer.notifyRepaintRequired`.  Otherwise, the targeting cursor stops updating once the Cesium render loop stops.
- Can we show the name of the catalog item that each feature came from....somewhere?

RacingTadpole:
Thanks. I've added the call to `notifyRepaintRequired`. Is the catalog item name available from the feature object? I couldn't find it. If so, I can prepend it to the feature name. Tho it might look ugly if the names get long, or if there's only one catalog item selected.

kring:
> Is the catalog item name available from the feature object?

Not directly.  Chloe just added a reference back to the `ImageryLayer` for raster features.  From that it's possible (but not especially easy) to find the associated catalog item.  We should extend our (and maybe Cesium's as necessary!) picking system so that picked features always know their `CatalogItem`.

> I can prepend it to the feature name. Tho it might look ugly if the names get long, or if there's only one catalog item selected.

There are different things we can do.... prepend it, append it, group by it, and probably other things I haven't thought of.  Let's just do something that isn't too terrible and then get UX input on it.

kring:
Was the call to `notifyRepaintRequired` lost somewhere along the way?  I don't see it, and the targeting cursor stops animating after the map is idle for a couple of seconds.

lib/Models/Cesium.js
```diff
@@ -639,6 +644,7 @@ function pickObject(cesium, e) {
         result.error = 'An unknown error occurred while picking features.';
     });
 
+    result.features = result.features.filter(uniqueElements);
```
kring: This happens too soon, right?  The list of features won't actually be populated until the promise above resolves.  Also, I'm curious why it's needed at all.  Are duplicates being added somewhere?

RacingTadpole: Good point, thanks. Yes, I found that the AREMI generation/available capacity data was giving duplicate features.

RacingTadpole: Improved version now pushed - thanks

RacingTadpole:
FYI, I've submitted a pull request for the changes to Cesium - https://github.com/AnalyticalGraphicsInc/cesium/pull/3130

RacingTadpole:
I've just noticed the scroll bar lets you scroll the title and close button off the screen - will fix this now.

RacingTadpole:
Hi @kring, I've fixed the remaining irritations I had with this now... :-)

lib/Models/CzmlCatalogItem.js
```diff
@@ -29,7 +29,7 @@ var readJson = require('../Core/readJson');
 var CzmlCatalogItem =  function(terria, url) {
      CatalogItem.call(this, terria);
 
-    this._czmlDataSource = undefined;
+    this.dataSource = undefined;
```
kring: This property should be named with an underscore, and then exposed as a public, non-underscored property with a getter only.  See `ImageryLayerCatalogItem.imageryLayer`.  Otherwise, `CatalogMember.serializeToJson` will try to write the data source to the JSON representation of the catalog item in some situations, which we don't want.

lib/Models/GlobeOrMap.js
```diff
@@ -91,6 +91,16 @@ GlobeOrMap.prototype._createEntityFromImageryLayerFeature = function(feature) {
         }
     };
 
+    // TODO: replace the above with this:
+    //
+    // entity.description = feature.description;  // already defined by the new Entity
+
+    // if (entity.propertyNames.indexOf('properties') === -1) {  // not defined yet, but could be in future
+    //     entity.addProperty('properties');
```
kring: I think we should be doing this `entity.addProperty` instead of the manual `entity.properties = {` above, and then change the code elsewhere to assume that entity.properties is never time-varying.  I think we talked about this before, but I can't remember: was there a reason that wouldn't work?

lib/Models/CzmlCatalogItem.js
RacingTadpole: good point, done!

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -0,0 +1,133 @@
+'use strict';
+
+/*global require*/
+var defined = require('terriajs-cesium/Source/Core/defined');
+var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
+var destroyObject = require('terriajs-cesium/Source/Core/destroyObject');
+var defaultValue = require('terriajs-cesium/Source/Core/defaultValue');
+// var Entity = require('terriajs-cesium/Source/DataSources/Entity');
```
kring: remove this?

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -0,0 +1,133 @@
...
+        var dataSource = feature.entityCollection.dataSource;
+        for (i = nowViewing.items.length - 1; i >= 0; i--) {
+            if (nowViewing.items[i].dataSource === dataSource) {
+                result = nowViewing.items[i];
+                break;
+            }
+        }
+        return result;
+    }
+    // if there is no data source, for ArcGis (at least) we can match up the imagery layer on the feature with a now-viewing item
```
kring: Yep this should work for all raster catalog items.

lib/ViewModels/FeatureInfoPanelViewModel.js
```diff
@@ -90,11 +112,39 @@ FeatureInfoPanelViewModel.prototype.closeIfClickOnBackground = function(viewMode
     return true;
 };
 
+FeatureInfoPanelViewModel.prototype.close = function() {
+    this.isVisible = false;
+    // this.resetSections();  // would briefly flash an empty feature info panel as it closes
```
kring: Right, because the panel fades out when isVisible is set to false, rather than disappearing instantaneously.  If resetting the sections is valuable, you could arrange to do it later (though it's a bit ugly).  If it's not needed, might as well just remove this commented-out code.

kring:
Just those few things, this is looking great!

RacingTadpole:
Back to you!

kring:
I can get double scrollbars with lots of layers enabled:

image

RacingTadpole:
Weird.  The double scroll bars were due to some css that I'm sure I changed already.  Must have gotten lost at the same time as that notify repaint call.  Anyway, fixed now.

PR #997, opened 2015-10-26
Added Proj4 projections to the location bar

keattang:
- Clicking on the location bar switches between lats/longs and projected coordinates.
- The initial state can be set through the `useProjection` property.


kring:
Thanks @keattang.  I'm a little hesitant to merge this, though, mainly because it seems hard-coded to only work with UTM as the projected coordinates.  Do you think it can be generalized?  Or, alternatively, you could consider keeping a modified copy of this code in your own application and used that instead of the one built into TerriaJS (that approach lets you avoid maintaining a fork of TerriaJS).

keattang:
I'll take a look at generalising it. Do you have an example of another projected coordinate system?

kring:
Sure, there are thousands of them, if not more.  The US loves their state planes:
https://en.wikipedia.org/wiki/State_Plane_Coordinate_System

Another one we use a lot on the web is Web Mercator (although probably not too many people would want to display those coordinates to their users):
http://spatialreference.org/ref/sr-org/7483/

Many countries use projections that are convenient within their borders.  Here's a Swiss one:
http://www.swisstopo.admin.ch/internet/swisstopo/en/home/topics/survey/sys/refsys/swiss_grid.html

lib/ViewModels/LocationBarViewModel.js
```diff
@@ -26,12 +27,18 @@ var LocationBarViewModel = function(options) {
...
+    this.useProjection = defaultValue(options.useProjection, false);
 
     this.latitude = '43.199°S';
     this.longitude = '154.461°E';
     this.elevation = '28m';
+    this.utmZone = '56';
+    this.northing = '5216050.13';
+    this.easting = '618699.50';
+    this.proj4longlat = '+proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees +no_defs';
+    this.proj4utm = '+proj=utm +ellps=GRS80 +units=m +no_defs';
```
keattang: Is the projection not defined by this line here? Could this then be changed after creating the `LocationBarViewModel` or set as an option to change the projection and displaying the UTM zone could be made optional?

kring: Yes, it should be possible to define any projection with a PROJ4 line like that.  I think the problem is mostly terminology...  Easting/Northing/Zone, as well as `proj4utm`.

keattang: Cool I'll make some tweaks and get back to you

keattang:
How does that look now?

lib/ViewModels/LocationBarViewModel.js
```diff
@@ -117,11 +135,27 @@ function updateCoordinatesFromLeaflet(viewModel, mouseMoveEvent) {
 }
 
 function cartographicToFields(viewModel, coordinates, errorBar) {
-    viewModel.latitude = Math.abs(CesiumMath.toDegrees(coordinates.latitude)).toFixed(3) + '°' + (coordinates.latitude < 0.0 ? 'S' : 'N');
-    viewModel.longitude = Math.abs(CesiumMath.toDegrees(coordinates.longitude)).toFixed(3) + '°' + (coordinates.longitude < 0.0 ? 'W' : 'E');
+
+    var zone = 1 + Math.floor((coordinates.longitude+180)/6);
```
kring: I don't think this is working correctly, it always come up with 30 or 31.  Probably because `coordinates.longitude` is radians and this line (and others below) are assuming degrees.

kring: Also, if you profile while moving the mouse around quickly, a not insignificant amount of time is spent in `proj4`.  Can we skip the projection when only showing longitude and latitude?

CHANGES.md
```diff
@@ -5,6 +5,7 @@ Change Log
 ### 1.0.46
 
 * Fixed an incorrect require (`URIjs` instead of `urijs`).
+* Added Proj4 projections to the location bar. Clicking on the bar switches between lats/longs and projected coordinates. To enable this, set `useProjection` to `true`
```
kring: Can you move this up to a new 1.0.47 section while you're making the other changes?

keattang:
This should be good to go now :)

PR #1002, opened 2015-10-27
Show lat/lon in 2D, don't show elev in 3D smooth.

### LINKED ISSUE #995: Turn off elevation display when using 3D Smooth mode
We continue to show the elevation display next to the lat/long display at the bottom right when we are in the 3D Smooth mode, but it just says `+-107m`. We should not show it when we don't have a sensible elevation to show.

kring:
image

Fixes #995

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,8 @@ Change Log
 * Handle WMS time interval specifications (time/time and time/time/periodicity)
 * Moved `url` property to base CatalogItem base class.  Previously it was defined separately on most derived catalog items.
 * Most catalog items now automatically expose a `dataUrl` that is the same as their `url`.
+* The `LocationBarViewModel` now shows the latitude and longitude coordinates of the mouse cursor in 2D as well as 3D.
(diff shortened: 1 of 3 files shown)


kring:
@RacingTadpole do you want to review and merge this one?

RacingTadpole:
This looks like a great improvement! Just two thoughts -
- Would it be worth renaming "e" to "event", following the general principle of not abbreviating things?
- Is there a Spec we can add?

kring:
> Would it be worth renaming "e" to "event", following the general principle of not abbreviating things?

I've always avoided `event` because I thought it was a reserved word.  It's highlighted a different color by Sublime Text, for example.  But I just looked it up and that was apparently only true in ES3.  In ES5 it's fair game.  So ok, done. :)

> Is there a Spec we can add?

I'll think about it.  It's tough because it requires interaction with both the DOM (`mousemove` event) and Leaflet (to get latitude and longitude from a mouse location).

kring:
This is ready for another look.  The test isn't great, but it's ok.

RacingTadpole:
Nice!

PR #1012, opened 2015-10-29
Infobox transition

chloeleichen:
@RacingTadpole @kring  is this what you mean?

### DIFF
--- a/lib/Styles/FeatureInfoPanel.less
+++ b/lib/Styles/FeatureInfoPanel.less
@@ -31,12 +31,38 @@
 
 .feature-info-panel {
     .panel;
+    width: auto;
+    -webkit-transition: all @explorer-panel-close-animation-length ease-out, visibility 0s @explorer-panel-close-animation-length;
+    -moz-transition: all @explorer-panel-close-animation-length ease-out, visibility 0s @explorer-panel-close-animation-length;
+    -ms-transition: all @explorer-panel-close-animation-length ease-out, visibility 0s @explorer-panel-close-animation-length;
+    -o-transition: all @explorer-panel-close-animation-length ease-out, visibility 0s @explorer-panel-close-animation-length;
+    transition: all @explorer-panel-close-animation-length ease-out, visibility 0s @explorer-panel-close-animation-length;
+
     background-color: @panel-background-color;
     top: 60px;
     right: 95px;
     max-width: 600px; // adjusted dynamically in FeatureInfoPanelViewModel.js
     position: absolute;
     box-shadow: @panel-box-shadow;
+}
+
+.feature-info-panel-section-content-wrapper{
+    width: 0;
+    height: 0;
+    -webkit-transition: all @explorer-panel-close-animation-length linear;
+    -moz-transition: all @explorer-panel-close-animation-length linear;
+    -ms-transition: all @explorer-panel-close-animation-length linear;
+    -o-transition: all @explorer-panel-close-animation-length linear;
+    transition: all @explorer-panel-close-animation-length linear;
+
+}
+
+.feature-info-panel-section-open{
+    .feature-info-panel-section-content-wrapper{
+        width: 1000px;
+        max-width: 100%;
+        height: 485px;
+    }
 }
 
 .feature-info-panel-header {
@@ -48,9 +74,11 @@
     font-weight: 700;
 }
 
+
 .feature-info-panel-body {
     .opaque-to-input;
 }
+//data-bind="style: { 'max-width': maxWidth + 'px' }
 
 .feature-info-panel-sections {
     overflow-y: auto;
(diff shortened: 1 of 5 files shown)


RacingTadpole:
Thanks! That's close, but I can't always tell what it's doing.
When I close a section, it sometimes animates down and sometimes doesn't (eg. see attached)
image
Otherwise when open it always stays at its max-width.
Can we make it animate to the browser-determined width for the content displayed?  Or, can we calculate the max width required for any of the sections, and keep it at that width always?

chloeleichen:
hm ok, if you just want a consistent width you can just make it width: 100% and it will be the max-width of the container which you already have. if you want transition i'll try to find out what's the problem with the width transition. @RacingTadpole

chloeleichen:
hi @RacingTadpole  I'm having trouble replicating the issue you are talking about, isn't it like this?

screen

In which browsers are you having the issue? Do you see any issue with the logic in my changes?

chloeleichen:
@RacingTadpole, the transition is fixed :D, will also check with @hilarycinis on Monday about the max-width

RacingTadpole:
That's looking good to me - thanks! Let me know what Hilary says too.

RacingTadpole:
As a slightly different enhancement, I notice that Cesium slides its info box in from the right hand side when it first appears (eg. http://cesiumjs.org/Cesium/Apps/Sandcastle/index.html?src=KML.html&label=Showcases ), which looks quite good. Would it be easy to do that?

chloeleichen:
@RacingTadpole  @hilarycinis is ok with the going to the max-width
The Cesium slide-in effect is not hard, but it changes the way the info box interact with other elements on the page, so don't think we'll do that for now

RacingTadpole:
Great, thanks Chloe. Do you want to submit a pull request?

chloeleichen:
hmm... isn't this already a pull request?

RacingTadpole:
oh yeah :-)
@kring, do you want to take a look at this again?

hilarycinis:
Checked @chloeleichen 's prototype, looks fine so lets go with that.

kring:
@chloeleichen and @hilarycinis this is a nice improvement, but it makes the info boxes always horizontally quite large (1000 pixels), even when they don't need to be.  On a large screen and with an infobox with a lot of content, it looks fine.  But some screens are small and some infoboxes have only a little content, and I think it looks pretty bad in that case.  It would be really nice if we could keep the dynamic sizing.

Here's an idea: instead of hiding the sections entirely (with a `ko if`), what if we just set the height of the hidden sections to 0.  We'd have to set the overflow, too, of course, to keep the content from being visible or creating scroll bars.  I _think_ that would make the box automatically size itself to the widest of its sections.  Does that sound like it might work @chloeleichen?

chloeleichen:
@kring but you can't apply css transition to a width/height that's `auto` :(

kring:
@chloeleichen yeah understood, but I think if we're sizing to the biggest section, we no longer need the transitions at all.

chloeleichen:
yeah that should work. I'll have a try later this afternoon @kring

chloeleichen:
@kring did it as you described, please review :)

kring:
This looks much better, thanks @chloeleichen!  The only thing I might change, if it's not too difficult, is to keep the same width even when all sections are collapsed.

chloeleichen:
That'll be a simper solution, but don't you want to have more screen space for the map when the sections are collapsed? @kring

kring:
That'd be nice, but I think the jump is worse than the extra horizontal space!

hilarycinis:
@chloeleichen what's causing the full width to be so wide?

chloeleichen:
@hilarycinis with my previous implementation, when a section is open, it goes to the max width, the max-width calculated based on the screen size etc. With this we can have a smooth transition. 

With the current implementation suggested by Kevin, the info box is the size of its widest section. So if i click on 10 narrow sections, the info box will be narrow, but if I click on 9 narrow section and 1 wide section, the info box will always be as wide as the widest section. In this way currently we can't have a smooth transition, but the width of the info box will not change either once it's opened.

hilarycinis:
@chloeleichen thanks. Lets review tomorrow and discuss. I understand Kevin's thoughts so maybe we can set a max % width that is smaller eg 50%

kring:
> maybe we can set a max % width that is smaller eg 50%

I don't think so.  That will look bad (e.g. horizontal scroll bars, tall skinny table columns) when the content of the infobox is naturally wide.  Dynamic sizing is what we want, and we also don't want that popping effect.  @chloeleichen's current implementation is good on both fronts, so I don't think there's much to discuss!

The only thing that _might_ be slightly better is if the panel were dynamically sized to the currently open section (rather than the largest section, opened or closed), and we had an animated transition between the sizes when expanding/collapsing sections.  But I think that'll be pretty hard to get working, and I'm not convinced it'll actually be better in the end (I can argue either way), so I wouldn't worry about it unless @chloeleichen or someone else wants to try it just for the challenge. :)

hilarycinis:
ok :)

kring:
Thanks @chloeleichen!

PR #1017, opened 2015-10-30
Use latest jsdoc, stop using gulp-jsdoc, hook to tutorials

RacingTadpole:
I've removed the dependency on gulp-jsdoc, which is not maintained any more, and use gulp-shell to access the command line version of jsdoc instead.
Even better, this lets us:
- use the latest jsdoc, still built using the same `gulp docs` command
- build our tutorials as part of the doc process
- link from the docs to our tutorials
- lets us include static files (eg. images) in our tutorials


RacingTadpole:
btw - not sure what the branch "jsdoc" is for, but it's 510 commits behind master - can we delete it?

PR #1018, opened 2015-10-30
better feature naming for TableDataSource

RacingTadpole:
I noticed that the feature info box was just showing "Site Data" for the collapsible heading titles for any csv data.
It now looks for "name", "station name" and "id" fields in that order to use instead, falling back to "Site Data" if none found.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
+### 1.0.47
+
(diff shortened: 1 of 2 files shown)


RacingTadpole:
I've changed the logic to match Cesium's geojson naming scheme, ie.: in this order, match:
- title
- title (case insensitive)
- name (case insensitive)
- containing title (case insensitive)
- containing name (case insensitive)
  If none are found, fall back to "Site Data" (the existing name).

PR #1022, opened 2015-11-02
Make it possible to disable CSV region mapping warnings 

meh9:
…with the `showWarnings` init parameter.

### DIFF
--- a/lib/Models/CsvCatalogItem.js
+++ b/lib/Models/CsvCatalogItem.js
@@ -100,7 +100,7 @@ var CsvCatalogItem = function(terria, url) {
     /**
      * Should any warnings like failures in region mapping be displayed to the user?
      * @type {Boolean}
-     * @default false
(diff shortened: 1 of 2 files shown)


meh9:
@stevage what do you think of this way to control it?

lib/Models/CsvCatalogItem.js
```diff
@@ -98,6 +98,13 @@ var CsvCatalogItem = function(terria, url) {
     this.keepOnTop = false;
 
     /**
+     * Should any warnings like failures in region mapping be displayed to the user?
+     * @type {Boolean}
+     * @default false
+     */
+    this.showWarnings = true;
```
kring, line 105: The actual default doesn't match your documented default.

kring:
Looks good other than the one doc thing, so I fixed it and merged this.  Thanks @meh9!

PR #1027, opened 2015-11-03
use templating engine, update tests

### LINKED ISSUE #1025: Use Handlebars (or equiv) in feature info box templates
As suggested in #955 for issue #496

RacingTadpole:
Interprets feature info templates using [Handlebars](https://github.com/wycats/handlebars.js/).

I chose Handlebars because it has #if, #each, #with helpers, but is otherwise consistent with (indeed built on) [Mustache](https://github.com/janl/mustache.js), so is fairly familiar to many people.

Other alternatives include [underscore templating](http://underscorejs.org/#template) and [dust](https://github.com/akdubya/dustjs). There's a good comparison done by LinkedIn [here](https://engineering.linkedin.com/frontend/client-side-templating-throwdown-mustache-handlebars-dustjs-and-more) - they found dust, mustache, handlebars and google closures to be the top four. It looks like dust has a few more features - but its git repo hasn't been updated for 4 years.

Fixes #1025.


RacingTadpole:
Note one downside of using a full templating engine is size. Unminified, the build appears to be 300K bigger with handlebars than with our simple `readTemplate` version.

```
git checkout master
ls -al wwwroot/build/TerriaJS-specs.js

3989332  3 Nov 11:27 wwwroot/build/TerriaJS-specs.js

git checkout 1025_handlebars
ls -al wwwroot/build/TerriaJS-specs.js

4294020  3 Nov 11:44 wwwroot/build/TerriaJS-specs.js
```

kring:
Wow, 300k is quite a lot!  How much bigger is the minified version?

RacingTadpole:
OK, the minified aremi build goes from 2651213 to 2758613, ie. handlebars adds 100K to the minified result. Wow.

RacingTadpole:
Changed it to use mustache instead. It doesn't have #if, but it does have an equivalent to #each. And it can recurse using partial templates (see the example in [featureInfoPanelSectionViewModelSpec](https://github.com/TerriaJS/terriajs/blob/1025_handlebars/test/ViewModels/FeatureInfoPanelViewModelSpec.js#L127-L163)).
The rationale is that it only adds 5 K to the minified build instead of 100K for handlebars.

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -82,7 +83,17 @@ function htmlFromFeature(viewModel, clock) {
...
-        return readTemplate(data, viewModel.template);
+        // template may be a string, eg. '<div>{{{Foo}}} Hello {{name}}</div>'
+        if (typeof viewModel.template === 'string') {
+            return Mustache.render(viewModel.template, data);
+        } else {
+            // or template may be an object with a main 'template' key, and additional partials
+            // eg. {template: '<div>test {{>foobar}}</div>', foobar: '<b>{{Foo}}</b>'}
+            var template = viewModel.template.template;
+            var partials = viewModel.template;
+            delete partials.template;
```
kring, line 96: This `htmlFromFeature` can be called multiple times, right?  So deleting this here will mean it's missing the next time around.

kring:
This looks good other than that one comment.  I think it's reasonable to merge this even if it doesn't completely address @axman6's templating needs, because it's clearly better than what we have in master now and more likely to be forward compatible.

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
RacingTadpole, line 96: nice one. yes, fixed (via clone).

axman6:
Yeah don't hold out because of us, just as long as the features we need are kept in mind, we can stick with our hacks for now.

PR #1032, opened 2015-11-04
New multi feature info box should use owner; csvCatalogItem has a dataSource

RacingTadpole:
Change to using "owner" not "dataSource" due to change in https://github.com/AnalyticalGraphicsInc/cesium/pull/3130 .

Update csvCatalogItem so it too has a dataSource.


RacingTadpole:
Not sure why the travis build is failing - when I do `gulp lint docs release` on my machine, it passes...?

PR #1038, opened 2015-11-06
FeatureInfoTemplates work even without imageryLayers

### LINKED ISSUE #1014: FeatureInfoTemplate only works if there is an imagery layer
Although featureInfoTemplate is defined on the CatalogItem, it is only actually displayed if it is on an imagery layer. We should make this work on all item types, eg. geojson.

RacingTadpole:
Fixes #1014. The fix itself only touches three lines. 

Note the `imageryLayer.featureInfoTemplate` property is not needed any more.

However it makes the tests (specs) trickier to write, since:
- featureInfoTemplate can no longer be defined at the feature level, but must be at the CatalogItem level
- the nowViewing model is needed to compare data sources to find the CatalogItem corresponding to this feature
- so we also need to maintain DataSources

The resulting tests are in a new `describe` call, "FeatureInfoPanelViewModel templating".

As a byproduct, this is effectively a new test for the Catalog model, so I have added that in its own spec file too.


kring:
@RacingTadpole mind merging in master and updating `CHANGES.md`?

RacingTadpole:
Thanks. I've merged and added changes.md now.

PR #1043, opened 2015-11-09
Filter out very similar Gazetteer results. Fixes #739.

### LINKED ISSUE #739: Place Name Search duplicates
Hi Team,

I've been asked to raise an issue with the "Official Place Names" search duplicate results. For example, "Wagga Wagga". "Perth" is another example, 3 results with same name and coordinates.

image

I understand these duplicates are due to the GA web service exposing a dataset that is probably not  suited to the searching common locations within NM, but would it be possible to filter out results on the client to remove entries with the same name and coordinates? 

Since you are only displaying the name and the click action is to zoom, this should at least make it perceived as removing duplicates. I can see that you are already doing some client side manipulation of the result to better handle data quality/applicability issues from the web service. 

There will be some results that have the same name and only slightly different coordinates, but this will at least improve results for users.

stevage:
Weirdly, I think I forgot to actually submit a pull request for this back in July. Oops!

### DIFF
--- a/lib/ViewModels/GazetteerSearchProviderViewModel.js
+++ b/lib/ViewModels/GazetteerSearchProviderViewModel.js
@@ -124,7 +124,7 @@ function stripDuplicates(results) {
         var lng = Number(results[i].location.split(',')[1]).toFixed(1);
 
         var hash = results[i].name + '_' + lat + ' ' + lng;
-        if (placeshash[hash] === undefined) {
+        if (!defined(placeshash[hash])) {
             placeshash[hash] = results[i];
             stripped.push(results[i]);
         }
--- a/lib/ViewModels/GazetteerSearchProviderViewModel.js
+++ b/lib/ViewModels/GazetteerSearchProviderViewModel.js
@@ -91,6 +91,7 @@ GazetteerSearchProviderViewModel.prototype.search = function(searchText) {
         if (defined(solrQueryResponse.response) && solrQueryResponse.response.numFound > 0) {
             var results = solrQueryResponse.response.docs.sort(
                 function (a,b) { return sortResults(a,b,searchText); } );
+            results = stripDuplicates(results);
             results.slice(0,10).forEach(function(result) {
                 that.searchResults.push(new SearchResultViewModel({
                     name: result.name + (result.state_id !== 'N/A' ? ' (' + result.state_id + ')' : ''),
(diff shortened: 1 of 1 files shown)


lib/ViewModels/GazetteerSearchProviderViewModel.js
```diff
@@ -113,6 +114,24 @@ GazetteerSearchProviderViewModel.prototype.search = function(searchText) {
...
+function stripDuplicates(results) {
+    var i;
+    var placeshash={};
+    var stripped = [];
+    for (i=0; i < results.length; i++) {
+        var lat = Number(results[i].location.split(',')[0]).toFixed(1);
+        var lng = Number(results[i].location.split(',')[1]).toFixed(1);
+
+        var hash = results[i].name + '_' + lat + ' ' + lng;
+        if (placeshash[hash] === undefined) {
```
kring: Use `!defined(placeshash[hash])`.

lib/ViewModels/GazetteerSearchProviderViewModel.js
```diff
@@ -113,6 +114,24 @@ GazetteerSearchProviderViewModel.prototype.search = function(searchText) {
     });
 };
 
+// Given a list of results sorted in decreasing importance, strip results that are close to another result with the same name
+function stripDuplicates(results) {
+    var i;
+    var placeshash={};
+    var stripped = [];
+    for (i=0; i < results.length; i++) {
+        var lat = Number(results[i].location.split(',')[0]).toFixed(1);
```
kring, line 123: `toFixed(1)` means places that are actually pretty far apart will be considered the same place, right?

kring:
Looks good other than those two comments.

lib/ViewModels/GazetteerSearchProviderViewModel.js
stevage, line 123: Yes. "Pretty far apart" meaning 0.1 degrees of latitude or longitude, or ~20km. This was intentional, so the user isn't presented with meaningless distinctions like "Sydney", "City of Sydney", etc.

kring, line 123: Is the result much worse if we use, say, 0.01 degrees instead of 0.1?  20km seems far enough apart that it could have meaningfully distinct results with similar names within that distance.

stevage:
Ready for re-review @kring

Issue #907, opened 2015-09-09
Allow base maps to be specified in init files


JJediny:
Testing/Playing around with terriajs through a fork of the nationalmap. One observation is that in order to [modify Basemaps](https://github.com/TerriaJS/terriajs/blob/master/lib/ViewModels/createAustraliaBaseMapOptions.js)

I'm pretty sure I'll have to use a modify this in TerriaJS -> and then manually compile TerriaJS instead of installing through npm as another stable dependency. Is there any reason why basemaps couldn't instead be handled as [a json file in the map wrapper under datasources directory](https://github.com/TerriaJS/map/tree/master/datasources) as datasources are currently from template/concatenating and load when the client is accessed [from the "Map" wrapper](https://github.com/TerriaJS/map)? or Can they be currently overwritten if so where/how would this be done

kring:
Hi @JJediny, thanks for checking out TerriaJS!  To modify the list of default base maps, there's no need to actually modify TerriaJS itself.  Instead, copy the entire source file from TerriaJS into your own project, modify it as appropriate, and use your function instead of the one in TerriaJS.

I do like your idea of specifying base maps using a JSON config file, though.  We may do that at some point, and we'd certainly welcome a pull request for it if you're up for it.

JJediny:
Without having dived in (yet) I guess the approach would be dependent on if Basemaps are currently handled any differently at run time. I saw three of the Basemap services used the ESRIgroup and WMS datasource object type - add X/Y/Z tile service (or lesser used TMS or ESRI's WMTS) and that's about all the services one could pick for available/external basemap services...

From past work in OL3 and Leaflet - I recall these libraries define whether a datasource is 'basemap' or 'layer' as a variable of the datasource object (ex. isBasemap: "True") with default set to False... ? Seems like that wouldn't disrupt the great work you all are putting into the datacatalog schema but get the same effect of managing basemaps as json datasource directory, thoughts?

meh9:
@JJediny, controlling the basemaps from the config does indeed sound like a nice improvement. Another simple option in the meantime, if you just want to add a particular new basemap, is to use the existing Terria function and just add your own basemap to the list:

https://github.com/NICTA/aremi-natmap/blob/2015-08-26/index.js#L157

JJediny:
@meh9 Thank you this is exactly what I was looking for - a place to hardcore new basemaps before load... If there's conscience on the best approach to move this into a json configuration let me know and I can try to pull together a PR?

[...]

PR #1046, opened 2015-11-09
Basemap json config

JJediny:


JJediny:
DO NOT PULL... Placeholder for Issue #907 

Approach for discussion is to extend CatalogItem w two new variables in `isBasemap` and `thumbnailUrl`. isBasemap to be used as a variable within an `item` as it avoids defining a new BaseMapModel and avoids creating a new highlevel filter to separately handle a new process for readJson for both `Basemap` and `Catalog`. This way [CompositeCatalogItem](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CompositeCatalogItem.js) need only be modified or emulated with a new model for Basemap that only sets an item when isBasemap = True...

Does this make sense - also not confident with my JavaScript skills on the next/best step?

kring:
Thanks @JJediny!  As we've discussed before, I do think it's reasonable to add a `thumbnailUrl` to `CatalogItem`.

However, I'm not so sure about the `isBasemap` bit.  It seems like you're heading toward a pattern where TerriaJS would search through the entire catalog to find all the base maps, and populate the base map selector from that.  But that concerns me on two fronts:
- it requires that all base maps be in the catalog, and
- it will take a long time, and generate lots of requests, to search through the entire catalog.

I think it would be better to add a new section to the JSON init file, `baseMaps`, to explicitly define the list of basemaps.  It can be deserialized in the same way as `catalog` so it shouldn't take more than a few lines of code to do it.

Does that sound reasonable?

JJediny:
Absolutely... Wasn't sure if catalog would be filtered once and parse out basemaps concurrently - but if it would require an additional iteration over catalog items then completely agree... I'll take another stab but it might be helpful If we can schedule a quick google hangout to knock it out in ~20 or so next week?

kring:
> Wasn't sure if catalog would be filtered once and parse out basemaps concurrently

We load sections of the catalog on demand.  For example, if a group is populated by querying a CKAN server, we don't query the CKAN server until the group is opened.  If we made the assumption that base layers would only be found in non-delay-loaded groups, performance would be fine.  But we'd still have the first problem I mentioned above.

>  it might be helpful If we can schedule a quick google hangout to knock it out in ~20 or so next week

Ok, I'll email you!

rsignell-usgs:
@kring & @JJediny , did indeed some offline (google hangout) progress take place?

kring:
@JJediny I'm going to close this for now, but feel free to re-open when something new needs review or discussion.

PR #1047, opened 2015-11-09
Allow templates for feature names

### LINKED ISSUE #1041: Let the init file specify features' name field
As per #1018, csv and geojson features are named in the feature info panel according to this rule: in this order, match:
- title
- title (case insensitive)
- name (case insensitive)
- containing title (case insensitive)
- containing name (case insensitive) 
- if none are found, csv falls back to "Site Data"'; I haven't checked what happens with geoJson.

It would be nice if we could specify the name field in the init file, so the info box doesn't fill up with lots of sections all called `Site Data`.

We could either add a new property, eg. `featureNameField` (alongside `featureInfoTemplate`), or else we could allow `featureInfoTemplate`, which can already be a json object with a `template` key (see #1027), to also take a `nameField` key.

I haven't thought yet about whether this has applicability beyond csv and geoJson datatypes.

RacingTadpole:
Fixes #1041. Note this is a pull request that adds on to PR #1038.


meh9:
Works really nice.

PR #1048, opened 2015-11-09
Improve the style of Markdown/HTML using the markdown binding

chloeleichen:
To test, go to http://127.0.0.1:3001/#build/TerriaJS/test/init/arcgis.json and select a feature from Gravity Anomaly (single dynamic layer).

@kring:  currently the `markdown` class is hardcoded in the template, not sure if this is the right way to go? 

We should probably create a styleguide to demo and document these styles?


kring:
@chloeleichen the class looks good to me.  Rather than requiring the users to specify it in the template, though, we should add it to any div that uses the `markdown` binding.  e.g. in `CatalogItemInfo.html`, change this:

```
<div class="catalog-item-info-section">
    <h2>Description</h2>
    <div class="catalog-item-info-description" data-bind="markdown: catalogItem.description"></div>
</div>
```

To this:

```
<div class="catalog-item-info-section">
    <h2>Description</h2>
    <div class="catalog-item-info-description markdown" data-bind="markdown: catalogItem.description"></div>
</div>
```

We could also change `KnockoutMarkdownBinding.js` so that every element with a `markdown` binding automatically gets this class applied, rather than doing it manually in each of the views.  I can't decide if that's a great or terrible idea. :)

chloeleichen:
@kring added `markdown` class to `KnockoutMarkdownBinding.js`

PR #1050, opened 2015-11-09
Limit how many features are shown in the featureInfoPanel

### LINKED ISSUE #1010: Support FEATURE_COUNT parameter for GetFeatureInfo requests
`GetFeatureInfo` requests by default return information for a single feature under the location clicked (default `FEATURE_COUNT=1`). With the new work to show information from each layer in #994 it would now be good to specify the `FEATURE_COUNT` to get all features. Ideally it would be possible to specify both an instance wide default and layer specific setting for value used.

RacingTadpole:
Fixes #1010. It limits the number of features in the featureInfoPanel to 100 per catalog item. If the limit is exceeded, a message appears at the top of the panel.

The limit of 100 is configurable in two ways: globally and per catalog item.

To change it globally, add to your `config.json` a line like `"defaultMaximumShownFeatureInfos": 25` inside `parameters`.

To change it for a single catalog item, add a property "maximumShownFeatureInfos", eg.:

```
    {
      "name": "my item",
      "url": "http://example.com/geoserver/wms",
      "type": "wms",
      "maximumShownFeatureInfos": 1000,
      "parameters": {
        "tiled": true
      }
```

Note an error is now thrown if WMS sets `parameters.feature_count` directly, since this would conflict with the new parameter.

Finally, for the record, we thought about setting the limit very high instead (eg. 1,000,000,000), and letting the catalogItem itself restrict the number if relevant, but we want the UI to work in a reasonable way out of the box.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Change Log
   * Added support for the glTF extensions [KHR_binary_glTF](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_binary_glTF) and [KHR_materials_common](https://github.com/KhronosGroup/glTF/tree/KHR_materials_common/extensions/Khronos/KHR_materials_common).
   * `ImageryLayerFeatureInfo` now has an `imageryLayer` property, indicating the layer that contains the feature.
   * Make KML invalid coordinate processing match Google Earth behavior. [#3124](https://github.com/AnalyticalGraphicsInc/cesium/pull/3124)
+* Show information for all WMS features when a location is clicked.
 
 ### 1.0.46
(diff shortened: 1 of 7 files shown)


meh9:
Works beautifully.

RacingTadpole:
Thanks @meh9.  I had one further thought about this - it is (technically) not backwards compatible that the `parameter.feature_count` now throws an error. I'm thinking I should change this before the release so that feature_count will still work - and an only throw an error if both are set.

meh9:
Well, sure, technically it isn't backwards compatible, but I have never seen anybody else use it, of course that's no guarantee...

PR #1053, opened 2015-11-10
Linkify content in feature info panel

### LINKED ISSUE #1045: Make URL in feature info into clickable link
I previously raised a narrower version of this as #316 which Keith closed (I think incorrectly).  Rather than reopen that, I'm creating a broader one that doesn't just relate to WMS services (eg this should be done for CSVs as well).

RacingTadpole:
Fixes #1045.

Builds the feature description from properties _with markdown applied to each property individually_ - for WMS, csv and geojson types.

In particular, all links become linkable in the FeatureInfoPanel.  (They were not before, because markdown was only being applied to the entire table, and markdown does not drill into a table if it is the only thing supplied.)

As before, you can override this behaviour by providing your own featureInfoTemplate.

Because markdown puts a `<p>` tag around a string, I have also turned off Chloe's new `<p>` padding-bottom. I'm sure there's a better approach to this.


lib/Core/KnockoutMarkdownBinding.js
```diff
@@ -10,6 +9,7 @@ var KnockoutMarkdownBinding = {
     allowUnsafeHtml : false,
 
     register : function(knockout) {
+        var allowUnsafeHtml = this.allowUnsafeHtml;
```
kring: This is a pretty minor point, but you should do `var that = this` here and then access `that.allowUnsafeHtml` later, instead of saving `this.allowUnsafeHtml`.  That way changing the value after registration will have an effect.

RacingTadpole: Very nice - thanks.

lib/Core/formattedPropertyValue.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+* Format the value for the description, used by the Feature Info Panel.
+* Strings have markdown applied to them, and numbers are formatted with commas.
+* Anything else is returned as-is.
+*
+* @param {} value The value to format.
+*
+*/
+function formattedPropertyValue(value) {
```
kring: Another nitpicky thing: how about `formatPropertyValue` (format acts like a verb) instead of `formattedPropertyValue` (formatted acts like an adjective).  Functions do things, so they should be verbs.

kring: The same is true of `numberWithCommas` above, but it's not part of the public API so it doesn't matter as much.

lib/Models/GeoJsonCatalogItem.js
```diff
@@ -162,12 +163,15 @@ var simpleStyleIdentifiers = ['title', 'description', //
...
     var html = '';
-    for ( var key in properties) {
+    for (var key in properties) {
         if (properties.hasOwnProperty(key)) {
             if (key === nameProperty || simpleStyleIdentifiers.indexOf(key) !== -1) {
                 continue;
             }
             var value = properties[key];
+            if (typeof value === 'string') {
+                value = markdownToHtml(value, false);
```
kring: Any reason not to use `formattedPropertyValue` here too?

kring:
> Because markdown puts a `<p>` tag around a string, I have also turned off Chloe's new `<p>` padding-bottom. I'm sure there's a better approach to this.

I've re-added Chloe's `<p>` styles, and also added one that removes the `margin-bottom` from the last element.  I think that's a reasonable way to handle it.

RacingTadpole:
Excellent comments, thanks. Back to you...

PR #1062, opened 2015-11-11
Color map

stevage:
Turns TableStyle and ColorMap into classes, and allows `colorMapString` for csv files.


kring:
@stevage is this still a work in progress, based on our conversation yesterday?  Or do you want to merge this now and consider further improvements separately?

stevage:
@kring I think this is complete, if not yet earth-shattering. Merging it would be better than not merging it, I believe.

Issue #137, opened 2014-07-03
Reduce size of embed text box in share


kgrochow:
The embed text is only going to get bigger and uglier, so we should make the embed url, just like the share url and hide 99% of it.  Given that I would also just move both under the image and possibly increase the size of the image a bit.  I talked with Hilary about this and she agrees in principal if not in specifics.

PR #1066, opened 2015-11-12
Ga color basemap

stevage:
This is for https://github.com/NICTA/nationalmap/issues/137


kring:
Now that the GA map covers the whole world, let's get rid of that Natural Earth II layer underneath it.  That will improve performance a bit.

PR #1073, opened 2015-11-13
Improve WMS Title matching.

kring:
`WebMapServiceCatalogItem` had some code to try to match each of its `layers` against either the `Name` or `Title` property in `GetCapabilities`.  This was intended to be a usability feature.  You can specify which layer to use by its human-oriented name rather than just by its identifier.  But it had some problems:
- It could match by `Title` even when a better match by `Name` existed (e.g. on a child layer).
- It didn't require that the matched-by-Title layer have a `Name` property.  Only layers with a `Name` are valid for `GetMap` requests.
- Upon finding a match by title, it wouldn't use the `Name` of the matched layer, so the `GetMap` would likely fail.

So this change does two things:
- It only tries for a match against `Title` when there is no match by `Name`, and it requires that the matched layer have a `Name` (even if it doesn't match).
- It uses the `Name` of the matched layers in later `GetMap` requests.

CC @meh9

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Change Log
 * Supports FeatureInfoTemplates on all catalog item types (previously only available on ImageryLayers).
 * Apply markdown to properties shown in the Feature Info Panel.
 * Add `includeCzml` option to CkanCatalogGroup.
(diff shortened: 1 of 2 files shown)


meh9:
Looks good to me, everything I've tested shows the Abstract/Data Description for the layer now.

PR #1076, opened 2015-11-13
Allow tags to pass through formatPropertyValue

RacingTadpole:
Here's a better fix than moving `formatPropertyValue` to `Models` - just let unknown tags pass through it, by passing `allowUnsafeHtml = true`.

This sounds scary, but it is no worse that what it was doing before we passed it through markdown.  

In fact, the property value will be passed through the markdown knockdown binding when it is actually displayed anyway, with `allowUnsafeHtml = false` at that point.

And I think conceptually this is more "core" than "models".


kring:
Ah, clever!

PR #1077, opened 2015-11-13
Add support for navigation with the keyboard

### LINKED ISSUE #507: Data navigation - Should be able to use TAB key to navigate data/now viewing selections

chloeleichen:
@kring : Done :D

*removed blue box
*close modal window with esc key
*keyboard navigation 

Fixes #507


lib/Core/escClose.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+var defined = require('terriajs-cesium/Source/Core/defined');
+
+var escClose = function(obj){
```
kring: How about a more descriptive name, without abbreviations?  Something like `closeWhenEscapeIsPressed`.

lib/ViewModels/DataCatalogTabViewModel.js
```diff
@@ -32,6 +33,18 @@ var DataCatalogTabViewModel = function(options) {
...
     this.svgArrowRight = defaultValue(options.svgArrowRight, svgArrowRight);
     this.svgInfo = defaultValue(options.svgInfo, svgInfo);
+
+    this.panelHasFocus = knockout.observable(false);
+
+    var that = this;
+
+    knockout.getObservable(this, 'isActive').subscribe(function(newValue) {
+        if (newValue) {
+            window.setTimeout(function(){
```
kring: Why is this `setTimeout` necessary?  In my (limited) testing it seems to work ok without it.

lib/ViewModels/NowViewingTabViewModel.js
```diff
@@ -57,8 +58,13 @@ var NowViewingTabViewModel = function(options) {
             if (!oneIsOpen && nowViewingItems.length > 0) {
                 nowViewingItems[0].isLegendVisible = true;
             }
+
+            window.setTimeout(function(){
```
kring, line 62: Same question here.

kring:
Thanks @chloeleichen, this looks great other than those couple of comments!

lib/ViewModels/DataCatalogTabViewModel.js
chloeleichen: I think you are right.

lib/ViewModels/NowViewingTabViewModel.js
chloeleichen, line 62: @kring  I think this is needed, otherwise when you switch from data catalogue tab to now viewing tab using the keyboard, it breaks the transition :)

kring:
Thanks @chloeleichen!

PR #1078, opened 2015-11-13
Add Urthecast API as catalog member type

### LINKED ISSUE #964: Add Urthecast API as catalog member type
Possible API library candidate: https://github.com/missinglink/urthecast

Bounty added: https://www.bountysource.com/issues/27290954-add-urthecast-api-as-catalog-member-type

jasonjlock:
Add Urthecast catalog group that contains available sensor platforms
(theia, landsat 8) and available map tile service renderers as catalog
items.
- Add Urthecast API client as a dependency
- Add sinon.js as a dependency to help mock XHR

resolves #964

### DIFF
--- a/lib/Models/UrthecastCatalogGroup.js
+++ b/lib/Models/UrthecastCatalogGroup.js
@@ -1,0 +1,140 @@
+'use strict';
+
+/*global require*/
+var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
+var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
+
+var ModelError = require('./ModelError');
+var raiseErrorToUser = require('../Models/raiseErrorToUser');
+var CatalogGroup = require('./CatalogGroup');
+var inherit = require('../Core/inherit');
+var UrthecastCatalogItem = require('./UrthecastCatalogItem');
+var urthecastClient = require('urthecast');
+
+/**
+ * A {@link CatalogGroup} representing a collection of avalible Urthecast sensor platforms.
+ *
+ * @alias UrthecastCatalogGroup
+ * @constructor
+ * @extends CatalogGroup
+ *
+ * @param {Terria} terria The Terria instance.
+ */
+var UrthecastCatalogGroup = function(terria) {
+    CatalogGroup.call(this, terria);
+
+    /**
+     * Gets or sets a hash of names of blacklisted data layers.  A layer that appears in this hash
+     * will not be shown to the user.  In this hash, the keys should be the Title of the layers to blacklist,
+     * and the values should be "true".  This property is observable.
+     * @type {Object}
+     */
+    this.blacklist = undefined;
+
+    knockout.track(this, ['blacklist']);
+};
+
+inherit(CatalogGroup, UrthecastCatalogGroup);
+
+defineProperties(UrthecastCatalogGroup.prototype, {
+    /**
+     * Gets the type of data member represented by this instance.
+     * @memberOf UrthecastCatalogGroup.prototype
+     * @type {String}
+     */
+    type : {
+        get : function() {
+            return 'urthecast-group';
+        }
+    },
+
+    /**
+     * Gets a human-readable name for this type of data source, such as 'Urthcast Sensor Platforms'.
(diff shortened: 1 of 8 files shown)


jasonjlock:
We tried to match the code and commit styles, but feel free to let me know if we didn't get it quite right. 

Might want to add Deimos-1, Deimos-2, and Iris (HRC) sensor platforms to the config blacklist since that data is not available yet.

kring:
Thanks @jasonjlock!  I'm waiting on an API key from Urthecast so I can try it out.  I'll add a few comments in the meantime.

lib/Models/UrthecastCatalogGroup.js
```diff
@@ -0,0 +1,140 @@
...
+    typeName : {
+        get : function() {
+            return 'Urthcast Sensor Platforms';
+        }
+    },
+});
+
+UrthecastCatalogGroup.prototype._load = function() {
+    var catalogGroup = this;
+    var apiKey = catalogGroup.terria.configParameters.urthecastApiKey;
```
kring, line 67: We should inform the user when the API key and secret are not set, by throwing a `ModelError` with an appropriate `title` and `message`.  It took me a minute to realize why nothing was working.

lib/Models/UrthecastCatalogGroup.js
```diff
@@ -0,0 +1,140 @@
...
+    var apiSecret = catalogGroup.terria.configParameters.urthecastApiSecret;
+    var urthecast = urthecastClient(apiKey, apiSecret);
+
+    urthecast.v1.satellite_tracker('/sensor_platforms', {}, function(err, res) {
```
kring: This asynchronous bit needs to be turned into a promise, and the promise needs to be returned, so that the group displays "Loading..." while the load is in progress instead of displaying "This group contains no items."  You can turn a normal callback into a promise by creating a deferred:

```
var deferred = when.defer();
```

And then returning its promise:

```
return deferred.promise;
```

Inside the callback, on success:

```
deferred.resolve()
```

And on failure:

```
deferred.reject();
```

If you pass a `ModelError` instance to `reject`, TerriaJS will display it to the user.

kring:
This functionality adds over 18k to the minified version of TerriaJS.  Since it also requires an API key, and most apps won't use it, we should make it opt-in rather than included by default.  To do that, remove it from `registerCatalogItems` and instead add a `registerUrthecastCatalogItems` function that does the same thing for just the Urthecast catalog item and group.  Apps that want to use Urthecast only need to require-in that function and call it, while apps that don't won't include the code at all.

kring:
It'd be great to include an example init file with the Urthecast group in TerriaJS.  You could put it in `wwwroot/test/init/urthecast.json` and it would look like this (I think):

```
{
  "catalog": [
    {
      "name": "Urthecast",
      "type": "urthecast-group"
    }
  ]
}
```

Then, you can use it in any TerriaJS with Urthecast support by visiting: `http://www.example.com/#build/TerriaJS/test/init/urthecast.json`.

kring:
This looks good other than those few comments, thanks!  I may have a few more once I get an API key and can try it out.

kring:
@jasonjlock I got an email back from Urthecast, and it sounds like it could be weeks before they give me an API key.  Maybe I could borrow yours for a few minutes so I can try this out?  My email address is on the web page linked from my GitHub profile.

jasonjlock:
@kring Awesome feedback, thank you. I'll work to get this updated today! You should now have access to developers.urthecast.com, from there you can obtain your API key and secret (assuming you used the email address mentioned above). Please let me know if you have any questions or run into any issues getting your API credentials.

kring:
Thanks @jasonjlock!

kring:
@jasonjlock I almost forgot: would you mind opening a pull request to add yourself to `CONTRIBUTORS.md`?

RacingTadpole:
One of the new tests is failing - `UrthecastCatalogGroup creates hierarchy of catalog items`. It’s not immediately obvious to me what’s going on… can one of you please take a look?

jasonjlock:
Apologies, taking a look now. Added #1101.

PR #1082, opened 2015-11-17
Dispose of feature info panel sections after close

RacingTadpole:
This fixes this issue: "We have an interesting thing happen on our OSX laptops using Chrome, when viewing ACMA, after having clicked somewhere so the "more than feature_count" warning message is shown, after you close the feature info box the globe rendering speed drops to about 1fps..."

It's not an issue while the feature info box is open - only after you close it, and only if there were lots of features shown (but not necessarily the full feature_count). Reopening the info box fixes it. And it's happening without any clock listeners.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
+### 1.0.49
+
+* Fixed a bug that caused poor performance when clicking a point on the map with lots of features and then closing the feature information panel.
+
 ### 1.0.48
 
 * Added the ability to disable feature picking for `ArcGisMapServerCatalogItem`.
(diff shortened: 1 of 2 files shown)


kring:
Wow, this is really interesting.  The profiler shows that not much is happening in our code while things are slow, except that a bit of time is spent in `getBoundingClientRect` (less than 1%).  But if you use the timeline view, you can see that Chrome is spending a massive amount of time in "Update Layer Tree".  So apparently, Chrome's layout gets extremely slow when that panel exists and has a lot of sections, but has `visibility: hidden`.  Maybe it has something to do with the fact that we're putting `pointer-events: none` on it as well... so we end up with a window that is "there" but that the browser has to ignore for mouse input events.  Or something.  In any case, removing the sections seems like a good idea.  Setting `display: none` might help, too.

kring:
Even setting `display: none` doesn't fix the performance problem (without also removing the sections).  So strange.

RacingTadpole:
Thanks for looking into it too - and for updating CHANGES for me... hopefully I'll remember that one day...

PR #1086, opened 2015-11-17
Show Feature Info properties using linkify, not markdown

### LINKED ISSUE #1084: markdown is displaying "-" as "<li></li>" in feature info
As raised in the comments of #1045... markdown is rendering some data badly. We may need to use linkify instead of markdown to display.

RacingTadpole:
Fixes #1084.
Note the `linkify-it` package doesn't explicitly include a function to linkify a string. The `linkifyContent` function I define below is based on the example they provide in their [demo](http://markdown-it.github.io/linkify-it/index.js).

### DIFF
--- a/lib/Styles/Core.less
+++ b/lib/Styles/Core.less
@@ -206,9 +206,11 @@ select {
     margin-top: 0;
     margin-bottom: 1rem;
   }
-  p:last-child {		
+  
+  p:last-child {
     margin-bottom: 0;
   }
+  
   strong { font-weight: bold }
   em { font-style: italic }
   small {font-size: 80%;}
--- a/lib/Styles/Core.less
+++ b/lib/Styles/Core.less
@@ -206,11 +206,11 @@ select {
     margin-top: 0;
     margin-bottom: 1rem;
   }
-  
+
   p:last-child {
     margin-bottom: 0;
   }
-  
+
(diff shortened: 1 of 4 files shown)


lib/Styles/Core.less
```diff
@@ -204,11 +204,7 @@ select {
 
   p {
     margin-top: 0;
-    margin-bottom: 1rem;
-  }
-
-  p:last-child {
-    margin-bottom: 0;
+    margin-bottom: 0; // 1rem;
```
kring: Did you mean to change this style?

kring:
I have mixed feelings about this, because markdown in properties is kind of awesome.  But I guess it's the safer way to go.  Users can always use HTML if they want some fancy formatting for a property.  So just that one comment and then I'll merge this.

lib/Styles/Core.less
RacingTadpole: yep - because markdown put everything in `<p>` tags, but linkify doesn't.  So we can go back to the normal `<p>` styling

kring: Normal styling has `margin-bottom: 1rem` not 0, right?  Plus, I think avoiding extra space after the last paragraph is a good thing anyway.

RacingTadpole: ah yes, ok.

PR #1088, opened 2015-11-19
Remove czmlDataSource from TableDataSource - so csv can have infoFeatureTemplates

### LINKED ISSUE #1085: featureInfoTemplate not working for csv data sources
Extend #496, #1038 to csv data sources too.

RacingTadpole:
Fixes #1085 

This is a big internal change, and I'm not completely confident that I've covered off all the necessary changes. But the specs still pass (when updated for the different way of describing colours), and it all seems to work. It even gets featureInfoTemplates working with csv data sources.

### DIFF
--- a/lib/Map/TableDataSource.js
+++ b/lib/Map/TableDataSource.js
@@ -5,9 +5,10 @@
 TableDataSource object for displaying geo-located datasets
 */
 
-var Cartographic = require('terriajs-cesium/Source/Core/Cartographic');
+var Cartesian3 = require('terriajs-cesium/Source/Core/Cartesian3');
 var ClockRange = require('terriajs-cesium/Source/Core/ClockRange');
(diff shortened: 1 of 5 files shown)


RacingTadpole:
I think it works now @kring! Even share works. Thanks.

kring:
The `Test Data -> Temporal -> Auto Incidents (CSV)` dataset (in the `#test` init file) is not correctly choosing the `incident_duration` property as its data variable.  Also, the `scale` / `scaleByValue` bit isn't working (all the icons are the same size).

RacingTadpole:
Thanks. Solved the dataVariable problem. The feature scaling problem has been with us since commit 51a51d5 (which I think is the problem, but it doesn't compile) and/or 05e9dd4 (which does compile), so I'm going to submit a separate PR for that.

RacingTadpole:
OK, together with PR #1094 I think this one now looks ok. I've added some specs this time too.

PR #1090, opened 2015-11-20
Refactor the Now Viewing tab to display configurable sections

kring:
Previously, the various bits shown on the Now Viewing tab were hard-coded, including special handling for various types of catalog items such as `AbsIttCatalogItem`.  This pull request refactors `NowViewingTabViewModel` to instead display a configurable list of sections (`NowViewTabViewModel.sections`).  Each section is given the opportunity to determine for itself whether it applies to each catalog item.  Added `LegendSectionViewModel`, `OpacitySectionViewModel`, `ConceptsSectionViewModel` (for CSV and ABS filtering), and `AbsPercentageSectionViewModel` (for choosing whether ABS data is display as a percentage or an absolute count).

CC @stevage and @RacingTadpole.  Both of you guys mentioned you may find this useful for some other work you are/will be doing.


kring:
This PR also has two bonus features:
- Added `Terria.cesiumBaseUrl` property. (useful when creating a separate `Terria` instance for embedding another map elsewhere in the map)
- The user interface container DOM element may now be provided to `TerriaViewer` by specifying `uiContainer` in its options.  Previously it always used an element named `ui`.  (will probably help some work @chloeleichen is doing)

kring:
@RacingTadpole can you review and merge this when you get a chance?

RacingTadpole:
If there's no legend, it no longer shows "No legend available for this data item." - do you want to bring it back?

kring:
> If there's no legend, it no longer shows "No legend available for this data item." - do you want to bring it back?

I think it's ok.  I removed the message because, in the branch where I originally did this, I have catalog items that do not have legends and that don't even make sense to have legends (specifically, some types of analysis results).

RacingTadpole:
I searched on `'ui'` and found quite a few hardcoded references to it (in 5 files).  Is it worth using your new `uiContainer` for them too?
There are no other refs to `'cesiumContainer'` that I could see.

kring:
Yes, that would be great.  Although, given that a new UI is coming, and these view models (unlike `TerriaViewer`) won't be used in the new UI, I don't think it's a very high priority.

RacingTadpole:
Finally - I know you've just copied this from the existing code, but what's the thinking behind the `TEST` text in the ABS section view? 

Otherwise, this looks great!

kring:
> what's the thinking behind the TEST text in the ABS section view in the ABS section view?

There's no good reason for it, I'll get rid of it.

kring:
Done

lib/Models/CsvCatalogItem.js
```diff
@@ -274,6 +275,19 @@ defineProperties(CsvCatalogItem.prototype, {
         get : function() {
             return this._tableDataSource;
         }
+    },
+
+    /**
+     * Gets the concepts that can be used to filter this data, such as Region Type, Age, etc.
+     * @type {AbsConcept[]}
```
RacingTadpole, line 282: I guess it's a bit late to comment on this now - but just noticed this jsdoc ref isn't right. (also the one in AbsIttCatalogItem)

Issue #1087, opened 2015-11-18
Images in popup message labels have `markdown` CSS style, may render badly.


stevage:
The latest release of National Map has squished images in the Related Maps popup.

![screenshot 2015-11-19 10 06 28](https://cloud.githubusercontent.com/assets/678344/11257372/43d651b0-8ea5-11e5-8fcf-c8b204bfb897.png)

The immediate cause is the images are inside an element with a `markdown` class, which [is defined with](https://github.com/TerriaJS/terriajs/commit/f9faa7bfc892ad8db81edcc814729435be6a0a5c) `max-width: 100%`:

The popup message labels get that class [here](https://github.com/TerriaJS/terriajs/commit/f93730bd2e332cab59ff780b22a8a2f07056bd8f).

We should:
- work out if we really want all message popups styled as markdown (possibly we do, I'd definitely rather be writing copytext in markdown than HTML)
- decide where and when that 100% max-width should apply. What's the downside of just getting rid of it?

kring:
@chloeleichen can you tell us the purpose of this stlye:
https://github.com/TerriaJS/terriajs/blob/e2a53cb8624e332f8c1c04c87c7bf0a7444cb8be/lib/Styles/Core.less#L172

i.e. what bad effects might we see if we just remove it?

We can also fix this by simply changing the HTML for the Related Maps panel to specify a width for the `<td>`.  Yes, it's a table.  Yes, I know I'm a horrible person.

stevage:
I don't think it's that horrible to use `<table>` to display what is basically a table, to be honest. It's a table of related maps, and the two columns are image preview and description.

kring:
Yeah, it's totally really a table!  I'm not using a table for layout at all! :)

chloeleichen:
@kring `{max-width: 100%}` is for if you have a very big image it will automatically shrink to the size of the container, and to remain the aspect ratio you should have `height = auto`; 

However, that does not work in table because `tb` takes up the space it needs to display the content, and in this case either the `td` nor the `image` know what's the correct width so it makes the `td` very small, so the `td` should be given at least a `min-width`. 

Will create a fix for this

chloeleichen:
<img width="684" alt="screen shot 2015-11-20 at 11 55 52 pm" src="https://cloud.githubusercontent.com/assets/7508939/11300382/cfd91fe2-8fe2-11e5-91ce-1cb58049dcb9.png">

chloeleichen:
[...]

PR #1092, opened 2015-11-20
Fix image styles tables in popup window

chloeleichen:
Issue #1087

### DIFF
--- a/lib/Styles/Core.less
+++ b/lib/Styles/Core.less
@@ -225,7 +225,6 @@ button.block{
   h6 { font-size: .75rem }
 
   td {
-    min-width: 100px;
     h1, h2, h3, h4, h5, h6{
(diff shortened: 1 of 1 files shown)


lib/Styles/Core.less
```diff
@@ -223,6 +224,14 @@ button.block{
   h5 { font-size: .875rem }
   h6 { font-size: .75rem }
 
+  td {
+    min-width: 100px;
```
kring: This line makes me a little nervous.  There are legitimate reasons to have a narrow column, right?  If we leave this off, the "Related Maps" image will be really small..  but that's ok, because we've already added a `width` to that `<td>` to fix that anyway.  So are you ok with just getting rid of this `min-width`?

chloeleichen:
@kring removed min-width for td

PR #1095, opened 2015-11-24
Issue 988 proxy Url for legend

tephenavies:
@kring issue https://github.com/TerriaJS/terriajs/issues/988. Legend Urls are proxied in LegendSectionViewModel.show(), but that could be moved to the constructor if that would be a better place.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Change Log
 * `CsvCatalogItem` and `AbsIttCatalogItem` now expose a `concepts` property that can be used to adjust the display.
 * Added `Terria.cesiumBaseUrl` property.
 * The user interface container DOM element may now be provided to `TerriaViewer` by specifying `uiContainer` in its options.  Previously it always used an element named `ui`.
+* Legend URLs are now accessed via the proxy, if applicable.
 
 ### 1.0.48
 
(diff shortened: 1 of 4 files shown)


chloeleichen:
@steve9164 please sign the CLA

kring:
@steve9164 yeah that's not the right place to do it.  The constructor won't work either.  In both cases, the list of proxied URLs won't be updated if the list of legendUrls changes (which can and does happen when new information comes in asychronously).

Instead, use `knockout.defineProperty` in the constructor to define a proxiedLegendUrls property.  It will just return the `map` that you currently have in `show`.  The difference is that a knockout property keeps track of the fact that it depends on the `legendUrls` property, and will be recomputed when `legendUrls` changes.

tephenavies:
@kring Ok. I didn't realise that the legendUrls changed.

kring:
@steve9164 no worries.  TerriaJS is crazy: everything can change! :)

chloeleichen:
Oh this is my fault i told @steve9164 to put it there :(

kring:
@chloeleichen it was a reasonable suggestion, but using `knockout.defineProperty` is better.

tephenavies:
@kring Is that better? Should I do any caching of the proxiedLegendUrls, or is having them recalculated every time okay?

kring:
Yep, much better, thanks!  You don't need to cache because Knockout will do that automatically.

Two more things:
- Can you add a note to CHANGES.md?  Something like "Legend URLs are now accessed via the proxy, if applicable."
- Nitpick: please use spaces for indentation instead of tabs.  A lot of us use [Sublime Text](http://www.sublimetext.com/3), which will do that automatically based on the `.editorconfig` file in the root TerriaJS directory.

tephenavies:
I added the changelog entry, converted the tabs into spaces and fixed the tab setting on my editor.

kring:
Thanks @steve9164!

PR #1096, opened 2015-11-24
Fixes #462. Query using WMS DescribeLayer to obtain a WFS/WCS URL

### LINKED ISSUE #462: Query DescribeLayer to get the WFS or WCS associated with a WMS
See section 8.2 in the "OpenGIS Styled Layer Descriptor Profile of the Web Map Service Implementation Specification":
http://portal.opengeospatial.org/files/?artifact_id=22364

`DescribeLayer` provides a standard way to find the WFS or WCS URL associated with a given WMS (assuming that WMS supports this optional operation, of course).

We should use this, when it's available, to populate the "Data URL" section of the info popup.

tephenavies:
@kring Fixes #462. Queries WMS server using DescribeLayer to obtain a WFS/WCS Url if no dataUrl is present.


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -427,10 +427,43 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
...
         var that = this;
-        return loadXML(proxyCatalogItemUrl(this, this.metadataUrl)).then(function(xml) {
+		
+        var p1 = loadXML(proxyCatalogItemUrl(this, this.metadataUrl)).then(function(xml) {
             var metadata = capabilitiesXmlToJson(xml);
             that.updateFromCapabilities(metadata, false);
         });
+		
+		// Query WMS for wfs or wcs URL if no dataUrl is present
+		if (!defined(this.dataUrl)) {
```
kring: We should do this even if `!defined(this._rawMetadata)` (above) is false.  For example, click Add Data in the UI and enter this URL:
http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms?service=WMS&version=1.3.0&request=GetCapabilities

Then, click the Info button next to Ballarat Aquatic Facilities.  The WFS DataURL doesn't appear because DescribeLayer is not invoked because the catalog item already has metadata (provided by its group).

I'm not sure if I've explained that well, so let me know if it's unclear.

kring:
In the test case I described before (add `http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms?service=WMS&version=1.3.0&request=GetCapabilities`, click info) we're currently crashing because `that` is not defined.  This also causes an error to be displayed when we attempt to enable the layer.

kring:
This reminded me of another thing... we need to make sure that a `DescribeLayer` failure doesn't fail the load entirely.  The layer should still be usable even if `DescribeLayer` isn't implemented by the server or if it returns something crazy.  To do that, add an `otherwise` call after the `.then`.  It doesn't need to do anything, but a comment explaining why we're handling it would be great.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -425,13 +425,43 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
...
+        promises.push(loadXML(proxyCatalogItemUrl(this, this.metadataUrl)).then(function(xml) {
             var metadata = capabilitiesXmlToJson(xml);
             that.updateFromCapabilities(metadata, false);
-        });
+        }));
+    }
+        
+    // Query WMS for wfs or wcs URL if no dataUrl is present
+    if (!defined(this.dataUrl)) {
+        var describeLayersURL = cleanUrl(this.url) + '?service=WMS&version=1.1.1&request=DescribeLayer&Layers=' + encodeURIComponent(this.layers);
```
RacingTadpole: A few little questions on this line - Table 1 of the pdf document in #462 refers to version=1.3.0 and sld_version=1.1.1, saying both fields are mandatory. Also, can I suggest consistent capitalisation, ie. use "layers=" instead of "Layers="? (The spec says "LAYERS" but we use lower case everywhere else.)

RacingTadpole:
Looks good @steve9164! Do you have a use case you've been testing against, that I can try out? Also, the branch now has conflicts with master - could you merge with master and re-push?

RacingTadpole:
It would be great to add a test for this too, to test/Models/WebMapServiceCatalogItemSpec.js - if you're up for a challenge :-).

tephenavies:
@RacingTadpole I merged master into my branch and I didn't get any messages about conflicts. For the future, how can I find the conflicts with master without just running a merge and seeing if it breaks?

RacingTadpole:
Ah, the conflict was probably in CHANGES.md. There's a `.gitattributes` file in terriajs with the line `CHANGES.md merge=union`, which gives git some extra rules about merging. Our computers respect it but apparently github doesn't, so it reports extra merge conflicts.

RacingTadpole:
Another possibility is that you might not have git pulled your master?
Also - you can see a warning on the github page - look at the bottom of this page and you'll see "This branch has conflicts that must be resolved".  
If your local version merged fine, just push it back to github and we should be good.

tephenavies:
I can see the warning, but is there any way I can see the conflicts?

I only see:

> This branch has conflicts that must be resolved
> Only those with write access to this repository can merge pull requests.

RacingTadpole:
good question - not that I know of.

tephenavies:
Test: Add data item `http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms?service=WMS&version=1.3.0&request=GetCapabilities`, click on info and check Data URL (this is fetched using DescribeLayer).

RacingTadpole:
Should we use the same `version` (but not `sld_version`) number in the `DescribeLayer` request as in the original WMS request (ie. `1.3.0` in your test above)?

RacingTadpole:
Is this meant to happen:? 

When I look at the Ballarat aquatic centres, the old way sent me to a quite nice page at this URL: 

> http://www.data.gov.au/dataset/ballarat-aquatic-facilities 

and the new way sends me to an error message at this URL: 

> http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms?service=WFS&version=1.1.0&request=GetFeature&typeName=ballarat-aquatic-facilities:9039af7f_c53d_421a_97f5_f01f5457e171&srsName=EPSG%3A4326&maxFeatures=1000 

`java.lang.RuntimeException: Encode failed for _Feature. Cause: INVALID_CHARACTER_ERR: An invalid or illegal XML character is specified. Encode failed for _Feature. Cause: INVALID_CHARACTER_ERR: An invalid or illegal XML character is specified.`

tephenavies:
@RacingTadpole I think @kring told me that DescribeLayer is not available in versions above 1.1.1.

Also, the new way of finding a data URL doesn't overwrite the more pretty URL that might already be present, it only provides URLs for items that don't have a data URL.

How did you get that URL? I get http://data.gov.au:80/geoserver/ballarat-aquatic-facilities/wfs? when I add http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms as a data source.

tephenavies:
Here's an interesting find. From the [same server](http://www.data.gov.au/dataset/ballarat-aquatic-facilities):
- DescribeLayer is not mentioned in a [GetCapabilities query with WMS version 1.3.0](http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms?service=WMS&version=1.3.0&request=GetCapabilities)
- DescribeLayer is mentioned in a [GetCapabilities query with WMS version 1.1.1](http://data.gov.au/geoserver/ballarat-aquatic-facilities/wms?service=WMS&version=1.1.1&request=GetCapabilities)

kring:
Yep, OGC removed `DescribeLayer` in v1.3.0.  They were trying to make the spec more focused, I guess.  It's ok for us to use the older interface when it's available, though.

tephenavies:
Tests:
- WFS - Add data source http://ogi.state.ok.us/geoserver/wms and get info on entry `nrcs_lidar_2013`
- WCS - Add data source http://ogi.state.ok.us/geoserver/wms and get info on entry `NAIP2008`

Seems to have trouble with data.gov.au sources. I'm not sure why.

(That server has some cool test data)

kring:
This looks good!  Can you merge in master to fix the conflict?  I'm not sure what's going on with data.gov.au.  I _think_ their WFS server is broken, but we should confirm that and report it to them if it is.

PR #1098, opened 2015-11-25
Don't ignore csv `displayDuration`

RacingTadpole:
The main motivation for this PR is to fix this issue: when you first select a csv data item, it would show hollow circles because the dataVariable is ignored.
In the process I also fixed another issue, which is that `displayDuration` was being ignored, and the specs incorrectly tested it.


RacingTadpole:
Actually, I now can't reproduce the original problem on the master branch. But this PR should at least fix the other issue.

PR #1100, opened 2015-11-25
Abstract out a base Concept class

RacingTadpole:
I’m hoping this will start to make the csvCatalogItem easier to understand (and maybe more robust), and also help me add a chart data catalog item.


RacingTadpole:
You might find it easiest to look at the differences as of commit 532633f (if that's possible).

42d98b8 just renames `csvVariable` and `csvDataset` to `VariableConcept` and `DisplayVariablesConcept` respectively. 

I think that explains what they are better. It also reduces the scope for confusion with the DataTable, which is frequently referred to as a dataset - thanks to this line in `TableDataSource`: `this.dataset = new DataTable();`. But technically it's a non-backwards-compatible change.

kring:
@RacingTadpole can you merge in master?  There are conflicts with #1098.  I tried to resolve the conflicts but ended up with lots of test failures so I probably don't know what I'm doing.

RacingTadpole:
thanks, done.

PR #1105, opened 2015-11-25
Make displayVariablesConcept a computed property of DataTable

RacingTadpole:
Csv data currently has two parallel structures for storing its data variables:
- `TableDataSource` stores its variables, and selected variable, in a `DataTable` (called `this.dataset`).
- `CsvDataset` and `CsvVariable` (which #1100 rename to `DisplayVariablesConcept` and `VariableConcept`), which are exclusively used to display the variable names in the Now Viewing tab.

Since the latter are purely for UI, this PR makes them a knockout computed property of the former. That way it's clear which is the source of truth, and any changes in the DataTable are automatically reflected in the UI without needing to maintain them both.

Builds on #1100 .

One outstanding question - what should `csvCatalogItem.disableUserChanges` do? I've left it there, but it doesn't do anything.


RacingTadpole:
@kring - I've updated this PR for the latest master too.

PR #1109, opened 2015-11-27
Improve the message that is shown for a WebGL major performance caveat

### LINKED ISSUE #1058: Fix misleading "WebGL not supported" message
When WebGL _is_ supported, but we fail to create a context because `majorPerformanceCaveat` is true, we still put up a message box that says "your system does not support WebGL."  This is confusing, especially when the user then clicks on the WebGL link and it reports that WebGL _is_ available.  We should tweak the wording when WebGL has a major performance caveat so that it is not so misleading.

kring:
Previously, if we couldn't create a WebGL context to our liking, we would put a message saying "your web browser doesn't support WebGL."  This was potentially confusing because some users would then go to other sites (such as the [one linked in the message](http://get.webgl.org)!) and it would say, "Sure thing, boss, your browser supports WebGL just fine!"

The difference of opinion came down to a little thing called `failIfMajorPerformanceCaveat`.  This flag, provided when creating the WebGL context, says "if the WebGL performance is going to be terrible, like if it's going to use software rendering or something, don't give me a WebGL context at all."  We're using this flag, of course, because we have a nice fallback of using Leaflet.

So this pull request changes `TerriaViewer` to distinguish between "no WebGL at all" and "no WebGL but it's going to be slow".  When it's going to be slow, the message now says:

> Your web browser reports that performance with WebGL, a technology used for the 3D view in NationalMap, is likely to be very poor, so you will see a limited, 2D-only experience.  Consider trying a different web browser, upgrading your video drivers, or upgrading your operating system.  Alternatively, you may dismiss this message and continue to use NationalMap in 2D.

Fixes #1058


RacingTadpole:
I'm finding it a little hard to quickly understand your message... do you mind if I change it to this one, which is based on the other slow-performance message we show?

> Your web browser reports that it has performance issues displaying National Map in 3D, so you will see a limited, 2D-only experience. For the full 3D experience, please try using a different web browser, upgrading your video drivers, or upgrading your operating system.

PR #1113, opened 2015-11-30
Fixes #1107. Click propagation from animation controls and timeline disabled.

### LINKED ISSUE #1107: in 2D mode, interaction with time controls cause map actions
When in 2D mode, interation with the time controls cause map actions.
Doing a single-click on the "Play Faster" (or anywhere in the control panel) causes a "feature info" query on the map at that point.
Doing two quick clicks to go faster-faster causes the map to zoom in.
Simlarly when doing double-click anywhere on the time slider.

All is okay in 3D mode.

tephenavies:
@kring Fixes #1107. Also removed what looks like a hack to disable dragging the map while adjusting the slider on the timeline.


RacingTadpole:
This looks great Steve, thanks!

PR #1118, opened 2015-12-01
Fixed documentation for CsvDataset.js setSelected function

### LINKED ISSUE #1117: Documentation not correct for CsvDataset.js setSelected function
The [documentation above the setSelected function](https://github.com/TerriaJS/terriajs/blob/1991d9fb267fdc6d39663a554963753e757c364a/lib/Models/CsvDataset.js#L74) is incorrect as it has been copied from the toggleOpen function without modification.

tephenavies:
Fixes #1117.


RacingTadpole:
Nice catch.

PR #1124, opened 2015-12-03
Generalise display variables to allow more than one to be selected (optionally)

RacingTadpole:
This is preparing the ground for charts, where we want to be able to plot multiple y-values from one file. 

It shouldn't change any behaviour for `csvCatalogItem` - but allows for a `allowMultiple` parameter to be passed to the DataTable's constructor.


kring:
@RacingTadpole is this intended to be a PR into `computed_variable_concept` instead of master?

PR #1128, opened 2015-12-03
Trigger a resize event when explorer panel is open or closed

RacingTadpole:
This should have no effect on anything yet. However, in future - eg. when we may have a chart panel which needs to know when to resize itself - it will be helpful to know when the explorer panel opens or closes.

There is a discussion of how to do this on Stack Overflow [here](http://stackoverflow.com/questions/1818474/how-to-trigger-the-window-resize-event-in-javascript).

I have yet to test this across browsers, but it works on the latest Chrome, Firefox and Safari, and the above discussion suggests it will work on IE.

Triggering the resize twice is ugly, but it gives the UI a chance to clean up after itself if the animation is running a bit slow.

Very happy if there's a better way to do this.


kring:
There's already similar logic to this in `ExplorerPanelViewModel.js`.  Should we either 1) generalize that to work for charts, too (e.g. make `mapElementToDisplace` something like `elementsToDisplace` instead), or 2) use this new approach there, too?

RacingTadpole:
Hmm, yes. Funnily enough this PR only changes `ExplorerPanelViewModel.js`, and I missed that it already had an approach to deal with something similar.

The existing approach is this: When you create the explorer panel, you pass `option.mapElementToDisplace = 'cesiumContainer'` or equivalent. If defined, then a 'map-displaced' class is put on that element only when the explorer panel is open. But also, when isOpen is toggled, leaflet receives a separate invalidateSize() call, and notifyRepaintRequired() is called.

So ExplorerPanel includes knowledge of what calls are needed to update leaflet and Cesium when they are resized, and some extra css is required.

I like my suggested approach better conceptually, in which ExplorerPanel only does "triggerResize" and lets anything that cares about that deal with it itself. As an added bonus, if we ever have another panel that also needs to change the map size (perhaps a chart panel?), we don't have to repeat the logic again.

What do you think?

kring:
I'm convinced!

RacingTadpole:
OK @kring, I've left the "`map-displaced`" class being added, so that you can control animations nicely with css. But the trigger resize in itself already does the leaflet `invalidateSize()` call and the `notifyRepaintRequired()`, as far as I can tell, so we can rationalise the code a bit.

Can I ask you to check this in IE9 though? I couldn't get a working version of IE9 going.

PR #1129, opened 2015-12-03
Add custom components framework, <collapsible>

RacingTadpole:
This is the framework into which I will be introducing the `<chart>` tag as a separate PR.

I'm submitting this as a PR now to see what you think about it. Is this a reasonable way to go about it?

To test it out, you can use the `662_custom_cpts` branch of national map.

### DIFF
--- a/lib/Models/customComponents.js
+++ b/lib/Models/customComponents.js
@@ -1,0 +1,127 @@
+"use strict";
+
+/*global require*/
+var AssociativeArray = require('terriajs-cesium/Source/Core/AssociativeArray');
+var defined = require('terriajs-cesium/Source/Core/defined');
+
+var CustomComponentElement = require('./CustomComponentElement');
+
+var types = new AssociativeArray();
+var elements = [];
+
+/**
+ * A store of registered custom component types, eg. <chart>,
+ * and elements of custom components.
+ */
+var customComponents = {};
+
+customComponents.register = function(customComponentType) {
+    types.set(customComponentType.name, customComponentType);
+};
+
+customComponents.isCustomComponent = function(componentName) {
+    return (types.contains(componentName.toLowerCase()));
+};
+
+
+customComponents.resetElements = function() {
+    elements = [];
+};
+
+customComponents.removeElements = function(owner) {
+    for (var i = elements.length - 1; i >= 0; i--) {
+        var thisElement = elements[i];
+        if (thisElement.owner === owner) {
+            elements.splice(i, 1);  // remove it from the array
+        }
+    }
+
+};
+
+customComponents.addElement = function(element) {
+    elements.push(element);
+};
+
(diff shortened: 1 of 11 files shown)


lib/ViewModels/CollapsibleViewModel.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+CollapsibleViewModel.prototype.toggleOpen = function() {
+    this.isOpen = !this.isOpen;
+};
+
+CollapsibleViewModel.prototype.renderCollapsible = function() {
+    // this is not the ideal way to find the element with class .chart-inject
+    // the first child is the <script> tag, etc
+    var collapsibleContainer = this.container.children[1].children[1].children[0];
```
kring, line 60: why not `var collapsibleContainer = this.container.getElementsByClassName('chart-inject')[0];`?

lib/ViewModels/CollapsibleViewModel.js
```diff
@@ -0,0 +1,70 @@
...
+    // move back existing content
+    var nextNode;
+    for (var node = this.wrapper.firstChild; node; node = nextNode) {
```
kring, line 63: Might as well just loop over `this.wrapper.children` right?

lib/Models/customComponents.js
```diff
@@ -0,0 +1,127 @@
...
+var CustomComponentElement = require('./CustomComponentElement');
+
+var types = new AssociativeArray();
+var elements = [];
+
+/**
+ * A store of registered custom component types, eg. <chart>,
+ * and elements of custom components.
+ */
+var customComponents = {};
```
kring, line 16: Since this is a namespace (or maybe we could think of it as a static class), it should be PascalCased.

lib/Models/registerCustomComponentTypes.js
```diff
@@ -0,0 +1,28 @@
...
+ * 
+ * @param  {Object} [options] Options.
+ * @param  {ChartPanelViewModel} [options.chartPanelViewModel] The chartPanelViewModel, if available.
+ */
+var registerCustomComponentTypes = function(options) {
+
+    var collapsibleComponentType = new CustomComponentType({
+        name: 'collapsible',
+        render: function(element) {
+            var collapsibleViewModel = new CollapsibleViewModel(element.attributes.name, element.attributes.open);
```
kring, line 20: This logic will be pretty much the same for all components, right?  Wouldn't something like this be sufficient?

```
customComponents.register('collapsible', CollapsibleViewModel);
```

The render would always just pass `element.attributes` to the constructor.

kring:
@RacingTadpole I think we discussed this offline, and thought a different approach (a custom Knockout binding if I remember correctly) might be better?  Ok to close for now?

PR #1134, opened 2015-12-04
map height

chloeleichen:
@RacingTadpole , the `var cContainer = document.getElementById('cesiumContainer');` bit probably could be done better? let me know if you need me to change anything


lib/ViewModels/ChartPanelViewModel.js
```diff
@@ -12,6 +12,9 @@ var combineData = require('../Core/combineData');
 var loadView = require('../Core/loadView');
 var removeView = require('../Core/removeView');
 var ChartViewModel = require('./ChartViewModel');
+var cContainer = document.getElementById('cesiumContainer');
```
RacingTadpole: Hi, thanks for this Chloe - looks good. Yes, is there a way to get the cesium container direct from the terria object?

lib/ViewModels/ChartPanelViewModel.js
```diff
@@ -90,7 +102,8 @@ var ChartPanelViewModel = function(options) {
             return [
                 new ChartViewModel(this, this, combineData(dataPairArrays), {
                     showRangeSelector: true,
-                    canDownload: true
+                    canDownload: true,
+                    height: this.height - 60
```
RacingTadpole: Also is there a way to not hardcode this 60? Eg. if it's all due to the chart panel title bar, let's define that height somewhere and use that.

lib/ViewModels/ChartPanelViewModel.js
chloeleichen: I think so, but i dunno how, @kring must know :dart:

kring: `Terria` knows the `currentViewer`, which surely has a reference to its DOM element somehow, but it's not really exposed in a useful way.  We could probably expose it.

RacingTadpole:
@chloeleichen - can I put this back on your radar? Just a few little things above. Thanks!

chloeleichen:
@RacingTadpole done :smile:

lib/ViewModels/ChartPanelViewModel.js
```diff
@@ -49,6 +57,16 @@ var ChartPanelViewModel = function(options) {
 
     updateMaxDimensions();
 
+    knockout.getObservable(this, 'isVisible').subscribe(function(isVisible) {
+        var cesiumContainer = this.terria.currentViewer.viewer.container;
```
RacingTadpole, line 61: Thanks @chloeleichen! Unfortunately this doesn't work with Leaflet... (it doesn't have a `viewer` property) - any ideas?

chloeleichen, line 61: `document.getElementById()?`

RacingTadpole, line 61: Ha! That's what you removed in the last commit! I've just poked around in terriajs's Leaflet class and found this:
`map.getContainer()`, ie. we can use `this.terria.currentViewer.map.getContainer()` here - this returns the DOM element we need.

chloeleichen, line 61: sounds cool, do you want me to do that or you've already done it?

RacingTadpole, line 61: I can do it... actually I'm thinking that we should really define a `getContainer` method on `currentViewer` directly, and then use that.

PR #1141, opened 2015-12-08
#902: Active Tab In Share Url

### LINKED ISSUE #902: Include active 'tab' state in share URL
When using the Share feature to get an iframe for inclusion on other people web pages, users are requesting that we switch to the NowViewing tab automatically.

Since we have existing behaviour we might want to just make it possible to track the state of the tabs, so that we can view the state as it was when share was clicked.

AlexGilleran:
Fixes #902.

Preserves the currently opened tab when sharing a link.

I've moved the current tab index from the explorer viewmodel to the terria model so that the Share popup viewmodel has access to it, and have put it in another initSource that's read when the app is reinitialized.

Is pulling that kind of ui logic into the model OK?

Also not sure about just persisting the index - would it be more robust if each tab had a unique id?


kring:
Thanks @AlexGilleran!

> Is pulling that kind of ui logic into the model OK?

It's certainly not ideal for the model layer to have this sort of knowledge of the state of the UI.  One way to avoid this is with a bit of indirection.  `Terria` already has a `userProperties` property that is used to get information from the URL to various parts of Terria (often the UI) that care.  For example, if you put `#hideExplorerPanel=1` in the URL, Terria sets `userProperties.hideExplorerPanel` to `1` without necessarily knowing what that actually means.  Then other parts of Terria (an application's `index.js` in this case) can look for that `hideExplorerPanel` property and act accordingly.

You could use something like that here.  But you'll also need to serialize this user property for sharing, which is not currently done (although it would be a good change if `hideExplorerPanel` were serialized for sharing too!).  I'm not sure if it makes sense to serialize _all_ `userProperties` for sharing, or if we need some way to specify which ones are meant to persist after sharing.

All that said, it's not the end of the world for `Terria` to just track this bit of UI state if the alternative is overly complicated.

> Also not sure about just persisting the index - would it be more robust if each tab had a unique id?

Yeah persisting the index isn't great.  Using the name of the tab might be a bit better, though of course is causes problems if we ever want to change the name of a tab.  I like the idea of adding an `id`.  Maybe give it a default for each of our current tab types to preserve backward compatibility.

AlexGilleran:
- Active tab state now moved to `terria.userProperties`
- Share view now puts`terria.userProperties` into share urls
- Active tab state now tracked by a string `id` instead of index
- Existing tabs now have an `id`, if no id is specified it falls back to the name, if no name is specified it falls back to "unknown"
- If tab ids collide, they're resolved by modifying whatever the second tab with that id was so its id is now `tab.id += 'i'`. So if all your tabs are called `id` you'll get "id", "idi" "idii" etc.

I couldn't think of a reason we wouldn't want to put user properties into the share URL (they're in the app URL anyway) so I just persisted all of them, can add a whitelist if necessary but I think it's a lot simpler for users if whatever user properties they have just automatically get shared?

kring:
Thanks @AlexGilleran, this is looking good!  I see two problems though:
1. `http://localhost:3001/#hideExplorerPanel=1` no longer works in this branch (but http://nationalmap.gov.au/#hideExplorerPanel=1 does).  I'm not completely sure it was this change that broke it, though.
2. Because changing the UI doesn't update the URL or the `userProperties`, it's easy to get into a bad state.  For example: visit `http://localhost:3001/#map=2d`.  You'll get 2D, as intended.  Then, use the Base Maps panel to switch to 3D.  Click Share and copy the URL into a separate window.  The share URL will contain `map=2D` and you'll get 2D, even though 3D is expected.

AlexGilleran:
Drat. OK. As far as I can see `hideExplorerPanel` is broken even without this commit, so I'll raise another bug for that.

Given that `userProperties` is supposed to be used by users also, I gather that we can't actually resolve this by making sure that all changes to app state that `userProperties` tracks (like the 2D/3D thing you described) also re-sync the `userProperties` object itself. 

So does a whitelist for `userProperties` keys that should be included in the URL make sense to you? I'm thinking by default it includes `hideExplorerPanel` and `activeTabId` but can be added to by the user when they create the share view?

EDIT: Actually maybe this did break hideExplorerPanel, I just wasn't fully refreshing when I tested it.

kring:
Yeah, a whitelist makes sense to me.  We'll need to fix the `hideExplorerPanel` bug before the release tomorrow because I know of at least one group that relies on it working.

AlexGilleran:
You probably missed my sneaky edit, but looks like `hideExplorerPanel` was this commit after all.

AlexGilleran:
Oops, yeah that was me, left out the second argument in 

```
this.activateTab(this.tabs[0], false);
```

AlexGilleran:
- Fixed hideExplorerPanel = 1
- Added an override-able whitelist for userProperties ending up in the share URL with `activeTabId` and `hideExplorerPanel` as defaults.
- Made the ExplorerPanel keep updating the hideExplorerPanel user property so it doesn't end up out of sync.

lib/ViewModels/SharePopupViewModel.js
```diff
@@ -12,9 +12,20 @@ var inherit = require('../Core/inherit');
...
+ * @param options {Object}
+ * @param options.terria {Terria} the terria instance to get data from
+ * @param options.userPropWhiteList {Object} A white list of userProperty ids that will be reflected in the share link.
+ *          Defaults to SharePopupViewModel#defaultUserPropWhiteList. Override this if you have custom userProperties
+ *          that you want to be shared along with the rest of the map.
+ */
 var SharePopupViewModel = function(options) {
     PopupViewModel.call(this, options);
     this.terria = options.terria;
+    this.userPropWhiteList = options.userPropWhiteList || SharePopupViewModel.defaultUserPropWhiteList;
```
kring: We usually use `defaultValue(options.usePropWhiteList, SharePopupViewModel.defaultUserPropWhiteList)` (unless the second parameter is expensive).

lib/ViewModels/SharePopupViewModel.js
```diff
@@ -87,30 +98,39 @@ var SharePopupViewModel = function(options) {
...
         baseMapName: this.terria.baseMap.name,
-        viewerMode: this.terria.leaflet ? '2d' : '3d'
+        viewerMode: this.terria.leaflet ? '2d': '3d'
     });
 
+    // Add user properties if applicable
+    var userPropertiesQuery = '';
+    this.userPropWhiteList.map(function(key) {
+        var val = this.terria.userProperties[key];
+        if (typeof val !== 'undefined') {
```
kring: Use `defined(val)`.

kring:
This looks great, only those nitpicky things left.

AlexGilleran:
Done :)

PR #1144, opened 2015-12-08
Sort ABS data variables numerically. Fixes #1143.

### LINKED ISSUE #1143: ABS DataVariable sort order is unfriendly
screenshot 2015-12-08 14 02 01

Not sure if we can fix this in the general case, or just have to apply specific sorting for these groups. It would be fairly easy to extract the first number with a regex, convert to integer, then use that as key.

stevage:


lib/Models/AbsIttCatalogItem.js
```diff
@@ -25,7 +25,8 @@ var proxyCatalogItemUrl = require('./proxyCatalogItemUrl');
 var AbsDataset = require('./AbsDataset');
 var AbsConcept = require('./AbsConcept');
 var AbsCode = require('./AbsCode');
-
+var naturalSort = require('javascript-natural-sort');
+naturalSort.insensitive = true;
```
kring: You need to do this right before it is used.  Otherwise, someone may have switched it back to case-sensitive sorting in the meantime.

kring:

PR #1149, opened 2015-12-09
Make displayVariablesConcept a computed property of DataTable (in master)

RacingTadpole:
This is #1105 but merging into master, instead of the `concept_class` branch.


RacingTadpole:
@kring, I've fixed the problem with dataVariable now for the `computed_variable_concept` PR too.

PR #1150, opened 2015-12-09
Generalise display variables to allow more than one to be selected (optionally)

RacingTadpole:
Same as #1124 , but merging into master.


RacingTadpole:
@kring, I've fixed the problem with dataVariable now, and added a test for it. Thanks for catching it!

lib/Map/DataTable.js
```diff
@@ -288,26 +332,71 @@ DataTable.prototype.getCsvTable = function () {
...
+DataTable.prototype.setDataVariable = function (varName, makeActive) {
+    // Note we take care to ensure selectedNames is a new object so it triggers knockout's tracking.
+    if (defined(this.variables[varName])) {
+        if (this.allowMultiple) {
+            // If it's not in the list, add it; if it is, remove it.
+            var index = this.selectedNames.indexOf(varName);
+            if (index === -1) {
+                this.selectedNames = this.selectedNames.concat([varName]);  // use concat not push to clone the array.
+            } else {
+                this.selectedNames.splice(index, 1);
```
stevage, line 350: This changes the semantics of "setDataVariable" to be "toggle data variable". Is this required for some reason? It doesn't seem right. Also this part of the decision tree is ignoring `makeActive`...

lib/Map/DataTable.js
```diff
@@ -288,26 +332,71 @@ DataTable.prototype.getCsvTable = function () {
...
+        } else {
+            // If this don't allow multiple selections, just reset the list to the new selection.
+            if (makeActive === false) {
+                this.selectedNames = [];
+            } else {
+                this.selectedNames = [varName];
+            }
+        }
+    } else {
+        this.selectedNames = [];
```
stevage, line 362: Maybe this should raise an exception instead of quietly failing. Setting a non-existent variable is pretty wrong.

lib/Map/DataTable.js
```diff
@@ -288,26 +332,71 @@ DataTable.prototype.getCsvTable = function () {
...
+    }
+};
+
+/**
+ * Set the current data variable to a reasonable default. Return the data variable.
+ * @returns {DataVariable} The chosen data variable.
+ */
+DataTable.prototype.setDefaultDataVariable = function() {
```
stevage, line 370: Should this be `assignDefaultDataVariable` or `chooseDefaultDataVariable`? Convention has `setX` takes an argument, no?

Also we should make this choice smarter - picking the leftmost column often means we choose a name field, which works out badly.

lib/Map/DataTable.js
```diff
@@ -288,26 +332,71 @@ DataTable.prototype.getCsvTable = function () {
...
+* @exception {DeveloperError} If more than one variable is selected (which is possible if allowMultiple = true).
 */
 DataTable.prototype.getDataVariable = function () {
-    return this.selected.data;
+    if (this.selectedNames.length === 0) {
+        return;
+    } else if (this.selectedNames.length === 1) {
+        return this.selectedNames[0];
+    }
+    throw new DeveloperError('DataTable.getDataVariable called with more than one variable selected.');
```
stevage, line 390: Would returning an array be better than throwing an error? Or maybe just returning the first variable, if we're saying that it has a special status.

lib/Map/DataTable.js
```diff
@@ -288,26 +332,71 @@ DataTable.prototype.getCsvTable = function () {
...
 * Get the current variable name.
 *
-* @returns {String} The current variable name
-*
+* @returns {String} The current variable name.
+* @exception {DeveloperError} If more than one variable is selected (which is possible if allowMultiple = true).
 */
 DataTable.prototype.getDataVariable = function () {
-    return this.selected.data;
+    if (this.selectedNames.length === 0) {
```
stevage, line 385: Incidentally, wouldn't it be nice to change the pattern of using variable names everywhere to instead be passing references to the actual `DataVariable` objects? It seems a bit backward somehow.

stevage:
Wow, a couple of epic pull requests here :) Still trying to get my head around it. Would you mind spelling out the use case for multiple selected variables for a single dataset? And is it assumed that the first selected variable in the list is the "main" one?

RacingTadpole:
Hi Steve - thanks for taking a look at this. To answer your last point first, this is all groundwork for charts. There will be a new catalog item for a time-series (or other chartable data) file. This will list the "variables" on the Legends tab, where you can pick which ones are shown on the chart - so you can see what I mean, I'll send you a link to my prototype separately.  I settled on the idea of reusing `DataTable` for the chart data, but if you have other ideas please let me know! In particular this may be the time to refactor the entire csvCatalogItem approach...?

RacingTadpole:
Your comments raise some excellent points. I've tried to walk a fine line between maintaining backwards compatibility and adding new functionality, and I think I've made it more confusing as a result. Further, as you point out, lots of these functions are named like they return or set instances of classes, but in fact deal with their names; and I've wound up half adopting the alternative paradigm, eg. adding `setDefaultDataVariable` which sets and returns the data variable, not the name. So... you've convinced me that I should either: a) give up backwards compatibility, rename everything and refactor as needed (eg. so they return objects not strings); or, b) essentially do the same thing, but don't call it DataTable, and leave the old DataTable around for backwards compatibility.

stevage:
Yes, I think we'd all love to refactor DataTable/DataVariable/TableDataSource/CsvCatalogItem/etc some more! Despite spending quite a bit of time refactoring CsvCatalogItem a while ago, I still don't have a particularly clear idea what concepts DataTable and TableDataSource are supposed to encapsulate.  (In particular, I don't really get why DataTable has a "selected variable" concept - it feels much too low down. A table doesn't know which variable is "selected", a GUI does though...)

I don't really like the idea of leaving some class around for "backward compatibility" unless we have a clear idea who depends on it and why.

kring:
@RacingTadpole should we close this?  It sounds like it might be superseded by your new CSV work.  Just trying to clean up old PRs!

RacingTadpole:
Closing this PR as it will need to be completely revised following #1224.

PR #1153, opened 2015-12-10
Added extra tests for searching catalog items

tephenavies:
Added tests for changes made in #1138 as well as tests for searching by catalog item type and url.


kring:
This is great, thanks @steve9164!

PR #1163, opened 2015-12-14
Fix HTTPS-related problems

kring:
The WFS request used in region mapping wasn't set up to ever go through the proxy.  Now it will do so when necessary.

Also, `corsProxy.shouldUseProxy` now returns true whenever the URL's protocol is `http` and the page's protocol is `https` (assuming we're allowed to proxy for the host at all).  This avoids a mixed-content warning.


RacingTadpole:
I can confirm that this loads National Map's ABS age census data WFS query using the proxy, where it didn't before.
However I haven't tested whether the corsProxy https change works. Is there a data source which does not use the proxy and has http? I will hardwire the pageIsHttps to check it works.

kring:
There's not, but `data.gov.au` supports CORS, so you could try adding it to the `corsDomains` list as a test.

RacingTadpole:
OK, I can confirm that once added to the `corsDomains`, data.gov.au does not use the proxy on localhost; and then when `pageIsHttps` is set to true, it does use the proxy.

PR #1164, opened 2015-12-14
Legend image background

meh9:
Wrap legend images in a div and give the image a white background so that we can see transparent legends.


kring:
Looks good, thanks!
FYI I fixed a problem with wide images (e.g. `Land -> Catchment Scale Land Use 2015`).

PR #1165, opened 2015-12-15
Add Progress Bar

### LINKED ISSUE #539: Add a "loading/working..." indicator
Some layers can take a while to load, it would be very nice to have a "loading..." spinner in a corner somewhere (probably best not showing it in the middle of the screen) showing that things are still happening. 

This would be particularly nice for WFS layers where we don't even get individual tiles loading showing there is progress, there is just nothing for a long period until everything is drawn all at once.

AlexGilleran:
Fixes #539.

Adds a `MapProgressBar` that can be added to the `cesiumContainer` in order to display a simple progressive loading bar as per the design in #539, that reflects tile loads in both Cesium and Leaflet. The `MapProgressBar` needs to be specifically added by the app that uses Terria.

As part of the infrastructure for the progress bar, adds a `terria.tileLoadProgressEvent` that raises events with the current number of tiles to be loading and the maximum number loading, so that other incremental progress bars can be plugged in.


AlexGilleran:
This can be merged now that Terria is using a version of Cesium with the commits this depends on merged in.

PR #1174, opened 2015-12-15
Translation test task using

jeanpul:
This is an example of how I actually manage translation
You need  gulp-gettext-static-tags (https://github.com/Innovisite/gulp-gettext-static-tags), 
please get the latest release from github and use npm link


kring:
Hi @jeanpul, this approach looks reasonable enough to me.  It's a little surprising to me that you had to develop your own gulp plugin for facilitating the translation though.  I would have assumed there are some fairly standard tools for doing localization.

In any case, since I think you meant this as an example (not something you'd like us to merge), I'm going to close the PR.  Thanks for putting together the example!

PR #1176, opened 2015-12-16
Revert html2canvas to 0.4.1

rsignell-usgs:
Since 0.5.0-alpha is broken, revert to 0.4.1.   I realize this PR might not be the right approach, but if it is, great!

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "terriajs",
-  "version": "1.0.50",
+  "version": "1.0.51",
   "description": "Geospatial data visualization platform.",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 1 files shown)


kring:
`0.5.0-alpha2` is ok.  You were just seeing the problem in https://github.com/TerriaJS/map because it uses an older version of TerriaJS which specifies `^0.5.0-alpha2`.  That carat means "anything with the same major version number greater than or equal to 0.5.0-alpha2".  Thus you get `0.5.0-beta3`, which is broken.

Short version: TerriaJS doesn't need to be changed; we only need to upgrade TerriaJS/map to use TerriaJS 1.0.50.

rsignell-usgs:
Awesome.  Thanks for the explanation!

kring:
I'm updating TerriaJS/map right now, I'll push it in a moment.

kring:
The `master` and `usgs` branches are both updated now (but not extremely well tested, so let me know if you run into any problems).

PR #1180, opened 2015-12-17
Merge colormap work

RacingTadpole:
Adds all of Steve's color mapping work into master, with passing tests.


stevage:
Issues raised by @kring:
- [x] "more postcode testing" test shows the value `2` twice
- [x] "more postcode testing" test shows the value `42.42` as `42`
- [x] "no color bins (linear scale)" has some pixel alignment issues

PR #1181, opened 2015-12-17
Made WebMapServiceCatalogItem get legend url from first Style tag

### LINKED ISSUE #699: Use the WMS LegendURL to show the legend
Currently when showing a WMS layer Terria creates its own URL to get the WMS legend instead of using the one specified in the WMS GetCapabilities. This can be a problem when a service specifies a custom legend to use, as we then do not show it. See for example:

http://www.ga.gov.au/gis/services/topography/National_Electricity_Transmission_Lines/MapServer/WMSServer?request=GetCapabilities&service=WMS

```
<LegendURL width="173" height="377">
<Format>image/png</Format>
<OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="http://www.ga.gov.au/gis/server/capabilities/National_Electricity_Transmission_Lines.png" xlink:type="simple"/>
</LegendURL>
```

AlexGilleran:
If any style tag is present in the Capabilities of a WMS layer with a LegendURL, this will be used by default for the legend image. If multiple <Style>s or <LegendURL>s are present, the first one will be used. If none are present, the legend URL will revert back to a default constructed url based on the WMS standard, as it currently does now.

As far as I can understand this _should_ fix #699, although the original use case in that issue is no longer valid because that dataset uses ArcGis now and it has a custom legend url set.

I did test against the distance from powerlines dataset in AREMI, which uses WMS, and it works with that - although the legend image is the same despite now coming from a different URL.

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -14,6 +14,7 @@ var when = require('terriajs-cesium/Source/ThirdParty/when');
 
 var arraysAreEqual = require('../Core/arraysAreEqual');
 var Metadata = require('./Metadata');
+var LegendUrl = require('./LegendUrl');
 var CatalogMember = require('./CatalogMember');
 var inherit = require('../Core/inherit');
 var raiseErrorOnRejectedPromise = require('./raiseErrorOnRejectedPromise');
@@ -56,7 +57,7 @@ var CatalogItem = function(terria) {
     /**
      * Gets or sets the URL of the legend for this data item, or undefined if this data item does not have a legend.
      * This property is observable.
-     * @type {String}
+     * @type {LegendUrl}
      */
     this.legendUrl = undefined;
 
@@ -168,7 +169,7 @@ var CatalogItem = function(terria) {
     knockout.defineProperty(this, 'legendUrls', {
         get: function() {
             if (!defined(this._legendUrls) || this._legendUrls.length === 0) {
-                if (defined(this.legendUrl) && this.legendUrl.length > 0) {
+                if (defined(this.legendUrl) && defined(this.legendUrl.url) && this.legendUrl.url.length > 0) {
                     return [this.legendUrl];
                 }
             }
@@ -235,8 +236,6 @@ var CatalogItem = function(terria) {
 
 inherit(CatalogMember, CatalogItem);
 
-var imageUrlRegex = /[.\/](png|jpg|jpeg|gif)/i;
-
 defineProperties(CatalogItem.prototype, {
     /**
      * Gets a value indicating whether this data item, when enabled, can be reordered with respect to other data items.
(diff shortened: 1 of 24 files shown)


meh9:
Testing this with the Bureau Solar Climatology services, which have two styles with legends specified, the legend is replaced with a `Open legend in a separate tab` message. When clicked it does show the correct legend image. I don't think that is the behaviour we want. This is one of the services:

http://neii.bom.gov.au/services/solarclim/wms/data/monClim_gloHorExp1Day.nc?service=WMS&version=1.3.0&request=GetCapabilities

AlexGilleran:
I've made the way that the legend determines whether to display a link or an image a bit more intelligent. Previously it just looked at the extension in the link, but I've now changed legends from `string`s to a new `LegendUrl` class that incorporates both a url and a mimeType - this allows the mime type in WMS XML to be read in and used by the `LegendSectionViewModel` to determine whether or not it describes an image.

I also found I was continually tripping over way various `CatalogItem`s use `legendUrl` and `legendUrls`, so I made everything use a consistent `legendUrls` property that's an array of `LegendUrl` objects. This can still be read in from a single `legendUrl` string in a json file through a custom updater.

lib/Models/CatalogItem.js
```diff
@@ -56,9 +58,9 @@ var CatalogItem = function(terria) {
     /**
      * Gets or sets the URL of the legend for this data item, or undefined if this data item does not have a legend.
      * This property is observable.
-     * @type {String}
+     * @type {Array<LegendUrl>}
```
kring: Just FYI, you can also do `@type {LegendUrl[]}`

lib/Models/CatalogItem.js
```diff
@@ -56,9 +58,9 @@ var CatalogItem = function(terria) {
...
      */
-    this.legendUrl = undefined;
+    this._legendUrls = undefined;
```
kring: Now that this is a private property, the doc shouldn't go here, because the doc ends up looking like this:

image

Document the public property instead.

lib/Models/CatalogItem.js
```diff
@@ -212,18 +200,27 @@ var CatalogItem = function(terria) {
...
                 return 'direct';
             }
         },
-        set : function(value) {
+        set: function(value) {
             this._dataUrlType = value;
         }
     });
 
+    knockout.defineProperty(this, 'legendUrls', {
```
kring: Not much reason to use `knockout.defineProperty` for a property like this that just delegates to a private field.  Just make a public field instead.

lib/Models/CatalogItem.js
```diff
@@ -367,19 +347,35 @@ CatalogItem.defaultUpdaters.rectangle = function(catalogItem, json, propertyName
...
+CatalogItem.defaultUpdaters.legendUrl = function(catalogItem, json, prototypeName) {
+    if (json.legendUrl) {
+        catalogItem.legendUrls = [new LegendUrl(json.legendUrl)];
+    }
+};
+
+CatalogItem.defaultUpdaters.legendUrls = function(catalogItem, json, prototypeName) {
+    if (defined(json.legendUrls)) {
+        catalogItem.legendUrls = json.legendUrls && json.legendUrls.map(function(rawLegend) {
+            return new LegendUrl(rawLegend.url, rawLegend._mimeType);
```
kring: Why is mimeType named with an underscore?  It's not actually private is it?

AlexGilleran: Yeah - the logic for deciding whether a certain mimeType is actually an image lives in `LegendUrl.prototype.isImage`.

AlexGilleran: ... but that breaks serialization so I'll just make it public :)

AlexGilleran:
As per discussion with @kring, legendUrl/s change has been removed.

Also just did a full amend and got rid of all the formatting cruft so it's all my changes now.

PR #1182, opened 2015-12-17
Replace . and # with _ in property keys for feature info template

### LINKED ISSUE #1169: Allow feature info templates to reference . in property names
We have a key in the feature info that ends with a full-stop, and cannot retrieve the data using the key

`<td>{{Physical Unit No.}}</td>`

To quote from http://stackoverflow.com/questions/14901683/how-do-i-reference-a-field-name-that-contains-a-dot-in-mustache-template, "You can't read a key with a . in it from Mustache."

One suggestion would be to search and replace . (and #?) in property names. Replace with _?

RacingTadpole:
Fixes #1169 .


lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -92,11 +92,25 @@ FeatureInfoPanelSectionViewModel.prototype.toggleOpen = function() {
     this.terria.currentViewer.notifyRepaintRequired();
 };
 
+function propertyValues(properties, clock) {
+    // Replace '.' and '#' in property keys with _, since Mustache cannot reference keys with these characters.
+    // In future, we will check each property for a getValue function here too; if it exists, use it to get the current value.
+    var badCharacters = new RegExp('[.#]', 'g');
```
kring: I think It's a little better to use `/[.#]/g` instead of an explicit `RegExp` object.

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -92,11 +92,25 @@ FeatureInfoPanelSectionViewModel.prototype.toggleOpen = function() {
...
+    var result = {};
+    for (var key in properties) {
+        if (properties.hasOwnProperty(key)) {
+            var cleanKey = key.replace(badCharacters, '_');
+            result[cleanKey] = properties[key];
```
kring: Should this be recursive?

RacingTadpole: yeah, good call.

RacingTadpole:
ok, done thanks @kring! (And thanks for updating CHANGES!)

Issue #1021, opened 2015-11-02
Opacity slider won't slide in Firefox


kring:
Clicking and dragging the opacity slider on the Now Viewing tab does not work in Firefox - the slider won't move.  However, clicking a spot _will_ move the slider there successfully, it's only the dragging that doesn't work.

Dragging works correctly in Chrome, IE10, IE11, and Edge, so I'm not sure where we're going wrong in Firefox.

kring:
@chloeleichen can you take a look at this one?

chloeleichen:
@kring : This is because the horizontal drag in firefox conflicts with the horizontal swipe/pan by hammer.js, which is used to implement swipe detection (`swipeLeft: close, swipeRight: open`) in ExplorerPanel.html. If we remove the swipe, the opacity slider will work in Firefox. 

This issue has been reported to hammer.js here: 
https://github.com/hammerjs/hammer.js/issues/654

I guess it depends on what do you think is more important: opacity able to slide in firefox, or the panel able to be swiped. Please let me know if you want me to remove the swipe

kring:
Thanks for investigating this, @chloeleichen!  Sorry I'm just getting back to it now.
What we want, I suppose, is for hammer to respond to swipes only outside the opacity slider.  "Swiping" inside the opacity slider should just slide the slider.  Any idea if that's possible somehow?

chloeleichen:
@kring i think to do that we'll need to change `hammer.js` and check if event target is `<input type='range'>` then do not apply its event handler? I can give it a try and submit a pull request to them, though absolutely no guarantee that i can find a fix :cry:

chloeleichen:
This is for old ui, can it be closed now? @kring

kring:
Yep, thanks!

PR #1184, opened 2015-12-17
Opacity UI enhancement for tablets and smartphones devices

jeanpul:
I had some issues with my IPad and Smartphone with the opacity slider so I've added a select box. I've reduced the opacity slider granularity for 0.01 to 0.1 in order to match the select box entries but this can be easily changed

Fabien

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 
 ### 1.0.50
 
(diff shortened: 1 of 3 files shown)


kring:
@jeanpul you're probably running into #1021.  I don't think adding another UI element is the right solution, though, so I'm going to close this.

PR #1185, opened 2015-12-17
Fly to with elevation

chris-cooper:
When flying to a rectangle that has a significant terrain elevation the camera ends up at ground level.  This PR performs a single terrain elevation query at the centre of the rectangle and offsets the camera by the appropriate amount.

@kring, I guess this could go into Cesium itself though it will complicate the logic of flyTo slightly due to the asynchronous sampleTerrain.


lib/Models/Cesium.js
```diff
@@ -338,10 +339,35 @@ Cesium.prototype.zoomTo = function(viewOrExtent, flightDurationSeconds) {
...
+        // Work out the destination that the camera would naturally fly to
+        var destinationCartesian = camera.getRectangleCameraCoordinates(viewOrExtent);
+        var destination = Ellipsoid.WGS84.cartesianToCartographic(destinationCartesian);
+
+        var terrainProvider = this.scene.globe.terrainProvider;
+        var level = 6; // A sufficiently coarse tile level that still has approximately accurate height
+        var positions = [Rectangle.center(viewOrExtent)];
+
+        // Perform an elevation query at the centre of the rectangle
+        sampleTerrain(terrainProvider, level, positions).then(function(results) {
```
kring: It would be good to return this promise, and then change `CatalogItem.zoomTo` to return the result of `terria.currentViewer.zoomTo`.  That way the `zoomTo` promise doesn't resolve until the zoom has actually started.

chris-cooper:
Was that along the right lines?

kring:
@chris-cooper I don't think zoom-to is working at all in this branch anymore, because `when(function() { ... })` returns a promise that immediately resolves to a function, without ever invoking that function.  This should work instead: `return when().then(function() { ... });`, as would `return when((function() { ... })());` but there might be a slightly cleaner way.

chris-cooper:
Sorry, my mistake.  I've done it the way you've suggested for now.

kring:
Thanks @chris-cooper!

PR #1191, opened 2015-12-18
Use CZML properties in the feature info template

### LINKED ISSUE #1140: Support property objects in CZML objects for use in templating FeatureInfo boxes
Currently CZML files don't have a well defined way to attach structured information to CZML objects which can then be templated in the front end. @RacingTadpole and @kring have discussed how this might look, and how it would fit in with the rest of Cesium's time varying structure (if one of you two could add some more details of what you've discussed and the potential implementations that would be great). This is likely something we will need early in 2016 for the network opportunity opportunity mapping project on AREMI, which is one of the most highly sought out datasets on the platform.

RacingTadpole:
Depends on https://github.com/TerriaJS/cesium/pull/7 .
Fixes #1140.

Adds a test that uses and completes a string-form featureInfoTemplate with CZML properties.


RacingTadpole:
I've just merged in 1169_dots to this, since it would have conflicted.

kring:
The test `FeatureInfoPanelViewModel CZML templating uses and completes a string-form featureInfoTemplate if present` is failing for me.

RacingTadpole:
Are you using this branch of Cesium: TerriaJS/cesium#7 ?

kring:
Oops, I wasn't, sorry.  That PR is now in `terriajs`.

RacingTadpole:
thanks for adding the time-varying test too!

PR #1201, opened 2015-12-22
Add getContainer() to GlobeOrMap

RacingTadpole:
This unifies the interface for referring to the containing DOM element. We'll need this for the charts panel.


kring:
We should probably add an implementation to `NoViewer`, too, right?

RacingTadpole:
Thanks, nice catch. Done.

PR #1204, opened 2015-12-23
Automatically run test suite as part of Travis build

kring:
This pull request uses Karma and Sauce Labs to automatically run our unit tests on a bunch of browsers on each check-in / pull request.  Currently we run on:
- Chrome 46 on Windows 7
- Safari 9 on Mac OS X 10.11
- Firefox 42 on Windows 7
- Firefox 38 (the current extended support release) on Windows 7

We'll add some Internet Explorer flavors once the tests are in better shape there.  We can also add mobile devices, but I haven't tried it yet.  This handy tool can be used to configure more platforms:
https://wiki.saucelabs.com/display/DOCS/Platform+Configurator?_ga=1.115415494.1515790378.1415140375#/

As an open source project, we have a free Sauce Labs account that allows us to run on up to 5 VMs simultaneously.  I think I can create sub-accounts for anyone that wants to run the tests across browsers on their own system before committing.  Let me know if you want to try it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,12 @@
 language: node_js
 node_js:
 - '0.10'
-addons:
-  sauce_connect:
-    username: "kevin-nationalmap"
-    access_key:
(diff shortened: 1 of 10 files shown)


kring:
Important limitation I should have mentioned above: Sauce Labs does not support WebGL, so any tests that require it must be protected by an `if (supportsWebGL())` check.

PR #1212, opened 2016-01-05
Internet Explorer 9 compatibility and test fixes

### LINKED ISSUE #1193: Lots of test failures in IE9
Even though TerriaJS works reasonably well in IE9, the test suite shows over 50 test failures.

Some problems I can see at a glance:
- We use `console.warn`, but in IE9 no console methods are available unless the debugger is open.  We work around this for `console.log` by adding it in `Terria.js` if it doesn't already exist.  We should do the same for `warn`.
- The tests for `EarthGravityModel1996` are timing out, probably because that class uses `ArrayBuffer`, which is not supported in IE9.  But since that class is only used with Cesium anyway, we should probably just exclude these tests in IE9.
- A bunch of "access denied" errors in `CsvCatalogItem` tests.  I'm not sure what we're being denied access to.
- And more...

kring:
This PR fixes all the test failures in IE9 (including some which were actual bugs in production code, not just test problems) and enables IE9 as a test target for our Travis / Sauce Labs automated tests.

Fixes #1193


kring:
Someone merge me, I'm awesome! :)

You don't necessarily need to be able to run the tests in IE9 because Travis does it automatically starting with this PR.

RacingTadpole:
I just searched on `.children`. which is not supported in IE, and found one stray instance:

https://github.com/TerriaJS/terriajs/blob/master/lib/Models/Leaflet.js#L68

Is this a problem? Can we change it to `childNodes` (noting that you would need to find the first one with the right node type à la https://github.com/AnalyticalGraphicsInc/cesium/pull/3395/files)?

RacingTadpole:
OK, my bad - Leaflet.js is fine on IE9. Turns out `children` _is_ supported on `Element` - just not on `DocumentElement`.

PR #1213, opened 2016-01-05
Catalog members can now have stable ids set in config JSON

### LINKED ISSUE #986: Catalog changes breaks existing Share URL's
We have partners that are embedding AREMI on their website as various iframes with specific layers turned on, which is great. Unfortunately, since the Share URL's use the name of the groups and layers as the key, whenever you change the structure or spelling of anything all existing Share URL's for layers in those parts of the catalog tree break. When these are iframes on somebody's website you effectively break their website...

This is a problem, since we do occasionally change the structure and names of things. Since we also have very close relationships with certain data custodians, they occasionally insist on changes as well.

For now, I've started writing custom init files hosted in a WWW S3 bucket and sending iframe blocks specifying the custom init file as a `#` URL, so that when we change anything I can then manually update those to reflect the new name or structure. However, it will be tedious to maintain these.

Possible solutions:
1. Bill suggested looking at SIRF to see if maybe we could do persistent identifiers: https://wiki.csiro.au/display/SIRF/The+SIRF+API
2. Some form of persistent UID in the init file/CKAN directory that never changes for a layer.
3. Our own URL shortening database where we can update the structure/layer names when they change.

Others?

AlexGilleran:
Fixes #986
- Catalog items/groups now look for an "id" field in the JSON, this
  enables share links to work even if the item is renamed or moved.
- If no id is present, share link generation falls back to the path of the
  catalog member
- Catalog members can specify shareKeys when renaming or changing an id to
  preserve compatibility with old share links.
- If no id is present in a share link, the old name-based matching
  functionality is used to preserve compatibility with old share links.
- Matching when performing updates on catalog groups is now done by id
  rather than name where possible.

### DIFF
--- a/lib/Models/Catalog.js
+++ b/lib/Models/Catalog.js
@@ -2,15 +2,10 @@
 
 /*global require*/
 
-var defaultValue = require('terriajs-cesium/Source/Core/defaultValue');
 var defined = require('terriajs-cesium/Source/Core/defined');
 var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
 var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
 var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
-var RuntimeError = require('terriajs-cesium/Source/Core/RuntimeError');
-var when = require('terriajs-cesium/Source/ThirdParty/when');
-
-var createCatalogMemberFromType = require('./createCatalogMemberFromType');
 var CatalogGroup = require('./CatalogGroup');
 
 /**
@@ -102,45 +97,8 @@ defineProperties(Catalog.prototype, {
  * @returns {Promise} A promise that resolves when the update is complete.
  */
 Catalog.prototype.updateFromJson = function(json, options) {
-    if (!(json instanceof Array)) {
-        throw new DeveloperError('JSON catalog description must be an array of groups.');
-    }
-
-    options = defaultValue(options, defaultValue.EMPTY_OBJECT);
-    var onlyUpdateExistingItems = defaultValue(options.onlyUpdateExistingItems, false);
-
-    var promises = [];
-
-    for (var groupIndex = 0; groupIndex < json.length; ++groupIndex) {
-        var group = json[groupIndex];
-
-        if (!defined(group.name)) {
-            throw new RuntimeError('A group must have a name.');
-        }
-
-        // Find an existing group with the same name, if any.
-        var existingGroup = this.group.findFirstItemByName(group.name);
-        if (!defined(existingGroup)) {
-            // Skip this item entirely if we're not allowed to create it.
-            if (onlyUpdateExistingItems) {
-                continue;
-            }
-
-            if (!defined(group.type)) {
-                throw new RuntimeError('A group must have a type.');
-            }
-
-            existingGroup = createCatalogMemberFromType(group.type,  this.terria);
-
-            this.group.add(existingGroup);
-        }
-
-        promises.push(existingGroup.updateFromJson(group, options));
-    }
-
     var that = this;
-    return when.all(promises, function() {
-        that.group.sortItems();
+    return CatalogGroup.updateItems(json, options, this.group).then(function() {
         that.terria.nowViewing.sortByNowViewingIndices();
     });
 };
@@ -168,7 +126,7 @@ Catalog.prototype.updateFromJson = function(json, options) {
  * @return {Object} The serialized JSON object-literal.
  */
 Catalog.prototype.serializeToJson = function(options) {
-     this.terria.nowViewing.recordNowViewingIndices();
+    this.terria.nowViewing.recordNowViewingIndices();
 
     var json = {};
     CatalogGroup.defaultSerializers.items(this.group, json, 'items', options);
--- a/lib/Models/Catalog.js
+++ b/lib/Models/Catalog.js
@@ -3,6 +3,7 @@
 /*global require*/
 
 var defined = require('terriajs-cesium/Source/Core/defined');
+var defaultValue = require('terriajs-cesium/Source/Core/defaultValue');
 var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
 var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
 var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
@@ -98,7 +99,11 @@ defineProperties(Catalog.prototype, {
  */
 Catalog.prototype.updateFromJson = function(json, options) {
     var that = this;
-    return CatalogGroup.updateItems(json, options, this.group).then(function() {
+    options = defaultValue(options, {});
+
+    options.shareKeyIndex = this.group.generateShareKeyIndex();
+
+    return CatalogGroup.updateItems(json, options, this.group).then(function () {
         that.terria.nowViewing.sortByNowViewingIndices();
     });
 };
--- a/lib/Models/Catalog.js
+++ b/lib/Models/Catalog.js
@@ -8,6 +8,7 @@ var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
 var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
 var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
 var CatalogGroup = require('./CatalogGroup');
+var when = require('terriajs-cesium/Source/ThirdParty/when');
 
 /**
  * The view model for the geospatial data catalog.
@@ -135,5 +136,38 @@ Catalog.prototype.serializeToJson = function(options) {
 
     return this.group.serializeToJson(options).items;
 };
+
+/**
+ * Resolves items in the catalog based on the share keys provided, and enabled them along with all their ancestors in
+ * the catalog hierarchy. This is asynchronous, as to enable an item it must have finished loading first.
+ *
+ * Note that because of the lazily-loaded nature of the catalog, items within it may not be resolvable by shareKey until
+ * their parents have loaded. As a result this loads shareKeys in serial from left to right. If a catalog member is the
+ * child of an asynchronously-loaded catalog group (like a ckan or socrata group), then that group's shareKey precede the
+ * child member.
+ *
+ * @param {String[]} shareKeys An array of string-based share keys.
+ * @returns {Promise} A promise that will resolve when all the items have been loaded and enabled.
+ */
+Catalog.prototype.enableByShareKeys = function (shareKeys) {
+    var shareKeyIndex = this.shareKeyIndex;
+
+    return shareKeys.map(function(shareKey) {
+        return shareKeyIndex[shareKey];
+    }).filter(function (member) {
+        return defined(member);
+    }).map(function (member) {
+        var loadPromise = member.load() || when(); // load() sometimes returns undefined.
+
+        return loadPromise.then(function () {
+            member.enableWithParents();
+        });
+    }).reduce(function (aggregatedPromise, promise) {
+        return aggregatedPromise.then(function() {
+            return promise;
+        });
+    }, when());
+};
+
 
 module.exports = Catalog;
--- a/lib/Models/Catalog.js
+++ b/lib/Models/Catalog.js
@@ -151,24 +151,58 @@ Catalog.prototype.serializeToJson = function(options) {
  *      possible to pass an empty object if nothing needs updating.
  * @returns {Promise} A promise that will resolve when all the items have been loaded and enabled.
  */
-Catalog.prototype.updateByShareKeys = function (sharedObjects) {
-    var shareKeyIndex = this.shareKeyIndex;
+Catalog.prototype.updateByShareKeys = function(sharedObjects) {
+    return Object.keys(sharedObjects).reduce(function(aggregatedPromise, shareKey) {
+        var itemJson = sharedObjects[shareKey];
 
-    return Object.keys(sharedObjects).reduce(function (aggregatedPromise, shareKey) {
-        return aggregatedPromise.then(function() {
-            var existingMember = shareKeyIndex[shareKey];
-
-            if (existingMember) {
-                return existingMember.updateFromJson(sharedObjects[shareKey])
-                    .then(function () {
-                        return when(existingMember.load());
-                    }).then(function () {
-                        existingMember.enableWithParents();
-                    });
-            }
-        });
-    }, when());
+        return aggregatedPromise
+            .then(this._loadInSerial.bind(this, itemJson.parents || []))
+            .then(this._updateAndEnable.bind(this, shareKey, itemJson));
+    }.bind(this), when());
 };
 
+/**
+ * Calls {@link CatalogMember#load} on a number of catalog members, identified by their share key, one after the other
+ * from left to right. Doing this in serial is important because sometimes calling load() on a catalog group will
+ * reveal more catalog items under it, which will be added to the catalog's shareKeyIndex - if these were done in
+ * parallel, share keys towards the right of the array would not able to be resolved at all.
+ *
+ * @param shareKeys An array of catalog member share keys to use for resolving catalog members from the catalog
+ * @returns {Promise} A promise that will be resolved when all the catalog members have either loaded or failed to load.
+ * @private
+ */
+Catalog.prototype._loadInSerial = function(shareKeys) {
+    return shareKeys.reduce(function(aggregatedPromise, shareKey) {
+        return aggregatedPromise.then(function() {
+            if (this.shareKeyIndex[shareKey]) {
+                return this.shareKeyIndex[shareKey].load();
+            }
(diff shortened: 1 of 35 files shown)


lib/Models/CatalogGroup.js
```diff
@@ -367,10 +340,25 @@ CatalogGroup.prototype.toggleOpen = function() {
...
 
 /**
+ * Finds the first item in this group with a value in {@link CatalogItem#allShareKeys} that matches the given key.
+ * The search is case-sensitive.
+ *
+ * @param {String} key The key to search for.
+ * @return {CatalogMember} The first item with the given key, or undefined if no item with that name exists.
+ */
+CatalogGroup.prototype.getChildForShareKey = function(key) {
+    return this.items.find(function(item) {
```
kring: `find` doesn't work in IE, unfortunately.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

kring:
@AlexGilleran does this allow preservation of share links even if an item moves to an entirely new group?  For example, if we have a catalog with this organization, and C is enabled, and then we share:

```
A
  B
    C
```

Can it be made to successfully load once the catalog is updated to, for example:

```
D
   E (former C)
```

It's not obviously looking at the code that this can be done, but maybed I've missed it.  I understand that AREMI needs this sort of thing though.

AlexGilleran:
Nope, back to the drawing board...

kring:
@axman6 @meh9 can you guys confirm that this is actually required? :)

AlexGilleran:
@meh9 said it was required in the other issue - unfortunately I didn't realise that the way that enabled catalog members are resolved was hierarchical, which was silly seeing how it was right in front of me :(.

I guess this means I need to change the resolution code so it just considers the `uniqueId` and scans through the entire catalog of layers/groups to find each catalog member.

meh9:
Yeah, so we have moved this between groups in the past, not just renamed the item. We work a lot with UX and occasionally redo the structure of things (i.e. move entire groups around etc).

AlexGilleran:
Have modified it so it works with moving members between groups, by first generating an index of all the items in the catalog by their share keys and doing a lookup against it when updating.

lib/Models/CatalogGroup.js
```diff
@@ -412,4 +402,135 @@ CatalogGroup.prototype.sortItems = function(sortRecursively) {
...
+    }, {});
+};
+
+/**
+ * Adds an item to the provided index under the provided key, checking to make sure there's no item for that key already.
+ * Will throw an error if so.
+ */
+function addToIndex(indexSoFar, key, item) {
+    if (indexSoFar[key]) {
+        throw new DeveloperError('Duplicate shareKey: ' + key);
```
kring: I think this should be a `RuntimeError`, not a `DeveloperError`, because it's probably caused by a duplicate key in the init file, not by a coding error, right?

As a general rule, we should be able to remove all code that checks for and throws `DeveloperError`, and our app will still work correctly.  In fact, Cesium does exactly this in release builds.

lib/Models/CatalogGroup.js
```diff
@@ -412,4 +402,135 @@ CatalogGroup.prototype.sortItems = function(sortRecursively) {
...
+            if (!defined(item.type)) {
+                throw new RuntimeError('A catalog member must have a type.');
+            }
+
+            existingItem = createCatalogMemberFromType(item.type, catalogGroup.terria);
+
+            catalogGroup.add(existingItem);
+        } else {
+            // We only want to use the id for matching, not overwrite it.
+            delete item.id;
```
kring: Can we avoid this delete somehow?  Maybe by explicitly skipping `id` in `updateFromJson`?  Two reasons:
- Deleting properties pretty much destroys the browser's ability to optimize property access to this instance.  Maybe not a big deal here, but still...
- It's better to avoid mutating the parameters to a function whenever possible.

lib/Models/CatalogGroup.js
```diff
@@ -346,6 +315,8 @@ CatalogGroup.prototype._getValuesThatInfluenceLoad = function() {
  * @param {CatalogMember} item The item to add.
  */
 CatalogGroup.prototype.add = function(item) {
+    item.parent = this;
```
kring: This `parent` property should be defined in the `CatalogMember` constructor, rather than added dynamically later.  It can be undefined initially, of course.

Currently CkanCatalogGroup can and does add a single `CatalogItem` to multiple `CatalogGroups`.  :fearful: 

We could potentially change that, or maybe there's another way to do this?

lib/Models/CatalogGroup.js
```diff
@@ -355,6 +326,8 @@ CatalogGroup.prototype.add = function(item) {
  * @param {CatalogMember} item The item to remove.
  */
 CatalogGroup.prototype.remove = function(item) {
+    delete item.parent;
```
kring: Just set it to undefined, don't delete it.

lib/Models/Terria.js
```diff
@@ -550,17 +550,25 @@ function generateInitializationUrl(url) {
...
+        var promises = [], updateSources = [];
 
         for (var i = 0; i < initSources.length; ++i) {
             var initSource = initSources[i];
             if (!defined(initSource)) {
                 continue;
             }
-            promises.push(terria.addInitSource(initSource));
+
+            if (initSource.catalogOnlyUpdatesExistingItems) {
```
kring: Why is this necessary?

lib/Models/CatalogGroup.js
AlexGilleran: Sure - I'm thinking of the init file as being something that's made by developers though ;). Changed.

lib/Models/Terria.js
AlexGilleran: Because I need to build a complete map of all the ids in the initial data before I update them with values from the share link.

Without splitting it into two steps like this, generating the map just returns a map of all the root catalog groups as the promises that read the children into those groups haven't executed yet.

kring: hmm there might be a deeper problem then (with the existing code I mean).  The intention is that the initSources would layer.  Later initSources are applied on top of earlier ones.  But that isn't necessarily true as soon as any part of applying an init source is asynchronous, right?

If we instead change the code to do `addInitSource` on the first init source, and wait for its promise to resolve before doing `addInitSource` on the second, I think that would eliminate the need for the two stages you've added here, and possibly also fix some other current/potential weird problems.  Does that sound right?

AlexGilleran: Sure, if that's the intention then it make sense to do it that way. I was worried about causing previously concurrent data loads to now happen one after the other.

kring: Yeah that's a reasonable concern.  I think it's ok, though.

meh9:
I've been doing some testing of this branch and have noticed what I think is a problem. If I move a Group (in an expanded state with a layer in the group enabled) somewhere else after getting a share url for it, then it enables the layer but doesn't expand the group. It's pretty minor though!

AlexGilleran:
So I've spent this week simplifying the way that serialization works and the way that sharing enabled catalog members works. Hopefully this seems like a step forward for you too :grimacing: .
- Refactored catalog serialization to use composable filter functions
  instead of a series of flags. So now the logic for whether you serialize certain items is passed in from whatever originally calls `serializeToJson` rather than being done by a big ever-expanding series of flag checks.
- Made CatalogMember#isUserSupplied false by default - this seems the more reasonable default? Certainly surprised me.
- Made catalog hold a constantly-up-to-date shareKeyIndex rather than generating it before loading enabled items, as enabling items causes more loads (e.g. ckan, socrata) that mutate the catalog further.
- Made all mutations of CatalogGroup#items go through CatalogGroup#add

kring:
Some high level thoughts before I get into the code:

> Made CatalogMember#isUserSupplied false by default - this seems the more reasonable default? Certainly surprised me.

"user supplied" in this case really means "it didn't come from loading a built-in catalog, so we need to serialize the whole thing."  `true` is the correct default so that catalog items created programmatically, for example, are serialized correctly.  It's easy for us to set it to false when we know we created it from a built-in catalog, and that's exactly what we do.

> Made all mutations of CatalogGroup#items go through CatalogGroup#add

`items` is part of the public API, and it's mutable, so if using it instead of calling `add` leads to bugs, there will be bugs.  You can probably subscribe to notification on `items` to do whatever you need to do, rather than forcing the user to go through `add`.

lib/Models/Catalog.js
```diff
@@ -149,30 +125,49 @@ Catalog.prototype.updateFromJson = function(json, options) {
...
+ */
+Catalog.prototype.enableByShareKeys = function (shareKeys) {
+    var shareKeyIndex = this.shareKeyIndex;
+
+    return shareKeys.map(function(shareKey) {
+        return shareKeyIndex[shareKey];
+    }).filter(function (member) {
+        return defined(member);
+    }).map(function (member) {
+        var loadPromise = member.load() || when(); // load() sometimes returns undefined.
```
kring: `when(member.load())` accomplishes the same thing slightly more cleanly.

lib/Models/Catalog.js
```diff
@@ -149,30 +125,49 @@ Catalog.prototype.updateFromJson = function(json, options) {
...
+
+    return this.group.serializeToJson(options).items;
+};
 
-    var json = {};
-    CatalogGroup.defaultSerializers.items(this.group, json, 'items', options);
-    return json.items;
+/**
+ * Resolves items in the catalog based on the share keys provided, and enabled them along with all their ancestors in
+ * the catalog hierarchy. This is asynchronous, as to enable an item it must have finished loading first.
```
kring: It's not actually necessary to load before enabling.  You can just set `isEnabled = true` and the catalog item will load if necessary.  The only reason to load explicitly is if you need a promise.

AlexGilleran: Yeah, I need the promise so I can wait for this to load before trying to enable its children (because otherwise I won't know whether they exist to match keys with). Probably can just immediately enable after calling `load` then.

kring:
It seems in this branch we now only serialize the IDs of catalog items that are opened or enabled?  That's not sufficient.  We lose the `opacity` setting on sharing, for example.  And some catalog items (e.g. CSV, ABS) have other settings that need to be serialized as well, such as the selected data variable.

kring:
Just FYI: the ABS layers are broken in this branch at the moment, but I think merging in master should fix it.

AlexGilleran:
- Re-added serialization of properties like opacity for sharing.
- Changed isUserSupplied default back to true

lib/Models/CatalogGroup.js
```diff
@@ -257,11 +232,41 @@ freezeObject(CatalogGroup.defaultSerializers);
  * @type {String[]}
  */
 CatalogGroup.defaultPropertiesForSharing = clone(CatalogMember.defaultPropertiesForSharing);
-CatalogGroup.defaultPropertiesForSharing.push('items');
 CatalogGroup.defaultPropertiesForSharing.push('isOpen');
+CatalogGroup.defaultPropertiesForSharing.push('items');
 
 freezeObject(CatalogGroup.defaultPropertiesForSharing);
 
+CatalogGroup.prototype.setupItemListeners = function() {
```
kring: This doesn't need to be part of the public API does it?

lib/Models/CatalogGroup.js
```diff
@@ -257,11 +232,41 @@ freezeObject(CatalogGroup.defaultSerializers);
...
+                } else {
+                    this.terria.catalog.shareKeyIndex[key] = item;
+                }
+            }, this);
+        }.bind(this),
+        deleted: function(item) {
+            item.parent = undefined;
+
+            item.allShareKeys.forEach(function(key) {
+                this.terria.catalog.shareKeyIndex[key] = undefined;
```
kring: I think you need to do this before setting `item.parent = undefined`.  Otherwise the key you're removing won't be the same as the one you previously added.

lib/Models/CatalogGroup.js
```diff
@@ -367,10 +372,27 @@ CatalogGroup.prototype.toggleOpen = function() {
...
+        }
+    }
+};
+
+/**
  * Finds the first item in this group that has the given name.  The search is case-sensitive.
  *
  * @param {String} name The name of the item to find.
  * @return {CatalogMember} The first item with the given name, or undefined if no item with that name exists.
+ * @deprecated Use {@link CatalogGroup#getChildForShareKey} where possible, this is retained for backwards compatibility
```
kring: Consider using the `deprecationWarning` function from Cesium.

lib/Models/CatalogGroup.js
AlexGilleran: Fixed.

lib/Models/CatalogGroup.js
AlexGilleran: Yeah, good catch. Made a test for this and fixed.

lib/Models/CatalogGroup.js
AlexGilleran: Done.

AlexGilleran:
Addressed all those :)

kring:
It seems catalog items won't be shared if they're inside groups that are closed.  For example:
- Open Data.gov.au -> Australian Taxation Office.
- Enable both 2011-2012 and 2012-2013 datasets.
- Close the Australian Taxation Office group.
- Click Share and load the URL in a new window.

The Now Viewing will be blank.

AlexGilleran:
Yeah that's an issue. The parent needs to be opened in order to know that the child exists, but because the child can have moved since the generation of the share link you can't be 100% whether opening any particular parent is actually going to reveal that child.

I guess it could make a better effort in this case - serializing the ids of every parent of a shared catalog item whether open or not, perhaps.

AlexGilleran:
So fixed that. Also realised that the existing code doesn't actually check if groups are open at all (whoops) so removed that - this makes it a lot simpler. Now it just shares each enabled `CatalogItem`'s JSON against the `CatalogItem`'s id, along with a `parents` property which has the id of all that `CatalogItem`'s ancestors. When a share link is opened, it loads each ancestor sequentially from parent to child, so that when it tries to look up the shared `CatalogItem`'s id, it can find it.

kring:
Order of layers isn't preserved on sharing, but it's not preserved in the version in nationalmap.gov.au either (haven't looked into why yet), so no need to fix it here unless it's easy.

kring:
Sharing is including too much information about custom items.  For example:
- Click Add Data, enter `http://geofabric.bom.gov.au/simplefeatures/ows`
- Open `Geofabric 2.1.1 - Web Map Service`, `Aquifer Contours 2.1`
- Enable `AHGFAquiferContour`
- Click Share

The share link goes on for pages, and copying it makes Chrome noticeably slow on my system.  The one on nationalmap.gov.au is (relatively) short in comparison.  The problem is that the URL in this branch unnecessarily includes details of all the layers queried from the WMS GetCapabilities

AlexGilleran:
- Rebased against master with the csv refactor, actual changes since the last review are [here](https://github.com/TerriaJS/terriajs/compare/59a170ef9f4b912de789d28aaa423059cc401388...986-stable-catalog-item-ids)
- Fixed the problem with sharing user-added catalogs by making it so that children of catalog groups with a url aren't serialized. There's no need for this anymore as the descendants of these will be loaded from the url anyway, then if the user has changed anything for a descendant of these groups (like changing opacity etc) then that will be set from `sharedCatalogMembers` after they've loaded.
- Made it so that CatalogMembers are only added to `terria.catalog.shareKeyIndex` when they're actually part of `terria.catalog`. This prevents a situation where items without a specific id could be added to a disconnected parent and hence indexed against a path like `ParentName/Name`, then have their parent added to the tree and have their default uniqueId change to `Root Group/ParentName/Name`, but the index not updated.
- Turned `combineFilters` into a smarter function that makes sure duplicate filters are removed.
- Got rid of the `items` serializer in a number of `CatalogGroup` subclasses switching flags back and forth to ensure that only sharable properties and enabled items are serialized - these now all call into a single static function in `CatalogGroup` that uses the new `itemFilter`/`propertyFilter` system.

kring:
There are two deprecation warnings at startup:

```
Use CatalogGroup.getChildForShareKey where possible, this is retained for backwards compatibility with existing share links.
dataSetID is deprecated. Please set datasetId instead.
```

If we want to remove these from the public API but we still need to use them internally, create a new private API for our use and make the public one call `deprecationWarning` before calling the private one.

Or maybe these (especially `findFirstItemByName`) are still useful and we shouldn't deprecate them at all?  I previously suggested adding `deprecationWarning`, but that might have been bad advice.  Maybe I should have suggested just toning down the "this is deprecated" language in the documentation instead (e.g. `Consider using {@link CatalogGroup#getChildForShareKey} where possible as it provides an improved way of finding a catalog member that is resistant to renames.`)

AlexGilleran:
Yeah good point, removed the deprecationWarning as I suppose there's legitimate backwards-compatibility reasons to match on name the same way we are. Also removed getChildForShareKey completely because I just realised that the code doesn't use it at all anymore, all lookups are done against the catalog index.

PR #1214, opened 2016-01-05
Fixed timeline hiding as soon as any time-series layer is hidden.

### LINKED ISSUE #965: Timeseries controls disappear when any layer is disabled even when other timeseries are enabled
When multiple timeseries layers are enabled, turning any layer off removes the timeseries controls.

Steps:
1. Go to: http://nationalmap.gov.au/#test
2. Turn on at least two layers in Test Data -> Temporal.
3. Turn any layer off.
4. Observe timeseries controls disappear, even though there are still timeseries layers visible.

AlexGilleran:
Fixes #965.

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -602,13 +602,14 @@ var scratchRectangle = new Rectangle();
 CatalogItem.prototype.useClock = function() {
     var that = this;
     return when(this.load(), function() {
-        if (defined(that.clock) && that.isEnabled && that.isShown) {
-            that.clock.getValue(that.terria.clock);
-            that.terria.showTimeline++;
-        }
-        else {
-            that.terria.clock.shouldAnimate = false;
-            that.terria.showTimeline = 0;
+        if (defined(that.clock)) {
+            if (that.isEnabled && that.isShown) {
+                that.clock.getValue(that.terria.clock);
+                that.terria.showTimeline++;
+            }
+            else {
+                that.terria.showTimeline--;
+            }
         }
     });
 };
--- a/lib/Models/Terria.js
+++ b/lib/Models/Terria.js
@@ -309,6 +309,13 @@ var Terria = function(options) {
             this.currentViewer.zoomTo(this.initialView, 2.0);
         }
     }, this);
+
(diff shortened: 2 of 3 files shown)


kring:
It's not just the shown/hidden state of the timeline that matters, though; the interval it's showing matters too.  When you enable a time-dynamic layer, Terria shows the timeline and sets the timeline interval to that layer's interval.  If you enable another one, that new one's interval is shown.  If you then click on the first (i.e. "zoom to"), the interval will change back to that layer's interval.

Now, what should happen when you disable a time-dynamic layer?  Currently it just hides the timeline entirely.  To bring it back, just zoom to one of your remaining time-dynamic layers.  Better would be to only hide the timeline if the layer you just hid is the one whose interval is currently showing on the timeline.  Better still (maybe?) would be to jump to the next time-dynamic layer's interval, unless there isn't one, in which case hide the timeline.

But the current behavior in this PR, of keeping the timeline up but showing an interval that is no longer applicable, doesn't make a lot of sense to me.  Am I thinking about it wrong?

AlexGilleran:
Nah I was thinking about it wrong - because there's only one clock I thought that one timeline was supposed to control all the layers and that it must adjust its interval to contain all of them - the fact that both monthly solar datasets have the same interval didn't help.

It sounds like we should make it more obvious somehow that timelines are associated with specific layers rather than being global? It certainly fooled me.

Either that or actually make it so that the timeline automatically expands to fit the earliest and latest points in all the layers, but that might mean that layers with really small intervals are hard to use.

meh9:
Hah, another feature that I had no idea existed - I've never seen the timeline change when using Zoom To! This certainly does complicate things.

kring:
> Either that or actually make it so that the timeline automatically expands to fit the earliest and latest points in all the layers

This sounds ok to me.  Ideally we'd indicate on the timeline (somehow!) which sub-intervals actually have data.  The Cesium timeline (which we're not using anymore) has basic support for this:

http://cesiumjs.org/releases/1.17/Apps/TimelineDemo/index.html

image

AlexGilleran:
Will reevaluate and come up with a better solution to the core problem, discussion continues in #965, closing this.

PR #1216, opened 2016-01-05
Fetch Esri MapServer legends in JSON and hence draw our own.

### LINKED ISSUE #1168: Display legends provided in JSON by ArcGis MapServer
Although Mapserver's JSON format is a little messy, it definitely gives us more flexibility than having to display a whole pre-formatted image, including layers we're not interested in.

[Example](http://www.asris.csiro.au/ArcGis/rest/services/TERN/BDW_ACLEP_AU_NAT_C/MapServer/legend?f=pjson)

```

  "layers": [
    {
      "layerId": 1,
      "layerName": "BDW_000_005_EV_N_P_AU_NAT_C",
      "layerType": "Raster Layer",
      "minScale": 0,
      "maxScale": 0,
      "legend": [
        {
          "label": "< 1.0 (g/cm3)",
          "url": "6b11a35cb97d3d60e4487134548947d4",
          "imageData": "iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAYAAACNiR0NAAAAAXNSR0IB2cksfwAAAAlwSFlzAAAOxAAADsQBlSsOGwAAADxJREFUOI1jYaAyYKGlgf+pYB4jigv/P3xIvkny8gwMDDT28qiBowaOGjhqIHYDYUUQmYAR3UBGSkyDAQDGRwSSq4VGyQAAAABJRU5ErkJggg==",
          "contentType": "image/png",
          "height": 20,
          "width": 20
        },
        {
          "label": "1.0 - 1.2 (g/cm3)",
          "url": "a7f788fe78d8a838b5f9eb47aed17569",
          "imageData": "iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAYAAACNiR0NAAAAAXNSR0IB2cksfwAAAAlwSFlzAAAOxAAADsQBlSsOGwAAAD1JREFUOI1jYaAyYKGlgf+pYB4jigs/79hItkm8Hv4MDAw09vKogaMGjho4aiB2A2FFEJmAEd1ARkpMgwEA6NsE9zyMX40AAAAASUVORK5CYII=",
          "contentType": "image/png",
          "height": 20,
          "width": 20
        },
        {
```

[Documentation](http://www.asris.csiro.au/ArcGis/sdk/rest/index.html#//02ss0000007n000000)

stevage:
Fixes #1168


lib/Models/ArcGisMapServerCatalogItem.js
```diff
@@ -178,28 +180,43 @@ defineProperties(ArcGisMapServerCatalogItem.prototype, {
...
-        var that = this;
-        return when.all([serviceMetadata, layersMetadata]).then(function(results) {
+        var serviceUri = baseUri.clone();
+        var layersUri = baseUri.clone().segment(layers); // either 'layers' or a number
+        var legendUri = baseUri.clone().segment('legend');
+
+        var serviceMetadata = this.mapServerData || getJson(serviceUri);
+        var layersMetadata = this.mapServerLayersData || getJson(layersUri);
+        var legendData = getJson(legendUri).
+            then(this.loadLegendFromJson.bind(this)).
```
kring: You should only do this if the catalog item doesn't already have a `legendUrl`.  If one is supplied in the init file, it should take precedence over an automatically generated one.

stevage: Thanks for the reminder, fixed.

kring:
I tried adding `http://www.ga.gov.au/gis/rest/services/topography/Dynamic_National_Map_Elevation_Relief_and_Physiography/MapServer` via Add Data, and then enabled some layers and looked at their legend.  They all got the default "Open the legend in a separate tab legend" instead of your fancy new ones.  I think that's because the legend is generated in the wrong place in `ArcGisMapServerCatalogItem`.  It should happen in `updateFromMetadata` instead of in `_load`, and `ArcGisCatalogGroup` should be modified to load the legend data, too, before calling `updateFromMetadata`.

stevage:
Ok, fixed the first part of that. I'm a little nervous that now it's firing off a bunch of legend.json requests before the user has even indicated they want to display it. That is, in the case of a catalog group, we fetch and generate all the legends, but maybe the user only wants to show one layer.

stevage:
Hmm I don't think I quite get how this LegendUrl is supposed to operate. Should `Legend.asPngUrl()` be returning a `LegendUrl`?

kring:
> Hmm I don't think I quite get how this LegendUrl is supposed to operate. Should Legend.asPngUrl() be returning a LegendUrl?

I don't think it needs to.  The purpose of `LegendUrl` is to pair a legend URL with some information about what type of legend it is (image versus an HTML page) for cases where we can't infer that it's an image from the URL (e.g. the URL ends in `.png`).  It's necessary because WMS-provided legend URLs sometimes don't indicate PNG-ness anywhere in their URL.

stevage:
One issue remaining perhaps is that generated legends shouldn't be serialised. Probably this should be handled as a flag on the LegendUrl class, and a `defaultSerializers` on `CatalogItem`?

kring:
> Probably this should be handled as a flag on the LegendUrl class, and a defaultSerializers on CatalogItem?

Yeah, that makes sense to me!

kring:
@stevage I tweaked a few things.  The legend URL serialization was a mess (not from this PR, just in general) and it was causing the generated URL to be serialized in at least one case that I found.  I think I have that fixed up now.  I also restored some of the testing logic that was in place before.  Better (less fragile) tests would be nice, of course, but it wasn't hard to fix them and I think that's better than changing them to not really test the same thing anymore.  So I'm ready to merge this, but I want to give you a chance to look at it first.

PR #1217, opened 2016-01-06
Added ability to show raw data of feature

### LINKED ISSUE #1171: Ability to show original key/value/data of templated feature info
Now that we have the awesome new capability to template the feature info data, it would be useful to have the capability to show the original data - i.e. the untemplated version. In many cases the templates will not display all the data, so there could be more details in the raw data that is interesting to power users. 

The way I imagine it is we have a link/button on a templated view that would expand a panel with the raw data, something like:

screen shot 2015-12-15 at 17 08 32

When clicked, this expands to show something like the original untemplated data, something like this:

viewraw

AlexGilleran:
Fixes #1171 

There's now a button in the feature panel that allows the raw data for a feature to be shown if templated data is normally available, as the templated data is usually only a subset of everything available.

Probably don't merge this yet as there's an ongoing discussion over how the show/hide button should look in #1171.

### DIFF
--- a/lib/Styles/FeatureInfoPanel.less
+++ b/lib/Styles/FeatureInfoPanel.less
@@ -71,7 +71,6 @@
 .feature-info-panel-body {
     .opaque-to-input;
 }
-//data-bind="style: { 'max-width': maxWidth + 'px' }
 
 .feature-info-panel-sections {
     overflow-y: auto;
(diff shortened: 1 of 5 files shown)


kring:
Is this still waiting on some discussion?

axman6:
How will this work with @RacingTadpole's work on putting properties on CZML packets as JSON objects? Will it (format and) print the raw JSON?

AlexGilleran:
No longer waiting on discussion now.

@axman6: All this does is display exactly what would normally get displayed if there was no template. It doesn't make any changes to that table, so @RacingTadpole altered that raw data table then this will show that as well.

axman6:
Ok, good to know - @RacingTadpole Maybe we should support showing the raw JSON in the feature info box for the new CZML stuff? (I'm still working on that btw, it's just tedious)

AlexGilleran:
Close enough with the raw data download issue that depends on this, so I might as well close this and do two issues in one PR

PR #1218, opened 2016-01-06
Tweaked the location bar

keattang:
- Removed spacing in UTM zone to be in line with the standard.
- Call `sampleAccurateHeight` when not using `EllipsoidTerrainProvider` so as to allow accurate sampling with providers other than `CesiumTerrainProvider`


kring:
Thanks @keattang!

PR #1219, opened 2016-01-07
Use terriajs-cesium 1.15.5.

kring:


keattang:
@kring what's the timeline on using cesium 1.17?

kring:
We haven't been in a hurry to due AnalyticalGraphicsInc/cesium#3386.  If you have a pressing need we could potentially cherry pick the fix, though.

keattang:
Not particularly pressing, just looking forward to the performance benefits :)

kring:
Here are the differences between 1.15.4 and 1.15.5:
https://github.com/TerriaJS/cesium/compare/terriajs-1.15.4...terriajs-1.15.5

1.15.6 upgraded knockout, but I'm going to revert that because there seems to be an incompatibility between the new version of knockout and our markdown binding.
https://github.com/TerriaJS/cesium/compare/terriajs-1.15.5...terriajs-1.15.6

kring:
Done, back to 1.15.5.

Issue #661, opened 2015-04-30
Split out server-side services


stevage:
I've been talking with @kring about splitting out the NodeJS services. Terria (NM) is essentially a static HTML app (albeit with a lot of clientside JavaScript and a moderately heavy build process), with a fairly loose dependency on a couple of web services:
- `/proj4def` - a trivial REST lookup service for proj4 definitions
- `/convert` - a web service wrapper for ogr2ogr, allowing users to view geospatial files in the browser that aren't directly supported browser-side
- `/proxy` - a proxy service which could be handled by nginx (except that it has the list of acceptable proxy targets)

The dependency is loose, in that there could be lots of client installations all depending on one proj4def/convert/proxy service, or vice versa, or they could all be split out, or whatever.

Benefits to splitting them out:
- Terria becomes a static HTML/JS app, which is brain-deadly simple to deploy and requires no server side infrastructure at all. People could deploy it to Github, to site44.com, etc. Which means more uptake, which perhaps means more developers etc etc. (We should provide a bundled zip for download.)
- It also means that _we_ can deploy instances more easily, and could look at CDNs etc for hosting.
- We can separately manage the actual proxying service(s), with clearer discussion about who we proxy for and why, and replace it with a tool that's really good at proxying (nginx). Instead of thinking of the proxy as some kind of hack, we can treat it as a service in its own right, and use it to solve other problems like #658 and #659.

By "splitting them out" I propose:
- this repo contains only the client-side code and its related test and build system
- we start a separate repo "TerriaConvert" (open to other names...) which is just the conversion service (maybe other stuff while we work out what do about them permanently)

Incidentally, is there anything that `/proj4def` does that spatialreference.org doesn't? 

For instance:

http://spatialreference.org/ref/epsg/28354/proj4/
http://nationalmap.nicta.com/proj4def/epsg:28354

stevage:
The server side stuff now lives here: https://www.npmjs.com/package/terriajs-server

For now, I'm leaving aside the question of whether we really need the proj4 service etc.

stevage:
Demo of static site here: http://stevage.github.io/nm-static/

PR #1221, opened 2016-01-07
Configurable services

stevage:
Centralise management of proxy/, convert/ and proj4def/ URLs, and allow overriding
them in config.json. This facilitates the decoupling of TerriaJS and TerriaJS-Server.

Supports #661.


kring:
As discussed on slack, I think it would be good to use `Terria.configParameters` instead of teaching `Terria` about each of these services specifically.  This is because Terria itself doesn't really need any of them, only specific (and optional) parts of the library need them.

PR #1224, opened 2016-01-13
Csv and ABS Refactor

### LINKED ISSUE #1037: Multi feature info and CSV region mapping problem
When multiple layers are enabled together with a CSV region mapped layer, and you click on a location where there is no mapped region but feature info for the other layers are returned, then an empty entry for the region mapped CSV layer is still included in the feature info box. When you click on a location where no region has been mapped and there is also no feature info returned for the other layers, then there i no data for the region mapped CSV.

### LINKED ISSUE #1091: Code issues in CsvCatalogItem/TableDataSource
@kring and I were talking about some of the remaining issues (after recent CSV refactoring) that would be nice to fix up at some point:
- Clarify what purpose TableStyle serves. At the moment there is a lot of wasted code setting properties on CsvCatalogItem.TableStyle, passing it to TableDataSource, which then copies those values onto itself. Perhaps TableStyle should encapsulate the user's desired style, as distinct from choices Terria makes about how to display stuff.
- Closely related, the ability to choose a column to display at run time lacks a lot of features:
  - You can only specify styling for the primary `dataVariable`
  - Once the user selects a different column, we don't have a way to remember what that original styling was
- Various guesses are made about data types and how to display them in many different places: sometimes bottom up (DataTable), sometimes top down (CsvCatalogItem). It would be nice to have a clear strategy for inferring data types (integer, string,...), semantics (depth, lat, lon, ...), and hence presentation (how many color bins...), and how the user can override these guesses.

### LINKED ISSUE #1207: Can show wrong feature descriptions of animated csvs
Eg. with /#test, droughts by country, clicking on Australia in 1991 shows the Botswana 1965 info.
image

RacingTadpole:
A major rewrite of the csv framework.

Clearly defines responsibilities of each class. `TableDataSource` handles features. `RegionMapping` handles region-mapping. Both of these use `TableStyle` (which is the only repository of style information). `TableStructure` holds the data. Each `TableColumn` knows whether it is selected. These two inherit from the relevant `Concept` classes, so there is no doubling up of variables for data and UI.  `LegendHelper` handles information that combines a `TableColumn` and `TableStyle`. `CsvCatalogItem.load` loads the `TableStructure`, then decides whether it needs `TableDataSource` or `RegionMapping`; it then ensures a column is active. 

`TableColumn` also does a lot of processing of date-times.

I have tried to follow https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Documentation/Contributors/CodingGuide/README.md .

Breaking changes include the APIs of (not including those which begin with `_`):
- `CsvCatalogItem`: `rowProperties`, `rowPropertiesByCode`, `dynamicUpdate` have been removed.
- `TableDataSource`: Completely rewritten and moved to `Models` directory.
- `DataTable` and `DataVariable` have been replaced with new classes, `TableStructure` and `TableColumn`.
- `RegionProvider`: `loadRegionsFromWfs`, `processRegionIds`, `applyReplacements`, `findRegionIndex` have been made internal functions.
- `RegionProviderList`: `chooseRegionProvider` has been changed and renamed `getRegionDetails`.
- `ColorMap`: `fromArray` and `fromString` have been removed, with the constructor taking on that functionality.
- `LegendUrl` has been moved to the `Map` directory.
- `TableStyle`: `loadColorMap` and `chooseColorMap` have been removed.
- `FeatureInfoPanelSectionViewModel`: its constructor now takes a `FeatureInfoPanelViewModel` as its first argument, instead of `Terria`.

Fixes #1091. Fixes #1207. Fixes #1037.


lib/Map/MapError.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @alias MapError
+ * @constructor
+ *
+ * @param {Object} options Object with the following properties:
+ * @param {Object} [options.sender] The object raising the error.
+ * @param {String} [options.title='An error occurred'] A short title describing the error.
+ * @param {String} options.message A detailed message describing the error.  This message may be HTML and it should be sanitized before display to the user.
+ */
+var MapError = function(options) {
```
kring: Maybe we should just rename `ModelError` to `TerriaError` and move it to Core?

RacingTadpole:
OK, this PR is ready to go again. I've updated the ABS catalog item to use the new CSV work.
You'll notice that the ABS items now use the new default choroplething too. We can set them back by using a suitable colorMap in the json.

RacingTadpole:
Down to just one error in the specs on IE9 with SauceLabs now:
`ArcGisMapServerCatalogItem can load json FAILED: Expected 4 to equal 2.`
https://github.com/TerriaJS/terriajs/blob/csv_refactor/test/Models/ArcGisMapServerCatalogItemSpec.js#L91-L105
It loads the url `http://my.arcgis.com/`. Any ideas why IE9 would hit more urls than other browsers for that url? Note the next test loads `http://my.arcgis.com/MapServer/3` and passes on IE9.

RacingTadpole:
Good news - I've fixed the specs on IE9.

kring:
In NationalMap:
- Enable `Data.gov.au -> Australian Securities and Investments Commission (ASIC) -> ASIC - Credit Licensee Register`.
- Click Share
- Copy the URL to a new window.

There's an exception on load of the new window in `CsvCatalogItem`:
image

RacingTadpole:
Thanks @kring, all bugs so far uncovered now fixed.

kring:
The display variable for CSVs is not preserved on sharing.

kring:
When I share an ABS layer with "Display as percent of region population" disabled, I get an error on load of the share URL:
image

RacingTadpole:
ok, those are fixed too.

kring:
Mind cleaning up the TODOs in this code?  If it's something that still needs to be verified, it probably should be done before merging.  If you're reasonably convinced it's ok now, just remove it.  If it can't be evaluated right now for whatever reason, remove it and write a github issue.

lib/Models/TableDataSource.js
```diff
@@ -0,0 +1,1008 @@
...
+            if (defined(dataSource._regionImageryLayer)) {
+                layerIndex = dataSource._regionImageryLayer._layerIndex;  // Would prefer not to access an internal variable of imageryLayer.
+                ImageryLayerCatalogItem.hideLayer(dataSource._catalogItem, dataSource._regionImageryLayer);
+                ImageryLayerCatalogItem.disableLayer(dataSource._catalogItem, dataSource._regionImageryLayer);
+                dataSource._regionImageryLayer = undefined;
+            }
+            setNewRegionImageryLayer(dataSource, layerIndex, regionIndices);
+            ImageryLayerCatalogItem.showLayer(dataSource._catalogItem, dataSource._regionImageryLayer);
+        }
+    });
```
tephenavies: @RacingTadpole I think this promise needs error handling, but I'm not sure how to implement it.

RacingTadpole: Nice catch, thanks Steve. All fixed now! In fact I've obliterated regionPromise in favour of a simpler approach - load() only resolves once regionDetails is available.

RacingTadpole:
I've massively simplified the region promise, so it is simply part of the load.
And cleaned up the TODOs.
Thanks @kring (fingers crossed it passes the travis checks!)...

kring:
There seems to be some lag in the legend and map view when choosing ABS concepts.  For example:
- Enable Country of Birth
- Switch to Now Viewing
- Change Region Type to States and Territories
- Change Region Type to Statistical Area 2

I get a view like this:

image

The map is showing states instead of SA2, and the legend shows that some region apparently has 200% of its population born in Australia.

RacingTadpole:
Thanks, fixed that one too. I've also done a nested refactor, so that `RegionMapping` is separated out from `TableDataSource`. `CsvCatalogItem` does a little more work now, choosing between the two. This should also help later when we have non-geo-spatial csvs.

RacingTadpole:
oops - the opacity slider has gone missing. Let me fix that first.

RacingTadpole:
ok, all good now. I hope...

RacingTadpole:
I've now merged in here my further changes from #1244, so that it handles time varying ABS data, eg. ABS_ANNUAL_ERP_ASGS.

In more detail -
- Removes 'display as percent' if not applicable
- Adds json parameter uniqueValuedConcepts
- Checks that region values are in same order in dataset and total population datafiles
- Filters out unsupported region codes such as GCCSA

You can try this out in combination with the `more_abs` branch of [national map](https://github.com/NICTA/nationalmap/tree/more_abs).

PR #1225, opened 2016-01-13
Make sure things in `Map` do not call things in `Models`

RacingTadpole:
Merge this after csv_refactor #1224.

Make sure things in `Map` do not call things in `Models`

Moves `TableDataSource` into Models, which is required because it references a number of Models itself (for region mapping).

Creates a new `MapError` class, identical to `ModelError`, which `RegionProvider` can call. `RegionProvider` is in the Map directory, but was calling `ModelError`, which was in Models.


RacingTadpole:
Note we could go further here, and eliminate classes that refer to `knockout` from Map. That would be `PickedFeatures` and `CesiumSelectionIndicator`. What do you think?

RacingTadpole:
I have now merged this directly into #1224.

PR #1226, opened 2016-01-13
Fix problem with multi feature info and CSV region mapping (#1037)

### LINKED ISSUE #1037: Multi feature info and CSV region mapping problem

RacingTadpole:
Merge after #1224 and #1225. 
Fixes #1037.


RacingTadpole:
btw, it looks like the travis build failed, but the push test passed; only the PR test failed. Its log doesn't show any errors. So I think it's a CI glitch.

RacingTadpole:
Merged into #1224.

PR #1233, opened 2016-01-14
Fixes `gulp` command so it runs without errors

RacingTadpole:
We left a stray `var LegendUrl = require('./LegendUrl')` in `AbsIttCatalogItem.js`.


kring:
Thanks @RacingTadpole!

PR #1236, opened 2016-01-15
Add a few JSdoc tags to support DataSourceEditor generation.

stevage:
Some were missing, some were wrong, and some are just extra niceties.

### DIFF
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -44,7 +44,7 @@ var ImageryLayerCatalogItem = function(terria) {
     /**
      * Gets or sets the opacity (alpha) of the data item, where 0.0 is fully transparent and 1.0 is
      * fully opaque.  This property is observable.
-     * @type {Number|String}
+     * @type {Number}
      * @default 0.6
      */
     this.opacity = 0.6;
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -44,7 +44,7 @@ var ImageryLayerCatalogItem = function(terria) {
     /**
      * Gets or sets the opacity (alpha) of the data item, where 0.0 is fully transparent and 1.0 is
      * fully opaque.  This property is observable.
-     * @type {Number}
+     * @type {Number|String}
      * @default 0.6
      */
     this.opacity = 0.6;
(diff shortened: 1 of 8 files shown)


lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -44,7 +44,7 @@ var ImageryLayerCatalogItem = function(terria) {
     /**
      * Gets or sets the opacity (alpha) of the data item, where 0.0 is fully transparent and 1.0 is
      * fully opaque.  This property is observable.
-     * @type {Number}
+     * @type {Number|String}
```
kring: Opacity can't actually be a string can it?

kring:
Looks good other than the one question.

lib/Models/ImageryLayerCatalogItem.js
stevage: You wouldn't think so, but: https://github.com/NICTA/nationalmap/blob/master/datasources/09_Tasmanian_Government.json#L8

stevage: And https://github.com/terriajs/terriajs/blob/0615a601b87538362a2617d1bf87797518111449/wwwroot/test/init/composite.json#L29-L29

And https://github.com/terriajs/terriajs/blob/0615a601b87538362a2617d1bf87797518111449/wwwroot/test/init/test-tablestyle.json#L18-L18

kring: Haha ok.  Does that actually work?  Should we just fix the JSON files?

stevage:
Three ways to fix:
1. Accept as-is (acknowledge that actually we do accept strings)
2. Add `@editortype {Number|String}`, indicating that we don't really want strings, but they are kind of legitimate, and the editor will allow them.
3. Fix the JSON files.

Just to clarify: if the schema says "Number" only, then the editor rejects JSON files that contain strings, without a very helpful error message.

kring:
3 sounds like the right approach to me.  It might be nice if the editor provided a more helpful error in this situation (because it can easily come up other places, too), but that's a separate issue.

stevage:
Want me to update all other strings-that-shouldn't-be-strings while I'm at it? (I'm looking at you, Rectangle...)

kring:
Which rectangles are strings?

stevage:
Heaps of them: https://github.com/nicta/nationalmap/blob/bf52ae434bbebcea4e8abcfbf764ea6b056fe2ff/datasources/00_National_Data_Sets/00_02_Land.json#L40-L40

kring:
Oh, right.  Yeah, that's dodgy. :)

stevage:
Ok, fixed. Ready for another look, @kring.

PR #1240, opened 2016-01-20
Replace Models/ModelError with Core/TerriaError

RacingTadpole:
#1225 prevented things in `Map` from using `ModelError`, which meant we had two identical error classes, `MapError` and `ModelError`. This recombines them as `Core/TerriaError`.

Merge this after #1224.

Per https://github.com/TerriaJS/terriajs/pull/1224#discussion_r49815910


lib/Map/MapError.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @alias MapError
+ * @constructor
+ *
+ * @param {Object} options Object with the following properties:
+ * @param {Object} [options.sender] The object raising the error.
+ * @param {String} [options.title='An error occurred'] A short title describing the error.
+ * @param {String} options.message A detailed message describing the error.  This message may be HTML and it should be sanitized before display to the user.
+ */
+var MapError = function(options) {
```
kring: `MapError` isn't needed anymore right?

PR #1241, opened 2016-01-20
Add 'make-schema' gulp and npm task, included in 'gulp release'.

stevage:


gulpfile.js
```diff
@@ -39,7 +39,11 @@ gulp.task('release-specs', ['prepare-cesium'], function() {
     return build(specJSName, glob.sync(testGlob), true);
 });
 
-gulp.task('release', ['release-specs']);
+gulp.task('make-schema', function(done) {
+    child_exec('npm run make-schema', undefined, done);
```
kring: This is a little weird.  Usually npm calls gulp, not the other way around.  Gulp is a normal node environment, so you could package `gen-schema` as a CommonJS module, require it in, and call it directly from gulp.

stevage: Ok, it was a little weird. Reworked. (gen-schema already has a CommonJS module, `generateSchema.js`).

PR #1242, opened 2016-01-20
Show Raw Data and Download as CSV/JSON

### LINKED ISSUE #1171: Ability to show original key/value/data of templated feature info

### LINKED ISSUE #1173: Download feature info data as CSV/JSON/XML or similar
Some users have mentioned that it would be useful to be able to download or export the feature info data as a CSV. 

Since data can be other than CSV, it should probably be a copy of the original data received from the feature info http request, and so would probably be in whatever format was received from the server? Or we could just export the processed data from the feature info datastructure (which I guess could mean that it is structured data, so might need to be JSON or XML).

I imagine it would look something like:

screen shot 2015-12-15 at 17 23 08

AlexGilleran:
Fixes #1171 
Fixes #1173

Added a 'Show Raw Data' and 'Hide Raw Data' link to feature info panels that have templated data - this link shows/hides a table with the original data table that is normally seen on features with no template configured.

Also added a drop-down to just below this table that allows the raw data to be downloaded as CSV (default) or JSON. If the data is too hierarchical to do as CSV, only JSON is available.

This drop-down was a bit of an issue as it often lives at the bottom of a scrollable modal, and is expected to overlap the edge of the modal. As such the bit that actually drops down is fixed positioned, with its position reset every time it's shown. If any click is made or the modal is scrolled then the dropdown closes again. This event handling uses native javascript instead of knockout because it seemed like a neater way to do it - there only needs to be a handler attached to either of these events, that's immediately unattached once executed.

The other problem is that this works by putting the data into a data URI, which works fine on everything except MS browsers, as these don't allow downloading data URIs at all. Unfortunately I can't think of a good way to feature-detect this, so I've just browser-sniffed IE/Edge and display a warning if the user clicks the download link with these browsers.

A small issue is that the filename will be "Unknown" in Safari as it doesn't read the "download" attribute on links, but it does respect that it should be downloaded rather than opened.


kring:
It seems that multi-feature selection is broken in this branch (I didn't try other branches, so there's a chance it's broken in master).  Enable two layers and click a spot where both have data.  Only one feature shows up in the feature info panel.

Much smaller problem: if the drop down list is partially off the screen and I click it, I get this:

image

Is it crazy to just use the browser's built-in `select` element instead of a custom thing?

AlexGilleran:
A `select` would work but not quite the same. Ask @cam-grant :)

kring:
Right, I guess it wouldn't also function as a clickable button, for one thing.

AlexGilleran:
- Fixed broken multi-feature select (caused by a swallowed exception in a binding)
- Fixed the problem in the screenshot - now when clicked the entire dropdown overlaps the edge of the panel.

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -29,6 +35,18 @@ var FeatureInfoPanelSectionViewModel = function(terria, feature, catalogItem) {
     this.feature = feature;
     this.name = feature.name ? feature.name : '';
     var template = defined(catalogItem) ? catalogItem.featureInfoTemplate : undefined;
+    var data = propertyValues(feature.properties, terria.clock);
+
+    knockout.track(this, ['name', 'templatedInfo', 'rawData', 'catalogItemName', 'rawDataVisible', 'downloadOptionsVisible', 'downloadDropdownPosition']);
+    knockout.track(this.terria, ['selectedFeature']);
```
kring: Terria is already rigging up `selectedFeature` to be observable.  I don't think this line is needed.

kring: Ah I see you just moved this from below.  Probably worthwhile to remove it anyway.

AlexGilleran:
Rebased and removed the double tracking.

AlexGilleran:
Fixed those tests :)

kring:
Looks good!  Please update `CHANGES.md` in master.

PR #1244, opened 2016-01-22
Handle wider range of ABS ITT data

RacingTadpole:
In particular, handles time varying ABS data, eg. ABS_ANNUAL_ERP_ASGS.
Merge after #1224.

In more detail -
- Removes 'display as percent' if not applicable
- Adds json parameter uniqueValuedConcepts
- Checks that region values are in same order in dataset and total population datafiles
- Filters out unsupported region codes such as GCCSA

You can try this out in combination with the `more_abs` branch of [national map](https://github.com/NICTA/nationalmap/tree/more_abs).


RacingTadpole:
Merged into #1224.

PR #1247, opened 2016-01-25
Timeline shows current layer's name, open timelines are tracked by stack

### LINKED ISSUE #965: Timeseries controls disappear when any layer is disabled even when other timeseries are enabled

AlexGilleran:
Fixes #965.

Currently the Terria timeline keeps track of the most recently enabled time-series layer. If two time-series layers are enabled then the one currently being tracked is disabled, the timeline disappears, which confuses users - it's not clear that the timeline is only tracking one layer at a time, and they don't realise that they have to get the timeline for the still-open layer back by clicking on it.

This adds a stack that keeps track of the time-series layers currently enabled. When a new time-series layer is enabled, it's placed on the top of the stack. When it's disabled, the new top of the stack becomes the layer that was added previously, and the timeline is switched to track this layer instead of disappearing. If the user has multiple time-series layers open and clicks on one to zoom to it, that layer is moved to the top of the stack and hence tracked by the timeline.

In order to make this obvious to the user, the timeline now displays the name of the layer that it's currently tracking.


kring:
@AlexGilleran can you merge in master when you get a chance?

AlexGilleran:
Done :)

lib/Models/TimeSeriesStack.js
```diff
@@ -0,0 +1,73 @@
...
+
+    knockout.track(this);
+
+    /**
+     * The highest time-series layer, or undefined if there are no time series layers.
+     */
+    knockout.defineProperty(this, 'topLayer', {
+        get: function () {
+            if (this._layerStack.length) {
+                return this._layerStack[this._layerStack.length - 1];
```
kring, line 29: You probably want `_layerStack` in the `knockout.track` above.  Otherwise, I can't see much point in `topLayer` being a knockout property, or in subscribing to it, because it will never change after it's computed once.

lib/Models/TimeSeriesStack.js
```diff
@@ -0,0 +1,73 @@
...
+
+    knockout.getObservable(this, 'topLayer').subscribe(function() {
+        if (defined(this.topLayer)) {
+            // If there's a top layer, make the clock track it.
+            this.topLayer.clock.getValue(this.clock);
+        } else {
+            // If there's no layers, stop the clock from running.
+            this.clock.shouldAnimate = false;
+        }
+    }.bind(this));
```
kring: `subscribe` takes a second parameter which is the `this` to use.

AlexGilleran:
Done - I thought the the knockout tracking did its knockout magic and figured out what it needed to track to keep the computed property up to date, but it's actually working because if `knockout.track` is called without a list of properties it just tracks everything. Have narrowed it down to just `_layerStack`.

kring:
Aah, I didn't know that about `knockout.track`, makes sense!

kring:
This is pretty slick!  Please update CHANGES.md in master for this one, too.

Issue #799, opened 2015-07-01
Don't use transparent as "no data" color (because Cesium-drawn empty circles are nearly unclickable)


meh9:
When adding tabular data to Cesium containing points and some lines do not have a value for whatever column is currently being visualised, then these points are drawn "empty", i.e. no colour fill. Unfortunately this makes the point very hard to click since it appears you have to click the line of the circle in order to bring up the data for that point.

The `Test` entry in this CSV shows the issue once dragndropped onto NM:

```
Some data, Place, Lat, Lon, Data1, Data2
1, NICTA,-35.2817828,149.1248208, Some more data from the CSV, And some more
4, Parkwood Road Substation,-35.2195879,148.996752, Some more data from the CSV, And some more
5, Cotter Dam,-35.3202023,148.9398087, Some more data from the CSV, And some more
,Test,-35.29,149, Some more data from the CSV, And some more
```

stevage:
Closely related issue: points in GeoJSON files with marker icons are very hard to click on as well - you have to click on the very base of the signpost (not the sign itself).

kring:
The difficulty in clicking marker icons is caused by AnalyticalGraphicsInc/cesium#2894.

meh9:
This has been highlighted by GA testing of AREMI. It would be nice to have this fixed, as it is a problem we just live with at the moment.

The best way to test is to turn on `Electricity Inf.` -> `Generation` -> `Current Power Generation` -> `All` in AREMI and play around clicking inside and on the edge of the unfilled circles:

http://nationalmap.gov.au/renewables/#share=0fR0dV

stevage:
Ok, so let's reframe this issue as "Don't leave circles unfilled, even when there is no data". Then it becomes a lot simpler. At the least, we could use a very low opacity colour.

meh9:
I do like the look of having the circles empty, as that really says "no value" to me. If we can retain that look, or something like it using a a nearly transparent colour (or white?) would be great.

RacingTadpole:
FYI, #1250 added a `nullColor` option to `tableStyle`: "defaults to a dark blue. This colour is used to display null values (but it does not appear on the legend; we should add it). It is also used to colour points when no variable is selected."

RacingTadpole:
I believe this is fixed now by the addition of `nullColor`, `replaceWithNullValues` and `replaceWithZeroValues`.
[...]

PR #1250, opened 2016-01-27
Improve treatment of missing values, hyphens and 'na' in numeric csv columns

RacingTadpole:
Many csv files (eg. in AREMI) have unusual markers for missing data, eg. '-' or 'na'.
As a result their data is shown as if it were enumerated data instead of numerical data.
This fixes that.
It introduces three new json `tableStyle` parameters:
- `replaceWithZeroValues`: defaults to `[null, '-']`. These values are coloured as if they were zero if they appear in a list with numbers. `null` catches missing values.
- `replaceWithNullValues`: defaults to `['na', 'NA']`. These values are coloured as if they were null if they appear in a list with numbers.
- `nullColor`: defaults to a dark blue. This colour is used to display null values (but it does not appear on the legend; we should add it). It is also used to colour points when no variable is selected.


RacingTadpole:
A possible future enhancement: if a column looks sufficiently numeric (say at least 10 rows and >50% numbers), always replace any non-numbers with either zero or null. Is that a good idea?

axman6:
I assume you only mean for colouring that point/region, the data still needs to be available in the feature info as is.

RacingTadpole:
Yes, good point - that is right.

lib/Core/TableColumn.js
```diff
@@ -134,6 +149,11 @@ var TableColumn = function(name, values, options) {
 
 inherit(VariableConcept, TableColumn);
 
+function replaceValues(values, replaceWithZeroValues, replaceWithNullValues) {
+    // Replace "bad" values like "-" with zero, and "na" with null.
+    var replaced = values.map(function(value) { return (replaceWithZeroValues.indexOf(value) >= 0) ? 0 : value; });
+    return replaced.map(function(value) { return (replaceWithNullValues.indexOf(value) >= 0) ? null : value; });
```
kring, line 155: It might be nice to do this in a single pass over the array instead of 2, especially since each pass makes a copy of the array.

kring:
Looks good other than the one comment, which isn't all that important, so I'm merging this.  You can do it in master or a separate PR if you're up for it.

lib/Core/TableColumn.js
RacingTadpole, line 155: good point. revision pushed.

RacingTadpole:
This addresses #799, but I'm not sure if it actually solves it.

PR #1251, opened 2016-01-28
Recognise (and test) more time formats

RacingTadpole:
Adds tests for YYYY and YYYY-MM formats, which were already recognised.

Newly recognises 2016-01-03 12:15:59" (ie. without the "T") as local times.

Merge after #1250.

Also displays times more nicely in the feature info panel - fixes #1249 .

PR #1252, opened 2016-01-28
Czml test files

RacingTadpole:
Just adds a CZML test file with time-varying description, and an init file that can be drag-and-dropped straight into a working instance.
Merge after #1251.


RacingTadpole:
Closing in favour of #1259.

PR #1255, opened 2016-01-29
Created UrlTemplateCatalogItem

keattang:


kring:
Thanks @keattang!  Mind adding the new catalog item to `registerCatalogMembers`?  A really basic test or two would be nice too, though admittedly this is pretty simple.

keattang:
Yeah no worries will get onto that when I get some time. 

Not sure why the travis builds are failing? Doesn't seem related.

Also do you have any opinions on getting the meta data from `tilemapresource.xml` as discussed here #825 ?

kring:
Unfortunately Travis builds don't work for PRs from outside the main repo, due to Travis encryption limitations.

stevage:
So is this basically a more general version of [OpenStreetMapCatalogItem](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/OpenStreetMapCatalogItem.js)? Seems a little bit unnecessary to have both?

kring:
Yeah, I pretty much agree with that.  `OpenStreetMapCatalogItem` has more assumptions about how the URLs are built, but it's pretty easy to access OSM layers using this.  Cesium's `UrlTemplateImageryProvider` wasn't available when `OpenStreetMapCatalogItem` was created, FWIW.  Also, Cesium's corresponding `OpenStreetMapImageryProvider` is now deprecated.

tobybellwood:
I came across a small issue with OpenStreetMapCatalogItem when trying to utilise basemaps from Here and CartoDB Retina:
Here - https://{1-4}.base.maps.cit.api.here.com/maptile/2.1/maptile/{map id}/{scheme}/{zoom}/{column}/{row}/{size}/{format} (per https://developer.here.com/rest-apis/documentation/enterprise-map-tile/topics/request-constructing.html)

CartoDB Retina - http://{s}.basemaps.cartocdn.com/{basemap-name}/{z}/{x}/{y}@2x.png (per https://cartodb.com/basemaps/)

The OpenStreetMapCatalogItem has a hardcoded period before format, meaning that custom formats needed a rewrite of the code to move the period (or slash, or @2x.) into the this.fileExtension

Is there now a move to use the new UrlTemplateImageryProvider in Terria?

keattang:
My main reasoning behind this was that I wanted to serve my own TMS tiles through to Terria with a list of subdomains and @kring suggested creating a `UrlTemplateCatalogItem` instead of a `TMSProviderCatalogItem` as it is the more general case.

stevage:
In my experience, most apps that allow custom tiles request the user to type a format like this:

`{s}.tiles.site.com/mylayer/{z}/{x}/{y}.png`

I think it would be better for us to use that pattern as the standard, rather than the less flexible approach written into OpenStreetMapCatalogItem.

tobybellwood:
@stevage - from what I can see, UrlTemplateImageryProvider doesn't hardcode the period into the format, so I'm happy!

kring:
Thanks for getting the ball rolling on this, @keattang.  I've added some tests and made some minor tweaks and opened #1385.

PR #1257, opened 2016-01-29
Fix Leaflet picking when zoomed far enough out that the world is repeated

### LINKED ISSUE #1256: Feature picking in Leaflet doesn't work when zoomed out so that the map is repeated
Clicking on a feature in the repeated part of the map (either to the left or right of the "main" map (which seems to be the map in the middle) causes a selection circle to appear on the equivalent "main" map point but no feature information is shown.

tephenavies:
`CesiumTileLayer.pickFeatures` can receive longitude values that are not in the range [-pi, pi). This change will wrap longitude values into the range [-pi, pi) before calling `imageryProvider.pickFeatures`.

Fixes #1256.


lib/Map/CesiumTileLayer.js
```diff
@@ -170,6 +170,10 @@ var CesiumTileLayer = L.TileLayer.extend({
     },
 
     pickFeatures: function(map, longitudeRadians, latitudeRadians) {
+        // Limit longitudeRadians to the range [-pi, pi)
+        function mod(x, n) { return ((x % n) + n) % n; } // Fix JS weird modulo operator
+        longitudeRadians = mod(longitudeRadians + Math.PI, 2*Math.PI) - Math.PI;
```
kring: Use `CesiumMath.negativePiToPi` instead.

kring:
Also, please update CHANGES.md.

tephenavies:
@kring Where should I put the changelong entry? In 2.0.0 or above it?

kring:
In 2.0.0.

PR #1258, opened 2016-01-29
Do not show html tag type variables in Now Viewing

RacingTadpole:
This fixes this bug:
- Click the very first data item in http://nationalmap.gov.au/renewables/ , "All generation types".
- Go to the "Legends" tab.
- At the bottom is "Latest 24h generation", which is actually an `` tag. It does not make sense to color points according to this.

This PR hides such html tag type variables.
Merge after #1253.


axman6:
Hopefully our use of HTML in CSVs will disappear after the charts work is implemented, it's definitely a hack at the moment.

PR #1268, opened 2016-02-04
Newui Search Refactor

AlexGilleran:
@kring @chloeleichen @RacingTadpole 

There's an equivalent branch for this in nationalmap as well.

The initial objective of this was to refactor search in such a way that we were no longer passing flags in to determine what to search, but it snowballed a bit from there:
- A `viewState` is now defined in the root of the ui and this is passed down to all components that need it. This is essentially a viewmodel that keeps track of bits of the ui state that are mutated publically - e.g. what `CatalogItem` is currently previewed, or whether the modal is open.
- The `viewState` is then referenced by the `render()` methods of the components that actually use this state, and the `ObserveModelMixin` takes care of rerendering when they change.
- If we wanted to be super clever we could make the entire state of `viewState` readonly and then only allow it to be mutated by functions, creating something akin to Flux's concept of actions.
- I imagine in a full implementation `viewState` would just be the root of a tree with grouped bits of functionality, so that we're not passing the entire state of the app around everywhere.

Apart from that I split search up into different versions for the sidebar and the modal, seeing how they overlap very very little now - but they still use some of the same inner components (isn't react great?).

I also formatted and reorganised a bit of JSX that was getting messy. I always tend to put functions that are called by the main `render` function underneath rather than on top of it as it's way easier for me to read, not sure if it's the opposite for you guys?

I can't imagine everyone will agree with this so looking forward to some discussion :).


lib/ReactViews/DataCatalogGroup.jsx
```diff
@@ -1,64 +1,49 @@
...
-        previewedCatalogItem: React.PropTypes.object,
-        onPreviewedCatalogItemChanged: React.PropTypes.func,
-        isOpen: React.PropTypes.bool,
-        onToggleOpen: React.PropTypes.func
+        group: React.PropTypes.object.isRequired,
+        viewState: React.PropTypes.object.isRequired
     },
 
-    toggleOpen(e) {
-        if (defined(this.props.onToggleOpen)) {
```
kring, line 28: This strange-looking logic is needed for catalog item search.  When a group matches the search query, that group should be shown initially closed even if it was open in the catalog view.  So it can't use the normal `group.isOpen` property.

lib/ReactViews/DataCatalogGroup.jsx
```diff
@@ -1,64 +1,49 @@
...
+                        {this.renderGroup(group)}
+                    </ul>
+                )}
             </li>
-            );
+        );
+    },
+
+    renderGroup(group) {
+        if (!group.isLoading) {
```
kring: We want to incrementally show the items even while we're loading.  This is important for the CSW group, for example, because it takes _forever_ to load everything, but the first 100 items appear fairly quickly.

AlexGilleran:
- Added an override flag that can be passed to a group to allow it to manage its own isOpen state that's active for searches
- Made loading bar show up alongside results so far for groups.

AlexGilleran:
Merged in latest.

PR #1270, opened 2016-02-04
Add support in CswCatalogGroup to create groups in Data Catalog Tab

sppigot:
Needs to be coupled with pull request on NICTA/aremi-natmap#37 - this pull request has an example of the config for the CswCatalogGroup

### DIFF
--- a/lib/Core/overrideProperty.js
+++ b/lib/Core/overrideProperty.js
@@ -2,6 +2,7 @@
 
 /*global require*/
 var defined = require('terriajs-cesium/Source/Core/defined');
+var definedNotNull = require('terriajs-cesium/Source/Core/definedNotNull');
 var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
 
 /**
@@ -26,7 +27,7 @@ var overrideProperty = function(owner, propertyName, descriptor) {
     // Define the new property.
     knockout.defineProperty(owner, propertyName, descriptor);
 
-    if (defined(overridden)) {
+    if (definedNotNull(overridden)) {
         // Notify subscribers to the old property that it has changed.
         // This is a hacky way of getting most subscribers to now subscribe to the new property.
         overridden.notifySubscribers();
--- a/lib/Models/CswCatalogGroup.js
+++ b/lib/Models/CswCatalogGroup.js
@@ -9,6 +9,7 @@ var CsvCatalogItem = require('./CsvCatalogItem');
 var GeoJsonCatalogItem = require('./GeoJsonCatalogItem');
 var inherit = require('../Core/inherit');
 var KmlCatalogItem = require('./KmlCatalogItem');
+var LegendUrl = require('../Map/LegendUrl');
 var TerriaError = require('../Core/TerriaError');
 var WebMapServiceCatalogGroup = require('./WebMapServiceCatalogGroup');
 var WebMapServiceCatalogItem = require('./WebMapServiceCatalogItem');
@@ -315,8 +316,8 @@ CswCatalogGroup.prototype._load = function() {
                 }
 
 
-                // maybe more than one url that results in a data layer here - so check for 
-                // the acceptable ones, store the others as downloadUrls that can be 
+                // maybe more than one url that results in a data layer here - so check for
+                // the acceptable ones, store the others as downloadUrls that can be
                 // displayed in the metadata summary for the layer
                 var downloadUrls = [], acceptableUrls = [], legendUrl;
                 for (var m = 0;m < uris.length; m++) {
@@ -336,20 +337,20 @@ CswCatalogGroup.prototype._load = function() {
                     downloadUrls.push({
                       url: url.toString(),
                       description: defined(url.description) ? url.description : url.name
-                    });  
+                    });
                   }
                 }
 
-                // Now process the list of acceptable urls and hand the metadata 
+                // Now process the list of acceptable urls and hand the metadata
                 // record and the downloadUrls to each data layer item we create
                 for (var j = 0; j < acceptableUrls.length; ++j) {
                   var uri = acceptableUrls[j];
-  
+
                   var group = that;
                   if (that.metadataGroups.length > 0) {
                     group = findGroup(that, that.metadataGroups, record);
                   }
-  
+
                   if (defined(group)) {
                     var catalogItem = createItemForUri(that, record, uri, downloadUrls, legendUrl);
                     if (defined(catalogItem)) {
@@ -371,7 +372,7 @@ CswCatalogGroup.prototype._load = function() {
     }
 
     function loadDomain() {
-        var getDomainUrl = cleanUrl(that.url) + '?service=CSW&version=2.0.2&request=GetDomain&propertyname='+that.domainSpecification.domainPropertyName;  
+        var getDomainUrl = cleanUrl(that.url) + '?service=CSW&version=2.0.2&request=GetDomain&propertyname='+that.domainSpecification.domainPropertyName;
         return loadXML(proxyUrl(that.terria, getDomainUrl)).then(function(xml) {
           if (!xml || !xml.documentElement || (xml.documentElement.localName !== 'GetDomainResponse')) {
             throw new TerriaError({
@@ -387,8 +388,8 @@ sending an email to <a href="mailto:'+that.terria.supportEmail+'">'+that.terria.
           var json = xml2json(xml), listOfValues = json.DomainValues.ListOfValues.Value;
           for (var i = 0; i < listOfValues.length; i++) {
             var keys = listOfValues[i].split(that.domainSpecification.hierarchySeparator);
-            // recursively find the group that the last key in keys should belong to and add that key 
-            findLevel(keys, 0, that.metadataGroups, that.domainSpecification.hierarchySeparator, that.domainSpecification.queryPropertyName); 
+            // recursively find the group that the last key in keys should belong to and add that key
+            findLevel(keys, 0, that.metadataGroups, that.domainSpecification.hierarchySeparator, that.domainSpecification.queryPropertyName);
           }
         }).otherwise(function(e) {
           throw new TerriaError({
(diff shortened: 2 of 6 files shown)


sppigot:
[AddingHierarchyToCSWSourcesInNationalMap-ver0.2.pdf](https://github.com/TerriaJS/terriajs/files/118276/AddingHierarchyToCSWSourcesInNationalMap-ver0.2.pdf)

kring:
Thanks for the pull request, @sppigot!  A few questions.

In your document, you say:

> Why not specify each criteria for the groups in the NationalMap configuration as a CSW
> GetRecords search on the server?

And this is a great question, because that seems like the right approach to me.  The answer is not particularly satisfying:

> Since each and every server request has a fixed overhead, it is more efficient to make a single request, even though the response to that request may be relatively large.

What sort of overhead are we talking about here?  In my limited experience with CSW servers, getting back a large response with a simple query was excruciatingly slow.  It's better, in my opinion, to get back smaller results after a few seconds than to make the user wait a _very_ long time to get all the results (that's what I saw on the marlin server, at least, which is why we made the default behavior to request only 100 records at a time).

Plus, we don't need a bunch of new code if we just do a separate query per group.

In any case, I'm more or less ok with your changes to `CswCatalogGroup.js` if they're necessary.  Regarding the changes to `CatalogItemInfo.html`, I'll have to look at these more critically because they affect lots of other data sources across a dozen projects, not just AREMI.  I wonder if you could simply use the existing `info` property on catalog items and populate it with custom data, rather than making these changes to the panel?

sppigot:
Thanks for the comments @kring.

Overhead: I'm referring to is whether it is faster to do lots of searches each with a fixed overhead than it is to search once and then page through the set of search results. This needs to be tested on a decent sized catalogue - happy to go with which ever is fastest so I'll test and turn this gut feeling into something more substantial.

On your comment about needing to page the results from marlin: CSW search results are intended to be paged through - the search results must be marshalled into whatever form they are requested in (in this case the CSW dublin core standard) so requesting a large number of records without paging the results would indeed be slow.

As regards new code: Not much extra code in either approach I suspect? - the metadataGroups in the config still need to be traversed and added no matter which way we go, the regex matching code in the current approach would be replaced with code to build the CSW filter queries if we went that way. 

On the changes to CatalogItemInfo.html - The changes I have made are more about labelling though - they are about grouping the metadata obtained from the service description/getcapabilities so that it can be distinguished from the richer set of metadata that is available from the CSW service. For example: you can see why we don't want the service description/getcapabilities to be referred to as the metadata URL and why we don't want the service contact to be referred to as the 'data custodian'. The idea behind the suggested changes is that it is better to group all the service description/getcapabilities stuff together with the service URL - it relates to the service and it shouldn't distract from the richer (and potentially much more accurate metadata) that might be available from the CSW record for the dataset delivered by the service. Note: if the metadata from the CSW record isn't available then the service stuff appears as before (albeit in a different order). Hope this helps when you're thinking about the changes suggested here!

kring:
I'm playing around with some different ways of doing this, but I'm having a hard time getting the CSW server to behave the way I expect.  Maybe you can point me in the right direction @sppigot?  I'm doing a query like this:

`curl -X POST -d @test.xml http://waveatlas.marine.csiro.au/geonetwork/srv/eng/csw --header "Content-Type:text/xml"`

Where text.xml looks like this:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" version="2.0.2" service="CSW" resultType="results" startPosition="1" maxRecords="100">
    <csw:Query typeNames="csw:Record" xmlns:ogc="http://www.opengis.net/ogc" xmlns:gml="http://www.opengis.net/gml">
        <csw:ElementSetName>full</csw:ElementSetName>
        <csw:Constraint version="1.1.0">
            <ogc:Filter>
                    <ogc:PropertyIsLike wildCard="%" singleChar="#" escapeChar="!">
                        <ogc:PropertyName>dc:title</ogc:PropertyName>
                        <ogc:Literal>ww3.aus_4m.CgE_avg%</ogc:Literal>
                    </ogc:PropertyIsLike>
            </ogc:Filter>
        </csw:Constraint>
    </csw:Query>
</csw:GetRecords>
```

But this returns no records.  If I remove the `%` from the end of the `ogc:Literal`, it finds the one record with the title `ww3.aus_4m.CgE_avg`, as expected.  Does the server not support querying titles with wildcards, or am I doing something wrong?

sppigot:
It does support query with wildcards - nothing wrong with your query. 

I suspect there is a problem with the way in which the title field is indexed - probably the tokenizing going on around the punctuation and numbers in either the title or the search term.

As a temporary measure, switch your query to dc:subject - layer names are indexed as keywords and the tokenizer for this field leaves special chars alone.

sppigot:
Hmm - yes, upon further examination, one of those decisions had been made about the analyzer/tokenizer which could go either way: both decisions (the original and the new one I have made) are justifiable though I think the way I have changed it now is better: less likely to violate the principle of least astonishment. In short: the results of your original query on dc:title are what you should expect now ie. 13 records. 

BTW The titles of these records will change to something more descriptive in the near future.

kring:
Thanks @sppigot, that's working nicely, now.  However, I'm still having similar problems with other queries.  For example, this returns no results:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" version="2.0.2" service="CSW" resultType="results" startPosition="1" maxRecords="100">
    <csw:Query typeNames="csw:Record" xmlns:ogc="http://www.opengis.net/ogc" xmlns:gml="http://www.opengis.net/gml">
        <csw:ElementSetName>full</csw:ElementSetName>
        <csw:Constraint version="1.1.0">
            <ogc:Filter>
                    <ogc:PropertyIsLike wildCard="%" singleChar="#" escapeChar="!">
                        <ogc:PropertyName>dc:subject</ogc:PropertyName>
                        <ogc:Literal>Earth Science | Oceans | Ocean Waves</ogc:Literal>
                    </ogc:PropertyIsLike>
            </ogc:Filter>
        </csw:Constraint>
    </csw:Query>
</csw:GetRecords>
```

Even though this query works and returns results with `dc:subject` set to `Earth Science | Oceans | Ocean Waves`:

``` xml
<?xml version="1.0" encoding="utf-8"?>
<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" version="2.0.2" service="CSW" resultType="results" startPosition="1" maxRecords="100">
    <csw:Query typeNames="csw:Record" xmlns:ogc="http://www.opengis.net/ogc" xmlns:gml="http://www.opengis.net/gml">
        <csw:ElementSetName>full</csw:ElementSetName>
        <csw:Constraint version="1.1.0">
            <ogc:Filter>
                    <ogc:PropertyIsLike wildCard="%" singleChar="#" escapeChar="!">
                        <ogc:PropertyName>dc:subject</ogc:PropertyName>
                        <ogc:Literal>Earth%</ogc:Literal>
                    </ogc:PropertyIsLike>
            </ogc:Filter>
        </csw:Constraint>
    </csw:Query>
</csw:GetRecords>
```

I'll be trying to do similar queries on `dc:contributor` and `dc:custodian` next.

sppigot:
First query is better written as:

``` xml
<csw:GetRecords xmlns:csw="http://www.opengis.net/cat/csw/2.0.2" version="2.0.2" service="CSW" resultType="results" startPosition="1" maxRecords="100">
  <csw:Query xmlns:ogc="http://www.opengis.net/ogc" xmlns:gml="http://www.opengis.net/gml" typeNames="csw:Record">
      <csw:ElementSetName>full</csw:ElementSetName>
      <csw:Constraint version="1.1.0">
        <ogc:Filter>
          <ogc:PropertyIsEqualTo>
            <ogc:PropertyName>dc:subject</ogc:PropertyName>
            <ogc:Literal>Earth Science | Oceans | Ocean Waves</ogc:Literal>
          </ogc:PropertyIsEqualTo>
        </ogc:Filter>
      </csw:Constraint>
  </csw:Query>
</csw:GetRecords>
```

kring:
Ah good point, and that works, thanks!

sppigot:
@kring Just rebuilt and added quite a few wave model layers to our catalogue. We're now using a thesaurus with a hierarchy of terms to classify the records into groups. The config for metadataGroups is below but have also extended the code to read a catalogue property (via a GetDomain call) that contains the hierarchical terms and generate the config for metadataGroups automatically. Will update the pull request and document shortly. The code is the same except that if the GetDomain request is enabled, the metadataGroups config is automatically built from the response and does not need to be specified in the aremi/nationalmap config. 

``` json
metadataGroups: [{
   "field":"subject",
   "value":"^Multiple Use\\ \\|\\ ",
   "regex":true,
   "group":"Multiple Use",
   "children":[{
      "field":"subject",
      "value":"Multiple Use | National",
      "regex":false,
      "group":"National"
    },{
      "field":"subject",
      "value":"Multiple Use | New South Wales",
      "regex":false,
      "group":"New South Wales"
    },{
      "field":"subject",
      "value":"Multiple Use | South Australia",
      "regex":false,
      "group":"South Australia"
    },{
      "field":"subject",
      "value":"Multiple Use | Victoria",
      "regex":false,
      "group":"Victoria"
    },{
      "field":"subject",
      "value":"Multiple Use | Western Australia",
      "regex":false,
      "group":"Western Australia"
   }]
  },{
   "field":"subject",
   "value":"^Wave Models\\ \\|\\ ",
   "regex":true,
   "group":"Wave Models",
   "children":[{
      "field":"subject",
      "value":"Wave Models | Direction of Maximum Directionally Resolved Wave Power",
      "regex":false,
      "group":"Direction of Maximum Directionally Resolved Wave Power"
   },{
      "field":"subject",
      "value":"Wave Models | Maximum Directionally Resolved Wave Power",
      "regex":false,
      "group":"Maximum Directionally Resolved Wave Power"
   },{
      "field":"subject",
      "value":"Wave Models | Wave Direction",
      "regex":false,
      "group":"Wave Direction"
   },{
      "field":"subject",
      "value":"Wave Models | Wave Directionality Coefficient",
      "regex":false,
      "group":"Wave Directionality Coefficient"
   },{
      "field":"subject",
      "value":"Wave Models | Wave Energy Flux",
      "regex":false,
      "group":"Wave Energy Flux"
   },{
      "field":"subject",
      "value":"Wave Models | Wave Height",
      "regex":false,
      "group":"Wave Height"
   },{
      "field":"subject",
      "value":"Wave Models | Wave Period",
      "regex":false,
      "group":"Wave Period"
   },{
      "field":"subject",
      "value":"Wave Models | Wave Spectral Width",
      "regex":false,
      "group":"Wave Spectral Width"
   }]
}]
```

kring:
Thanks @sppigot, that sounds like a major improvement to have the CSW server itself provide the grouping, rather than building it into the TerriaJS init file.

sppigot:
[AddingHierarchyToCSWSourcesInNationalMap-ver0.3.pdf](https://github.com/TerriaJS/terriajs/files/145883/AddingHierarchyToCSWSourcesInNationalMap-ver0.3.pdf)

kring:
Thanks @sppigot, sorry for the delay in merging this!

PR #1281, opened 2016-02-08
Handle multiple final linefeeds in csv files

### LINKED ISSUE #1279: Lat-lon csvs failing if lat or lon is ever undefined
Eg. using the latest master (TerriaJS 2.0.0), with AREMI, try to view Weather -> Solar Station Historical Observations.
It fails because `column.values[21] === undefined` for all columns.

RacingTadpole:
Some csv files have 3 or more final line feeds, which was causing terria to fail to read them, eg. Weather -> Solar Station Historical Observations.

For good measure, this PR also handles a missing lat or lon in a csv more gracefully.

Fixes #1279.


axman6:
@meh9 should probably also look into fixing the CSVs themselves

lib/Core/TableStructure.js
```diff
@@ -182,13 +182,17 @@ TableStructure.fromCsv = function(csvString, result) {
...
-        var last = json[json.length - 1];
-        if (last.length === 1 && last[0] === null) {
-            json.pop();
+    // Remove extra blank lines at the end.
+    function removeFinalBlank(json) {
+        if (json.length > 1) {
+            var last = json[json.length - 1];
+            if (last.length === 1 && last[0] === null) {
+                json.pop();
+                removeFinalBlank(json);
```
kring: JavaScript doesn't support tail recursion (I'm pretty sure), so this will blow out the stack if there are a lot of blank lines at the end.  Why not just use a loop and a single call to splice?

kring: I mean, it doesn't support optimizing tail recursion to not actually be recursion.

kring:

lib/Core/TableStructure.js
RacingTadpole: thanks - but are you saying recursion should be avoided in js always?

kring: No, recursion is ok.  It only needs to be avoided when the depth of recursion is controlled by the content of an external data file and is therefore potentially unbounded.

kring: I guess technically even our init files could trigger a stack overflow with crazy enough input, because they are recursive data structures.  But this seems like a case where it's easily avoided and results in both faster and clearer (IMO, admittedly) code.

PR #1282, opened 2016-02-09
Show "blank" values in CSV legends more nicely

### LINKED ISSUE #1265: Show "blank" values in CSV legends more nicely
- Change caption to "(No value)" perhaps, instead of "blank"
- Make this value either first or last in the legend list.
- Make this value optionally hideable through `tableStyle`

RacingTadpole:
Fixes #1265.
- Changes caption to `tableStyle.nullLabel` (defaulting to "") instead of "blank".
- Makes this value last in the legend list, shown both for ENUM and SCALAR types.

I also changed the nullColor to black, from dark blue, to distinguish it better from the other colours.

I didn't do the last dot point, "optionally hideable through tableStyle" - do you want that @stevage @meh9 ?

You can test most of this PR using the first catalog item on the AREMI site, switching to various display variables.


RacingTadpole:
This is a little nicer now: there is a small gap between the null color and the scalar colors.
Check it out using a link such as http://localhost:3001/#build/terriajs/test/init/csv.json

kring:
`http://localhost:3001/#build/TerriaJS/test/init/csv.json` for those of us with case-sensitive file systems :)

Some screen shots from Arthur's example init file:

image

image

Issue #1235, opened 2016-01-15
IE11 display issue with data catalogue


aherald:
On IE11, win 7, there is a display issue with the display of data catalogue.  A screen shot is attached

 Scenario
1.      User adds data from the catalogue
2.      Clicks on ‘Now Viewing’ and gets display as per screenshot

If the user then clicks on Search and back to ‘Now Viewing’, all displays as expected.

![datacataloguedisaplyissue](https://cloud.githubusercontent.com/assets/16713545/12344716/938444b2-bb94-11e5-9933-f0110cb7a43e.jpg)

stevage:
Fyi @chloeleichen if you make the commit message "Fixes #1235" then the issue closes automatically.

PR #1289, opened 2016-02-11
fix issue #1235

chloeleichen:


lib/ViewModels/SearchTabViewModel.js
```diff
@@ -49,8 +49,11 @@ var SearchTabViewModel = function(options) {
 
     // Focus the search box when the tab is activated.
     knockout.getObservable(this, 'isActive').subscribe(function(newValue) {
-        if (newValue) {
```
chloeleichen, line 52: This is probably not necessary but i added for consistency, maybe not a good idea?

PR #1290, opened 2016-02-11
Newui voldemort

chloeleichen:


RacingTadpole:
Thanks Chloe! I don't think I can handle this being called Voldemort in the code though... given that they are called Concepts everywhere else, can we either stick with Concept, or change all the references to Concept to an agreed-upon new name?

chloeleichen:
@RacingTadpole it's only temporary until we come up with a proper name

lib/ReactViews/NowViewing/Voldemort.jsx
```diff
@@ -0,0 +1,76 @@
...
+        return content;
+    },
+
+    render() {
+        return (<div className='voldemort'>
+                  <ul>{this.renderVoldemort()}</ul>
+                </div>);
+    }
+});
+module.exports = Voldemort;
```
RacingTadpole, line 111: I notice you're using radio buttons regardless of whether > 1 can be selected. Can you use checkboxes instead?
AbsConcepts (which are the top-level concepts for ABS catalog items) have an `isUnique` property that tells you if only 1 can be selected. DisplayVariableConcepts (which are the top-level concepts for csv files) have an `allowMultiple` property that tells you > 1 can be selected.  Hmmm... that's crazy. Let me rationalise this first.

lib/ReactViews/NowViewing/Voldemort.jsx
```diff
@@ -0,0 +1,76 @@
...
+    propTypes: {
+        nowViewingItem: React.PropTypes.object
+    },
+
+    toggleOpen(item) {
+        item.isOpen = !item.isOpen;
+    },
+
+    toggleActive(item) {
+        item.isActive = !item.isActive;
```
RacingTadpole: Don't set active items this way - selecting one doesn't unselect the old one, even when only one can be selected. Use `item.toggleActive()`.

lib/ReactViews/NowViewing/Voldemort.jsx
chloeleichen, line 111: @RacingTadpole the icons are also temporary for now, but i'll add an extra class indicating whether multiple items can be selected.

so i need to check both `isUnique` and `allowMultiple`? can we in the modal do something to unify this term?

RacingTadpole, line 111: No, I've just pushed through a change so they both use `allowMultiple` now.

RacingTadpole:
OK, @kring tells me he's ok with voldemort for now...

RacingTadpole:
Looking pretty cool! One detail: can you check each concept's `isVisible` property and only display it if true?

RacingTadpole:
In fact you might want to check all the logic in this section: https://github.com/TerriaJS/terriajs/blob/master/lib/Views/NowViewingTab.html#L2-L30 and make sure it works in with voldemort...

PR #1292, opened 2016-02-12
Add CkanCatalogItem

### LINKED ISSUE #1115: Populate metadata from CKAN, even when catalog item comes from elsewhere
It's sometimes useful to specify that a catalog item's metadata (e.g. its description) should come from a CKAN dataset, even if the catalog item itself is a direct reference to somewhere else (e.g. a WMS layer).  I can think of two ways to do this:
1. Add a `CkanCatalogItem` (currently we only have `CkanCatalogGroup`) to point to the CKAN resource.  TerriaJS would query the CKAN resource to find out what type of thing it really is, and then access it as WMS, GeoJSON, etc. depending on the type of resource.  This is a bit tricky in our object model, but makes some sense conceptually.
2. Give `CatalogItem` a `ckanResourceUrl` property.  Any catalog item with this property defined would query CKAN for metadata, but would otherwise be a normal catalog item of some other type (e.g. `WebMapServiceCatalogItem`).

kring:
Fixes #1115 

Adds a `CkanCatalogItem` that is configured with a CKAN URL and a `resourceID`, `datasetID`, or both.  When loaded, it queries CKAN to find a resource based on the provided information and then creates a new catalog item (e.g. a `WebMapServiceCatalogItem` or a `GeoJsonCatalogItem`) to access that resource.  Through some feats of semi-hackery, that new catalog item gets added to the Now Viewing instead of the `CkanCatalogItem`.

The semi-hackery is general enough to be used for other types of resources-from-catalogs, e.g. to create a `CswCatalogItem`.


kring:
I'm still going to write some tests for this, so don't merge yet.  I'm just opening it for early review.

RacingTadpole:
A nitpicky initial comment while I get into this... I've used `datasetId` in ABS (deprecating the old `dataSetID`)... I see you've introduced `datasetID` - would it be worth changing this to `datasetId` to keep notation consistent across catalog items?
(I guess `resourceID` -> `resourceId` too?)

Issue #1267, opened 2016-02-04
Layer hiding behaviour with region mapped data


tephenavies:
There are a couple of bugs regarding layer hiding:
- [x] A layer hidden by clicking the eye can be re-shown by changing selected variables (most likely intended) but this doesn't change the eye icon
- [ ] A layer can with no variable selected still displays data (based on the last variable that was selected) but has no legend

tephenavies:
The second problem seems to be a logic error in [`AbsIttCatalogItem.js: rebuildData`](https://github.com/TerriaJS/terriajs/blob/16caf395f0722f4729187705e653f222a09794a3/lib/Models/AbsIttCatalogItem.js#L440-L452). When there are no variables selected, `tableStructuresAndCombinations` is `undefined`. This means that `regionDetails` will also be `undefined`, so the imagery is not updated.

tephenavies:
I think this is also related:

When I add [this file](https://github.com/TerriaJS/terriajs/files/147677/demo.txt) as a csv and deselect the `Statistic` column, the layer is redrawn with different colours.

tephenavies:
What should the behaviour be when you try to deselect the only selected concept of a certain concept type? Should TerriaJS:
A. Disallow deselecting
B. Deselect the concept and hide the layer (meaning you would have to select a concept **and click the show button** to reshow it), or
C. Deselect it and not display the layer (probably the most obvious option UI-wise, but it would complicate the possible states that a layer could be in – i.e. shown, hidden by lack of concept selection but with attribute `isShown` still true, hidden by hide button/errors & basemap `isShown` false but not hidden)

RacingTadpole:
Sorry - the second checkbox in your issue description might be covered by issue https://github.com/TerriaJS/terriajs/issues/791, which is solved via https://github.com/TerriaJS/terriajs/pull/1338 (an outstanding PR).

stevage:
Or, as in that PR I think:

D. Deselect it, and now the layer displays, using a default color instead of choroplething.

RacingTadpole:
Yep, good point - D is the current behaviour for region-mapped csvs when no column is active. Also no legend is displayed in that case. As I understand it, @steve9164 has uncovered two issues here:
- ABS items don't behave well when you have nothing selected for a particular concept type
- a region-mapped csv file with only a region column doesn't load.

I recommend we close this issue and open two separate issues for these.

PR #1294, opened 2016-02-12
Make eye icon change when hidden layer is shown via a change in variable selection

tephenavies:
A hidden layer is displayed when you change the concept selection, however the underlying item's `isShown` property is not updated. This change fixes that behaviour so that the layer is displayed and the `isShown` property is set to true (which in turn changes the eye icon in the Now Viewing tab).

Fixes the first point in #1267.

### DIFF
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -623,7 +623,7 @@ ImageryLayerCatalogItem.showLayer = function(catalogItem, layer) {
         return;
     }
 
-    catalogItem.isShown = true;
+    if (catalogItem.isEnabled) catalogItem.isShown = true;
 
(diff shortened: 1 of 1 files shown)


kring:
Wouldn't it be better to put a check in `redisplayRegions` in `RegionMapping.js` to only call `showLayer` if `regionMapping._catalogItem.isShown` is true?  That way toggling the display variable wouldn't change the shown state at all.

tephenavies:
That is another solution, but is there any time that you would want to change the variables while keeping the layer hidden?

kring:
I see them as independent things.  Just like we don't flip Show back on when Opacity changes, we shouldn't do it here either.

steve9164:
Toggling display variables now doesn't change the `isShown` state or redisplay a hidden layer.

PR #1298, opened 2016-02-15
Escaped commas in exported CSVs.

### LINKED ISSUE #1287: Feature CSV Generation Doesn't Escape ","
Just realised that the feature CSVs don't include quotes in between the commas to escape commas that are in the normal markup.

AlexGilleran:
Fixes #1287.

Pretty simple, CSVs were previously naively generated by `.join`ing values together with commas in between, this wraps quotes around the values and escapes quotes by doubling them as per CSV conventions.


RacingTadpole:
This looks good. Can you update CHANGES.md?

PR #1302, opened 2016-02-17
Implement per-column table styling

### LINKED ISSUE #1097: Per-column CSV table styling
Something in the back of my mind for a while. Our init file format only lets you specify one style, which applies to one column. So when the user selects a different column in the interface, Terria is flying blind.

We should perhaps have something like:

```
tableStyle: {
  columns: {
    lga: {
      regionType: "LGA_NAME",
    }
    pop: {
      colorMap: "black-red",
      title: "Population"
    }
    migration: {
      colorMap: "red-black-blue",
      title: "Net migration",
      colorBinMethod: "quantile"
    }
    type: {
      colorMap: "Set3",
      title: "LGA type",
      varType: "ENUM"
    }
  },
  dataVariable: "pop"
}
```

RacingTadpole:
Fixes #1097. 

Columns can be specified by either their name in the csv file, or their position (0 for the first column, etc).  I've added `name` and `type` to the list of tableStyle parameters that apply in `tableStyle.columns`.

An example is the final catalog item at http://localhost:3001/#build/terriajs/test/init/test-tablestyle.json .

One implementation decision I've made, which we can discuss, is: if you supply a new `name` for a column, but not a `type`, its type is guessed based on the _new_ name, not its old name.  

Unsupported in this PR:
- Setting `columns.x.type` to `region`
- `columns.x.regionType`
- Columns cannot be specified by regex.

### DIFF
--- a/lib/Models/TableColumnStyle.js
+++ b/lib/Models/TableColumnStyle.js
@@ -8,6 +8,7 @@ var freezeObject = require('terriajs-cesium/Source/Core/freezeObject');
 var ColorMap = require('../Map/ColorMap');
 var serializeToJson = require('../Core/serializeToJson');
 var updateFromJson = require('../Core/updateFromJson');
+var VarType = require('../Map/VarType');
 
 /**
  * A set of properties that define how a table column should be displayed.
@@ -143,10 +144,28 @@ var TableColumnStyle = function(options) {
     this.legendTicks = defaultValue(options.legendTicks, 3);
 
     /**
-     * A nice name for this column.
+     * Display name for this column.
      * @type {String}
      */
     this.name = options.name;
+
+    /**
+     * The variable type of this column. Convert strings, which are case-insensitive keys of VarType, to numbers. See VarType for further information.
+     * @type {String|Number}
+     */
+    this.type = undefined;
+    if (defined(options.type)) {
+        var typeNumber = parseInt(options.type, 10);
+        if (typeNumber === typeNumber) {
+            this.type = typeNumber;
+        } else {
+            for (var varTypeName in VarType) {
+                if (options.type.toLowerCase() === varTypeName.toLowerCase()) {
+                    this.type = VarType[varTypeName];
+                }
+            }
+        }
+    }
 };
 
 // When colorMap is updated, we need to convert it to a colorMap.
--- a/lib/Models/TableColumnStyle.js
+++ b/lib/Models/TableColumnStyle.js
@@ -18,6 +18,29 @@ var updateFromJson = require('../Core/updateFromJson');
  * @constructor
  *
  * @param {Object} [options] The values of the properties of the new instance.
+ * @param {Float} [options.minDisplayValue] All data values less than or equal to this are considered equal for the purpose of display.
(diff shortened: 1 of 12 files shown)


lib/Models/TableColumnStyle.js
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * A set of properties that define how a table column should be displayed.
+ * If not set explicitly, many of these properties will be given default or guessed values elsewhere,
+ * such as in CsvCatalogItem.
+ *
+ * @alias TableColumnStyle
+ * @constructor
+ *
+ * @param {Object} [options] The values of the properties of the new instance.
```
AlexGilleran, line 20: Should this have the values of the potential options like in other constructors that take options hashes?

lib/Models/TableStyle.js
```diff
@@ -37,163 +46,53 @@ var TableStyle = function(options) {
...
-    },
-    colorPalette: function(tableStyle, json, propertyName) {
-        return ColorMap.loadFromPalette(json[propertyName]).then(function(colorMap) {
-            tableStyle.colorMap = colorMap;
-        });
-    }
+TableStyle.prototype.updaters = clone(TableStyle.prototype.updaters);
+TableStyle.prototype.updaters['columns'] = function(tableStyle, json, propertyName) {
+    // Do not update columns here; do it directly in the updateFromJson function.
+    // tableStyle.columns = objectToTableColumnStyle(json[propertyName]);
```
AlexGilleran: If this isn't updating columns does it still need the commented out line?  Or this updater at all?

lib/Models/TableColumnStyle.js
RacingTadpole, line 20: Very true. I just copied it over from TableStyle, which didn't have it either. But I'll add it.

lib/Models/TableStyle.js
RacingTadpole: My idea is, by including an empty function here, lib/Core/updateFromJson.js calls this (and does nothing) instead of doing `target[propertyName] = json[propertyName];` , ie. it does not set `columns`.  
Instead, columns is set in the preceding function, the tableStyle's own `updateFromJson`.
I did it this way because TableColumnStyle's colorPalette actually returns a promise. The updaters can't handle a promise, but updateFromJson can.
Of course, I settled on this admittedly strange pattern after all the other things I tried failed...

RacingTadpole: Maybe a better approach would be make colorPalette behave more normally? I don't know if that's possible.

AlexGilleran: Ah cool, that makes sense - maybe add a comment noting that the noop is intentionally included? And remove the commented line of code unless you're gonna need it later? :grimacing:

AlexGilleran:
So cool :rainbow: .

As discussed on slack though:

```
 "columns": {
               "Value": {
                 "name": "Postal code"
               },
}
```

results in:

image

Probably should say "Postal code" not "Value" :).

RacingTadpole:
@AlexGilleran: I've added your suggestions now. I've also built on this quite a bit in https://github.com/TerriaJS/terriajs/pull/1307, and I might have cleaned up some things there that really should be here... sorry...

Issue #1172, opened 2015-12-15
Commas in numbers in the Feature Info panel


RacingTadpole:
In the Feature Info panel, if there is no template given, a large number is formatted with commas (eg. so `1234567` is shown as `1,234,567`). But if a template is given, the number is left as `1234567`. We should rationalise this.

RacingTadpole:
I think the default should be the same as with no template, ie. numbers are formatted with commas.

One way we could do this is to define some new properties in the feature info panel, but only if a template is defined.  

E.g. if `amount` is a number (say 1234567), then we change it to the string `"1,234,567"`, and add a new property `"amount | noCommas"` = `1234567`.

Is that too much of a hack? I guess it immediately lends itself to adding more such, eg. "`| isArray`"...

kring:
A less hacky (but more verbose) approach is to do something like this:

```
{{#nocommas}}{{amount}}{{/nocommas}}
```

Where `nocommas` is a mustache lambda function.  One nice thing about this is you can wrap it around your entire template and the entire thing will be rendered without commas.

kring:
The folks at ABR would also like to see the ability to format numbers without commas (e.g. ABNs).  Also, they noted an inconsistency between how numbers are formatted on the feature info panel (commas are added) and how they're formatted when they appear in the legend (commas are not added).

kring:
Oh, they also found a workaround for preventing TerriaJS from adding commas to numbers: suffix the value with `&nbsp;`.  Unfortunately, if the numbers do appear in the legend, the `&nbsp;` is visible, e.g:

![image](https://cloud.githubusercontent.com/assets/924374/13064341/ebba6c36-d4a5-11e5-8fc2-3c0121c3df02.png)

RacingTadpole:
Interesting! Note in the current release there will be commas in the legend - although that's not the direction of consistency the ABR wants...
It should be easy enough to add the `{{#nocommas}}` syntax.

kring:
My screen shot above was taking with the version of NM we released yesterday, for what it's worth.

RacingTadpole:
Do you know if the ABR can set up a json file with `featureInfoTemplate`?

kring:
@RacingTadpole I don't see why they couldn't...  I'll suggest that.

PR #1303, opened 2016-02-17
Optionally format numbers in the Feature Info panel according to user locale

RacingTadpole:
Implements a template tag which formats the number according to the user's locale, if available. 
Reverses the default so that, without this tag, numbers are shown without commas. This was chosen because frequently numbers are IDs or phone numbers.

Eg. `{{#localeFormat}}1234567.8{{/localeFormat}}` => `1,234,567.8`.

Addresses #1172. It does not quite fix it, as to be consistent, the same behaviour should apply to the legend. This will be the subject of an upcoming PR.


RacingTadpole:
Actually I have a better idea, based on #1302 - set the formatting directly in tableStyle, where it can be used by both the Feature Info Panel and the Legend (and any future functionality).

PR #1304, opened 2016-02-17
Typo and options in GeoJsonCatalogItem

chloeleichen:


AlexGilleran:
+1 but I've got no idea how to test this, when I change `marker-size` in geojson it doesn't look any different even if I change the it to "large".

chloeleichen:
@AlexGilleran try to change it to 50 then to 200 maybe?

AlexGilleran:
Didn't seem to change the look of the point at all - this was with the GeoJSON test file that Steve made. Not sure what is supposed to happen though :/. Code looks fine.

stevage:
Thanks @AlexGilleran, definitely something wrong here, I'll investigate. Looks like colour and marker size aren't making it through.

stevage:
Sigh. Our test file for "unstyled" features actually had styles in them! Argh. Once I remove that, then applying styles in the init file works as expected.

screenshot 2016-02-23 12 08 35

I suspect I made a bad choice when I originally decided that the styles in the GeoJSON feature should override those in the init file. Maybe it should be the other way around. Thoughts?

PR #1307, opened 2016-02-18
Per-column number formatting

### LINKED ISSUE #1172: Commas in numbers in the Feature Info panel

RacingTadpole:
Builds on #1302.

Adds the column-specific `tableStyle` parameter `format` to set the column number format. This applies in both the Feature Info Panel and the Legend.

Eg. 

```
"tableStyle": {
    "format": {
        "useGrouping": true,
        "maximumFractionDigits": 2
    }
}
```

to add thousand separators to numbers and show only two decimal places. The two attributes above, plus `"style": "percent"`, are guaranteed to work on all browsers. In most browsers (Safari being the most notable exception), all the properties of the [Javascript Intl options parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) are supported. Note the thousand separator is now locale-specific.

`"style": "percent"` shows eg. `0.855` as `85.5%`.

An example is the final catalog item at http://localhost:3001/#build/terriajs/test/init/test-tablestyle.json .

Fixes #1172.
Replaces #1303.


test/Models/CsvCatalogItemSpec.js
```diff
@@ -437,6 +437,112 @@ describe('CsvCatalogItem with lat and lon', function() {
...
+                        replaceWithNullValues: ['bad'],
+                        nullColor: '#A0B0C0',
+                        name: 'Temperature'
+                    }
+                }
+            });
+            var nullColor = new Color(160/255, 176/255, 192/255, 1);
+            csvItem.load().then(function() {
+                expect(csvItem.tableStructure.columns[2].name).toEqual('Temperature');
+                function cval(i) { return csvItem.dataSource.entities.values[i]._point._color._value; }
```
AlexGilleran: Was this supposed to be separated out? Seems a bit weird to be defining this as a function (twice) when you could just go `csvItem.dataSource.entities.values[1]._point._color._value`. Not a big deal though.

RacingTadpole: yeah, good point. I'll rationalise those unnecessary functions.

AlexGilleran:
:+1:  apart from that, looks awesome :D

RacingTadpole:
Fixed that funny issue with `missingNumberFormatting.csv` - the issue was that the latitudes were +30, not -30, so we were looking in the wrong spot...

AlexGilleran:
Looks good but you've fallen foul of the merge again :(

RacingTadpole:
OK, all merged in...

PR #1310, opened 2016-02-18
Update null-value table styling defaults

### LINKED ISSUE #1081: CSV column type detection is not robust and not coercive
I ran into an issue today with the SEIFA test dataset. A single row out of 2483 had the value "-" instead of a number, which broke a new choroplething function I'm using. `DataTable` had happily marked the column as `VarType.SCALAR` because it passed this trivial test:

```
    if (this.varType === VarType.SCALAR && this.minVal > this.maxVal) {
        this.varType = VarType.ENUM;
    }
```

It would be pretty easy to more robustly check that every value is actually a number, but:
1. That would seriously slow down the loading of large datasets
2. In cases like this, it would throw out the baby with the bathwater, by treating 2482 rows as enums instead of scalars, because of one bad egg.

I guess the solution is to somehow combine type detection with type coercion: once a column has been marked as `VarType.SCALAR` it should never return non-scalars. But since code tends to access the raw array (`DataVariable.vals`) I'm not sure how one can practically implement that without a serious performance penalty.

Thoughts?

RacingTadpole:
Changes these table style defaults, as discussed with @meh9:
- `tableStyle.replaceWithZeroValues` to `[]`, ie. nothing will be changed to 0 by default.
- `tableStyle.replaceWithNullValues` to `["-", "na", "NA"]`, so that "-" is treated as null now.
- `tableStyle.nullLabel` to '(No value)', instead of ''.

Fixes #1081.


AlexGilleran:
:+1: but needs a merge/rebase.

RacingTadpole:
Merged with master, thanks.

PR #1311, opened 2016-02-18
Made showWarnings flag for CSV/ABSITT items work again

### LINKED ISSUE #1309: CsvCatalogItem showWarnings is no longer being checked
There is a regression where the `CsvCatalogItem` `showWarnings` is no longer being checked. As a result, region mappings that fail will show a warning popup even when `showWarnings` has been set to `false`.

To test turn on the AREMI layer `Electricity Inf. -> Generation -> Small Scale Installations -> Air Source Heat Pumps` and a warning message saying that it couldn't match a number of regions will appear.

AlexGilleran:
Fixes #1309

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 ### 2.0.2
 
 * Fixed a bug that caused the content on the feature info panel to be rendered as pure HTML instead of as mixed HTML / Markdown.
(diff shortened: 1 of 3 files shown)


kring:
Update CHANGES.md please! :)

RacingTadpole:
Yep, this looks good - just Kevin's comment above.

AlexGilleran:
image

PR #1312, opened 2016-02-19
Rationalise the core, map and model directories

RacingTadpole:
Unfortunately I discovered a few places that so-called `Core` modules were requiring things from `Map` and even `Models`. This fixes that.

Builds on #1307, which I thought would be a good idea to save merging later, but is actually poor, because now this PR depends on that one.
- Moved `TableColumn`, `TableStructure`, and the classes based on `Concept` to `lib/Map`.
- Moved `LegendHelper` to `lib/Models`.


AlexGilleran:
Looks good, really needs a merge though because now the diff with master is actually pretty small ;).

RacingTadpole:
thanks, done.

PR #1314, opened 2016-02-19
Get app name and support email from config parameters, not Terria opt…

stevage:
…ions.


AlexGilleran:
:+1: but needs a merge

stevage:
Ok now @AlexGilleran ?

PR #1319, opened 2016-02-22
Fix region mapped layer becoming unhidden when changing concepts

tephenavies:
Continuing on from #1294 (which I can't reopen because I force pushed t the branch).

Toggling display variables now doesn't change the isShown state or redisplay a hidden layer.

Fixes the first point in #1267.


AlexGilleran:
:+1:, needs a merge :)

PR #1323, opened 2016-02-23
Alert user when a WMS layer doesn't match what's on the server.

stevage:
Fixes NICTA/NationalMap#196.


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -319,18 +319,24 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
...
+                        console.log('A layer with the name or ID \"' + layers[i] + '\" does not exist on the WMS Server - ignoring it.');
+                        thisLayer.splice(i, 1);
+                        layers.splice(i, 1);
+                        --i;
+                    } else {
+                        throw new TerriaError({
+                            title: 'No layer found',
+                            message: 'The WMS dataset "' + this.name + '" has no layers matching "' + this.layers + '". (Perhaps it should be "' + capabilities.Capability.Layer.Layer.Name + '").' + 
+                            '\n\nEither the catalog file has been set up incorrectly, or the WMS server has changed.' + 
+                            '\n\nPlease report this error by sending an email to <a href="mailto:' + this.terria.supportEmail + '">' + this.terria.supportEmail + '</a>.'
```
AlexGilleran, line 339: This ends up in HTML, so these will have to be `<br>`s right?

stevage, line 339: screenshot 2016-02-24 12 02 28

You'd think so, but apparently not. We must do a replacement somewhere.

kring, line 339: Markdown FTW!

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -319,18 +319,24 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
...
-                    --i;
+                    if (thisLayer.length > 1) {
+                        console.log('A layer with the name or ID \"' + layers[i] + '\" does not exist on the WMS Server - ignoring it.');
+                        thisLayer.splice(i, 1);
+                        layers.splice(i, 1);
+                        --i;
+                    } else {
+                        throw new TerriaError({
+                            title: 'No layer found',
+                            message: 'The WMS dataset "' + this.name + '" has no layers matching "' + this.layers + '". (Perhaps it should be "' + capabilities.Capability.Layer.Layer.Name + '").' +
```
kring: The `capabilities.Capability.Layer.Layer.Name` bit is dangerous.  It's possible (likely?) that one of those dotted parts will evaluate to undefined, and if it does NM will display a really ugly message instead of the nice one you're trying to generate.

kring:
Looks good other than the one comment.  Oh and update CHANGES.md too please.

lib/Models/WebMapServiceCatalogItem.js
stevage: Yeah true. Ugh. am I really going to have to write `capabilities && capabilities.Capability && ...` ?

kring: Yeah, sorry :(

You should propose a new feature for ES2017 that lets you evaluate a big long expression to undefined if any part along the way evaluates to undefined :)

stevage:
Just confirming that we think the build failure is spurious, so please review when ready @kring

PR #1328, opened 2016-02-24
Skip time intervals with no data

### LINKED ISSUE #896: Support timelines without specific years/months/days
Some layers that we have to show (currently WMS and CZML) have timeseries data that doesn't really work with the current time control.

In one case we have a dataset that has been averaged down to an arbitrary 24 hours - i.e. there isn't really a year/month/day attached to the 24 hours. However, in order to show it in Terria we have add a date to those 24 hours, and that date is part of what is displayed on the time slider. This gives the incorrect notion that the 24 hours are on some specific date, when we need the time slider to display just hours. The way to solve this one would probably be to specify that no date should display on the time slider.

In another case, we have the new Bureau of Meteorology Solar Climatology WMS services, e.g.:

http://neii.bom.gov.au/services/solarclim/wms/data/monClim_gloHorExp1Hou.nc?service=WMS&request=GetCapabilities

This has 12 months of 24 average hours per month, i.e. many years of 24 hour days for January have all been averaged down to 24 hours representing an average January day, and so on. 

In order to get a WMS time dimension that works, they have put the arbitrary year of 1993 on all timestamps, and the 24 hours span from the 15th to the 16th of the month. There is no data or timestamps outside of the 15/16 days of each month:

```
1993-01-15T16:00:00.000Z, 1993-01-15T17:00:00.000Z, [...], 1993-01-16T15:00:00.000Z,
1993-02-15T16:00:00.000Z, 1993-02-15T17:00:00.000Z, [...], 1993-02-16T15:00:00.000Z,
[...]
```

In this case, we need to specify that the time slider should not display the year (so this probably generalises together with not showing a date at all with the other case given above in this issue), and it should somehow put the 24 hours for each month next to each other on the time slider instead of what it does currently, which is leaving 30 days of no data between the single 24 hours of data for each month.

Showing data with specific timestamps, like specific individual datapoint observations, is all well and good and really useful. But the real utility of using Big Data comes from analysing that data, and aggregating it down into useful abstractions like averages etc. It is when displaying these aggregated processed versions that we need to be able to display the times that these aggregations are concerned with in a non-time-instant way.

### LINKED ISSUE #946: Snap timeline scrubber to closest datapoint
When dealing with time series data, the timeline scrubber should snap to the closest datapoint on click-drag-release. This should help to communicate the granularity of the data currently being displayed.

AlexGilleran:
Fixes #896 
Fixes #946 

Extends `Clock` to override `currentTime` setter when the clock is representing the timeline of a `CatalogItem` with intervals. Uses the `TimeIntervalCollection` binary search to determine if the new time will be in a gap for which that `CatalogItem` has no data, and if so advances it to the start of the next interval. This means that as time advances, every tick this check is made, so the timeline on the screen jumps through areas of no data.

This requires some changes to the timeline to allow configurable date formats. This was done specifically for use cases like a TMY, in which the year and date are moot (and misleading) but the month and time are very important. They look like: 

```
"dateFormat": {
    "timelineTic": <format that appears on the timeline itself>
    "currentTime": <format that appears to the left of the timeline>
},
```

There's also an extra config field, `defaultDuration`, that is used to determine what duration an interval should be if there's no end specified by the WMS server.

There's still some debate around how this should be represented (see discussion in #896) in the UI - this branch simply darkens regions with no data.

https://cloud.githubusercontent.com/assets/900555/13208992/d8b6da30-d972-11e5-849f-966bbe61cabe.png

Should probably wait for input from @cam-grant before merging.


AlexGilleran:
- Now works for CSVs too
- All durations are `displayDuration` not `defaultDuration`

AlexGilleran:
Updated with latest from `master`.

AlexGilleran:
Updated from `master` again :)

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -109,6 +109,8 @@ var WebMapServiceCatalogItem = function(terria) {
      */
     this.maxRefreshIntervals = 1000;
 
+    this.displayDuration = undefined;
```
kring, line 112: Can you add doc for this?  It's especially useful because it gets extracted from the source code and used in the data source editor UI.

PR #1332, opened 2016-02-24
Use UriJS to join URLs, to support a baseUrl of "/".

stevage:


stevage:
Useful for City of Sydney.

PR #1337, opened 2016-02-25
Improve UX of drag-dropping unsupported file types.

### LINKED ISSUE #560: Make supported filetypes clearer
I wanted to test the "add data" function, so attempted to upload a shapefile (.shp). It seems like a reasonable thing to try to do (although probably it wouldn't work because you need the auxiliar files, hence zipped shapefiles work better.)

The current UX isn't great:

screenshot 2015-03-02 12 26 20

Followed by:

> An error occurred while attempting to convert this file to GeoJson. This may indicate that the file is invalid or that it is not supported by the National Map conversion service. If you would like assistance or further information, please email us at nationalmap@lists.nicta.com.au.

Suggest:
- including a list of supported file types and protocols in the "add data" dialog
- skipping the first dialog (IMHO the average user is going to always click "OK" because we've just told them it _might_ work). 
- add filetype detection even for non-supported filetypes, in order to give a more specific message (eg, "Shapefiles aren't supported by NationalMap. Try uploading a GeoJSON or a KML file.", and even better (if possible), aborting the upload at this stage.
- remove references to "National Map conversion service". I don't think the user needs to know there is a "conversion service" - there's just National Map.
- distinguish between the two causes of failure: unsupported file type or invalid file.
- hence perhaps reword the second dialog as: "We couldn't process the file [filename]. National Map supports GeoJSON, KML, [...]." or "We couldn't process the file [filename] - it might be corrupt or incompatible with National Map. For assistance, try nationalmap@lists.nicta.com.au"

stevage:
Specifically:
- reject known bad file types immediately, with a helpful message
- show that message using a nicer window than the JS alert()
- provide a more readable question about the conversion service
- if user declines the conversion service, don't show an extra error message afterwards

Fixes #560.


stevage:
upload

stevage:
I'm not sure if that's a real build failure or not. If it's not, could someone please review? :)

lib/Models/createCatalogItemFromFileOrUrl.js
```diff
@@ -30,47 +32,65 @@ var createCatalogItemFromFileOrUrl = function(terria, fileOrUrl, dataType, confi
...
-    } else if (dataType === 'other') {
-        if (defaultValue(confirmConversion, true) && !confirm('\
-In order to convert this file to a format supported by '+terria.appName+', it must be uploaded to the '+terria.appName+' conversion service. \
-Click OK to upload the file to the '+terria.appName+' conversion service now.  Or, click Cancel \
-and the file will not be uploaded or added to the map.')) {
-            return when.reject('The user declined to use the conversion service.');
-        }
-
-        newCatalogItem = createCatalogMemberFromType('ogr', terria);
+    if (name.match(/\.(shp|jpg|jpeg|pdf|xlsx|xls|tif|tiff|png|txt|doc|docx)$/)) {
```
kring: Would it make sense to flip this logic a bit?  That is, first call `createCatalogItemFromUrl` before checking the extension.  If that fails (i.e. returns `ogr`), _then_ check for some common extensions that definitely won't work with the conversion service before prompting the user to invoke it.

The reason is that the set of supported formats is a plugin point in TerriaJS.  An app can add new ones without changing our library.  And at least one of the types in your reject list (TIFF) is a sensible geospatial format that some apps might actually want to support.  So we should let user-defined handlers take precedence over a hard-coded reject list.

kring:
Looks great other than the one comment!

lib/Models/createCatalogItemFromFileOrUrl.js
stevage: Yeah, nice one.

lib/Models/createCatalogItemFromFileOrUrl.js
```diff
@@ -30,47 +32,68 @@ var createCatalogItemFromFileOrUrl = function(terria, fileOrUrl, dataType, confi
...
     } else {
         newCatalogItem.data = fileOrUrl;
         newCatalogItem.dataSourceUrl = fileOrUrl.name;
     }
 
-    return newCatalogItem.load().then(function() {
-        return newCatalogItem;
-    });
-};
+    return newCatalogItem.load().yield(newCatalogItem);
```
RacingTadpole, line 99: Is `yield` supported in all browsers? https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/yield suggests IE and Safari might be questionable (although it works in my Safari). Can you add a spec which tests that `loadItem` works? Then we could be sure.

lib/Models/createCatalogItemFromFileOrUrl.js
```diff
@@ -30,47 +32,68 @@ var createCatalogItemFromFileOrUrl = function(terria, fileOrUrl, dataType, confi
...
+        return when(true);
+    
+    var d = when.defer(); // there's no `when.promise(resolver)` in when 1.7.1
+    PopupMessageConfirmationViewModel.open('ui', { 
+        confirmText: 'Upload', 
+        title: 'Use conversion service?',
+        message: message,
+        confirmAction: d.resolve,
+        enableDeny: true,
+        denyAction: function() { this.close(); } // we don't call resolve, so just leave the promise hanging?
```
RacingTadpole: Good question (when the popup is cancelled, "we don't call resolve so just leave the promise hanging?") - can I refer this to @kring?

RacingTadpole: Actually could you resolve it with a true/false value, and then take action accordingly?

lib/Models/createCatalogItemFromFileOrUrl.js
kring, line 99: That's not the yield operator, it's the yield function, provided by when.js.  Should work everywhere.

lib/Models/createCatalogItemFromFileOrUrl.js
kring: Why not reject the promise?  `d.reject();`

lib/Models/createCatalogItemFromFileOrUrl.js
```diff
@@ -30,47 +32,68 @@ var createCatalogItemFromFileOrUrl = function(terria, fileOrUrl, dataType, confi
...
+    });
+    return d;
```
kring: This should be `return d.promise;`

lib/Models/createCatalogItemFromFileOrUrl.js
stevage, line 99: Yeah. when.yield is basically defined as doing the thing I replaced with it.

lib/Models/createCatalogItemFromFileOrUrl.js
stevage: Yes. (because `deferred.then` is deprecated)

lib/Models/createCatalogItemFromFileOrUrl.js
stevage: @RacingTadpole's approach is better. A user abort is not an exception (and if we just did that it would throw an error message).

stevage:
Ok, resolved all those. @kring

RacingTadpole:
Looks good to me!

kring:
:+1:

PR #1338, opened 2016-02-25
Shows lat-lon and region-mapped CSVs even if no data variable, improves their coloring

### LINKED ISSUE #791: Allow lat-lon CSVs to be shown with no data variable (all points the same color)
I'm putting this in for CSV files that have lat/lons but it probably applies more broadly. At the moment, it seems that there needs to be at least one variable checked.  However, for some csvs, it only makes sense to show points for the lat/lons and not use different colours for different values in the value range. At the moment, you can't do this from the UI.  It would be good to be able to have all checked off (actually this is probably a good default for CSVs which have not had an initial set of variable states specified)
Eg see csv file of bus stops at http://data.gov.au/dataset/bus-stops-brisbane-city-council

RacingTadpole:
Fixes #791 and #1356.
Sample file in `wwwroot/test/csv/lat_lon.csv`.


stevage:
Hmm, there doesn't seem to be a legend generated.

screenshot 2016-02-29 11 44 08

Also just pointing out an inconsistency: when you load a GeoJSON file of points with no style information, we [generate a random color based on the dataset name](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/GeoJsonCatalogItem.js#L590). But for CSVs we always use grey?

I'm not sure that either behaviour is ideal, to be honest. Probably the right thing would be to keep a global count of the number of unstyled datasets, and assign a new colour to each new dataset in order from a palette.

kring:
How about a random color from a palette based on the dataset name?  That avoids any non-determinism based on load order.

RacingTadpole:
Good ideas, thanks. 

FYI, it's currently using the `nullColor` to display the points if there's no value, which is settable in via `tableStyle`; I was mentally equating "no column" to "no value". I think it's fine to break that association. Let me know if you can think of any reason to keep it. The only downside I can see is that you could no longer set the colour using `tableStyle.nullColor`. But we could always add a new parameter.

I'll align geojson to draw from the same palette.

Re the legend, good point; for a single dataset it didn't seem necessary, but we'd certainly need it if there are multiple no-value-column datasets open, each using different colours.

stevage:
> How about a random color from a palette based on the dataset name? That avoids any non-determinism based on load order.

Yeah, that's what the GeoJSON does. Things I don't like about this approach:
- there's no telling what the generated color will be. Sometimes it turns out to be very close to something else (for instance, to a default color), and that's confusing. (Well, it has confused me in the past)
- having the color dependent on name can be incredibly confusing when say, you change the name of a file, and load it again - and now it's a different color. This has also confused me in the past - and in both cases, I was already aware of this behaviour, although it wasn't top of mind.

So, I think the behaviour that "the first thing I load is blue, and the next one is red" is much more obvious to the user than "things called `BlahBlah.csv` are a kind of murky green, but things called `BlahBlah2.csv` are a purplish scarlet".

stevage:
> Re the legend, good point; for a single dataset it didn't seem necessary, but we'd certainly need it if there are multiple no-value-column datasets open, each using different colours.

I think it's still necessary. There are grey dots on the map, so the legend should say what grey dots mean.

kring:
> there's no telling what the generated color will be. Sometimes it turns out to be very close to something else (for instance, to a default color), and that's confusing.

If the name forms a hash into a carefully selected list of colors, we wouldn't have the "very close to something else" problem any more with this approach than we would with sequential mapping, right?

> So, I think the behaviour that "the first thing I load is blue, and the next one is red" is much more obvious to the user than "things called BlahBlah.csv are a kind of murky green, but things called BlahBlah2.csv are a purplish scarlet".

Ok, but just watch out for things changing colors after they're shared and that sort of thing.

RacingTadpole:
@kring - just to be clear - when you say palette above, do you mean choose a colour from a defined list of allowed colours, eg. from `['#8dd3c7', '#ffffb3', '#bebada']` (though in practice we'd want many more in the list) ?
In contrast the geojson code defines a "palette" as an allowed range of R,B and G values, https://github.com/TerriaJS/terriajs/blob/master/lib/Models/GeoJsonCatalogItem.js#L49-L57.

The only problem with the list-of-allowed-colours approach is that even with as many as 12 colours in the list, there is an 8% chance of two files using the same colour. So I wonder if the existing geojson approach might actually be best.

kring:
I mean the former, choose a color from a list.  Ideally we'd have a pretty large list (2-3 dozen colors?) to avoid collisions.

RacingTadpole:
OK, this is done. I've used Kelly's 22 maximum-contrast colors http://www.iscc.org/pdf/PC54_1724_001.pdf as the palette, and made geojson and csv use the same set, based on the catalog item name, both at 60% opacity.

Note that geojson still does not show a legend. Should we adapt the csv legend framework to show it, or are there situations where geojson is already showing a legend? Not sure that should be part of this PR...

RacingTadpole:
I fixed the potentially very confusing "colorPalette.js" in the latest commits. Now there is a `standardCssColors` object which you can draw some basic arrays of colors from - currently only `highContrast`, but #1384 moves the default ColorBrewer 9-color palette in there too. I wanted to combine this with `colorMap.js` but that is quite a different thing, as colormaps have more structure.

This solves the issue - lat/lon csvs with no data variable (or with none checked) are shown on the map, with a single random color for every point; the color tries to be different for different csv files.

Note it also applies the same handling to region-mapped csvs, ie. it fixes #1356 . (I've improved the specs for this case in the commit below.)

There is room for future improvement - with only 22 unique colors to choose from, and the choice made by hashing the file name, there is about a 5% chance that two files will have the same color. Also, the idea with the max-contrast color set is that you should start at the first one and work down. We could solve this by coloring based on order of adding, rather than the filename. But that would break consistency when sharing.

All in all, this is ready for review again.

lib/Map/DisplayVariablesConcept.js
```diff
@@ -75,12 +75,12 @@ inherit(Concept, DisplayVariablesConcept);
...
 defineProperties(DisplayVariablesConcept.prototype, {
     /**
-     * Gets a value indicating whether this item has child items.
+     * Gets a value indicating whether this item has visible child items.
      * @type {Boolean}
      */
     hasChildren : {
         get : function() {
-            return this.items.length > 0;
+            return this.items.filter(function(concept) {return concept.isVisible;}).length > 0;
```
kring: Minor, but this can be slightly more efficiently written as `this.items.some(function(concept) { return concept.isVisible; });`

lib/Models/GeoJsonCatalogItem.js
```diff
@@ -408,20 +401,16 @@ function loadGeoJson(geoJsonItem) {
 
     */
 
-    function defaultColor(colorString, palette) {
+    function defaultColor(colorString, name) {
         if (colorString === undefined) {
-            return getRandomColor(palette, geoJsonItem.name);
+            var color = Color.fromCssColorString(getRandomCssColor(standardCssColors.highContrast, name));
+            color.alpha = 0.6;
```
kring: We probably want points to be opaque, right?  Code below gives polygons (via the fill property) an alpha of 0.75, and that could maybe default to a bit lower (like 0.6).

kring:
This looks good to me, other than those two minor things.  @stevage do you want to look at this before we merge it?

RacingTadpole:
I've done the `.some` change, thanks Kevin.

On looking at this again, I have made some minor additional changes in the last commit:
- Using opacity 0.6 looks awful en masse, eg. #test airports. So I have changed it to 100%.
- Bike racks were being shown in black, which looks funny, and clashes with the default "no value" color. So I have removed black from the palette.
- Randomly choosing the stroke as well as the marker-color can look good, but it can look terrible - eg. for #test airports, they are both khaki. I personally think always using a black stroke would look much better - so I've made this change, but really it's up for discussion.

PR #1340, opened 2016-02-26
Update to terriajs-cesium 1.18.0

### LINKED ISSUE #1139: Feature info on timeseries raster contain values for two times
When you click on a raster layer that has a timeseries, like this one:

http://neii.bom.gov.au/services/solarclim/wms/data/monClim_dirNorExp1Day.nc

Then when the feature info window pops up you get two results, one for the time you are currently on, and one for the next time. I'm guessing this is because Cesium pre-loads the tiles for the next time so they are ready to switch to (or something like that).

It would be great to make it only return feature info for the current time interval.

As can be seen in the screenshots below there are two values returned, but only one has the layer name associated with it.

screen shot 2015-12-07 at 15 45 16

screen shot 2015-12-07 at 15 45 23

kring:
Aside from updating package.json to use the new version, this updates TerriaJS for a few breaking changes in Cesium.  The biggest change was in DrawExtentHelper, which used a method of drawing polygons that has been removed, but as a bonus the shift-click-zoom thing works across the antimeridian now.

Fixes #1139


kring:
So this pull request now also switches Travis to run our build and tests on Node 5.x instead of 0.10.x.

This is to appease the kraken :octopus:  (even master is broken as of this writing).  Here are the ways of the kraken:

Our `UrthecastCatalogItem` (which we don't use in NationalMap and our other maps but we do use in the TerriaJS tests, of course) depends on the `urthecast` package, which depends on `"superagent": "^1.4.0"`.  Today, a new version of `superagent` was published, `1.8.0-beta.1`.  Now in today's versions of npm, a version with a beta tag will never match `^1.4.0`, and so the latest non-beta-tagged version, `1.7.2`, will continue to be used and all is well.

However, the old version of npm (1.4.2) installed by nvm on travis when we request node 0.10 apparently doesn't follow this rule and so it installs this new beta version.  The new beta version, either accidentally or purposely, doesn't work on IE9 because it uses the undefined `__defineGetter__`.

Thus, our tests started failing today without us changing anything.

"But wait," you say, "we can't switch to Node 5.x, because our production servers run a LTS release of Ubuntu and so they still only have node 0.10."  This is true, but keep in mind that `terriajs-server` is a separate project.  It's entirely possible - and perhaps even reasonable - to _build_ our apps with Node 5.x, and then run terriajs-server in production on Node 0.10.x.  There's no reason these versions need to be the same.  Of course, we should continue to run the `terriajs-server` tests on the version of Node we intend to use in Production.

Or, even better, we could start using Node 5.x in production, but the point is we don't have to prior to merging this pull request.

RacingTadpole:
Thanks Kevin! Two questions -
- Do we have any example open street map catalog items I can check this on? (I just searched for `open-street-map` in national map and aremi, with no luck.)
- I didn't even know we had a `DrawExtentHelper` - how do I activate this to check it works?

kring:
> Do we have any example open street map catalog items I can check this on?

The CartoDB base maps use OSM.  Aaaand they don't work.  I'll fix this now.

> I didn't even know we had a DrawExtentHelper - how do I activate this to check it works?

Shift-click drag on the 3D map.

kring:
OSM work again and this is ready!

PR #1341, opened 2016-02-26
Made CatalogMember disclaimers display when loaded from share link

### LINKED ISSUE #1316: Layers popup disclaimer does not activate when using share URL
When sharing a layer with a popup disclaimer using a share URL and there is a sitewide disclaimer, the layer disclaimer does not activate once you click through the site disclaimer.

To test, the Electricity -> Network Opportunities layers in AREMI has a popup disclaimer and there is a sitewide disclaimer, when using the following share URL you only see the site disclaimer:

http://nationalmap.gov.au/renewables/#share=kb6JKy

(Best to use incognito mode since the layer disclaimer only shows once per browser.)

AlexGilleran:
Fixes #1316. 

Previously disclaimers were shown by the side panel on a click event, this binds them to the model's `isEnabled`/`isOpen` property directly, via a new event in `Terria` called `disclaimerEvent`.

Of course Terria and hence shared items are loaded before any part of the UI, meaning that this event was ignored, so I created a `QueueingEvent` that remembers events raised until the first listener is added, then immediately replays. This means that `disclaimerEvent`s triggered during the load of Terria are remembered for when the UI loads and modals can be displayed.

The other problem is that loading from a share key means it's quite easy to have the same disclaimer pop up a bunch of times for the same disclaimer key, as isEnabled can be set to true multiple times before the disclaimer is clicked, and multiple data items can have the same disclaimer. To prevent this I've created a `DisclaimerViewModel` which listens for `Terria.disclaimerEvent` and pops up a disclaimer when one is triggered. If more events come in for the same disclaimer before it's dismissed by the user, the existing disclaimer stays on screen and the callback of the second call is remembered. When the disclaimer is dismissed, all callbacks that have come in for that disclaimer are executed.

This requires `DisclaimerEvent` to be added to `index.js` in NationalMap (and hence in all clients that display a disclaimer), which could be said to be a breaking change seeing how it will stop disclaimers from showing if this isn't done - possibly I could put it in `TerriaViewer` to avoid this?


CHANGES.md
```diff
@@ -5,6 +5,10 @@ Change Log
 ### 2.1.2
 
 * Upgraded to terriajs-cesium 1.18.0.
+#### 2.2.0
```
stevage: This should be above ### 2.1.2, right?

stevage:
Looks superficially good to me, but probably makes more sense for @kring or @RacingTadpole to review all the event stuff.

CHANGES.md
AlexGilleran: Yeah, copped a wonky merge there, fixed :)

AlexGilleran:
heh @RacingTadpole said the same thing so I guess it's up to you @kring :)

lib/Models/CatalogGroup.js
```diff
@@ -82,7 +82,22 @@ var CatalogGroup = function(terria) {
         }
     };
 
-    knockout.track(this, ['isOpen', 'isLoading', 'items']);
+    knockout.track(this, ['_isOpen', 'isLoading', 'items']);
+
+    knockout.defineProperty(this, 'isOpen', {
```
kring: What not just subscribe to change notification on `isOpen` instead of overriding the property?  In fact, we're already subscribing to it below, so hopefully you can just add some more logic to the existing handler.

lib/Models/CatalogItem.js
```diff
@@ -268,6 +268,21 @@ var CatalogItem = function(terria) {
         }
     });
 
+    knockout.defineProperty(this, 'isEnabled', {
```
kring: Same here, subscribe instead of overriding.

lib/Models/CatalogItem.js
```diff
@@ -984,7 +999,14 @@ function isEnabledChanged(catalogItem) {
             catalogItem._loadForEnablePromise = resolvedOrRejected ? undefined : loadPromise;
         }
 
-        catalogItem.isShown = true;
+        if (catalogItem.isShown) {
+            // If item already had the isShown flag but wasn't enabled, it wasn't actually shown, so call isShownChanged
+            // to update the view.
```
kring: When would this happen?  Setting `isShown` will set `isEnabled` first and then wait for the enabling to happen for continuing with the showing.  It seems like adding this logic here could result in the showing logic happening twice in some cases?

AlexGilleran:
I'm overriding instead of subscribing because I want to stop the property from changing until the disclaimer is accepted. If I subscribe that'll be too late because as the disclaimer pops up the layer will already be loading behind it.

Unless I can immediately set the value back to false in the subscriber and that will act the same as stopping it from changing?

lib/Models/CatalogItem.js
AlexGilleran: Normally it would set `isEnabled = true`, except now `isEnabled` isn't set until the disclaimer is accepted. When it does get set, `isShown` is already true so the layer doesn't update and actually show itself.

Will take another look re: the showing logic happening twice.

kring:
Hmmm interesting.  Ok, so, in a perfect world, when the user clicks the checkbox next to an item with a disclaimer:
- The check would actually appear, giving the user feedback that they actually clicked it.
- The disclaimer would appear.
- Ideally the layer would _not_ appear on the map until the user agreed to the disclaimer.  IMO this is not critical, showing the data but disallowing interaction via the modal would be just fine, though I can imagine some debate on this.
- If the user agrees, the modal closes and the user sees the data on the map.  If they do not agree, the check mark disappears and the data disappears (if it ever appeared).

IMO overriding the `isEnabled` is the wrong way to do this, because:
- If the set of the value to true isn't allowed to "go through", the checkbox won't appear in the UI.
- It's pretty dodgy to have something that acts like a simple property, but when you try to set it, it ignores you for awhile and then maybe later actually acquires the value you tried to set.

I have to think some more about a better way, though.  One option is to hook into the load process.  i.e. the load isn't complete until the disclaimer is accepted.

AlexGilleran:
Yeah ok. It's hard to be sure a user accepts the disclaimer before a layer is enabled without specifically hooking into isEnabled, I guess the compromise would be to check for the disclaimer specifically in the explorer panel and when loading a share link and be careful when introducing other ways to enable layers.

>  IMO this is not critical, showing the data but disallowing interaction via the modal would be just fine, though I can imagine some debate on this.

Does this sound right @meh9 ? I recall you saying that there might be legal problems if someone was able to view a layer without accepting the disclaimer but I'm not sure how serious this was.

If it's not serious do we need to have a disclaimer that accepts a callback at all?

kring:
Yeah I _think_ the best way is to hook into `isEnabledChanged`.  Instead of:

```
var loadPromise = when(catalogItem.load(), function() { ... });
```

Do this:

```
var loadPromise = when.all([catalogItem.load(), promptForDisclaimerAsPromise(catalogItem)]).then(function() {
    // user accepted the disclaimer and load succeeded
    // ... same logic as before
}).otherwise(function(e) {
    // user rejected the disclaimer or the load failed.
    catalogItem.isEnabled = false;
    throw e;
});
```

Or something like that...  I've changed the code a bunch of times but I think this is close now.

kring:
> If it's not serious do we need to have a disclaimer that accepts a callback at all?

I think we still need that.  If the user rejects the disclaimer, we still need to turn the layer off.  I'm just arguing that having it in the background behind the modal window prior to accepting/rejecting isn't terrible.

AlexGilleran:
Yeah I like the promise idea.

AlexGilleran:
It shows promise :grimacing:

AlexGilleran:
- Changed the disclaimer event to hook into the `load` promise activated when isEnabled / isOpen changes. This does however mean that in some cases it's possible to open a group without accepting the disclaimer (but you still see it, I don't think this is a massive deal).
- Pulled the common `load()` code in `CatalogItem` and `CatalogGroup` into `CatalogMember` - this was done as part of a different approach that turned out to be a dead end but I still think it's an improvement in general.

lib/Models/CatalogItem.js
kring: Is this still need after the changes to make the disclaimer sort of hook into the load process?  I can't think of why it would be, but maybe I've missed something.

lib/Models/CatalogMember.js
```diff
@@ -399,6 +408,55 @@ CatalogMember.prototype.enableWithParents = function() {
     throw new DeveloperError('Types derived from CatalogMember must implement a "enableWithParents" function.');
 };
 
+CatalogMember.prototype.waitForDisclaimerIfNeeded = function() {
+    if (this.needsDisclaimerShown) {
+        var deferred = when.defer();
+        this.terria.disclaimerEvent.raiseEvent(this, deferred.resolve);
+        return deferred;
```
kring: Should be `return deferred.promise;`

AlexGilleran:
Done :)

AlexGilleran:
Removed QueueingEvent, made disclaimer run through a callback rather than an event and used `alert` as a fallback if `DisclaimerView` isn't used.

PR #1350, opened 2016-02-29
Fix error when adding single line csv

### LINKED ISSUE #1348: csv file with one data line gives error
Eg drag on csv file with 
Organisation, Lat, Lon
Test, -33.5055944, 151.3006284

Gives error:
"Data could not be added
The specified data could not be added because it is invalid or does not have the expected format."

tephenavies:
Fixes #1348.


stevage:
Test case?

RacingTadpole:
Cool - thanks Steve! I notice it doesn't show a legend though - you could do this instead:

```
/**
 * Given a simple array of css color strings, eg. ["red", "orange", "black"],
 * return an evenly spaced array suitable to instantiate a color map.
 * If a single color is provided, eg. ["red"], returns a color map with two entries: red at offsets 0 and 1.
 * @param  {String[]} simpleArray A simple array of css color strings, eg. ["red", "orange", "black"].
 * @return {Object[]} An array of {color, offset} objects.
 */
function simpleArrayToArray(simpleArray) {
    if (simpleArray.length === 1) {
        var color = simpleArray[0];
        return [{color: color, offset: 0}, {color: color, offset: 1}];
    }
    return simpleArray.map(function(el, index) {
        return {color: el, offset: index/(simpleArray.length - 1)};
    }).filter(function(o) {
        return o.color !== '';
    });
}
```

RacingTadpole:
That produces a legend for the lat-lon case, but it still doesn't produce one for the region-mapped one-line csv...

tephenavies:
@stevage  Here is a test case:
[test.txt](https://github.com/TerriaJS/terriajs/files/151878/test.txt)

RacingTadpole:
Great - could you add a spec for both of Bill's examples in `CsvCatalogItemSpec`?

tephenavies:
@RacingTadpole Sure

tephenavies:
There is a test in `CsvCatalogItemSpec` that has only 1 line of data, but it is added using `updateFromJson`:

``` javascript
it('identifies "latitude" and "longitude" fields', function(done) {
        csvItem.updateFromJson( { data: 'latitude,longitude,value\n-37,145,10' });
        csvItem.load().then(function() {
            expect(csvItem.dataSource.tableStructure.hasLatitudeAndLongitude).toBe(true);
        }).otherwise(fail).then(done);
    });
```

RacingTadpole:
Ah, but that has a scalar 'value' field, and so the legend works fine for it; it is only when the field is an ENUM, as in Bill's example, that you get this problem. Try this test:

```
    it('handles one-line data with enum', function(done) {
        csvItem.updateFromJson( { data: 'lat,lon,org\n-37,145,test' });
        csvItem.load().then(function() {
            expect(csvItem.dataSource.tableStructure.hasLatitudeAndLongitude).toBe(true);
        }).otherwise(fail).then(done);
    });
```

tephenavies:
Ahh, I (wrongly) assumed that the difference was caused by loading from json compared to opening a csv, so I made a CSV and a test case using that:

``` javascript
it('handles a csv with a single line of data', function(done) {
        csvItem.url = 'test/csv/lat_lon_single_line.csv';
        csvItem.load().then(function() {
            expect(csvItem.dataSource).toBeDefined();
            expect(csvItem.dataSource.tableStructure).toBeDefined();
            expect(csvItem.dataSource.tableStructure.columns.length).toEqual(3);
        }).otherwise(fail).then(done);
    });
```

But your test case doesn't need a file, so I'll replace mine with that.

RacingTadpole:
This looks good, thanks @steve9164 . I just merged in the latest master, and added a check that `legendUrl` is defined to those two new tests, since we want to be sure there's a legend as well. I'll merge once travis is done.

PR #1352, opened 2016-03-01
Improve performance of SA1 region mapped csvs

### LINKED ISSUE #1313: Large performance regression with SA1 region mapped csv since csv_refactor
There is a very large performance difference between changing variables for a large region mapped csv on [staging](http://staging.nationalmap.nicta.com/) (running `2016-02-15 terriajs: 2.0.1`) and [nationalmap live](http://nationalmap.gov.au/) (running `2016-01-19 terriajs: 1.0.54`).

Testing with [terriajs/wwwroot/test/csv/55k-SA1s.csv](https://github.com/TerriaJS/terriajs/blob/master/wwwroot/test/csv/55k-SA1s.csv), switching variables takes under a second to start reloading tiles and redrawing on nationalmap but over a minute on staging (and in master) – actually it took about 6 minutes to change variables.

I tried to run a `git bisect` to find the change, but wasn't able to, however I think the issue could be somewhere in csv_refactor.

RacingTadpole:
This fixes the performance slowdown of SA1 region-mapped csvs - fixes #1313.
There were two pain points:
- `formatNumberForLocale` was called for every cell in the csv file, and it was slow; this has been sped up by caching the formatter.
- We had inadvertently started using `ckmeans` when there is no TableColumn active (which is true briefly when the active variable is changed from one to another), and this is slow for 55,000 rows.

As a result the first 1000 lines of 55k-SA1s.csv load in approx 2s on my computer, very close to the earlier performance noted. The full 55k-SA1s file takes about 15s to load. It changes variable very quickly again.


lib/Core/formatNumberForLocale.js
```diff
@@ -32,10 +35,14 @@ function formatNumberForLocale(number, options) {
     options.useGrouping = defaultValue(options.useGrouping, false);
     options.maximumFractionDigits = defaultValue(options.maximumFractionDigits, 20);
     if (toLocaleStringSupportsOptions()) {
+        // The order of options may lead to equivalent keys being cached multiple times, but this is still better than creating a new Intl.NumberFormat every time.
```
AlexGilleran: Can you change options into a sorted array of two-tuples and stringify that? Just a thought.

AlexGilleran:
:+1: , the comment I had is just a suggestion. Needs a merge too.

lib/Core/formatNumberForLocale.js
RacingTadpole: Yes... but I'm not sure it's worth it. The real gain here is saving recreating the formatter 1000s of times (once per row of the csv file). In contrast the minor performance hit from caching the same formatter twice (or even several times), instead of once, doesn't seem too bad, vs. the complexity of adding the extra code. What do you think?

RacingTadpole:
The last 5 commits all make the csv load time much faster - see https://github.com/TerriaJS/terriajs/issues/1313#issuecomment-192035911 .

PR #1354, opened 2016-03-01
Clean up documentation and error handling around CKAN filter queries.

stevage:


lib/Models/CkanCatalogGroup.js
```diff
@@ -305,6 +307,13 @@ CkanCatalogGroup.prototype._load = function() {
     var that = this;
 
     var promises = [];
+    if (!defined(this.filterQuery) || typeof this.filterQuery !== 'object' || typeof this.filterQuery[0] !== 'string') {
```
AlexGilleran: You can use `Array.isArray()` in IE9+

AlexGilleran:
Looks fine apart from the edge case where filterQuery is an object but not an array, not sure if that's possible?

Also is there any simple way to test it?

PR #1363, opened 2016-03-02
Added `fogSettings` option to `TerriaViewer`

keattang:
Added `fogSettings` option to `TerriaViewer` which allows you to alter the Cesium fog settings that were introduced in Cesium 1.16

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Change Log
 * Warn user when the requested WMS layer doesn't exist, and try to provide a suggestion.
 * Fixed the calculation of a csv file's extent so that missing latitudes and longitudes are ignored, not treated as zero.
 * Upgraded to terriajs-cesium 1.18.0.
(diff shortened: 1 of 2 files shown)


RacingTadpole:
Great, thanks. This would require the application to set the fog settings in code, eg.

```
TerriaViewer.create(terria, { fogSettings: {density: 5.0e-4, screenSpaceErrorFactor: 0} });
```

What about putting this into the initialisation json file instead? Eg. it currently has things like:

```
{
    "homeCamera": {"west": 116, "south": -27, "east": 149, "north": -8},
    "baseMapName": "Australian Topography",
    "initialViewerMode": "3d",
    "services": [], ...
}
```

to which we could add `"fogSettings": {...}`.

I think we'd have to add a `fogSettings` to `Models/Terria.js`, set it in https://github.com/TerriaJS/terriajs/blob/master/lib/Models/Terria.js#L420-L468, and then `TerriaViewer` can pick it up from `terria` directly.

What do you think?

kring:
Thanks for getting this rolling @keattang, but I'm going to close this in favor of #1410.  Let me know if that PR doesn't cover your use case.

PR #1372, opened 2016-03-03
Automated testing improvements

kring:
There are now three gulp tasks for running tests:

`gulp test`

Detects which browsers are installed on your local system and runs the tests in all of them.

`gulp test-saucelabs`

Runs the tests in 7 browsers on Sauce Labs.  Set these environment variables with your authentication information: `SAUCE_USERNAME` and `SAUCE_ACCESS_KEY`.

`gulp test-browserstack`

Runs the tests in 7 browsers on BrowserStack.  Set these environment variables with your authentication information: `BROWSER_STACK_USERNAME` and `BROWSER_STACK_ACCESS_KEY`.

Also updated `.travis.yml` to do `gulp test-saucelabs test-browserstack`, which means it will (for now) run the tests on both services.  We'll probably decide pretty quickly which one is more reliable / faster and drop the other one.


kring:
I've just removed `test-browserstack` from Travis because our open source account isn't activated yet and we've already hit the time limit on our free trial.  I'll turn it back on when the account is activated, which will probably be late tonight.  Feel free to merge in the meantime if you want, you can try it out by creating your own BrowserStack account.

kring:
BrowserStack testing is back on.

AlexGilleran:
Looks good. Do you need a changelog entry for this?

kring:
> Do you need a changelog entry for this?

I originally thought no, because it doesn't impact users of TerriaJS (only developers).  But it can't hurt, I'll add a mention.

kring:
Done

AlexGilleran:
As discussed, it's krakeny :( :octopus:

kring:
Ok, I removed BrowserStack from Travis for now.  By the way our problem is described here, but no real solution yet:
https://github.com/karma-runner/karma-browserstack-launcher/issues/62

PR #1373, opened 2016-03-03
Wms transparent legend

### LINKED ISSUE #461: Request transparent background from WMS servers
Geoserver supports a bunch of options to control legend styling:
http://docs.geoserver.org/stable/en/user/services/wms/get_legend_graphic/legendgraphic.html#controlling-legend-appearance-with-legend-options

Other WMS servers (Esri for sure) don't support `legend_options`.

Both Geoserver and Esri support an additional parameter, `transparent=true`, that gives the legend a transparent background instead of a white one.  This doesn't go well when we put the transparent image on a dark background, though, because there doesn't appear to be any way to tell the Esri server to make the label text anything other than black.  In a geoserver, it works great, because we can also tell the geoserver to use white text.

Esri servers also support `width` and `height`.  By specifying the width to be the width of the legend window, we get much better results from Esri servers.  For example (from `Social and Economic -> Population Estimates`), this:

Without Width

becomes this:

With Width

But Geoserver interprets `width` differently, so we get terrible results if we pass `width=350` to a geoserver.

The upshot of all this is that, for best results, we need to do detection of the WMS server vendor, and tailor our `GetLegendGraphic` requests accordingly.

stevage:
This kind of half-heartedly fixes #461 .


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -372,8 +394,27 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
 
         if (style.LegendURL) {
             // According to the WMS schema, LegendURL is unbounded.
-            var legendUrl = Array.isArray(style.LegendURL) ? style.LegendURL[0] : style.LegendURL;
-            var legend = new LegendUrl(legendUrl.OnlineResource['xlink:href'], legendUrl.Format);
+            var legendUrl = Array.isArray(style.LegendURL) ? style.LegendURL[0] : style.LegendURL, legend;
```
kring: Why declare `legend` up here at the end of a long line?

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -141,8 +154,9 @@ var WebMapServiceCatalogItem = function(terria) {
                 return undefined;
             }
             var layer = this.layers.split(',')[0];
+            // This is a double fallback. Generally we get a legend URL from the getCapabilities, or it's not supplied in the init file.
             return new LegendUrl(cleanUrl(this.url) +
-                '?service=WMS&version=1.3.0&request=GetLegendGraphic&format=image/png&layer=' + layer, 'image/png');
+                '?service=WMS&version=1.1.0&request=GetLegendGraphic&format=image/png&transparent=True&layer=' + layer, 'image/png');
```
kring, line 170: When this is a geoserver, shouldn't this URL include the font size and antialiasing tweaks too?  Otherwise the Mobile Black Spot Programme datasets don't get a nice legend because they don't have an explicit style-with-legend listed in GetCapabilities.

kring:
Looks good to me other than those couple of comments.

lib/Models/WebMapServiceCatalogItem.js
stevage: Because in strict mode, all variables must be declared. :p

lib/Models/WebMapServiceCatalogItem.js
stevage, line 170: Wow, nice pick up. So many edge cases. In this situation:
- the legend Url wasn't explicitly set
- The GetCapabilities doesn't define a `<Style>` for that layer (and hence doesn't include an explicit `<LegendURL>`. I didn't know that could happen.

stevage:
There. I fixed it.

kring:
Steve did you see my comment on this one from 7 days ago?

> Is this part fixed? It doesn't look like it..?

Also please update CHANGES.md.

kring:
@stevage just a reminder that this one is waiting for changes from you.  Specifically this comment hasn't been addressed yet:

> When this is a geoserver, shouldn't this URL include the font size and antialiasing tweaks too? Otherwise the Mobile Black Spot Programme datasets don't get a nice legend because they don't have an explicit style-with-legend listed in GetCapabilities.

Also merge in master and update CHANGES.md please.

PR #1375, opened 2016-03-04
Improve csv loading to ignore any completely blank lines

### LINKED ISSUE #1366: csvs with completely blank lines do not zoom
Here's a csv file abstracted from a real example, where `extent` calculation fails

```
Organisation name,Organisation type,Description of Services,Main theme,Address,Postcode,Lat,Lon,Telephone,Contact Person,Email,Data source(s),Service Classification,Lead Administrative Agency,Geographical Region Served
,,,,,,,,,,,,,,
ABCD,Other NGO,Desc .åÊ,Education,"Level 4, A Building
Albert Avenue
Chatswood NSW 2067",2067,-33.1,151.2, 1234 5678 or 0400 000 000,A Smith,,http://example.com/resources/delivery,,,
,,,,,,,,,,,,,,
EFGH,Other NGO,Desc,Education,Address,2067,,151.3, 1234 5678 or 0400 000 000,A Smith,,http://example.com/resources/delivery,,,

,,,,,,,,,,,,,,
IJKL Australia,Other NGO,"Mon to Fri 9am to 5pm.Ê",Community,A StreetÊÊÊTumbi UmbiÊÊNSW,,-33.5,151.7,(02) 1234 5678,,,,,,
,Extra info,,,,,,,,,,,,,
```

It fails because the blank line in the middle leads to an `undefined` value, not a `null` value, and then the min/max lat/lon calculations give `NaN`, which means the extent rectangle is ill-defined.

RacingTadpole:
Fixes #1366.


lib/Map/TableStructure.js
```diff
@@ -219,14 +219,12 @@ TableStructure.fromCsv = function(csvString, result) {
     var json = csv.toArrays(csvString, {
         onParseValue: castToScalar
     });
-    // Remove extra blank lines at the end.
-    var n = 0;
+    // Remove any blank lines.
     for (var i = json.length - 1; i >= 0; i--) {
```
AlexGilleran: Might be simpler to use `Array.filter` here? Not sure if CSVs get long enough for it to matter, but the `json.splice` call makes this run in quadratic-time (I think, it's been a while since I read Javascript the Good Parts).

RacingTadpole: yeah, good idea - will do.

RacingTadpole:
back to you @AlexGilleran - thanks!

PR #1379, opened 2016-03-07
Geojson display

stevage:
- Fixed error choosing the fill colour of GeoJSON catalog items.
- Made GeoJSON catalog items a lighter shade, and with a thicker border for better visibility of adjoining polygons.


lib/Models/GeoJsonCatalogItem.js
```diff
@@ -448,12 +441,15 @@ function loadGeoJson(geoJsonItem) {
...
-        strokeWidth : defaultValue(style['stroke-width'], 2),
+        strokeWidth : defaultValue(style['stroke-width'], 4),
         stroke: defaultColor(style.stroke, lineAndFillPalette),
         markerOpacity: style['marker-opacity'] // not in SimpleStyle spec or supported by Cesium but see below
     };
 
-    options.fill = defaultColor(style.fill, options.stroke.clone);
+    options.fill = defaultColor(style.fill, options.stroke.clone());
+    if (!defined(style['fill'])) {
+        options.fill = options.fill.brighten(0.2, options.fill);
```
kring: Arthur's recently-merged pull request now makes polygon borders black, regardless of their fill color (unless overridden of course).  Maybe check that out and see if you still want to switch it to a lighter version of the fill color.

lib/Models/GeoJsonCatalogItem.js
```diff
@@ -448,12 +441,15 @@ function loadGeoJson(geoJsonItem) {
         markerSize : defaultValue(parseMarkerSize(style['marker-size']), 20),
         markerSymbol: style['marker-symbol'], // and undefined if none
         markerColor : defaultColor(style['marker-color'], pointPalette),
-        strokeWidth : defaultValue(style['stroke-width'], 2),
+        strokeWidth : defaultValue(style['stroke-width'], 4),
```
kring: This is also used for the outline of points and for polylines.  Do we want those to be width 4 too?  Also, polygon width > 1 doesn't work on Windows anyway, so I'm not sure how much value there is in bumping it up to 4.

stevage:
Looks like nothing left in this branch after merging master.

PR #1381, opened 2016-03-10
Highlight clicked polygons and polylines

### LINKED ISSUE #922: Highlight selected geometry
It would be very useful to highlight "selected" (clicked or hovered) geometries for the purpose of data analysis. For example, hovering over a linestring would highlight it (e.g. bolder and/or different colour) over the other linestrings in the same layer. Same applies to other geometries. This is somehow related to #553 although within a layer instead of between layers.

@stevage and @axman6 told me they would also be happy to see that feature, but my first attempt at finding where/how this should be done in such a massive code base wasn't successful, so I'm resorting to posing an issue. I guess it should carry the UX/Design tag (@philipgrimmett or @hilarycinis ?).

kring:
image

This works for vector features and for any raster server that provides feature info geometry in GeoJSON or Esri REST format.  So it works for most - but not all - of the items in National Data Sets.

Fixes #922


lib/Map/featureDataToGeoJson.js
```diff
@@ -0,0 +1,112 @@
...
+            }
+        };
+    } else if (spatialReference.wkid === 4326) {
+        return {
+            type: 'name',
+            properties: {
+                name: 'EPSG:4326'
+            }
+        };
+    } else if (defined(spatialReference.wkid)) {
```
AlexGilleran, line 94: Can't this branch be used for everything except `102100` ?

kring, line 94: Yep, good call!

AlexGilleran:
Works well in my testing as long as you're not worried about coarse outlines being retained when zooming in :).

image

kring:
This is ready.

RacingTadpole:
Looks great to me! It even works with region-mapped csvs and the ABS items.
Note though the highlighted boundary has turned yellow in the latest version - not sure if that was intentional?

tephenavies:
(Testing with an ABS region mapped dataset). Should the border show up in FeatureInfo when it's clicked? Currently it adds an extra entry that describes the region that was bordered. This also removes the border (since it's not longer the top item).

kring:
> Should the border show up in FeatureInfo when it's clicked? 

Ideally, no.  I'll see if I can stop that from happening.  Good catch!

> Note though the highlighted boundary has turned yellow in the latest version - not sure if that was intentional?

No, it wasn't intentional!  I'll take a look.

kring:
The highlighted boundary looks like it's still white to me...?  Actually on the white-ish basemaps it's black.

kring:
Oh, I know what it is.. you need to install Cesium 1.18.1.

kring:
Ok the highlight itself is no longer pickable.  This is ready.

PR #1383, opened 2016-03-10
Use superagent 1.7.x.

kring:
We don't actually need superagent directly, but urthecast uses it, and 1.8.0 (the latest version currently) uses form-data 1.0.0-rc3 which is not compatible with Internet Explorer 9 (it uses FormData, a DOM type not available in IE9, without guarding against undefined references).  So by adding a dependency specifically on 1.7.x here, we force urthecast to get that version and our tests pass in IE9 once again.


kring:
I'm insta-merging this because it unbreaks master.

PR #1384, opened 2016-03-10
Improve legend and coloring of ENUM columns of csv files

### LINKED ISSUE #1374: Improve legend and colouring of ENUM columns
ENUM columns currently use the `ckmeans` method to find colour bins:
https://github.com/TerriaJS/terriajs/blob/1313_csvperf/lib/Models/LegendHelper.js#L302

If the number of unique ENUM values is less than `binCount`, this gives you back the ENUM values, which is fine.

However, there are two problems if there are lots of different ENUM values (ie. more than `binCount`):
- Speed: If there are 55,000 unique values, it can take over a minute to calculate.
- Interpretability:  Eg. in the below, the regions north of Sydney have the first letter "W", which is not on the legend, and which are coloured orange (the same as "Q"):
  image

Originally identified in #1313 .

RacingTadpole:
Fixes #1374.

Depends on #1313 and #1338.

Improved legend and coloring of ENUM (string) columns of csv files; only the most common values are colored differently, with the rest shown as 'Other'.


AlexGilleran:
Has a conflict, can you merge it? :grimacing:

RacingTadpole:
Thanks, all merged. Note this includes #1338, so you might want to wait for (or merge) that one first.

AlexGilleran:
Ooh thanks for the heads up. Looks like that's already being discussed by other people so I'll leave this til then :)

RacingTadpole:
Ready to go, thanks.

PR #1389, opened 2016-03-11
Hide ABS regions when a concept is deselected

### LINKED ISSUE #1355: ABS region mapped csv display doesn't change when variable is deselected
Moved from #1267 into it's own issue.

When the last selected concept of a concept group is deselected, the legend is removed but the layer on the map is not updated.

The problem seems to be caused logic error in AbsIttCatalogItem.js: rebuildData. When there are no variables selected, tableStructuresAndCombinations is undefined. This means that regionDetails will also be undefined, so the imagery is not updated.

RacingTadpole:
Fixes #1355.


AlexGilleran:
So it does hide the regions, but this is only reflected on the map when your mouse actually leaves the side panel and hovers over it, so it looks initially like it's done nothing. Could it do a manual refresh or something?

kring:
Sounds like we just need to add a call to `terria.currentViewer.notifyRepaintRequired`.

RacingTadpole:
Strange - I can't reproduce this behaviour on either Chrome or Safari.
However I have discovered a problem - it dies if you turn off the ABS layer and then switch 2D/3D. I'll fix that in the meantime.

RacingTadpole:
OK, I've fixed the error I saw - can you take a look and see if you're still getting your original error? (On the optimistic hope that one fix magically fixes the other...)

RacingTadpole:
Fixed. For the record, it was indeed calling `notifyRepaintRequired` when changing the imageryLayer, but not when only hiding it, via the `regionMapping._changed.raiseEvent` call in `RegionMapping` and the `item._regionMapping.changedEvent.addEventListener(dataChanged.bind(null, item), item)` call in `CsvCatalogItem`. `dataChanged` just calls `notifyRepaintRequired`.

PR #1392, opened 2016-03-11
Newui mobile

chloeleichen:


lib/ReactViewModels/ViewState.js
```diff
@@ -7,6 +7,9 @@ var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
  * Root of a global view model. Presumably this should get nested as more stuff goes into it. Basically this belongs to
  * the root of the UI and then it can choose to pass either the whole thing or parts down as props to its children.
  */
+
+// mobileViews:'data', 'preview', 'nowViewing', 'search';
+
```
kring, line 12: you could put this doc on the `mobileView` property itself.  Use the jsdoc `/**` style comments.  There's a Sublime plugin called DocBlockr that makes it easy to write them.

kring, line 12: Also doesn't ES2015 have an enum type?  Even if not, you can define a type with `data: 0`, `preview: 1`, and `nowViewing: 2` (or even use the strings instead of numbers) because it's less prone to typos that way.

lib/ReactViews/MobileSearch.jsx
```diff
@@ -0,0 +1,20 @@
...
+const MobileSearch = React.createClass({
+    propTypes: {
+        viewState: React.PropTypes.object,
+        terria: React.PropTypes.object
+    },
+
+    render() {
+        return (
+            <div className="search--mobile">
+                mobile search
```
kring, line 14: I guess this is a work in progress?  Doesn't look like it does much yet. :)

kring:
This all looks ok to me!  Maybe just delete the commented-out code for tidyness and then go ahead and merge it.

kring:
wow Github has reactions now...  pretty limited set though. :octocat:

PR #1393, opened 2016-03-14
Make feature info respect current time, for region-mapped csvs

### LINKED ISSUE #1248: Get feature info doesn't respect current time
Issue from @TobyBellwood:

> So creating the dataset with an ISO-8601 date YYYY-MM-DD creates the slider properly, and the csv reader interprets the min and max values for the scale correctly (ie max of all years).  The allowance of a year-only date makes sense (only where multiple years exist in the dataset – and  accepted as per https://github.com/NICTA/nationalmap/issues/192 )
> 
> What I did notice however is that the feature info box doesn’t update once the timeseries advances and updates the coloring for the polygons.  Even in 2011/2012 the infobox still displays the initial 2009 data.

unnamed

RacingTadpole:
Fixes #1248 .
You can try this out with the #test droughts.csv and/or [SA3-Region-Innovation-Data-2009-14 csv-geo-au](https://data.gov.au/dataset/sa3-region-innovation-data-2009-14).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,7 @@ Change Log
 * Improved ABS display (to hide the regions) when a concept is deselected.
 * Improved readability of ArcGis catalog items and legends by replacing underscores with spaces.
 * `ArcGisMapServerCatalogItem` metadata is now cached by the proxy for only 24 hours.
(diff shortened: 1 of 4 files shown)


RacingTadpole:
Ready to go again.

PR #1395, opened 2016-03-15
Improve legend when tableStyle.colorBins is 0

RacingTadpole:
Improved legend when tableStyle.colorBins is 0.
See it with localhost:3001/#build/TerriaJS/test/init/test-tablestyle.json.


RacingTadpole:
Don't merge this yet.

RacingTadpole:
This appears to work, but I've actually deleted a chunk of code for displaying gradient legends which appears to be redundant. That makes me a little bit nervous, so I suggest we merge it after the release today.

stevage:
What was the thinking here? Obviously the result changes the semantics of 'colorBins: 0'.
- Previously: "Don't assign values to bins, instead map them linearly onto colors and generate a smooth gradient legend".
- Now: "Don't assign values to bins, instead give them all the same grey color".

Although, judging from the behaviour at NM, the former behaviour has been broken for a while:

screenshot 2016-04-05 15 45 59

stevage:
screenshot 2016-04-05 15 49 11

This is how it looked previously.

RacingTadpole:
Ah thanks, I didn't realise what the old semantics were - so I was changing the legend to match the display, which it looks like is broken. I'll rework this PR to match the old behaviour.

RacingTadpole:
Great - got it matching the original behaviour again, and even added some tests. Here's the latest:
image

PR #1398, opened 2016-03-15
Use terriajs-cesium 1.19.0.

kring:


kring:
Don't merge this yet.  There are lots of deprecation warnings and probably bugs as a result of Cesium making `defined` now equivalent to `definedNotNull`, and deprecating `definedNotNull`.

RacingTadpole:
That's an interesting change. I see the discussion here https://github.com/AnalyticalGraphicsInc/cesium/pull/3522

kring:
Yeah I'm not sure I 100% agree with the change, but probably not worth fighting. :)

kring:
This is ready now.

RacingTadpole:
@AlexGilleran discovered that we distinguish `null` and `undefined` for `tableStyle.timeColumn`.
https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CsvCatalogItem.js#L368

I'll update the PR to handle this. However, maybe this is a sign to implement this in a different way...?

RacingTadpole:
Oops - looks like we already had a spec in place for this, and that you've already updated `CsvCatalogItem`, @kring! Thanks. @AlexGilleran - can you replicate the problem you were seeing?

RacingTadpole:
@AlexGilleran: I can replicate the problem. I'm looking into fixing it.

AlexGilleran:
OK great, I was just finding I could still reproduce it and presumed I'd messed up an npm link somewhere :(

RacingTadpole:
Ah, it's a subtle one - we fall afoul of the `defined` check in https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CsvCatalogItem.js#L368 . 

_This means that any `json` that sets a property to `null` will now be ignored._

The [spec](https://github.com/TerriaJS/terriajs/blob/master/test/Models/CsvCatalogItemSpec.js#L307-L316) didn't pick it up because it sets the properties of the tableStyle directly, so doesn't go through `TableStyle`'s `updateFromJson`.

I propose to change https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CsvCatalogItem.js#L368 so that it recognises `null` as distinct from `undefined` again. This will fix the problem for any other `json` properties we've missed, if any.

AlexGilleran:
:shipit:

PR #1399, opened 2016-03-16
Included features in share links

### LINKED ISSUE #1080: Share view with feature info
There are times when it would be useful if the share URL included the currently clicked feature, so you could point someone to a specific feature - not just a whole layer.

For instance, I've just come across what I think is a data error, but the only way to share it is to take a screenshot, or include verbal instructions about what to click on.

AlexGilleran:
Fixes #1080.

I thought I was just about done with this a few days in. That was nearly two weeks ago. This change has nearly killed me, and now, dear reviewer, you can finish the job by nit picking it :stuck_out_tongue: 
- `SharePopupViewModel` now picks up `terria.pickedFeatures` and `terria.selectedFeature` and serialises them as part of the share link if they're present, `Terria` picks up them up from the link.
- How do we serialise a bunch of features? 
  - For raster features we record the tile coords (x, y, height) for each ImageryLayer that features were picked from by hooking into the `ImageryProvider.pickFeatures` method.
  - For vector features we record the id and name (as ids are sometimes autogenerated) for each feature
- This involved a bunch of messing about with `Cesium.js` and `Leaflet.js`, so I did some refactorings that you may or may not agree with.
- On the upside I also wrote some tests for both the above :).


RacingTadpole:
This is awesome.
As requested, here's an initial nitpick. It is probably really hard to fix too. :-)
I managed to select some power stations at the same location with the same name (eg. Northern Power Station near Port Augusta, SA); when shared, all the power stations with that name were open, not just the one that was open. And collapsing & opening one of them did both.

RacingTadpole:
This is awesome (again). It even works when you share a 3D view on a browser that only supports 2D.

Here's another little thing - I think much easier to fix. If you already have a feature picked, with a polygon highlighted, and then paste in a shared URL on top, it leaves the old polygon highlighted.

Also, if the feature info panel is already open, sometimes it doesn't update to the newly opened feature.

AlexGilleran:
- Fixed the Port Augusta Problem - now instead of trying to match on id and then name I just dumped the id completely (given how often they're unstable they're next to useless) and instead I've made it resolve by a hash of the feature's name and all its properties, resolving collisions by name. This is a best-effort solution (it won't work for layers where they have the same name and the same properties, for instance) but it works better than the old solution.
- Fixed the features not being selected when just changing the hash, was because I always waited for the viewer to load even if it had already loaded.

lib/ViewModels/FeatureInfoPanelSectionViewModel.js
```diff
@@ -290,29 +291,6 @@ function replaceBadKeyCharacters(properties) {
...
-                result[key] = properties[key];
-            }
-        }
-    }
-    return result;
-}
-
 function applyFormatsInPlace(properties, formats) {
     // Optionally format each property. Updates properties in place, returning nothing.
     for (var key in formats) {
```
RacingTadpole, line 296: Good idea to break this out!

lib/Map/CesiumTileLayer.js
```diff
@@ -169,22 +170,26 @@ var CesiumTileLayer = L.TileLayer.extend({
...
                 return undefined;
             }
 
-            return that.imageryProvider.pickFeatures(tileCoordinates.x, tileCoordinates.y, level, longitudeRadians, latitudeRadians);
+            return that.imageryProvider.pickFeatures(tileCoordinates.x, tileCoordinates.y, level, longitudeRadians, latitudeRadians).then(function(features) {
+                return {
+                    features: features,
+                    coords: {x: tileCoordinates.x, y: tileCoordinates.y, level: level}
+                };
+            });
```
RacingTadpole: I'm just trying to understand the implications of this `then` clause... so `ImageryProvider.pickFeatures` says it returns a promise which resolves to an array of `ImageryLayerFeatureInfo` instances. So you're changing the result type of the `CesiumTileLayer.pickFeatures` promise... will it be confusing to have two different types of `pickFeatures` calls?

Is there any way to add some jsdoc-style explanation to this function?

lib/Models/Cesium.js
```diff
@@ -597,6 +599,161 @@ Cesium.prototype.lowerToBottom = function(item) {
...
+ * @returns {PickedFeatures} A {@link PickedFeatures} object that is a combination of everything passed.
+ * @private
+ */
+Cesium.prototype._buildPickedFeatures = function(providerCoords, pickPosition, existingFeatures, featurePromises, imageryLayers, defaultHeight) {
+    var result = new PickedFeatures();
+
+    result.providerCoords = providerCoords;
+    result.pickPosition = pickPosition;
+
+    result.allFeaturesAvailablePromise = when.all(featurePromises).then(function(allFeatures) {
```
RacingTadpole, line 727: Following on from my question about the changes to what `pickFeatures` returns... it's not obvious to me what sort of objects these `featurePromises` resolve to? Especially as they can come from `imageryProvider.pickFeatures` or `this.scene.imageryLayers.pickImageryLayerFeatures`...

test/ViewModels/SharePopupViewModelSpec.js
```diff
@@ -248,21 +254,101 @@ describe('SharePopupViewModel', function() {
...
+                });
+            });
+
+            it('should generate pickCoords from pickPosition', function() {
+                // Conversion of coordinates isn't exact
+                expect(parsed.pickCoords).toEqual({
+                    lng: 2,
+                    lat: 3.999999999999998,
+                    height: 6.000000649901364
+                });
```
RacingTadpole: ummm...! Perhaps a good place to use `expect().toEqualEpsilon()`?

lib/Models/Cesium.js
AlexGilleran, line 727: Yeah I actually came back to this myself and realised that I didn't know what it did anymore - clarified :)

test/ViewModels/SharePopupViewModelSpec.js
AlexGilleran: Ooh that's cool, done.

lib/Map/CesiumTileLayer.js
AlexGilleran: Yeah actually I've refactored this whole thing into two different methods that are a bit more clearly named (one to figure out what to pass into the picking function and the other one to do the picking).

AlexGilleran:
Updated :) :elephant:

AlexGilleran:
Also fixes #1430

PR #1410, opened 2016-03-21
Add fogSettings option to init files

RacingTadpole:
This implements my suggestion in #1363.

Now that I've done it, I'm not sure whether doing it this way is really worthwhile compared to @keattang's approach. I'm submitting it as a PR so another reviewer can see the alternatives and choose one.

The approach here means that you can change the fog settings in the init file, rather than in code. It also means you can change them by dragging and dropping on a json file such as:

```
{
    "fogSettings": {"density": 3.0e-3, "screenSpaceErrorFactor": 1},
    "catalog": []
}
```

Could there ever be a use case for this?


AlexGilleran:
Makes sense to me, the only problem would be if `Fog` starts taking in something that can't be expressed in JSON like a custom type.

PR #1413, opened 2016-03-23
Fix broken URLs for ABS ITT catalog items. Fixes Nicta/NationalMap#233

stevage:


RacingTadpole:
Good plan. Note the ABS data is usually combined from several URLs (eg. the Age 0-4 is one download, Age 5-9 is another).

This PR changes it to show only one of these download URLs, which may be misleading, especially as the downloaded values are in a column labelled simply "value". The user may think this has the values they selected, but in fact it might only have values for Age 0-4.

I suggest we either show no URL, show all the URLs, or show only one but include a warning that this is not the complete data.

stevage:
Ah, thanks, I didn't realise that. All the URLs would be ideal, if not too hard?

RacingTadpole:
I agree. 

Currently `dataUrl` is assumed to be a string, we could make it either a string or an array and update downstream code accordingly... although then its name is misleading, so I'm not sure I like that idea. Alternatively we could invent a new `dataUrls` array... but that will add a layer of complexity to the code. 

Maybe the simplest would be to concatenate the URLs into a comma-separated string, and store it in `dataUrl`. One trick there is removing the `proxy/` from each. The naming is still a bit odd, but at least it's always a string.

RacingTadpole:
I'm getting the idea you'd like me to do this one...? :-)

stevage:
Yes please :) Pretty low priority though really..

RacingTadpole:
This one may as well wait until we've implemented the new SDMX-JSON reader, since the urls will be different.

Issue #1419, opened 2016-04-04
Build process / tools roadmap


kring:
- [ ] Build UMD module(s) and publish with the npm package.  This allows TerriaJS to be used without Webpack.
  - [x] Generate UMD module(s).  `gulp build-libs`.
  - [ ] Add tests (or at least a demo project) that demonstrates the UMD module works.
- [ ] NPM Publish built CSS along with built UMD module.
- [ ] Make TerriaJS work with Browserify again.  One challenge is that we're now loading HTML and XML using the Webpack `raw-loader` (so `require('./whatever.xml')`) instead of the Browserify `brfs` (so `require('fs').readFileSync(__dirname + './whatever.xml', 'utf8')`).  Another is the Babel transformation.  Ideally we'd just publish individual modules that have already been transformed by loaders but _not_ combined.
- [x] Let Webpack build our CSS.
  - [x] Use CSS modules?
- [x] Add support for [hot module replacement](https://webpack.github.io/docs/hot-module-replacement.html) so we need to reload the app less often
- [ ] Enable use of [Flow](http://flowtype.org/)
- [x] Separate our build into separate [chunks](https://webpack.github.io/docs/code-splitting.html).  e.g. one for Cesium, one for Leaflet, one for WPS support, etc.
- [ ] Tighten up eslint rules.  Currently they're chosen primarily to not flag anything that jshint didn't already flag (so we get a clean report without changing all our code).  We should evaluate what rules we want to turn back on.
- [x] Use the Webpack dev server to get faster builds and maybe other benefits.

kring:
Not much here that we need an issue to track, closing.

PR #1420, opened 2016-04-04
Switch to Webpack instead of Browserify

kring:
And also clean up the build process quite a bit.  Some of the improvements here:
- Added support for ES2015 via Babel.
- Incremental builds now take less than 1 second on my system, instead of 5-6 seconds.
- There's now a clearer separation between gulp tasks run when TerriaJS is `npm install`ed and those that are run in the course of development.  This should fix problems like the recent one where an explicit `npm install karma` was necessary when using TerriaJS from git instead of npm.
- It puts us in a good position to do a bunch of other cool things, like described in #1419.
- Adds a handy landing page on `http://localhost:3001/` with links to the tests, doc, etc.
- Removes the custom `server.js` and instead uses `terriajs-server` to host the web server used to run the tests.


kring:
(NationalMap branch to use with this coming soon!)

.eslintrc
```diff
@@ -0,0 +1,199 @@
...
+        "constructor-super": 1,
+        "generator-star-spacing": 0, // see Babel section
+        "no-this-before-super": 1,
+        "no-var": 0,
+        "object-shorthand": 0, // see Babel section
+        "prefer-const": 1
+
+    },
+    "plugins": [
+        //"react"
```
AlexGilleran, line 197: so this isn't needed because of babel-eslint?

EDIT: oh right this is master so no react, I get it

buildprocess/configureWebpack.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+/*global require*/
```
AlexGilleran: Does this file get linted? If so doesn't the commonJS thing in eslintrc fix it?

kring: It doesn't get linted by `gulp lint` but your editor will probably lint it (mine does).  I'm not sure exactly what you're asking though.  Are you asking if the `global require` is needed?

AlexGilleran: Yeah - all of these `/\* global require */ things should be superfluous after you've told .eslintrc to allow it

kring: Ah, good point.  I may have turned off the rule that would yell at us for the superfluous global because I didn't want to change every source file in TerriaJS.  :)

In any case, no reason to add it to new source files!

kring: Is the `'use strict'` superfluous as well?  I think it might be.

AlexGilleran: Yeah afaik babel adds it

AlexGilleran:
+1 from me

stevage:
Just noting that this didn't build with Node 5.4.1:

```
[12:38:17] Error: Cannot find module 'entities/maps/entities.json'
    at Function.Module._resolveFilename (module.js:339:15)
    at Function.require.resolve (internal/module.js:23:19)
    at configureWebpack (/Users/stevebennett/odev/nicta/nationalmap/terriajs/buildprocess/configureWebpack.js:41:23)
    at Object.<anonymous> (/Users/stevebennett/odev/nicta/nationalmap/terriajs/buildprocess/webpack.config.js:27:1)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
```

Upgrading to 5.10 worked.

kring:
I just merged in master, so now is a good time to merge this! ;)

PR #1426, opened 2016-04-05
Added mechanism for hiding the source of catalog items.

AlexGilleran:
Merging this into jsonTreeview because we need a hotfix for AREMI.

This allows catalog items to specify a `hideSource` property in their JSON, which defaults to `false` but when set to `true` will hide the details of the source of data for a service in the catalog item info panel. This is done by not showing the sections that would normally have this info, and filtering `CatalogMember.info` by excluding info items that have names in a new `_sourceInfoItemNames` property on `CatalogMember`. What info items to exclude varies from item to item, this includes `'GetCapabilities URL'` for `WMSCatalogItem` as that's the use case for this fix.

### DIFF
--- a/lib/Models/CatalogMember.js
+++ b/lib/Models/CatalogMember.js
@@ -41,16 +41,16 @@ var CatalogMember = function(terria) {
     this.description = '';
 
     /**
-    * Gets or sets the array of section titles and contents for display in the layer info panel.
-    * In future this may replace 'description' above - this list should not contain
-    * sections named 'description' or 'Description' if the 'description' property
-    * is also set as both will be displayed.
-    * The object is of the form {name:string, content:string}.
-    * Content will be rendered as Markdown with HTML.
-    * This property is observable.
-    * @type {Object[]}
-    * @default []
-    */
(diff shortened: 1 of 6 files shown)


lib/Models/CatalogMember.js
```diff
@@ -264,11 +281,42 @@ defineProperties(CatalogMember.prototype, {
...
+    _infoItemsWithSourceInfoLookup: {
+        get: function() {
+            if (!defined(this._memoizedInfoItemsSourceLookup)) {
+                this._memoizedInfoItemsSourceLookup = this._sourceInfoItemNames.reduce(function(lookupSoFar, name) {
+                    lookupSoFar[name] = true;
+                    return lookupSoFar;
+                }, {});
+            }
+
+            return this._memoizedInfoItemsSourceLookup;
```
RacingTadpole, line 321: Should you add `this._memoizedInfoItemsSourceLookup = undefined;` to the constructor?

AlexGilleran, line 321: Yes.

AlexGilleran:
Updated with that :)

PR #1432, opened 2016-04-08
Made ArcGis copyrightText property show up in info section

### LINKED ISSUE #1188: Add `copyrightText` to ArcGIS MapServer License info page details
Some ArcGIS layers appear to use the key `copyrightText` to put the licensing details in. We should grab that text and add it to the Licensing details on the layer info page.

Here is a service that uses `copyrightText`:

http://www.ga.gov.au/gis/rest/services/topography/National_Electricity_Transmission_Lines/MapServer?f=pjson

AlexGilleran:
Fixes #1188. 

Pretty simple, ArcGis servers have a copyrightText property under both layer and server json, this shows the layer one under "Copyright Text" if it's there, otherwise it shows the server one, otherwise it doesn't add the section at all.


RacingTadpole:
I've already merged it, but can you update CHANGES.md too? (I reckon do it in master directly.)

PR #1437, opened 2016-04-08
Wait for GeoJSON region selection layer to load before hiding it

tephenavies:
When you select a feature during the GeoJSON region selection layer load time of a previously selected feature, an exception is thrown because `_hide` is called on the `GeoJsonCatalogItem` before it has loaded, leading to a `DeveloperError`. Waiting on the `_highlightPromise` before hiding solves this issue.


AlexGilleran:
Would be nice to have a unit test :)

tephenavies:
@AlexGilleran How would I add a reliable unit test for this? Would it not (in some way) rely on one promise resolving before another, which wouldn't be 100% reliable?

AlexGilleran:
It should be possible to control the order of the promises by faking them. I was thinking fake `_load` on the CatalogItem with `when.defer` to simulate the catalog item being in a loading state, then call `_highlightFeature` and `_removeHighlightCallback` to ensure that no error occurs as it did before. Then you can resolve the `_load` promise and make sure the item hides as usual.

tephenavies:
I'm not sure that is possible. The error was thrown from the `GeoJsonCatalogItem.prototype._hide` function:

``` javascript
if (!defined(this._dataSource)) {
        throw new DeveloperError('This data source is not loaded.');
    }
```

To create a working and useful test, I'd have to:
1. Be able to access the feature highlight `GeoJsonCatalogItem` (which only exists as a a local variable of `GlobeOrMap.prototype._highlightFeature` and in the closure of `_removeHighlightCallback`)
2. Replace the loading promise with a defer before it loads from outside the `_highlightFeature` function so that `_dataSource` doesn't get assigned to
3. Then call `_removeHighlightCallback` and see if it errors

Any implementation that I can think of would rely too heavily on implementation details of feature highlighting.

lib/Models/GlobeOrMap.js
```diff
@@ -246,12 +246,14 @@ GlobeOrMap.prototype._highlightFeature = function(feature) {
                   'marker-color': this.terria.baseMapContrastColor
                 };
 
+                var that = this;
                 var removeCallback = this._removeHighlightCallback = function() {
-                    catalogItem._hide();
-                    catalogItem._disable();
+                    that._highlightPromise.then(function() {
+                        catalogItem._hide();
+                        catalogItem._disable();
```
kring, line 268: I think this is still not quite right.  The problem is that `catalogItem._enable()` only gets called if `removeCallback === that._removeHighlightCallback`, which won't be true if a new feature has been selected in the meantime.  So even after waiting for the highlight promise to resolve, there's no guarantee that `_enable` and `_show` have been called, so an exception may still be raised.

However, now that the exception will be thrown inside a `then` callback, it will be silently ignored rather than causing feature picking to abort.  So that's definitely an improvement.

I think the only way to avoid the exception entirely is to track whether the layer has actually been enabled.  It's a bit ugly, but that can be done by slamming a property on the highlightPromise that gets set to true when the layer is enabled.  Maybe you can think of a cleaner way?

kring, line 268: I guess another option is to just catch the exception. That's not too much different from the current situation, but it does at least make it clear that the situation is expected and doesn't break anything. :)

tephenavies:
The original modifications that I made do fix the issue since for a `GeoJsonCatalogItem`, `_hide()` only fails when the item's `_load()` hasn't been called (since `_load` adds the `_dataSource` property and hide throws if this doesn't exist. `_enable` and `_show` don't have to have been called). It's probably not a good idea to rely on this. I think the best solution without adding extra properties to the promise is:

``` javascript
var removeCallback = this._removeHighlightCallback = function() {
    that._highlightPromise.then(function() {
        if (removeCallback !== that._removeHighlightCallback) {
            return;
        }
        catalogItem._hide();
        catalogItem._disable();
    }).otherwise(function(){});
};
```

This will mean that `_hide` cannot be called before `load` finishes, and also won't be called if `_enable` and `_show` weren't called.

kring:
Yeah, good idea, that makes sense to me!

kring:
Can you update changes.md?  Then I'll merge this.

tephenavies:
I added a changelog entry under 3.0.0 because it looks like v3 hasn't been published yet.

PR #1439, opened 2016-04-11
Added ability to set date formats for the timeline in the layer json

AlexGilleran:
Pulled out from the becalmed #1328 for AREMI and GEOGLAM.

`timelineTic` and `currentTime` can be specified in the config JSON, and these override the default format for times in the timeline.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,12 @@
 Change Log
 ==========
 
+### 2.3.1
+
+* Added support for configuration of how time is displayed on the timeline - catalog items can now specify a dateFormat hash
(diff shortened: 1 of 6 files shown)


RacingTadpole:
The usual questions ... how big is the dateFormat package? And how did you choose this one?

AlexGilleran:
6 kB unminified: https://github.com/felixge/node-dateformat/blob/master/lib/dateformat.js

Picked because it was small, very flexible and basically the de facto formatting standard based off it's popularity (5,018,006 people per month can't be wrong, I hope!).

PR #1440, opened 2016-04-11
Use Webpack in the newui branch

kring:
This merges the `webpack` branch into `newui`.

The build will fail on this PR because there are 180 eslint problems.  I suggest we merge it as-is and then start fixing the eslint problems directly in `newui`.  Fixing all the lint problems here would make this PR even harder to review.


AlexGilleran:
:+1:

RacingTadpole:
This works great for me, though I have to admit I haven't looked at the details of the code. @AlexGilleran do you want to review as well?

AlexGilleran:
Yeah I'm all good with it 👍

PR #1446, opened 2016-04-15
Add support for proxyabledomains service.

kring:


lib/Models/Terria.js
```diff
@@ -363,6 +364,16 @@ Terria.prototype.start = function(options) {
         corsProxy.proxyDomains.push.apply(corsProxy.proxyDomains, config.proxyDomains);
         corsProxy.baseProxyUrl = that.configParameters.corsProxyBaseUrl;
 
+        var promises = [];
+
+        promises.push(loadJson(cp.proxyableDomainsUrl).then(function(proxyableDomains) {
```
RacingTadpole, line 369: This means that if `proxyableDomainsUrl` is left at the default, you get this 404 every time you start up: `GET http://localhost:3001/proxyabledomains/ 404 (Not Found)` - are we OK with that?

kring, line 369: You won't once you're using terriajs-server 2.0.0.

lib/Models/Terria.js
```diff
@@ -363,6 +364,16 @@ Terria.prototype.start = function(options) {
...
+            if (proxyableDomains.proxyAllDomains) {
+                corsProxy.isOpenProxy = true;
```
RacingTadpole, line 371: Two minor thoughts here -
- Why use two different names here? I'd vote for using `corsProxy.proxyAllDomains`. The "open" terminology confused me at first - I thought it was referring to opening a file for reading.
- Why are we giving users the power to set this to true at all? Isn't that a bad idea?

kring, line 371: `proxyAllDomains` is incorrect in this context.  It won't proxy all domains, it's just an indication that the server will allow all domains to be proxied.  It's also not a security problem, the security is in the server.  Setting it to true when the server doesn't agree will just make connections fail.

PR #1448, opened 2016-04-15
Support a range of projections for ArcGis MapServer layer extents

### LINKED ISSUE #1447: Support other projections in Esri Mapserver extents
We [currently support a small number of projections](https://github.com/terriajs/terriajs/blob/e72e5fd351b75acb038fced9649785332b634149/lib/Models/ArcGisMapServerCatalogItem.js#L391-L391). Perhaps we could use the proj4 library, which we already require, to support extents in any projection.

Required for https://github.com/NICTA/nationalmap/issues/240

stevage:
Fixes #1447, fixes NICTA/NationalMap#240


kring:
There's an error in this branch when enabling `Land -> Surface Geology` the first time (it works thereafter).  The problem is that it uses wkid 102100, which is not listed in our Proj4Definitions.js.  The quick solution is to add it, of course, but we also need to think about what happens when we don't know how to handle a projection.  It should probably just return undefined as it did before.

kring:
Also please update CHANGES.md.

PR #1459, opened 2016-04-21
Added injection of lat/long into templates data in features

AlexGilleran:
This is a rush job for AREMI - will cherry pick this into master with tests later.


axman6:
Is this getting the location from the feature returned? What we're actually after is the position clicked according to terria, whether the feature has a position or not.

meh9:
@AlexGilleran is it easier to get into master if I leave your original branch?

AlexGilleran:
Doesn't really matter to me - this uses the feature's position if it's available though, should I change it to use the click position always as per @axman6 's comment?

meh9:
So, in the case where there is an existing lat/lon for a feature, as far as we can tell that is always available to us already in the template, so replacing the `terria.coords` with those actually only hides information. Having them separate allows us to decide which one we want. Could we please have the click position always. :)

PR #1463, opened 2016-04-22
Improve previewing in new UI

### LINKED ISSUE #1436: Leaflet preview draws ABS region mapped items differently the first time you preview them
abs preview newui

When you first select the item, it displays data for the "AUS" region map, then when you select it again it displays using "SA4" (which is the default when the dataset is added to the main map).

kring:
This PR eliminates the copy-for-Terria thing we had going on before and instead teaches catalog items how to add themselves to a separate map and then uses that for the preview.  The main benefit of this approach is that catalog items no longer need to be loaded a bunch of times.  A single load is done no matter how many main/preview maps are in use.

Fixes #1436 

I'm going to insta-merge this into newui, but I'm opening it as a PR in case anyone is curious.


kring:
I should mention, the one case where I know this isn't currently working is with time-varying datasets.

PR #1472, opened 2016-04-27
Add `CesiumTerrainCatalogItem`, and a few other things

kring:
This PR starts with #826 but fixes up some things and adds some specs.

It also:
- Switches all builds (even watch) to use the `source-map` webpack devtool instead of `eval-source-map` because breakpoints don't work reliably in the latter (which is unfortunate because it makes incremental builds much faster and the source maps themselves are great, it's just breakpoints that are broken).
- Upgrades to terriajs-cesium 1.20.0.
- Adds a custom webpack loader that removes the debug code (checks that lead to DeveloperErrors) from Cesium.  This should give us a bit of a performance improvement.


RacingTadpole:
Can you add your usual summary of the update to Cesium in CHANGES.md?

RacingTadpole:
When I go to http://localhost:3001/#build/terriajs/test/init/terrain.json, the two new `cesium-terrain` items don't seem to show anything different?

kring:
CHANGES.md is updated.

PR #1475, opened 2016-04-28
Set timeslider init time in config file

rdengate:
This addresses "Start timeslider at specified time including Present" https://github.com/TerriaJS/terriajs/issues/1382. It is a behaviour change as the default position of the time slider when opening a layer is currently "start", but after this change will be "present". It seems like this is what people wanted, though; see discussion in https://github.com/TerriaJS/terriajs/issues/889.


lib/Models/CatalogItem.js
```diff
@@ -167,8 +167,18 @@ var CatalogItem = function(terria) {
      */
     this.maximumShownFeatureInfos = terria.configParameters.defaultMaximumShownFeatureInfos;
 
+    /**
+     * Options for formatting current time and timeline tic labels. See tests for examples.
```
kring: These comments get turned into documentation for the data source editor, which is used by non-developers.  So referring them to the tests doesn't really work, you need to enumerate the options here.  Sorry.

lib/Models/CatalogItem.js
```diff
@@ -167,8 +167,18 @@ var CatalogItem = function(terria) {
...
+     * @type {Object}
+     */
     this.dateFormat = {};
 
+    /**
+     * Options for the value of the animation timeline at start. See tests for examples.
+     * @type {String}
+     */
+    this.timesliderInitTime = undefined;
```
kring: Avoid abbreviations (initial instead of init).

Also it seems slightly odd to mention the timeslider here.  Maybe something like `initialTimeSource`?

lib/ViewModels/AnimationViewModel.js
```diff
@@ -119,6 +124,74 @@ function updateForNewTopLayer(animationViewModel) {
     }
 }
 
+/**
+ * Set time to nearest time to specified (may be start or end of time range if time is not in range).
```
kring: This doc will show in in the generated jsdoc, which doesn't make sense because this function isn't accessible.  Either don't use jsdoc-style comments (regular comments are fine), or add `@private`.  Same with the one below.

lib/ViewModels/AnimationViewModel.js
```diff
@@ -119,6 +124,74 @@ function updateForNewTopLayer(animationViewModel) {
...
+                break;
+            default:
+                // Note that if it's not an ISO8601 timestamp, it ends up being set to present.
+                // Is it an ISO8601 timestamp?
+                var timestamp;
+                try {
+                    timestamp = JulianDate.fromIso8601(startTime);
+                }
+                catch (e) {
+                    throw new DeveloperError('No good timesliderInitTime specified in config file: ' + startTime);
```
kring: `DeveloperError` should only be thrown when a developer uses an API wrong.  The idea is (and Cesium actually does this!) that all DeveloperErrors and related checks can be removed completely from release builds.  Use `TerriaError` or maybe `RuntimeError`.

lib/Models/CatalogItem.js
kring: Actually, as I think about it more, I'm not sure it actually makes sense for this to be specified in `CatalogItem`.  I think this should be in the hands of the individual catalog item types.  For example, for time-dynamic WMSs, showing the latest and starting paused makes tons of sense.  For a CSV, that's almost certainly not what you want; starting at the beginning and playing is more sensible (I think, we don't have too many practical examples of this).  For CZML we don't have to decide, because the file itself specifies.

kring:
Don't forget to update `CHANGES.md` and add yourself to `CONTRIBUTORS.md`!

PR #1491, opened 2016-05-03
Search improvements

### LINKED ISSUE #1155: Test case for limiting search results incorrect
`FlipOutGroup` is defined but never instantiated, so `_load` is never called and consequently the test could pass when it should fail.

https://github.com/TerriaJS/terriajs/blob/6152e4529b4642a932778005857f3b2f206ae9c0/test/ViewModels/CatalogItemNameSearchProviderViewModelSpec.js#L104

### LINKED ISSUE #1325: Don't lock up browser during search
Currently it's pretty painful to use search on NationalMap, because as soon as you type, the browser fires off a lot of requests and gets very busy dealing with the responses. This means that even if the answer you need comes back straight away (eg, it's a dataset available in the catalogue without requiring a server request), you can't really use it because the browser is so locked up.

kring:
Two improvements to search:
- Instead of rapid-firing searches at the geocoder every 250ms or so, we now wait until either the user presses enter or three seconds pass.  This should greatly reduce or Bing Maps transaction usage.
- Made search less likely to lock up the browser by:
  - Only loading one catalog group at a time.
  - Using a much faster method to sort catalog items after load.  Previously we were using [javascript-natural-sort](https://www.npmjs.com/package/javascript-natural-sort) but it's extraordinarily slow.  Now, catalog members have a `nameSortKey` that is computed once (and recalculated if the name is changed) formed by splitting the name into string parts and number parts and comparing each separately.

Fixes #1325 
Fixes #1155

### DIFF
--- a/lib/ViewModels/CatalogItemNameSearchProviderViewModel.js
+++ b/lib/ViewModels/CatalogItemNameSearchProviderViewModel.js
@@ -2,6 +2,7 @@
 
 /*global require*/
 var inherit = require('../Core/inherit');
+var runLater = require('../Core/runLater');
 var SearchProviderViewModel = require('./SearchProviderViewModel');
 var SearchResultViewModel = require('./SearchResultViewModel');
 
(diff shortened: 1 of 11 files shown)


lib/ViewModels/CatalogItemNameSearchProviderViewModel.js
```diff
@@ -165,22 +166,33 @@ function findMatchingItemsRecursively(viewModel, searchInProgress, searchExpress
...
-            if (defined(loadPromise) && item.isLoading) {
-                var searchPromise = loadPromise.then(
-                    findItemsInLoadedGroup.bind(undefined, viewModel, searchInProgress, searchExpression, item, path.slice(), searchFilters)
-                    );
-                searchPromise = searchPromise.otherwise(ignoreLoadErrors);
-                promises.push(searchPromise);
-            } else {
-                findMatchingItemsRecursively(viewModel, searchInProgress, searchExpression, item, path, promises, searchFilters);
-            }
+            promises.push(loadAndSearchGroup(viewModel, item, searchInProgress, searchExpression, path, promises[0], searchFilters));
```
AlexGilleran: So `promises[0]` is passed and waited for to avoid running every promise at once? Do you reckon you could explain that in a comment, took me a while to figure it out.

AlexGilleran:
a

Seems to work really well - just a bit concerned about passing `promise[0]` without explaining why

kring:
Yeah good point, it's a bit hacky.  I'll see if I can clean it up, or comment it at the very least.

kring:
@AlexGilleran this is ready, no more arrays of promises and even some bonus sorting tests.

AlexGilleran:
Looks good :).

... but do the bonus tests pass? ;)

kring:
The kraken has been defeated, ready for another look.

kring:
Ping @AlexGilleran :)

PR #1497, opened 2016-05-04
Add search for WPS services in Csw Catalog Groups

sppigot:
Also add findIndex to avoid breaking IE11

### DIFF
--- a/lib/Models/CswCatalogGroup.js
+++ b/lib/Models/CswCatalogGroup.js
@@ -139,6 +139,12 @@ var CswCatalogGroup = function(terria) {
      */
     this.includeWps = false;
 
+    /**
+     * Gets or sets a regular expression that, when it matches the protocol attribute of a URI element of a record, indicates that the resource is a WPS resource.
+     * @type {RegExp}
+     */
+    this.wpsResourceFormat = /\bwps\b/i;
+
     knockout.track(this, ['url', 'getRecordsParameters', 'includeWms', 'includeKml', 'includeCsv', 'includeEsriMapServer', 'includeGeoJson', 'includeWps']);
 };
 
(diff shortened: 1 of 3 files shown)


lib/Models/CswCatalogGroup.js
```diff
@@ -252,7 +261,13 @@ var resourceFormats = [
 ];
 
 CswCatalogGroup.prototype._load = function() {
-    var postDataTemplate = defaultValue(this.getRecordsTemplate, CswCatalogGroup.defaultGetRecordsTemplate);
+    var postDataTemplate;
+
+    if (this.includeWps) {
```
kring: The way this is set up, if `includeWps` is true, any WMS items are not included, regardless of the state of `includeWms`.

lib/Models/CswCatalogGroup.js
```diff
@@ -447,13 +481,29 @@ Otherwise, if reloading doesn\'t fix it, please report the problem by sending an
...
+    wps.name = title;
+    wps.url = uri;
+
+    return wps.load().then(function() {
+        return wps;
+    });
+}
+
+function addProcessesToGroup(wpsPromise, wpsGroup) {
+    wpsPromise.then(function(wpsItem) {
```
kring: Without an `otherwise`, a WPS load error will silently fail telling the user about it.

kring: The best way to handle this case is to make sure that all promises are returned from `_load`.  Then other parts of TerriaJS will automatically handle errors and display them to the user.

lib/Models/CswCatalogGroup.js
```diff
@@ -599,4 +649,37 @@ function cleanUrl(url) {
     return uri.toString();
 }
 
+function cleanAndProxyUrl(catalogGroup, url) {
+    return proxyCatalogItemUrl(catalogGroup, cleanUrl(url), '1d');
+}
+
+function proxyUrl(terria, url) {
```
kring: This function isn't used.

lib/Models/CswCatalogGroup.js
```diff
@@ -599,4 +649,37 @@ function cleanUrl(url) {
...
+    if (defined(terria.corsProxy) && terria.corsProxy.shouldUseProxy(url)) {
+          return terria.corsProxy.getURL(url);
+    }
+    return url;
+}
+
+// polyFill from ecma6 js
+function findIndex(array, predicate) {
```
kring: It would be good to make this a separate module in `Core`.

kring:
Thanks @sppigot!  Just those few comments.

sppigot:
Thanks @kring! - removal of (old) special case handling for wps stuff should deal with all the comments.

kring:
Thanks @sppigot!  I'm going to merge this via #1525.

PR #1498, opened 2016-05-04
ParameterEditor improvements mainly to date-time handling

sppigot:
Split date time handling for parameter editor between two inputs (sid…e by side)
Default time for easier handling
Also add extra identification to parameters so that inputs don't get confused when switching between
different processes


kring:
Thanks @sppigot!

PR #1501, opened 2016-05-05
Provide csv properties with both old and new names to feature info panel

### LINKED ISSUE #1483: Allow old and new names in FeatureInfoTemplate (with TableStyle.columns...name)
If you have a csv column titled `funding` and want to display it as `Funding ($)`, you can use:

```
      "tableStyle": {
        "columns": {
          "funding": {
            "name": "Funding ($)"
          }
      }
```

But then you need to use `{{Funding ($)}}` in `FeatureInfoTemplate`, which is inconvenient and prone to break.

We should allow both `{{funding}}` and `{{Funding ($)}}`.

Note this will mean swapping names will be impossible, but that seems a small price to pay.

RacingTadpole:
Fixes #1483.

This works (test it with the new item at http://localhost:3001/#build/terriajs/test/init/test-tablestyle.json)... but there’s a snag: if both old and new names for a property are available to the feature info, then when you “Download as csv”, you wind up with separate columns for the two.

Solving that problem would require introducing some new complexity - a new category of property that can be accessed by templates but does not appear in the download.

Alternatively we could change it so that templates can only use the original name. In that case, the downloaded csv would also have the original column names. (In the current master, they have the new names.)

@axman6 ?


axman6:
Hmm, this does seem difficult. We should discuss this next week.

RacingTadpole:
Yes, a better way might be to tell the feature info panel that the template should populate some aliases, rather than duplicating the properties. That could be via a custom property of "properties". (Eg. I set `properties._terria_key_aliases = {oldName1: newName1, oldName2: newName2}`.
Of course, the downside is now you can never have a property called "_terria_key_aliases".

axman6:
I think that's a cost we're willing to pay =)

RacingTadpole:
I've now implemented this by passing the map from the csv logic to the feature info panel via `properties._terria_columnAliases`. What do people think?
In particular, given that we already overwrite `properties.terria` to add the picked lat and lon, it might be cleaner to adapt that instead of adding a new one.

RacingTadpole:
Also note - when this is merged into master, we'll need to make corresponding changes to the `newui` feature info panel.

kring:
This looks fine to me.  Can you update CHANGES.md and remove the commented-out code, and then I'll merge it?  Assuming @axman6 is happy with it of course.

axman6:
Just checking, what does this mean for using this? Which column names are used in the data download? And which names are accessible in templates?

RacingTadpole:
Both old and new are available in the templates. The new names are used as the column headings in the data download.

axman6:
Great, sounds good to me, happy to close the issue.

PR #1503, opened 2016-05-05
Allow WebMapServiceCatalogItem to override pickFeatures for GetFeatureInfo

sppigot:
Also add function to build WPS process for pickFeatures override - triggered by catalog with reference to a WPS URL template

### DIFF
--- a/lib/Models/CswCatalogGroup.js
+++ b/lib/Models/CswCatalogGroup.js
@@ -354,7 +354,7 @@ CswCatalogGroup.prototype._load = function() {
                     var excludedProtocol = false;
                     for (var l = 0; l < resourceFormats.length; l++) {
                         var f = resourceFormats[l];
-                        if (url.protocol.match(that[f[0]])) {
+                        if (url.protocol.match(that[f[0]]) && that[f[1]]) {
                             excludedProtocol = true;
                             acceptableUrls.push(url);
                         }
--- a/lib/Models/CswCatalogGroup.js
+++ b/lib/Models/CswCatalogGroup.js
@@ -6,6 +6,7 @@ var URI = require('urijs');
 var ArcGisMapServerCatalogItem = require('./ArcGisMapServerCatalogItem');
 var CatalogGroup = require('./CatalogGroup');
 var CsvCatalogItem = require('./CsvCatalogItem');
+var createPickFeaturesUsingWebProcessingService = require('./createPickFeaturesUsingWebProcessingService');
 var GeoJsonCatalogItem = require('./GeoJsonCatalogItem');
 var inherit = require('../Core/inherit');
 var KmlCatalogItem = require('./KmlCatalogItem');
@@ -14,6 +15,7 @@ var proxyCatalogItemUrl = require('./proxyCatalogItemUrl');
 var TerriaError = require('../Core/TerriaError');
 var WebMapServiceCatalogGroup = require('./WebMapServiceCatalogGroup');
 var WebMapServiceCatalogItem = require('./WebMapServiceCatalogItem');
+var WebProcessingServiceCatalogGroup = require('./WebProcessingServiceCatalogGroup');
 var xml2json = require('../ThirdParty/xml2json');
 
 var clone = require('terriajs-cesium/Source/Core/clone');
@@ -130,10 +132,18 @@ var CswCatalogGroup = function(terria) {
      */
     this.domainSpecification = undefined;
 
-    knockout.track(this, ['url', 'getRecordsParameters', 'includeWms', 'includeKml', 'includeCsv', 'includeEsriMapServer', 'includeGeoJson']);
+    /**
+     * True to allow OGC:WPS service resources to be added to the catalog; otherwise, false.
+     * @type {Boolean}
+     * @default false
+     */
+    this.includeWps = false;
+
+    knockout.track(this, ['url', 'getRecordsParameters', 'includeWms', 'includeKml', 'includeCsv', 'includeEsriMapServer', 'includeGeoJson', 'includeWps']);
 };
 
 CswCatalogGroup.defaultGetRecordsTemplate = require('./CswGetRecordsTemplate.xml');
+CswCatalogGroup.wpsGetRecordsTemplate = require('./CswGetRecordsWPSTemplate.xml');
 
 inherit(CatalogGroup, CswCatalogGroup);
 
@@ -240,7 +250,7 @@ CswCatalogGroup.defaultSerializers.geoJsonResourceFormat = regexSerializer('geoJ
 freezeObject(CswCatalogGroup.defaultSerializers);
 
 CswCatalogGroup.prototype._getValuesThatInfluenceLoad = function() {
-    return [this.url, this.filterQuery, this.blacklist, this.filterByWmsGetCapabilities, this.minimumMaxScaleDenominator, this.allowEntireWmsServers, this.includeKml, this.includeWms, this.includeCsv, this.includeEsriMapServer];
+    return [this.url, this.filterQuery, this.blacklist, this.filterByWmsGetCapabilities, this.minimumMaxScaleDenominator, this.allowEntireWmsServers, this.includeKml, this.includeWms, this.includeCsv, this.includeEsriMapServer, this.includeWps];
 };
 
 var resourceFormats = [
@@ -252,7 +262,13 @@ var resourceFormats = [
 ];
 
 CswCatalogGroup.prototype._load = function() {
-    var postDataTemplate = defaultValue(this.getRecordsTemplate, CswCatalogGroup.defaultGetRecordsTemplate);
+    var postDataTemplate;
+
+    if (this.includeWps) {
+        postDataTemplate = defaultValue(this.getRecordsTemplate, CswCatalogGroup.wpsGetRecordsTemplate);
+    } else {
+        postDataTemplate = defaultValue(this.getRecordsTemplate, CswCatalogGroup.defaultGetRecordsTemplate);
+    }
 
     var that = this;
     var startPosition = 1;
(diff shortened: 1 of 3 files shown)


kring:
@sppigot I'm a little hesitant on this one because it seems so specific to your situation.  A WMS where we prefer GetFeatureInfo to use WPS, and that WMS is discovered via a CSW query, and the CSW record describes, in a pretty bespoke way, that this is the desired behavior.

I can see two ways of looking at this:
1. It's _generally useful_ for a layer of one type to query feature info using a totally different type of service.  In this case, we should generalize it, so we can use it for, e.g., a WMS that gets feature info via WFS or an ArcGIS MapServer that gets feature info via WMS GetFeatureInfo.  Or we can even plug in a totally custom feature info provider to use some bespoke interface.
2. WMS layers should always query feature info the WMS way.  If you want to implement the WMS GetFeatureInfo on your server to actually do a WPS query behind the scenes and then wrap it up in a regular WMS GetFeatureInfo response, great!  But it's outside the scope of TerriaJS to get involved in that.

I'm kind of leaning toward 2.  Maybe you can convince me otherwise by telling me more about your use case?

sppigot:
@kring Thanks for the response, Uwe and I have been thinking about a generic approach (basically point 1 in your response) using the catalog. 
- The catalogue already has the concept of describing services->operations->parameters. eg. a WMS with a GetFeatureInfo operation and parameters of that operation (eg. REQUEST etc).
- The service descriptions also include the layers that the services act upon - see abridged example service record below.
- The generic approach would be for TerriaJS to query the catalogue for its list of services with a particular operation and filter using the layer its want to apply the operation to - it will receive an ISO19119 record.
- TerriaJS could then use this and info from the layer metadata record to build a request for the particular operation. If no result then just fall back to standard operation if present.
- This could also be applied to other services already used and services to come.
- An additional use case and potential advantage could be that other processing services (like yours concerned with the ABS data) could interrogate layers if they provide the data necessary to be relevant.

Example service record - somewhat abridged but you get the idea:

``` xml
....

<srv:containsOperations>
   ..
   <srv:operationName>GetFeatureInfo</srv:operationName>
   <srv:operationDescription>..</srv:operationDescription>
   <srv:parameters>
     ..
     <srv:name>
         <gco:aName>LatLonPosition</gco:aName>
         <gco:TypeName>GeoJSON</gco:TypeName>
     <srv:name>
     ..
     <srv:optionality>required</srv:optionality>
     <srv:repeatability>false</srv:repeatability>
    ..
   </srv:parameters>
   <srv:parameters>
      ... similar description for layerName parameter, aName=LayerName, typeName=TEXT ...
   </srv:parameters>

   <srv:connectPoint>
      <gmd:URL>http://oa-gis.csiro.au/wps?version=1.0.0&request=Execute&service=WPS&Identifier=get-value-at-point</gmd:URL>
      <gmd:protocol>OGC:WPS</gmd:protocol>
   </srv:connectPoint>
</srv:containsOperations>

....

<srv:operatesOn uuidref="0dc35208-2f5b-3a74-e053-12a3070a69ed"/>

....


```

The reasoning behind taking the extra steps for this generic approach here is twofold:
1. ​​The catalog approach is declarative - anyone using TerriaJS can set up the records in their catalogue in the prescribed manner and use the steps here to provide an alternative implementation for any of the current operations eg. GetFeatureInfo and GetLegendGraphic - some users eg. GA already have their records set up in this manner and others (eg. IMOS) would like to. Note that this generic approach would also replace the GetLegendGraphic 
2. There are quite a lot of people who want to do this (eg. when GetFeatureInfo returns useless RGB values). They are currently using server side hacks such as intercepting the GetFeatureInfo request and changing the name of the layer.... :-( A declarative approach using the metadata catalogue to solving this would seem to be much better and another advantage of using the TerriaJS client.
3. What we are doing here is to use service metadata from the catalog to provide a "less bespoke", more generic approach to overriding the default behaviour.

Challenge is to store the layer information in a consistent manner eg. layer name in the example above. More to come on that.

kring:
Thanks for the explanation @sppigot!  I'm still not really convinced, though.  IMO, WMS GetFeatureInfo is the _right_ way to get feature info on click from a WMS.  It takes into account the multiple levels of detail that a WMS image may be rendered at and the size of a pixel.  E.g. for a point, you don't have to click on the point's exact latitude and longitude; simply clicking on a pixel that the point covers is sufficient.

Also, using GetFeatureInfo enables maximum compatibility: If I grab that WMS URL from a TerriaJS app and put it in a GIS tool, I'll get a similar experience to what I got in TerriaJS.  That's one of the purposes of a standard like WMS.  If we instead invent a way to describe arbitrary operations-on-click in a metadata record, only TerriaJS is going to support that.

I'm sympathetic to the fact that it's a pain with current geospatial servers to provide a custom implementation of `GetFeatureInfo`, but I'd still argue that it makes more sense to open a pull request into Geoserver (for example) to fix that rather than hack around it in one particular client (TerriaJS).  And I also think that the server-side hack you described to change the layer name is much better than doing something fancy in the client, because it benefits all clients.

sppigot:
Thanks @kring! All of which I agree with - in a perfect world GetFeatureInfo is the preferred way to do all this and who would do it any other way if they had the choice? but:
- I have investigated the charts approach Mats mentioned and (unless I'm missing something) I can't see it working for this use case as 1. the wps url we use needs the x,y from the getfeatureinfo call and the layer name - Mats is using stuff that comes back from the GetFeatureInfo in his urls (eg. an id of the power station whatever) - we don't have this coming from our GetFeatureInfo call, 2. even if we did have access to these we would still need code to handle the XML from the wps etc - not sure how that would/could be config'd - but please let me know if there is another way we might be able to adapt this to suit....
- standards are notorious for being rather constrained and it is often helpful for clients to have documented ways in which some of these constraints can be removed/relaxed 
- to return to the nub of your original argument: that what we are proposing is a bespoke solution (ie. too specific to our use case): might be helpful to examine what the specifics are: records in the catalog must be set up in a certain way to use it ie. with a GetFeatureInfo online resource that points to the WPS process and that a WPS handler exists within TerriaJS to do the getFeatureInfo (ie. the original pull request) 

But one of the main ideas of this approach is that it builds on the other catalogue features that have already been accepted into TerriaJS ie. 
1. get records from the catalog and classify them according to the keyword with hierarchy (CswCatalogGroup with GetDomain call etc)
2. GetLegendGraphic pointing to an image URL and overriding the default GetLegendGraphic URL constructor in TerriaJS/Cesium

These functions also require the records in the catalog to be set up in a certain way and I don't see why the GetFeatureInfo WPS handler is any different? In its favour: it is a declarative approach, based on the WPS standard, uses a template URL so it is quite flexible (doesn't add much code either :-)). Or to put it another way: if this is too specific to our use case, why were the other catalogue functions already added to TerriaJS, that have similar catalog requirements, not treated in the same way?

meh9:
@sppigot Regarding getting x,y to use in a WPS call, we recently made it possible to get the lat/lon of the clicked point in the template without having to have it returned in the feature info call, see: https://github.com/TerriaJS/terriajs/issues/1438 which we use here to generate a CSV Export URL: https://github.com/NICTA/aremi-natmap/blob/2016-05-11/datasources/aremi/custodian/Bureau_of_Meteorology/Solar_Satellite-template.ejs#L5

There is an optional WPS HTTP GET request that your WPS server may or may not support: http://geoprocessing.info/wpsdoc/1x0ExecuteGET

sppigot:
Thanks @meh9! - I hadn't seen that and yes we are using the GET request on the WPS already.

This is likely still not as nice as our existing pull request however as it will very likely require an extra button (like the downloads @meh9?), injecting the layer name as well, possibly some changes to the WPS process (it accepts geojson) and some styling of the wps response. By comparison, the original pull request is flexible and useful for us and others and very straightforward - it grabs the response from the wps and sticks it into the getfeatureinfo panel as json.

Sorry to be a bit pushy about this @kring but it would be nice if you could indicate why you object to the original pull request as per my comment above, in particular: that this pull request is flexible and useful for others if they set up their catalogue appropriately and is thus consistent with the other catalogue based extensions already accepted into terriajs.

kring:
@sppigot you're right that it's not very different from the keyword hierarchy classification thing, and I did have similar concerns about that.  The major difference is that the current problem seems fairly easy to solve on the server using existing standards.  In contrast, I didn't see a better/existing way of grouping CSW records.

When I say it's fairly easy to solve on the server, I'm specifically suggesting using common web server features to redirect the `GetFeatureInfo` request to a separate service that you write.  That service can do a WPS request (driven by metadata in a CSW record if desired) and return the result as a normal GetFeatureInfo response.  I don't think it would be much harder to implement there than in TerriaJS, and then it would benefit all clients.

But let me take another look at the PR...

kring:
@sppigot I'm finally coming back to this and I'm afraid I'm still not convinced.  My objection has two parts:
1. This can be relatively easily done on the server, and doing it there has the advantage that it will then work for all clients because it will follow the WMS standard.  I don't think you've addressed this point yet.
2. The implementation in this PR, while relatively simple, makes assumptions that are specific to your use case.

Specifically, on point 2, this PR checks for a URL in the metadata record with the description `GetFeatureInfoUrl` and then assumes that the service at that URL is not an actual WMS GetFeatureInfo, but is instead a WPS service with very specific inputs and outputs, when other similar inputs and outputs are totally reasonable (e.g. encoding the point as GeoJSON is valid but arbitrary).

Of course this can be generalized, and you've proposed that previously (https://github.com/TerriaJS/terriajs/pull/1503#issuecomment-218968853).  But that's drastically increasing the complexity, and at that point I come back to point 1 above: why not just wrap your service up in the standard WMS GetFeatureInfo on the server rather than requiring some complicated behavior from any client that wants to access your service?

PR #1510, opened 2016-05-09
Include "engines" attribute in package.json.

stevage:


package.json
```diff
@@ -3,6 +3,10 @@
   "version": "3.0.0",
   "description": "Geospatial data visualization platform.",
   "license": "Apache-2.0",
+  "engines" : { 
+    "node" : ">= 4.0.0",
```
kring: Are you sure this is correct?  We just saw problems on @chloeleichen's computer with 5.0.0, and I thought you had previously seen problems with 5.9.x even.  I know 5.10.x and 5.11.x work.

stevage:
Yeah, thanks - not much harm in being conservative. In reality I don't think the engine attribute gets used by much.

PR #1521, opened 2016-05-11
Revert Leaflet from 0.7.7 to 0.7.3

### LINKED ISSUE #1496: Csv feature selection broken in Leaflet 0.7.7
Issue #1121 has reoccured because Leaflet has been updated from 0.7.3 to 0.7.7.

To reproduce, open any csv region mapped layer (such as `ABS -> 2011 Census -> Age` or `#test -> Region Mapped (CSV) -> 2011 Census AUST (SA4)`), switch to 2D mode and click on regions a few times. Only the first 2 clicks regiester.
leaflet 0 7 7 feature selection

The previous fix #1151 was to downgrade Leaflet to 0.7.3, so to fix we could try either:
- Reverting to Leaflet 0.7.3 or 
- Fixing the underlying issue (if we need Leaflet 0.7.7)

tephenavies:
Fixes #1496.

Here's a comparison between Leaflet 0.7.3:
leaflet 0 7 3
And Leaflet 0.7.7:
leaflet 0 7 7

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "imports-loader": "^0.6.5",
     "javascript-natural-sort": "^0.7.1",
     "json-loader": "^0.5.4",
-    "leaflet": "0.7.7",
+    "leaflet": "0.7.3",
     "less": "^2.6.1",
     "less-plugin-npm-import": "^2.1.0",
     "linkify-it": "^1.2.0",


kring:
0.7.3 seems to work fine on high dpi displays, too.  We'll have to see if it helps with the preview problems in new UI once we merge it over there.

PR #1530, opened 2016-05-12
AnyValue now handled in WPS service.

rdengate:
LiteralValue can provide AnyValue rather than AllowedValues, which wasn't previously handled. It's now handled by creating a generic parameter which is edited with the GenericParameterEditor. 

Also, if output value is ComplexData, handle that by creating a chart.

### DIFF
--- a/lib/Models/GenericParameter.js
+++ b/lib/Models/GenericParameter.js
@@ -1,0 +1,43 @@
+'use strict';
+
+/*global require*/
+var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
+var FunctionParameter = require('./FunctionParameter');
+var inherit = require('../Core/inherit');
+
+/**
+ * A parameter that can be any value.
+ *
+ * @alias GenericParameter
+ * @constructor
+ * @extends FunctionParameter
+ *
+ * @param {Object} options Object with the following properties:
+ * @param {Terria} options.terria The Terria instance.
+ * @param {String} options.id The unique ID of this parameter.
+ * @param {String} [options.name] The name of this parameter.  If not specified, the ID is used as the name.
+ * @param {String} [options.description] The description of the parameter.
+ * @param {Boolean} [options.defaultValue] The default value.  If not specified, the first possible value is the default.
+ */
+var GenericParameter = function(options) {
+    FunctionParameter.call(this, options);
+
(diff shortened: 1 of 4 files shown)


lib/Models/GenericParameter.js
```diff
@@ -0,0 +1,43 @@
...
+ * @extends FunctionParameter
+ *
+ * @param {Object} options Object with the following properties:
+ * @param {Terria} options.terria The Terria instance.
+ * @param {String} options.id The unique ID of this parameter.
+ * @param {String} [options.name] The name of this parameter.  If not specified, the ID is used as the name.
+ * @param {String} [options.description] The description of the parameter.
+ * @param {Boolean} [options.defaultValue] The default value.  If not specified, the first possible value is the default.
+ */
+var GenericParameter = function(options) {
```
kring: We have a `GenericParameterEditor` as a catch-all, but for the actual parameter types I think it would be better to be specific.  This isn't a generic parameter, it's a `TextParameter`, right?

kring: Actual `StringParameter` is probably better.

lib/Models/WebProcessingServiceCatalogItem.js
```diff
@@ -155,14 +155,26 @@ WebProcessingServiceCatalogItem.prototype._load = function() {
...
                 continue;
             }
+            if (defined(outputs[i].Data.LiteralData)) {
+                that.shortReportSections.push(new ShortReportSection({
+                    name: outputs[i].Title,
+                    content: formatOutputValue(outputs[i].Title, outputs[i].Data.LiteralData)
+                }));
+            } else if (defined(outputs[i].Data.ComplexData)) {
+                var content = outputs[i].Data.ComplexData;
+                if (outputs[i].Identifier === "csv" || outputs[i].Identifier === "json")
```
kring: `Identifier` is more like the name of the output than it's type, right?  It seems questionable to assume that all outputs named `json` or `csv` are charts, and outputs with other names are not.  Is there a better way to do this?

kring: The WPS spec says `ComplexData` has a mandatory `Format` property, so that seems like the thing to use.

kring:
Looks good other than those comments!

PR #1556, opened 2016-05-19
Add WFS support to CKAN catalog items and groups

### LINKED ISSUE #1553: WFS items are not supported from CKAN
Eg. Even after changing a CKAN item's json to read:

```
            "type":"ckan",
            "filterQuery":[
                "fq=+(res_format%3aWFS%20OR%20res_format%3awfs%20OR%20res_format%3awms%20OR%20res_format%3aWMS%20OR%20res_format%3acsv-geo-au%20OR%20res_format%3aaus-geo-csv)"
            ],
```

It still doesn't show up under the CKAN group item. We simply haven't implemented it in CkanCatalogGroup or CkanCatalogItem.

RacingTadpole:
Adds WFS support to CKAN. Fixes #1553.

In fact, in most cases, you don't really want to use this. You really want to pretend that WFS servers are WMS servers, because most of them double up as WMS, and WMS is better.

You can do that by using init json like so - note the `filterQuery` and `wmsResourceFormat` lines:

```
{
    "catalog": [
        {
            "blacklist": {...},
            "filterByWmsGetCapabilities": false,
            "filterQuery": [  // Only gets resources with format = WFS, wfs, WMS, wms and csv-geo-au
                "fq=+(res_format%3aWFS%20OR%20res_format%3awfs%20OR%20res_format%3awms%20OR%20res_format%3aWMS%20OR%20res_format%3acsv-geo-au)"
            ],
            "url": "example.com",
            "type": "ckan",
            "name": "Name",
            "includeCsv": true,
            "allowEntireWmsServers": true,  // optionally
            "useResourceName": true,  // optionally
            "wmsResourceFormat": "(wms|wfs)$"
        }
    ]
}
```

Still, if you ever do want to use WFS, you'd need this PR.

### DIFF
--- a/lib/Models/CkanCatalogGroup.js
+++ b/lib/Models/CkanCatalogGroup.js
@@ -113,12 +113,12 @@ var CkanCatalogGroup = function(terria) {
     this.useResourceName = false;
 
     /**
-     * True to allow entire WMS servers (that is, WMS resources without a clearly-defined layer) to be
-     * added to the catalog; otherwise, false.
+     * True to allow entire WMS AND WFS servers (that is, WMS/WFS resources without a clearly-defined layer) to be
+     * added to the catalog; otherwise, false. (Yes, this is poorly named.)
      * @type {Boolean}
      * @default false
      */
-    this.allowEntireWmsServers = false;
+    this.allowEntireWmsServers = false;  // and WFS servers too.
 
     /**
      * True to allow WMS resources to be added to the catalog; otherwise, false.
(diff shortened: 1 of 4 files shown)


lib/Models/CkanCatalogGroup.js
```diff
@@ -113,12 +113,12 @@ var CkanCatalogGroup = function(terria) {
...
     /**
-     * True to allow entire WMS servers (that is, WMS resources without a clearly-defined layer) to be
-     * added to the catalog; otherwise, false.
+     * True to allow entire WMS AND WFS servers (that is, WMS/WFS resources without a clearly-defined layer) to be
+     * added to the catalog; otherwise, false. (Yes, this is poorly named.)
      * @type {Boolean}
      * @default false
      */
-    this.allowEntireWmsServers = false;
+    this.allowEntireWmsServers = false;  // and WFS servers too.
```
kring: It'd be cleaner to rename this.  Then add a new property named `allowEntireWmsServers` which just passes through to the new one with a `deprecationWarning` along the way.

kring:
Looks good aside from the one nit.

lib/Models/CkanCatalogGroup.js
kring: Or, have a separate property for WFS?

RacingTadpole: True, you are voicing something I knew to be true but which I tried to ignore... :-)
The reason I hesitated is that `CkanCatalogItem.createCatalogItemFromResource` takes an option called `allowGroups`, and I didn't want to change its interface to split out `allowWmsGroups` and `allowWfsGroups`.
So what I'll do is: change it to this, but allow the old `allowGroups` as a synonym for `allowWmsGroups`, but deprecate it.
Perfect.

RacingTadpole:
Thanks @kring, ready to go (when travis is done).

PR #1560, opened 2016-05-20
Fixed loading from shortened URL not working in IE9

### LINKED ISSUE #1550: Shortened share URL doesn't work for IE9
It appears that using a shortened share URL with IE9 doesn't work, instead producing an error like:

pasted_image_at_2016_05_18_05_29_pm

This was the share URL: http://nationalmap.gov.au/renewables/#share=OJYjXl

(Note that unless #1490 is fixed, once the share URL bug is fixed you will still get Catalogue Service error.)

AlexGilleran:
Fixes #1550 

Problem occurred because IE9 has to proxy the URL expand call, but it would make that call before it had a list of proxyable domains, so when it determined whether it should proxy the call `proxyableDomains === []` and it didn't proxy it, so the call would fail.

Now I've made it so that the call to try to load the catalog from a link happens after the call to get `proxyableDomains` is complete.

In the process I made it so more of the proxying logic was within `CorsProxy` itself - now it initialises itself and it has a method that encapsulates the "if I should proxy then proxy otherwise use the original url" logic that's found throughout Terria.

Needs some more tests/jsdoc but I need to get more fixes done before Monday for the AREMI release, will follow up later.


RacingTadpole:
Nice one. I haven't _actually_ checked this works in IE9, but it doesn't seem to break anything anywhere else. I like your improvements to CorsProxy too. Only prepare for case-insensitivity nightmares on Macs...

PR #1567, opened 2016-05-23
Added an undefined check in region mapping for no properties.

AlexGilleran:
I've got no idea what this really does or how it works or how feature properties can be undefined and yet still display - maybe @RacingTadpole could explain this to me when he's back? But this fixes up feature picking when you've got both a region mapped dataset and a vector point dataset at the same time.


kring:
@RacingTadpole can you review this one?

RacingTadpole:
That's funny. This is exactly the same as #1600, which is already merged. Except that I didn't update the README in that one.

RacingTadpole:
BTW - the situation in which this mattered for me, was when some regions had data, but others did not. If you clicked on a region that had no data, it caused an error. (But only if it was non-time-varying data, because that was handled properly already.)

Issue #1578, opened 2016-05-31
Lat/Lon in templates don't work in Safari


axman6:
It seem's @RacingTadpole work to make the lat/lon clicked available in feature info templates doesn't work properly in Safari. I believe this used to work but has now broken. Can be tested on http://preprod.nationalmap.gov.au/renewables/ via `Renewable energy -> Solar -> Solar Satellite DNI - 2014`, then clicking on any location. The two URLs listed should end in `v1/DNI/<lat>/<lon>` but in safari they are `v1/DNI//`.

RacingTadpole:
Interesting - when I compare going to the solar satellite DNI data in Chrome and Safari, I get a different problem: the timeline starts at different dates - Chrome at 1/1/2015, Safari at 5/10/2014 (at which time all readings are zero). These are in fact the end dates of the data - they appear to be different ranges in each browser.
When they're all zero, I get the problem you state above - and I get it in both Chrome and Safari. But at other times, the URLs work in Safari.

So I think there are two problems:
- missing/zero data leads to missing lat/lon (for some reason!)
- Safari's different choice of a start date

Plus, it needs to implemented in `newui`.

AlexGilleran:
This was fixed :)

PR #1595, opened 2016-06-06
Add lat/lon to feature data when it updates with time

RacingTadpole:
This fixes the first part of #1578.

There were two compounding problems 
- a zero property value erroneously led to the property being considered time-varying,
- and lat-lon was not added to time-varying data.


AlexGilleran:
Looks good - could you add a test? Should be as simple as setting `terria.pickedFeatures.properties` to `undefined` and making sure the lat/lng still work :).

Also changes.md

PR #1601, opened 2016-06-07
Changed WMS interval creation to use JulianDates instead of JS Dates.

AlexGilleran:
This resolves a problem in Safari where any WMS item with an interval inside an hour of time that doesn't exist because of DST (e.g. Oct 5 2014 2am - 3am in Sydney) can't show data beyond that time, because a call to the JS date `setHours` (or minutes etc) function in Safari that would put the new value of the date inside one of these non-existent time intervals is simply ignored, so the time never advances beyond it.

This would normally deserve a test but I can't think of a good way to write one, given that it depends on being in a certain timezone and now the code doesn't use timezone-sensitive dates anymore there's no way to reach in and set one.


RacingTadpole:
Nice one - the code looks simpler too!

PR #1612, opened 2016-06-09
NewUI: Check if nextLayer is defined before accessing properties

rdengate:
If you have a catalog item that is enabled by default in the config, it doesn't load in the newui, but it does in master.

I've tested this change with GEOGLAM and it works.

### DIFF
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -546,10 +546,9 @@ function onClockTick(catalogItem, clock) {
         setOpacity(catalogItem, catalogItem._nextLayer, catalogItem.opacity);
         fixNextLayerOrder(catalogItem);
         ImageryLayerCatalogItem.disableLayer(catalogItem, catalogItem._imageryLayer);
-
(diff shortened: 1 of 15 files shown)


lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -547,8 +547,11 @@ function onClockTick(catalogItem, clock) {
         fixNextLayerOrder(catalogItem);
         ImageryLayerCatalogItem.disableLayer(catalogItem, catalogItem._imageryLayer);
 
-        catalogItem._imageryLayer = catalogItem._nextLayer;
-        catalogItem._imageryLayer.imageryProvider.enablePickFeatures = true;
+        if (defined(catalogItem._nextLayer))
+        {
+            catalogItem._imageryLayer = catalogItem._nextLayer;
```
kring: I think this line should be outside the `if`.  We don't want the previous layer lingering around if there's no next layer.

rdengate: Yep, good idea.

PR #1616, opened 2016-06-09
Enumeration color binning improvements

### LINKED ISSUE #1538: Legend and coloring of ENUM columns improvements reduce functionality
The fixes implemented in https://github.com/TerriaJS/terriajs/pull/1384 have removed the ability to distinguish more than 8 string encoded variables.  Eg in the attached file, the column RDA_NAME is the display column for colouring (unsophisticated and n-color unaware as it is), and only displaying 8 of the 55 regions isn't sufficient.  Maybe we need to think harder about the algorithm to choose bins - perhaps {colorBins: 9} is overly prescriptive, and maybe should only be applied when row-count = unique-values?

Can we discuss, in light of requirements for ABR, as I've now run into this problem too!!

image

[RDA_LGA_2015_16.txt](https://github.com/TerriaJS/terriajs/files/266007/RDA_LGA_2015_16.txt)

kring:
Fixes #1538 

Enumeration in `colorBinMethod: top` mode (default):

image

Enumeration in `colorBinMethod: cycle` mode:

image

### DIFF
--- a/lib/Models/LegendHelper.js
+++ b/lib/Models/LegendHelper.js
@@ -339,32 +339,8 @@ function buildBinColors(legendHelper, colorBins) {
     }
 
     // Convert the output formats of the binning method into our format.
-    if (method === 'top') {
-        if (binCount === 1) {
-            binColors.push({
-                upperBound: 1,
-                colorArray: getColorArrayFromColorGradient(colorGradient, 0.5)
-            });
-        } else {
-            var numberOfSpecificColors = numericalValues.length;
-            if (numericalValues.length > colorBins) {
-                numberOfSpecificColors = colorBins - 1;
-            }
-            for (i = 0; i < numberOfSpecificColors; i++) {
-                binColors.push({
-                    upperBound: i + 0.5,
-                    colorArray: getColorArrayFromColorGradient(colorGradient, i / (binCount - 1))
-                });
-            }
-            if (numericalValues.length > colorBins) {
-                // Add an "other" category.
-                binColors.push({
-                    upperBound: tableColumn.uniqueValues.length + 1,
-                    colorArray: getColorArrayFromColorGradient(colorGradient, 1)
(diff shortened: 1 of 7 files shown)


stevage:
Kind of an interesting edge case. I added a test case with no explicit mode, and selected the 'LGAname' column:

screenshot 2016-06-10 10 57 26

Thanks to the ambiguous name "Campbelltown (C)" there are 567 unique values out of 568 rows. So it falls back to the "unique colours for most common values" method.

kring:
There's currently no special behavior when #rows=#unique values, so you'd get that even without the Cambelltown thing.

lib/Models/LegendHelper.js
```diff
@@ -466,18 +505,32 @@ function buildLegendProps(legendHelper, colorMap) {
...
+                title: defaultVariousValuesEnumLabel,
+                multipleColors: binColors.map(function(bin) {
+                    return convertColorArrayToCssString(bin.colorArray);
+                })
+            }];
+        } else {
+            items = reversedBinColors.map(function(bin, index) {
+                var label = tableColumn.uniqueValues[reversedBinColors.length - 1 - index];
+                // It would probably be cleaner to combine the binColor calculation and this calculation, so we don't have to do this test in each place.
+                if (tableColumn.uniqueValues.length > binColors.length && legendHelper._cycleEnumValues) {
```
stevage: Something strange here - testing if `&& legendHelper._cycleEnumValues` but we're on the false side of that exact test above.

kring: Quite right! :monkey:
I'll clean it up.

kring: That's leftover code from when I was putting "Various Values" as the label for each of the colors, rather than sticking all  the colors in one box.

kring:
ok, all set.

PR #1619, opened 2016-06-10
Master: Use user-provided style rather than picking first

rdengate:
Styles can be passed as a parameter to style the WMS layer, but it is ignored for GetLegendGraphic, which currently uses the URL from the first style if a list is provided. Now it uses the specified style if provided, but otherwise chooses the first style as it previously did.

GEOGLAM requires this change in both master and newui. I'm intending to submit a separate PR for merge into newui.

### DIFF
--- a/lib/Models/WebMapServiceCatalogItem.js
+++ b/lib/Models/WebMapServiceCatalogItem.js
@@ -409,8 +409,24 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
     }
 
     var legendUri, legendMimeType;
-    // There can be multiple styles - just get the first if so.
+    // If style is defined in parameters, use that, but only if a style with that name can be found.
+    // Otherwise use first style in list.
     var style = Array.isArray(thisLayer.Style) ? thisLayer.Style[0] : thisLayer.Style;
+    if (defined(this.parameters.styles)) {
+        var styleName = this.parameters.styles;
+        if (Array.isArray(thisLayer.Style)) {
+            for (var ind=0; ind<thisLayer.Style.length; ind++) {
+                if (thisLayer.Style[ind].Name === styleName) {
+                    style = thisLayer.Style[ind];
+                }
+            }
+        } else {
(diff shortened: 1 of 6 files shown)


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -409,8 +409,24 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
     }
 
     var legendUri, legendMimeType;
-    // There can be multiple styles - just get the first if so.
+    // If style is defined in parameters, use that, but only if a style with that name can be found.
```
kring, line 412: We don't verify that a style exists before passing it to `GetMap`, so I'm not sure why we would here.  IMO `parameters` should be passed through to the server without any judgement or interpretation, just do what the user tells us to do.  I think we should pass all of the `parameters` to `GetLegendGraphic` too, not just `style`.

rdengate, line 412: I didn't realize that was an option. It does sound like a much better idea. Thanks!

wwwroot/test/WMS/multiple_style_legend_url.xml
```diff
@@ -102,7 +102,7 @@
                         <Format>image/gif</Format>
                         <OnlineResource xmlns:xlink="http://www.w3.org/1999/xlink"
                                         xlink:type="simple"
-                                        xlink:href="http://www.example.com/legendUrl"/>
+                                        xlink:href="http://www.example.com/request=GetLegendGraphic?firstUrl"/>
```
kring: This isn't actually a valid `GetLegendGraphic` URL.  It should be:
http://www.example.com?request=GetLegendGraphic&firstUrl

kring: Our regex currently matches anything with the string `GetLegendGraphic`, but we might tighten it up in the future, in which case it'd be nice if this test still passed.

kring:
I made those small changes and merged it so we can include this in the NM release today.

PR #1620, opened 2016-06-10
NewUI: Use user-provided style rather than picking first

rdengate:
Styles can be passed as a parameter to style the WMS layer, but it is ignored for GetLegendGraphic, which currently uses the URL from the first style if a list is provided. Now it uses the specified style if provided, but otherwise chooses the first style as it previously did.

GEOGLAM requires this change in both master and newui.


kring:
Do we need this PR?  We can pick this up just by merging master into newui once the other PR is merged to master, right?

rdengate:
Oh right, I didn't know that happened. Thanks!

PR #1626, opened 2016-06-14
Cache metadata and single data files for just one day

### LINKED ISSUE #1617: Layer metadata requests are cached too long
The layer metadata requests are no longer cached for just 1 day, but are now being cached at the default of 14 days which is much too long.

This is broken in both ArcGIS and WMS layers, so may be broken across all?

kring:
Fixes #1617


meh9:
I have tested this for WMS and ArcGIS which both now work as expected. However, I can't figure out what I need to do to test the other catalogue types.

kring:
Just load a dataset of each type and verify that the cache headers are correct?  Or just press that big green merge button, because it was the same change in every catalog item type. ;)

PR #1630, opened 2016-06-14
Support JSON5 (http://json5.org/) use in init files and config files,…

stevage:
… so comments

can be used and object keys don't need to be quoted.

### DIFF
--- a/lib/Core/loadJson5.js
+++ b/lib/Core/loadJson5.js
@@ -1,0 +1,31 @@
+"use strict";
+/* global require  */
+
+var json5 = require('json5');
+var loadText = require('terriajs-cesium/Source/Core/loadText');
+var defined = require('terriajs-cesium/Source/Core/defined');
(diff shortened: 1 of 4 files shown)


kring:
@stevage I'll review/merge this one after the NM release today.

lib/Core/loadJson5.js
```diff
@@ -0,0 +1,31 @@
+"use strict";
+/* global require  */
+
+var json5 = require('json5');
+var loadText = require('terriajs-cesium/Source/Core/loadText');
+var defined = require('terriajs-cesium/Source/Core/defined');
+var clone = require('terriajs-cesium/Source/Core/clone');
+
+var defaultHeaders = {
+    Accept : 'application/json,*/*;q=0.01'
```
stevage: there's no official json5 mimetype, but we should probably pre-empt `application/json5`

PR #1636, opened 2016-06-15
Adding isNcWMS to WebMapServiceCatalogItem

jiwalker-usgs:
Using the boxfill/rainbow style as the way to detect, from what I can tell this is always included for layers in ncWMS.  I was hoping for something like a keyword as in the geoserver case, but this was not there.  I'm open to suggestions on other possibilities (included a getCaps document in wwwroot/test).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 ### 4.0.0
 
 * `CswCatalogGroup` will now include Web Processing Services from the catalog if configured with `includeWps` set to true.
+* `WebMapServiceCatalogItem` will now detect ncWMS servers and set isNcWMS to true.
 
 ### 3.3.0
 
--- a/lib/Models/WebMapServiceCatalogItem.js
+++ b/lib/Models/WebMapServiceCatalogItem.js
@@ -122,6 +122,12 @@ var WebMapServiceCatalogItem = function(terria) {
     this.isEsri = undefined;
 
     /**
+     * Gets or sets whether this WMS has been identified as being provided by ncWMS.
+     * @type {Boolean}
+     */
+    this.isNcWMS = undefined;
+
+    /**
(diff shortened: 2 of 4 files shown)


rsignell-usgs:
The Travis CI Build issues appear to have nothing to do with this PR, right?
https://travis-ci.org/TerriaJS/terriajs/builds/137918372#L1656-L1670

kring:
Thanks for the PR @jiwalker-usgs!  It's disappointing that ncWMS doesn't offer a better way to detect it (or at least detect the presence of its custom capabilities!) but this seems like a reasonable heuristic.  I'll review more thoroughly and merge tomorrow morning.

> The Travis CI Build issues appear to have nothing to do with this PR, right?

The Travis build is failing because the part that runs the tests on Sauce Labs doesn't work for pull requests originating from outside the main TerriaJS repo,  This is because of the way Travis encrypts secret bits of information (the Sauce Labs API key in this case).

rsignell-usgs:
@kring , so this is the travis issue?
https://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests

kring:
Yep, that's it.  We should follow the advice on that page and skip the test run for external PRs.  Or, better yet, to get some level of testing, we could run most of our tests on Electron (roughly, Chrome running inside Node) right on the Travis VM for external PRs, rather than use Sauce Labs.

rsignell-usgs:
@kring, that would be great, as having all external PRs failing is a less-than-awesome experience.  :grimacing: 

If you wanted to do this as a task under our contracted work, that would be fine by me.

kring:
Looks great, thanks @jiwalker-usgs!  Can you please add yourself to this file in a separate PR:
https://github.com/TerriaJS/terriajs/blob/master/CONTRIBUTORS.md

jiwalker-usgs:
Will do!

PR #1637, opened 2016-06-16
Made auto-generated legend titles wrap on non-MS browsers.

### LINKED ISSUE #1580: Autogenerated legends should wrap long titles
Some of the autogenerated legends have long titles, we should wrap them: http://nationalmap.gov.au/#share=htCdkl

screen shot 2016-05-31 at 22 48 59

AlexGilleran:
Fixes #1580.

This is a bit awkward because SVG doesn't flow at all. What this does is move the bars down enough to accomodate two lines of text (for all legends, but it looks ok IMHO) then inserts a `<p>` tag inside a `<foreignObject>` in the SVG to handle the auto wrapping. If the browser doesn't support `<foreignObject>` (IE/Edge except IE11) it falls back to what it did before - a one-line `<text>` element that doesn't wrap at all.

Also made auto generated legends a bit nicer in MS browsers in general - no more second white background inherited from non-autogenerated legends.

Also got rid of a second `overflow: auto` inside the sidebar - not sure why that's there? Was causing a horizontal scroll bar for me.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,10 @@
 Change Log
 ==========
 
-### 3.3.1
+### 3.4.0
 
 * Fixed a bug that caused the `corsProxyBaseUrl` specified in `config.json` to be ignored.
+* Long auto-generated legend titles now word wrap in most web browsers.
 
 ### 3.3.0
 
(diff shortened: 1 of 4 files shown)


kring:
I get this in IE11 too:
image

Anyway, this is a big improvement in Chrome and Firefox, and not really any worse in IE/Edge, so I'm merging.  It'd be nice to fix this better in the new UI though.

PR #1638, opened 2016-06-17
Run tests using Electron when Sauce Labs is not available

kring:
Sauce Labs can't be used for testing of pull requests originating from outside the main TerriaJS repo, because the API key is not available there due to the (totally sensible) Travis security model.  This PR detects that situation and runs the tests right on the Travis VM using Electron instead.

See discussion in #1636.


kring:
@stevage mind reviewing and merging this one?

stevage:
There are now three redundant lines in the gulp file for `test-sauceLabs`: https://github.com/TerriaJS/terriajs/blob/7c024c4bc03afe6e27c4a6454e7ca472b25a3bab/gulpfile.js#L144

Other than that, ok. (So you can self-merge.)

kring:
Redundant as in, unnecessary?  The `test-saucelabs` task is still useful for explicitly running on SauceLabs from a local machine (as I just did today to track down the kraken).

PR #1640, opened 2016-06-17
Added WMS <MetadataURL> to layer description

### LINKED ISSUE #1545: Add support for WMS GetCapabilities <MetadataURL/> elements
We have had some requests to expose `<MetadataURL/>` elements from the WMS `GetCapabilities` data. In most cases these link to sources of standards compliant metadata pages for the datasets in question. We should probably show these links on the layer Info page as sources of more information about the dataset.

Example `GetCapabilities` with `<MetadataURL/>` elements:

http://services.ga.gov.au/gis/services/Australian_Coastal_Sediment_Compartments/MapServer/WMSServer?request=GetCapabilities&service=WMS

AlexGilleran:
Fixes #1545.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 ### 3.3.1
 
 * Fixed a bug that caused the `corsProxyBaseUrl` specified in `config.json` to be ignored.
+* Added support for the WMS Style/MetadataURL tag in layer description.
 
 ### 3.3.0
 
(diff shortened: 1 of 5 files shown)


kring:
I've changed the title to just "Metadata Links" instead of "Data Metadata Links", cause the latter seems redundant to me.  Any objections?

kring:
Oh I see the discussion about this in the bug now.  hmmm

AlexGilleran:
I certainly don't object. Come to think of it "data metadata" does sound a bit tautological

PR #1641, opened 2016-06-17
Fixes #1581 (no feature info data in CSV format if it contained nulls).

### LINKED ISSUE #1581: Features with null attributes can't generate CSV downloads
If you have a layer (say, CED 2016) with null attributes (population in Lingiari, WA) then [this line](https://github.com/terriajs/terriajs/blob/d1f6118fe7eecb80051080be35b9cdf213edae03/lib/ViewModels/FeatureInfoPanelSectionViewModel.js#L272-L272) reports the null value as an object and CSV data generation fails. Whereas it should be possible to download as CSV in that case (null is just two consecutive commas...).

### LINKED ISSUE #1586: A way of selecting a layer identifier as input to WPS
A WPS process can require a layer as input. We need a way to select a layer and pass it to the WPS function.

Things to think about is if:
- the layer should be from the enabled layers only
- from all layers
- only ones valid for the WPS function
- some intersection of the above

stevage:


stevage:
Err, that should be fixes #1581.

PR #1642, opened 2016-06-17
Filter out duplicate mousemove events

### LINKED ISSUE #1523: Terrain height is continually polled on high DPI displays
On my Windows laptop (which is not actually high DPI) with "Change the size of text, apps, and other items" set to 175% (so it acts like it is), the location bar at the bottom continually flips between showing an elevation +/- some number and a precise elevation.  Presumably it thinks the mouse is moving around?

kring:
Fixes #1523


kring:
@chloeleichen mind reviewing and merging this one?

PR #1645, opened 2016-06-17
Newui file dnd

chloeleichen:
Refactor drag and drop:

This branch: 
1. Moves the drag-and-drop overlay out of explorer window so it uses browser coordinates 
2. Fixes drag-and-drop and drop overlay stuck issue: https://github.com/TerriaJS/terriajs/issues/1435
3. Drag-and-drop file now works on extra small screen
4. The drag-and-drop event is bind to div#ui rather than the entire document


lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
```diff
@@ -50,7 +51,7 @@ const StandardUserInterface = React.createClass({
             that.acceptDragDropFile();
         };
 
-        document.addEventListener('dragover', this.dragOverListener, false);
+        document.getElementById('ui').addEventListener('dragover', this.dragOverListener, false);
```
AlexGilleran: This is configurable by the implementation of TerriaJS, could we change it to using a `ref` of the outermost element?

chloeleichen:
@AlexGilleran using `ref` rather than `id` now. Would you please have a look again?

PR #1647, opened 2016-06-20
NewUI: Added ability to format timeslider tics and timeline label

rdengate:
Previously in master, we could control the label format on the timeslider (tics, and label on left). This is the same feature ported to newui. Note that because of changes in how timeslider works, the implementation is different from master.

Any maps that used this feature previously will need to be updated, because the way of specifying format in config.json has changed. 
Old:

```
"dateFormat":
 {
    "currentTime": "mmmm yyyy",
    "timelineTic": "yyyy"
 }
```

New, using javascript toLocaleDateString/toLocaleTimeString options for both currentTime and timelineTic:

```
"dateFormat":
{
    "currentTime": {"month": "short", "year": "numeric"},
     "timelineTic": {"year": "numeric"}
}
```


kring:
@AlexGilleran you're probably the right person to review this one.

AlexGilleran:
So why the switch away from `dateFormat` ? Unfortunately `toLocaleTimeString` doesn't support options in a lot of browsers https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString. If you dig through all the cruft in the travis logs you can see your test fails in Safari :(.

rdengate:
I did it because all the timeline code is using the DateFormat functions formatDate, formatTime, and formatDateTime. It didn't make sense for me to use a separate date formatter just for this feature. However, since it isn't working on Safari, I guess we have no choice.

Issue #1514, opened 2016-05-10
Implement SDMX-JSON catalog item


RacingTadpole:
SDMX is the standard used by the OECD and many statistical agencies, including the Australian Bureau of Statistics (ABS). The ABS is implementing SDMX-JSON.

Useful links include:
- [OECD API Documentation](https://data.oecd.org/api/sdmx-json-documentation/)
- [OECD query builder](http://stats.oecd.org/sdmx-json)
- [The Technical Standards Working Group's guide to SDMX-JSON, with examples](https://github.com/sdmx-twg/sdmx-json/blob/master/data-message/docs)
- [Official SDMX website](https://sdmx.org/)
- [ECB SDMX RESTful web service](https://sdw-wsrest.ecb.europa.eu/)
- [Introduction to SDMX-JSON (slides)](http://airosa.github.io/sdmx-prototype-json/sdmxjson-intro-slides/)

RacingTadpole:
@sosna, @airosa: Hi, @tobybellwood suggested I get in contact with you. I'm implementing an SDMX-JSON reader in TerriaJS, which underlies nationalmap.gov.au, so that we can display statistical agencies' data - in particular Australian Bureau of Statistics data - on a map.

We currently use an unsupported API for the ABS data; you can see an example using this API at http://nationalmap.gov.au/#share=JBZ9pC.

The SDMX-JSON-reading code so far is at: https://github.com/TerriaJS/terriajs/blob/1514_sdmx/lib/Models/SdmxJsonCatalogItem.js

The idea is the reader can either handle a specific endpoint like `http://stats.oecd.org/sdmx-json/data/QNA/.B1_GE.VOBARSA.A/all`, or an open-ended one like `http://stats.oecd.org/sdmx-json/data/QNA`. In the latter case it uses `http://stats.oecd.org/sdmx-json/dataflow/QNA` to identify the structure of the data first.  It then identifies time and region concepts and treats those specially.

The main outstanding features I'd still like to implement are passing through `attributes` like units, handling potentially hierarchical dimensions, and knowing when the data can be meaningfully divided by regional population to show a %.

Do you have any advice or suggestions?

sosna:
This project looks very interesting :)

[...]

PR #1671, opened 2016-06-23
Add SDMX-JSON CatalogItem

RacingTadpole:
This adds a new catalog item which can read the SDMX-JSON format used by the OECD and (soon) the ABS, per #1514.  Try out the demos at http://localhost:3003/#build/terriajs/test/init/sdmx.json .

This should handle the following cases -
- a TIME_PERIOD dimension
- multiple frequencies
- multiple region types (though this is waiting for an example to test)
- both specific datafile and generic API endpoints
- ability to set the initially selected values in json
- a REGION dimension with no REGIONTYPE dimension

Some limitations:
- It can’t handle hierarchical dimensions (eg. like the ABS-ITT currently has for Age->0-4 years->0,1,2,3,4)
- It does not display units or other attributes, which would be great. In particular it shouldn't try to add values with different units.
- It does display non-spatial data on a chart, but this needs more thought, since it only displays the total, and the way of choosing which concepts to display doesn't make intuitive sense here. 
- It doesn't offer a "show as % of regional population" option.

I'm submitting this PR now, even without any live catalog items using it, because there are a few small changes to other files that we may as well bed down now.

### DIFF
--- a/lib/Models/SdmxJsonCatalogItem.js
+++ b/lib/Models/SdmxJsonCatalogItem.js
@@ -1,0 +1,896 @@
+'use strict';
+
+/*global require*/
+var URI = require('urijs');
+
+var clone = require('terriajs-cesium/Source/Core/clone');
+var defaultValue = require('terriajs-cesium/Source/Core/defaultValue');
+var defined = require('terriajs-cesium/Source/Core/defined');
+var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
+// var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
+var freezeObject = require('terriajs-cesium/Source/Core/freezeObject');
+// var deprecationWarning = require('terriajs-cesium/Source/Core/deprecationWarning');
+var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
+var loadJson = require('terriajs-cesium/Source/Core/loadJson');
+var when = require('terriajs-cesium/Source/ThirdParty/when');
+
+var arrayProduct = require('../Core/arrayProduct');
+// var arraysAreEqual = require('../Core/arraysAreEqual');
+var CsvCatalogItem = require('./CsvCatalogItem');
+var DisplayVariablesConcept = require('../Map/DisplayVariablesConcept');
+var inherit = require('../Core/inherit');
+var TerriaError = require('../Core/TerriaError');
+var overrideProperty = require('../Core/overrideProperty');
+var proxyCatalogItemUrl = require('./proxyCatalogItemUrl');
+var RegionMapping = require('../Models/RegionMapping');
+var TableColumn = require('../Map/TableColumn');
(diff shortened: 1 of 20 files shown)


lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -0,0 +1,931 @@
...
+var overrideProperty = require('../Core/overrideProperty');
+var proxyCatalogItemUrl = require('./proxyCatalogItemUrl');
+var RegionMapping = require('../Models/RegionMapping');
+var TableColumn = require('../Map/TableColumn');
+var TableStructure = require('../Map/TableStructure');
+var VariableConcept = require('../Map/VariableConcept');
+var VarType = require('../Map/VarType');
+
+/*
+    The SDMX-JSON format.
```
kring: Should some of this text go in the jsdoc so that it is included in our generated reference documentation?  Some of it is implementation notes and need not go in the jsdoc, but parts of it seem useful to folks trying to use this class.

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -0,0 +1,931 @@
...
+    }
+});
+
+/**
+ * Gets or sets the default set of properties that are serialized when serializing a {@link CatalogItem}-derived for a
+ * share link.
+ * @type {String[]}
+ */
+SdmxJsonCatalogItem.defaultPropertiesForSharing = clone(CsvCatalogItem.defaultPropertiesForSharing); // TODO: do we need to add dataUrlComponent etc to this list?
+SdmxJsonCatalogItem.defaultPropertiesForSharing.push('regionDimensionId');
```
kring: The rule of thumb for `defaultPropertiesForSharing` is this: if it can be changed via the UI, it goes in `defaultPropertiesForSharing`, otherwise it doesn't.  So most of these things don't need to be included when sharing, right?

RacingTadpole:
Thanks, I've made those changes. I've also fixed sharing so the selected concepts are shared properly.

test/Models/AbsIttCatalogItemSpec.js
```diff
@@ -6,61 +6,6 @@ var Rectangle = require('terriajs-cesium/Source/Core/Rectangle');
 var Terria = require('../../lib/Models/Terria');
 var AbsIttCatalogItem = require('../../lib/Models/AbsIttCatalogItem');
 
-describe('AbsIttCatalogItem._generateAllCombinations', function() {
```
kring, line 9: Looks like you removed the tests, but forgot to remove the code that was tested here.  I'm assuming you meant to change it to use your new `arrayProduct`.

kring:
This looks good as a first cut!  As you mentioned, it's pretty weird how all of the selected values get added up and used to color the map, but it's not at all obvious what would be better.

I'm merging this, please update CHANGES.md and remove the old `AbsIttCatalogItem._generateAllCombinations` in newui.

PR #1677, opened 2016-06-23
Word wrap in legends, add support for gradients with blank values

### LINKED ISSUE #1660: Newui: Bad legends with long names
image

kring:
Fixes #1660 

Added support for long titles in legends by manually measuring text length and wrapping.
image

Also fixed gradient legends, which were broken in the new UI due to a CSS problem:
image

And added support for "blank" values with gradient legends (this wasn't supported before even in the old UI):
image


lib/Map/Legend.js
```diff
@@ -115,36 +115,24 @@ var Legend = function(props) {
      */
     this.variableNameTop = defaultValue(props.variableNameTop, 17);
 
+    /**
+     * Gets or sets the CSS class that will be applied to the legend when it is displayed.
+     * This is used to ensure that the correct font is used when measuring text for word wrapping.
+     * @type {String}
+     */
+    this.cssClass = defaultValue(props.cssClass, 'legend__legend');
```
kring: This is the dodgiest bit of the pull request.  In order to do manual word wrapping, we need to know what font will be used.  The fundamental problem is that all the code that creates legends has no actual idea of what CSS will eventually be used to render them.  So putting `legend__legend` here is a pretty serious violation of the principle that the `Map` layer shouldn't know about the `Views` layer.  Ideas on how to fix this are welcome, otherwise I guess we'll just live with it.

chloeleichen: I think i totally missed the discussion on slack, but why not use `foreignobject` to treat those text as html? 

Also this one does wrapping as well: http://d3plus.org/examples/utilities/a39f0c3fc52804ee859a/

kring: Alex used `foreignObject` initially, but it doesn't work in Microsoft browsers, even new ones.  Also, it's hard to make the rest of the content move down to make room, with the result that you see in the bug that this one fixes.

Arguably the right solution is to let legends generate HTML instead of SVG.  I think that might  need to wait until after 15 July though.

kring:
@stevage can you merge this one?

stevage:
How do I replicate the issue? Where's the "Language and proficiency in English" dataset?

kring:
It's easy to create really long titles with ABS datasets.  If it's not obvious I'll give you an exact dataset name when I'm next at a computer.

stevage:
Yeah, surprisingly it's not that obvious (to me). I get legends like this:

screenshot 2016-06-29 17 41 41

kring:
`B12 Proficiency in Spoken English/Language of Parents by Age of Dependent Children` (under ABS Data Service) is great for long names.  Turn off `Display as a percentage of region population` and then start checking other boxes, but only one per group (otherwise the title changes to just `Total selected`).

So the problem in your screenshot maybe has something to do with high DPI displays, I'll check it out.

kring:
Nope, it was just broken in general because the name of the CSS class used for legends changed (it gained the `tjs-` prefix).  It works fine now on an (emulated) high DPI display, so hopefully it will on a real one, too.

stevage:
Side note: we probably shouldn't generate such long titles in ABS layers, eh, @RacingTadpole ?

stevage:
screenshot 2016-06-30 11 04 05

kring:
Yeah it only wraps the legend title, not the name of each box.  Wrapping the names in the boxes, too, would be a good idea, but there's no attempt at implementing that here.  Separate issue?

PR #1678, opened 2016-06-24
Show the map but not the UI when hideUi=true

kring:
@meh9


lib/ReactViews/StandardUserInterface/map-column.scss
```diff
@@ -18,7 +18,6 @@
 
 .map__cell {
   display: table-cell;
-  position: relative;
```
kring, line 21: This was the key line that made the map reappear.  Hopefully removing it doesn't break IE or something crazy.  It seemed ok in my testing.  @AlexGilleran

RacingTadpole:
Hi - happy to review this one but it looks like there's some background to it that I've missed - so I'm going to assign @AlexGilleran - hope that's ok!

AlexGilleran:
Unfortunately this makes the timeline/graph pop up over the top of the map :(

image

image

kring:
:scream: thought I tested that.  I'll look again.

AlexGilleran:
It's a subtle one because somehow it moves the disclaimer but not the legend :/

kring:
Ok this is ready now.

PR #1679, opened 2016-06-25
Fix several Add Data problems

### LINKED ISSUE #1665: problem adding this WMS endpoint to newUI
I tried to add the WMS service endpoint
http://mesonet.agron.iastate.edu/cgi-bin/wms/nexrad/n0r-t.cgi
to the newUI instance at
http://nationalmap.research.nicta.com.au/newui/
by using the "Add data=>My Data=>Add Web Data", selecting "WMS", pasting the endpoint URL and selecting the 2nd layer for display (there are 2 layers).   

When I do this, I seem to get some view of the galaxy (with no Earth in sight) that I can't get out of:

http://tinyurl.com/juj2o9s

Any idea what is going on here?

It works okay with the old UI:

kring:
Fixes #1665 
(actually the "view the galaxy" problem seems to no longer exist, but this fixes new problems when trying to do the same thing).
- Fixed an exception when adding new user data due to an attempted call of a function that doesn't exist.  I think the intention was to switch to the list of user datasets, so that's what I've done.
- Fixed an exception when attempting to add user-added data to the map, because it was incorrectly assuming that the added item would be the view state's `previewedItem`, but user-data has a separate preview.

### DIFF
--- a/lib/ReactViews/DataCatalog/DataCatalogItem.jsx
+++ b/lib/ReactViews/DataCatalog/DataCatalogItem.jsx
@@ -39,7 +39,7 @@ const DataCatalogItem = React.createClass({
         this.props.viewState.viewCatalogItem(this.props.item);
         // mobile switch to nowvewing
         this.props.viewState.switchMobileView(this.props.viewState.mobileViewOptions.preview);
-        if (this.props.viewState.previewedItem.isEnabled === true &&
+        if (this.props.item.isEnabled === true &&
             this.props.viewState.closeModalAfterAdd === true &&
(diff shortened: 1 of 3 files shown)


RacingTadpole:
This looks good. Two questions (which are related to this PR, but not due to it, since I see them in `newui` as well):
- I see an empty "User-added data" group is at the bottom of the Data Catalog, even at the start - is that intentional?
- When I try to add the URL `http://data.gov.au`, my localhost:3003 refreshes completely (I've tested in both Chrome and Firefox). Are you seeing that too?

PR #1680, opened 2016-06-25
Fix up Leaflet picking.

### LINKED ISSUE #1627: Cannot view GeoJSON Properties Table When Leaflet (2D) is Present and "marker-symbol" is used.
When the map is set to 2D mode (Leaflet) clicking on anything that's defined with a marker-symbol (or image) does not show the properties passed in with GeoJSON. 

If you add "Test overriding Styled and unstyled features" GeoJSON dataset to the map and toggle between 3D and 2D you'll see that on 3D the table with the properties is displayed. On 2D that table is not. 

I attempted to debug and it looks like the click event on the leaflet marker is gobbling up the event that attempts to find the features at that spot.

kring:
For unknown reasons, Leaflet billboards (unlike other feature types) swallow click events, so the map click event is never raised when the user clicks on a billboard.  I've reorganized (and simplified, I think) the Leaflet picking to handle this possibility.  The updated comment in the code describes how it works.

Sharing the feature info panel doesn't work, but it doesn't work in the new UI branch before this change either.  I don't _think_ I've made it any worse.

Fixes #1627

### DIFF
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -722,8 +722,8 @@ LeafletVisualizer.prototype.visualizersCallback = function(leafletScene, dataSou
     return [new LeafletGeomVisualizer(leafletScene, entities)];
 };
 
-function featureClicked(visualizer, entity) {
-    visualizer._leafletScene.featureClicked.raiseEvent(entity);
+function featureClicked(visualizer, entity, e) {
+    visualizer._leafletScene.featureClicked.raiseEvent(entity, e);
 }
 
 module.exports = LeafletVisualizer;
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -722,8 +722,8 @@ LeafletVisualizer.prototype.visualizersCallback = function(leafletScene, dataSou
     return [new LeafletGeomVisualizer(leafletScene, entities)];
 };
 
-function featureClicked(visualizer, entity, e) {
-    visualizer._leafletScene.featureClicked.raiseEvent(entity, e);
+function featureClicked(visualizer, entity, event) {
+    visualizer._leafletScene.featureClicked.raiseEvent(entity, event);
 }
 
 module.exports = LeafletVisualizer;
--- a/lib/Models/Leaflet.js
+++ b/lib/Models/Leaflet.js
@@ -341,6 +341,13 @@ Leaflet.prototype.lowerToBottom = function(item) {
     item._imageryLayer.setZIndex(0);
 };
 
+/**
+ * Picks features based off a latitude, longitude and (optionally) height.
+ * @param {Object} latlng The position on the earth to pick.
+ * @param {Object} imageryLayerCoords A map of imagery provider urls to the coords used to get features for those imagery
+ *     providers - i.e. x, y, level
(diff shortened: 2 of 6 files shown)


lib/Models/Leaflet.js
```diff
@@ -342,13 +342,7 @@ Leaflet.prototype.lowerToBottom = function(item) {
 };
 
 Leaflet.prototype.pickFromLocation = function(latlng, imageryLayerCoords, existingFeatures) {
```
RacingTadpole, line 351: Can you add jsdocs to this function?

lib/Map/LeafletVisualizer.js
```diff
@@ -722,8 +722,8 @@ LeafletVisualizer.prototype.visualizersCallback = function(leafletScene, dataSou
     return [new LeafletGeomVisualizer(leafletScene, entities)];
 };
 
-function featureClicked(visualizer, entity) {
-    visualizer._leafletScene.featureClicked.raiseEvent(entity);
+function featureClicked(visualizer, entity, e) {
```
RacingTadpole: Can you be more specific than `e`? What sort of object is this? (Here and in `featurePicked`.)

lib/Models/Leaflet.js
```diff
@@ -415,42 +409,51 @@ function createLeafletCredit(attribution) {
...
+* - The map click is basically ignored, unless it is the first click event we see.
+* - Once this runLater executes, we can assume that all the click handlers for that one click will have fired, so we
+*       set _pickedFeatures back to undefined (this.terria.pickedFeatures is still set)
+* - When another click happens, it re-initialized _pickedFeatures and the process starts again.
+*
+* For most types of features, the map click will be the first click event, which means the raster pick will happen
+* exactly where the user clicked.  Leaflet billboards, however, seem to swallow click events, so if the user clicks
+* on a billboard _only_ the click for that feature will be raised.  That means the raster pick will happen at the
+* location of the billboard, rather than at the actual location that the user clicked.
+*/
```
RacingTadpole, line 435: I know this largely pre-dates your PR, but this whole comment block is really confusing (to me at least). "Basically" ignored? What if the map click _is_ the first we see? Do we really rely on multiple clicks for vector features, when we just use one of them? (Maybe "rely on" should be "receive"?) If "multiple clicks" is always two, why not just say two? Etc.

I propose this replacement:

```
/*
* There are two "listeners" for clicks which are set up in our constructor.
* - One fires for any click: `map.on('click', ...`.  It calls `pickFeatures`.
* - One fires only for vector features: `this.scene.featureClicked.addEventListener`. 
*    It calls `featurePicked`, which calls `pickFeatures` and then adds the feature it found, if any.
* These events can fire in either order.
* Billboards do not fire the first event.
*
* Note that `pickFeatures` does nothing if `leaflet._pickedFeatures` is already set. 
* Otherwise, it sets it, runs `runLater` to clear it, and starts the asynchronous raster feature picking.
*
* So:
* If only the first event is received, it triggers the raster-feature picking as desired.
* If both are received in the order above, the second adds the vector features to the list of raster features as desired.
* If both are received in the reverse order, the vector-feature click kicks off the same behavior as the other click would have;
* and when the next click is received, it is ignored - again, as desired.
*/
```

RacingTadpole:
Just those few minor doco-related things - thanks!

PR #1681, opened 2016-06-27
PolylineGlowMaterialProperty in Leaflet

rdengate:
A PolylineGlowMaterialProperty with colour (0, 0, 0, 0.1) produces a white with subtle black falloff outline in Cesium map. However, Leaflet doesn't know much about PolylineGlowMaterialProperty and produces a line that is flat shaded (0, 0, 0, 0.1), so barely visible.

As far as I could tell, Leaflet doesn't have anything similar to PolylineGlowMaterialProperty. We could try to fake it by replacing leaflet line with polyon that has white fill and specified colour stroke, or by drawing specified colour line and then white line on top, but both of those options are very hacky though and would require large changes in Leaflet. 

This PR changes line to use default line properties for Leaflet if PolylineGlowMaterialProperty is in use.

### DIFF
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -15,6 +15,7 @@ var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
 var Ellipsoid = require('terriajs-cesium/Source/Core/Ellipsoid');
 var Property = require('terriajs-cesium/Source/DataSources/Property');
 var writeTextToCanvas = require('terriajs-cesium/Source/Core/writeTextToCanvas');
+var PolylineGlowMaterialProperty = require('terriajs-cesium/Source/DataSources/PolylineGlowMaterialProperty');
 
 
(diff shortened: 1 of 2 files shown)


kring:
Drawing two lines on top of each other, one wider, makes a lot of sense to me.  And it sounds like that should be doable in the LeafletVisualizer without changing Leaflet itself.  Am I mistaken?

rdengate:
Ah sorry! Yes, I'm saying "Leaflet" when actually I mean "LeafletVisualizer". It's absolutely possible in LeafletVisualizer, but probably not pretty.

RacingTadpole:
Hi! So I can understand what this does, can you add an example (eg. to `test-styles.geojson`) that shows difference between the current and the proposed behaviours?

rdengate:
@RacingTadpole I've added wwwroot/test/CZML/glowing_polyline.czml, which has a glowing line through America.

RacingTadpole:
Thanks! Looks good to me!

PR #1693, opened 2016-06-27
Add optional context for CatalogFunctions

### LINKED ISSUE #1557: Add capability to enable another CatalogItem when using an item specified by WebProcessingServiceCatalogGroup
WPS functions can sometimes only be used in particular geographical areas. In order to show the user where they can click for the WPS we should show another layer indicating where it is "safe to click".

Add support to `WebProcessingServiceCatalogGroup` so that another `CatalogItem` can be specified for this that will be turned on when a item in the group is clicked. This layer should then then be turned off at some point, possibly when any other thing in the catalog is clicked on or something similar.

kring:
Fixes #1557 

Also fixes some styles on the analytics pages.

Currently this is rigged up to show a `CatalogFunction`'s `contextItem` layer when that function is selected in the catalog UI, even if the catalog UI is not visible.  So if we click Add Data and click a Wave Atlas WPS function, the context layer appears on the map and in the workbench.  Then we can close the catalog panel and the context will still be visible.  If we open the catalog panel and select a different dataset, the context layer will disappear.

For the most part this seems pretty ok, but it can get a little weird in corner cases.  For example:
1. Click on a WPS function so the context appears.
2. Close the catalog panel.
3. Click the `About this dataset` button on the context layer in the Workbench.

The catalog will appear and show the details of the context layer (good!) but the context layer will simultaneously disappear from the workbench because the original WPS item is no longer selected (weird!).

### DIFF
--- a/lib/Models/CatalogFunction.js
+++ b/lib/Models/CatalogFunction.js
@@ -4,9 +4,12 @@
 var arraysAreEqual = require('../Core/arraysAreEqual');
 var CatalogItem = require('./CatalogItem');
 var CatalogMember = require('./CatalogMember');
+var clone = require('terriajs-cesium/Source/Core/clone');
+var createCatalogMemberFromType = require('./createCatalogMemberFromType');
 var defined = require('terriajs-cesium/Source/Core/defined');
 var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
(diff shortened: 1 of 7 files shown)


meh9:
I understand the downsides. I think they are acceptable as this is not going to be a feature that is used very often. If we start using it a lot I think we can put more time into polishing it.

PR #1696, opened 2016-06-28
Fix chart download button; reorganise Chart directories

RacingTadpole:
4 fixes:
- The Chart download button went blank somewhere along the way, so this fixes that (try Sheep and Star stations on http://localhost:3003/#build/terriajs/test/init/charts.json).
- Improves directory structure - eg. we previously had "Collapsible.jsx" in the "Chart" directory.
- Improves naming of "ChartExpandButton", which actually shows both Expand and Download buttons.
- Improved some comments.


kring:
Travis is reporting a build error for this PR.

RacingTadpole:
Sorted.

PR #1697, opened 2016-06-28
Implement new ShareDataService to replace Google URL Shortener.

stevage:
This shouldn't be merged until we work out what to do with `/usable` but I'd appreciate a review.


kring:
Looks ok to me.  I like the idea of having the server config report whether it supports URL shortening.  We should use a similar mechanism to determine if the feedback feature is available.  Currently it's enabled on the client by the presence of a feedback URL in the client config file, but that's really ugly.

stevage:
Ok, I added a wrapper around `/serverconfig` which gets initialised before anything else, sort of where `/proxyabledomains` used to get called.

kring:
There are test failures in this branch.

kring:
Also, do you have any thoughts on terriajs's backward compatibility with older versions of terriajs-server?  After this PR, terriajs will crash on startup if using a version of terriajs-server without the new `/serverconfig` service.  We should try to be more accommodating than that, if it's not a huge hassle.

stevage:
Yeah I should be able to fix that. In general, I don't think there is very much need for a new TerriaJS to support an old TerriaJS-server, although I guess TerriaMapStatic could make an exception or two.

kring:
Yeah I'd say that TerriaJS should be mostly usable without any of our fancy services in the server.  A simple static web server should be sufficient for many use cases.

If that works, it's not much different from using an old terriajs-server that is only missing some of the services.

stevage:
Ok, I think I achieved the above.

stevage:
Kind of hard to know what to do about the second test failure, which is caused because I removed support for this in config.json:

```
    "parameters": {
        "proxyableDomainsUrl": "test/init/proxyabledomains.json"
    }
```

That is, before it was possible to load the list of proxyable domains from some alternative file, whose URL is given here. Is it necessary that we continue to support this? Or is it configurability for the sake of configurability?

If we support it, that means you can define proxyable domains:
- in terriajs-server's server config (`/serverconfig -> allowProxyFor` or `proxyAllDomains`)
- in config.json's `proxyDomains`
- by creating a separate file, and pointing at it in `config.json` (as above)

Seems like an awful lot of flexibility for a relatively minor feature.

lib/Models/Terria.js
```diff
@@ -418,7 +459,7 @@ Terria.prototype.start = function(options) {
  * @param {String} newUrl The new URL of the application.
  * @return {Promise} A promise that resolves when any new init sources specified in the URL have been loaded.
  */
-Terria.prototype.updateApplicationUrl = function(newUrl) {
+Terria.prototype.updateFromApplicationUrl = function(newUrl) {
```
kring: This is a breaking change.  Add a new function with the old name and call this one after calling `deprecationWarning`.  Also, look for all the places we call it.  There's at least one.

stevage: oh, I didn't think of that. It's not important, I'll just unchange it.

PR #1698, opened 2016-06-28
Map Marker for Search Results

### LINKED ISSUE #296: Search for address or placename should indicate search result
User request (from a local council): Currently when you search for a placename, it just zooms in to the location but it isn't totally obvious where the successful search hit is.  It would be good to highlight this. Assuming we don't have access to the actual text on the basemap, we could just put a marker showing where the search hit is.

AlexGilleran:
Fixes #296 (three digit issue #!!!!)

When the sidebar search is visible, it now adds an extra layer to terria that puts a map marker on the map when a search result is clicked. When the sidebar search is removed, it removes the map layer, so you only ever see map markers while the search is visible.

To figure out where to drop the pin, the search results have a `location` property now. This kind of breaks the interface where catalog item search and location search use the same results, but now that they're in separate views I expect we'll be dropping the commonality anyway? They have very very little in common.

This uses the google material icons pin - I expect we'll replace this once Chloe's finished with SVG sprites.

This also has an additional fix for billboards in Leaflet - it wraps the click handler in `LeafletVisualizer` for non-circle markers and retriggers the click event on the map directly, as these block propagation in our version of Leaflet.


kring:
Not just 3 digits, also the oldest open issue!

lib/Map/LeafletVisualizer.js
```diff
@@ -731,6 +731,26 @@ LeafletVisualizer.prototype.visualizersCallback = function(leafletScene, dataSou
...
+ * refiring the click event on the base map after the feature is clicked.
+ *
+ * This still won't fix multiple vector layers on top of each other :(.
+ *
+ * TODO: Can remove this when we upgrade leaflet, as later versions have a completely overhauled event bubbling system.
+ *
+ */
+function featureClickedReraiseClick(visualizer, entity, e) {
+    if (visualizer._leafletScene.map._fireMouseEvent) {
+        visualizer._leafletScene.map._fireMouseEvent(e.originalEvent);
```
kring, line 748: Does this mean the map click event will be raised twice when the user clicks on a `CircleMarker`?

kring, line 748: Oh, nevermind, I see... You're only using this handler for the feature types known to have the problem.  Nice.

PR #1701, opened 2016-06-28
Update all packages, except the ones we purposely downgraded.

kring:
In response to #1699.  Merging this should close that, too.


kring:
Ideally, we'd write tests that fail when the two packages kept downgraded, `leaflet` and `superagent`, are upgraded, so greenkeeper can do its thing.  In the Leaflet case, at least, that might be pretty tricky.

AlexGilleran:
Damn Greenkeeper, sign the CLA!

Yeah we might run into more issues with that given that our coverage is pretty low :(

kring:
Yep, but it should at least help us avoid Travis krakens.  We still might find problems when e.g. actually running an app on IE9, but those are generally easier to debug.

kring:
The `superagent` bug was fixed, so that's now upgraded too.  Meaning, the only package from #1699 that this doesn't upgrade is Leaflet.  So this should be ready.

Issue #1686, opened 2016-06-27
Add specs for ReactViews (and check we haven't lost any key functionality)


RacingTadpole:
We need to make sure all the relevant ViewModel specs make it into `newui`.

RacingTadpole:
## AnimationViewModel

This is now `Timeline`.
- [ ] on init, showAnimation should be false
- when a layer is added to the time series stack
  - [ ] should set showAnimation to true
  - [ ] should set layerName to the name of the topmost layer
- when the last layer is removed from the time series stack
  - [ ] should set showAnimation to false
- timelineTic format
  - [ ] should be used if provided
  - [ ] should not be used if not provided
- currentTime format
  - [x] should be used if provided
  - [x] should not be used if not provided
- scrubbing
  - [ ] should set the current time to the time in the event
  - with gap between intervals
    - [ ] should set time to time within an interval as usual
    - should change a time to the previous interval's end
      - [ ] just after the start
      - [ ] just before the midpoint
    - should change a time to the next interval's start
      - [ ] just before the end
      - [ ] at the exact midpoint

## BrandBarViewModel
- [ ] to allow no elements
- [ ] to add elements from options

RacingTadpole:
## DisclaimerViewModel
- [ ] should allow an already-accepted disclaimer to go straight through without displaying
- for a non-confirmation disclaimer
  - [ ] should open a popup message
  - [ ] should immediately execute the success callback
  - [ ] should record the popup message as being shown in local storage
  - [ ] should pass the ui container straight through to the modal
- for a non-confirmation disclaimer
  - [ ] if no key is specified, opens a confirmation dialog with the correct success callback
  - [ ] should pass options from the catalogItem correctly
  - [ ] should pass the ui container straight through to the modal
  - if a key is specified
    - [ ] should not record the popup message as being accepted in local storage until after the success callback is executed
    - [ ] should only a modal once for a certain key
  - if multiple items with keys are specified
    - [ ] should open 2 modals for 2 different keys regardless of how many events for each are triggered
    - [ ] should trigger all callbacks for a key after that key's modal is dismissed

## ExplorerPanelViewModel
- on init
  - [ ] sets the initial tab to the value in terria.userProperties.activeTabId
  - [ ] selects the first tab if no value for terria.userProperties.activeTabId
  - [ ] selects the first tab if terria.userProperties.activeTabId doesn't match a tab id
  - [ ] avoids tab id collisions by modifying the ids of colliding tabs
[...]

PR #1702, opened 2016-06-28
Add some newui feature info specs

RacingTadpole:
Starts to address a small small part of #1686.


kring:
Looks like a legit failure on IE10 and 11 (extra space before `rgb(...)` on those browsers for some reason).

Issue #1694, opened 2016-06-27
NewUI: Make top right buttons configurable


rdengate:
There's no way for the user to choose which buttons on the top right (compass, locate me, zoom) should be present by changing config options. It would be great to have this, because for the measure tool we're about to have a button there which most maps won't want.

Currently the height of the buttons is hard coded which makes it doubly difficult to have one that's turned on and off.

@AlexGilleran has an idea of how this could be done.

kring:
This was fixed in #1713.

Issue #1659, opened 2016-06-20
New UI mobile: About, Related maps, and Support links go nowhere


kring:
They're just links to the same page.

tephenavies:
@AlexGilleran, @kring: Should this issue be migrated to NICTA/nationalmap? Or possibly TerriaMap? Can ZenHub even move issues across organisations? (NICTA/nationalmap doesn't show up in my "Move Issue" dropdown)

AlexGilleran:
Soooorta. It really belongs to both seeing how we've got to add the ability to do these things then exploit that ability in NM

AlexGilleran:
(note that this isn't really done yet because it still needs a way to work in mobile)

kring:
Done now, right?

PR #1713, opened 2016-06-29
Custom Panels/Buttons and easier DropdownPanels

AlexGilleran:
This adds a `customElements` prop to `StandardUserInterface` that allows extra elements to be injected into the standard user interface without completely recreating it. This allows for #1659 to be fixed in NationalMap (so it can inject its own related maps panel and about button) and also paves the way for #1694 using the same mechanism.

Is this the right way to open up `StandardUserInterface`? I'm not sure, but it works for now and we'll learn by doing it.

In addition, because the idea is that client maps can inject their own `DropdownPanel`s, I've updated this to use javascript to figure out where to place itself and where to place the caret that points at its button. Without this we have to plug exactly where the panel will be into the CSS, which makes it annoying to add buttons/panels to the right of it.

I've also made the leaflet styles global (which they should've been originally).

For a commit that uses this see https://github.com/NICTA/nationalmap/pull/276.


buildprocess/configureWebpack.js
```diff
@@ -2,7 +2,7 @@ var path = require('path');
 var StringReplacePlugin = require("string-replace-webpack-plugin");
 
 // If node-sass starts hanging, uncomment this line:
-//process.env.UV_THREADPOOL_SIZE = 128;
+process.env.UV_THREADPOOL_SIZE = 128;
```
kring, line 5: Did you mean to commit this?  I don't have a problem with it, just want to make sure.

kring:
This looks great!

There's a small z-order problem (not new exactly, but it looks like you tried to fix related problems in this PR) that I'll write a separate issue for.

buildprocess/configureWebpack.js
kring, line 5: Having this line uncommented seems to make my Windows build hang, or at least take a really really long time.

AlexGilleran, line 5: Hmm, _not_ having it makes my build hang :(.

PR #1717, opened 2016-06-30
Fix add data with non-WMS URL

### LINKED ISSUE #1684: Weird behavior when using Add Data for URLs
From #1679:

> When I try to add the URL http://data.gov.au, my localhost:3003 refreshes completely (I've tested in both Chrome and Firefox). Are you seeing that too?

For me it doesn't refresh, but it also doesn't display an error or present the user with the option to use the conversion service.

RacingTadpole:
Fixes #1684.  Was getting stuck in an infinite loop!

However this raises another issue - we have hardcoded all the catalog items to check for into `AddData.jsx`, ie. `registerCatalogMembers`'s calls to `createCatalogItemFromUrl` are ignored. Also, we should write some tests for `AddData.jsx`. New issue on the way...

### DIFF
--- a/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
+++ b/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
@@ -97,7 +97,7 @@ const AddData = React.createClass({
                 loadFunctions = wmsThenWfs.concat(others);
             }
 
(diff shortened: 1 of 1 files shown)


RacingTadpole:
This doesn't seem to solve the problem with _invalid_ URLs - for these, my browser refreshes the page. I've added a new issue #1723 for that.

PR #1727, opened 2016-06-30
Draw lines and polygons

rdengate:
This is the polygon drawing code only (no measuring tool). The measuring tool is still in progress and is waiting on issue #1694.

It's useful to have the polygon drawing code merged now because it's needed for WPS polygon drawing functions.


test/Map/UserDrawing.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
kring: We usually name specs `UserDrawingSpec.js` instead of just `UserDrawing.js`.  It makes it easier to find the right file in the debugger and in editors.

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
+'use strict';
+
+import MapInteractionMode from '../Models/MapInteractionMode';
+
+var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
```
kring: Probably good to avoid mixing styles like this in a single flie; either use `import` or `require` throughout.

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
...
+/**
+ * @alias UserDrawing
+ * @constructor
+ *
+ * For user drawings, which includes lines and/or a polygon
+ * @param {Terria} terria The Terria instance.
+ * @param {Object} options Object with the following properties, all of which are optional:
+ * @param {String} [options.messageHeader] Heading for the dialog which pops up when in user drawing mode.
+ * @param {Bool}   [options.allowPolygon] Let the user click on first point to close loop.
+ * @param {Object} [options.onPointClicked] Callback for when a point is clicked. Function that takes a CustomDataSource.
```
kring: JSDoc has syntax for documenting the signature of callbacks:
http://usejsdoc.org/tags-callback.html

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
...
+ * @param {Object} [options.onCleanup] Callback for when clean up is happening, i.e., for done or cancel.
+ * @param {Object} [options.onMakeDialogMessage] Callback for when the dialog is displayed, to provide a custom message.
+ */
+var UserDrawing = function(terria, options) {
```
kring: We're not 100% consistent about this, but for new code I think it's better to put the `terria` inside `options`.  Mixing the two ways of passing parameters just makes it harder to understand how to call it.

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
...
+
+    if (!defined(terria)) {
+        throw new DeveloperError('Terria instance is required.');
+    }
+
+    // Client is not required to pass options.
+    if (!defined(options)) {
```
kring: You can replace this with `options = defaultValue(options, defaultValue.EMPTY_OBJECT);`

lib/Map/UserDrawing.js
kring: Also, things in `Map` shouldn't depend on `Terria` because it's in `Models`.  If it doesn't make sense to separate this `UserDrawing` class from `Terria`, you can just move this class to `Models` instead.

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
...
+    // Handle what happens when user picks a point
+    knockout.getObservable(pickPointMode, 'pickedFeatures').subscribe(function(pickedFeatures) {
+        when(pickedFeatures.allFeaturesAvailablePromise, function() {
+            if (defined(pickedFeatures.pickPosition)) {
+                var pickedPoint = pickedFeatures.pickPosition;
+                var firstPointEntity = new Entity({
+                    name: 'First Point',
+                    position: pickedPoint,
+                    billboard : {
+                        image : that.svgPoint,
```
kring: I thought Cesium had problems using SVGs in billboards in some browsers.  Did I make that up?

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
...
+UserDrawing.prototype.enterDrawMode = function() {
+
+    // If we have finished a polygon, don't allow more points to be drawn. In future, perhaps support multiple polygons.
+    if (this.inDrawMode || this.closeLoop) {
+        // Do nothing
+        return;
+    }
+
+    this.inDrawMode = true;
+    document.getElementById("cesiumContainer").setAttribute("style", "cursor: crosshair");
```
kring: Don't hardcode `cesiumContainer`.  You can get it from `terria.cesium._mapContainer`.  Might be nice to expose that as a public property instead of accessing the underscore name.

lib/Map/UserDrawing.js
```diff
@@ -0,0 +1,400 @@
...
+
+/**
+ * Called after a point has been added, prepares to add and draw another point, as well as updating the dialog.
+ * @private
+ */
+UserDrawing.prototype._prepareToAddNewPoint = function() {
+    var pickPointMode = this._mapInteractionModeUpdate();
+    var that = this;
+
+    knockout.getObservable(pickPointMode, 'pickedFeatures').subscribe(function(pickedFeatures) {
```
kring: Does this ever get unsubscribed?

kring: I guess the idea is it doesn't matter, because the map interaction mode that you're subscribing to is only temporary.

test/Map/UserDrawing.js
```diff
@@ -0,0 +1,21 @@
...
+describe('UserDrawing', function() {
+    var terria;
+
+    beforeEach(function () {
+        terria = new Terria({
+            baseUrl: './'
+        });
+    });
+
+    it('will use default options if options are not specified', function() {
```
kring: It would be great to see some more tests for this. ;)

kring: For the user drawing in general, I mean.

kring:
Looks great! Just those mostly nitpicky comments.

lib/Models/UserDrawing.js
```diff
@@ -0,0 +1,418 @@
...
+/**
+ * @alias UserDrawing
+ * @constructor
+ *
+ * For user drawings, which includes lines and/or a polygon
+ * @param {Object} options Object with the following properties, all of which are optional except terria:
+ * @param {Terria} [options.terria] The Terria instance.
+ * @param {String} [options.messageHeader] Heading for the dialog which pops up when in user drawing mode
+ * @param {Bool}   [options.allowPolygon] Let the user click on first point to close loop
+ * @callback {onPointClicked} options.onPointClickedCallback Optional way to subscribe to point clicks
```
kring: I think these should be `@param` not `@callback`.

lib/Models/UserDrawing.js
```diff
@@ -0,0 +1,418 @@
...
+ * @return {String} Message to add to dialog
+ */
+
+/**
+ * @alias UserDrawing
+ * @constructor
+ *
+ * For user drawings, which includes lines and/or a polygon
+ * @param {Object} options Object with the following properties, all of which are optional except terria:
+ * @param {Terria} [options.terria] The Terria instance.
```
kring: The brackets around `options.terria` mean the parameter is optional.  Remove them for required parameters, and you can also remove the comment above.  You can specify what default value is used when an optional parameter isn't supplied with something like `[options.allowPolygon=true]`.

kring: The Cesium doc guide does a good job of covering this sort of thing:
https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Documentation/Contributors/DocumentationGuide/README.md

lib/Models/UserDrawing.js
```diff
@@ -0,0 +1,418 @@
...
+     * If true, user can click on first point to close the line, turning it into a polygon.
+     * @type {Bool}
+     * @default true
+     */
+    this.allowPolygon = defaultValue(options.allowPolygon, true);
+
+    /**
+     * Callback that occurs when point is clicked (may be added or removed). Function takes a CustomDataSource which is
+     * a list of PointEntities.
+     * @type {function}
```
kring: `@type` should be `{onPointClicked}`.  Actually my personal preference is to avoid `on` in the type of a callback function.  Something like `PointClickedCallback` is a good name IMO.  `onPointClicked` is fine for the name of the property itself.

lib/Models/UserDrawing.js
rdengate: Thanks, those docs are very good!

PR #1728, opened 2016-06-30
abs buttons styling

chloeleichen:


kring:
image

It's a bit weird the way these icons line up, but definitely much better than it was!  Merging.

chloeleichen:
i agree, i should mandate the caret always on the left and checkbox always on the right, will do that when I get everything else fixed. Thanks @kring

PR #1744, opened 2016-07-01
Ungrouped items in CKAN catalog items are now grouped under an item w…

### LINKED ISSUE #1610: Group uncategorized items in CKAN when using groupBy="group"
Unlike when grouping by organisation, not every CKAN item necessarily has a group. So when `groupBy="group"`, we can end up with many items at the top level, which is a pretty disastrous UX. We should put them all in a (possibly customizably-named) group like "ungrouped".

stevage:
…hose title

is determined by .ungroupedTitle (default: "No group") Fixes #1610.

### DIFF
--- a/lib/Models/CkanCatalogGroup.js
+++ b/lib/Models/CkanCatalogGroup.js
@@ -537,6 +537,8 @@ function createItemFromResource(resource, ckanGroup, itemData, extras, parent) {
     });
 }
 function populateGroupFromResults(ckanGroup, json) {
+    var ungrouped;
+
     var items = json.result.results;
     for (var itemIndex = 0; itemIndex < items.length; ++itemIndex) {
         var item = items[itemIndex];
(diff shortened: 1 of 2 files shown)


lib/Models/CkanCatalogGroup.js
```diff
@@ -548,7 +558,7 @@ function populateGroupFromResults(ckanGroup, json) {
         for (var resourceIndex = 0; resourceIndex < resources.length; ++resourceIndex) {
             var resource = resources[resourceIndex];
 
-            var groups;
+            var groups, ungrouped;
```
kring: This should move outside the loop for clarity.  It doesn't actually matter with the silly `var` scoping rules, but it's misleading to locate it here.  If we replaced `var` with `let`, the code would still run just fine but wouldn't work right anymore (it would create multiple "No group" groups).  I'll move it before merging this.

stevage:
Ok, done. @kring

stevage:
Heh, oops - didn't see you were going to do this so I tried to do it.

PR #1754, opened 2016-07-05
Color scale range

### LINKED ISSUE #774: Allow users to specify COLORSCALERANGE on WMS requests
It's awesome that we can drop THREDDS WMS endpoints in the National Map, and it would be even more awesome if we could specify the color range.  I know this is a non-standard WMS feature, but it's super useful for model products served via THREDDS ncWMS service.   As an example, here's the NOAA Global WaveWatch 3 model significant wave height in the National Map:
6-22-2015 8-59-19 am
which doesn't show the user a whole lot because the default color range is -50,50 meters.

The National Map request currently looks like this:
http://thredds.ucar.edu/thredds/wms/grib/NCEP/WW3/Global/Best?time=2015-06-18T18%3A00%3A00.000Z&transparent=true&format=image%2Fpng&exceptions=application%2Fvnd.ogc.se_xml&styles=&tiled=true&service=WMS&version=1.1.1&request=GetMap&layers=Significant_height_of_combined_wind_waves_and_swell_surface&srs=EPSG%3A3857&bbox=17532819.79994059%2C-7514065.628545966%2C20037508.342789244%2C-5009377.085697312&width=256&height=256
which looks like: 
sample tile

but with the addition of the `COLORSCALERANGE` parameter on the end:
http://thredds.ucar.edu/thredds/wms/grib/NCEP/WW3/Global/Best?time=2015-06-18T18%3A00%3A00.000Z&transparent=true&format=image%2Fpng&exceptions=application%2Fvnd.ogc.se_xml&styles=&tiled=true&service=WMS&version=1.1.1&request=GetMap&layers=Significant_height_of_combined_wind_waves_and_swell_surface&srs=EPSG%3A3857&bbox=17532819.79994059%2C-7514065.628545966%2C20037508.342789244%2C-5009377.085697312&width=256&height=256&COLORSCALERANGE=0.28%2C5.53
it could look like this:
sample tile

It would be cool if this could be specified by the user in the interface on "layer settings" or something.

jiwalker-usgs:
Support changing the color scale range for ncWMS servers.

Fixes #774.


kring:
Thanks @jiwalker-usgs!  I'll review this today.

lib/ReactViews/Workbench/Controls/ColorScaleRangeSection.jsx
```diff
@@ -0,0 +1,50 @@
...
+        this.minRange = event.target.value;
+    },
+
+    changeRangeMax(event) {
+        this.maxRange = event.target.value;
+    },
+
+    render() {
+        const item = this.props.item;
+        this.minRange = item.parameters.colorscalerange.split(',')[0];
```
kring: We crash here when using a WMS server without a `colorscalerange` parameter.

lib/ReactViews/Workbench/Controls/ColorScaleRangeSection.jsx
```diff
@@ -0,0 +1,50 @@
...
+        this.maxRange = item.parameters.colorscalerange.split(',')[1];
+        if (!item.isNcWMS) {
```
kring: This check needs to be moved earlier, and you should probably check for the presence of `colorscalerange` parameter, too, in case it's possible for a server we think is ncWMS to to not have it.

lib/ReactViews/Workbench/Controls/ColorScaleRangeSection.jsx
```diff
@@ -0,0 +1,50 @@
...
+            return null;
+        }
+        return (
+            <div className={Styles.colorscalerange}>
+                <label>Colorscalerange: </label>
+                <label htmlFor="rangeMin">Min: </label>
+                <input type='text' name='rangeMin' defaultValue={this.minRange} onChange={this.changeRangeMin} />
+                <label htmlFor="rangeMax">Max: </label>
+                <input type='text' name='rangeMax' defaultValue={this.maxRange} onChange={this.changeRangeMax} />
+                <button type='button' onClick={this.updateRange} title="Update Range" className={Styles.btn}>Update Range</button>
```
kring: Put this in a form with an `onSubmit` so that the user can update the range by pressing enter.  I think that may fix the styling too.  If not, define `btn` in `colorscalerange-section.scss` to compose the one in `_buttons.scs`.

kring:
Thanks @jiwalker-usgs!  Looks good, just those few comments.  Since you've opened a PR from the branch I created inside the TerriaJS repo, you'll probably need to bring it over to your own repo before you can make changes, and then you'll have to open a separate PR from there.

rsignell-usgs:
@jiwalker-usgs , any chance you can do these small changes before noon tomorrow (July 11)? 
We have a CDI mid-project report at 3pm tomorrow and it would be nice to show this issue resolved so that we have some progress toward our [CDI milestone](https://github.com/TerriaJS/terriajs/milestone/17)!

PR #1755, opened 2016-07-06
Csv with addresses

rdengate:
This uses the GNAF geocoder to translate addresses in csv files to new lat long columns, so the addresses can be shown as data points on the map.


lib/Map/AddressConverter.js
```diff
@@ -0,0 +1,123 @@
...
+ * A promise to convert a single address to lat long coordinates, which resolves to
+ * a length two array with (longitude, latitude) coordinates.
+ *
+ * @param {String} [address] The address to convert
+ * @param {String} [corsProxy] Proxy for cross origin resource sharing
+ * @return {Promise} A promise that eventually returns a pair of coordinates.
+ */
+function convertAddress(address, corsProxy) {
+
+    var gnafApi = new GNAFApi(corsProxy);
```
kring: This is very specific to the Australian GNAF.  Either the name (`AddressConverter`) should be changed to reflect that, or it should be made more generic.

lib/Models/CsvCatalogItem.js
```diff
@@ -419,14 +425,19 @@ function loadTableFromCsv(item, csvString) {
...
     item._tableStructure = tableStructure;
+    // Does the csv have addresses we can translate to long and lat?
+    if (!tableStructure.hasLatitudeAndLongitude && tableStructure.hasAddress)
+    {
+        var addrConverter = new AddressConverter(tableStructure, item.terria.corsProxy);
+        return addrConverter.convertEachAddress().then(function() {
+            return createDataSourceForLatLong(item, tableStructure);
+        })
+        .otherwise(function(e) {
+            console.log("Unable to map addresses to lat-long coordinates.", e);
```
kring, line 559: This otherwise will prevent the error from being raised up higher and being shown to the user.  Either remove it, or add a `throw e` to the end.

lib/Models/CsvCatalogItem.js
```diff
@@ -419,14 +425,19 @@ function loadTableFromCsv(item, csvString) {
     tableStructure.loadFromCsv(csvString);
     setActiveTimeColumn(tableStructure, tableStyle);
     item._tableStructure = tableStructure;
+    // Does the csv have addresses we can translate to long and lat?
+    if (!tableStructure.hasLatitudeAndLongitude && tableStructure.hasAddress)
+    {
+        var addrConverter = new AddressConverter(tableStructure, item.terria.corsProxy);
```
kring: Ok IMO it's not ok for `CsvCatalogItem` to be hard-coded to resolving Australian addresses using GNAF.  This needs to be a plugin point somehow.

kring:
Looks great, just need to find a way to make it less Australia-specific!

rdengate:
Waiting on:
- _msearch to be publicly accessible
- https://github.com/TerriaJS/terriajs/pull/1848
- https://github.com/TerriaJS/terriajs/pull/1842

rdengate:
_msearch is now publicly accessible. Once #1842 and #1848 go through, this is able to be merged.

rdengate:
Dependent PRs have been merged, so this is ready to merge.

lib/Models/Terria.js
```diff
@@ -53,6 +53,7 @@ var defaultConfigParameters = {
  * @param {String} [options.cesiumBaseUrl='(options.baseUrl)/build/Cesium/build/'] The base directory in which Cesium can find its static assets.
  * @param {String} [options.appName] The name of the app.
  * @param {String} [options.supportEmail] The support email for the app.
+ * @param {String} [options.batchGeocoder] Geocoder to use for geocoding addresses in CSV files.
```
kring: Wrong type.

lib/Models/Terria.js
```diff
@@ -110,6 +111,12 @@ var Terria = function(options) {
     this.supportEmail = defaultValue(options.supportEmail, "support@terria.io");
 
     /**
+     * The geocoder to use for batch geocoding addresses in CSV files.
+     * @type {String}
```
kring: Wrong type here too.

lib/Map/AddressConverter.js
```diff
@@ -0,0 +1,38 @@
...
+/*global require*/
+var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
+
+/*
+Determine lat and long coordinates for a given address.
+If TableStructure has address column, adds two new columns: lat and long.
+*/
+
+/**
+ * Instantiate an address converter with a TableStructure that it must convert.
```
kring: This isn't true is it?  The `tableStructure` is passed to `bulkConvertAddresses`, it's not passed to the constructor.

kring: Anyway this should describe the class, not its constructor.  Something like `Geocodes addresses to a longitude and latitude` maybe?

lib/Map/AddressConverter.js
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @alias AddressConverter
+ * @constructor
+ */
+var AddressConverter = function() {
+    /**
+     * Missing addresses from last geocode.
+     * @type {Array}
+     */
+    this.missingAddresses = [];
```
kring: Why not make the promise resolve to this, instead of storing it here?

lib/Map/AddressConverter.js
```diff
@@ -0,0 +1,38 @@
...
+If TableStructure has address column, adds two new columns: lat and long.
+*/
+
+/**
+ * Instantiate an address converter with a TableStructure that it must convert.
+ *
+ * @alias AddressConverter
+ * @constructor
+ */
+var AddressConverter = function() {
```
kring: Is `Converter` the best name for this?  Why not simply `Geocoder`?

lib/Map/GnafAddressConverter.js
```diff
@@ -0,0 +1,146 @@
...
+var when = require('terriajs-cesium/Source/ThirdParty/when');
+var VarType = require('../Map/VarType');
+var AddressConverter = require('../Map/AddressConverter');
+var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
+var GnafApi = require('../Models/GnafApi');
+var defined = require('terriajs-cesium/Source/Core/defined');
+var JulianDate = require('terriajs-cesium/Source/Core/JulianDate');
+
+/*
+Australia-only address converter, which uses the GNAF api to determine lat and long coordinates for a given address.
```
kring: This comment should be in the jsdoc comments below.

lib/Map/GnafAddressConverter.js
```diff
@@ -0,0 +1,146 @@
...
+            matchedAddresses.push(info[j].name);
+            j++;
+        }
+        that.tableStructure.addColumn("Matched Address", matchedAddresses);
+        that.tableStructure.addColumn("Lon", longValues);
+        that.tableStructure.addColumn("Lat", latValues);
+        that.missingAddresses = missingAddresses;
+        return when(true);
+    }).otherwise(function(e) {
+        throw e;
```
kring: An `otherwise` that simply re-throws the error can be removed completely.

lib/Map/GnafAddressConverter.js
```diff
@@ -0,0 +1,146 @@
...
+            longValues.push(info[j].location.longitude);
+            latValues.push(info[j].location.latitude);
+            matchedAddresses.push(info[j].name);
+            j++;
+        }
+        that.tableStructure.addColumn("Matched Address", matchedAddresses);
+        that.tableStructure.addColumn("Lon", longValues);
+        that.tableStructure.addColumn("Lat", latValues);
+        that.missingAddresses = missingAddresses;
+        return when(true);
```
kring: All you need here is `return true;`

lib/Map/GnafAddressConverter.js
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+}
+
+/**
+ * Do not try to geocode addresses that don't look valid.
+ *
+ * @param {Array} addressList List of addresses that will be considered for geocoding
+ * @return {Object} Probably shorter list of addresses that should be geocoded, as well as indices of addresses that
+ *                  were removed.
```
kring: Add `@private`, otherwise this private function will show up in the doc.

lib/Models/CsvCatalogItem.js
```diff
@@ -486,7 +487,17 @@ CsvCatalogItem.setActiveTimeColumn = function(tableStructure, tableStyle) {
...
+/**
+ * Loads the TableStructure.
+ * Chooses what sort of object to place it in:
+ *  - TableDataSource if it has latitude and longitude, or if it has addresses that we can convert to coordinates.
+ *  - RegionMapping if it has a region column
+ *  - a non-geospatial class otherwise (pending)
+ *
+ * @param {CsvCatalogItem} item Item that tableDataSource is created for
+ * @param {String} csvString String in csv format.
+ * @return {Promise} A promise that resolves to true if it is a recognised format.
```
kring, line 499: Need `@private` here too.

lib/Models/CsvCatalogItem.js
```diff
@@ -511,15 +522,45 @@ function setupTableStructure(item, tableStructure) {
...
+
+    // Does the csv have addresses we can translate to long and lat?
+    if (!tableStructure.hasLatitudeAndLongitude && tableStructure.hasAddress && defined(item.terria.batchGeocoder))
+    {
+        var addressConverter = item.terria.batchGeocoder;
+        return addressConverter.bulkConvertAddresses(tableStructure, item.terria.corsProxy).then(function() {
+            var timeTaken = JulianDate.secondsDifference(JulianDate.now(), addressConverter.startTime);
+            var message = "Bulk geocode of " + addressConverter.numberOfAddressesConverted + " addresses took "
+                          + timeTaken.toFixed(2) + " seconds, which is "
+                          + (addressConverter.numberOfAddressesConverted/timeTaken).toFixed(2)
```
kring: Are you sure we want to show all this to the user?

lib/Models/CsvCatalogItem.js
```diff
@@ -548,6 +589,25 @@ function setupTableStructure(item, tableStructure) {
...
 }
 
+/**
+ * Creates a datasource based on tableStructure provided and adds it to item. Suitable for TableStructures that contain
+ * lat-lon columns.
+ *
+ * @param {CsvCatalogItem} item Item that tableDataSource is created for
+ * @param {TableStructure} tableStructure TableStructure to use in creating datasource.
+ *
+ * @return {Promise}
```
kring, line 599: `@private`

lib/Models/CsvCatalogItem.js
rdengate: I think it's nice. The geocode is not that fast, so having a dialog at the end with speed per address seems ok to me. However, I will remove it if you object.

kring:
I don't think the addresses are being resolved correctly.  I dragged the `address_to_lat_long.csv` file onto the map and got this:

image

kring:
Also I noticed that dragging that file onto the map when there's no `batchGeocoder` configured in `Terria` causes a crash in charting code.  I'm not sure if this is a new problem in this branch.

rdengate:
Er, yes, it seems there's been a change in the way the GNAF batch API works :-(. I'm following that up with Neil. 

I was testing in TerriaMap today and the behaviour when dragging a csv that contains addresses when no geocoder was set was to display them by region, so that's interesting... what are you running that in?

kring:
> I was testing in TerriaMap today and the behaviour when dragging a csv that contains addresses when no geocoder was set was to display them by region, so that's interesting... what are you running that in?

NationalMap with index.js changed to use the `GnafAddressConverter`.

rdengate:
Thanks, I will test.

rdengate:
> Also I noticed that dragging that file onto the map when there's no batchGeocoder configured in Terria causes a crash in charting code. I'm not sure if this is a new problem in this branch.

This behaviour is in master too. I don't believe it's related to address geocoding at all. I have logged https://github.com/TerriaJS/nationalmap/issues/308 for this.

rdengate:
> I don't think the addresses are being resolved correctly

Fixed now. Totally my bug and nothing to do with the GNAF API, for the record.

rdengate:
@kring Thanks for the code review. I've addressed all the changes.

lib/Map/AddressGeocoderData.js
```diff
@@ -0,0 +1,37 @@
...
+    this.numberOfAddressesConverted = numberOfAddressesConverted;
+
+    /**
+     * Number of addresses that were null.
+     * @type {Number}
+     */
+    this.nullAddresses = nullAddresses;
+    /**
+     * Addresses that couldn't be geocoded.
+     * @type {JulianDate}
```
kring: Wrong type.

lib/Map/AddressGeocoderData.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+/**
+ * Holds data for what the AddressGeocoder returns.
+ *
+ * @alias AddressGeocoderData
+ * @constructor
+ */
+var AddressGeocoderData = function(startTime,
```
kring: Maybe `BulkAddressGeocoderResult`?

kring:
When I drop that `address_to_lat_long.csv` file I now get:

image

It looks like the geocoding works well though!

PR #1756, opened 2016-07-06
Poll for new csv data

### LINKED ISSUE #1494: Add setting to poll layers for new data (CSV)
On AREMI, we have a service which produces a CSV which is updated every 5 minutes. We would like to have Terria fetch this every N minutes from the same URL and update the information shown on the map when it's enabled so we can have live feature info and map rendering. @RacingTadpole has been working on related periodic data updates for the Charts work (#1278).

RacingTadpole:
This PR adds polling for csvs.  The idea is that this will be extended to other catalog item types, esp. czml. You can test it with http://localhost:3003/#build/terriajs/test/init/polling.json .

There are a few aspects to this, as discussed in #1494:
- 1. We poll for new data in a CsvCatalogItem (either replace or update the existing data), and:
  - a) Redisplay it on the map (if it is spatial).
  - b) Update any displayed features in the Feature Info Panel (if it is spatial).
  - c) Update the Chart Panel (if it is not spatial).
- 2. Poll for new csv data in any `<chart>` tags in the Feature Info Panel, and update those charts.

Key changes:
- Adds a `Polling` model which holds the polling information (seconds, url, etc).
- Knockout-tracks TableStructure column values.
- Use typed arrays to communicate with the WebWorker in the Chart Panel download.

There is one thing that ought to be done before releasing: `polling.idColumns`. Currently lat/lon features are identified by their line number in the file.  However, I'm opening this PR so that we can get these changes into the codebase and get comfortable with them before 15 July.

Fixes #1494.


lib/Map/TableColumn.js
```diff
@@ -164,7 +164,9 @@ var TableColumn = function(name, values, options) {
 
     this._formattedValues = getFormattedValues(this);
 
-    knockout.track(this, ['_type']);  // so that TableStructure can change columnsByType if type changes.
+    // Track _type so that TableStructure can change columnsByType if type changes.
+    // Track _values so that charts can update live with new data.
+    knockout.track(this, ['_type', '_values']);
```
kring, line 169: Did you by any chance check CSV performance before/after this change?  There's at least a possibility that it could add a lot of overhead.

lib/Models/CustomComponentType.js
```diff
@@ -45,16 +45,54 @@ const CustomComponentType = function(options) {
      * @type {Object[]}
      */
     this.furtherProcessing = options.furtherProcessing;
+
+    /**
+     * Gets or sets a function which checks a ReactComponent and returns a Boolean indicating if that react component
```
kring: Even though there's no actual React here, it's a big dodgy that this type in Models talks about ReactComponents so specifically.  Can you state it more generally in the doc?  It's ok to describe it generally and then say "in the standard TerriaJS React UI, this is a ReactComponent" or something like that.

lib/Models/CustomComponents.js
```diff
@@ -1,7 +1,11 @@
 'use strict';
 
+import React from 'react';
```
kring: My previous comment was a bit academic, but it's definitely no good to actually require `React` inside a module in `Models`.

lib/Models/Polling.js
```diff
@@ -0,0 +1,71 @@
...
+     */
+    this.url = options.url;
+
+    /**
+     * Gets or sets a flag which says whether polled data should replace the existing data, or update it.
+     * If defined, the data received is added to the existing TableStructure. Otherwise, it replaces the TableStructure.
+     * @type {Boolean}
+     * @default false
+     */
+    this.replace = options.replace;
```
kring, line 48: Why default this to the unsupported case?

lib/Models/CsvCatalogItem.js
```diff
@@ -562,6 +582,56 @@ CsvCatalogItem.prototype.activateColumnFromTableStyle = function() {
...
+
+/**
+ * Loads data from a URL into a (usually temporary) table structure.
+ * @param  {String} url The URL.
+ * @return {Promise} A promise which resolves to a table structure.
+ */
+function loadIntoTableStructure(url) {
+    // Load in the data file as a TableStructure. Currently only understands csv.
+    const tableStructure = new TableStructure();
+    return loadText(url).then(tableStructure.loadFromCsv.bind(tableStructure));
```
kring: This needs to go through the proxy if so configured.  Use the '0d' form when polling so it doesn't cache.

lib/Models/registerCustomComponentTypes.js
```diff
@@ -143,22 +161,31 @@ const registerCustomComponentTypes = function(terria) {
                 yColumns: yColumns,
                 id: id,
                 canDownload: true,
-                raiseToTitle: !!getInsertedTitle(node)
+                raiseToTitle: !!getInsertedTitle(node),
+                pollSources: pollSources,
+                pollSeconds: node.attribs['poll-seconds'],
+                pollReplace: (node.attribs['poll-replace'] === 'true'),
+                updateCounter: updateCounter,  // Change this to trigger an update.
+                // pollSource: node.attribs['poll-src'],
```
kring: remove commented-out code.

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -371,4 +377,79 @@ function describeFromProperties(properties, time) {
...
+    const foundCustomComponents = CustomComponents.find(info);
+    foundCustomComponents.forEach((match, componentNumber) => {
+        const updateSeconds = match.type.selfUpdateSeconds(match.reactComponent);
+        if (updateSeconds > 0) {
+            setTimeoutForUpdatingCustomComponent(that, match.reactComponent, updateSeconds, componentNumber);
+        }
+    });
+}
+
+function setTimeoutForUpdatingCustomComponent(that, reactComponent, updateSeconds, componentNumber) { // eslint-disable-line require-jsdoc
```
kring, line 429: I don't understand why all this logic is needed in the UI.  Entity has a `definitionChanged` event that is raised when any of its properties (e.g. `description`) changes.  The UI should only need to subscribe to that.  All the logic for actually updating the Entity based on polling or whatever else should be elsewhere.  Or am I missing something?

kring:
That's all I have for now!  This is very cool!

lib/Models/Polling.js
RacingTadpole, line 48: Yes, it does seem a bit obtuse, but I'm trying to think ahead: eg. for CZML the natural default will be to update, not replace. For clarity I suggest we make this the default for all item types (although I don't really have a strong opinion for CSV itself). 
Since this will be locked in once we release, I want to establish this default now, even though it's not yet supported.

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
RacingTadpole, line 429: Re logic in the UI - Note this piece only handles self-updating `<chart>` tags (ie. point 2 in my description at the top of this PR). These are only rendered when you've opened a Feature Info Section, and they don't actually change any of the feature's properties or description - so `definitionChanged` would never get triggered. So I saw <chart> updates as updates to the UI, not the feature. Is that plausible?

Using `definitionChanged` might simplify the implementation of point 1b in my description, though.

kring, line 429: Ok, that makes sense.  It still bothers me a bit that this logic is here, though, especially given how complicated it is.  For example, if we wanted to add support for embedded charts on the Workbench, we'd have to duplicate all this logic, right?  I guess we can just plan to refactor later when we get to that, but it might be worth at least thinking about how that might work.

RacingTadpole, line 429: Yes, true. One simple solution could be to create a reusable React Component, eg. `<CustomContent>`, that effectively just calls `renderMarkdownInReact`, and sets up the above timeout logic.

kring, line 429: That sounds good.  I'll leave it to you if you think that is a sensible thing to do now, or if it can just as easily wait until we need it later.

lib/Map/TableColumn.js
RacingTadpole, line 169: Here are some tests with a 55,000-line region-mapped csv and a stopwatch on my computer, with Cesium:

```
             load     change var to First Letter
newui        8.5s     4.5s
newui_poll   7.4s     4.3s
```

There's probably a wide margin of error on this :-), but at least it seems ok. 

I also tried loading ABS Age data (and changing to SA2), and they were similarly quick.

Do you have any other test files in mind?

kring, line 169: Nope, that's great, thanks!

lib/Models/CustomComponents.js
RacingTadpole: There might be a bigger issue here, I've just realised - `Models/registerCustomComponents` also requires `React`.
Would you agree that one ought to go in a new `ReactCustom` directory (or similar)?
Actually I'm leaning towards just putting it in `ReactViews/Custom`.

kring: Yeah `ReactViews/Custom` makes sense to me.

RacingTadpole:
OK, back to you thanks Kevin. This was a very good change to have made before we go live with the new UI - I've shifted 4 files from Models (and one from Core! :-o) into ReactViews/Custom.
This will require a change in path in TerriaMap / index.js too.

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
RacingTadpole, line 429: Hmm. I've used the `Feature` class to hold the component-updating information, so making this work for a non-feature (eg. a standalone chart) will take more thought than I have right now. I think it won't make a huge difference to the public API though, so we can do it in a later version.

PR #1766, opened 2016-07-07
Fix displayName error

RacingTadpole:
Fixes this error:
`Unknown prop `displayName` on <a> tag`


kring:
So this PR makes eslint complain at build time.  Seems like the error is spurious in this case, because we're not actually defining a component.  So should we just disable the eslint warning locally?

PR #1791, opened 2016-07-12
This -> That

stevage:


lib/Models/CkanCatalogGroup.js
```diff
@@ -382,7 +382,7 @@ CkanCatalogGroup.prototype._load = function() {
 
     }
     this.filterQuery.forEach(function(query) {
-        var uri = new URI(this.url)
+        var uri = new URI(that.url)
             .segment('api/3/action/package_search')
             .addQuery({ rows: 100000 });
```
kring, line 388: There's another `this` further down in this function.

Issue #1777, opened 2016-07-09
CSW hierarchical catalogue problems


meh9:
It appears that sometime in the last couple of months, the code that parses a CSW catalogue and turns it into a hierarchical tree has changed.

The AREMI Wave Atlas categories now shop up as two long lists of layers instead of having subcategories like they used to. (There is one exception, the Wave Model subcategory of "Direction of Maximum Directionally Resolved Wave Power", I don't know what makes this different.) See this share which has two layers enabled, one in each long flat list:

http://nationalmap.gov.au/renewables/#share=MMbz3c

The Wave Atlas guys run their own frontend that loads the same catalogue in the exact same way we do but using older code, and they still have the layers in subcategories. Go to this URL and then see the categories in "Renewable Energy -> Wave -> Australian Wave Atlas (I couldn't get the sharing to work):

http://awavea.csiro.au/

(Note that this is using the OldUI, but the results are currently the same in the NewUI, the AREMI NewUI sharing doesn't work at the moment.)

kring:
Fixed by #1777.

PR #1792, opened 2016-07-13
Fix for CSW hierarchical catalogue problems #1777 plus test

rdengate:
This fixes the problem with the catalog item levels being flattened as described in #1777.

### DIFF
--- a/test/Models/CswCatalogGroupSpec.js
+++ b/test/Models/CswCatalogGroupSpec.js
@@ -1,0 +1,24 @@
+'use strict';
+
+/*global require,describe,it,expect*/
+var CswCatalogGroup = require('../../lib/Models/CswCatalogGroup');
+
+describe('CswCatalogGroup', function() {
+    it('findLevel function test', function() {
+        var metadataGroups = [];
+
+        var keyList = [["Multiple Use", "National"],
+                       ["Multiple Use", "Western Australia"],
+                       ["Wave Models", "Direction of Maximum Directionally Resolved Wave Power", "Averages"],
+                       ["Wave Models", "Maximum Directionally Resolved Wave Power", "Averages"]];
+        for (var i=0; i<keyList.length; i++) {
+            var keys = keyList[i];
+            CswCatalogGroup._findLevel(keys, 0, metadataGroups, " | ", "subject");
+        }
+
+        // If you're trying to debug this, run it through a json formatter. Then the structure becomes clear.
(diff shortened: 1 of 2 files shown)


test/Models/CswCatalogGroupSpec.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+/*global require,describe,it,expect*/
+var CswCatalogGroup = require('../../lib/Models/CswCatalogGroup');
+
+describe('CswCatalogGroup', function() {
+    it('findLevel function test', function() {
```
kring, line 7: Bit of a nitpick, but Jasmine specs should generally be named so they read like a specification in English.  So if you have:

```
describe('foo', function() {
    it('should return 4', function() {
    });
});
```

you can read that as "foo should return 4".  Or even include the function name "it should return 4".

test/Models/CswCatalogGroupSpec.js
```diff
@@ -0,0 +1,24 @@
...
+        var strExpected = '[{"field":"subject","value":"^Multiple Use\\\\ \\\\|\\\\ ","regex":true,"group":"Multiple Use","children":[{"field":"subject","value":"Multiple Use | National","regex":false,"group":"National"},{"field":"subject","value":"Multiple Use | Western Australia","regex":false,"group":"Western Australia"}]},{"field":"subject","value":"^Wave Models\\\\ \\\\|\\\\ ","regex":true,"group":"Wave Models","children":[{"field":"subject","value":"^Wave Models\\\\ \\\\|\\\\ Direction of Maximum Directionally Resolved Wave Power\\\\ \\\\|\\\\ ","regex":true,"group":"Direction of Maximum Directionally Resolved Wave Power","children":[{"field":"subject","value":"Wave Models | Direction of Maximum Directionally Resolved Wave Power | Averages","regex":false,"group":"Averages"}]},{"field":"subject","value":"^Wave Models\\\\ \\\\|\\\\ Maximum Directionally Resolved Wave Power\\\\ \\\\|\\\\ ","regex":true,"group":"Maximum Directionally Resolved Wave Power","children":[{"field":"subject","value":"Wave Models | Maximum Directionally Resolved Wave Power | Averages","regex":false,"group":"Averages"}]}]}]';
```
kring, line 20: Maybe clearer to put a JavaScript object literal here instead of a string, and `JSON.stringify` the expected value too?

kring:
I'm just nitpicking tests, so I'm going to merge this..  If you feel like making those improvements to the tests you can do a separate PR, or even do it right in master if it's straightforward.

rdengate:
Thanks @kring! I've made the changes and pushed straight to master.

PR #1793, opened 2016-07-13
Added New API for Customising Standard User Interface

AlexGilleran:
So this is a pretty big one in terms of complexity inside TerriaJS, but I reckon it delivers a lot of simplicity for the actual maps that use TerriaJS so it's worth it overall. While trying to figure it all out it went through a number of iterations - we might be able to refine it further but this is a start.

At its most basic level what this does is allow users of Standard User Interface to add extra items to the menu in such a way that they get added as buttons on the top of the map in desktop, and links in the menu on mobile.

How it does this is a big shift. With this change, customisations to `StandardUserInterface` can now be passed as children to `StandardUserInterface` using JSX as the API, in a similar way to how React Router uses JSX to specify its routes even though Routes aren't actual elements. So a customised `StandardUserInterface` will look something like this:

```
<StandardUserInterface>
    <Menu>
        <MenuItem href="about.html" />
        <MenuPanel>
            <div>Custom panel</div>
        </MenuPanel>
    </Menu>
    <Nav>
        <button onClick={onButtonClick}>Button</button>
    </Nav>
</StandardUserInterface>
```

As part of this I've created the `MenuItem` and `MenuPanel` components - these provide a means for users of TerriaJS to determine what they want to add semantically, which automatically figure out where/how they should render based on `viewState.useSmallScreenInterface`. So a `MenuItem` is either a link on the menu or a button on the map that goes to an href or triggers an onclick, and `MenuPanel` is also either a link on the menu or a button on the map, but it drops down a panel, either under the button if on desktop or to the left of the screen for mobile.

Longer-term this could be a very general-purpose way of customising the interface - it could just as easily handle branding, for instance, or if users wanted to add extra elements to the sidebar. More generally it provides a way of decoupling what goes into the interface from where it goes, which might be interesting in the future - you could define a set of components to go into a map and then separately determine how you wanted them laid out.

See my PR on TerriaMap to see how it works more concretely: https://github.com/TerriaJS/TerriaMap/pull/36


kring:
Crisp!

Issue #1778, opened 2016-07-11
Poor performance adding WMS with lots of time intervals


kring:
Adding this WMS group using the Add Data feature takes ages (I haven't seen it complete yet) and locks up the UI in the process:
http://geoport.whoi.edu/thredds/wms/coawst_4/use/fmrc/coawst_4_use_best.ncd?service=WMS&version=1.3.0&request=GetCapabilities

The time is spent building time intervals.  We should be able to make this go much faster.

meh9:
I believe that this is actually affecting AREMI as well - most of our timeseries layers now causes the browser to become completely unresponsive.

To test, build the latest aremi-natmap#newui, and try loading either of these two layers:

Renewable Energy -> Solar -> Solar Satellite DNI - 2014
Electricity Infrastructure -> Network Opportunities -> Available Capacity

It must have broken fairly recently - the current newui build we have in prototype is a week or two old, and these layers work fine there.

kring:
I think that might be a separate problem, but in any case it sounds bad!  I'll take a look today.

meh9:
Ok, thanks, let me know if you want me to log another issue, I just thought these sounded very similar.

kring:
Yeah, they do sound similar.  The reason I say they're probably different is because the performance problem with the geoport WMS was very specific to the WMS interval handling.  No need to write a separate bug, at least until I investigate a bit more.

kring:
Removed AREMI label because the problem described by @meh9 above was a separate problem and is now fixed.
Nahid to take a look at this.

PR #1794, opened 2016-07-13
Change the markdown-is-really-just-HTML regex.

kring:
The old one was totally hanging the browser when processing some text, such the disclaimer for the AREMI `Solar Satellite DNI 2014` dataset.

Fixes the problem @meh9 described in #1778, but not the original #1778 problem.


kring:
@rdengate can you verify that you don't have the chart-inside-p tag problem you saw before with this PR?

rdengate:
@kring yes, there's no problem with this PR. Thanks!

PR #1799, opened 2016-07-13
Measure tool

rdengate:
This has CSS but no code for hooking it into the UI yet. I'm not 100% clear on what the code hooking it into the UI will look like yet, but I figure that would make a good separate PR anyway. The upshot of that is that the CSS might change.


kring:
Looks good to me!  @rdengate can you update CHANGES.md?

rdengate:
Thanks @kring! Have updated changelog.

PR #1823, opened 2016-07-19
Default label for graph x axis download

AlexGilleran:
Used to be blank, now it's 'date'

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,9 +3,9 @@ Change Log
 ==========
 
 ### 4.0.3
+
 * Made the column title for time-based CSV exports from chart default to 'date'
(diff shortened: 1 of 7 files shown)


RacingTadpole:
Cool - I see, so as long as real changes don't occur within 100ms of each other, it won't recalc the downloadable csv data. I didn’t know about `debounce`.

Wouldn't it be simpler to perform all the updates on a local variable and then set the item equal to the local variable, so it only triggers a single change? Not that I've looked into whether that makes sense at all here.

kring:
This is ok by me, but I'll let @RacingTadpole merge.  We have other places that do manual debouncing, `SearchBox.jsx` and `MouseCoords.js`.  It could be worth using the lodash one now that we've taken that dependency?

PR #1825, opened 2016-07-19
Prevent text select when drag compass

chloeleichen:
Prevent text select when drag compass, fixing https://github.com/TerriaJS/terriajs/issues/1809


kring:
Needed CHANGES.md update, so I did it in master.

PR #1830, opened 2016-07-19
Add charts, customisable tags and polling to Changes.md

RacingTadpole:
A little terse, but at least they're mentioned now.
Are you ok with my linking them back to comments in the source code?
I missed lots of other things though, eg. vector tile capability, polygon drawing.


kring:
Yeah that's fine, though it'd be nice in the future to be able to link to documentation instead. ;)

RacingTadpole:
very true... maybe we should add that to our CI too!

kring:
> maybe we should add that to our CI too!

Not sure what you mean?

PR #1835, opened 2016-07-19
Made reordering work in IE9+

### LINKED ISSUE #1819: Reordering of layers broken in Firefox
Reordering tile based layers does not work in Firefox. There are no error messages, the layers just won't budge.

I'm using Firefox v47.0.1.

AlexGilleran:
Replaced our HTML5 drag-and-drop solution with react-anything-sortable. This uses touch/mousedown events to drag in one dimension and is a lot more reliable in my experience, and also works all the way back to IE8. Plus we get to get rid of all our fiddly drag and drop code and never debug it again.

Fixes #1819


kring:
This looks awesome!  I do see one serious problem though: scrolling workbench items no longer works on mobile.

AlexGilleran:
As in on a mobile phone in the mobile view? I'm guessing you're testing on Android? Works for me on iOS.

kring:
> Works for me on iOS.

Yeah, I was on Android, but it's working fine for me now too.  :confused:   Merging!

PR #1838, opened 2016-07-20
make preview number on mobile bigger when single digit

### LINKED ISSUE #1781: NewUI Mobile Issues
# In rough groupings of priority:
### High
- [x] Scrolling the catalogue on iOS moves the map underneath instead
### Med
- [x] The marker showing the selected location search result doesn't appear on mobile.
- [x] Remove the quick add button on mobile, replace it with a chevron pointing right
- [x] Geolocation button needs to be bigger and spaced evenly from the side of the screen
- [x] The feature info comes up under the location button (but Kevin has fixed it I think?)
- [x] Top row of feature info should be the same height as the top row of the catalog (the row with the "Done" button) so that the close button can be tapped
- [x] Top row of selected layers should also be that height so that the show/hide button can be tapped
- [x] Search button needs to be wider and have space to the left of it (a divider) to make it more tappable.
### Low
- [x] Catalog items need more padding above and below
- [ ] ~~The add-to/remove-from-map button on the preview screen would be better if it were slightly thinner than the screen, and was white when not added and dark grey when added~~ Phil changed his mind on this one.
- [x] Remove the icon from the branding to save space - possibly responsively re-introduce it as width increases?
- [x] Change the DATA button to Data
- [x] The menu icon is aligned slightly wrong on iOS
- [x] The padding on the bar at the top of the screen is slightly wrong on iOS
- [x] The small number indicating how many layers are visible should change to a full-size number as width becomes available.
- [x] The opacity slider should be full width
- [] ~~The search box doesn't need the black frame around it.~~
### Super-Low
- [x] Search doesn't work at all :)

chloeleichen:
This fixes https://github.com/TerriaJS/terriajs/issues/1781#issuecomment-233806737

### DIFF
--- a/lib/ReactViews/Mobile/MobileHeader.jsx
+++ b/lib/ReactViews/Mobile/MobileHeader.jsx
@@ -6,6 +6,7 @@ import Branding from '../SidePanel/Branding.jsx';
 import Styles from './mobile-header.scss';
 import Icon from "../Icon.jsx";
 import MobileMenu from './MobileMenu';
+import classNames from 'classnames';
 import { removeMarker } from '../Search/SearchMarkerUtils';
 
 const MobileHeader = React.createClass({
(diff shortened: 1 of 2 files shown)


chloeleichen:
screen shot 2016-07-20 at 10 55 48 am
screen shot 2016-07-20 at 10 55 33 am

PR #1842, opened 2016-07-20
Show load indicator when user-added file is loading

### LINKED ISSUE #1820: Load indicator for csv item
When dragging and dropping a csv, there's no indication that it's loading, so if it takes a long time to load the user is left wondering if anything is going on. In addition, there's no way to cancel.

rdengate:
Adds a temporary catalog item while the file is loading, which means a loading indicator is shown. Fixes #1820.


lib/Models/addUserFiles.js
```diff
@@ -1,5 +1,8 @@
+'use strict';
+
 import addUserCatalogMember from './../Models/addUserCatalogMember';
 import createCatalogItemFromFileOrUrl from './../Models/createCatalogItemFromFileOrUrl';
+import AsyncFunctionResultCatalogItem from './../Models/AsyncFunctionResultCatalogItem';
```
kring: Clever use of `AsyncFunctionResultCatalogItem` here, but maybe we should rename it, since this has nothing to do with async functions.

lib/Models/addUserFiles.js
```diff
@@ -23,28 +25,42 @@ add data from a file on your system.  Please upgrade your web browser.  For the
...
         }
+        tempCatalogItem.loadPromise = loadPromise;
+        tempCatalogItem.isEnabled = true;
+        tempCatalogItemList.push(tempCatalogItem);
     }
 
     return when.all(promises, addedItems => {
-        return addedItems.filter(item => item && !(item instanceof TerriaError));
+        const items = addedItems.filter(item => item && !(item instanceof TerriaError));
+        tempCatalogItemList.map(function(value) { terria.catalog.userAddedDataGroup.remove(value); });
```
kring: `forEach` would be a little cleaner than `map` here, since you're not using the result at all.

kring:
Can we find a way to flip to the `User-Added Data` list and show the loading spinner there, too?  After this PR, the item appears immediately in the workbench with a `Loading...` indicator, which is a huge improvement.  But on smaller screens, the workbench is fairly hidden.  It'd be great to show signs of progress on the catalog page itself, too.

RacingTadpole:
Hi @rdengate, this looks great! Did you come back to @kring's comment above? @kring, I'm not sure I understand you - on a small screen, the "Add Data" panel closes when you add it, right?

rdengate:
@RacingTadpole thanks! Yep, Kevin and I had a discussion in Slack about it, which resulted in https://github.com/TerriaJS/terriajs/issues/1880.

lib/Models/ResultPendingCatalogItem.js
```diff
@@ -25,12 +25,12 @@ var AsyncFunctionResultCatalogItem = function(terria) {
     knockout.track(this, ['loadingMessage', 'shortReport']);
```
RacingTadpole: Wondering if "shortReport" belongs in `ResultPendingCatalogItem`, now this has been repurposed? I notice that `invokeTerriaAnalyticsService` and `PlacesLikeMeFunction` don't use `shortReport`, so we may be ok just dropping it. @kring?

Similarly, can you change the "typeName" from "Analysis in Progress"?

kring: `shortReport` is a property on `CatalogMember` now, so yeah, we can get rid of it here.

kring: See `WhyAmISpecialFunction.js` for an example of something that uses it (but not on the `ResultPendingCatalogItem` specifically).

rdengate:
Thanks @RacingTadpole. I've made the changes you suggested.

PR #1843, opened 2016-07-20
hide legend image if there is an error

### LINKED ISSUE #1773: Broken legend image symbol in newui
The Topography > ... > Aspect in AREMI shows like this in the new ui:

aremi aspect

In the old UI, there's no legend image.

old aspect

chloeleichen:
This fixes https://github.com/TerriaJS/terriajs/issues/1773

It makes the legend image behaves the same with the old UI: when there is an error loading the legend image, the image is hidden: 

screen shot 2016-07-20 at 4 44 24 pm

### DIFF
--- a/lib/ReactViews/Workbench/Controls/Legend.jsx
+++ b/lib/ReactViews/Workbench/Controls/Legend.jsx
@@ -55,7 +55,7 @@ const Legend = React.createClass({
                                     <When condition={legend.isImage && legend.insertDirectly}>
                                         <li key={i}
                                             onError={this.onImageError.bind(this, legend)}
-                                            className={Styles.legendSvg, {[Styles.legendImagehasError]: legend.imageHasError}}
+                                            className={classNames(Styles.legendSvg , {[Styles.legendImagehasError]: legend.imageHasError})}
                                             dangerouslySetInnerHTML={legend.safeSvgContent}
                                         />
                                     </When>
--- a/lib/ReactViews/Workbench/Controls/Legend.jsx
+++ b/lib/ReactViews/Workbench/Controls/Legend.jsx
@@ -1,14 +1,13 @@
 'use strict';
 
+import classNames from 'classnames';
 import defined from 'terriajs-cesium/Source/Core/defined';
-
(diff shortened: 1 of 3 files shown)


lib/ReactViews/Workbench/Controls/Legend.jsx
```diff
@@ -17,6 +18,11 @@ const Legend = React.createClass({
 
     onImageError(legend) {
         legend.imageHasError = true;
+        this.forceUpdate();
+    },
+
+    componentWillMount() {
+        this.legends = this.getLegends();
```
kring: I think doing this on mount only will cause problems.  The list of legends may change asynchronously as the catalog item loads, and the UI needs to react to that.

lib/ReactViews/Workbench/Controls/Legend.jsx
```diff
@@ -17,6 +18,11 @@ const Legend = React.createClass({
 
     onImageError(legend) {
         legend.imageHasError = true;
+        this.forceUpdate();
```
kring: I'd probably use knockout computed properties for this.  If you move the `this.props.item.legendUrls.map(...)` bit into the render function, and make it look something like this:

``` js
const legends = this.props.item.legendUrls.map(legendUrl => {
                 const legend = {
                     url: proxyCatalogItemUrl(this.catalogMember, legendUrl.url),
                     isImage: legendUrl.isImage(),
                     imageHasError: false,
                     onImageError: this.onImageError,
                      insertDirectly: !!legendUrl.safeSvgContent, // we only insert content we generated ourselves, not arbitrary SVG from init files.
                      safeSvgContent: {__html: legendUrl.safeSvgContent}
                  };
                  knockout.track(legend, ['imageHasError']);
                  return legend;
 });
```

The render should happen automatically if there's an error loading the image.

The only danger I can think of is the browser will try loading each legend again each time render is triggered.  The only really reliable way to fix that, though, is to put more of this state into the `LegendUrl` model object instead of keeping it in the UI.

chloeleichen:
@kring can you have another look please?

PR #1847, opened 2016-07-21
make opacity slider look like design

chloeleichen:
Fixes https://github.com/TerriaJS/terriajs/issues/1846, make opacity look like better
screen shot 2016-07-21 at 11 26 52 am

Note that it stops working after hot reload, need to refresh the page


kring:
@philipgrimmett previously requested (see #1781) "the opacity slider should be full width".  It's ok to have it on the same line as the Opacity label and percentage, IMO, but it should use the remaining width, or at least close to it.

chloeleichen:
screen shot 2016-07-22 at 10 26 38 am
done ! @kring

PR #1848, opened 2016-07-21
Bulk geocoding for GnafApi

### LINKED ISSUE #1786: Improve GNAF Scoring
Scoring on a fuzzy result in elastic search makes things weird, so it's recommended we make our GNAF queries look like this:

```
"{"query":{"match":{"d61Address":{"query":"405 CAMBERWELL ROAD  CAMBERWELL VIC 3124","fuzziness":2,"prefix_length":2}}},"rescore":{"query":{"rescore_query":{"match":{"d61Address":{"query":"405 CAMBERWELL ROAD  CAMBERWELL VIC 3124"}}},"query_weight":0}},"size":10}"
```

This does the search fuzzily but the scoring unfuzzily.

rdengate:
Added ability to bulk geocode addresses. This also fixes #1786 because I changed the scoring at the same time.

Note: this won't work until the _msearch URL becomes publicly accessible, but that should happen within the next few days. Please don't merge yet. I will request a merge when the mearch URL is accessible.


rdengate:
_msearch is now accessible, and this is able to be reviewed and/or merged. Thanks!

lib/Models/GnafApi.js
```diff
@@ -0,0 +1,311 @@
...
+ */
+GnafApi.prototype.bulkGeoCode = function(searchTermList, rectangle, batchSize, maxSimultaneousRequests) {
+    batchSize = batchSize || DEFAULT_BATCH_SIZE;
+    maxSimultaneousRequests = maxSimultaneousRequests || DEFAULT_MAX_SIMULTANEOUS_REQUESTS;
+    var searchTermBatches = splitIntoBatches(searchTermList, batchSize);
+    var searchTermBatchesRequests = splitIntoBatches(searchTermBatches, maxSimultaneousRequests);
+
+    var i = 0;
+    var mergedResults = [];
+    return this._runThrottledRequests(i, mergedResults, searchTermBatchesRequests, rectangle);
```
AlexGilleran, line 101: Is this done recursively so you can time the batches properly? It's a good solution but it might be worth looking into having `runThrottledRequests` return the promise for the batch it's passed, then doing something like:

```
searchTermBatchesRequests
    .map(requests => this._runThrottledRequests(requests, rectangle))
    .reduce((accumulatedPromise, batchPromise) => accumulatedPromise.then(batchPromise), when())
    .then(resultsOfBatches => resultsOfBatches.reduce((acc, batchResult) => acc.concat(batchResult), []))
```

I'm happy with merging it like this too though.

AlexGilleran:
Looks good, personally I think handling batches by reducing an array of promises with `then` is a bit neater than doing it recursively but either way is fine :)

AlexGilleran:
Is there a test CSV for this btw? We should put it on so it's accessible with `#test` in the URL if so.

lib/Models/GnafApi.js
rdengate, line 101: That is nice! I haven't really used reduce before. I might not change it at this stage, but thanks for the suggestion. I'll bear this in mind for future work.

rdengate:
Thanks for the code review @AlexGilleran! For the CSV, I assume you mean one with addresses? It wouldn't geocode just with this PR. I'll add it into the PR for geocoding addresses though as I agree, that would be great to have.

AlexGilleran:
Oh so this is just progress towards having that? No worries then :). Happy to :shipit: if you fix the merge conflicts.

rdengate:
The PR for geocoding addresses in csv files is https://github.com/TerriaJS/terriajs/pull/1755.

I have fixed the merge conflicts. Thanks!

PR #1851, opened 2016-07-21
remove display name on div

chloeleichen:
@RacingTadpole:

I'm getting an error on Aremi -> All Power Stations: 

```
warning.js:44 Warning: Unknown prop `displayName` on <td> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in td (created by FeatureInfoSection)
    in tr (created by FeatureInfoSection)
    in tbody (created by FeatureInfoSection)
    in table (created by FeatureInfoSection)
    in div (created by FeatureInfoSection)
    in section (created by FeatureInfoSection)
```

Because `displayName` is added to a div. This is fixed after removing it. However, in the warning, it says `Unknown prop`displayName`on <td> tag`, but the displayName is actually added to `div`, so I am not sure why it's complaining about `td`. Any ideas?

### DIFF
--- a/lib/ReactViews/Custom/registerCustomComponentTypes.js
+++ b/lib/ReactViews/Custom/registerCustomComponentTypes.js
@@ -39,7 +39,7 @@ const chartAttributes = [
     'poll-replace'
 ];
 
-/** 
+/**
  * @private
  */
 function splitStringIfDefined(string) {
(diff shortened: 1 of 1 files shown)


kring:
I can't see any harm in removing an invalid attribute from the `div`, so I'm merging this.  It might be an incomplete fix, though, because the `displayName` doesn't really belong on the `td` below the changed line, either.  eslint thinks it does, but that's just because it's overzealous in its detection of functional-style React components, which this is not.

PR #1853, opened 2016-07-21
Unmount+Destroy preview map when explorer panel is hidden

### LINKED ISSUE #1832: Timeseries performance problem
There appears to be a performance problem when previewing and the adding and playing a timeseries layer.
## Steps to see problem:
1. Open the catalogue and add click to preview the particular timeseries layer.
2. Add the layer to the map.
3. Press play on the timeseries.
   Observe how when the timeseries is playing through 2016 it progresses slowly, and one into 2017 it starts progressing very jerkily.
## Steps to show the layer without performance problem:
1. Do the steps above and then generate a share url.
2. Follow the share url which adds the layer to the map without going through the catalogue layer preview step.
3. Press play on the timeseries.
   Observe how the entire timeseries plays smoothly without performance problems.

AlexGilleran:
Fixes #1832 (now tested!)

Right now the explorer panel is in the background and the preview map is still rendering and updating even when it's hidden. This turns into a major issue when playing time-series data, as the preview map is still connected up to the global terria clock, so if it was left previewing a time-series layer it'll continue re-rendering. This turns into a major-super-duper problem when rendering heavy time series vector layers.

So all this does is puts the explorer panel behind an `<If>` and adds the same logic to destroy the TerriaViewer as the main map has.


kring:
- Click a dataset so it's previewed.
- Click Done
- Click Add Data

The panel appears immediately, then disappears, then fades back in.

AlexGilleran:
Fixed - because the preview map gets created on a `ref` callback it results in a tonne of work being done in the gap between the explorer window rendering and the explorer window getting its `enter` style, enough to actually cause them to be done over two separate DOM updates and hence causing this behaviour. So I put the preview map setup in a `setTimeout` which should both fix that and make the explorer panel pop up a bit smoother.

kring:
Sorry, still having problems.  Now if I do the procedure above, the preview map is blank when the panel pops back up.  There's also a noticeable delay (feels like up to half a second maybe) between when I click Add Data and when the panel appears, but maybe that was always true.

kring:
By clicking around wildly in the catalog, I also saw cases where the preview wouldn't appear, and cases where it would crash:

```
destroyObject.js:45 Uncaught DeveloperError: This object was destroyed, i.e., destroy() was called.
Error
    at new DeveloperError (http://localhost:3001/build/nationalmap.js:571:19)
    at new throwOnDestroyed (http://localhost:3001/build/nationalmap.js:45912:19)
    at throwOnDestroyed.<anonymous> (http://localhost:3001/build/nationalmap.js:56700:14)
    at NewClass.fireEvent (http://localhost:3001/build/nationalmap.js:47850:25)
    at HTMLImageElement._tileOnLoad (http://localhost:3001/build/nationalmap.js:50380:10)
    at HTMLImageElement._tileOnLoad (http://localhost:3001/build/nationalmap.js:93850:38)
```

AlexGilleran:
I've made some more changes to make sure everything's cleaned up when the preview map mounts/unmounts, and also added a property to viewstate that makes the preview map not load until the explorer window has finished animating. This stops it from going at a low framerate because it's both loading the preview and animating at the same time - it's a problem outside of this branch too.

kring:
`preview-placeholder.png` doesn't appear to be used.

kring:
Something is still not right.  Now, by clicking around randomly, I can get into a state where the preview map doesn't reflect the selected layer.  Usually it's one click behind.  So if I click on "Mobile Black Spot Database" the blue dots that should be shown for that later don't appear.  But then if I click on something else, I'll see those blue dots instead of whatever I clicked on.  I can reproduce this reliably, so let me know if you can't and I'll show you.

AlexGilleran:
Yeah turns out I was using `this.props` instead of `newProps` in `componentWillReceiveProps` :(

PR #1854, opened 2016-07-21
Fix vector tile feature selection for vector-tile@1.3.0

tephenavies:
Fixes problems with and applies Greenkeeper PR #1822.


package.json
```diff
@@ -40,7 +40,8 @@
     "mustache": "^2.2.1",
     "mutationobserver-shim": "^0.3.1",
     "pbf": "^1.3.2",
-    "point-in-polygon": "1.0.0",
+    "point-geometry": "^0.0.0",
```
kring, line 43: Does this need to be mentioned in LICENSE.md?

package.json
```diff
@@ -40,7 +40,8 @@
...
+    "point-in-polygon": "steve9164/point-in-polygon",
```
kring: Why the fork?  We should put it under the TerriaJS org, at least.

tephenavies: Maybe a fork is the wrong thing, but before I was doing point-in-polygon calculations with a ring and a point from GeoJSON data. The ring was an array of `[x,y]` coordinates. Now I'm doing the calculations with a ring that isn array of `Point(x,y)` coordinates. Basically, I changed every `ring[i][0]` to `ring[i].x` and the same for `[1] -> .y` so it would work with the new data layout.

The package is just 1 function. I could instead copy and paste it into `MapboxVectorTileImageryProvider.js` and comment that it is from `point-in-polygon` as we can still include the licence for that.

package.json
tephenavies, line 43: Yes

package.json
kring: Ok, thanks, that makes sense.  A fork is ok, we should just put it in TerriaJS.  But I think I agree with you; this thing is so simple and unlikely to change that just putting the one function right in our code doesn't really have any downsides that I can think of.

kring: Also, if all you need is something with an `x` and `y` property to represent a point, you can use Cesium's `Cartesian2` instead of pulling in `point-geometry`.

tephenavies: I'll move the `point-in-polygon` repo over.

`point-geometry` is a Mapbox module already used by `vector-tile`, and the output of some of the functions use the `Point` type from `point-geometry`.

kring: > point-geometry is a Mapbox module already used by vector-tile

Are you sure about that?  `npmgitdev list point-geometry` doesn't show it.

tephenavies: Actually I'll just copy the function in and attribute it to the writer.

kring: Oh yes it does, what was I looking at before?  Sorry.

tephenavies: From the [`vector-tile` readme](https://github.com/mapbox/vector-tile-js/blob/master/README.md)

#### Methods
- **loadGeometry()** &mdash; parses feature geometry and returns an array of
  [Point](https://github.com/mapbox/point-geometry) arrays (with each point having `x` and `y` properties)

kring: Yeah for some reason `npmgitdev list point-geometry` only shows the dependency directly from `terriajs`, not the one from `vector-tile`, despite the fact that `point-geometry` is clearly in `vector-tile`'s package.json.  So basically `npm list` doesn't work right.  Shocking, I know.

PR #1855, opened 2016-07-22
Lodash debounce

### LINKED ISSUE #1827: Use lodash's debounce
We have just added `lodash.debounce` in `ChartPanelDownloadButton`. There are other places that do manual debouncing, eg. SearchBox.jsx and MouseCoords.js. It may be worth using the lodash one now that we've taken that dependency.

See #1823.

chloeleichen:
This fixes https://github.com/TerriaJS/terriajs/issues/1827


lib/ReactViewModels/MouseCoords.js
```diff
@@ -133,20 +134,18 @@ export default class MouseCoords {
 
         const terrainProvider = globe.terrainProvider;
         if (!(terrainProvider instanceof EllipsoidTerrainProvider)) {
-            clearTimeout(this.accurateHeightTimer);
-            this.accurateHeightTimer = setTimeout(()=> {
-                this.sampleAccurateHeight(terrainProvider, position);
-            }, this.accurateSamplingDebounceTime);
+            if(defined(this.accurateHeightTimer)) this.accurateHeightTimer.cancel();
```
kring: I think this is just using debounce as a timer, rather than actually using it to debounce.  All our existing code to do the debounce is still in place.  By really using `debounce`, you should be able to remove our `debounceSampleAccurateHeight` function entirely.

lib/ReactViews/Search/SearchBox.jsx
```diff
@@ -29,6 +29,10 @@ export default React.createClass({
         };
     },
 
+    componentWillMount() {
+        this.debounced = debounce(this.search, 200);
```
kring: Should be `2000` not `200`.

lib/ReactViews/Search/SearchBox.jsx
```diff
@@ -38,13 +42,10 @@ export default React.createClass({
     },
 
     searchWithDebounce() {
```
kring: You should be able to remove this function entirely, and instead do `this.searchWithDebounce = debounce(this.search, 2000);` in `componentWillMount`.

chloeleichen:
@kring can you have another look please 😄

lib/ReactViewModels/MouseCoords.js
```diff
@@ -127,26 +134,12 @@ export default class MouseCoords {
...
-    }
 
     sampleAccurateHeight(terrainProvider, position) {
-        this.accurateHeightTimer = undefined;
         if (this.tileRequestInFlight) {
             // A tile request is already in flight, so reschedule for later.
-            this.accurateHeightTimer = setTimeout(()=> {
-                this.sampleAccurateHeight(terrainProvider, position);
-            }, this.accurateSamplingDebounceTime);
+            this.debounceSampleAccurateHeight.cancel();
```
RacingTadpole, line 140: This works, and looks good to me! However I haven't quite got my head around this part, so I'll leave it to @kring to review again.

lib/ReactViewModels/MouseCoords.js
```diff
@@ -75,9 +76,15 @@ export default class MouseCoords {
                 const maxHeightGeoid = maxHeight + (this.geoidModel ? this.geoidModel.maximumHeight : 0.0);
                 errorBar = Math.max(Math.abs(approximateHeight - minHeightGeoid), Math.abs(maxHeightGeoid - approximateHeight));
             }
+            Cartographic.clone(position, this.lastHeightSamplePosition);
+            const terrainProvider = globe.terrainProvider;
 
             this.cartographicToFields(intersection, errorBar);
-            this.debounceSampleAccurateHeight(globe, intersection);
+            if (!(terrainProvider instanceof EllipsoidTerrainProvider)) {
+                this.debounceSampleAccurateHeight.cancel();
```
kring: Is this cancel necessary?  I think calling the debounce function again will automatically reset the debounce timer.

kring:
It seems like the accurate height is never being sampled.  If you the mouse around on the map, it displays the elevation with a +/-, as expected.  But when you stop moving the mouse, it's supposed to display an exact height after a second or so.  But it never does.

chloeleichen:
Ready to be reviewed again

PR #1862, opened 2016-07-25
button layout for chart item

chloeleichen:
Fixes: https://github.com/TerriaJS/terriajs/issues/1861

screen shot 2016-07-25 at 11 36 55 am


kring:
Hmmm something is not quite right:
image

chloeleichen:
Ready for another look:

screen shot 2016-07-29 at 2 35 01 pm

PR #1873, opened 2016-07-26
Three fixes for WPS polygons

rdengate:
- AllowedValues for LiteralData WPS input now works even if only one value specified
- Fixed bug in WPS polygon datatype to return valid polygon geojson
- Fix regression: cursor changes in UserDrawing now functions in 2D as well as 3D


RacingTadpole:
What should I test this on? Thanks!

rdengate:
@RacingTadpole: I have tested all these changes manually, but you're welcome to test yourself if you like. Here's how:

AllowedValues for LiteralData WPS input now works even if only one value specified
- Since I fixed this the client has added in a few more values for their timeSeriesPoly WPS, but if you're super keen to test it, I can run a test WPS service that has a single AllowedValue input. You'll need Geoglam#newui to test this, and I can provide you with an extra bit in the config file to point to my test WPS.

Fixed bug in WPS polygon datatype to return valid polygon geojson
- Using Geoglam#newui, add data, choose timeSeriesPoly WPS, draw a polygon, and submit it. If it returns a chart, the polygon passed is valid geojson. Previously it was returning nothing. You can also check out the Execute request in the network developer tools to visually inspect the geojson that is passed to timeSeriesPoly.

Fix regression: cursor changes in UserDrawing now functions in 2D as well as 3D
- With Geoglam#newui, change to 3D, add data, choose timeSeriesPoly, and draw a polygon. Note that the cursor is changing to crosshair. Change to 2D and repeat. Without my change, the cursor doesn't change to crosshair in 2D.

lib/ReactViews/Analytics/PolygonParameterEditor.jsx
```diff
@@ -38,7 +38,7 @@ const PolygonParameterEditor = React.createClass({
     },
 
     getValue() {
-        const pointsLongLats = this.props.parameterValues[this.props.parameter.id];
+        const pointsLongLats = this.props.parameterValues[this.props.parameter.id][0];
```
RacingTadpole, line 41: I'm just digging through the code to understand what sort of a thing `this.props.parameterValues[key]` is.

It looks like:
`<MyDataTab>` → `<DataPreview previewed={this.props.viewState.userDataPreviewedItem}>` → `<InvokeFunction previewed={this.props.previewed}>` → `<ParameterEditor parameterValues={this._parameterValues}>`
where in the last step, 
`this._parameterValues[parameter.id] = parameter.defaultValue` (`parameter` is an element from `this.props.previewed.parameters` - [link here](https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Analytics/InvokeFunction.jsx#L65)).

Ie. `this.props.parameterValues[key]` is from `ViewState`'s `userDataPreviewedItem.parameters[x].defaultValue` where `userDataPreviewedItem`.

When I look at `WhyAmISpecialFunction.js` for example, it has a `parameters` array. However none of its elements has a `defaultValue` property...  what am I missing here?

Thanks!

RacingTadpole, line 41: I found a `defaultValue` in `PolygonParameter`. There it's listed in the doc string as:
`@param {Boolean} [options.defaultValue] The default value.`
Is that right? I suspect I've made a mistake in my code tracing somewhere - doesn't your new code expect an array whose first element is `pointsLongLats`?

rdengate, line 41: this.props.parameterValues[key] data type varies depending on which ParameterEditor has set it.

In InvokeFunction.jsx we have this._parameterValues (https://github.com/TerriaJS/terriajs/blob/rd_wps_fixes/lib/ReactViews/Analytics/InvokeFunction.jsx#L21), which in initializeParameters, goes through and sets any defaults defined for every parameter (https://github.com/TerriaJS/terriajs/blob/rd_wps_fixes/lib/ReactViews/Analytics/InvokeFunction.jsx#L65). So this._parameterValues might look something like:
{"layer" : "some_default_layer", "point": undefined}, where "layer" has come from a LiteralData parameter and "point" has come from a PointGeometry parameter.

this._parameterValues is passed to the ParameterEditor for every type of Parameter (https://github.com/TerriaJS/terriajs/blob/rd_wps_fixes/lib/ReactViews/Analytics/InvokeFunction.jsx#L90), so the this._parameterValues for each ParameterEditor (PolygonParameterEditor.jsx, say) contains a default for the parameters for that editor but also defaults for the other parameters.

It seems every *ParameterEditor.jsx is sticking in whatever data type it wants for parameterValues. That's okay, because each parameter editor is processing that value before it's returned to calling functions, with:
- *Parameter.js's formatValueAsString to display the value in the WPS dialog
- *ParameterEditor.jsx's getValue to retrieve a value which will be used to construct a WPS Execute request: https://github.com/TerriaJS/terriajs/blob/rd_wps_fixes/lib/Models/WebProcessingServiceCatalogFunction.js#L251

Does that make sense?

rdengate, line 41: Another piece of this puzzle is that the defaultValue for a parameter is not actually getting set, ever, in the code you're looking at. I've fixed this in https://github.com/TerriaJS/terriajs/pull/1910.

RacingTadpole, line 41: Ah! That explains it. However I don't think any of the `jsx` files are setting the defaultValue for any parameters - the only place (apart from your change in #1910) is in the `Models/*Parameter.js` files. And there, the jsdocs say it is a Boolean.  I would suggest changing the jsdoc info. :-)

RacingTadpole, line 41: Although I realise this isn't caused at all by your PR - sorry!  Perhaps I should really open this as a new issue.

rdengate, line 41: @RacingTadpole good observation. Yeah, if you could log that as a separate issue that would be great. Thanks!

PR #1874, opened 2016-07-26
Solve multi-y axis labels, x-label flying in from left

### LINKED ISSUE #1849: Get stacked Y axis even though chart series are in the same unit
If in AREMI you enable one of the layers in Electricity -> Generation -> Current Power Generation and then try to add two of the power stations timeseries to a chart, then even though they are both in MW the chart ends up showing stacked Y axis:

screen shot 2016-07-21 at 12 06 16

If you then remove the last one added and re-add it then you _do not_ get stacked axis - just the single one that you want!

I would have put a share URL in but it doesn't seem to happen when using a share URL.

RacingTadpole:
Fixes #1849.
Also stops the "Date" label flying in from the left when you first view a chart.

As you can see from the number of code changes, this turned out to be trickier than I expected. :-)


lib/Charts/LineChart.js
```diff
@@ -279,7 +285,7 @@ function render(element, state) {
             .style('text-anchor', 'middle')
             .text(state.axisLabel.x)
             .style('opacity', hasData ? 1 : 1e-6)
-            .transition().duration((isFirstLine || state.mini) ? 1 : transitionDuration)
+            // .transition().duration((isFirstLine || state.mini) ? 1 : transitionDuration)
```
AlexGilleran: Can this be deleted?

AlexGilleran:
Tests fine for me, just not sure whether that commented line was supposed to stay in there :)

RacingTadpole:
Cool, thanks. I've deleted that comment line now.

PR #1882, opened 2016-07-28
Add support for ArcGis FeatureServers

### LINKED ISSUE #993: Support ESRI FeatureServer
The URLs provided by ESRI OpenData sites ([eg](http://data.melbournewater.opendata.arcgis.com/datasets/b0181addb35840779802f66bf5d98a07_0?geometry=143.237%2C-38.122%2C147.099%2C-37.471)) look like this:

`https://services5.arcgis.com/ZSYwjtv8RKVhkXIL/arcgis/rest/services/HWV_Overview/FeatureServer/0/query?outFields=*&where=1%3D1`

But we can't paste that directly into TerriaJS. We should accept a wider range of ESRI URLs.

### LINKED ISSUE #1877: Support ArcGIS FeatureServer
As described here http://resources.arcgis.com/en/help/rest/apiref/index.html?featureserver.html

Eg. 
https://sampleserver6.arcgisonline.com/arcgis/rest/services
https://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire/FeatureServer
https://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire/FeatureServer/0

It looks like to get actual geometry data you need a URL like this:
https://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire/FeatureServer/query?f=json&layerDefs={0:"1=1"}

Note with MapServers, these are both ArcGisCatalogGroups:
- http://www.ga.gov.au/gis/rest/services/earth_observation/Landcover_WM/MapServer, which contains a single layer, land cover.
- http://www.ga.gov.au/gis/rest/services/, which contains a number of ArcGisCatalogGroups following the hierarchy below.

image

```
{
   "sharedCatalogMembers":{
      "Root+Group/User-Added+Data/":{
         "isOpen":true,
         "type":"esri-mapServer-group",
         "parents":[
            "Root+Group/User-Added+Data"
         ]
      },
      "Root+Group/User-Added+Data//earth+observation":{
         "isOpen":true,
         "type":"esri-mapServer-group",
         "parents":[
            "Root+Group/User-Added+Data",
            "Root+Group/User-Added+Data/"
         ]
      },
      "Root+Group/User-Added+Data//earth+observation/Landcover+WM":{
         "isOpen":true,
         "type":"esri-mapServer-group",
         "parents":[
            "Root+Group/User-Added+Data",
            "Root+Group/User-Added+Data/",
            "Root+Group/User-Added+Data//earth+observation"
         ]
      }, ...
   }
}
```

RacingTadpole:
Fixes #1877 and fixes #993.

Note `ArcGisCatalogGroup` was used in two different ways before: both as a group for MapServer layers, and as a group for rest services. (Eg. http://www.ga.gov.au/gis/rest/services/earth_observation/Landcover_WM/MapServer vs http://www.ga.gov.au/gis/rest/services/). These are now split into two separate group types: `ArcGisCatalogGroup` (with type `esri-group`) for the rest services and `ArcGisMapServerCatalogGroup` (with type `esri-mapServer-group`) for the MapServer layer group. I've then added `ArcGisFeatureServerCatalogGroup` and -`Item`.

This could potentially break pre-canned catalogs, as previously the types `esri-group` and `esri-mapServer-group` were interchangeable. For National Map it looks ok though.


RacingTadpole:
Note I've merged part of #1886 into this one now, so you can see the feature info of the feature server.

lib/Models/ArcGisFeatureServerCatalogGroup.js
```diff
@@ -0,0 +1,213 @@
...
+var inherit = require('../Core/inherit');
+var proxyCatalogItemUrl = require('./proxyCatalogItemUrl');
+var replaceUnderscores = require('../Core/replaceUnderscores');
+var ArcGisFeatureServerCatalogItem = require('./ArcGisFeatureServerCatalogItem');
+
+/**
+ * A {@link CatalogGroup} representing a collection of layers from an ArcGIS Feature Service.
+ * Eg. https://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire/FeatureServer
+ *
+ * @alias ArcGisFeatureServerCatalogGroup
```
AlexGilleran, line 25: why alias it to itself?

lib/Models/ArcGisFeatureServerCatalogGroup.js
```diff
@@ -0,0 +1,213 @@
...
+     */
+    typeName : {
+        get : function() {
+            return 'ArcGIS Feature Server Group';
+        }
+    },
+
+    /**
+     * Gets the set of functions used to serialize individual properties in {@link CatalogMember#serializeToJson}.
+     * When a property name on the model matches the name of a property in the serializers object lieral,
```
AlexGilleran, line 92: typo on `literal`

RacingTadpole, line 92: ha! I copied that directly from another CatalogItem. Looks like I'm not the only one - try searching for "lieral" in the /lib directory, there are 23 hits :-)

lib/Models/ArcGisFeatureServerCatalogGroup.js
RacingTadpole, line 25: I have no idea, but all the constructors do it...

AlexGilleran, line 25: Huh. Funny how you see it a thousand times but never notice it.

kring, line 25: It used to be required in order for jsdoc to work right.  But that was about a hundred years ago, I have no idea if it still is.

AlexGilleran:
Possible issues:
- The first URL there (http://www.ga.gov.au/gis/rest/services/earth_observation/Landcover_WM/MapServer) only loads if you have AutoDetect on, not if you specify ESRI MapServer as the type in the dropdown (this works on http://preview.nationalmap.terria.io/). Maybe we need another dropdown item?
- The only layer at that URL appears to have metadata but doesn't actually load onto the map (this works on http://preview.nationalmap.terria.io/)

RacingTadpole:
Nice catch. I've expanded that "Esri ArcGIS Server" to five different options now, so the user can really pick the one they want...

RacingTadpole:
I've updated this so it will work better with #1886.

RacingTadpole:
@AlexGilleran - over to you again?

Issue #1787, opened 2016-07-12
Templating is removed from feature info panel when layer is removed


axman6:
On http://newui.aremi.nicta.com.au/prototype/
1. turn on a later (say Elec. Inf -> Current Power -> All)
2. click on something to view its feature info
3. click “Remove All” in the workbench. The feature info panel removes the templated view of the feature info, but keeps the standard table based view of the data
   ![screen shot 2016-07-12 at 3 10 13 pm](https://cloud.githubusercontent.com/assets/39567/16755709/d2d1cbe0-4842-11e6-803f-bf18017004f9.png)
   ![screen shot 2016-07-12 at 3 10 31 pm](https://cloud.githubusercontent.com/assets/39567/16755713/d8852b2c-4842-11e6-93d3-c2e59885818e.png)

PR #1883, opened 2016-07-28
Show Feature Info even for WFS and ArcGis Feature Servers

RacingTadpole:
Previously, if the catalog item could not be identified for feature info, the feature was ignored. 
This was a mistake. It should be shown.

Also, if the title bar of the feature info section would be blank, revert to "Site Data".

### DIFF
--- a/lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
@@ -61,7 +61,6 @@ const FeatureInfoPanel = React.createClass({
         const {catalogItems, featureCatalogItemPairs} = getFeaturesGroupedByCatalogItems(this.props.terria);
 
         return catalogItems
-            .filter(catalogItem => defined(catalogItem))
             .map((catalogItem, i) => {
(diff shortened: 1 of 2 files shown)


RacingTadpole:
I traced where this `filter` came from - it was introduced in https://github.com/TerriaJS/terriajs/pull/1812/files .  I haven't seen any harm in removing it. @AlexGilleran do you think removing the `filter` will cause any problems?

AlexGilleran:
That was actually introduced in https://github.com/TerriaJS/terriajs/pull/1797. The idea is that it gets rid of features that have had their layer removed, so if we remove the filter then #1787 will reopen :/.

RacingTadpole:
Ah, thanks. Unfortunately `catalogItem` can be `undefined` there simply because it couldn't be found from the information available, in which case we do still want to show the feature. (Eg. Bike racks in #test).

In the old UI, if all items were removed, the feature info panel was left on the screen (which looked weird if a template was used, since the template disappeared with the catalog item). With #1797, removing all catalog items also removes the info from the feature info panel.

Clearly the new behaviour better is better.

It looks like we need a more robust method of passing a feature's catalog item through. Unfortunately at the point of picking, only the entity is known, not its catalog item...

PR #1885, opened 2016-07-28
Turn all feature `Entity`s into `Feature`s

RacingTadpole:
Any features which came through `DataSourceItem`s, eg. `geoJson`, did not actually get turned into `Feature`s.
This is important so that `FeatureInfoSection` can be track them properly if their current properties/descriptions change.

I haven't found a way to manually test my changes to `Terria.js` yet - any ideas?


AlexGilleran:
Code seems fine from reading it, but what does it actually do as an end result? Is this to cater for a GeoJSON layer that autorefreshes like the powerplants in AREMI?

RacingTadpole:
Well, the PR was motivated by consistency - `FeatureInfoSection` expects a `Feature` not an `Entity`, and in many cases it was getting an `Entity`. In particular it expects `'currentProperties', 'currentDescription', 'updateCounters'` properties on the feature, and for these to be tracked by knockout.

But yes, if the feature had time-varying properties or was being polled, then we'd need this change. Time-varying features certainly applies to CZML data sources.

I'm glad you asked because I now see that the CZMLs with time-varying description/properties do not update at
http://localhost:3003/#build/TerriaJS/test/init/czml.json (under master) - and this PR makes it even worse!  It has shifted the feature to the north (!?) and always returns "No results" in the feature info.

So hold off merging this, and I'll investigate. (I've raised a new issue, #1915).

RacingTadpole:
OK, this is done properly now. Fixes #1915.

PR #1886, opened 2016-07-28
Show Feature Info for more CatalogItems

RacingTadpole:
As a result of #1797, which was a fix for #1787, a number of catalog item types stopped showing their feature info (Gpx, Ogr, WebFeatureServiceCatalogItem, ArcGisFeatureServerCatalogItem, WebProcessingServiceCatalogItem).

This fixes that.

This is a better solution than #1883.

### DIFF
--- a/lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
@@ -61,7 +61,6 @@ const FeatureInfoPanel = React.createClass({
         const {catalogItems, featureCatalogItemPairs} = getFeaturesGroupedByCatalogItems(this.props.terria);
 
         return catalogItems
-            .filter(catalogItem => defined(catalogItem))
             .map((catalogItem, i) => {
                 // From the pairs, select only those with this catalog item, and pull the features out of the pair objects.
                 const features = featureCatalogItemPairs.filter(pair => pair.catalogItem === catalogItem).map(pair => pair.feature);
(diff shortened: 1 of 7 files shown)


lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
```diff
@@ -200,8 +204,9 @@ function determineCatalogItem(nowViewing, feature) {
         }
 
         for (i = nowViewing.items.length - 1; i >= 0; i--) {
-            if (nowViewing.items[i].dataSource === dataSource) {
-                result = nowViewing.items[i];
+            item = nowViewing.items[i];
+            if (item.dataSource === dataSource || (defined(item._geoJsonItem) && item._geoJsonItem.dataSource === dataSource)) {
```
rdengate: I'm confused about how this works. Isn't item.dataSource the same as item._geoJsonItem.dataSource? It seems like in https://github.com/TerriaJS/terriajs/blob/master/lib/Models/WebProcessingServiceCatalogItem.js#L97 that's what it's returning...

RacingTadpole: That's a great pickup - actually the WPS is special, it's the only catalog item that maps it across - but you're right, I think a much better way to effect this would be to expose `_geoJsonItem.dataSource` as the `dataSource` for all the other ones too. I was feeling uncomfortable referring to an underscored property outside the class!

rdengate: Aha, that makes sense. I agree, might be good to do that. Other than that I think this looks fine :+1:

RacingTadpole:
Thanks, all done now.

PR #1889, opened 2016-07-29
d3 v4

chloeleichen:
@RacingTadpole 😄


lib/Charts/Tooltip.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
-import d3 from 'd3';
+import * as d3Collection from 'd3-collection';
+import * as d3Selection from 'd3-selection';
+import * as d3Transition from 'd3-transition'; // eslint-disable-line no-unused-vars
```
RacingTadpole: I think it's better to be explicit about what we're importing - instead of `import *`, can you use eg.
`import {selection as d3Selection} from 'd3-selection'`?
(This applies to all the files here.)

lib/Charts/LineChart.js
```diff
@@ -1,6 +1,10 @@
 'use strict';
 
-import d3 from 'd3';
+// import d3 from 'd3';
```
RacingTadpole: I'd say just delete the old line, instead of commenting it :-)

lib/Charts/LineChart.js
```diff
@@ -118,56 +121,66 @@ function render(element, state) {
...
         .attr('height', state.height);
 
-    const g = d3.select(element).selectAll('.line-chart');
+    const g = d3Selection.select(element).selectAll('.line-chart');
+
+    const t = d3Transition.transition().duration(transitionDuration);
 
     const lines = g.selectAll('.line').data(data, d => d.id).attr('class', 'line');
-    const isFirstLine = (!defined(lines[0][0]));
+    // NOTE
```
RacingTadpole: Note what?

lib/Charts/LineChart.js
```diff
@@ -214,7 +228,7 @@ function render(element, state) {
             .style('text-anchor', 'end');
     yAxisElements.exit().remove();
 
-    // No data. Show message if no data to show.
+    // // No data. Show message if no data to show.
```
RacingTadpole: Super minor... but you've double-commented this line!

lib/Charts/LineChart.js
```diff
@@ -224,24 +238,22 @@ function render(element, state) {
         .attr('x', element.offsetWidth / 2 - margin.left - size.yAxesWidth)
         .attr('y', (size.height - 24) / 2);
 
-    // Axes.
+    // // Axes.
```
RacingTadpole: And this one

lib/Charts/LineChart.js
```diff
@@ -260,18 +272,18 @@ function render(element, state) {
     }
 
     g.select('.x.axis')
-        .transition('1st').duration(transitionDuration)
+        .transition(t)
```
RacingTadpole, line 272: What does it mean to use the same transition `t` here as elsewhere, especially since this one had a specific name before?

lib/Charts/LineChart.js
```diff
@@ -260,18 +272,18 @@ function render(element, state) {
...
         .attr('transform', 'translate(0,' + y0 + ')')
         .style('opacity', 1)
         .call(xAxis);
     if (defined(state.grid) && state.grid.x) {
         // Recall the x-axis-grid lines only extended up; we need to translate the ticks down to the bottom of the plot.
         g.selectAll('.x.axis line').attr('transform', 'translate(0,' + (size.plotHeight - y0) + ')');
     }
-    // If mini or no data, hide the ticks, but not the axis, so the x-axis label can still be shown.
+    // // If mini or no data, hide the ticks, but not the axis, so the x-axis label can still be shown.
```
RacingTadpole: Double commenting trick again :-)

lib/Charts/LineChart.js
```diff
@@ -260,18 +272,18 @@ function render(element, state) {
...
     g.select('.x.axis')
         .selectAll('.tick')
-        .transition('2nd').duration(transitionDuration)
+        .transition(t)
```
RacingTadpole, line 283: Same `t` question...

lib/Charts/LineChart.js
```diff
@@ -287,16 +299,18 @@ function render(element, state) {
...
-        .transition('1st').duration(transitionDuration)
+    const yAxisElementsElements = g.select('.y.axes').selectAll('.y.axis');
+
+    yAxisElementsElements
+        .transition(t)
         .attr('transform', (d, i) => ('translate(' + (- i * Size.yAxisWidth) + ', 0)'));
 
-    yAxisElements[0].forEach(yAxisElement => {
-        const yAxisD3 = d3.select(yAxisElement);
+    yAxisElementsElements.nodes().forEach(yAxisElement => {
```
RacingTadpole, line 305: Thanks for cleaning up all my `[0]` stuff with `.nodes()` - that looks much more professional!

lib/Charts/Title.js
```diff
@@ -70,20 +73,20 @@ const Title = {
...
                 addedTitleComponents.append('span')
                     .attr('class', 'color');
             }
             addedTitleComponents.append('span')
                 .attr('class', 'name');
             // Enter and update.
-            titleComponents.select('.color').style('background-color', d=>d.color);
-            titleComponents.select('.name').text(getName);
+            addedTitleComponents.select('.color').style('background-color', d=>d.color);
+            addedTitleComponents.select('.name').text(getName);
```
RacingTadpole, line 85: I can't remember why I had `titleComponents` here instead of `addedTitleComponents` - what made you switch?

RacingTadpole:
This looks great, thanks Chloe - just the minor nitpicks above! 👍

lib/Charts/Title.js
chloeleichen, line 85: `selection` in v4 are immutable,  so after `selection.data()` , the `addedTitleComponents` is the new selection returned, so trying to modify `titleComponents` at the first render does not actually modify the attributes of latest selection: https://github.com/d3/d3/blob/master/CHANGES.md#selections-d3-selection

RacingTadpole:
When I bring up sheep stations (for example) in the chart panel, and then choose "power", the existing orange chart line disappears in the transition, instead of animating smoothly.
image

lib/Charts/Tooltip.js
chloeleichen: We want to import the entire module here tho?

RacingTadpole: I think he's got it set up so that [`selection`](https://github.com/d3/d3-selection/blob/master/index.js) has everything under it, so my example replacement would work. 
If not, we're only using `event` and `select`, so you could also try
`import {event as d3Event, select as d3Select} from 'd3-selection'`, and then use `d3Select` instead of `d3Selection.select`.

lib/Charts/LineChart.js
chloeleichen, line 272: I assume you used named transition to do concurrent transition, which no longer works this way in d4. The selection.transition method now takes an optional transition instance which can be used to synchronize a new transition with an existing transition: https://github.com/d3/d3/blob/master/CHANGES.md#transitions-d3-transition

chloeleichen:
Fixed, @RacingTadpole ready for another look

lib/Charts/LineChart.js
```diff
@@ -287,16 +296,18 @@ function render(element, state) {
     if (state.mini) {
```
RacingTadpole, line 296: I realise the transition on line 291 may have a different duration - but this syntax looks the same as in v3 - should you use the `d3Transition()` syntax here too?

RacingTadpole:
This works really nicely, I think maybe even better than before - thanks! But I can't see any changes in `3425e56` that explain why the transitions now work, where this morning they didn't - what was the trick?

lib/Charts/LineChart.js
chloeleichen, line 296: The reason I kept it is that I saw in this PR this line has been removed: 💃 https://github.com/TerriaJS/terriajs/pull/1874/files

RacingTadpole, line 296: you rock!

chloeleichen:
the problem was because originally I did `lines.merge(lines)` which updated the whole selection and did transition opacity from 0 to 1 on every updates. I removed it in the updated code so it doesn't do that anymore.

Issue #1868, opened 2016-07-26
Catalog UI does not show selected group after it's closed and re-opened


kring:
When you click a group in the catalog UI, the group is highlighted in blue and the name of the group, plus its description and other info, is displayed in the preview area.  If you then close the explorer window and re-open it, the group is still displayed in the preview panel but is no loner highlighted in blue on the left.

This came up as a minor issue in GA's testing of the new NationalMap release.

RacingTadpole:
I'm just reviewing #1902 and I'm not sure I understand the problem. For me, master shows this, even after closing and re-opening the "Add Data" panel:
![image](https://cloud.githubusercontent.com/assets/1757858/17286337/4975937a-580c-11e6-9e62-326ab410e859.png)
Is that the highlighting in blue you mean?

kring:
Yeah it works fine when an item is selected, but not when a group is selected.

PR #1902, opened 2016-08-01
catalog btn blue on open

chloeleichen:
Fixes https://github.com/TerriaJS/terriajs/issues/1868

### DIFF
--- a/lib/ReactViews/DataCatalog/CatalogGroup.jsx
+++ b/lib/ReactViews/DataCatalog/CatalogGroup.jsx
@@ -18,7 +18,8 @@ function CatalogGroup(props) {
                     className={classNames(
                             Styles.btnCatalog,
                             {[Styles.btnCatalogTopLevel]: props.topLevel},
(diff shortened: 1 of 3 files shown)


RacingTadpole:
I'm just clarifying the issue over at #1868. But with this PR I can get a lot of highlighted blue bars in the navigation pane, which looks a bit confusing. Eg.
image

chloeleichen:
yeah because they are all 'open', do we want to high light the top level that is open? 

Previously the highlight happens when the button is focused, that is why only one group can be highlighted at all time. 

Maybe it's a question for @philipgrimmett

kring:
Groups should only be blue when they're focused/selected.  But the catalog should remember which group was focused/selected when you close and reopen the catalog UI.  It currently does that for catalog _items_ but not for catalog _groups_.  If that's unclear, stop by and I'll show you what I mean.

chloeleichen:
ah I see, let me give it another go

chloeleichen:
Ready for another look please! @RacingTadpole

PR #1905, opened 2016-08-01
More general SDMX-JSON reader

### LINKED ISSUE #1514: Implement SDMX-JSON catalog item

RacingTadpole:
Files such as `wwwroot/test/SDMX-JSON/data-as-obs-repeated-dim.json` break the previous implementation of `SdmxJsonCatalogItem`.

There are two issues:
- The details of where the data is stored are completely different, eg. all the structure info is in `structure.dimensions.observation` not `structure.dimensions.series`, and the data is in `dataSets[0].observations[key]`, not `dataSets[0].series[key].observations[timeIndex][0]`.
- It has two identical dimensions, State and Region, with data only for `0:0:N:0:M:0:0` for `N=M`.

The first problem was solved by using @airosa's https://github.com/airosa/sdmxjsonlib under the hood.
The second problem was solved by adding a new `aggregatedDimensionIds` parameter, which can be used more generally to sum over a dimension's values and suppress it from the UI.

Try it out with http://localhost:3003/#build/terriajs/test/init/sdmx.json.

Fixes #1514.


AlexGilleran:
(haven't read the code yet)

I've noticed with the non-spatial graph one, adding more data (e.g. deaths and births instead of just births) results in the chart line changing colour but changing to the sum of both instead of graphing them against each other as with other graphable datasets. Is that unavoidable? If so should the line change colour? Seems a bit weird.

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -590,53 +601,84 @@ function buildRegionAndTimeColumns(item, dimensions) {
...
+        if (!defined(x)) {
+            return y;
         }
-    }
-    return result;
+        if (!defined(y)) {
+            return x;
+        }
+        return x + y;
+    }, 0);
```
AlexGilleran: This might be neater as `array.filter(defined).reduce((x, y) => x + y, 0)` but it's a nitpick.

RacingTadpole:
Yeah, you're right. The interface for non-spatial SDMXs isn't thought through at all yet. It doesn't crash, but that's about all I can say for it. In fact it's still showing the total of the selected columns (just like for region-mapped SDMXs), but with weird color choices.
Since no one is using SDMX yet, I figured I'd leave this be, and fix it when a use-case arises.

lib/Models/SdmxJsonCatalogItem.js
RacingTadpole: nice! although there's one slight difference in behaviour which is important - my code sums [undefined, undefined] as undefined, yours as 0...

RacingTadpole: Actually, on rereading my code, I'm not sure I'm right...?

AlexGilleran: Haha on reading your code the first time I thought it summed to zero but actually it does sum to `undefined`

AlexGilleran: UPDATE: On third thought I'm pretty sure it sums to 0 because it always takes the one that isn't undefined so it'll keep taking 0 the whole time.

AlexGilleran: However it does return undefined if I actually run it. I don't know what to think anymore.

AlexGilleran: FINAL EDIT: Actually it only worked because I let `defined = x => !!x` in my test, with `defined` properly... defined it does actually return 0 for `[undefined, undefined]`.

I'm not sure it really matters but it was a fun learning experience.

RacingTadpole:
OK, nice catch on the undefined stuff there - all fixed and back to you!

PR #1910, opened 2016-08-02
Enumeration parameter default

### LINKED ISSUE #1881: EnumerationParameter default option in dropdown
Currently an enumeration parameter in WPS presents as a dropdown, with the first allowed value shown in the UI. However, if the user submits this, no value is passed because they didn't change the dropdown.

Add a blank or "<Not specified>" to the drop down for optional enumeration parameters, or if one is required, default to the one shown in the UI.

rdengate:
Enumeration parameter now defaults to what is shown in UI, and if parameter is optional, '' is default. Fixes #1881.

This depends on https://github.com/TerriaJS/terriajs/pull/1873.

Most *Parameter types don't set defaultValue. DateTimeParameter tried, but it wasn't passed in via options, so it still would have been undefined. Even if it had been used in DateTimeParameter, it wouldn't have worked because in invokeFunction, the part where it's getting defaulted doesn't have any parameters to iterate over.

componentWillMount (where parameters are initialised and defaulted) used to get called with 0 parameters, then render() would get called with 0 parameters. Then render() would get called again, this time with 3 parameters, but they haven't been initialised because componentWillMount will only be called before the first render().


lib/ReactViews/Analytics/PolygonParameterEditor.jsx
```diff
@@ -79,7 +79,7 @@ const PolygonParameterEditor = React.createClass({
             points.push(CesiumMath.toDegrees(cartographic.latitude));
             pointsLongLats.push(points);
         }
-        this.props.parameterValues[this.props.parameter.id] = pointsLongLats;
+        this.props.parameterValues[this.props.parameter.id] = [pointsLongLats];
```
RacingTadpole, line 82: You can't change the props!

rdengate, line 82: But it seems all *ParameterEditor.jsx classes do this? E.g. https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Analytics/DateTimeParameterEditor.jsx#L43
https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Analytics/PointParameterEditor.jsx#L49
https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Analytics/RegionParameterEditor.jsx#L97

I wrote PolygonParameterEditor based on PointParameterEditor. If setting props is wrong (and it seems like it is) then I think we will have to change this in all *ParameterEditor.jsx classes?

RacingTadpole, line 82: Interesting! Alright, let's write that up as a separate issue too... would you be up for doing that?

kring, line 82: My two cents on changing props:
- Actually changing a value in props is illegal, I think React will actually stop you from doing that.
- Changing the value of something inside props (e.g. a model-ish object like `parameterValues`) inside `render` or something called by `render` is bad because `render` should idempotent.
- Changing a value of something inside props outside of render is fine, we do that all the time!
- It _may_ be ok to change something inside props even from within a render if the "change" isn't so much a change as it is setting an initial value, e.g. initialize on first access.  Ideally the model layer would do this sort of thing instead of the UI, though.

lib/ReactViews/Analytics/InvokeFunction.jsx
```diff
@@ -32,6 +32,10 @@ const InvokeFunction = React.createClass({
         this.initializeParameters(nextProps);
     },
 
+    componentWillUpdate(nextProps, nextState) {
+        this.initializeParameters(nextProps);
+    },
+
```
RacingTadpole, line 43: Out of interest - what's driving this change?

RacingTadpole, line 43: It feels a little funny to "initialize" the parameters on an "update".

lib/ReactViews/Analytics/PolygonParameterEditor.jsx
RacingTadpole, line 82: Yes, that sounds good. And here we are changing it in an object in response to a click, not in render, so it should be ok.
It still feels a little mucky to me... I think I'd prefer to have a function you can call on the model itself to update its parameter values. But admittedly there's not much difference between setting it directly and calling a function to set it.
I'm happy not to raise this as an issue.

lib/ReactViews/Analytics/InvokeFunction.jsx
rdengate, line 43: So, previously initializeParameters was only called in componentWillMount. But this got called the first time with 0 parameters, then render() would get called with 0 parameters. Then render() would get called again, this time with 3 parameters, but they would not have been initialised because componentWillMount is only called before the first render(). I was looking for a way to call initalizeParameters before render() that triggered the first time. I agree this doesn't sound quite right (although it does work). Do you have a different idea of how I could fix this?

RacingTadpole, line 43: Hmm... InvokeFunction does some interesting contortions, doesn't it? 
Eg. It seems strange to `knockout.track` the same thing in both `componentWillMount` and `initializeParameters`. And that `initializeParameters` also "enable[s] the context item, if any" - can that be moved somewhere more appropriate?
But regarding this PR, I'm worried that the `initializeParameters` will overwrite desirable parameters with the default values in some situation we haven't come across yet.
Perhaps you need a function "updateParameters", which updates the parameters using the default values just as - er - default values?

RacingTadpole, line 43: Also, I haven't seen `knockout` tracking on a view variable before... I guess there's no reason we can't do it. But since that view variable is just a reworking of a variable defined on a model, wouldn't it be cleaner to track the model variable instead? Maybe this is a question for @kring ... :-)

rdengate, line 43: - It seems strange to knockout.track the same thing in both componentWillMount and initializeParameters

That does seem odd. I am going to remove the knockout.track in initializeParameters and see what happens.
- initializeParameters also "enable[s] the context item, if any" - can that be moved somewhere more appropriate?

I'll have a look at moving this.
- I'm worried that the initializeParameters will overwrite desirable parameters with the default values in some situation we haven't come across yet.

Yeah, fair enough. I will see if I can fix this.

Thanks @RacingTadpole!

rdengate:
Ok, hopefully that's an improvement. I'm happy to make more changes if required :-)

lib/ReactViews/Analytics/InvokeFunction.jsx
```diff
@@ -19,21 +19,26 @@ const InvokeFunction = React.createClass({
...
     },
 
     componentWillReceiveProps(nextProps) {
-        this.initializeParameters(nextProps);
+        // This will be called when component is already mounted but props change, for example if you go from one WPS
+        // item to another. In that instance, we'll need to clear parameter values.
+        this._parameterValues = {};
+        this.defaultPropsIfNeeded(nextProps);
+        this.enableContextItem(nextProps);
     },
```
RacingTadpole, line 36: this looks good to me! thanks. only question is - do you need to call `enableContextItem` in both places?

rdengate, line 36: By calling it in both places, it's behaving as it did before I made any changes. I'm not sure what it's doing though. Do you know?

RacingTadpole, line 36: Well... I guess it does no harm. But if we don't expect `props.previewed.contextItem` to ever change, and it's present on mounting, then you wouldn't need it on receiving props as well. But I'm happy to let it go. :-)

PR #1912, opened 2016-08-02
Overzoom geometry in vector tiles

tephenavies:
Waiting on/incorporates #1854.

Allows the server to serve only pre-generated tiles to a certain zoom level. The client can then use these tiles for higher zoom levels by expanding the geometry (since they are vector tiles). Fixes TerriaJS/vector-tile-server#1.

~~Test at the nationalmap instance http://crisp.terria.io~~
Test using nationalmap branch `mvt_first_release`


tephenavies:
This PR is ready to merge. Once merged, ~~I can terminate http://crisp.terria.io and~~ finalise release methods for vector tile servers. This will make vector tiles ready to be released in nationalmap by the end of this week.

tephenavies:
I have terminated http://crisp.terria.io. The production vector tile server is now stable and located at http://vector-tiles.terria.io (Note: this is a vector tile endpoint only, there is no nationalmap deployment or web interface on the server).

kring:
This looks great!  The only problem I see is that the vector tile region mapping doesn't work in IE9, but that probably wasn't introduced in this PR.  I'll write a separate issue.

PR #1924, opened 2016-08-05
Add support for moving point CSV files

RacingTadpole:
Fixes #1275. Also fixes #1916.

This also lays the groundwork for #1921 (hence the strange branch name), #1922 and eventually #1424 and #1474.

Check it out using http://localhost:3003/#build/terriajs/test/init/csv.json and the "Moving features" datasets. You'll see there are new `idColumns` and `isSampled` properties.

PR #1927, opened 2016-08-08
Implement polling using `idColumns` and allowing updates, not just replacements

### LINKED ISSUE #1922: Implement `polling.idColumns` and `polling.replace=false`
These two loose ends remain from #1756.

Currently:
- the polled csv [completely replaces the original](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CsvCatalogItem.js#L626) (it would be nice to have the option of it updating only), and
- features are updated assuming they appear in [exactly the same order](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/TableDataSource.js#L271) as before (for lat/lon csvs).

I set up the parameters `idColumns` and `replace` to implement these, but never did.

For bonus points we could also implement polling on czml and other catalog items.

RacingTadpole:
Fixes #1922.  Also makes polling work for moving features csv files (#1924). Merge after #1924.

To try it out, go to http://localhost:3003/#build/terriajs/test/init/polling.json, and try the "Moving features updated/replaced in 20s" datasets.

### DIFF
--- a/lib/Map/TableStructure.js
+++ b/lib/Map/TableStructure.js
@@ -533,10 +533,13 @@ TableStructure.prototype.append = function(table2, rowNumbers) {
         }
         updatedColumnValueses.push(this.columns[columnNumber].values.concat(valuesToAdd));
     }
-    this.columns = this.columns.reduce((updatedColumns, column, columnNumber) => {
-        updatedColumns.push(new TableColumn(column.name, updatedColumnValueses[columnNumber], column.options));
+    var updatedColumns = this.columns.reduce((updatedColumns, column, columnNumber) => {
+        updatedColumns.push(new TableColumn(column.name, updatedColumnValueses[columnNumber], column.getFullOptions()));
         return updatedColumns;
     }, []);
+    // activeTimeColumn is not tracked, but columns are, so change activeTimeColumn first.
+    this.activeTimeColumn = getColumnWithSameId(this.activeTimeColumn, updatedColumns);
+    this.columns = updatedColumns;
 };
 
 /**
@@ -556,11 +559,22 @@ TableStructure.prototype.replaceRows = function(table2, replacementMap) {
             }
         }
     }
-    this.columns = this.columns.reduce((updatedColumns, column, columnNumber) => {
-        updatedColumns.push(new TableColumn(column.name, updatedColumnValueses[columnNumber], column.options));
+    var updatedColumns = this.columns.reduce((updatedColumns, column, columnNumber) => {
+        updatedColumns.push(new TableColumn(column.name, updatedColumnValueses[columnNumber], column.getFullOptions()));
         return updatedColumns;
     }, []);
+    this.columns = updatedColumns;
 };
+
+function getColumnWithSameId(column1, columns) {
+    if (defined(column1)) {
+        var matchingColumns = columns.filter(column => column.id === column1.id);
+        if (matchingColumns.length !== 1) {
+            throw new DeveloperError('Ambiguous column: ' + column1.name);
+        }
+        return matchingColumns[0];
+    }
+}
 
 /**
  * Merges table2 into this table.
@@ -579,14 +593,7 @@ TableStructure.prototype.merge = function(table2) {
     var table2RowNumbersMap = table2.getIdMapping(this.idColumnNames);
     var rowsFromTable2ToAppend = [];  // An array of row numbers.
     var rowsToReplace = {};  // Properties are {table 1 row number: table 2 row number}.
-    var table2ActiveTimeColumn;
-    if (defined(this.activeTimeColumn)) {
-        var table2ActiveTimeColumns = table2.columns.filter(column => column.id === this.activeTimeColumn.id);
-        if (table2ActiveTimeColumns.length !== 1) {
-            throw new DeveloperError('Cannot merge tables with ambiguous time column.');
-        }
-        table2ActiveTimeColumn = table2ActiveTimeColumns[0];
-    }
+    var table2ActiveTimeColumn = getColumnWithSameId(this.activeTimeColumn, table2.columns);
     for (var key in table2RowNumbersMap) {
         if (table2RowNumbersMap.hasOwnProperty(key)) {
             var table2RowNumbers = table2RowNumbersMap[key];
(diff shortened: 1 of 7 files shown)


lib/Map/TableStructure.js
```diff
@@ -503,6 +514,118 @@ TableStructure.prototype.getIdMapping = function() {
...
+ * row numbers from table2.
+ * Changes all the columns in one go, to avoid partial updates from tracked values.
+ * @param {TableStructure} table2 The table to add to this one.
+ * @param {Integer[]} [rowNumbers] The row numbers from table2 to add (defaults to all).
+ */
+TableStructure.prototype.append = function(table2, rowNumbers) {
+    if (this.columns.length !== table2.columns.length) {
+        throw new DeveloperError('Cannot add tables with different numbers of columns.');
+    }
+    var updatedColumnValueses = [];
```
AlexGilleran: Valueses?

image

RacingTadpole: too golumy? It's an array of values arrays... `valueses` seemed to capture it better than `valuesArray`, which could be confused with `values` itself. But happy to change it if you think it looks too weird :-)

AlexGilleran: `updatedColumnValuesArrays` maybe? When I read 'valueses' I tend to apply gollum-logic and assume it's the same as 'values' ;)

lib/Map/TableStructure.js
```diff
@@ -503,6 +514,118 @@ TableStructure.prototype.getIdMapping = function() {
...
+    for (var columnNumber = 0; columnNumber < table2.columns.length; columnNumber++) {
+        var valuesToAdd = table2.columns[columnNumber].values;
+        if (defined(rowNumbers)) {
+            valuesToAdd = valuesToAdd.filter((_, rowNumber) => rowNumbers.indexOf(rowNumber) >= 0);
```
AlexGilleran: If there's potential for high numbers of rows could we make rowNumbers a lookup table so this isn't quadratic-time?

lib/Map/TableStructure.js
```diff
@@ -503,6 +514,118 @@ TableStructure.prototype.getIdMapping = function() {
...
+        updatedColumnValueses.push(this.columns[columnNumber].values);
+        for (var table1RowNumber in replacementMap) {
+            if (replacementMap.hasOwnProperty(table1RowNumber)) {
+                var table2RowNumber = replacementMap[table1RowNumber];
+                updatedColumnValueses[columnNumber][table1RowNumber] = table2.columns[columnNumber].values[table2RowNumber];
+            }
+        }
+    }
+    var updatedColumns = this.columns.reduce((updatedColumns, column, columnNumber) => {
+        updatedColumns.push(new TableColumn(column.name, updatedColumnValueses[columnNumber], column.getFullOptions()));
```
AlexGilleran: is this actually a `map`?

RacingTadpole: yeah! I don't know what planet I was on there.

lib/Map/TableStructure.js
RacingTadpole: sounds good

lib/Map/TableStructure.js
RacingTadpole: maybe... it's O(n(cols) \* n(rows)), so not as bad as quadratic time though.
Do you think doing a `rowNumbers.reduce` to build up a new array would be better than `valuesToAdd.filter`? Unless there's only one or two features, `rowNumbers` should be (much?) smaller than `valuesToAdd`.

AlexGilleran: Right so I remember what I was thinking here - using `filter` is fine, it's `indexOf` that I'm worried about. Every value in `valuesToAdd` will result in a linear search over `rowNumbers`, whereas if you turned `rowNumbers` into a lookup table at the start your filter function could just be `defined(rowNumbersLookup[rowNumber])`

lib/Map/TableStructure.js
```diff
@@ -503,6 +514,121 @@ TableStructure.prototype.getIdMapping = function() {
...
+        throw new DeveloperError('Cannot add tables with different numbers of columns.');
+    }
+    var updatedColumnValuesArrays = [];
+    function mapRowNumberToValue(valuesToAdd) {
+        return rowNumber => valuesToAdd[rowNumber];
+    }
+    for (var columnNumber = 0; columnNumber < table2.columns.length; columnNumber++) {
+        var valuesToAdd = table2.columns[columnNumber].values;
+        if (defined(rowNumbers)) {
+            valuesToAdd = rowNumbers.map(mapRowNumberToValue(valuesToAdd));
```
AlexGilleran: that works too :). Although it means line 533 should probably be in an else block.

RacingTadpole: fair enough, done. :-)

lib/Map/TableStructure.js
```diff
@@ -503,6 +514,123 @@ TableStructure.prototype.getIdMapping = function() {
...
+    }
+    if (this.columns.length !== table2.columns.length) {
+        throw new DeveloperError('Cannot merge tables with different numbers of columns.');
+    }
+    var table1RowNumbersMap = this.getIdMapping();
+    var table2RowNumbersMap = table2.getIdMapping(this.idColumnNames);
+    var rowsFromTable2ToAppend = [];  // An array of row numbers.
+    var rowsToReplace = {};  // Properties are {table 1 row number: table 2 row number}.
+    var table2ActiveTimeColumn = getColumnWithSameId(this.activeTimeColumn, table2.columns);
+    for (var key in table2RowNumbersMap) {
```
AlexGilleran: I'm having trouble following this loop - could you maybe rename `key` to `table2Id`? That's what it is right?

lib/Map/TableStructure.js
```diff
@@ -503,6 +514,123 @@ TableStructure.prototype.getIdMapping = function() {
...
+        if (table2RowNumbersMap.hasOwnProperty(key)) {
+            var table2RowNumbers = table2RowNumbersMap[key];
```
AlexGilleran: and this to something like `table2RowNumbersWithThisId` ?

lib/Map/TableStructure.js
RacingTadpole: actually the keys of `table2RowNumbersMap` are feature ids (but not `feature.id` - they are the concatenation of the `idColumns` for that feature) ... so I'll rename `key` to `featureIdString`

AlexGilleran: Sounds good 👍

lib/Map/TableStructure.js
RacingTadpole: thanks, that looks nicer now.

PR #1929, opened 2016-08-09
Restore behaviour of esri-group and esri-mapServer-group

RacingTadpole:
PR #1882 introduced breaking changes to the catalog:

> Catalog types `esri-group` or `esri-mapServer-group` were previously synonyms. Now `esri-group` must be used for REST services (eg. URLs ending in `gis/rest/services`), while `esri-mapServer-group` must be used for MapServers (eg. URLs ending in /MapServer).

This PR makes the two types interchangeable again, while keeping the new FeatureServer functionality.

It does it by making `ArcGisCatalogGroup` function load just like an `ArcGisMapServerCatalogGroup` if it sees `MapServer` in the URL.

It also extends the capability of `ArcGisCatalogGroup` (and hence `esri-group`) so that it can handle FeatureServers too.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,10 +2,9 @@
 Change Log
 ==========
 
-### 5.0.0
+### 4.2.0
 
-* Breaking change: Catalog types `esri-group` or `esri-mapServer-group` were previously synonyms. Now `esri-group` must be used for REST services (eg. URLs ending in `gis/rest/services`), while `esri-mapServer-group` must be used for MapServers (eg. URLs ending in /MapServer).
-* Added support for ArcGis FeatureServers.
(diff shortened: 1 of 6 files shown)


RacingTadpole:
Here are some ways to test this:

Does auto-detecting type of added URLs work? Eg. Try it with:
- https://services5.arcgis.com/ZSYwjtv8RKVhkXIL/arcgis/rest/services
- https://services5.arcgis.com/ZSYwjtv8RKVhkXIL/arcgis/rest/services/HWV_Overview/FeatureServer
- http://www.asris.csiro.au/arcgis/rest/services/TERN/SRTM_attributes_3s_ACLEP_AU/MapServer

The above should also work when you choose the type as "Esri ArcGis Server".

Does the existing catalog still work? Try searching data for `is:esri-mapServer-group` and `is:esri-group` (the Sunshine Coast Council is the only example of the latter) and checking they still work.

PR #1931, opened 2016-08-10
Add a time series chart to the Feature Info Panel for moving features

### LINKED ISSUE #1921: Show a chart of time-varying feature data in the Feature Info panel
We now have charts which show data associated with a point or region over time, eg. power generation. 

When a table of time-series and geographic data is available, eg. if we had a table with columns:

`date, featureId, lat, lon, value`

It would be great if the Feature Info panel also showed a chart of the time series at that point.

This would be a useful building block in implementing #1424 and #1474, for example.

It should also work for region-mapped tables.

RacingTadpole:
Fixes #1921.  Merge after #1927.

Test it out with http://localhost:3003/#build/terriajs/test/init/csv.json - see the three sampled Moving Features datasets, which show how to use it with and without feature info templates.

### DIFF
--- a/lib/Map/TableStructure.js
+++ b/lib/Map/TableStructure.js
@@ -279,33 +279,37 @@ TableStructure.prototype.toArrayOfColumns = function() {
 * Return data as an array of rows of formatted data, eg. [ ['x', 'y'], ['1', '12,345'], ['2.1', '20'] ].
 * @param {String} [dateFormatString] If present, override the standard date format with a string (see https://www.npmjs.com/package/dateformat)
 *                 Eg. "isoDateTime" or "dd mmm yyyy HH:MM:ss"
+* @param {Integer[]} [rowNumbers] An array of row numbers to return. Defaults to all rows.
 * @returns {Object} An array of rows of formatted data, the first of which is the column names.
 */
-TableStructure.prototype.toArrayOfRows = function(dateFormatString) {
+TableStructure.prototype.toArrayOfRows = function(dateFormatString, rowNumbers) {
     if (this.columns.length < 1) {
         return;
     }
     var that = this;
     var rows = that.columns[0].values.map(function(value0, rowIndex) {
-        return that.columns.map(function(column) {
-            if (dateFormatString && column.type === VarType.TIME) {
-                return dateFormat(column.dates[rowIndex], dateFormatString);
-            }
-            return column._formattedValues[rowIndex];
-        });
-    });
+        if (!defined(rowNumbers) || rowNumbers.indexOf(rowIndex) >= 0) {
+            return that.columns.map(function(column) {
+                if (dateFormatString && column.type === VarType.TIME) {
+                    return dateFormat(column.dates[rowIndex], dateFormatString);
+                }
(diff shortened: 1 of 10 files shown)


lib/Map/TableStructure.js
```diff
@@ -279,33 +287,37 @@ TableStructure.prototype.toArrayOfColumns = function() {
...
     var that = this;
     var rows = that.columns[0].values.map(function(value0, rowIndex) {
-        return that.columns.map(function(column) {
-            if (dateFormatString && column.type === VarType.TIME) {
-                return dateFormat(column.dates[rowIndex], dateFormatString);
-            }
-            return column._formattedValues[rowIndex];
-        });
-    });
+        if (!defined(rowNumbers) || rowNumbers.indexOf(rowIndex) >= 0) {
```
AlexGilleran: Once again this might not affect performance that much, but wouldn't it make more sense to loop over the rowNumbers and get that row out of `that.columns[0].values` for each one, instead of looping over `that.columns[0].values` and for each one checking all of `rowNumbers` to see if that row is in it?

RacingTadpole: yep, will do

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -172,6 +174,11 @@ const FeatureInfoSection = React.createClass({
                             <When condition={reactInfo.showRawData || !this.hasTemplate()}>
                                 <If condition={reactInfo.hasRawData}>
                                     {reactInfo.rawData}
+                                    <If condition={reactInfo.timeSeriesChart}>
+                                        <div className="time-series-chart">
```
AlexGilleran: Is `time-series-chart` built into D3 or something?

RacingTadpole: ah, good question. No, this is an arbitrary class name ostensibly to allow user-specific styling, but which is really needed so that React gives this a different key to anything else in this section.

AlexGilleran: Could you change it to `Styles.timeSeriesChart` and define `@include empty-module("time-series-chart")` in `feature-info-section.scss` then :)? That means it'll be namespaced the same as all the other classes as `tjs-feature-info-section__time-series-chart`

RacingTadpole: excellent, done.

lib/Map/TableStructure.js
```diff
@@ -295,16 +295,20 @@ TableStructure.prototype.toArrayOfRows = function(dateFormatString, rowNumbers)
...
+            }
+            return column._formattedValues[rowNumber];
+        });
+    }
+    var rows;
+    if (defined(rowNumbers)) {
+        rows = rowNumbers.map(getRow);
+    } else {
+        rows = that.columns[0].values.map((_, rowNumber) => getRow(rowNumber));
+    }
```
AlexGilleran, line 311: :crisp:

PR #1935, opened 2016-08-11
Fixes WPS parameter selection.

### LINKED ISSUE #1918: Revise tracking of InvokeFunction parameters
We are creating and knockout-tracking `parameterValues` inside a view, https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Analytics/InvokeFunction.jsx.

It would be cleaner to track the parameter values where they occur in the models.

See discussion in #1910.

### LINKED ISSUE #1933: WPS point selection broken in terriajs#master
When trying to use WPS with a point location parameter, clicking `Select location` and then clicking a point on the globe returns you to the WPS function screen but no point lat/lon is entered in the location text field.

In addition, the `contextItem` for the WPS function is displayed when viewing the function in the catalogue window, but when clicking `Select location` it is turned off, and only reappears when the catalogue window returns.

RacingTadpole:
Fixes #1933 and fixes #1918.

A recent change meant that the entire ExplorerPanel, and hence any InvokeFunction component, is unmounted while you click out a point/polygon/rectangle/etc on the map.

This required a complete reworking of how the selected data is stored, moving it from `InvokeFunction` to `WebProcessingServiceCatalogFunction`.

### DIFF
--- a/lib/Models/CatalogFunction.js
+++ b/lib/Models/CatalogFunction.js
@@ -29,6 +29,7 @@ var CatalogFunction = function(terria) {
 
     this._loadingPromise = undefined;
     this._lastLoadInfluencingValues = undefined;
+    this._parameters = [];
 
     /**
      * Gets or sets a value indicating whether the group is currently loading.  This property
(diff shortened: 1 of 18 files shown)


RacingTadpole:
This isn't quite ready - I've found a subtlety in RegionTypeParameterEditor that will take more thought.

RacingTadpole:
OK, this almost works now. The remaining issue is: if you try it with TerriaMap and look at some of the Analytics options, you cannot actually select a region.  But this is a problem with master too, so this PR still represents an improvement.

I guess this points to the need for more thorough tests around the WPS and analytics stuff, and their ReactViews, so we don't accidentally break it all again.

PR #1936, opened 2016-08-11
Filtered region mapped features that don't have descs or props

### LINKED ISSUE #1926: AREMI Vic Bio layers break feature selection
When we enable two or more layers in the Vic Bioenergy categories (all CSV based layers), it breaks feature selection. It will highlight the polygon that is clicked, and the `Feature Information` will pop up and say `No results` even though there is data. When closing the empty feature info popup a number of exceptions are printed in the console.

When having just a single layer enabled the correct feature info comes up.

Two layers together:

http://nationalmap.gov.au/renewables/#share=s-t7WN8obCICVwWhdj

Both those layers individually:

http://nationalmap.gov.au/renewables/#share=s-khGUBZprwNpMorlF
http://nationalmap.gov.au/renewables/#share=s-xgiU7dEwB3zhbrs7

AlexGilleran:
Fixes #1926 

@RacingTadpole you might want to make sure this makes sense.


RacingTadpole:
OK, I think I can make a story for why this happens.
When one imageryLayerFeatureInfos is returned by this function, and it's invalid, it all works. When it returns a valid and an invalid one, something breaks. So your solution works nicely.
I also can't find anything that your solution breaks (ABS ITT, time-varying CSVs etc).

PR #1939, opened 2016-08-12
Removed Global Styles

AlexGilleran:
Right now terria requires a bunch of global styles that hit individual element names, which makes it hard to integrate into other UIs on a component-by-component basis (you've pretty much got to use an iframe), and means we can't generate a decent styleguide.

This removes those styles, making terriajs more modular, and is the first step to allowing it to be used as a toolkit where you pick certain components rather than being forced to use the StandardUserInterface.


lib/ReactViews/Preview/MappablePreview.jsx
```diff
@@ -38,7 +38,7 @@ const MappablePreview = React.createClass({
             <div className={Styles.root}>
                 <If condition={catalogItem.isMappable}>
                     <DataPreviewMap terria={this.props.terria} previewedCatalogItem={this.props.previewed}
-                                    showMap={!this.props.viewState.explorerPanelAnimating} />
+                                    showMap={!this.props.viewState.explorerPanelAnimating || this.props.viewState.useSmallScreenInterface} />
```
AlexGilleran, line 41: derp

chloeleichen:
All look good to me, I think it can be merged once https://github.com/TerriaJS/TerriaMap/pull/61 is merged into NM

AlexGilleran:
Can we merge that then @chloeleichen ? It's super small :)

chloeleichen:
Does the PR on TerriaMap automatically get merged into NM when merged? or do we need to create seperate PR? @AlexGilleran

AlexGilleran:
Nah it doesn't get automatically merged, but I don't think it needs a PR either.

AlexGilleran:
That TerriaMap one is merged into NM now :)

PR #1943, opened 2016-08-15
1778 time interval performance

nahidakbar:


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -511,7 +511,11 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
     updateValue(this, overwrite, 'minScaleDenominator', thisLayer.MinScaleDenominator);
     updateValue(this, overwrite, 'getFeatureInfoFormats', getFeatureInfoFormats(capabilities));
     updateValue(this, overwrite, 'rectangle', getRectangleFromLayers(this._allLayersInRawMetadata));
-    updateValue(this, overwrite, 'intervals', getIntervalsFromLayer(this, thisLayer));
+    this._lazyLoadInterval = function() {
```
AlexGilleran: Maybe call it `_lazyLoadIntervals` instead? And make sure it's declared as undefined in the constructor :).

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -567,6 +571,11 @@ WebMapServiceCatalogItem.prototype._load = function() {
     var that = this;
     var promises = [];
 
+    if (this._lazyLoadInterval) {
+        this._lazyLoadInterval();
+        delete this._lazyLoadInterval;
```
AlexGilleran: We usually prefer to set things to `undefined` instead of using `delete` because changing the shape of the object causes chrome performance problems.

test/Models/WebMapServiceCatalogItemSpec.js
```diff
@@ -381,6 +381,9 @@ describe('WebMapServiceCatalogItem', function() {
             layers: '13_intervals'
         });
         wmsItem.load().then(function() {
+            // interval is now lazy loaded. '_load' gets called only prior to
+            // use of the WMS data.
+            wmsItem._load();
```
AlexGilleran: `_load` should be getting called as a result of `load` being called

kring:
@nahidakbar can you fix your git settings so that your commits are associated with your GitHub account, and sign the CLA if you haven't already?

kring:
Maybe I'm missing something @nahidakbar, but I don't understand what this PR is accomplishing.  The interval loading gets delayed slightly, but only slightly, right?  Instead of happening inside the `updateFromCapabilities` call, it happens immediately after it.  And in any case, we eventually still load the intervals in the same way, and it still takes forever.  Or have I missed something?

nahidakbar:
In the add data modal, only updateFromCapabilities gets called while the full GetCapabilities response is parsed. The full load is not called unless use selects on an item to get more detailed layer information. This is exploited to defer parsing of time interval.

kring:
Thanks for this @nahidakbar.  I ended up fixing it in a similar way as part of a different PR, though, so i'm going to close this.

Issue #1945, opened 2016-08-15
Custom font download


chloeleichen:
Since some government computers cannot download custom font, it seems that our google fonts fall back to `helvetica`.
'
I think it's worth a try loading font data inline using Base64 encoding. Here is an example of it: https://raw.githubusercontent.com/ustwo/ustwo.com-frontend/master/src/app/lib/font.css

chloeleichen:
base64 encoding doesn't solve the problem

kring:
Nothing more to do here, right?  I'm closing it.

PR #1954, opened 2016-08-16
fixing #1945, use base64 encoding for font

chloeleichen:
Issue: https://github.com/TerriaJS/terriajs/issues/1945

Tested working in all modern browsers. 

Though I cannot test whether this approach actually solve the problem of PM&C not able to use google font. Guess we can put it on staging somewhere and ask them to send a screenshot? 

It seems that it also make gulp watch stops sometimes, i think it's because the css file to watch is too big?


kring:
I set up a temporary server for this branch:
https://nationalmap-base64-font.nationalmap.nicta.com.au/

I'll ping some folks in the government to send us a screenshot.

kring:
Here's what it looks like for @tobybellwood:
image

Still default fonts @chloeleichen?

chloeleichen:
nope, looks great! tho did you notice the preview map looks very strange @kring 

No it doesn't i was lying

kring:
Latest screenshot:
image

I've asked for a screenshot with the Add Data panel open, but it looks good so far!

chloeleichen:
Maybe we should ask for a screenshot(again) with add data panel open and a preview open? 👯

kring:
Looks like it's working!

image

chloeleichen:
Yay! looks good to me!

kring:
Toby says he can now see the icons on fontawesome too!  So apparently something has changed such that custom fonts are no longer blocked.  That's good in general, but makes it hard to tell if this change is accomplishing anything.  

Here's a side-by-side screenshot:
image

chloeleichen:
but we fixed fontawesome icon (we no longer use fontawesome icons) from ages ago?

kring:
Right, fontawesome is just a way to test if custom fonts are supported, because it's super obvious on that site when they're not.

kring:
Ok here we have side-by-side images on a system confirmed not to be able to show custom fonts (at least on fontawesome):

Without this change:
image

With this change:
image

They look the same to me. :(

chloeleichen:
no good :(

PR #1956, opened 2016-08-16
Chart conflict

chloeleichen:
Fixing https://github.com/TerriaJS/terriajs/issues/1831

Can you have a look please @RacingTadpole ?


lib/Map/VariableConcept.js
```diff
@@ -92,8 +92,10 @@ VariableConcept.prototype.toggleActive = function() {
             return;
         }
     }
+    // deactivate others
+    this.parent.disableColumnsWithDifferentXType(this);
```
RacingTadpole: I think this will fail if `disableColumnsWithDifferentXType` is undefined.

lib/Map/DisplayVariablesConcept.js
```diff
@@ -103,4 +103,5 @@ DisplayVariablesConcept.prototype.toggleActiveItem = function(variable) {
     }
 };
 
+
```
RacingTadpole: To keep it clean can you get rid of this extra line?

lib/Map/TableStructure.js
```diff
@@ -46,7 +46,7 @@ var defaultDisplayVariableTypes = [VarType.ENUM, VarType.SCALAR, VarType.ALT];
  */
 var TableStructure = function(name, options) {
     options = defaultValue(options, defaultValue.EMPTY_OBJECT);
-    DisplayVariablesConcept.call(this, name, options.getColorCallback);
+    DisplayVariablesConcept.call(this, name, options.getColorCallback, options.disableColumnsWithDifferentXType);
```
RacingTadpole: Actually it disables columns from other tables, which is a bit lost with this name.
To also make it consistent with `getColorCallback`, how about `disableIncompatibleTableColumnsCallback`?

I don't think this ever worked actually, as the constructor's signature is:
`var DisplayVariablesConcept = function(name, allowMultiple, getColorCallback)`

Also - I don't see where this is ever stored by `DisplayVariablesConcept`?

lib/Models/CsvCatalogItem.js
```diff
@@ -553,6 +553,7 @@ CsvCatalogItem.prototype.setChartable = function() {
     tableStructure.allowMultiple = true;
     this.isMappable = false;
     tableStructure.getColorCallback = this.getNextColor.bind(this);
+    tableStructure.disableColumnsWithDifferentXType = this.disableColumnsWithDifferentXType.bind(this);
```
RacingTadpole: This is valid JS but we always try to declare all possible properties in the constructor. So add `this.disableColumnsWithDifferentXType = undefined;` to `DisplayVariablesConcept` (which is the parent class of `TableStructure`).

lib/Map/TableStructure.js
chloeleichen: yeah i'm a bit confused by that too but it actually works 🎅

RacingTadpole:
Also a few lint problems - otherwise looking good! (Not that I've actually run the code yet...:-))

lib/Models/CsvCatalogItem.js
```diff
@@ -553,6 +553,7 @@ CsvCatalogItem.prototype.setChartable = function() {
     tableStructure.allowMultiple = true;
     this.isMappable = false;
     tableStructure.getColorCallback = this.getNextColor.bind(this);
+    tableStructure.disableIncompatibleTableColumns = this.disableIncompatibleTableColumns.bind(this);
```
chloeleichen: Isn't it better to just add `disableIncompatibleTableColumns` here and get only set in `setChartable`, in this way we don't need to pass in an extra options to `DisplayVariablesConcept` ?

RacingTadpole: Yep, good point. Can you name it like this though?
```tableStructure.toggleActiveCallback = this.disableIncompatibleTableColumns.bind(this);`

chloeleichen: yup! I see what you mean, done!

RacingTadpole:
This is cool... but I've found a remaining issue: when I have two data sets with different x-units, eg.
image
I can't click on any of the columns from the second one.

PR #1957, opened 2016-08-16
Fix CSV with NaN depth crashing Cesium

tephenavies:
Currently, a value in a depth/height/elevation column that is coerced to `NaN` in arithmetic will cause Cesium to crash and won't allow switching to 3D mode until you refresh (because it messes up scene frustum calculations in Cesium rendering with `NaN`). This PR fixes that problem.

Test csv file:

```
Lat,Lon,Depth,Val
-33,150,not number,4
-34,151,text,5
```

Note: there is another problem with having text in an altitude column. The altitude column is marked as numeric, so trying to colour by that column causes a crash in `ckmeans.js` because there are no numeric values. This is a much smaller problem (since it doesn't crash Cesium), but it's harder to solve.


kring:
@steve9164 can you add this to CHANGES.md?

tephenavies:
Added a changelog entry. I had a go at writing a test, but it the problem crashes Cesium's Scene frustrum calculations, which I'm not sure how (or if?) you can induce from a test.

PR #1984, opened 2016-08-25
Reintroduced StandardUserInterface to enable backwards compatibility.

AlexGilleran:


lib/ReactViews/Map/Panels/setting-panel.scss
```diff
@@ -26,6 +26,9 @@
 .btn--viewer {
   composes: btn from '../../../Sass/common/_buttons.scss';
   composes: btn-grey from '../../../Sass/common/_buttons.scss';
+
+  color: $text-light;
```
RacingTadpole, line 30: Just wondering if it's worth adding a "deprecated" comment to these added lines as well, so we know we can safely remove them again one day?

RacingTadpole:
oop - found one (or it could be my own dodgy build)
image

RacingTadpole:
oop - another one under ABS :-)
image

RacingTadpole:
one more (looks like a few things in this screenshot though)...
image
http://localhost:3003/#build/terriajs/test/init/charts.json

lib/ReactViews/Map/Panels/setting-panel.scss
AlexGilleran: Unfortunately I don't think we'll ever be sure of that - I'm pretty sure a lot of these are actually just fixing stuff I missed too :(.

PR #1987, opened 2016-08-25
Make URL to serverconfig service configurable.

kring:
This is necessary for e.g. the Northern Australia map because the `serverconfig` service is found at `/serverconfig/` instead of at `/northernaustralia/serverconfig/`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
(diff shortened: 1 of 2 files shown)


kring:
I'm insta-merging this but someone (@stevage?) should review it anyway.

stevage:
Looks ok to me.

PR #1991, opened 2016-08-26
shorten url btn style

chloeleichen:
I can't actually test this on my local, can someone test this please? 
Problem is: 
screen shot 2016-08-26 at 9 35 18 am
should be:
screen shot 2016-08-26 at 10 32 30 am


RacingTadpole:
It is a bit tricky to test locally, since we don't have URL shortening enabled in dev. I did it by overriding`<If condition={this.isUrlShortenable()}>`. However I couldn't test that clicking on the button actually works.

What's your thinking with removing the `Styles.btnCheckboxOn` class?

chloeleichen:
@RacingTadpole  the `Styles.btnCheckboxOn` was useful when we were using icon font, so css know what icons to use, and it's no longer useful since we now use svg sprites to do icons

PR #2004, opened 2016-08-30
Ensure units are available for moving features

### LINKED ISSUE #2003: TableStyle Units don't flow through to moving point csvs
eg.

```
"tableStyle": {
                    "dataVariable": "speed",
                    "columns": {
                        "speed": {
                            "name": "speed (km/day)",
                            "units": "km/day"
                        }
                    }
                }
```

RacingTadpole:
Fixes #2003.


kring:
I don't see any units on the small preview chart embedded in the feature info panel, but the big chart looks good.  That might be intentional on the preview, but even if not, it's a separate issue, so I'm merging this.

RacingTadpole:
Thanks - yeah, that's intentional according to the design, although I agree some sort of x & y-axis labels would be desirable.

PR #2016, opened 2016-09-02
Recognise JSON5 everywhere instead of JSON

### LINKED ISSUE #2014: Drag-and-drop init files don't support JSON5
Ran into this while demoing. Tried to drop an init file in JSON5 format onto nationalmap.

RacingTadpole:
Fixes #2014.

I have a feeling there was a reason we didn't do this before... but what was it? @kring?


kring:
`JSON5` parsing is much slower than regular `JSON` parsing.  Also, this won't catch any place that we use Ceisum's `loadJson`.  I suggest we close this and be more selective in where we accept JSON5.  For example, we arguably don't want to use JSON5 for GeoJSON parsing.

RacingTadpole:
OK. How about I update this PR to create a new `lib/Core/readJson5.js` command, and then use it explicitly in the Add Data process?
I'll also check where `lib/Core/readJson.js` is used, and Cesum's `loadJson`, and see which are appropriate to switch to `json5`.

Oh - there will be an issue with geojson, won't there... since when you add a json file, you don't know whether it's an init file or a geojson file until you've read it. Maybe read it as json first, then if that fails, try json5...?

kring:
> Maybe read it as json first, then if that fails, try json5...?

Yeah that makes sense to me.

stevage:
If we were being more proper about this (sorry, probably my fault), we would require json5 files to have the `.json5` extension, and therefore the server would apply the `json5` mimetype. We probably shouldn't be sticking non-json in `.json` files.

kring:
> We probably shouldn't be sticking non-json in .json files.

Yeah that seems sensible to me.  How painful is it to change at this point?

RacingTadpole:
OK, I've gone with the "read it as json first, then if that fails, try json5" approach - what do you think?

Note the existing logic that any filename containing `.json` is interpreted as json means that `.json` and `.json5` are both handled by this logic.
(See https://github.com/TerriaJS/terriajs/blob/master/lib/Models/addUserFiles.js#L41)

CHANGES.md
```diff
@@ -9,6 +9,7 @@ Change Log
 * Refactored Csv, AbsItt and Sdmx-Json catalog items to depend on a common `TableCatalogItem`. Deprecated `CsvCatalogItem.setActiveTimeColumn` in favour of `tableStructure.setActiveTimeColumn`.
 * Error in geocoding addresses in csv files now shows in dialog box.
 * Fixed css styling of the timeline and added padding to the feature info panel.
+* Enhanced json support to recognise JSON5 format everywhere.
```
kring, line 12: This changelog entry isn't really true, is it?  Because `loadJson` does not support JSON5.  I've already merged this PR though so maybe you could tweak the wording in master?

RacingTadpole: Done, thanks - now reads "Enhanced json support to recognise JSON5 format for user-added files."

PR #2029, opened 2016-09-06
Drag user drawn points

rdengate:
Once points are drawn on the map, they can be dragged to reposition them. Works for both Leaflet and Cesium.


lib/Models/UserDrawing.js
```diff
@@ -427,5 +414,47 @@ UserDrawing.prototype._clickedExistingPoint = function(features) {
...
 };
 
+/**
+ * Remove point from list of pointEntities, which essentially deletes the point
+ * @param {Feature} feature Feature that user has clicked on
+ * @private
+ */
+UserDrawing.prototype._removePointFromPointEntities = function(feature) {
+    // We can't just 'remove' the feature from the entities list, because it's not necessarily the same entity,
+    // even though the id is the same.
```
kring: `EntityCollection` has a `removeById` function.

kring:
It looks like the measure tool's distance doesn't update when you drag a point.

kring:
The points lose their draggability if you switch between 2D and 3D while the measure tool is open.

RacingTadpole:
@rdengate - did you get a chance to respond to Kevin's comments?

rdengate:
@RacingTadpole I'm still working on the measure tool distance updating, sorry (it turned out to be a bigger deal than I realized).

rdengate:
@kring thanks for the code review. I've addressed all the points, so this is ready for round 2 or merge. Thanks!

tephenavies:
In Cesium, if you drag a point back to it's original location, it gets deleted:
drag points

rdengate:
Wow, that's interesting. Thanks for spotting that @steve9164! I will fix it.

rdengate:
Fixed, thanks.

tephenavies:
Do you know why the polylines sometimes have the jagged line? You can see in the gif that when the angle between adjacent line segments becomes small, one of those line segments is jagged.

rdengate:
Yeah, I think it's a Cesium issue. I have logged it as something to look into in https://github.com/TerriaJS/terriajs/issues/1981, but not sure when I'll get to that.

PR #2031, opened 2016-09-06
Geojson linestring in known WPS input parameters

rdengate:
If type of WPS input is http://geojson.org/geojson-spec.html#linestring, deliver geojson line string to WPS. This is a straightforward extension of what is already there for other geojson features.


rdengate:
Can be tested with https://github.com/TerriaJS/nationalmap/pull/325.

PR #2034, opened 2016-09-07
Support reversed lon-lat GML coords for WFS

stevage:
I have no idea why we needed this but I'm sure it's super important.


kring:
Closing this to re-open it as a PR into `oldui` in support of an NEII release.

PR #2044, opened 2016-09-13
Don't show repeated errors on base map failures

### LINKED ISSUE #1869: Layer errors show repeated popups
In the new UI (but not the old) an error downloading tiles can cause a separate "error" modal to appear for every failed tile.  This requires the user to click OK very many times before they can use the app again.  After a few tile errors and showing the error to the user, the layer should be hidden so we don't bother the user with further modals.  We saw this when the Black Marble base layer was returning errors.

kring:
Fixes #1869 
(which turned out to be specific to base maps, not a general error handling problem).

### DIFF
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -492,6 +492,9 @@ is invalid.  Please select a different base map using the Map button in the top-
 
                     layer.show = false;
                     catalogItem.terria.baseMap = undefined;
+
+                    // Don't use this base map again on startup.
+                    catalogItem.terria.setLocalProperty('basemap', undefined);
                 } else {
                     globeOrMap.terria.error.raiseEvent(new TerriaError({
                         sender: catalogItem,
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -15,6 +15,7 @@ var TimeInterval = require('terriajs-cesium/Source/Core/TimeInterval');
 var TimeIntervalCollection = require('terriajs-cesium/Source/Core/TimeIntervalCollection');
 
 var CatalogItem = require('./CatalogItem');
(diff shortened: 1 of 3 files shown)


lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -475,22 +476,43 @@ ImageryLayerCatalogItem.enableLayer = function(catalogItem, imageryProvider, opa
...
+                    layer.show = false;
+                    catalogItem.terria.baseMap = undefined;
+
+                    // Don't use this base map again on startup.
+                    catalogItem.terria.setLocalProperty('basemap', undefined);
+                } else {
+                    globeOrMap.terria.error.raiseEvent(new TerriaError({
+                        sender: catalogItem,
+                        title: 'Error accessing catalogue item',
+                        message: '\
```
RacingTadpole, line 502: Do you have any suggestions for how to test this? Black Marble isn't giving me any errors any more.
From looking at the code changes, I'm curious that `indexOf(catalogItem) >= 0` stops lots of errors, as opposed to, say, `indexOf(catalogItem) === 0` ?

kring: You can test this by switching to the Australian Hydrography base map, which is currently broken.  `=== 0` would assume that the failing layer is the bottom-most, which is not true when the base is a composite like Hydrography and the upper layer is the one that fails.

RacingTadpole:
This just presents a blue globe after the error message. You can even add data and zoom around on it, all on the blue globe.
Should we revert to the previous base map (if any) if there's an error?

kring:
Well, we don't really know the previous base map at the time we detect this failure.  We could keep track of that, but that won't help if the initial basemap (either the default, or the last one the user used) is broken.  I thought about just jumping to the "next" basemap, but in the end I thought it would be ok to leave the globe blue and tell the user in the error message to change to a different base map.

PR #2047, opened 2016-09-13
Show contextItem during location selection

### LINKED ISSUE #1942: WPS contextItem is not shown during location selection
This was brought up in #1933 and still remains.

The analysis study area flickers but then disappears during the selection.

rdengate:
Fixes #1942. It seems that the contextItem was displayed according to whether the invokeFunction panel was displayed, but when location selection occurs, this is hidden. I think it should instead be the responsibility of MapInteractionWindow, which controls all location selection (point, line, rectangle, polygon).


rdengate:
See https://github.com/TerriaJS/terriajs/pull/2047

rdengate:
Whoops. I meant to link to the PR in the issue, not the PR :-)

PR #2052, opened 2016-09-13
Fix legend for column with all null replaced values

tephenavies:
Changing the order of null replacements and the generation of `numericalValues` leads to several improvements including not running ckmeans/quantile on null values, and it also fixes the following legend bug:

pasted image at 2016_09_13 17_40


RacingTadpole:
Looks good! Can you update CHANGES.md?

RacingTadpole:
Actually it'd be great if you could add a new spec to `TableColumnSpec.js` covering `numericalValues` too.

PR #2063, opened 2016-09-14
Assign explicit keys to elements in repeated <Choose>s.

kring:
This avoids a warning from React and also causes the WMS styles and WFS typenames to actually show up on the dataset info page.


RacingTadpole:
This looks fine but I haven't seen this bug - eg. Broadband Availability is a WMS, and I don't get the console warning - can you give me an example where this applies?

kring:
What version of react do you have installed?  It may have only become a problem after an upgrade of react.

PR #2069, opened 2016-09-14
Chart legend flickers

chloeleichen:
Fixing https://github.com/TerriaJS/terriajs/issues/2040

@RacingTadpole  I couldn't see why it flickers now since all the logic you had were still there. All I can think of is to try a lower threshold for the tooltip to hide. Can you have a try and let me know if you think it's better?


kring:
@RacingTadpole can you review this one?

PR #2071, opened 2016-09-14
Fix issue where generic parameters not passed to wps service

rdengate:
After this fix, any input entered into GenericParameterEditor input fields is passed to the WPS service.


kring:
Please update CHANGES.md in master.

PR #2072, opened 2016-09-15
Improve handling of active time columns with changing table data

RacingTadpole:
An earlier PR introduced sorting of all tables by their time column, if they had any.  This is a good idea, so charts appear with ordered x-axes.
However, it has the (unproven) potential to interact badly with polled csvs, where features are identified by row number - which is currently the situation with aremi's power stations csv. This csv has a time column, but it is not an active time column - it gives the date of the last generation data.

This PR only sorts by an _active_ time column.

It also cleans up a potential problem where a pre-existing active time column would be left stranded after sorting.


RacingTadpole:
Actually hold off on this one. I will overhaul how I handle `activeTimeColumn` more generally - keeping a reference to an actual column is causing all sorts of problems with polling new data.

RacingTadpole:
OK, this is ready to go (once travis is happy)

PR #2074, opened 2016-09-15
Show a message when region mapping on old browsers.

kring:
These browsers won't support vector tile based region mapping when we roll it out.

### DIFF
--- a/lib/Models/RegionMapping.js
+++ b/lib/Models/RegionMapping.js
@@ -578,6 +578,22 @@ function createNewRegionImageryLayer(regionMapping, layerIndex, regionIndices, g
 
     globeOrMap = defaultValue(globeOrMap, catalogItem.terria.currentViewer);
 
+    var terria = globeOrMap.terria;
+
+    // Inform the user that region mapping will not be supported in old browsers in the future.
+    if (!terria.configParameters.regionMappingOldBrowserNotifiedThisSession && typeof ArrayBuffer === 'undefined') {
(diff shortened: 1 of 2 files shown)


lib/Models/RegionMapping.js
```diff
@@ -578,6 +578,22 @@ function createNewRegionImageryLayer(regionMapping, layerIndex, regionIndices, g
 
     globeOrMap = defaultValue(globeOrMap, catalogItem.terria.currentViewer);
 
+    var terria = globeOrMap.terria;
+
+    // Inform the user that region mapping will not be supported in old browsers in the future.
+    if (!terria.configParameters.regionMappingOldBrowserNotifiedThisSession && typeof ArrayBuffer === 'undefined') {
+        terria.configParameters.regionMappingOldBrowserNotifiedThisSession = true;
```
RacingTadpole, line 585: This is probably over the top, but ideally we'd add this configParameter into Terria.js's `defaultConfigParameters` too, right?

kring: Nah, I don't think it's needed.  It's not the sort of config parameter we'd ever put in config.json, so this is a bit of a hack perhaps.  This is actually iteration number 3.  The first put it in local storage, but then it only showed once ever, which is probably not what we want.  Then I made it a module-scale variable, which was fine until the first occurrence was sometimes a result of preview (where errors are intentionally not shown) so it wouldn't be shown at all (sometimes).  So now I'm putting it on the Terria instance.  It's on `configParameters` only out of convenience.

PR #2106, opened 2016-09-20
add "long" to longitude type hint

nahidakbar:


kring:
@nahidakbar can you sort out your github account email addresses so that GitHub knows you committed this change?

PR #2112, opened 2016-09-21
helper messages 

chloeleichen:
Fixing https://github.com/TerriaJS/terriajs/issues/2087

@RacingTadpole this is a slightly different approach to https://github.com/TerriaJS/terriajs/pull/2097 
Can you have a look and let me know what you think?


RacingTadpole:
This looks good to me, thanks Chloe! I think others should take a look too though - @philipgrimmett, @stevage, @kring, @AlexGilleran.

So if you click on the map before adding a dataset you get:

image

(I find "maybe" a bit odd here - how about "Click 'Add Data' to add data to the map.")

When you add your first dataset to a session, you get:

image

(I think that's too verbose - I'd go with "Click on the map to learn more about a location")

And if you click somewhere there is no data, you get this:

image

(I find the grammar a bit jarring - how about "No data is available here - try another location.")

RacingTadpole:
Oh, can you also please merge in the latest master and update CHANGES.md?

chloeleichen:
Updated the wordings

lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
```diff
@@ -101,6 +101,21 @@ const FeatureInfoPanel = React.createClass({
         }
     },
 
+    getMessageForNoResults() {
+        if(this.props.terria.nowViewing.hasItems) {
+            // if no local storage flag
```
RacingTadpole: Sorry to be pedantic - these comments are out-of-date though :-)
And to be even more uselessly pedantic - I think everywhere else we're writing `if (...)` not `if(...)`... do you mind adding in the space here and in `DataCatalogItem` above?

chloeleichen: Done 💃

lib/ReactViewModels/ViewState.js
```diff
@@ -55,6 +55,9 @@ export default class ViewState {
          */
         this.featureInfoPanelIsCollapsed = false;
 
+        // is user adding data for the first time?
+        this.firstTime = true;
```
kring: This needs a better name.  First time for what?

kring: I'll change it since I already have this branch checked out.

PR #2114, opened 2016-09-21
Css updates

chloeleichen:
Removed `_layout.scss` which is not being used. 
Updated the `pre` look in the notification window
Hopefully fixing https://github.com/TerriaJS/terriajs/issues/2092 but I wasn't able to replicate it at the first place.


kring:
@RacingTadpole can you review, since you wrote the bug this is meant to fix?

RacingTadpole:
Yep, that fixes it for me. Thanks!

PR #2123, opened 2016-09-26
Add UI to select the WMS style when multiple are available

### LINKED ISSUE #1511: Allow users to select WMS styles
If the WMS `getCapabilities` advertises a list of styles (most do), it would be very useful to allow the user to select these from a dropdown list under "layer properties"

### LINKED ISSUE #1888: Enhancing the min/max WMS request behaviour
The new ability to specify COLORSCALERANGE min/max (https://github.com/TerriaJS/terriajs/issues/774) is great, but I just discovered the hard way that if you reverse the min/max values you get a very cryptic error.   

Could we catch the case where the user inputs a max value that is less than the min value?

And could we also reverse the order of the widgets?  (max on top, min on bottom, so it matches the sense of the y axis on the legend)

Here's a screen grab of my error:

@jiwalker-usgs can you take a look at this?

kring:
You can try it out with:
http://localhost:3001/#build/TerriaJS/test/init/wms.json

image

image

Also adds better error handling for the COLORSCALERANGE functionality.

Fixes #1511 
Fixes #1888


kring:
Failing tests, give me a minute before looking at this (I'm sure everyone was rushing to do so!) :)

kring:
Ok, all good now, rush away!

rsignell-usgs:
@kring, okay, I was on leave for 10 days, and at workshop today, but will try this out tomorrow am if not tonight.    Unless you have an instance going somewhere.   This all sounds awesome!

rsignell-usgs:
Okay, tested this out, and I see the proper controls, but I can't get this window to close:

kring:
Something is pretty broken in that deployment.  Check out the console:
image

I'm guessing there's some kind of problem with React?  Maybe try removing your `node_modules`, running `npm install` again, and rebuilding?

rsignell-usgs:
On my deployment, I can change the colorscalerange, but the legend is now longer updating correctly to reflect the values the user types in:

rsignell-usgs:
While we are mucking with that example, can we change the variable displayed to be wave height instead of wave direction?   And also change the position of the widgets in the UI so that max is on top of min, (so the same sense as in the legend, with max value on top)?   I keep typing in values backwards and getting scolded.   :smile_cat:

kring:
Ok, I've fixed that bug that prevented the legend from updating, flipped the min/max order, and changed the test init file to use wave height instead of direction.

rsignell-usgs:
Awesome!   I tried this out, and it worked great in 3D, but I noticed that it was loading extra tiles so I switched to 2D mode, and that seemed to work fine, but when I then switched the WMS style in 2D mode, only the legend updated.   Changing COLORSCALERANGE via the GUI in 2D mode also didn't seem to work:  only the legend got updated.

kring:
Ok it's working in Leaflet now too.  Thanks for the QA @rsignell-usgs!

rsignell-usgs:
Yep, i confirm it's working nicely in 2D:

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -134,12 +141,60 @@ var WebMapServiceCatalogItem = function(terria) {
...
 
+    /**
+     * Gets or sets a value indicating whether the user's ability to change the display properties of this
+     * catalog item is disabled.  For example, if true, {@link WebMapServiceCatalogItem#styles} should not be
+     * changeable through the user interface.
+     * This property is observable.
+     * @type {Boolean}
+     * @default false
+     */
+    this.disableUserChanges = false;
```
RacingTadpole, line 152: Interesting. There's a legacy `this.disableUserChanges` on `TableCatalogItem` too, which I see now is completely unused. I'll submit a PR to get rid of that too.

kring: Perhaps that's something that never made it over to the new UI?

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -377,12 +528,14 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
 
         if (defined(this.layers)) {
             var layers = this.layers.split(',');
+            var styles = (this.styles || this.parameters.styles || '').split(',');
```
RacingTadpole, line 531: Why do you define `styles` as a string with commas in it, rather than an array? Is it to match up with `layers`?

kring: Yep, exactly.  I'm not totally convinced it's the right thing, but consistency is good... right?

RacingTadpole: If it makes sense for the person constructing the json to specify `layers` and `styles` in the same way, yep. Or are they different enough concepts that there wouldn't be any great risk of catalog-constructor-error with `styles` as an array?
I'm guessing that you felt there was enough reason to do it like layers in the first place, so I'll go with that :-)

RacingTadpole:
This is great!  I'd recommend a UI person to take a look at it eventually too though... :-)

PR #2126, opened 2016-09-27
Document corsDomains. Fixes #1958.

### LINKED ISSUE #1958: Document corsDomains
The `corsDomains` property (defined [here](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/Terria.js#L507)) should be documented [here](http://terria.io/Documentation/Customizing/Config-JSON/)

stevage:


Documentation/CatalogManagement/Initialization-File.md
```diff
@@ -59,7 +61,14 @@ Catalog files can be edited three ways:
 
 ## Catalog file properties
 
-### Setting camera positions
+### `corsDomains`
+
+By default, Terria proxies all requests, assuming that the servers do not support CORS. You can add hosts that are known to support CORS to this property to avoid proxying them. (This property is also supported by TerriaJS-Server).
```
kring: Not quite accurate.  It proxies all requests that are in the proxy whitelist, not "all requests".  And there's no corresponding property on `terriajs-server` as far as I know.

stevage: So, in other words, totally inaccurate. :) I'll have another go.

PR #2131, opened 2016-09-28
Support for new ckan terria_view extension 

nahidakbar:


nahidakbar:
@kring

kring:
I tweaked this slightly:
- It no longer allows the same origin to post messages by default.  In NationalMap and such, messages from the same origin would be fine, but TerriaJS is a library and we can't assume that things on the same origin should be allowed to post messages in _all_ applications TerriaJS ends up being used in.  The parent or opener can use the `allowOrigin` message to whitelist the same origin if desired.
- Explicitly treated an undefined `allowOrigin` property as meaning no origins are allowed.  Previously if someone could make their message come from an empty string origin, the message would be allowed.  I don't know if this is possible in any browsers, but it's much easier to guard against the possibility than to convince myself it can't happen.

PR #2139, opened 2016-09-30
Change old browser region mapping message

tephenavies:
Region mapping will no longer work in old browsers as of TerriaJS/nationalmap#345. Change the message to an exception and throw it every time an outdated browser user tries to open a region mapped layer (not just the first time).

### DIFF
--- a/lib/Models/RegionMapping.js
+++ b/lib/Models/RegionMapping.js
@@ -630,18 +630,17 @@ function createNewRegionImageryLayer(regionMapping, layerIndex, regionIndices, g
 
     var terria = globeOrMap.terria;
 
-    // Inform the user that region mapping will not be supported in old browsers in the future.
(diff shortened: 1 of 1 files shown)


RacingTadpole:
Woo hoo!

tephenavies:
This fails IE9 tests. How can I disable region mapping tests on IE9?

kring:
You can conditionally disable tests using something like this:
https://github.com/TerriaJS/terriajs/blob/master/test/Models/CesiumSpec.js#L23

Maybe use `typeof ArrayBuffer !== 'undefined'` as the condition.

tephenavies:
@kring I've moved the region mapping in old browsers error into the section that actually uses vector tiles. The tests should no longer fail on IE9, and any first or third party maps that use WMS region mapping should continue to work on IE9.

Issue #2143, opened 2016-09-30
Lat/lon/elev box obscures disclaimer, text still jumps around


RacingTadpole:
The lat/lon/elev box is now sitting on top of the disclaimer, eg.:
![image](https://cloud.githubusercontent.com/assets/1757858/19009631/373d2704-87b9-11e6-8fda-aa6064d448c1.png)

It can also spill over onto two lines sometimes (I'm having trouble catching this to screenshot it).

Also, while the box is no longer changing size, the words (Lat, Lon, Elev) still move around inside it. Can they be made fixed?

chloeleichen:
@philipgrimmett says:

> re. covering the disclaimer: would could decide as a team to look past it, change if it ever gets questioned...

kring:
Don't think that'll fly.  From the Bing Maps terms of use (https://www.microsoft.com/maps/product/terms.html):

> Change, obscure, or minimize any logo, trademark, copyright or other notice of Microsoft or its suppliers, or digital watermarks in the Content; except that we may make alternative logo, trademark and copyright attribution requirements available for use with small maps or on small devices; if available you will find them here.

We can possibly move the credits somewhere else, but leaving them there and obscuring them is bad news.

PR #2145, opened 2016-10-04
Location distance legend

chloeleichen:
Fixes: https://github.com/TerriaJS/terriajs/issues/2143


kring:
I'll leave this one to yu @RacingTadpole.

RacingTadpole:
Sure!

RacingTadpole:
This is great, thanks @chloeleichen - just one thing - can you un-obscure the disclaimer again, per @kring's comment in #2143?

chloeleichen:
now the legend doesn't block the disclaimer in all base-maps except for one: 
screen shot 2016-10-10 at 9 48 37 am

is this acceptable ?

RacingTadpole:
Thanks! Doesn't the disclaimer height depend on the window's width though? Is there a way to float the lat/lon/elev box above the disclaimer regardless of its height?

chloeleichen:
@RacingTadpole no they don't interfere with each other.

RacingTadpole:
Sorry, I meant situations like this:
image
Although I see now that the live NationalMap does the same thing. @kring - what do you think?

chloeleichen:
unfortunately the legend does not know by default how tall the disclaimer is. besides, there just isn't enough space for it.

chloeleichen:
we could, however, move up the disclaimer and pin the legend to the bottom, let me have a try

RacingTadpole:
Maybe... although the disclaimer is a lot wider than the lat/lon box, so it might be more annoying higher up.

chloeleichen:
what about this? 
screen shot 2016-10-10 at 10 20 33 am

RacingTadpole:
I think the lat/lon info should be higher priority than the disclaimer, so should be above it ... but happy to hear from @philipgrimmett or others.

RacingTadpole:
Sorry @chloeleichen - I was too hasty merging that - there are three lint errors - could you please fix these (either in a new branch or direct in master)?

```
/home/travis/build/TerriaJS/terriajs/lib/ReactViews/DataCatalog/DataCatalogMember.jsx
  11:34  error  Component definition is missing display name  react/display-name
/home/travis/build/TerriaJS/terriajs/lib/ReactViews/Search/SearchBox.jsx
  14:34  error  Component definition is missing display name  react/display-name
/home/travis/build/TerriaJS/terriajs/lib/ReactViews/Search/SearchHeader.jsx
  7:34  error  Component definition is missing display name  react/display-name
```

chloeleichen:
@RacingTadpole 
I tried to look it up and found this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md

Tho i don't know why it only comes up now and only in those two files. Do you happen to know why?

Here is the PR anyway: https://github.com/TerriaJS/terriajs/pull/2157

PR #2147, opened 2016-10-05
Fix Terria analytics

kring:


lib/ReactViews/Analytics/BooleanParameterEditor.jsx
```diff
@@ -11,31 +12,47 @@ const BooleanParameterEditor = React.createClass({
...
+                <button type='button'
+                        className={Styles.btnRadio}
+                        title={description}
+                        onClick={this.onClick}>
+                    {value && <Icon glyph={Icon.GLYPHS.checkboxOn}/>}
+                    {!value && <Icon glyph={Icon.GLYPHS.checkboxOff}/>}
+                    {name}
+                </button>
+            </div>
+        );
```
rdengate, line 34: When testing a boolean input WPS parameter, I'm getting a text field...

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
```diff
@@ -11,31 +12,47 @@ const BooleanParameterEditor = React.createClass({
...
     },
 
     renderRadio(state) {
         let name;
         let description;
-        let classNames;
-        const value = this.props.previewed.parameterValues[this.props.parameter.id];
+        const value = this.props.parameter.value === state;
```
rdengate, line 40: If the default value of the parameter is false, doesn't that mean that no radio button will be true? And doesn't one button have to be true?

rdengate (commented):
Looks good! It's great to see how having the value in FunctionParameter.js simplifies the code. Certainly a big change though. I've made a few comments.

lib/Models/RegionParameter.js
```diff
@@ -46,63 +47,73 @@ defineProperties(RegionParameter.prototype, {
...
+            FunctionParameter.defaultValueSetter.call(this, value);
+        }
+    },
+
+    /**
+     * Gets the region provider indicating the type of region that this property holds data for.
+     * @memberof RegionParameter.prototype
+     * @type {RegionProvider}
+     */
+    regionProvider: {
```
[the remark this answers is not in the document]
kring: Good idea, done.

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
kring: As far as I can tell, WPSCatalogFunction doesn't support booleans?  It's created `StringParameter`s on that example catalog function, anyway.

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
kring: For booleans that use a radio button (rather than a checkbox), `renderRadio` is called twice, once with `state` true and once with it false.  So I think it will work correctly if the value is false.  Am I missing something?

kring:
Thanks for the review @rdengate, back to you.

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
rdengate: Ah -- are you saying this works for the analytics functions?

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
rdengate: I might have the wrong idea here, but if this.props.parameter.value is false, value is false for both states, isn't it? This is only relevant if the value of this.props.parameter.value can be false; I'm assuming this can pick up a default value of false from somewhere.

kring: No, when `this.props.parameter.value` is false, and `state` is false, `value` is true.  Am I misunderstanding you?

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
rdengate: Sure, that makes sense. We should add a boolean parameter for WPS at some stage, although I guess no one has asked for it yet. I'm happy with this then. Thanks!

lib/ReactViews/Analytics/BooleanParameterEditor.jsx
rdengate: oh, you're right! Yep. All good. Thanks!

PR #2158, opened 2016-10-10
Add the ability in the UI to select WMS dimensions like Elevation

### LINKED ISSUE #1512: Allow users to select WMS ELEVATION and other dimensions
When WMS `getCapabilities` advertise a list of `ELEVATION` values, allow user to select from available `ELEVATION` values in a dropdown list.

### LINKED ISSUE #1778: Poor performance adding WMS with lots of time intervals

### LINKED ISSUE #1976: Generalize test for ncWMS service
Right now we test for a ncWMS service by seeing if a `boxfill/rainbow` style is available:
https://github.com/TerriaJS/terriajs/blob/c8141034ba1025bcc115a0e2b5b4323b1ee94611/lib/Models/WebMapServiceCatalogItem.js#L367-L368

The thinking was that all ncWMS services would advertise this style, but ncWMS2 (which will be included in the THREDDS Data Server 5.0 when it is officially released) has dropped this style.

Here's an example of a ncWMS2 getCapabilities from a development THREDDS Data Server 5.0:

Perhaps it would be wise to distinguish between ncWMS and ncWMS2 and add a check for a style named `default-scalar/default`, something like:

```
            if (!defined(this.isNcWMS2) && myLayer.Style[j].Name && myLayer.Style[j].Name.match(/default\-scalar\/default/i)) {
              this.isNcWMS2 = true;
```

It seems prudent to identify ncWMS and ncWMS2 separately rather than just represent both as ncWMS, as their capabilities may be different.   

@guygriffiths, do you agree?

kring:
Fixes #1512 
Fixes #1778
Fixes #1976

image

image


kring:
This fixes #1976 now, too, to make the COLORSCALERANGE functionality work with ncWMS2 servers.

CHANGES.md
```diff
@@ -8,6 +8,7 @@ Change Log
 * Fixed calculation of end dates for moving-point csv files, which could lead to points disappearing periodically.
 * Fixed a bug that prevented fractional seconds in time-varying WMS periodicity.
 * Added the ability to the workbench UI to select the `style` to use to display a Web Map Service (WMS) layer when multiple styles are available.
+* Added the ability to the workbench UI to select from among the available dimensions of a Web Map Service (WMS) layer.
```
RacingTadpole, line 11: Worth adding to `CHANGES.md` that the COLORSCALERANGE functionality now works with ncWMS2 servers too?

RacingTadpole:
This isn't really the province of this PR - but I notice that sharing doesn't share the color min or max.
For me, it also switches the order of the two WMS test items I had enabled (one shown, one not).

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -515,14 +644,40 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
         myLayer = myLayer[0];
         if (myLayer && myLayer.Style && myLayer.Style.length > 0) {
           for (var j = 0; j < myLayer.Style.length; ++j) {
-            if (!defined(this.isNcWMS) && myLayer.Style[j].Name && myLayer.Style[j].Name.match(/boxfill\/rainbow/i)) {
+            if (!defined(this.isNcWMS) && myLayer.Style[j].Name &&
+                (myLayer.Style[j].Name.match(/boxfill\/rainbow/i) ||
+                 myLayer.Style[j].Name.match(/default-scalar\/default/i) ||
+                 myLayer.Style[j].Name.match(/default-vector\/default/i))) {
```
RacingTadpole, line 653: Are these standards-defined style names that clearly mark the server out as NcWMS? Could you put in a ref to the doc maybe?

kring: No, it's just a heuristic.  The ncWMS developer recently added a way to reliably detect the availability of COLORSCALERANGE (see #1976), but we can't rely on that yet exclusively yet because it's so new.  So basically we guess based on the presence of these styles, and the user can override our guess by explicitly setting the property values.  #1976 has more discussion.

CHANGES.md
kring: Good idea, done.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -515,14 +644,40 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
...
       }
     }
 
+    if (!defined(this.supportsColorScaleRange)) {
+        this.supportsColorScaleRange = this.isNcWMS;
+
+        if (!this.supportsColorScaleRange) {
+            var hasExtendedRequests = capabilities.Capability &&
+                capabilities.Capability.ExtendedCapabilities &&
+                capabilities.Capability.ExtendedCapabilities &&
```
RacingTadpole: I admire your desire to make this super thorough! But probably don't need the extra check for `capabilities.Capability.ExtendedCapabilities` :-)

kring: Hahah thanks!

kring:
> This isn't really the province of this PR - but I notice that sharing doesn't share the color min or max.
> For me, it also switches the order of the two WMS test items I had enabled (one shown, one not).

I'll fix the problem where the min/max aren't preserved on sharing for this PR.  The other is separate.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -662,18 +829,18 @@ WebMapServiceCatalogItem.prototype.updateFromCapabilities = function(capabilitie
...
     }
-    updateValue(this.parameters, overwrite, 'srs', srs);
+    updateValue(wmsItem.parameters, overwrite, 'srs', srs);
 
-    if (this.isNcWMS) {
-        updateValue(this, overwrite, 'colorScaleMinimum', -50);
-        updateValue(this, overwrite, 'colorScaleMaximum', 50);
+    if (wmsItem.supportsColorScaleRange) {
+        updateValue(wmsItem, overwrite, 'colorScaleMinimum', -50);
+        updateValue(wmsItem, overwrite, 'colorScaleMaximum', 50);
```
RacingTadpole, line 843: Out of interest... why -50 to 50?

kring: @jiwalker-usgs chose that range originally, but I don't know if there was any particular reason for it.  It might be fairly arbitrary.

test/Models/WebMapServiceCatalogGroupSpec.js
```diff
@@ -77,14 +78,19 @@ describe('WebMapServiceCatalogGroup', function() {
         group.flatten = true;
 
         loadWithXhr.load = function(url, responseType, method, data, headers, deferred, overrideMimeType, preferText) {
-            expect(url).toContain('GetCapabilities');
-            return loadWithXhr.defaultLoad('test/GetCapabilities/GeographicOnly.xml', responseType, method, data, headers, deferred, overrideMimeType, preferText);
+            if (url.indexOf('GetCapabilities') >= 0) {
```
RacingTadpole: Why not use `jasmine.Ajax.stubRequest`?

lib/Models/WebMapServiceCatalogItem.js
RacingTadpole: I just tried adding `"colorScaleMinimum": -25` to the json file, and it gets overridden to -50 - would it be worth letting this be configurable?

kring: oh yeah definitely, that's a bug.

test/Models/WebMapServiceCatalogGroupSpec.js
kring: Only because all the other tests in here already used this technique.  Let me see if I can easily switch them all over.

lib/Models/WebMapServiceCatalogItem.js
RacingTadpole: oh yeah - it kinda kills the sharing thing too :-)

rsignell-usgs:
I installed the latest TerriaMap from master, but that wasn't pulling the latest changes from terriajs, so I updated `package.json` to:

``` ...
    "svg-sprite-loader": "0.0.27",
    "terriajs": "git://github.com/TerriaJS/terriajs",
    "terriajs-catalog-editor": "^0.2.0",
    "terriajs-cesium": "1.23.2",
```

but now I get:


Can you help provide me with a TerriaMap/TerriaJS configuration that displays the COLORSCALERANGE controls for ncWMS2 ndpoints?

rsignell-usgs:
I'm also getting these warnings on `npm install`:

```
npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.14
npm WARN react-addons-pure-render-mixin@15.3.2 requires a peer of react@^15.3.2 but none was installed.
```

PR #2170, opened 2016-10-13
Region WPS

rdengate:
This allows region as a WPS input parameter, and also changes the widget so that the region is selected on the map rather than on a preview map in the data catalog (this change affects Analytics as well). 

Fixes some bugs as well. For example, countries can now be selected by clicking on the map rather than only via autocomplete, and autocomplete now works if user enters a capital letter as the first letter of the region.


Documentation/Customizing/WPS-parameters-guide.md
```diff
@@ -0,0 +1,19 @@
+## Guide to WPS parameters: which UI elements they map to
```
kring, line 1: Nice!

lib/Models/RegionParameter.js
```diff
@@ -96,7 +96,11 @@ RegionParameter.prototype.findRegionByID = function(regionID) {
     }
 
     return regionProvider.regions.filter(function(region) {
-        return region.id === regionID;
+        if (typeof region.id === 'string' && typeof regionID === 'string') {
```
kring: Why is this change necessary?  I think it's ok for region IDs to be case sensitive.  Or, put another way, the replacements system should take care of any adjustments to the ID that are necessary.

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -95,6 +99,10 @@ WebProcessingServiceCatalogFunction.parameterConverters = [
                 return undefined;
             }
 
+            if (input.Identifier === "regionType") {
```
kring: It seems a bit strange to check this here.  Can we handle it just by ordering the converters differently?  i.e. put the `regionType` converter first in the list, and then this converter works on any `LiteralData` that isn't handled by a prior converter?  That sort of thing will be necessary if someone wants to extend the set up parameters from outside TerriaJS, because they won't be able to add an actual `Identifier` check to our existing converters easily.

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -139,6 +147,44 @@ WebProcessingServiceCatalogFunction.parameterConverters = [
         }
     },
     {
+        id: 'RegionType',
+        inputToFunctionParameter: function(catalogFunction, input) {
+            if (!defined(input.LiteralData)) {
+                return undefined;
+            }
+
+            if (input.Identifier !== "regionType") {
```
kring: `Identifier` is basically the name of the parameter, right?  What if we want to select two region types?  That's not purely theoretical, the Spatial Detailing service does exactly that.

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -139,6 +147,44 @@ WebProcessingServiceCatalogFunction.parameterConverters = [
...
+                return undefined;
+            }
+
+            var allowedValues = input.LiteralData.AllowedValues;
+            var validRegionTypes;
+            if (defined(allowedValues)) {
+                var allowed = allowedValues.Value.slice();
+                if (typeof(allowed) === "string") {
```
kring: No parenthesis around `(allowed)`, use single quotes.

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -177,6 +223,60 @@ WebProcessingServiceCatalogFunction.parameterConverters = [
...
+             var schema = input.ComplexData.Default.Format.Schema;
+             if (schema.indexOf('http://geojson.org/geojson-spec.html#') !== 0) {
+                 return undefined;
+             }
+
+             if (schema.substring(schema.lastIndexOf('#') + 1) !== 'multipolygon') {
+                 return undefined;
+             }
+
+             if (input.Identifier.indexOf('region') === -1) {
```
kring: I think this test is pretty suspect, too.

It seems to me, anytime a WPS wants a polygon, the UI should allow the user to specify that polygon in a number of ways, including:
- By clicking out polygon points.
- By drawing a bounding box.
- By selecting a region.
- Maybe: by selecting an existing polygon on the map?

It's up to the user how they define their polygon, it's not necessarily specified by the server (though perhaps there could be a mechanism for the server to hint at a good default).

Is that totally crazy?

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -177,6 +223,60 @@ WebProcessingServiceCatalogFunction.parameterConverters = [
...
+                 return undefined;
+             }
+
+             if (!defined(catalogFunction.regionTypeParameter)) {
+                 throw new DeveloperError('No regionTypeParameter found, but regionGeometry requires a regionTypeParameter.');
```
kring: Why does it need a region type parameter?

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -177,6 +223,60 @@ WebProcessingServiceCatalogFunction.parameterConverters = [
...
+                 isRequired: (input.minOccurs | 0) > 0,
+                 regionProvider: catalogFunction.regionTypeParameter
+             });
+         },
+         functionParameterToInput: function(catalogFunction, parameter, value) {
+             if (!defined(value) || value === '' || !defined(value.id)) {
+                 return undefined;
+             }
+
+             // Value is an object with id and fid. We want an actual polygon for WPS, but not for analytics.
```
kring: I'd argue that's because the input is a polygon, and we're only specifying the polygon by selecting a region as a convenience.  It's possible a different kind of input (even if it's one we don't have yet) could accept a region type and identifier.

lib/Models/WhyAmISpecialFunction.js
```diff
@@ -29,6 +29,8 @@ var WhyAmISpecialFunction = function(terria) {
         name: 'Region Type',
         description: 'The type of region to analyze.'
     });
+    // Let the RegionParameter handle showing it in the UI, rather than displaying it directly.
+    this._regionTypeParameter.showInUi = false;
```
kring: We can probably do this automatically at the UI level.  Look for a `RegionParameter` with a `regionType` property equal to this `RegionTypeParameter`.  If we find one, don't show the UI for this `RegionTypeParameter`.

lib/ReactViews/Analytics/RegionDataParameterEditor.jsx
```diff
@@ -166,12 +166,13 @@ const RegionDataParameterEditor = React.createClass({
             );
         }
         return (
+            // Don't break the lines around the link to csv-geo-au, or whitespace stripping will ruin the formatting in
+            // the rendered version.
```
kring, line 170: Ugh, thanks React.

lib/Models/MapInteractionMode.js
```diff
@@ -24,6 +24,12 @@ function MapInteractionMode(options) {
     this.onCancel = options.onCancel;
 
     /**
+     * Opportunity to display extra custom UI elements.
+     * @type {Function}
+     */
+    this.customUi = defaultValue(options.customUi, function() { return ""; });
```
kring: This is pretty interesting.  From the standpoint of the `MapInteractionMode`, this is kind of like an opaque object.  In our React-based UI, it's a function that returns a React component.  In our old Knockout based UI, it might be a view-model or something like that, so not a function at all.  It might be good to document it this way.  Something like:

```
/**
 * Gets or sets the details of a custom user interface for this map interaction mode.  This property is not
 * used by the `MapInteractionMode` itself, so it can be anything that is suitable for the user interface.  In
 * the standard React-based user interface included with TerriaJS, this property is a function that is called
 * with no parameters and is expected to return a React component.
 * @type {Any}
 */
```

Also the default value should probably be undefined.

kring (changes requested):
I was skeptical at first, but I really like the new region picking UI!

lib/Models/UserDrawing.js
```diff
@@ -157,7 +157,9 @@ var UserDrawing = function(options) {
     // helper for dragging points around
     var that = this;
     this.dragHelper = new DragPoints(options.terria, function(customDataSource) {
-        that.onPointMoved(customDataSource);
+        if (typeof that.onPointMoved === "function") {
```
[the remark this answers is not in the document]
rdengate: Sure, will change. From what I understand, javascript allows both interchangeably though... is there a style guide I should be following?

lib/Models/RegionParameter.js
rdengate: This is the change that allows countries to be selected by clicking on them. Otherwise you get to here and (for example) Australia is compared to australia and the region id is not returned. Do I need to fix this in regionProvider instead though?

lib/Models/WebProcessingServiceCatalogFunction.js
rdengate: thanks, changed.

lib/Models/WebProcessingServiceCatalogFunction.js
rdengate: RegionParameter takes one, which it uses to determine region ID.

lib/Models/MapInteractionMode.js
rdengate: Thanks, changed.

lib/Models/WhyAmISpecialFunction.js
rdengate: Done, thanks.

rdengate:
Thanks @kring. I believe any comments I haven't answered above have been addressed in the offline discussion but let me know if I missed anything. This is ready for round 2 :-)

kring (changes requested):
Looks good other than these two small comments.

lib/Models/MapInteractionMode.js
```diff
@@ -24,6 +24,14 @@ function MapInteractionMode(options) {
     this.onCancel = options.onCancel;
 
     /**
+     * Gets or sets the details of a custom user interface for this map interaction mode. This property is not used by
+     * the `MapInteractionMode` itself, so it can be anything that is suitable for the user interface. In the standard
+     * React-based user interface included with TerriaJS, this property is a function that is called with no parameters
+     * and is expected to return a React component.  @type {Any}
+     */
+    this.customUi = defaultValue(options.customUi, function() { return undefined; });
```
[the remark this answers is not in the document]
rdengate: Thanks, changed.

lib/Models/RegionParameter.js
rdengate: Argh! This used to fix the problem described, but you're right, that now works without my change. I've removed my change. Note that it's still comparing Australia to australia and returning no match... but the country is still selected, so I guess there's something that has changed elsewhere.

kring (approved):

lib/Models/RegionParameter.js
rdengate: Is this with region type COUNTRY?

PR #2182, opened 2016-10-19
Reverse the order of items in ArcGIS legends to match the order used by ArcGIS itself

### LINKED ISSUE #1692: MapServer autolegend puts entries in wrong order
When we let Terria create a legend automatically from an ArcGIS MapServer the entries in the legend come out in the wrong order:

screen shot 2016-06-27 at 17 12 01

screen shot 2016-06-27 at 17 12 11

We should correct this so that it's at least in the same order in the Terria legend as the MapServer.

Bonus points if we can connect the continuous range somehow.

kring:
Fixes #1692

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
+### 4.5.1
+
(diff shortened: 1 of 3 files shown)


axman6:
LGTM AFAICT =)

PR #2185, opened 2016-10-21
Make GetCapabilities parsing more lenient.

### LINKED ISSUE #2184: Can't load sci-wms WMS endpoint
I tried loading this sci-wms WMS endpoint:
http://sci-wms.whoi.edu/wms/datasets/necofs_gom3_forecast?service=WMS&request=GetCapabilities&version=1.3.0&tiled=true
into TerriaJS via "Add Data=>My Data=>Web Data=>WMS Server", and it balked with a "Group not Available" error:

I have tested that CORS is enabled on the endpoint, so that's not the problem.  

Is this a bug, or is there is something non-standard in the sci-wms endpoint that terriaJS doesn't like?

kring:
Fixes #2184


rdengate:
Looks fine to me; tested and can confirm it fixes #2184.

Issue #2160, opened 2016-10-10
Search result location Pin should be clamped to terrain


geoffrowland:
Currently, on zooming into a 3D surface the location Pin/Marker floats around in a distracting way.
Not a huge issue, but would be nice if the Pin could be clamped to the terrain surface.

nahidakbar:
Should be closed as it is done.

geoffrowland:
Thanks! Fix works for me

PR #2186, opened 2016-10-21
Clamp search results billboard (pin) to terrain.

nahidakbar:
#2160


nahidakbar:
should be ready @kring

kring:
Nice, thanks @nahidakbar!

PR #2187, opened 2016-10-22
Default clipToRectangle to true.

kring:
This results in _much_ better performance for some layers, such as the time-dynamic Chlorophyll concentration layer in `#test`.  The downside is that, if a server doesn't account for the size of a graphical representation of a feature when reporting the bounds, features (e.g. points) may be cut off at the edges.  I don't know of any servers that have this problem anymore, so I think changing the default to favor performance in the much more common case is a wise choice.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ Change Log
 ### 4.6.0
 
 * Change in defaults:
-  * The `clipToRectangle` property of raster catalog items (`WebMapServiceCatalogItem`, `ArcGisMapServerCatalogItem`, etc.) now defaults to `true`.  It was `false` in previous releases.  Using `false` prevents features (especially point features) right at the edge of the layer's rectangle from being cut off when the server reports too tight a rectangle, but also causes the layer to load much more slowly in many cases.  Starting in this version, we favor performance and the much more common case that the rectangle can be trusted.
+  * The `clipToRectangle` property of raster catalog items (`WebMapServiceCatalogItem`, `ArcGisMapServerCatalogItem`, etc.) now defaults to `true`.  It was `false` in previous releases.  Using `false` prevents features (especially point features) right at the edge of the layer's rectangle from being cut off when the server reports too tight a rectangle, but also causes the layer to load much more slowly in many cases.  Starting in this version, we favour performance and the much more common case that the rectangle can be trusted.
 * Add support for 'Long' type hint to CSV data for specifying longitude.
 
 ### 4.5.1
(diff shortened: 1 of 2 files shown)


kring:
@rsignell-usgs this should drastically improve the performance of some of the WMS servers you've been working with.

rsignell-usgs:
I modified `package.json` to 

```
    "terriajs": "git://github.com/TerriaJS/terriajs.git#clipToRectangle",
```

then did:

```
 npm stop
 gulp sync-terriajs-dependencies
 npm install
 gulp
 npm start
```

I added this WMS endpoint: 
http://www.smast.umassd.edu:8080/ncWMS2/wms?SERVICE=WMS&REQUEST=GetCapabilities&VERSION=1.3.0

I switched to 2D mode, and requested the `u:v group`, which plots vectors.  There are indeed many fewer requests (only order 30, which is great)!

I did get the following error once when I zoomed in but perhaps it has nothing to do with this PR, and I couldn't repeat it.   Just including it FYI...

```
destroyObject.js:45 Uncaught DeveloperError: This object was destroyed, i.e., destroy() was called.
Error
    at new DeveloperError (http://gamone.whoi.edu/terriajs/build/TerriaMap.js:599:19)
    at new throwOnDestroyed (http://gamone.whoi.edu/terriajs/build/TerriaMap.js:47970:19)
    at throwOnDestroyed.<anonymous> (http://gamone.whoi.edu/terriajs/build/TerriaMap.js:220229:14)
    at handleMouseUp (http://gamone.whoi.edu/terriajs/build/TerriaMap.js:217774:21)
    at HTMLDocument.listener (http://gamone.whoi.edu/terriajs/build/TerriaMap.js:217618:13)throwOnDestroyed @ destroyObject.js:45(anonymous function) @ Cesium.js:88handleMouseUp @ ScreenSpaceEventHandler.js:224listener @ ScreenSpaceEventHandler.js:68
TerriaViewer.js:460 L.LatLngBounds {_southWest: L.LatLng, _northEast: L.LatLng}
TerriaViewer.js:460 L.LatLngBounds {_southWest: L.LatLng, _northEast: L.LatLng}
TerriaViewer.js:460 L.LatLngBounds {_southWest: L.LatLng, _northEast: L.LatLng}
```

kring:
I'm not sure what to make of that error, @rsignell-usgs.  Definitely let me know if you see it again, especially if you can make find a way to reproduce it.

rdengate:
I wasn't able to reproduce the zoom error either. This looks fine to me.

PR #2190, opened 2016-10-25
Fix region selection name box when region ID is non-numeric.

kring:
- Select the `Regions like this` function from Terria Analytics.
- Click `Selection region`
- Choose `COUNTRY`
- Click on Australia
- Before this PR, Australia would be selected but its name wouldn't show up in the name box.


kring:
I should mention that this was originally in #2170, but I (incorrectly) convinced @rdengate to remove it before I merged it.

rdengate:
Tested, and this looks good. Thanks @kring!

PR #2192, opened 2016-10-31
Uber WPS support

rdengate:
These are the changes required so that a Terria instance (e.g., GEOGLAM) is able to add a custom WPS parameter that prompts the user to select a point, draw a polygon, or select a region on the map. The result of the user defining location is always a geojson feature.


lib/Models/Feature.js
```diff
@@ -92,7 +92,7 @@ Feature.fromEntityCollectionOrEntity = function(entity) {
     if (entity.entityCollection) {
         feature = entity.entityCollection.getById(entity.id);
     }
-    if (!feature || !feature instanceof Feature) {
+    if (!feature || !(feature instanceof Feature)) {
```
kring, line 95: Oh good catch.  It probably means that we weren't correctly creating a `Feature` for something that started out as just an `Entity` though.  @RacingTadpole is there anything in particular we should test to make sure that doesn't have unintended side effects?

RacingTadpole: Yes, very nice catch. It would be good to check the feature info panel for time-varying features (eg. in #test and #csv), and with some of the #polling test cases too.

rdengate: Thanks... I only did it because lint complained, actually.

lib/Models/FunctionParameter.js
```diff
@@ -204,8 +204,22 @@ FunctionParameter.prototype.load = function() {
...
 FunctionParameter.prototype.formatValueAsString = function(value) {
     return defined(value) ? value.toString() : '-';
 };
 
+/**
+ * Returns a GeoJSON feature that represents value, if it makes sense to do so.
+ * @param {Object} value Value to get GeoJSON feature for.
+ * @return {Object} JSON feature or undefined.
+ */
+FunctionParameter.prototype.getGeoJsonFeature = function(value) {
```
kring: I'm not a huge fan of having this here.  It'd be better to just make it an optional method.  You could introduce an intermediate base class, like `GeometryFunctionParameter` if you want.

lib/Models/LineParameter.js
```diff
@@ -64,4 +64,15 @@ LineParameter.prototype.formatValueAsString = function(value) {
     }
 };
 
+LineParameter.prototype.getGeoJsonFeature = function(value) {
```
kring, line 79: Strange that this is an instance method but it doesn't ever use `this`.  Should it at least assume the current value of the parameter if one isn't supplied?

lib/Models/WebProcessingServiceCatalogItem.js
```diff
@@ -105,194 +99,162 @@ WebProcessingServiceCatalogItem.prototype._getValuesThatInfluenceLoad = function
     return [this.wpsResponseUrl, this.wpsResponse];
 };
 
+WebProcessingServiceCatalogItem.prototype._getGeojson = function() {
```
kring: If this is never meant to be called from outside this class, you can make it a simple function that takes the catalog item as its first parameter.  Generally a function on the prototype, named with an underscore, should be reserved for a "protected" method.  i.e. one that's meant to be called from base or derived classes but not from outside the class.

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -478,34 +479,38 @@ WebProcessingServiceCatalogFunction.prototype.invoke = function() {
...
 
-    var uri = new URI(this.url).query({
-        service: 'WPS',
-        request: 'Execute',
-        version: '1.0.0',
-        Identifier: this.identifier,
-        DataInputs: createWpsDataInputsFromParameters(this)
-    });
+    var that = this;
+    when.all(createWpsDataInputsFromParameters(this)).then(function(dataInputs) {
```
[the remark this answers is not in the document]
rdengate: It's in Geoglam's GeoJsonParameter, which is being added to WebProcessingServiceCatalogFunction's parameterConverters in the index file of Geoglam.

lib/ReactViews/Analytics/ParameterEditor.jsx
```diff
@@ -27,6 +27,177 @@ const ParameterEditor = React.createClass({
         previewed: React.PropTypes.object
     },
 
+    parameterTypeConverters: [
```
kring: Does this actually work?  We want this property to be static, so that it can be modified by an application without a specific instance handy.  I suppose you can do that by accessing `ParameterEditor.prototype.parameterTypeConverters` but that's ugly.  Probably better to remove this from here and instead add `ParameterEditor.parameterTypeConverters = ...` just before the return at the end of this file.

lib/ReactViews/Analytics/ParameterEditor.jsx
```diff
@@ -27,6 +27,177 @@ const ParameterEditor = React.createClass({
...
+                                    previewed={parameterEditor.props.previewed}
+                                    parameter={parameterEditor.props.parameter}
+                                />
+                            </div>);
+                }
+            }
+        },
+        {
+            id: 'region',
+            parameterTypeToDiv: function ParameterTypeToDiv(type, parameterEditor) {
```
kring: The generic name `ParameterTypeToDiv` makes eslint happy, but it'd be better to use a more descriptive name, like `RegionParameterToDiv`.  That way the more useful name will show up in the debugger.

lib/ReactViews/Analytics/ParameterEditor.jsx
```diff
@@ -37,124 +208,15 @@ const ParameterEditor = React.createClass({
...
-                return (
-                    <div>
-                        {this.renderLabel()}
-                        <GenericParameterEditor
-                            previewed={this.props.previewed}
-                            parameter={this.props.parameter}
-                        />
-                    </div>);
         }
+        const genericEditor = this.parameterTypeConverters.find(function(item) { return item.id === 'generic'; });
```
kring: `find` isn't available on Internet Explorer.  There's a `findIndex` function in TerriaJS that you can use instead, or you can do:
`this.parameterTypeConverters.filter(function(item) { return item.id === 'generic'; })[0];`

lib/ReactViews/Analytics/PointParameterEditor.jsx
```diff
@@ -20,61 +18,45 @@ const PointParameterEditor = React.createClass({
         viewState: React.PropTypes.object
     },
 
+    componentWillMount() {
+        if (!defined(this.pointParameterEditorCore)) {
+            this.pointParameterEditorCore = new PointParameterEditorCore(this.props.previewed,
```
kring: This won't work.  If the component receives new props, the `Core` component won't get them.  You can implement more React lifecycle methods, but there's probably a more Reacty way to do this.  Let's talk about it.

kring: I should say, it very well may work in given how we're using this component currently.. but it's still not a reliable thing to do.

lib/ReactViews/Analytics/RegionPicker.jsx
```diff
@@ -256,19 +265,27 @@ const RegionPicker = React.createClass({
...
-            return region.id;
+            val = region.id;
         }
+        return this.regionProvider.regionType + ": " + val;
     },
 
     render() {
+        let param;
+        if (defined(this.props.parameter.regionParameter)) {
+            param = this.props.parameter.regionParameter.regionTypeParameter;
```
kring: Same thing here, just expose a `regionTypeParameter` on your parameter.

kring:
That's everything from me @rdengate!

rdengate:
Thanks @kring! I will address the comments today :-)

lib/Models/LineParameter.js
rdengate: Yeah, good idea. Changed in *Parameter.js classes where applicable.

lib/Models/WebProcessingServiceCatalogItem.js
rdengate: Thanks, that makes sense. Changed.

lib/ReactViews/Analytics/ParameterEditor.jsx
rdengate: Yeah, I was accessing via the prototype. Better now!

lib/ReactViews/Analytics/ParameterEditor.jsx
rdengate: Arghh IE. Thanks, changed.

lib/ReactViews/Analytics/RegionPicker.jsx
rdengate: Thanks, changed.

rdengate:
I have made most of the changes suggested but have yet to look at the componentWillMount issue. Will discuss Monday.

rdengate:
@kring I have made changes as discussed, so this is ready for the next round.

lib/Models/FunctionParameter.js
```diff
@@ -59,6 +59,12 @@ var FunctionParameter = function(options) {
     this.name = defaultValue(options.name, options.id);
 
     /**
+     * Gets or sets the display value of the parameter.
+     * @type {String}
+     */
+    this.displayValue = defaultValue(options.displayValue, '');
```
kring: Shouldn't this be computed from the actual value, rather than a separate property?  It's ok if it's settable, but the setter should set the underlying value.

lib/Models/FunctionParameter.js
```diff
@@ -204,6 +210,11 @@ FunctionParameter.prototype.load = function() {
     return this._loadingPromise;
 };
 
+/**
+ * Represents value as string.
+ * @param {Object} value Value to format as string.
+ * @return {String} String representing value.
+ */
 FunctionParameter.prototype.formatValueAsString = function(value) {
     return defined(value) ? value.toString() : '-';
```
kring, line 214: This should use the parameter's own value if one isn't passed in, just like you did in the derived parameter types.

lib/Models/PointParameter.js
```diff
@@ -47,12 +47,37 @@ defineProperties(PointParameter.prototype, {
 });
 
 PointParameter.prototype.formatValueAsString = function(value) {
-    if (!defined(value)) {
+    var val = defined(value) ? value : this.value;
```
kring: A bit of a nitpick, but you can simplify this to `var val = defaultValue(value, this.value);`  or even `value = defaultValue(value, this.value);`

lib/Models/WebProcessingServiceCatalogItem.js
```diff
@@ -339,4 +275,32 @@ function formatOutputValue(title, value) {
     }, '');
 }
 
+function getGeoJson(catalogItem) {
+    if (!defined(catalogItem.parameters) || !defined(catalogItem.parameterValues)) {
+        return;
```
kring: It's equivalent, but `return undefined;` would be clearer here.

kring: Actually, it might be better to `return when(undefined);`.  That way the function always returns a promise, and that promise may or may not resolve to undefined, rather than needing to check for undefined in two places.  You could remove the `when` at the call site.

rdengate:
Please don't merge yet; I still have some changes to make. Thanks!

rdengate:
@kring this is good to go, thanks.

Issue #2127, opened 2016-09-27
Remember user's choice of 2D / 3D Flat / 3D terrain


kring:
Currently we persist the base map selection in local storage, but not the viewer mode.

PR #2194, opened 2016-11-01
make it so that zoom does not crush the terrain #2127

nahidakbar:


lib/ReactViews/Map/Navigation/ZoomControl.jsx
```diff
@@ -86,6 +86,19 @@ const ZoomControl = React.createClass({
             const direction = Cartesian3.subtract(focus, camera.position, cartesian3Scratch);
             const movementVector = Cartesian3.multiplyByScalar(direction, 2.0 / 3.0, cartesian3Scratch);
             const endPosition = Cartesian3.add(camera.position, movementVector, cartesian3Scratch);
+
+            // make sure that we don't crush the terrain through
+            const pickRay = new Ray(scene.camera.positionWC, scene.camera.directionWC);
+            const globe = scene.globe;
+            const surfacePosition = globe.pick(pickRay, scene);
```
kring, line 93: `globe.pick` will return undefined if the ray doesn't insersect the globe (i.e. if you tip the camera so that the globe horizon is below the center of the screen).  We need to make sure we don't crash in that scenario.

lib/ViewModels/TerriaViewer.js
```diff
@@ -315,6 +315,12 @@ us via email at '+that.terria.supportEmail+'.'
 
     var scene = viewer.scene;
 
+    // Set minimum zoom distance to 25 meters. Default is 1m.
+    // Zoom crashes through terrain if the rate of zoom is higher than terrain
+    // resolution rate and the height at the more detailed LOD is higher than
+    // previous LOD.
+    scene.screenSpaceCameraController.minimumZoomDistance = 25.0;
```
kring, line 322: The Cesium camera already has a way to deal with this problem when handling mouse input.  I think the problem is that we're just circumventing it by calling `flyToPosition`.  Can you take a look at how it does it (probably in `ScreenSpaceCameraController`) to see if we can do something similar to avoid adjusting the minimum distance?

kring: @nahidakbar I haven't seen that with the mouse while or with right-click dragging.  If it happens, I'd consider it it bug in Cesium.

kring: Ok if I zoom in really fast to a mountainous place, I can get some weird views (e.g. see stars) but zooming back out a bit always puts me where I think I should be.  I don't see the teleportation.

nahidakbar: Yeah I see that when I don't scroll fast enough. Magical teleportation can be accessed if you scroll faster still and keep scrolling. Do you want me to remove this threshold? FlyTo problem is catered for by changes to ZoomControl.jsx alone.

kring: Yeah, please do.  I don't think we want to impose a 25m limit.  It would be great to actually fix the underlying bug though.

kring:
Closing because of no movement in over a year, and I'm not sure the approach (set the minimum zoom distance to a larger value) is the right solution.

PR #2206, opened 2016-11-11
Added hover text that shows the position of data catalog members

### LINKED ISSUE #2136: Can't see where catalog item lives in catalog
I was looking for an item in AREMI. After finding it by search, there is not a way to know how to find it in the hierarchy:

screenshot 2016-09-29 14 41 40

Previously there was a way (hover over, maybe?) to see where it actually fits in the hierarchy.

RacingTadpole:
Fixes #2136.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Change Log
 * The marker indicating the location of a search result is now placed correctly on the terrain surface.
 * Region for CatalogFunctions now selected on map rather than preview map
 * Some regions that were previously not selectable (in Analytics) except via autocomplete are now selectable.
+* Added hover text that show the position of data catalog search results in the full catalog.
 
 ### 4.5.1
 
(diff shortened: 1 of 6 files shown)


kring:
I'll mention a couple of things for your consideration, even though I'm going to merge this anyway.
1. The search results already contain a tooltip with a path for each search result.  It also handles the possibility that a catalog item exists in two groups (which used to be allowed and happened in the Victorian Government group, but I'm not sure if it's still possible).
2. The implementation of `getAncestors` is pretty inefficient.  It does lots of heap allocations each time it is called, and the recursion is likely to be slow in JavaScript, especially since it's not even tail recursion.  And since it's called for every item and group, we might actually feel the pain in a big catalog.  It might be worth using a simple iterative implementation, which wouldn't really be any more code or any less clear IMO.

RacingTadpole:
Thanks @kring, I've replaced the getAncestors function with a loop implementation now.

PR #2210, opened 2016-11-11
Fix ArcGIS test init file, and improve its error message

RacingTadpole:
`CatalogGroup` instances do not have `nowViewingCatalogItem`, `dataCustodian` or `url` properties. So we might as well get rid of the code that checks for them and displays them in the preview pane.

I may have misunderstood something here, so please check this.

Also adds a `description` property (and fixes the URLs) to `#build/terriajs/test/init/arcgis.json` so you can at least see a description. (Adding `url` or `dataCustodian` doesn't work, because they never get put on the instance, which is how I found this issue.)

Also, since the arcgis test group I happened to test this on had some links that weren't working, I took the opportunity to update them, and improve the error handling of ArcGis items (try the "bad link" item to see it). (That really ought to be a separate PR, but these are both so tiny, I've combined them.)

### DIFF
--- a/lib/Models/ArcGisMapServerCatalogItem.js
+++ b/lib/Models/ArcGisMapServerCatalogItem.js
@@ -222,11 +222,11 @@ ArcGisMapServerCatalogItem.prototype._load = function() {
                 that.updateFromMetadata(results[0], {"layers": [results[1]]}, results[2], false, results[1]);
             } else {
                 var message = defined(results[0].error) ? results[0].error.message : 'This dataset returned unusable metadata.';
-                that.terria.error.raiseEvent(new TerriaError({
+                throw new TerriaError({
                     title: 'ArcGIS Mapserver Error',
                     message: '<p>' + message + '</p><p>Please report it by \
 sending an email to <a href="mailto:' + that.terria.supportEmail + '">' + that.terria.supportEmail + '</a>.</p>'
-                }));
+                });
             }
         });
     }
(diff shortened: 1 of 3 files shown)


kring:
`CatalogGroup` itself doesn't have all these properties, but some of them, at least, are sensible on derived group types.  For example, `CkanCatalogGroup` has a `url` and I think `dataCustodian` is sensible for an `ArcGisCatalogGroup`.  And all groups have `info` sections, via `CatalogMember`.  We definitely don't want to get rid of that.

RacingTadpole:
OK, thanks. I'll revert that change and just leave the improvements to arcgis in this PR.

RacingTadpole:
OK, ready for review again - thanks :-)

lib/Models/ArcGisMapServerCatalogItem.js
```diff
@@ -222,8 +221,12 @@ ArcGisMapServerCatalogItem.prototype._load = function() {
                 // Results of a single layer query. Make it look like a multi layer query result.
                 that.updateFromMetadata(results[0], {"layers": [results[1]]}, results[2], false, results[1]);
             } else {
-                console.error("Unusable ArcGIS Mapserver layers metadata: \n" + layersMetadata);
-                throw new RuntimeError('Unusable ArcGIS Mapserver layers metadata from ' + layersUri.toString() + '. See console log for details.');
+                var message = defined(results[0].error) ? results[0].error.message : 'This dataset returned unusable metadata.';
+                that.terria.error.raiseEvent(new TerriaError({
```
kring: `load` should usually return errors via promise rejection, rather than raising `terria.error` and returning success.

RacingTadpole: oops, thanks. Fixed.

PR #2214, opened 2016-11-13
Add data and service metadata to "About this Dataset"

RacingTadpole:
Fixes https://github.com/TerriaJS/neii-viewer/issues/50. Try it out on WMS and ArcGIS items.
I had this mostly done before the last comment on that issue, so here it is anyway :-)

Also makes the collapsible component slightly more general, with variable-defined fill colour and an "isInverse" option to allow for light or dark backgrounds.


RacingTadpole:
Not sure why travis is hitting a new lint warning on `InvokeFunction`, which I haven't touched... I've just added two commits (`9c0e42d` and `22b0f65`) to fix that too, though it doesn't really belong here.

kring:
Gets a bit broken looking on narrow screens:
image

kring:
Another concern I have is that these metadata tables can be enormous on some servers.  Thousands of styles, for example.  In fact, we've seen cases in the past where the UI would hang for 10 seconds to display these tables.  Maybe it's less an issue now that we're using 🐎 React, but it's probably better to be cautious.  One easy thing is to make sure that when the tables are not visible, their DOM elements are not created.  That way, at least the UI won't hang when the preview panel first comes up.  But it looks like `Collapsible` just hides its children with CSS rather than removing them from the DOM entirely.

RacingTadpole:
Fair enough - but I think `collapsible` _is_ removing the child from the DOM if it's closed - see https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Custom/Collapsible/Collapsible.jsx#L24-L30 .
Re small screens - I agree... but what to do? I could re-instate break-words in `th`, but that makes it look terrible on every size screen, I think.

kring:
>  I think collapsible is removing the child from the DOM if it's closed

You're right, not sure what I was looking at.. Sorry!

> Re small screens - I agree... but what to do? I could re-instate break-words in th, but that makes it look terrible on every size screen, I think.

Yeah I'm not sure.. maybe a horizontal scrollbar below a certain size would be an ok compromise?

RacingTadpole:
Sorry... the css to put a horizontal scrollbar here is defeating me - I tried `.root { overflow-x: scroll; width: 800px; }` in `metadata-table.scss`, but that adds a scrollbar to every table sub-element; and adding it to `.metadata` in `mappable-preview.scss` seems to suppress the scrollbar. @chloeleichen any chance you could take a look?

RacingTadpole:
@chloeleichen - do you know if you'll get a chance to look at this in the next week?

chloeleichen:
i'm afraid we will have to enable scrollbar on the outside container. can you have a look now @RacingTadpole

RacingTadpole:
Thanks @chloeleichen - looks good to me!

PR #2217, opened 2016-11-14
More general css, with some small changes to look

RacingTadpole:
Merge #2215 first.

This PR adds more flexibility for changing the look of terria by using css variables.
However it does change a few small things - the main one being the coloring of the location bar at the bottom now matches the coloring of the share buttons at the top - I think for the better, but worth looking over ( @philipgrimmett  ?).

This should be backwards-compatible.


RacingTadpole:
Ready to go again now. Thanks.

PR #2218, opened 2016-11-14
Fix CSW group loading when <references> does not have a protocol attribute

kring:
@rsignell-usgs reported that the TerriaJS reported an exception when trying to load the `CswCatalogGroup`  that I added to `csw.json` in this PR.  The problem was that the code was expecting `<references>` elements to have a `protocol` attribute.  It's surprisingly hard to find an authoritative reference for what attributes are allowed and required on a `<references>` element (the XML namespace links to here http://purl.org/dc/terms/) but I did find other examples of this element with only a `scheme` instead of a `protocol`, so clearly this is something we should support.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Change Log
 * Widened scrollbars and improve their contrast.
 * Removed the default maximum number of 10 results when searching the data catalog.
 * Allow users to browse for json configuration files when adding "Local Data".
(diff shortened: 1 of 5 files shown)


rsignell-usgs:
@kring, woohoo, this works great!  It fixes my issue -- pycsw CSW endpoints work fine now:


And it's good to support pycsw, since that's the CSW that data.gov, and many others use:
https://github.com/geopython/pycsw/wiki/Live-Deployments

PR #2234, opened 2016-11-21
Fix missing WMS legend when URL does not contain GetLegendGraphic

kring:


rsignell-usgs:
Does this fix https://github.com/TerriaJS/terriajs/issues/2201?

kring:
No, I don't think so, sorry.  The symptom of this bug was that the legend didn't appear at all.

PR #2243, opened 2016-11-23
2119 add shareurl to feedback

### LINKED ISSUE #2119: Add share link to feedback issues (with permission checkbox)
Problem: occasionally we receive troubleshooting feedback but can't diagnose the issue without knowing which base map was selected, which data sets were turned on etc.

Proposed solution: add a checkbox option in the feedback form to include a share link with their submission e.g. "[  ] Share my map view with the National Map's developers (this helps us to troubleshoot issues)"

axman6:
Sends the share url to the server for inclusion in github issues - this along with the fixes to terriajs-server should close #2119.


lib/ReactViews/Feedback/FeedbackForm.jsx
```diff
@@ -3,6 +3,7 @@
 import ObserveModelMixin from '../ObserveModelMixin';
 import React from 'react';
 import sendFeedback from '../../Models/sendFeedback.js';
+import shareLink from '..//Map/Panels/SharePanel/BuildShareLink.js';
```
kring: Extra slash, and also I think it's good practice to make the name of the import match the name of the file.  It's a lot less confusing that way.

lib/ReactViews/Feedback/FeedbackForm.jsx
```diff
@@ -17,6 +18,7 @@ const FeedbackForm = React.createClass({
     getInitialState() {
         return {
             isSending: false,
+            sendShareURL: false,
```
kring: Any reason we shouldn't default this to true?

kring:
Just those couple of comments, looks good!  Thanks @axman6!

lib/ReactViews/Feedback/FeedbackForm.jsx
```diff
@@ -51,7 +53,7 @@ const FeedbackForm = React.createClass({
                     this.props.viewState.feedbackFormIsVisible = false;
                 } else {
                     this.setState({
-                        isSending: false
+                        isSending: false,
```
kring: extra comma here

axman6:
One thing that probably needs to be fixed is the CSS for checkboxes, currently the form ends up looking like: 
screen shot 2016-11-23 at 3 10 13 pm

axman6:
Also, currently the phrase "the National Map" is hard coded in, I thought I saw somewhere a setting for the project name but couldn't find it; we should probably use that instead.

kring:
> currently the form ends up looking like

Strangely enough the checkbox is on the same line on my system.  Still, it'd be good to switch to using the checkbox style used elsewhere in TerriaJS.  Example here: https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx#L151

> Also, currently the phrase "the National Map" is hard coded in

Search for `terria.appName`.

axman6:
Looking much better now: 
screen shot 2016-11-23 at 4 43 25 pm

lib/ReactViews/Feedback/FeedbackForm.jsx
```diff
@@ -87,6 +95,13 @@ const FeedbackForm = React.createClass({
                       <input type="text" name="email" className={Styles.field} value={this.state.email} onChange={this.handleChange}/>
                       <label>Comment or question</label>
                       <textarea className={Styles.field} name="comment" value={this.state.comment} onChange={this.handleChange}/>
+                      <div className={classNames(Styles.shareUrl)}>
```
kring: Since you have only one style, you don't need `classNames`.  See `Styles.action` below.

PR #2246, opened 2016-11-27
Allow user to specify colours for csv data in WPS responses

rdengate:
If a json/csv ComplexData contains 'color' as an attribute (which is a comma separated list of css colors), these colours are applied to the columns of the csv. It allows the user to override the default behaviour of picking a colour that hasn't been used from a list of possible colours.

To test: 
1. With this change: https://github.com/TerriaJS/nationalmap/pull/373
2. In #test, Test Data > WPS > Csv Output (if you don't see that, you don't have the change in step 1)
3. Select a point on the map
4. Run analysis
5. Note that the colours for the chart are a strong red, green, blue, whereas previously they'd be orange, light green, purple.


kring:
@RacingTadpole can you review and merge?  The WPS side of this looks fine to me.

lib/Map/VariableConcept.js
```diff
@@ -62,7 +62,14 @@ var VariableConcept = function(name, options) {
...
         // If this concept has been activated, and its parent has a color callback, use it to set this concept's color now.
         if (nowActive && defined(this.parent) && !defined(this.color) && defined(this.parent.getColorCallback)) {
-            this.color = this.parent.getColorCallback();
+            var that = this;
+            var columnId;
+            for (var i=0; i<this.parent.columns.length; i++) {
+                if (that.id === this.parent.columns[i].id) {
+                    columnId = i;
+                }
+            }
```
RacingTadpole: I'd prefer not to have this object loop through its siblings in this way.  Not sure what the right way is though - let me think about it for a bit.

lib/ReactViews/Custom/registerCustomComponentTypes.js
```diff
@@ -148,6 +150,15 @@ const registerCustomComponentTypes = function(terria) {
         }
         const url = defined(sources) ? sources[0] : undefined;
         const tableStructure = tableStructureFromStringData(getSourceData(node, children));
+
+        const colors = splitStringIfDefined(node.attribs['colors']);
+        if (defined(colors)) {
+            const getColorCallback = function(columnId) {
+                return colors[columnId];
+            };
+            tableStructure.getColorCallback = getColorCallback;
```
RacingTadpole: This new attribute `colors` only works if there is a `tableStructure`, ie. if actual data is provided in the chart tag. We should at least note that. But it would be even better if there's a way to make it work more generally.

RacingTadpole: Minor suggestion - the new functional format would be more succinct - eg. `tableStructure.getColorCallback = index => colors[index];`. (It's not really a column id here - it's an index, right?)

RacingTadpole:
Just thinking about how to resolve the two slight concerns I have (looping over cousins in the Concepts, and having the color tag attribute only work for some situations):

There are two situations where colors need to be provided:
- In the preview panel (when the chart is coming directly from a chart tag). Luckily for this case there is already a `colors` prop available in `Chart.jsx` - so you could just pass it in from `registerCustomComponentTypes`.
- When the chart is produced from a TableCatalogItem. In this case, you could also pass colors to `ChartExpandAndDownloadButtons`, which could detect if it is present when the user clicks Expand. Then when it creates the chart's `CsvCatalogItem` (which derives from `TableCatalogItem`), it could override its `getNextColor` method at that point. It might be nice to pass an `options.colors` to `setChartable` which it could use to do this, rather than doing it directly.

What do you think?

RacingTadpole:
@rdengate commented elsewhere:
> But... I want the colours only to be used when expanded, and to keep using white in the preview.
> It was kind of a thick line and there's no legend, so it wasn't meaningful. I think the light grey looks neater.
> The preview is so small that you can't get a lot of info from it anyway.

OK, we could make it a blanket rule that the colors get passed through without being used in the preview. Because your comment would be true always.
If we ever want the preview to be colored as well, we can add another prop later, eg. `preview-colors`.

RacingTadpole:
Just to check we’re on the same page - I’m imagining it that the result will be 
• little or no change (compared to master) in `VariableConcept`,
• that `registerCustomComponentTypes` will just pass colors through to the two react components (without attaching it to the table structure),
• `Chart.jsx` won’t have the line ```color: defined(this.props.colors) ? this.props.colors[index] : undefined``` (odd - rereading the code I’m not sure how TableStructure colors ever get on the chart!);
• and you’ll add some stuff to the `expand` function of `ChartExpandAndDownloadButtons`
?
Thanks!

rdengate:
Thanks @RacingTadpole! Hopefully that addresses all your comments.

RacingTadpole:
Yep, that looks good to me! I haven't actually tried it out yet though, so I'll do that.
Could I ask one thing - to make it easier to test, could you put an example of `<chart>` with a `colors` attribute into `wwwroot/test/init/charts.json`?

rdengate:
Sure -- the test in WPS is not sufficient?

RacingTadpole:
I just thought a direct test that `colors` works in a `chart` tag would be good too - assuming it's meant to?

rdengate:
Fair enough, good idea. I will add a test.

rdengate:
Test added :-)

RacingTadpole:
Cool - are these the colours I am meant to see (black and blue/red/green)?
image


Also - do you want to remove the hover text in the workbench, ie:
image

rdengate:
Yes to both questions.

RacingTadpole:
I'm just looking at how to suppress the tooltip - in `Chart.jsx` it hides it if `this.props.styling === 'feature-info'` or if `defined(this.props.highlightX)`. Would the former work here or are you already passing a different `styling` parameter?

rdengate:
I'm passing styling=histogram. There was some reason to do that, but I can't remember what it was...

RacingTadpole:
maybe just try changing it to feature-info?

rdengate:
Thanks! That works. I might leave that for the next PR, though. There are going to be more chart tweaks soon.

RacingTadpole:
Last thing - can you add to CHANGES.md?

rdengate:
Done. Thanks!

PR #2254, opened 2016-12-04
Fix name of shareurl in feedback message to server

axman6:
Fixes bug where the share url was sent using a different key than the server was expecting.


RacingTadpole:
Incorporated into #2256.

PR #2259, opened 2016-12-05
Elevation function added

rdengate:
This allows the user to draw a line on the map and see the elevation (height above sea level) charted along that line. If water table data is available that will also be charted. This service is consumed by http://www.ga.gov.au/elvis/ but we've talked about adding it to terria and AREMI for a while now.

This also has some chart tweaks: don't allow LON, LAT, or ALT in charts (and change ALT to SCALAR if ALT is found), and hide x axis column so x isn't charted against itself.

To test, add this to an init json and run:
```
{
      "name": "Elevation Tools",
      "type": "group",
      "preserveOrder": true,
      "items": [
        {
          "name": "Elevation",
          "type": "elevation-function",
          "url": "http://www.ga.gov.au/elvis/service/path/"
        }
      ]
}
```


kring:
@rdengate very cool!  It needs a bit more doc around when this is applicable and how to use it.  Maybe a more specific name would be appropriate, too.  It's not a general `ElevationFunction`, it's the Elvis elevation service, right?

lib/Models/ElevationCatalogItem.js
```diff
@@ -0,0 +1,237 @@
...
+            '</tr>' +
+            '<tr>' +
+                '<td style="vertical-align: middle">distance</td>' +
+                '<td style="padding-left: 4px">' + this.parameters[0].getLineDistance() + '</td>' +
+            '</tr>' +
+            '<tr>' +
+                '<td style="vertical-align: middle">count</td>' +
+                '<td style="padding-left: 4px">500</td>' +
+            '</tr>' +
+        '</table>';
```
RacingTadpole: One thought that occurs to me here, to separate html out of the code - could you load this as a mustache template, eg. see https://github.com/TerriaJS/terriajs/blob/master/lib/Models/SensorObservationServiceCatalogItem.js#L344-L345

lib/Models/ElevationCatalogItem.js
```diff
@@ -0,0 +1,237 @@
...
+
+    var outputsSection =
+        '<table class="cesium-infoBox-defaultTable">' +
+            '<tr>' +
+                '<td style="vertical-align: middle">Path Elevation</td>' +
+                '<td><chart data=\'' + distanceVsElevationCsv + '\' styling: "feature-info"' + chartOptions + '></td>' +
+            '</tr>' +
+        '</table>';
```
RacingTadpole: and here too

lib/Models/ElevationFunction.js
```diff
@@ -0,0 +1,170 @@
...
+var CatalogFunction = require('./CatalogFunction');
+var defineProperties = require('terriajs-cesium/Source/Core/defineProperties');
+var ResultPendingCatalogItem = require('./ResultPendingCatalogItem');
+var loadWithXhr = require('terriajs-cesium/Source/Core/loadWithXhr');
+var inherit = require('../Core/inherit');
+var LineParameter = require('./LineParameter');
+var ElevationCatalogItem = require('./ElevationCatalogItem');
+var sprintf = require('terriajs-cesium/Source/ThirdParty/sprintf');
+
+var ElevationFunction = function(terria) {
```
RacingTadpole: yes, per Kevin's comment - can you add a doc string?

kring:
@rdengate I'm not able to get this to work, but it could just be a problem on the server.  It's returning:
image

RacingTadpole:
Yep, the changes on the charting side look good to me - thanks!

rdengate:
Thanks for the code review @kring and @RacingTadpole! I have made all changes suggested. I'm talking to GA about the elvis service as it seems to work all the time on their site, but only intermittently on terria (for the same request).

kring:
@meh9 is this still a thing?

meh9:
Well, I'd like it to be, but I'm thinking now that it's too specific a feature, and only nerds like me want it. Closed (for now).

PR #2261, opened 2016-12-07
GNAF API update

rdengate:
GNAF API updated to new Lucene-based backend, which improves performance. Once this change is in, the performance is sufficient to allow NM to have geocoding for addresses in csv files, so I'll create a PR for that after this is in.


kring:
Code changes look ok, but I don't think the geocoder is working well.  For example, here's current production NationalMap:

image

Looks good!  Now here's a local build with this change:

image

Not so good.

kring:
Oh wait, that's just because my map view was different in the local instance.  I'm not sure only showing results in NSW when I explicitly say ACT just because the map is looking at Redfern is the right thing, but there's no regression here anyway.  So I'll merge this.

PR #2263, opened 2016-12-08
Add help icon

rdengate:
Required for SoE.


RacingTadpole:
To help with reviewing this one - you might be interested that you can view the icons, if you know their names, at eg.
http://nationalmap.gov.au/build/TerriaJS/images/icons/left.svg

kring:
Unless we're going to use this icon in TerriaJS, it shouldn't be added to TerriaJS.  Just add it to SoE, and instead of specifying the icon as `Icon.GLYPHS.help`, require it in at the top of whatever file uses it:

```
var helpIcon = require('../wwwroot/images/icons/help.svg');
```

And then reference it using the required name:

```
<Icon glyph={helpIcon}/>
```

If that doesn't work, a small change to the WebPack config might be required.  Let me know.

rdengate:
Ok, thanks Kevin.

PR #2273, opened 2016-12-13
Help screens for SoE

rdengate:
This supports creating help screens which Phil designed for State of Environment. The actual creation of help screens is done in the State of Environment instance.

### DIFF
--- a/lib/Models/Terria.js
+++ b/lib/Models/Terria.js
@@ -264,9 +264,9 @@ var Terria = function(options) {
     /**
      * Gets or sets the help screen, a screen which can be used alone or in series to explain a concept in terria in
      * context.
-     * @type {HelpScreenData[]}
+     * @type {HelpScreen[]}
      */
-    this.helpScreen = [];
+    this.helpScreens = [];
 
     /**
      * Gets or sets the catalog of geospatial data.
@@ -352,7 +352,7 @@ var Terria = function(options) {
     this.selectBox = false;
 
     // TODO: Why track showTimeline, which doesn't appear anywhere in the code (in this file or any other)?
-    knockout.track(this, ['viewerMode', 'baseMap', 'baseMapName', 'fogSettings', '_initialView', 'homeView', 'pickedFeatures', 'selectedFeature', 'mapInteractionModeStack', 'helpScreen', 'configParameters', 'showTimeline', 'catalog', 'selectBox']);
+    knockout.track(this, ['viewerMode', 'baseMap', 'baseMapName', 'fogSettings', '_initialView', 'homeView', 'pickedFeatures', 'selectedFeature', 'mapInteractionModeStack', 'helpScreens', 'configParameters', 'showTimeline', 'catalog', 'selectBox']);
 
     /**
      * Gets or sets the camera's initial view.  This is the view that the application has at startup.  If this property
--- a/lib/Models/Terria.js
+++ b/lib/Models/Terria.js
@@ -262,6 +262,13 @@ var Terria = function(options) {
     this.mapInteractionModeStack = [];
 
     /**
+     * Gets or sets the help screen, a screen which can be used alone or in series to explain a concept in terria in
+     * context.
+     * @type {HelpScreenData[]}
+     */
+    this.helpScreen = [];
+
+    /**
      * Gets or sets the catalog of geospatial data.
      * @type {Catalog}
      */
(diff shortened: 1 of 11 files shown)


lib/ReactViewModels/ViewState.js
```diff
@@ -96,7 +98,9 @@ export default class ViewState {
             'myDataIsUploadView',
             'mobileMenuVisible',
             'panelVisible',
-            'explorerPanelAnimating'
+            'explorerPanelAnimating',
+            'highlightedComponentId',
+            'highlightedComponent'
```
RacingTadpole: typo - missing final `s` :-)

lib/Models/Terria.js
```diff
@@ -262,6 +262,13 @@ var Terria = function(options) {
     this.mapInteractionModeStack = [];
 
     /**
+     * Gets or sets the help screen, a screen which can be used alone or in series to explain a concept in terria in
+     * context.
+     * @type {HelpScreenData[]}
+     */
+    this.helpScreen = [];
```
RacingTadpole: A very minor point, but I think it's good practice to give arrays plural names, eg. `helpScreens` :-)
A related even more minor point - I like to keep the names of the instance the same as the names of the class, but with the first letter lower case - although this is going to get messy with the plural idea - eg. `helpScreenDatas` - or maybe this is really telling us that `HelpScreenData` should be simply `HelpScreen`...

lib/ReactViews/Notification/HelpScreenWindow.jsx
```diff
@@ -0,0 +1,46 @@
...
+    },
+
+    render() {
+        const helpScreen = this.props.terria.helpScreen && this.props.terria.helpScreen[this.props.terria.helpScreen.length - 1];
+        const windowClass = classNames(Styles.window, {
+            [Styles.isActive]: helpScreen
+        });
+        const buttonText = helpScreen && helpScreen.totalNumberOfScreens === helpScreen.currentScreenNumber ? 'DONE' : 'NEXT';
+        const positionLeft = helpScreen && helpScreen.left;
+        const positionTop = helpScreen && helpScreen.top;
```
RacingTadpole, line 24: Just looking at the code I'm trying to find where `helpScreen.left` and `top` get set, but can't see it - what am I missing?

RacingTadpole:
@kring - feel free to make comments on this one too :-)

lib/ReactViews/Notification/HelpScreenWindow.jsx
rdengate: It's in SOE. The help is specific to SOE, so I thought I should only use the help screens in SOE.

RacingTadpole: Ah right. What's the best way to test this do you think? Fire up an SOE instance? (Do you have a link to the repo?)

rdengate: Sure -- it's https://github.com/TerriaJS/soe. Private repo, but I've added you. You'll need rd_help_screens branch. Note only the first help item works; I wanted to make sure my approach was okay before fully implementing help. Alternatively you can access my running instance if you want... let me know.

kring:
This is a super cool feature @rdengate, so thanks for working on it!  Unfortunately I'm going to ask for some fairly big changes, though.

I have three main criticisms:
* There's a bunch of code in SoE in that is generally needed for managing help, like turning overlays on and off.
* It requires changes to React components in TerriaJS in order to support highlighting of those components in app-specific help.
* It does direct manipulation of the DOM, which is ill-advised in a React app.

Fixing these problems is a bit tricky, but @AlexGilleran and I talked about it and I think we came up with something workable.  Here's the idea:

* The component to highlight is specified to the help system as a class name.  (the TerriaJS class names are a bit obscure because of our use of CSS modules, but you can use Chrome dev tools to get them and then they are stable across non-breaking TerriaJS versions).
* To highlight an element, we find the first element in the DOM with that class name.  In some cases (React is re-rendering, for example), such a DOM element might not exist.  In that case, we just don't highlight anything.
* Once we know the DOM element we're highlighting, we get its rectangle on the screen using [getBoundingClientRect](https://developer.mozilla.org/en/docs/Web/API/Element/getBoundingClientRect).
* We gray out everything outside that rectangle by creating four overlay elements, like this:

image

The overlays should be normal React components, rather an explicitly-created DOM elements.  Periodically (maybe every 200ms using a `setInterval`), we'll need to look for the first DOM element with that class name again and adjust the overlays accordingly.  This handles the possibility of the DOM element moving around or being re-rendered by React.

This all can actually be done entirely in app code without any changes to TerriaJS, though making this help system a feature of TerriaJS is reasonable.  Ideally everything except that help text and highlight choices would be in TerriaJS and would be easily configurable from an app.

Let me know if any of that makes sense.

kring:
Come to think of it, this could pretty easily be a separate library for doing on-screen help in React apps.  If you want to be famous.

rdengate:
@kring thanks for your comments. I agree with your criticisms so am happy to make big changes. Just so I understand, though, is having a HelpScreenWindow in StandardUserInterface still ok? I'm picturing a PR for terria which has a lot of the changes in this PR, but reverts ViewState.js, SidePanel.jsx and global.scss, although I guess alternatively HelpScreenWindow could be added in NM instance-specific UserInterface, which would move everything out of Terria...

I'm not sure how to make it a separate library. Would a good first step be to implement it in SoE only, and then try to generalise and extract it?

kring:
Unless we're going to have help by default in TerriaJS (maybe someday, but probably not right now), I think we should leave `HelpScreenWindow` out of `StandardUserInterface`.  Adding it to the app-specific `UserInterface` makes sense to me.

Making it a separate library means putting it in a separate git repo and publishing it as a separate module on npm.  Then add it to SoE's package.json and require in the bits you use.  Yeah, you can start by putting it on SoE itself if that makes things easier.

rdengate:
Thanks @kring. That makes sense!

rdengate:
This is not going to be merged in current form, so closing PR.

Issue #2090, opened 2016-09-16
Inconsistent close buttons


stevage:
Apparently this week is open season on the designers? :) I have no idea if this matters, but:

![screenshot 2016-09-16 11 44 58](https://cloud.githubusercontent.com/assets/678344/18572987/40560e4e-7c03-11e6-9164-fcdc33ef13e1.png)

![screenshot 2016-09-16 11 44 50](https://cloud.githubusercontent.com/assets/678344/18572988/4080779c-7c03-11e6-8e3e-e4f5c77c8dce.png)

![screenshot 2016-09-16 11 44 46](https://cloud.githubusercontent.com/assets/678344/18572989/40843422-7c03-11e6-9761-3f6bbd5bdef0.png)

![screenshot 2016-09-16 11 44 44](https://cloud.githubusercontent.com/assets/678344/18572990/40a9a7d4-7c03-11e6-887f-5aef8ac83213.png)

RacingTadpole:
And also from memory that "Your system seems to be running slowly, switching to 2D" box has no close X, only an OK button. Would an X as well be helpful? Generally I expect people just want to get rid of the warning quickly.

chloeleichen:
@philipgrimmett

chloeleichen:
i think they should all be X

stevage:
Yep, and allow clicking outside the box to also close it, wherever that makes sense.

philipgrimmett:
Subtle differences for each case:
- Image 1: Dropdown elements generally don't require a close button (clicking away is more common)
- Image 2: Adding/browsing for data is a task to be completed, hence 'done'. Using this wording frames the action, and hopefully gives the user more confidence while toggling between the main map and the catalogue.
- Image 3: I think the 'X' close is appropriate here. When a user open and closes feature info, the states of the map remains unchanged. 'Done' might suggest otherwise.
- Image 4: hmm. We tried to match the '+' add action with a '-' remove action, but looking back at it now I don't think it really works. At the risk of contradicting my point on Image 3, I think an 'X' icon or rubbish bin icon alongside the 'remove' text would be more recognisable for the average user.

Thoughts?

stevage:
Yep. Specifically, yeah a rubbish bin (I didn't think of that) or X would be better for the Remove action. 

And I get your reasoning for Image 2, but there seems something slightly strange about putting the Done up in the top right. It's sort of like combining the "end of workflow" action with the "close window" action and yielding what is IMHO an odd result.

[...]

PR #2280, opened 2016-12-19
Move "Done" button text into css for easier overriding

RacingTadpole:
Makes it easier to address #2090.
Addresses https://github.com/TerriaJS/neii-viewer/issues/44.

Eg. after this PR is merged, you could replace the "Done" button with an "✕" in the client's `globals.scss` via:

    .tjs-explorer-window__btn--close-modal {
      &:before {
        content: '✕' !important;
      }
      color: #EDEDED;
      font-size: $font-size-large !important;
      border: none;
      &:focus, &:hover {
        border: none;
      }
    }

@chloeleichen - is this good practice?

### DIFF
--- a/lib/ReactViews/ExplorerWindow/ExplorerWindow.jsx
+++ b/lib/ReactViews/ExplorerWindow/ExplorerWindow.jsx
@@ -109,7 +109,6 @@ const ExplorerWindow = React.createClass({
                             className={Styles.btnCloseModal}
                             title="Close data panel"
                             data-target="close-modal">
-                        Done
                     </button>
                     <Tabs terria={this.props.terria} viewState={this.props.viewState}/>
                 </div>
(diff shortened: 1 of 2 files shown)


chloeleichen:
I think it's a smart way to configure this. however, i'm not sure the configuration itself should be in CSS or in javascript file? @kring @AlexGilleran what do you guys think?

AlexGilleran:
I'm not a fan - ideally we should be able to move all of the text into one place at some point for easier localisation. Although if we never do that I suppose it doesn't matter...

RacingTadpole:
Yeah, you're right. I guess the real solution here is to decide if we want `✕` or `Done`, and just adopt it. @philipgrimmett - would you support ✕ , as a partial solution to #2090? Or is `Done` better - in which case I'll push back on TerriaJS/neii-viewer#44 ?

philipgrimmett:
My recommendation would be to retain 'Done'. The wording was defined in response to usability testing, where users appeared to stall after turning on a dataset (It wasn't clear to them how they should proceed). We wanted to frame 'adding data' as a task to be completed, not the default state. Not a perfect solution, but I still think it's an improvement on the generic 'X'

screenshot 2016-12-21 11 52 21

PR #2281, opened 2016-12-19
Simplify 'Download Data for this Feature' to 'Download this Data'

RacingTadpole:
Suggested by @rdengate - what do you think? :-)

### DIFF
--- a/lib/ReactViews/FeatureInfo/FeatureInfoDownload.jsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoDownload.jsx
@@ -1,7 +1,10 @@
 import React from 'react';
-import Dropdown from '../Generic/Dropdown';
+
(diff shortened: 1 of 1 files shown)


rdengate:
Philip's idea actually. I am in favour. :-)

kring:
So I thought we changed it from "Download Data" because that terminology implied you were downloading the whole dataset, not just the contents of the feature info panel.  IMO "Download this Data" has the same problem. 

Since the button is only shown when we're looking at "raw" data (not the result of a feature info template), how about "Download this Table"?

RacingTadpole:
@kring - I've renamed this to Download this Table.
Also, I noticed this duplicated `DataUri`'s functionality, so I've stripped out the duplication. (A rare case of reducing the code base size!)
There was also an unused check for data uri compatability, which I've made real now.

PR #2286, opened 2016-12-20
Cancel pending tile requests when removing a layer from the 2D map

### LINKED ISSUE #2285: Leaflet generates and doesn't cancel hundreds of requests when dragging time slider
* Visit `http://localhost:3001/#test` or similar
* Switch to 2D.
* Enable `Test Data -> Temporal -> Chlorophyll-a concentration (monthly)`.
* Open the network tab.
* Drag the time slider around and watch and hundreds or thousands of requests are generated.  These requests appear to be handled in approximately the order they're made, rather than old irrelevant ones being canceled as you drag to a new time.

kring:
Fixes #2285

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
+### 4.7.3
+
+* Canceled pending tile requests when removing a layer from the 2D map.  This should drastically improve the responsiveness when dragging the time slider of a time-dynamic layer in 2D mode.
+
 ### 4.7.2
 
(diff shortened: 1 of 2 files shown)


kring:
@rdengate can you test this in Geoglam and merge if you're happy?

rdengate:
thanks @kring! Will do.

rdengate:
@kring I think we're still seeing the issue. After dragging the timeslider around for a bit, it takes 15-20 seconds for the layer to show in 2D, vs always a second or two in 3D. Although, Pablo said it used to take minutes for him, so this might be an improvement?

PR #2287, opened 2016-12-21
Fix IE11 hang when adding data

### LINKED ISSUE #2275: IE11 can freeze when adding data
We have seen this in both IE11 version 11.0.14 and 11.0.37.

When you add a few datasets (the exact number changes), the UI becomes unresponsive. The cursor continues to change when you move the mouse, but nothing else works.

Particularly noticeable on NEII-viewer (the new ui version), eg. by adding a few items from the very first catalog group, and a few from Environmental Monitoring Sites Register -> Air. See https://github.com/TerriaJS/neii-viewer/issues/55.

But has also been seen on Northern Australia and National Map.

kring:
Fixes #2275

Internet Explorer 11 has a bug under Windows 7, described here:
https://connect.microsoft.com/IE/feedback/details/796745/mouse-events-are-not-delivered-at-all-anymore-when-inside-an-svg-a-use-is-removed-from-the-dom

Basically if you remove an SVG USE element while the mouse is inside it, the browser stops delivering all mouse events.  This makes it appear as if the application is hung, even though keyboard navigation continues to work.  The workaround used here disables the delivery of pointer events to SVG USE elements.  This is ok because we generally handle events on the containing DIV, rather than on the SVG itself.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Change Log
 * Made "Search" and "Add Data" font sizes consistent.
 * Made Feature Info Panel styling more flexible.
 * Fixed a bug that could cause an extra `/` to be added to end of URLs by `ArcGisMapServerCatalogItem`, causing some servers to reject the request.
+* Added a workaround for a bug in Internet Explorer 11 on Windows 7 that could cause the user interface to hang.
 
 ### 4.7.1
 
--- a/lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
+++ b/lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
@@ -96,7 +96,7 @@ const StandardUserInterface = React.createClass({
         const terria = this.props.terria;
         const allBaseMaps = this.props.allBaseMaps;
         return (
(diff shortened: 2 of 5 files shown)


kring:
@AlexGilleran you may want to take a quick look at this to make sure I haven't done something horribly abusive with the CSS here.  For example, I'm not entirely sure if that rule I've added applies globally or if the CSS modules make it more localized.  If it's the former and there's a way to do the latter, we should try that.

AlexGilleran:
It should probably be in the global.scss file (there's still one in Terria right?) but it'll work either way - css modules can't rewrite the name of an element, only class names.

kring:
> css modules can't rewrite the name of an element, only class names

Right, of course.  We could still scope it to only apply inside our elements somehow though, right?  Like `.some-class-on-the-root-ui-element svg use { ... }`?

AlexGilleran:
Oh right, that's what you mean sorry - yep that makes sense. Maybe `[class*='tjs'] { ... }`

AlexGilleran:
Or on the root works too, assuming there is one for every bit of markup we create.

PR #2292, opened 2016-12-22
Fix up user guide generation process

kring:
Previously the doc generation process was in https://github.com/TerriaJS/MakeDocs and used bash, perl, and python.  Now everything needed is in this repo and it only needs node and python (for mkdocs).  It also uses a new and less buggy template.


kring:
New doc generated after the changes in this PR is at http://terria.io/Documentation/

PR #2294, opened 2017-01-03
Add some tableStyle variables for displaying charts

rdengate:
Per column colour and yAxis min/max values to allow more control over chart display.


lib/Models/TableColumnStyle.js
```diff
@@ -122,6 +137,12 @@ var TableColumnStyle = function(options) {
     this.featureInfoFields = options.featureInfoFields;
 
     /**
+     * Color for column (css string)
+     * @type {String}
+     */
+    this.color = options.color;
```
kring: This only applies to the chart, not to the map, right?  A more descriptive name (`chartLineColor`?) would make it clearer, or at least mention in the doc that it only applies to charting.

kring:
Please update CHANGES.md, too.

kring:
Looks good other than those two small things.

rdengate:
Thanks @kring. I have made all changes you suggested.

PR #2295, opened 2017-01-03
Treat json in ComplexData of WPS return as catalog item

rdengate:
This allows custom catalog items to be displayed. I've removed the ability to specify color directly in WPS return, as I think if per column colour is going to be specified it should be via catalogItem tableStyle, which is more flexible and less hacky.


lib/Models/WebProcessingServiceCatalogItem.js
```diff
@@ -27,6 +28,7 @@ function WebProcessingServiceCatalogItem(terria) {
     CatalogItem.call(this, terria);
 
     this._geoJsonItem = undefined;
+    this._terria = terria;
```
kring: This isn't necessary, the `CatalogItem` base class exposes terria on the `terria` property itself already.

kring: Sorry, `CatalogMember`, not `CatalogItem`.

lib/Models/WebProcessingServiceCatalogItem.js
```diff
@@ -153,15 +155,15 @@ WebProcessingServiceCatalogItem.prototype._load = function() {
...
-                        var colors = '';
-                        if (defined(outputs[i].colors)) {
-                            colors = 'colors="' + outputs[i].colors + '"';
-                        }
+                    if (outputs[i].Data.ComplexData.mimeType === 'text/csv') {
                         content = '<collapsible title="' + outputs[i].Title + '" open="' + (i === 0 ? 'true' : 'false') + '">';
-                        content += '<chart ' + colors + ' can-download="true" hide-buttons="false" title="' + outputs[i].Title + '" data=\'' + outputs[i].Data.ComplexData + '\' styling="histogram"></chart>';
+                        content += '<chart can-download="true" hide-buttons="false" title="' + outputs[i].Title + '" data=\'' + outputs[i].Data.ComplexData + '\' styling="histogram"></chart>';
                         content += '</collapsible>';
+                    } else if (outputs[i].Data.ComplexData.mimeType === 'application/json') {
```
kring: I'm not sure it's fair to assume that ComplexData with mime type `application/json` is always a TerriaJS init file.  Let's instead use `application/vnd.terriajs.init+json`.  In theory we can officially register that (http://www.iana.org/form/media-types) but it might be overkill.

kring: Actually since this isn't a full init file, but rather a single catalog member, `application/vnd.terriajs.catalog-member+json` would be better.

kring:
Looks great other than those nitpicks.  Please update CHANGES.md too while you're in there.

rdengate:
Thanks @kring. This should be good to go now.

PR #2298, opened 2017-01-04
SDMX-JSON reader enhancements

### LINKED ISSUE #2202: SDMX-JSON reader enhancements
Follows on from #1514 .  See the `sdmx5` branch.

- [x] Some measures are mutually exclusive and shouldn't be selectable together.
- [x] Accept `LGA2011` as a region code (currently only `LGA` or `LGA_2011` accepted)
- [x] Add a flag to sort value names alphabetically/numerically.
- [x] Have a way to hide some values from the config - eg. so we can show the 5-year age bands and hide the individual years under “Age”.
- [x] Treat “total” values differently (eg. so if you click on “Persons” you turn off the checks on “Male” and “Female") - probably by manual json config, unless there’s a way to specify which are the totals in the SDMX.
- [x] Include a checkbox to show values as a percent of the total (like we have now for ITT) - which requires a way to configure which query gives the totals.
- [x] Have a way to download the shown data (eg. as csv).
- [x] Put a better url in the "About this Dataset" info box.
- [x] Use Summarised concepts if appropriate.
- [x] Clean up TO DOs.
- [x] Fix a bug that can show "no value" when you change region types and percentage of region.
- [x] Improve SummarisedConcept behaviour when you deselect all values. Or prevent this from happening if allowMultiple is false.
- [x] Organise the catalog.

Could also show units (eg. number of households/people/businesses etc), if available.

RacingTadpole:
Fixes #2202.  Incorporates #2301.

Test using `http://localhost:3001/#build/terriajs/test/init/sdmx-abs.json` and `smdx.json`.

In addition to updating the SDMX-JSON catalog item, also brings some changes and features through for others to use:
- Can set `requireSomeActive` to prevent the user from deselecting all display variables.
- Reduce code duplication in `DisplayVariablesConcept.prototype.toggleActiveItem` and `VariableConcept`.
- Add the new `SummaryConcept`s.
- `TableCatalogItem`s now show a download URL in About This Dataset, which downloads the entire dataset as csv, even if the original data was more complex (eg. SOS or SDMX).
- Allows the legend to be placed after the variable choice via `displayChoicesBeforeLegend`. This is useful if changing variable can trigger a reload and legend change, so they don't jump around.


lib/Core/numberUtils.js
```diff
@@ -0,0 +1,34 @@
...
+function startsNumeric(n) {
+    // startsNumeric("15") = true
+    // startsNumeric("-1.1") = true
+    // startsNumeric("-1.1a") = true   // ** Note difference from isNumeric.
+    // startsNumeric("Infinity") = true  // ** Also different, but unlikely to be a problem.
+    // startsNumeric("NaN") = false
+    return !isNaN(parseFloat(n));
+}
+
+function compareStringsAndNumbers(a, b) {
```
kring: Have you looked at `javascript-natural-sort`?  It's a more robust (but probably much slower) version of this.

RacingTadpole: Oh, I didn't see we already had that installed - thanks. I won't try to comprehend its [code](https://github.com/Bill4Time/javascript-natural-sort/blob/master/naturalSort.js) though...

kring:
Clicking the "Download the currently selected data as csv" link does this in IE11:
image

If I say "Allow" I get this:

image

And if I look for an app in the store I get:
image

I think `data:attachment/csv` might be a bit suspect, though I'm far from an expert on this.  Isn't that supposed to be an MIME type, so `text/csv`?

lib/Core/numberUtils.js
kring: Yeah and in my experience it is _very_ slow.  If you're sorting more than, say, a couple hundred items, it might not be a good choice.

lib/ReactViews/Preview/MappablePreview.jsx
```diff
@@ -132,14 +132,18 @@ const MappablePreview = React.createClass({
                                             {catalogItem.dataUrlType.indexOf('wfs') === 0 && <a href="http://docs.geoserver.org/latest/en/user/services/wfs/reference.html" target="_blank" key="wfs">Web Feature Service (WFS) documentation</a>}
                                             {catalogItem.dataUrlType.indexOf('wcs') === 0 && <a href="http://docs.geoserver.org/latest/en/user/services/wcs/reference.html" target="_blank" key="wms">Web Coverage Service (WCS) documentation</a>}
                                             {' '} for more information on customising URL query parameters.
+                                            <br/>
+                                            <Link url={catalogItem.dataUrl} text={catalogItem.dataUrl}/>
+                                        </When>
+                                        <When condition={catalogItem.dataUrlType === 'data-uri'}>
+                                            <Link url={catalogItem.dataUrl} text="Download the currently selected data as csv" download="data.csv"/>
```
kring: `CSV` is an acronym (Comma Separated Values) so it should be capitalized.

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -393,27 +484,22 @@ SdmxJsonCatalogItem.prototype._load = function() {
     // which is not the case here.  We will need to pass them to each call to new TableColumn as well.
     this._tableStructure = new TableStructure(this.name, this._columnOptions);
     this._regionMapping = new RegionMapping(this, this._tableStructure, tableStyle);
-    this._regionMapping.changedEvent.addEventListener(dataChanged.bind(null, this), this);  // TODO: not sure we need this.
+    // this._regionMapping.changedEvent.addEventListener(dataChanged.bind(null, this), this);
```
kring: Remove commented out code?

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -393,27 +484,22 @@ SdmxJsonCatalogItem.prototype._load = function() {
...
-    // Polling is not implemented on SDMX-JSON items (yet).
-};
-
-// An event listened triggered whenever the dataSource or regionMapping changes.
+// An event listener triggered whenever the dataSource or regionMapping changes.
 // Used to know when to redraw the display.
-function dataChanged(item) {
-    item.terria.currentViewer.notifyRepaintRequired();
-}
+// function dataChanged(item) {
```
kring: Here too

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -427,6 +513,11 @@ function updateColumns(item, newColumns) {
     item.terria.pickedFeatures = undefined;
 }
 
+// Adds the wildcard's exclusive values to the front of those for this dimension id, if any.
+function gettotalValueIdsForDimensionId(item, dimensionId) {
```
kring: capital T in total

lib/Models/TableCatalogItem.js
```diff
@@ -83,9 +83,9 @@ var TableCatalogItem = function(terria, url, options) {
      * Gets or sets the opacity (alpha) of the data item, where 0.0 is fully transparent and 1.0 is
      * fully opaque.  This property is observable.
      * @type {Number}
-     * @default 0.6
+     * @default 0.8
      */
-    this.opacity = 0.6;
+    this.opacity = 0.8;
```
kring, line 88: A change of default is technically a breaking change.  We can live with it if there's a good reason, though.

lib/Models/TableCatalogItem.js
```diff
@@ -208,6 +208,40 @@ var TableCatalogItem = function(terria, url, options) {
...
+        }
+    });
+
+    overrideProperty(this, 'dataUrlType', {
+        get: function() {
+            if (defined(this._dataUrlType)) {
+                return this._dataUrlType;
+            }
+            if (defined(this._tableStructure)) {
+                return 'data-uri';
```
kring, line 237: Should we have a more specific `dataUrlType`?  I noticed elsewhere that the UI is assuming that `data-uri` always means CSV, which is kind of questionable.

RacingTadpole: good point. Although it's also useful for it to know that the url is a data-uri, not a link to an external csv file.  So perhaps `csv-data-uri`...

lib/Models/TableCatalogItem.js
RacingTadpole: I just found it looks much nicer, with richer colours, but still readable map text underneath. Eg. for demos I find myself wanting to dial up the opacity to show it off to best advantage. But it's not critical.
We could perhaps leave the TableCatalogItem default as 0.6, but introduce 0.8 on SDMXJSON catalog items. A bit inconsistent, but then it's not a breaking change.

lib/Core/numberUtils.js
RacingTadpole: I guess this won't be sorting huge lists, so we could use it. But maybe it would be better to switch the other uses of natural-sort to this new one instead..., and ditch the dependency?

lib/ReactViews/Workbench/Controls/SummaryConcept/SummaryConcept.jsx
```diff
@@ -0,0 +1,187 @@
...
+ *    activate any leaf nodes.)
+ *    (If summaryConcept.allowMultiple is false, you cannot 'cancel' or 'go back' on this.)
+ * If summaryConcept.allowMultiple is true, then an <./AddButton> is also shown,
+ *    which simply opens the root concept, at which point OpenInactiveConcept takes over.
+ *
+ * This design would need revision to handle concepts whose direct children are a mix of
+ * both leaf nodes and parent nodes.
+ *
+ * This component cheekily uses the active concepts' isActive flag -
+ * which is by rights the boolean true - to instead store an integer (>0) giving
```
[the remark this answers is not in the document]
RacingTadpole: you're right, that's a trick I was using in the cotf2 branch for something else, which doesn't appear here. I'll remove it from this PR.

test/Models/SdmxJsonCatalogItemSpec.js
```diff
@@ -238,13 +241,15 @@ describe('SdmxJsonCatalogItem', function() {
...
                 var regionDetails = item.regionMapping.regionDetails;
                 expect(regionDetails).not.toBeDefined();
-                expect(item.isMappable).toBe(false);
-                // Expect it to have created the right table of data (with no time dimension).
-                var columnNames = item.tableStructure.getColumnNames();
-                expect(columnNames.length).toEqual(2 + 5 * 2 * 2); // one for each value, plus date and total columns.
-                expect(item.tableStructure.columns[0].values.length).toEqual(3); // 3 dates.
-                // Expect it to show 3 concepts to the user.
-                expect(item.concepts.length).toEqual(3);
+                // Beyond this, assume nothing. But if you want it to work, you may want the following:
```
kring, line 244: Why was this test commented out?

lib/Core/numberUtils.js
kring: No, that one does lots of clever stuff.  For example, it sorts correctly when numbers are in the middle or at the end of the string.  It also handles dates I believe.

lib/Models/TableCatalogItem.js
kring: Ok, let's leave it, but make sure it's mentioned in CHANGES.md.

lib/Models/TableCatalogItem.js
kring: It could spot a data URI by the fact that they always start with `data:`.

kring:
This looks great, I love the new UI.  Honestly the best I can do is bike shed this (nitpick the simple parts I understand, and assume you know what you're doing in the complicated parts!) but I tried some things and I can't see any problems, so I'm happy to merge it once those minor things are addressed.

RacingTadpole:
thanks! I'll get on to them now.

test/Models/SdmxJsonCatalogItemSpec.js
RacingTadpole: Non-spatial SDMXs passed this test but gave weird results before. So I made them _really_ not work instead, so the rest of this test fails, but the user gets a more sensible experience.

RacingTadpole:
@kring - thanks for looking at this. I've hopefully addressed all your comments now.

PR #2301, opened 2017-01-05
Interpret dates of the form 2016-Q2

RacingTadpole:
This format is often used by the ABS in its SDMX-JSON datasets, eg. http://localhost:3001/proxy/http://stat.data.abs.gov.au/sdmx-json/data/RES_PROP_INDEX/1.3..Q/all/ . 

This increases how many ABS datasets we can display.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
+### 4.7.5
+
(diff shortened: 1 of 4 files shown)


kring:
This looks ok to me, but I'll merge it with #2298.

PR #2308, opened 2017-01-06
Upgrade to terriajs-cesium 1.29.1

kring:

PR #2314, opened 2017-01-07
Reduce the initial size of TerriaJS, especially in 2D

kring:
This builds on #2308, so merge that first.

With these changes, the release build of nationalmap.js goes from 3.8 MB to 2.7 MB.  nationalmap.js alone is sufficient for 2D and for most catalog item types.  Additional JS code is loaded for 3D and when loading KML, GeoJSON, or CZML (that is, clicking it in the preview or adding it to the map... just having it in the catalog won't trigger download of the extra code).  The total download size is about the same if you need everything, but this makes the download smaller in some important scenarios.


kring:
Fixed the test failures, so would you mind reviewing this one too @steve9164?

tephenavies:
I didn't get around to fully testing it. I got caught up checking that modules weren't being duplicated in different modules. I used this plugin:
``` javascript
var PrintChunksPlugin = function() {};
PrintChunksPlugin.prototype.apply = function(compiler) {
    compiler.plugin('compilation', function(compilation, params) {
        compilation.plugin('after-optimize-chunk-assets', function(chunks) {
            console.log(JSON.stringify(chunks.map(function(c) {
                return {
                    id: c.id,
                    name: c.name,
                    includes: c.modules.filter((m) => m.request).map(function(m) {
                        return m.request.split('/').slice(-2).join('/');
                    })
                };
            })));
        });
    });
};
```
(modified from webpack docs) to print out the modules in each chunk, copied the relevant output json, then put it into python to check for and duplicates:
``` py3
string = '''
    json output here
'''
import json
o = json.loads(string)
a,b,c = [set(c['includes']) for c in o]

print(a.intersection(b))
print(a.intersection(c))
print(b.intersection(c))
```

The only duplicate files are:
```
'DataSources/ReferenceProperty.js',
'DataSources/ScaledPositionProperty.js',
'Core/CornerType.js',
'Scene/ColorBlendMode.js',
'DataSources/CompositePositionProperty.js',
'DataSources/CompositeProperty.js'
```
which are in both `1.nationalmap.js` and `2.nationalmap.js`

kring:
Oh no, I should have told you!  Webpack has a great tool for doing that automatically.  Just uncomment the lines in `runWebpack.js` in this PR to write `stats.json`, set `profile: true` in NationalMap's `webpack.config.js` (not 100% sure this step is necessary), run the build, and then upload the generated stats.json file to this web site:
https://webpack.github.io/analyse/

The Hints tab will show duplicated modules if there are any.

tephenavies:
That's good to know for next time! Luckily it didn't take too long to analyse them myself.

PR #2320, opened 2017-01-11
CKAN Improvements

### LINKED ISSUE #2318: Add format appended to dataset name for CKANCatalogItems
Currently the CKANCatalogGroup creates CKANCatalogItems based on the name.  Where a single dataset has multiple spatial resources in CKAN, it creates a duplicate NM Catalog item for each resource, but you have to read the metadata for each one to figure out which is which.

If we can add the resource format to the name, it'll provide an explanation for the repetition, as well as allowing the more discerning data-adder their selection of sources.

I've created a simple mockup - it could be parentheses, hyphens, whatever works...
image

### LINKED ISSUE #2319: CkanCatalogItem sets res_format:ESRI REST to ArcGisMapServerCatalogItem explicitly, missing ArcGisFeatureServerCatalogItem
In https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CkanCatalogItem.js, resources detected as ESRI REST are automatically assumed to be ArcGisMapServerCatalogItem, but in many cases, they can also be ArcGisFeatureServerCatalogItems (current point in case http://www.data.gov.au/dataset/mbrc-libraries)

```
this.esriMapServerResourceFormat = /^esri rest$/i;
...
[options.esriMapServerResourceFormat, ArcGisMapServerCatalogItem],
```

The "ESRI REST" format is set explicitly in the client's data.json file (http://data.moretonbay.qld.gov.au/data.json)

What is the complexity of either checking both formats for ESRI Rest sources, or performing some test before choosing the correct model?

kring:
From CHANGES.md:
* `CkanCatalogGroup` now automatically adds the type of the resource (e.g. `(WMS)`) after the name when a dataset contains multiple resources that can be turned into catalog items and `useResourceName` is false.
* Added support for ArcGIS FeatureServers to `CkanCatalogGroup` and `CkanCatalogItem`.  In order for `CkanCatalogGroup` to include FeatureServers, `includeEsriFeatureServer` must be set to true.

To try this out, use http://localhost:3001/#clean&https://raw.githubusercontent.com/TerriaJS/NationalMap-Catalog/esri/build/nm.json and look for FeatureServers under `Data.gov.au -> Moreton Bay Regional Council` and multiple resources of different types under `Local Government -> Victoria -> Ballarat`.  To verify that MapServers still work, check out `Data.gov.au -> Geoscience Australia -> GA_SBAGD_Framework (MapServer)`.

Fixes #2318
Fixes #2319

### DIFF
--- a/lib/Models/CkanCatalogGroup.js
+++ b/lib/Models/CkanCatalogGroup.js
@@ -651,9 +651,7 @@ function populateGroupFromResults(ckanGroup, json) {
         if (resourceItems.length > 1 && !ckanGroup.useResourceName) {
             resourceItems.forEach(function(item) {
                 var typeName = CkanCatalogItem.shortHumanReadableTypeNames[item.type] || 'Other';
-                if (defined(typeName)) {
-                    item.name += ' (' + typeName + ')';
-                }
+                item.name += ' (' + typeName + ')';
             });
         }
     }
--- a/lib/Models/CkanCatalogGroup.js
+++ b/lib/Models/CkanCatalogGroup.js
@@ -567,6 +567,8 @@ function populateGroupFromResults(ckanGroup, json) {
             }
         }
 
+        var resourceItems = [];
+
         var resources = item.resources;
         for (var resourceIndex = 0; resourceIndex < resources.length; ++resourceIndex) {
             var resource = resources[resourceIndex];
@@ -576,7 +578,11 @@ function populateGroupFromResults(ckanGroup, json) {
                 groups = item.groups;
             } else if (ckanGroup.groupBy === 'organization' && item.organization) { // item.organization is sometimes null
                 groups = [item.organization];
+            } else {
+                groups = undefined;
             }
+
+            var addedItem;
 
             if (defined(groups) && groups.length > 0) {
                 for (var groupIndex = 0; groupIndex < groups.length; ++groupIndex) {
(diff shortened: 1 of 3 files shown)


kring:
@steve9164 can I bug you to review and merge this one too?  It'd be great to get it into the NationalMap release tomorrow.

tephenavies:
Oh sure it's today. I've heard that one before 😉

kring:
Hahaha no no it's _tomorrow_! ;)  Totally definitely not yesterday or the day before.

lib/Models/CkanCatalogGroup.js
```diff
@@ -597,25 +620,41 @@ function populateGroupFromResults(ckanGroup, json) {
...
+            }
+        }
+
+        // If there's more than one resource item, and we're not using the resource name to name
+        // our items, then they'll all have the same name.  Add the type to the name to help
+        // distinguish them.
+        if (resourceItems.length > 1 && !ckanGroup.useResourceName) {
+            resourceItems.forEach(function(item) {
+                var typeName = CkanCatalogItem.shortHumanReadableTypeNames[item.type] || 'Other';
+                if (defined(typeName)) {
```
tephenavies: This condition will always be true. Perhaps you meant to check if `item.name` was defined?

tobybellwood:
so i just need to set includeEsriFeatureServer:true in the ejs for NM-catalog as well as includeEsriMapServer:true now?

kring:
> so i just need to set includeEsriFeatureServer:true in the ejs for NM-catalog as well as includeEsriMapServer:true now?

Yep, that's right.  On top of the Solr changes you already made of course.

lib/Models/CkanCatalogGroup.js
kring: Oops, that `if` check isn't needed at all anymore, since I added the `|| 'Other'`.  I'll remove it.

kring: Good catch!

tephenavies (approved):

PR #2328, opened 2017-01-12
Don't crash when searching for a location in 2D.

### LINKED ISSUE #2321: Place/Location search marker not visible in 2D
On both NM and AREMI, a location/address search completes successfully, but doesn't render the placemarker, or zoom-to in 2D (tested on Chrome and IE11 here at PMC and iPad Chrome).

kring:
It was trying to set the height of the marker by querying Cesium terrain, which isn't available in 2D.

Fixes #2321 

Even after this PR, if you search in 2D, and then switch to 3D, the marker won't be at the right height so you'll get a parallax effect.  This is difficult to fix and IMO not too critical.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 Change Log
 ==========
 
+### 4.8.3
+
+* Fix a bug that prevented drawing the marker and zooming to the point when searching for a location in 2D.
+
 ### 4.8.2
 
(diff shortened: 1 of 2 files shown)


tephenavies:
Should the changelog be updated to put the change under 4.8.2? (since 4.8.1 is the current release)

kring:
> Should the changelog be updated to put the change under 4.8.2? (since 4.8.1 is the current release)

Yep, done.

PR #2334, opened 2017-01-17
Support for adding HelpScreens

rdengate:
If defined by the instance using TerriaJS, this adds a button to the top right titled help, which drops down a list of available help sequences, such as how to add data or how to change map settings. When a dropdown item is clicked, a sequence of help screens appears leading the user through a task.


lib/ReactViews/HelpScreens/HelpPanel.jsx
```diff
@@ -0,0 +1,151 @@
...
+    mixins: [ObserveModelMixin],
+
+    propTypes: {
+        viewState: React.PropTypes.object.isRequired,
+        helpSequences: React.PropTypes.object
+    },
+
+    componentWillMount() {
+        const that = this;
+        setInterval(function() {
```
kring: This interval needs to be cleared in `componentWillUnmount`.

kring: Ideally you would only set the interval when you have screens to show.  A fast interval like this one (10ms!) will run even when you're focused on another tab and will have implications on battery life.

lib/ReactViews/HelpScreens/HelpPanel.jsx
```diff
@@ -0,0 +1,151 @@
...
+            if (!defined(that.state.screens)) {
+                return;
+            }
+            if (that.props.helpSequences.cancel) {
+                // Has been cancelled from somewhere else. Abort!
+                that.cancel();
+            }
+            const i = that.state.index;
+            const currentScreen = that.state.screens[i];
+            if (defined(that.state.previousRectangle) && that.state.previousRectangle === that.state.currentRectangle) {
```
kring: Why not `updateCurrentRectangle` before checking to see if the rectangle changed?  Seems like it would simplify the logic and avoid the help panel being a "frame behind".

kring (changes requested):
Looks great!  I definitely think there's potential to make this a library entirely separate from TerriaJS, but don't worry about that for now.

rdengate:
Thanks @kring! All changes are there now.

kring:
Well @rdengate I have a couple more suggestions.  I know you're running low on time and have lots of other things to do, so if you want to merge this as-is for now and move on I'm ok with that (and maybe I'll do some of these suggestions myself at some point).  But here we go:

First, when the user clicks the highlighted element, it should probably move to the next help screen rather than close the help entirely.  The need for this is most obvious with the SOE "Load data from a file or external source" help.  The instructions tell you to click on the button, but if you actually do, it closes the help rather than moving to the next step.

Second, I think the JSX components can be factored a bit better.  Currently `HelpPanel` is both a UI element that lets the user launch the help, and also it installs the timer to track elements and manages the sequence of help screens.  Those two things are totally orthogonal and shouldn't be in the same React component.

So I think there should be two top-level, help-related components in SoE's UserInterface.jsx: `HelpMenuPanel` and `HelpOverlay`.  `HelpMenuPanel` is the Help button and the menu that appears when you click it.  `HelpOverlay` takes a single `HelpSequence` and executes it.  It installs the interval timer and manages the sequence of `HelpScreen`s.  It contains a `HelpScreenWindow` and `ObscureOverlay`.

To communicate between `HelpMenuPanel` and `HelpOverlay`, it may be helpful to introduce a `HelpViewState`.  Your `HelpSequences` is close, but there's no need for the view state to hold the list of sequences.  Only the `HelpMenuPanel` needs to know the possible set of sequences.  The rest of the help system just needs to know the current sequence and which scene within the sequence is currently displayed.  Canceling can be as simple as setting `currentSequence` to undefined.

Not sure if that totally made sense, so let me know what you think.  And like I said it's ok if you have more important things to do right now.

rdengate:
@kring, thanks for comments. I believe I've made all changes suggested! This should be good to go, hopefully.

kring (approved):

kring:
Whoo hoo, this looks great!  Thanks for all the iterations!

rdengate:
@kring thanks! :-D

PR #2342, opened 2017-01-23
Cotf values list of dict

tephenavies:
Merge front-end changes to match API changes. Also merges master.

There is still an issue where conditions cannot be removed due to an exception being thrown in the `isActive` toggling code for display variables.


RacingTadpole:
Any idea why we're getting the spec error `reprojectPoint reprojects a point from one CRS to another`?

tephenavies:
No. I'll take a look.

tephenavies:
https://github.com/TerriaJS/terriajs/pull/2341 seems to be the reason

PR #2346, opened 2017-01-25
Make link coloring a function of the background color

RacingTadpole:
Something like this is needed to address https://github.com/TerriaJS/neii-viewer/issues/84 in a non-hacky way.

We ought to apply the same approach to all text and button coloring - either in this or a separate PR.

@chloeleichen, @AlexGilleran, @kring ?

### DIFF
--- a/lib/ReactViews/ExplorerWindow/explorer-window.scss
+++ b/lib/ReactViews/ExplorerWindow/explorer-window.scss
@@ -41,7 +41,7 @@
     @include transform(none);
   }
   a{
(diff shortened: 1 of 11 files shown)


RacingTadpole:
Anyone want to take a look at this?  I've removed my comment "don't merge this", since it would be nice to have this functionality in the next release for the NEII viewer.

kring:
Looks good to me, so I'm merging.  @chloeleichen you may want to take a glance at this when you get a chance.

PR #2348, opened 2017-01-30
Fix ArcGisCatalogItem for proj4 v2.3.16

tephenavies:
Fix ArcGisCatalogItem relying on `proj4` correctly mutating a point, and change usage of `proj4` function across terriajs to the form `proj4(CRS, CRS, [lon, lat]) -> [lon, lat]`, which is the documented form on [`proj4` Github page](https://github.com/proj4js/proj4js) to avoid future issues.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,12 @@
 
+
 Change Log
 ==========
 
 ### 4.8.3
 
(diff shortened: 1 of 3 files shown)


kring:
@steve9164 can you update CHANGES.md?

PR #2352, opened 2017-02-01
Adds WPS POST Execute option

rdengate:
If useKvp is set to false in the catalog item in the json init file for a WPS process, the WPS Execute request will become a POST request with parameters encoded in the XML payload. Otherwise, parameters will be encoded in the GET request as key value pairs (KVP) as usual. This allows massive polygons such as Australia to be sent without worrying about URL length limits.


lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -53,6 +56,13 @@ function WebProcessingServiceCatalogFunction(terria) {
      */
     this.identifier = undefined;
 
+    /**
+     * Gets or sets whether to use key value pairs (KVP) embedded in Execute URL, or whether to make a POST request
+     * with XML data.
+     * @type {Boolean}
+     */
+    this.useKvp = undefined;
```
kring: We should document the default.  Also, is there a better name for this?  The spec describes it as "Execute HTTP GET request KVP encoding" (versus "Execute HTTP POST request XML encoding").  How about `executeWithHttpGet`?

You probably already know this, but I was interested to learn that WPS servers are required to support POST for Execute, but GET is optional.  So I think changing the default to POST makes a lot of sense.

lib/Models/DateTimeParameter.js
```diff
@@ -44,4 +44,22 @@ defineProperties(DateTimeParameter.prototype, {
...
 });
 
+/**
+ * Process value so that it can be used in an URL.
+ * @param {String} value Value to use to format.
+ * @return {String} Stringified JSON that can be used to pass parameter value in URL.
+ */
+DateTimeParameter.formatValueForUrl = function(value) {
+    return JSON.stringify({
+                type: 'object',
```
kring, line 54: This is... weird.  I mean, what is this format?  It seems like it might be WPS specific, in which case it doesn't belong in the DateTimeParameter.  I realize this ship has already sailed a bit, but the whole notion of `formatValueForUrl` doesn't make a lot of sense to me.  For the GeoJSON types, it seems like it does the same thing as `getGeoJsonFeature` but then wraps it up in a FeatureCollection and JSON.stringify's it.  It would be much better IMO to be specific about this.  Have a function `getGeoJsonFeatureCollection`, and if someone wants that as a JSON string they can call JSON.stringify themselves.

lib/Models/WebProcessingServiceCatalogFunction.js
```diff
@@ -451,38 +442,69 @@ WebProcessingServiceCatalogFunction.prototype.invoke = function() {
...
+    dataInputs = dataInputs.join(";");
+    var uri = new URI(that.url).query({
+        service: 'WPS',
+        request: 'Execute',
+        version: '1.0.0',
+        Identifier: that.identifier,
+        DataInputs: dataInputs
+    });
+    if (that._statusSupported) {
+        uri.addQuery('status', true);
```
kring, line 470: It looks like `status` and `storeExecuteResponse` don't get added in the POST path.  Is that a problem?

kring: Oh sorry I'm wrong, they're just added differently.  Nevermind!

lib/Models/WebProcessingServiceCatalogGroup.js
```diff
@@ -33,14 +33,16 @@ function WebProcessingServiceCatalogGroup(terria) {
      */
     this.url = '';
 
+    this.useKvp = true;
```
kring: Needs doc.

lib/Models/WebProcessingServiceCatalogFunction.js
kring: I noticed this is `knockout.track`'d in the group, but not in the function here.  It should be here too I think.

kring:
Looks good, just those few comments.  On the `formatValueForUrl` front, I'm ok with keeping that question separate from this PR.  But if you do have time to clean that up, that's even better!

lib/Models/WebProcessingServiceCatalogFunction.js
rdengate: I agree that the default should be POST, but since TerriaJS has been defaulting to GET, if we change this it's going to break WPS for anyone whose WPS server can't accept a POST request (if such people exist).

kring: Well the spec says POST is required, so theoretically such people shouldn't exist.  `GetCapabilities` also advertises which methods are supported, so we could check that to be super thorough:

image

rdengate: Hmm okay. I'll change it to default to POST. Thanks!

rdengate:
Thanks @kring! I think `formatValueForUrl` is going to be a can of worms, so considering it separately would be good. I've addressed the other comments.

PR #2353, opened 2017-02-01
Additional catalogs

rdengate:
This is a a stab at adding additional catalogs (which would be used for separating functions/tools/analysis/whatever-you-want-to-call-it from data layers). It's doing so in a way that's very naive and probably will break for some cases e.g. creating a share link. It seems to be sufficient for what Geoglam needs, though (analytics don't transfer via share link anyway).

Geoglam would use it by adding (in index.js):
```
        var toolsCatalog = new Catalog(terria);
        toolsCatalog.name = 'Tools';
        toolsCatalog.id = 'tools';

        loadJson5("init/some_file.json").then(function(json) {
            toolsCatalog.updateFromJson(json.catalog);
            terria.additionalCatalogs.push(toolsCatalog);
        });
```

Two changes to note here though:
1. ViewState no longer uses modalTabIndex to keep track of tabs (i.e. their position in the catalog) but rather an id which is set at creation. So it's still hardcoding some data, but at least not the position in the UI.
2. viewCatalogItem doesn't hardcode opening user data or add data, preferring to get the catalog id from the item passed. This fixes a bug in where if you add user data then close or open the User-Added Data group that's created, the explorer panel takes you to the Data Catalogue.


kring:
Closing in favor of #2692.

Issue #2026, opened 2016-09-06
Support bar charts


stevage:
This was requested at the National Map new UI launch. The use case was election data, showing first preferences as a bar chart when you click on a point.

AnaBelgun:
Ability to visualise time series data with other types of charts has been requested again by NEAR/EUDM team. 
Below is their feedback: 
"Having some alternate charting types might be useful"
More details here: https://docs.google.com/spreadsheets/d/1GPbQ9xeX4bgB2ZyZlpbqKZeBUvf3yf5qw9jb7wqs5CA/edit#gid=0

kring:
> Having some alternate charting types might be useful

That's some pretty vague feedback to build a major feature on. Do we know / can we find out more specifically what they're trying to accomplish and how alternate charting types would help?

We do have a mostly-implemented bar chart feature, though, so rolling that out wouldn't be too hard. It just might not be too useful, either.

AnaBelgun:
Ana to follow up and see what examples they have when a bar chart would be better

AnaBelgun:
when the time comes, there is an old PR here: https://github.com/TerriaJS/terriajs/pull/2362

kring:
> when the time comes, there is an old PR here: #2362

That's all in #3151 now.

AnaBelgun:
Waiting for NEAR CSIRO team to provide more details about the expected chart behaviour

Issue #2113, opened 2016-09-21
Misleading chart interpolation when values are missing


RacingTadpole:
Eg. With #test -> Droughts by country, Malaysia has a drought only in 1998 and 2014.  The chart ignores all missing data in between and interpolates between the two, which is misleading:

![image](https://cloud.githubusercontent.com/assets/1757858/18698126/04632ece-800b-11e6-933c-9970d2f9320d.png)

This might be a good place to use a bar chart instead.

chloeleichen:
or just show two points, or show a grey area in between and label it no data ? An example of gray area is here if you scroll to bottom and see the Device Usage Chart: http://dashboard.gov.au/dashboards/1-mygov

nahidakbar:
![image](https://cloud.githubusercontent.com/assets/3362993/22816288/53ca4dd4-efb5-11e6-9913-9449340c9fc1.png)
better?
@kring  @RacingTadpole

RacingTadpole:
whoa!! That's looking pretty awesome!

nahidakbar:
PR #2362 review and close

meh9:
@nahidakbar how does it determine if it should be showing a line or bars, just by detecting missing values? Or do you control it with config params or something?

nahidakbar:
Right now, just based on point density. If you have some more datasets you can point me to which should be rendered this way, I can test to see if they conform.

nahidakbar:
@meh9

PR #2362, opened 2017-02-10
2113 bar chart implementation

nahidakbar:
#2113


RacingTadpole:
Noticed the specs are failing with this one - ```IE 10.0.0 (Windows 7 0.0.0) ERROR
  Object doesn't support property or method 'assign'``` - can you take a look?

kring:
@nahidakbar we don't have a polyfill for `Object.assign` in TerriaJS, but you can use Cesium's `clone` which works in a lot of the same situations.  Probably worth reviewing the approach before worrying too much about IE10 compatibility though @RacingTadpole.

RacingTadpole:
yeah, good point @kring! Thought I'd start with an easy one though ;-)

RacingTadpole:
I'm having trouble checking this branch out - it keeps thinking I mean commit `2113c5c`!
I can do it by checking out `origin/2113` explicitly, then making my own branch.

RacingTadpole:
oop, now it's missing `lib/Core/assign`... think you might have just forgotten to add it to the commit (although as kring said, elsewhere we're using Cesium's `clone` for this)

lib/Charts/ChartData.js
```diff
@@ -53,7 +53,7 @@ var ChartData = function(points, parameters) {
      * Units of this set of points.
      * @type {String}
      */
-    this.units = parameters.units;
+    this.units = parameters.units || 'unknown';
```
RacingTadpole: Unfortunately adding "unknown" here then displays "unknown" in the y-axis label, eg.
image
image
(these from http://localhost:3001/#build/terriajs/test/init/charts.json)

RacingTadpole:
Can you add an example file to `wwwroot/test/init/charts.json` to make this easier to test?

RacingTadpole:
Somehow the initial animation has gone a little awry - I'm getting the x-axis and its labels descending from the top when I first view the chart (both for bar and line charts).

RacingTadpole:
Does the data shown by the bar chart match up with the data? Eg. with the droughts dataset for India, I got:
image
but now it's:
image

I think I'm getting a large value at the first date that you didn't get in your screenshot? Eg. for Indonesia:
image

lib/Charts/BaseChart.js
```diff
@@ -0,0 +1,392 @@
...
+            d3Element,
+            svg,
+            gTransform,
+            g,
+            t,
+            state,
+            element
+        };
+
+        this.renderChart(renderConext);
```
RacingTadpole: I can see you've made this an abstract base class on which the subclasses need to define the `renderChart` function. Nice! This should work well for when we want to display bar charts and line charts on the same plot, right?
One thing - could you add some js-doc comments for this, including making it clear the only responsibility of the subclasses is define `renderChart` (and override `getXpadding` if desired)? To see how we've done it in the past, take a look at eg. how the comments show that CsvCatalogItem extends TableCatalogItem.

lib/Charts/BarChart.js
```diff
@@ -0,0 +1,44 @@
...
+
+const barWidth = 6;
+
+const BarChart = assign({}, BaseChart, {
+    renderChart(renderContext) {
+        const {g, data, scales} = renderContext;
+        
+        // animations can kiss my ass
+        
+        d3Sync(g.selectAll('.line'), data, 'g', (dataline, datalines_count) => {
```
RacingTadpole: haven't come across this way of doing things before - can you explain it to me? It would be nice to have animations consistent with the line chart too...

lib/Charts/determineChartType.js
```diff
@@ -0,0 +1,56 @@
...
+                };
+            });
+        } else {
+            domain = domain[0];
+        }
+
+        if (domain.x[0] instanceof Date) {
+            const years = (domain.x[1].getTime() - domain.x[0].getTime()) / (3600 * 24 * 365.25 * 1000);
+            
+            const pointsPerYear = points / years;
```
RacingTadpole: So the idea is that annual data gets a barchart, but anything more frequent (eg. daily or monthly data) gets a line chart? That's a good first approx, but I'd guess the unit of time doesn't really matter - eg. given 10 hours of hourly data, then a bar chart might still be better. Conversely if we had 1000 years of annual data, a line chart is better.  So maybe it's more about the total amount of data, not the unit of time?

RacingTadpole: ...which makes me realise this PR doesn't address #2113 in all circumstances - it only solves it when it displays as a bar chart. Ie., this PR actually fixes #2026, and goes some way to solving #2113.

RacingTadpole:
Your approach to adding bar charts looks good to me @nahidakbar - thanks! Great to see that with your refactoring, adding new types of charts is just a handful of code!
A few funny effects going on in the UI, which I don't see in your screenshot, so that will be interesting, plus the other comments along the way too.
Also, we should hook in a way to specify the chart type directly, probably in here: https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Custom/registerCustomComponentTypes.js#L68-L116

lib/Charts/BaseChart.js
RacingTadpole: plus - minor typo - you've called it `renderConext` not `renderContext`.

RacingTadpole:
I've just tried showing a line chart and a bar chart on the same plot, and it shows the second plot as the same type of thing as the first plot. Do you think it would be hard to make this work? Eg. I first dragged this csv on:
```
date,value
1/01/1970,-10000000
1/04/1970,-20000000
1/06/1970,-30000000
1/09/1970,-25000000
1/01/1971,-10000000
1/04/1971,-20000000
1/07/1971,-30000000
1/10/1971,-25000000
```
which shows as a line chart - and then opened the droughts by country, and the drought data shows up on the expanded chart panel as a line chart.  If you do it in the other order, the first one shows as a bar chart, and the csv file above does not show up.

RacingTadpole:
Does this work with negative values too?

lib/Charts/BaseChart.js
```diff
@@ -0,0 +1,392 @@
...
+            transitionDuration,
+            hasData,
+            d3Element,
+            svg,
+            gTransform,
+            g,
+            t,
+            state,
+            element
+        };
```
RacingTadpole: This is quite a funny old bunch of things to pass around like this. It would be much nicer to figure out which ones we really need and put them into a defined class.

nahidakbar:
thanks

ill have a look tomorow

nahidakbar:
@RacingTadpole think I've finished implementing those changes

lib/Charts/ChartType.js
```diff
@@ -0,0 +1,64 @@
...
+    {
+        chartData.renderer = determineChartType(chartData);
+    }
+}
+
+function initialise(state) {
+    state.data.forEach(initialiseChartData);
+}
+
+module.exports.initialise = initialise;
```
RacingTadpole: Can you add some comments/docstrings to this file?
Also, we have an unwritten convention that `module.exports` is always set to an object with the same name as the file. (Eg. try `grep module.exports lib/Models/*`) - can you adjust this to follow that convention?

RacingTadpole:
This is looking really nice! Could you add an example to test/init/charts.json which shows bar and line charts together?

RacingTadpole:
Also the preview y-axis labels have gone black instead of white - 
image
and in the expanded chart the units label looks a bit bad:
image vs 
image

RacingTadpole:
Also with the animation - the charts used to fade in, now the axis flies in from the left and the points from above - can we go back to fading in?

lib/Charts/BarChart.js
```diff
@@ -0,0 +1,48 @@
...
+class BarChart extends BaseChart {
+
+    render(chart, chartData, renderContext) {
+        const {scales, state, chartTransition} = renderContext;
+        const chartDataSets = state.data.filter(data => data.renderer instanceof BarChart);
+        const offset = (barWidth * chartDataSets.length) / 2 - chartDataSets.indexOf(chartData) * barWidth;
+        const sx = scales.x, sy = scales.y[chartData.units], color = chartData.color || 'white';
+        
+        // preprocess for negative data; rect does not seem to like negative values.
+        // TODO: Maybe a better way to do this is with transforms.
```
RacingTadpole: Kevin always hassles me about leaving TODOs in the code... can you decide which way is better?

lib/Charts/ChartRenderer.js
```diff
@@ -0,0 +1,423 @@
...
+    Tooltip.show(Tooltip.html(selectedData), tooltip, state.tooltipSettings, chartBoundingRect);
+}
+
+function unhilightDataAndHideTooltip(g, tooltip) {
+    g.select('.selection').selectAll('circle').data([]).exit().remove();
+    g.select('.selection').selectAll('line').data([]).exit().remove();
+    Tooltip.hide(tooltip);
+}
+
+module.exports = new ChartRenderer();
```
RacingTadpole: Not sure about returning a `new ChartRenderer()` - none of our other classes do that, and it might trip someone up later.

lib/Charts/ChartRenderer.js
```diff
@@ -0,0 +1,423 @@
...
+
+const yAxisLabelWidth = 21;
+
+const defaultNoDataText = 'No preview available';
+
+const defaultTransitionDuration = 1000; // milliseconds
+
+const threshold = 1e-8; // Threshold for 'equal' x-values during mouseover.
+
+class ChartRenderer {
```
RacingTadpole, line 36: I haven't used `class` in js yet. According to the [docs](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes) don't you need a `constructor` for `new` to work?

RacingTadpole:
Any luck with trying to hook in a way to specify the chart type directly, eg. in here: https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Custom/registerCustomComponentTypes.js#L68-L116 ?

RacingTadpole:
Re the animations - I don't think the individual bars are tracking properly - eg. expand two countries' drought charts up, and turn one on and off. The bars of one morph into the bars of the other.
trial

lib/Charts/ChartData.js
```diff
@@ -72,6 +73,17 @@ var ChartData = function(points, parameters) {
...
+    /**
+     * Chart type. If you want these points to be rendered with a certain way. Leave empty for auto detection.
+     * @type {String}
+     */
+    this.type = parameters.type;
+    
+    /**
+     * Chart Renderer Object corresponding to type. Used internally for rendering the chart.
+     */
+    this.renderer = undefined;
```
RacingTadpole: Putting the renderer into a class called ChartData feels iffy - is there another way to do it?

lib/Charts/d3Sync.js
```diff
@@ -0,0 +1,39 @@
...
+ * If you are using D3, you are doing this all the time.
+ * Let's make that code more readable by having a common function we call for this purpose.
+ */
+function d3Sync(parent, arrayData, childElementTagName, updateCallback, transition=undefined)
+{
+    // remove stray elements
+    parent
+        .selectAll(childNodesSelector)
+        .filter(function() { return this.tagName !== childElementTagName; })
+        .remove();
```
RacingTadpole: I just tried commenting out this section (that removes stray elements) and can't see any change in behaviour - can we delete it?

lib/Charts/ChartType.js
```diff
@@ -0,0 +1,64 @@
...
+
+function determineChartType(chartData) {
+    
+    // respect chartData.type
+    if (chartType[chartData.type] !== undefined) {
+        return chartType[chartData.type];
+    }
+    
+    // insert hack here for figuring out which type of chart datasets need
+    try
```
RacingTadpole: Did you have any suggestions on how to choose the chart type in light of my earlier comment?

nahidakbar: If you assign a value to the type field in chart data, and it exists, it will use that type of chart. How that integrates with registerCustomComponentTypes.js, will leave future generations to figure out if they need to specify chart types for data.

nahidakbar:
@RacingTadpole @kring 

Ok I think it is time for me to stop bleeding from my eyes one more time.

Let me know if anything else is messed up.

RacingTadpole:
Thanks. It looks like the animation is still coming in from the corner, and the bars are being replaced, not updated - is that what you're seeing?
barcharts
Also, did you get a chance to add an example to `test/init/charts.json` which shows bar and line charts together?

kring:
@RacingTadpole where do we stand with this?

stevage:
I have rebased this branch and done some further refactoring/documentation here:

https://github.com/TerriaJS/terriajs/tree/2113-barcharts-rebased

stevage:
On that branch I have dealt with the weird transitions issue.

PR #2370, opened 2017-02-13
Further enhancements to SDMX-JSON catalog item

### LINKED ISSUE #2376: Update the draft catalog for ABS's SDMX-JSON offering
Now that SDMX-JSON is largely implemented, bring some order to how we will display the 100+ ABS datasets that are available.

RacingTadpole:
Improved the SDMX-JSON catalog item to handle huge dimensions, allow a blacklist, handle bad responses better, and a few smaller changes.

Fixes #2376.  Follows on from #2202  and #1514.


lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -188,10 +198,11 @@ var SdmxJsonCatalogItem = function(terria, url) {
     this.aggregatedDimensionIds = [];
 
     /**
-     * Gets or sets whether to re-sort the values that appear in the SDMX-JSON response, in the Now Viewing panel.
-     * The default is false, so that the order is maintained (except for totalValueIds, which are moved to the top).
-     * By setting this to true, the values are sorted into alphabetical and/or numerical order.
-     * @type {Boolean}
+     * Gets or sets how to re-sort the values that appear in the SDMX-JSON response, in the Now Viewing panel.
+     * The default is null, so that the order is maintained (except for totalValueIds, which are moved to the top).
```
kring, line 206: Probably not a big deal, but the code says the default is `false`, not `null`.

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -529,39 +575,55 @@ function getTotalValueIdsForDimensionId(item, dimensionId) {
...
+    // filterList can be either item.whitelist or item.blacklist.
+    // set isWhiteList false if is blacklist, so that the return values are negated (except for a missing list).
+    function getFilter(filterList, dimensionId, isWhiteList) {
+        var thisIdsFilterList = filterList[dimensionId];
+        if (!defined(thisIdsFilterList)) {
             return function() { return true; };
         }
         try {
-            var thisIdsWhitelistRegExps = thisIdsWhitelist.map(string => new RegExp('^' + string + '$'));
+            var thisIdsRegExps = thisIdsFilterList.map(string => new RegExp('^' + string + '$'));
```
kring, line 599: If you're supporting regular expressions, why add the implicit `^` and `$`?  Also, you should mention in the doc for the whitelist and blacklist that it's interpreted as a regular expression.  Otherwise people will be surprised when `.` matches anything rather than requiring a dot.

lib/Models/SdmxJsonCatalogItem.js
RacingTadpole: Nice catch, it should be `null`.

lib/Models/SdmxJsonCatalogItem.js
```diff
@@ -529,39 +575,55 @@ function getTotalValueIdsForDimensionId(item, dimensionId) {
...
+                if (isWhiteList) {
+                    return isPresent;
+                }
+                return !isPresent;
             };
         }
         return function(value) {
             // Test as a straight string, and if that fails, as a regular expression.
-            return thisIdsWhitelist.indexOf(value.id) >= 0 || thisIdsWhitelistRegExps.map(regExp => value.id.match(regExp)).some(defined);
+            var isPresent = thisIdsFilterList.indexOf(value.id) >= 0 || thisIdsRegExps.map(regExp => value.id.match(regExp)).some(defined);
```
kring: Slightly dodgy use of `thisIdsRegExps` here, because it's defined in a nested scope.  Of course, JavaScript doesn't pay attention to such things, but it's still a bit strange.  I think moving the `return function(value) ...` inside the `try` as well would make it clearer.

lib/Models/SdmxJsonCatalogItem.js
RacingTadpole: Yep, will add to the docs.
On ^ and $ - good question.  The only problem is an expression like "T" would then mean anything containing a "T", and that's the most common sort of expression.  I think that will trip people up.

kring: The only existing example I know of where we allow people to specify a regular expression in the init file is `CkanCatalogGroup`'s various format properties.  There, we construct it like this: `new RegExp(userSuppliedValue, 'i')`.  I think that's pretty reasonable if you're advertising it as a regexp.  Adding `^` and `$` makes it a lot harder to blacklist everything that starts with a certain string, for example.

PR #2383, opened 2017-02-21
Fix uneven map button height

### LINKED ISSUE #2250: Related Maps and About buttons smaller
In NM and related instances, the related maps and about buttons are smaller than the others by 2px, see screenshot below. This is in the released version of NM also.

size_difference

tephenavies:
Fixes #2250.


meh9:
Looks great.

PR #2384, opened 2017-02-23
Merge `info` instead of replacing it entirely

### LINKED ISSUE #2379: Adding `info` property to ckan-resource hides several other metadata fields
We have a number of `ckan-resource` catalogue items that point at MapServer data.gov.au resources. We want to add an extra metadata field to these using an `itemProperties` `info` property. When added, several other of the metadata fields are no longer shown.

The original About This Dataset looks like:

> screen shot 2017-02-15 at 18 17 01

Adding this to the `itemProperties` of the `ckan-resource`:

```
                "info": [
                  {
                    "name": "foo",
                    "content": "bar"
                  }
                ]
```

The result on the About This Dataset:

> screen shot 2017-02-15 at 18 18 37

The fields Dataset Description, Resource Description, and Licence are no longer shown, however the Service Description field is still being shown.

kring:
Fixes #2379

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 4.10.2
+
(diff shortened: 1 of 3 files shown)


meh9 (approved):
Looks awesome! :D

PR #2387, opened 2017-02-24
Stop csv files with a date column sometimes failing if a date is missing

RacingTadpole:
As reported by @meh9 recently.

Also updated all the corresponding tests to ensure nulls are handled properly.


meh9:
Fantastic!

PR #2400, opened 2017-03-08
Close polygons used as input to `CatalogFunction`.

### LINKED ISSUE #2382: Close polygons used for WPS input
At the moment when we submit Geojson of a closed polygon as input to a WPS process we only send 3 points for a triangle. However, looking at the Geojson spec it appears that closed polygons have to repeat the initial point in order to make them closed:

http://geojson.org/geojson-spec.html#id4

We should make our closed polygons do this.

kring:
Fixes #2382

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 4.10.4
+
+* Close polygons used as input to a `CatalogFunction` by making the last position the same as the first one.
+
 ### 4.10.3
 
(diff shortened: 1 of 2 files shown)


kring:
CC @meh9 
Heads up, master (and this branch) now have a new version of Cesium, so when testing this you'll need to make sure you only have one copy of `terriajs-cesium` installed and that it is version 1.31.0.

meh9 (approved):
Awesome, and so little code!

PR #2402, opened 2017-03-09
Webpack2 

tephenavies:
Use webpack 2 to build TerriaJS. Requires TerriaJS/cesium#23.


tephenavies:
Tests failing due to loading of cesium workers with worker-loader (fixed in TerriaJS/cesium#23).

kring:
I'm seeing test failures in IE11 that look like this:

image

It's certainly true that IE11 doesn't have a `Promise` type (that's an ES2015 feature).  Less obvious is why Webpack is using it, since presumably it's meant to work with pre-ES2015 browsers too, right?

kring:
Oh this explains it:
> nb require.ensure now depends upon native Promises. If using require.ensure in an environment that lacks them then you will need a polyfill.

https://webpack.js.org/guides/migrating/#require-ensure-and-amd-require-is-asynchronous

Ideally we'd polyfill `Promise` using the `when.js` we already have rather than adding another promise implementation, but I don't know how realistic that is.

tephenavies:
Tests passing on IE9 (according to Travis). Probably ready to 🚢 . Could merge before or after this release, I'm fine with either.

Closes #2349, #2350, #2359, #2373, #2381, #2388, #2389, #2398, #2408.

PR #2404, opened 2017-03-09
Add nameInCatalog property

### LINKED ISSUE #2378: Allow different names for the Catalog and once added
Especially with the new ABS data, often the catalog item name winds up only making sense in context, eg:
image

This is not ideal on the left of the screen, or for Feature info headings - especially if the sections are collapsed and there's more than one.
image

It would be nice if the name of the item in the now-viewing panel and feature info could be different to the catalog item name.

RacingTadpole:
Fixes #2378. This PR is useful in the case when it would be redundant to use the full name in the catalog given the group names above it.

You can try this out using the updated `csv.json`, ie. http://localhost:3001/#build/terriajs/test/init/csv.json

image

In the screenshot note the name on the left is the same catalog item as selected on the right, but the naming is different.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 4.10.4
+
+* Added support for a new `nameInCatalog` property on all catalog members which overrides `name` when displayed in the catalog, if present.
+
 ### 4.10.3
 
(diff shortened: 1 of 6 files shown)


RacingTadpole:
Now with added breadcrumbs as hover text for the workbench item name
image
(present whether the name is different or not)

PR #2407, opened 2017-03-09
Make charts respect all yAxisMax for all charting cases instead of ju…

meh9:
…st a single chart case  https://github.com/TerriaJS/geoglam-nm/issues/6

### DIFF
--- a/lib/Charts/Scales.js
+++ b/lib/Charts/Scales.js
@@ -60,14 +60,18 @@ const Scales = {
                     const dataWithTheseUnits = data.filter(l=>(l.units === theseUnits));
 
                     // Override y-domain if user has requested it.
-                    if (dataWithTheseUnits.length === 1 && defined(dataWithTheseUnits[0].yAxisMin)) {
-                        thisYDomain[0] = dataWithTheseUnits[0].yAxisMin;
+
+                    const yAxisMin = Math.min.apply(Math, dataWithTheseUnits.filter(d => defined(d.yAxisMin)).map(d => d.yAxisMin));
+                    const yAxisMax = Math.max.apply(Math, dataWithTheseUnits.filter(d => defined(d.yAxisMax)).map(d => d.yAxisMax));
+                    
+                    if (yAxisMin === yAxisMin && yAxisMin !== Infinity && yAxisMin !== -Infinity && thisYDomain[0] < yAxisMin) {
+                        thisYDomain[0] = yAxisMin;
                     }
-                    if (dataWithTheseUnits.length === 1 && defined(dataWithTheseUnits[0].yAxisMax)) {
-                        thisYDomain[1] = dataWithTheseUnits[0].yAxisMax;
+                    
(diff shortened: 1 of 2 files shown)


tephenavies:
@meh9: I'm not really sure what the ymax does to graphs. Does it cut them off so that graphs look they they plateau at ymax when they exceed that value?

If so, do these changes mean that if you display 2 graphs, one with a ymax and ymin defined (in catatlog config), and one that doesn't have ymax and ymin that both will be trimmed to ymin < y < ymax range?

meh9:
@steve9164 I assume so when the two graphs share the same unit. I'll give it a test and see what happens.

meh9:
Yes, it sets it for all graphs of that unit, and if several different `yAxisMax` are set it appears to use the `max(a, b)` of them.

tephenavies:
I think this should only set the yMax or yMin when they are specified for all graphs of that unit. ~~That change would mean that a graph's shape can only be deformed by vertical and horizontal scaling (+translation) by adding another graph, rather than allowing graphs to be truncated by the settings of other graphs.~~ Otherwise a graph can be truncated by the settings of another graph being viewed.

tephenavies:
An example I can think of that would be a problem with this change is this:
- A graph of watertable depth over time (truncated with a maximum of 0m above ground)
- User adds a graph (via catalog, WPS or other service or from expanding feature info) which is in the range 0-50m

That second graph is either not shown at all when added, or displayed as a solid line on the 0m mark, both of which might suggest the data is bad, or the service isn't working.

**Also, adding a graph whose [yMin, yMax] interval isn't the same as a currently displayed graph will cause both graphs to look (in some cases) very different from what each would look like on their own (with either of the suggested methods)**

Maybe @RacingTadpole has an opinion on how displaying multiple graphs with different [yMin, yMax] intervals should be done?

meh9:
I'm happy with any solution that works - note that the current state is that `yAxisMax` doesn't work at all, which is the worst of all states.

axman6:
At the moment, we have a specific use case for this fix and a deadline we need to meet which is very soon, so if it's possible to merge something non-optimal with the understanding a better/more general solution will need to be sorted out, that would be preferable.

Perhaps the way to fix this is for any series with a ymin/max to have its units treated as different to one without, so steven's case where two sets using the same units but one of which should be capped and span the whole height leads to two y axes, one for the capped metres, and one for the non-capped metres. My 2¢.

RacingTadpole:
I'm happy to merge this if this fixes your immediate problem, and we can write a new PR to fix the problems Steve's identified.

RacingTadpole:
Give me a sec to run a few tests on it first

RacingTadpole:
The first test I've tried is to force the max to be 20, on a plot where it only goes to 9. It fails, because it only applies the max if it is less than the max of the data (`if (thisYDomain[1] > yAxisMax)`).  Is that ok with you?

RacingTadpole:
When the max does apply, the chart doesn't respect its own boundaries. Let's add that to the next PR too.
image
Or maybe that's quite a good way to show that there is data there, just off the screen (as opposed to missing data ... not that we handle missing data right yet)

meh9:
Thanks guys, I appreciate there are problems, and also your help. :)

PR #2411, opened 2017-03-13
Added `terria.markdown` for markdown-support for data fields

### LINKED ISSUE #2386: URLs can be wrongly parsed in FeatureInfo
Two URLs with an "and" between can be parsed as a single URL. See https://github.com/TerriaJS/neii-viewer/issues/90.

Only in the normal view, not in the raw data view.
image

RacingTadpole:
Fixes #2386.  The problem is that the feature info template assumed `{{siteUrl}}` was a single tag, and wrapping it in `<a>`, but it actually contained two links in plain text. But if you just display it as `{{siteUrl}}` it comes out as text with no links.

With this PR you can do `{{#terria.markdown}}{{siteUrl}}{{/terria.markdown}}`.

Slightly worried that there is actually no problem here at all, and we don't need this ability. But I can't see how to do it otherwise, eg. this spec fails (with `'description': 'Check out www.example.com/1 and www.example.com/2.'`):
```
        it('does not render fields with markdown', function() {
            const template = '<div>Description: {{description}}</div>';
            const section = <FeatureInfoSection feature={feature} isOpen={true} clock={terria.clock} template={template} viewState={viewState}/>;
            const result = getShallowRenderedOutput(section);
            expect(findAllWithType(result, 'a').length).toEqual(2);
        });
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 4.10.5
+
+* Added support for the `terria.markdown` function in feature info templates, so that markdown-formatted data fields can be displayed.
+
 ### 4.10.4
 
 * Added the ability for `CkanCatalogGroup` to receive results in pages, rather than all in one request.  This will happen automatically when the server returns partial results.
--- a/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
@@ -13,6 +13,7 @@ import CustomComponents from '../Custom/CustomComponents';
 import FeatureInfoDownload from './FeatureInfoDownload';
 import formatNumberForLocale from '../../Core/formatNumberForLocale';
 import Icon from '../Icon.jsx';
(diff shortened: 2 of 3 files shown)


RacingTadpole:
After discussion with @kring we decided this PR is not necessary.

The ultimate issue is that once you start using html tags in a feature info template, it assumes you are in control of all links - but in this case, we still want terria to do the "linkifying" for us.

However a [markdown](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet)-formatted template doesn't make that same assumption. So, the solution is to use markdown - not html - in the feature info template.  You can define a table in markdown using the format:

```
|Column Heading 1|Column Heading 2|
|--------|--------|
|Site name|Sydney|
|Site urls|example.com/1 and example.com/2|
```

etc.  Also `<h1>x</h1>` is just `# x` in markdown.

Also note you can mix html and markdown side-by-side, but once you're inside an html tag, you can't use markdown.

See https://github.com/TerriaJS/neii-viewer/issues/90#issuecomment-286267363 for the solution to the example which led to this PR.

PR #2412, opened 2017-03-14
Rename misleading "full screen" button.

kring:
Fixes TerriaJS/nationalmap#429

image

image


RacingTadpole:
Wow, when you said a small change, you meant it! :-)
This is totally fine but I can bikeshed it a bit - "Show the Workbench" is quite long, would "Show Workbench" suffice?

I also noticed a bug while testing on small screens - when the workbench is hidden, on a small screen you get an empty gray bar along the top. I'll open a new issue for that.

kring:
Ok, done.

PR #2414, opened 2017-03-14
Improve feedback form resize behavior.

kring:
Don't ruin the layout when the text area is resized.

Before:
image

After:
image

Fixes TerriaJS/nationalmap#432


RacingTadpole:
Much nicer!

PR #2419, opened 2017-03-14
Show chart previews in About This Dataset, and support xAxis and x-column.

### LINKED ISSUE #2365: Preview panel for chart data
Eg. drag a file containing this onto the map:
```
date,value
1/01/1970,-10000000
1/04/1970,-20000000
1/06/1970,-30000000
1/09/1970,-25000000
1/01/1971,-10000000
1/04/1971,-20000000
1/07/1971,-30000000
1/10/1971,-25000000
```
and in the resulting Add Data panel view, you see this:
image
because it has `isMappable == false` (so cannot display a preview map). It would be cool if it showed a preview of the chart instead.

### LINKED ISSUE #2417: Charts `x-column` attribute doesn't work
This was never finished.
Also, we should get rid of `x` in this:
image
Note this problem doesn't happen when the x-axis is time.

RacingTadpole:
Fixes #2417.
Fixes #2365.
Also fixes y-columns, which weren't properly being respected.
Fixes the color of the download button on the feature info panel.
Also fixes some other subtle bugs we'd never noticed.
As a result I think the code is a bit cleaner, which is nice.
And finally makes a completely unrelated fix to the formatting of the "slow performance" alert, which markdown was interpreting as a very wide `pre` `code` block.

Try this out using http://localhost:3001/#build/terriajs/test/init/charts.json and in particular the various `XY Plots` and the feature info panel for the `Train Stations`. It also works for drag-and-dropped csvs.


kring:
:crisp:!

PR #2432, opened 2017-03-17
Use updated custom czml properties (Cesium PR#5105)

RacingTadpole:
We'll need this once https://github.com/AnalyticalGraphicsInc/cesium/pull/5105 is merged into https://github.com/TerriaJS/cesium/tree/terriajs. (Though it will do no harm to merge it now.)

In my earlier way of doing this, `properties` could either be an object directly or something with a `getValue`. With the Cesium PR, CZML `properties` is a `PropertyBag`, which has `getValue` that returns an object.

So this PR handles the new additional possibility that `properties.getValue` is not a single value, but itself is an object.

Test this with http://localhost:3001/#build/terriajs/test/init/czml.json.

### DIFF
--- a/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
@@ -421,11 +421,9 @@ const simpleStyleIdentifiers = ['title', 'description',
 function describeFromProperties(properties, time) {
     let html = '';
     if (typeof properties.getValue === 'function') {
-        const singleValue = properties.getValue(time);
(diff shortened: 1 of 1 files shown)


RacingTadpole:
@kring - we'll need this one too if we use the latest cesium, and it does no harm in the meantime.

kring:
Oh ok, I didn't realize this was backward compatible with older versions of Cesium, too.  We're not upgrading for this release, so might as well merge it after the release anyway, though, right?

RacingTadpole:
that's fine too :-)

PR #2433, opened 2017-03-17
Added in score column to GNAFAddressGeocoder csv output and map

tobybellwood:
Only file change is to GNAFAddressGeocoder.


tobybellwood:
Happy for this to be chopped apart - just PR'ing it for the record!

kring:
Thanks @tobybellwood!

PR #2434, opened 2017-03-17
Remove a bunch of deprecated code.

kring:
Because the Webpack 2 change necessitates a bump to TerriaJS 5.0.


kring:
@RacingTadpole you're probably the best person to review and merge this.

kring:
Looks like some of the deprecation removing broke a bunch of specs?

CHANGES.md
```diff
@@ -5,6 +5,16 @@ Change Log
...
+  * Removed the following previously-deprecated modules: `registerKnockoutBindings`, `AsyncFunctionResultCatalogItem`, `PlacesLikeMeFunction`, `SpatialDetailingFunction`, and `WhyAmISpecialFunction`.
+  * Removed `lib/Sass/StandardUserInterface.scss`.
+  * Removed the previously-deprecated third pararameter, `getColorCallback`, of `DisplayVariablesConcept`.
+  * Removed the following previously-deprecated properties from `TableColumn`: `indicesIntoUniqueValues`, `indicesOrValues`, `indicesOrNumericalValues`, and `usesIndicesIntoUniqueValues`.
+  * Removed the previously-deprecated `dataSetID` property from `AbsIttCatalogItem`.
+  * Removed the previously-deprecated `allowGroups` property from `CkanCatalogItem`.
+  * Removed the previously-deprecated `setRegionColumnType` function from `RegionMapping`.
+  * Removed the previously-deprecated `options.regionMappingDefinitionsUrl` parameter from the `Terria` constructor.
```
RacingTadpole: Can I suggest you include in here what to replace it with, since in the new version, the deprecation messages are gone, so it isn't obvious any more.

RacingTadpole:
OK, the problem was this: `CsvCatalogItemSpec` was using the now deprecated `terria.regionMappingDefinitionsUrl` to point to a test regionMapping file. So with this PR, it silently started using the default one instead, which then conflicts with `AbsIttCatalogItemSpec`'s stubbing of the default one. Since `Abs` doesn't stub the region types that `Csv` needs, it fails.  

If you run the `Csv` specs in isolation, the `Abs` stubs aren't included and then the specs pass - I'm not sure why.

CHANGES.md
kring: Ok, done.  You may want to take a glance that I haven't misrepresented how to update the CSV-related stuff.

RacingTadpole:
Great, thanks. I've just added one more line to CHANGES:

* Removed the previously-deprecated `regionMapping.regionDetails[].column` and `.disambigColumn`. Use `.columnName` and `.disambigColumnName` instead.

PR #2440, opened 2017-03-20
Fix crash when starting in 2D and switching to 3D Smooth.

kring:
I introduced this bug when I changed TerriaJS to defer load of 3D code when starting in 2D.  It made `terria.cesium` set asynchronously, so attempting to immediately set the globe's terrain provider would crash.  Since it's unnecessary to do this anyway when the viewer was just created, I adjusted the logic accordingly.


kring:
@meh9 this fixes the compass not appearing in 3D smooth in Geoglam.

meh9 (approved):

PR #2449, opened 2017-03-28
As augmented virtuality

a-stacey:
Note the developer end of this feature is labeled as Augmented Virtuality because this is the correct technical term for the feature that is implemented here, however all of the user facing documentation shows the feature as Augmented Reality as the term is more palatable.

Two public functions on the AugmentedVirtuaility class don't have test cases:
.hover()
.moveTo()

Have attempted to write test cases for these functions using Cesium widget and without using mock objects but neither of these tests cases came out cleanly and were overly complex. Happy to take further advice on possible solutions.

Further there are no tests around getting the alpha, beta, gamma and screen orientation from the browser, not sure if it is possible to test anything nicely and cleanly here.

As this is my first large feature I expect that there are probably lots of issues against style and codbase norms so please just let me know and I will try to clean it up as needed.

Will submit a corresponding pull request for TerriaMap to enable this feature.


tephenavies:
@a-stacey What are the conditions for displaying the tool. Something along the lines of:
- Cesium renderer
- Experimental features enabled

Is there any condition on the device being mobile, or any other condition?

a-stacey:
@steve9164 Yes you must have:
- Experimental features enabled (you will see this in the branch in the TerriaMap repo, along with adding the feature to the experimental menu).
- 3D Smooth or 3D Terrain enabled.
- The user agent must report as Android or IOS device (since it only makes sense on mobile devices and not desktop agents and screen size is not necessarily a determining factor nor the sensor events as the desktop reports on this even though they don't update). As you will see in code the plan is to extend this to all mobile browsers in future, but currently I have only found Android and IOS devices to test on, so am reluctant to enable other devices until we have found some suitable hardware to test on or the module has undergone more use and testing (see .suitableBrowser() function comments for more information here).

You can simulate on the desktop by turning on device emulation. In Chrome from the developer toolbar select "toggle device toolbar", from the list of devices select something like Galaxy S5, then load the page (the page must have the user agent set when it is loaded or at least by the time that either 3D smooth or 3D terrain is enabled). You can emulate the sensor values by selecting "More tools" and the selecting the "Sensors" pane from the Chrome developer toolbar and then selecting an orientation or manually dragging the device orientation shown using your mouse.

In hardware on Android or IOS device you should just be able to load this PR along with the PR for TerriaMap and switch to 3D Smooth of 3D Terrain. A demo of the live code in this PR is currently deployed on https://arvr.terria.io which you can use to verify your setup, but please don't rely on this demo staying in place is it was setup for debug and demo purposes only and is subject to change or go down at any time. From a local build I have been port forwarding to a USB tethered android device using port forwarding over ADB for development (I can provide detailed instructions here if you need for getting this setup working). I have not developed locally for IOS so I can't give anymore information here and only tested IOS using the like above.

Hopefully all of this makes sense, but let me know if you need any more information.

tephenavies (commented):
I'll push some other changes.

lib/Models/AugmentedVirtuality.js
```diff
@@ -0,0 +1,676 @@
...
+    // The following is disabled because chrome does not currently support deviceorientationabsolute correctly:
+    // if ('ondeviceorientationabsolute' in window)
+    // {
+    //     window.addEventListener('deviceorientationabsolute', function(event) {that._orientationUpdate(event);} );
+    //     // console.log('Debug: Device Orientation Absolute');
+    // }
+    // else
+    if ('ondeviceorientation' in window) {
+        window.addEventListener('deviceorientation', function(event) {that._storeOrientation(event);});
+        // console.log('Debug: Device Orientation');
```
tephenavies: Please remove these commented debugging calls

a-stacey:
Thanks for the feedback.

I have updated the outstanding issues that I am aware of.

Please have a look at d9964b8955aea42cbe70cb0e5db3bbf7b50cffbe and indicate if any further work is needed here either from my module or to fix the issue upstream (log bug).

Also it seems that master has been updated for new react versions but that the upgrade in the codebase is only partial and I still see a lot of warnings elsewhere in the codebase. I have attempted to update the style in this patch to match the remainder of the codebase, but please let me know if there is anything that I have done wrong here or missed.

tephenavies:
Close in favour of #2513 (which has Travis tests).

PR #2455, opened 2017-03-31
Ignore leap seconds when evaluating WMS time dimension periodicity

### LINKED ISSUE #2452: Rounding issue with large "maxRefreshIntervals"
I increased the value of `maxRefreshIntervals` from 1,000 to 10,000 and it worked fine, but I realized I had 35,000 time steps in the model run.  So I increased the value to 40,000, but then I get an error in the WMS request, because it is asking for a time value of `2017-03-30T00:59:58Z`, but there is no actual data at that time (since the data is actually at `2017-03-30T01:00:00Z`):

http://gamone.whoi.edu/terriajs/proxy/http://tds.marine.rutgers.edu/thredds/wms/roms/espresso/2013_da/his/ESPRESSO_Real-Time_v2_History_Best?time=2017-03-30T00%3A59%3A58Z&srs=EPSG%3A3857&transparent=true&format=image%2Fpng&exceptions=application%2Fvnd.ogc.se_xml&styles=&tiled=true&feature_count=101&colorscalerange=-50%2C50&service=WMS&version=1.1.1&request=GetMap&layers=salt&bbox=-10018754.171394622%2C0%2C0%2C10018754.171394622&width=256&height=256

Here's the "share" url for this test case: https://tinyurl.com/lcdoy7y

kring:
`WebMapServiceCatalogItem` now ignores [leap seconds](https://en.wikipedia.org/wiki/Leap_second) when evaluating ISO8601 periods in a time dimension.  As a result, 2 hours after `2016-06-30T23:00:00Z` is now `2016-07-01T01:00:00Z` instead of `2016-07-01T00:59:59Z` even though a leap second at the end of June 2016 makes that technically 2 hours and 1 second.  We expect that this is more likely to align with the expectations of WMS server software.

Fixes #2452

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 5.0.2
+
+* `WebMapServiceCatalogItem` now ignores [leap seconds](https://en.wikipedia.org/wiki/Leap_second) when evaluating ISO8601 periods in a time dimension.  As a result, 2 hours after `2016-06-30T23:00:00Z` is now `2016-07-01T01:00:00Z` instead of `2016-07-01T00:59:59Z` even though a leap second at the end of June 2016 makes that technically 2 hours and 1 second.  We expect that this is more likely to align with the expectations of WMS server software.
+
 ### 5.0.1
 
 * Breaking changes:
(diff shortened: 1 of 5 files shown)


kring:
Just realized this doesn't work with periods like `P1D2H` (one day two hours) even though it used to.  It also doesn't work with things like `P1M` (1 month) but it never did.  Only days, hours, minutes, and seconds work.  I'll fix the former problem, at least, and add a test.

kring:
This is ready now.

kring:
Superceded by #2456, closing.

PR #2456, opened 2017-03-31
Improvements to WMS TIME dimension handling

### LINKED ISSUE #2452: Rounding issue with large "maxRefreshIntervals"

kring:
From CHANGES.md:

* Fixed a bug that prevented `WebMapServiceCatalogItem` from acting as a time-dynamic layer when the time dimension was inherited from a parent layer.
* `WebMapServiceCatalogItem` now supports WMS 1.1.1 style dimensions (with an `Extent` element) in addition to the 1.3.0 style (`Dimension` only) style.
* `WebMapServiceCatalogItem` now passes dates only (rather than dates and times) to the server when the TIME dimension uses the `start/stop/period` form, `start` is a date only, and `period` does not include hours, minutes, or seconds.
* `WebMapServiceCatalogItem` now supports years and months (in addition to days, hours, minutes, and seconds) in the period specified of a TIME dimension.
* `WebMapServiceCatalogItem` now ignores [leap seconds](https://en.wikipedia.org/wiki/Leap_second) when evaluating ISO8601 periods in a time dimension.  As a result, 2 hours after `2016-06-30T23:00:00Z` is now `2016-07-01T01:00:00Z` instead of `2016-07-01T00:59:59Z` even though a leap second at the end of June 2016 makes that technically 2 hours and 1 second.  We expect that this is more likely to align with the expectations of WMS server software.

Fixes #2452


rsignell-usgs:
I can confirm that the combination of PR https://github.com/TerriaJS/terriajs/pull/2456 and PR https://github.com/TerriaJS/terriajs/pull/2450 fix my issue #2452 with WMS datasets with 30,000+ time intervals.  Here's the latest forecast for sea surface salinity on the Mid-Atlantic US East Coast displaying correctly as proof!

kring:
Anyone want to review this before I just hit merge myself? :)

PR #2459, opened 2017-04-05
Add a configuration parameter which specifies which view mode to use as default when running on a mobile platform.

a-stacey:
Now that I have added Augmented Virtuality mode it has been requested that there is an option to specify using 3DSmooth as the default on mobile platforms to make this feature more accessible from some instances.

This patch is somewhat conservative in that it preserves the previous functionality and yields to that specification if it is provided (that being if options.defaultTo2D is set). The reason for this is simply that I didn't know whether other implementations made use of this option for purposes other then mobile configuration. If this functionality is not needed beyond this then the following lines should also be removed:
             if (options.defaultTo2D && !defined(that.userProperties.map)) {
                 that.viewerMode = ViewerMode.Leaflet;
             }

I have not added test cases for this commit as I couldn't find test cases for the configuration file, if I have overlooked this please let me know. 

I have submitted a corresponding PR for TerriaMap which changes the default behavior for configuration of 2D mode from

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@ Change Log
 ==========
 
 ### 5.0.2
-* Added configuration option to specify default view mode when running on a mobile platform.
+
+* Added option to specify `mobileDefaultViewerMode` in the `parameters` section of `config.json` to specify the default view mode when running on a mobile platform.
 
 ### 5.0.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,9 @@
 Change Log
 ==========
 
+### 5.0.2
+* Added configuration option to specify default view mode when running on a mobile platform.
+
 ### 5.0.1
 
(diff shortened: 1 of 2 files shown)


a-stacey:
I'm not sure what I'm doing wrong in that I signed the CLA last time, but am again being prompted for it this time? It looks like somehow the commits are not associated with my github account? On looking at the agreement again it seems that it has been updated so possibly it was just the above message that was confusing along with the timing of my two PRs.

Please advise.

Thanks

AlexGilleran:
Looks like you committed under your data61 email but it isn't associated with your github user?

a-stacey:
Hey true thanks, just hadn't updated the domain.

kring:
Thanks @a-stacey.  Can you please also update https://github.com/TerriaJS/terriajs/blob/master/doc/customizing/client-side-config.md to include this new option?

PR #2475, opened 2017-04-12
Fix lat/lon coordinate wrapping in IE11

### LINKED ISSUE #2446: Latitude / longitude bar is cut off in IE11
image

Reported in TerriaJS/NationalMap-Feedback#347.

kring:
Fixes #2446


kring:
I'm just going to merge this because no one else has IE11, but @chloeleichen you should take a quick glance at the CSS change in case it might create some problem I haven't thought of.

PR #2476, opened 2017-04-12
Fix deprecated warnings with React 15.5

RacingTadpole:
There remains one dependency (that I'm aware of) that means we still get the deprecation warnings, unfortunately: https://github.com/jasonslyvia/react-anything-sortable - and it is not under active development.
Also, needs an update to the latest version of redbox in TerriaMap / NationalMap etc.


RacingTadpole:
woo hoo! A green tick - @kring.

PR #2498, opened 2017-05-02
Fix bug where csv files with missing dates failed to load

### LINKED ISSUE #2497: csv files with missing dates fail to load
Eg.
```
postcode,date,value
3121,2015-08-07,1
3122,2015-08-07,2
3123,2015-08-07,3
3124,2015-08-07,4
3121,2015-08-08,4
3122,,3
3121,2015-08-09,2
3122,2015-08-09,2
3123,2015-08-09,3
3124,2015-08-09,3
```
leads to an error in RegionMapping.js when it tries to query the timeInterval.

RacingTadpole:
Fixes #2497


meh9:
@RacingTadpole in your testcase from #2497, if I change the value for the line that has no date to something higher than the others, it shows that it still gets included in the legend even though it is not possible to show that value. As it's not possible to show it, I don't think it should be included in the legend. What do you think?

RacingTadpole:
That makes sense. It's a bit harder to do though, because I've tried to write the TableStructure so that each column is fairly independent of the others. So when the legend looks for the range of possible values, it just gets the min and max from the column.  So this would require implementing a bit more mixing of info between columns.
I agree it's worthwhile, but it's a slightly bigger job to implement.

RacingTadpole:
Can I suggest we merge this, which prevents an ugly crash?  I'll log a new issue for @meh9's comment.

meh9:
Merged!

PR #2507, opened 2017-05-04
Show disclaimers triggered by load (i.e. a share link)

kring:


meh9:
@kring, the layer disclaimer comes up before the global disclaimer, would it be easy to reverse that order?

kring:
Hmm not sure, I'll have a look.

kring:
Ok TerriaJS supports a global disclaimer directly now.  So, you'll need to add a new parameter to `terria.start` in your index.js:
```js
globalDisclaimerHtml: require('./lib/Views/GlobalDisclaimer.html'),
```

You can also optionally add a development server preamble:

```js
developmentDisclaimerPreambleHtml: require('./lib/Views/DevelopmentDisclaimerPreamble.html')
```

And delete the code starting with `if (defined(terria.configParameters.globalDisclaimer)) {` in your index.js.

Everything else is configured in config.json under `"globalDisclaimer"`:

* `enableOnLocalhost` - true to show the disclaimer even on localhost
* `devHostRegex` - if this regex matches `window.location.hostname`, the development preamble (if any) will be shown.
* `prodHostRegex` - if this regex does not match `window.location.hostname`, the develoment preamble (if any) will be shown.
* `title` - the title to show on the disclaimer dialog.  Defaults to `Disclaimer`.
* `buttonTitle` - the text to show on the button to dismiss the dialog.  Defaults to `I Agree`.
* `width` - the width of the dialog in CSS pixels.  It will never be bigger than the browser window no matter what you put here.  Defaults to 600.
* `height` - the height of the dialog in CSS pixels.  It will never be bigger than the browser window no matter what you put here.  Defaults to 550.
* `hideUi` - true to hide the user interface while showing the dialog.  Defaults to false.

kring:
@meh9 this is ready when you are. :)

meh9:
Works great!

PR #2509, opened 2017-05-09
Refresh Cesium data source when load recreates it.

kring:
This fixes some dodgy behavior when a `DataSourceCatalogItem` is loaded again (because one of its "values that influence load" changed) after it is enabled and possibly shown.

1. Use the Chome developer tools, Sensors tool to set a custom Geolocation.  Something like latitude -25, longitude 135.
2. Click "go to my location" in the TerriaJS user interface and allow the site to use your location if necessary.
3. Click the checkbox on the workbench next to "My Location" to hide the point.
4. Change your location in the Chrome dev tools to something nearby, like -25 latitude, 135.01 longitude.
5. Click "go to my location" again.  The map should pan to center on the new location but the point should still not be shown (so far so good).
6. Click the checkbox on the workbench again to re-show the point.

Two points will appear appear on the map after step 6.  This PR fixes that.

Another problem fixed in this PR:

1. Complete steps 1 and 2 as above.
2. With the point still shown, change your location in the Chrome dev tools.
3. Click "go to my location".

The location will not update.  This PR fixes that.

### DIFF
--- a/lib/Models/CzmlCatalogItem.js
+++ b/lib/Models/CzmlCatalogItem.js
@@ -114,8 +114,18 @@ CzmlCatalogItem.prototype._load = function() {
         var CzmlDataSource = require('terriajs-cesium/Source/DataSources/CzmlDataSource');
 
         promiseFunctionToExplicitDeferred(codeSplitDeferred, function() {
+            // If there is an existing data source, remove it first.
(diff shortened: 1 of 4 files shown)


RacingTadpole:
Update changes? :-)

PR #2512, opened 2017-05-09
Fix Feature Info chart updating for SOS

### LINKED ISSUE #2511: SOS doesn't always update feature info when you change concept
If you use the bom.json init file, and use one of the last two catalog items, and change from Annual Average to Monthly Average, you'll see the feature info doesn't update.

RacingTadpole:
Fixes #2511 .


RacingTadpole:
Actually I'm now seeing something a bit odd with one of the other SOS items, let me investigate.

RacingTadpole:
OK, the problem was that now half-updated features can appear in the feature info (but only as part of getting to the fully-updated feature), so the Feature Info section needs to be able to handle that.

This actually fixes a bug I hadn't noticed before, eg. for Storage Levels, choose Daily Average and choose a sensor near Perth. Then click Annual Average, and for the date it picks, there is no such feature.  In the past, it led to weird behaviour, like being able to click anywhere and have the cross-hairs go there (because it hit an error). Now it works properly.

RacingTadpole:
OK, this is ready for review - thanks! @kring

lib/Models/TableDataSource.js
```diff
@@ -326,7 +326,6 @@ function setOneFeaturePerRow(dataSource, tableStructure, specialColumns) {
     // If there more entities already exist than there are in the table,
     // remove the extras.
     var entities = dataSource._entityCollection;
-    entities.suspendEvents();
```
kring, line 329: Why was it necessary to remove these lines?  I have to imagine it makes loading a CSV much slower i some scenarios.

kring: Oh nevermind, I see now you just moved the calls to the calling function

PR #2528, opened 2017-05-15
Update LocateMe feature to enable toggleing for continueous updates while AR is on.

a-stacey:
As it says on the box.

Just let me know if there are any issues or if you have any questions.


lib/Models/AugmentedVirtuality.js
```diff
@@ -248,6 +249,10 @@ AugmentedVirtuality.prototype.hover = function(height, position) {
         position = camera.positionCartographic.clone();
     }
 
+    if (!defined(flyTo) || (flyTo !== false)) {
```
kring, line 252: You can replace this with `flyTo = defaultValue(flyTo, true);`.

lib/Models/AugmentedVirtuality.js
```diff
@@ -231,8 +231,9 @@ AugmentedVirtuality.prototype.toggleHoverHeight = function() {
  *                       {@link AugmentedVirtuality.MINIMUM_HOVER_HEIGHT} to avoid visual artifacts when the viewer
  *                       becomes to close to the surface.
  * @param {Cartographic} [position] The location to hover over. If not specified the current camera location will be used.
+ * @param {Boolean} [flyTo] Whether to fly to the location (true) or whether to jump to the location (false).
```
kring, line 234: Use `@param {Boolean} [flyTo=true] Whether...` to document the default.

kring:
Looks good other than those very minor comments, so I'm merging this.  You can make those small changes in master if you want.

PR #2538, opened 2017-05-24
Fix a bug with vector tiles and proxy

RacingTadpole:
If you add `forceProxy: true` to a region-mapped catalog item, it will try to hit URLs like

```http://localhost:3001/proxy/_0d/https://vector-tiles.terria.io/FID_POA_2011_AUST```

which fail.  Vector tiles should never be proxied.

Question though: another function of the proxy server is to serve http over https - is that an issue here?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 5.2.5
+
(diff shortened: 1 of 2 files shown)


tephenavies:
The vector tile server always serves over https, so this is not an issue.

PR #2541, opened 2017-05-29
button title bug

chloeleichen:
issue #2540


RacingTadpole:
@chloeleichen - not sure why but it thinks you need to sign the CLA - could you take a look?

PR #2545, opened 2017-05-30
Modify feature info download to include Lat + Lon in filename.

a-stacey:
Note: I don't particularly like that this patch effectively duplicates earlier functionality in the file:
```                const latLngInRadians = Ellipsoid.WGS84.cartesianToCartographic(this.props.position);
                propertyData.terria.coords = {
                    latitude: CesiumMath.toDegrees(latLngInRadians.latitude),
                    longitude: CesiumMath.toDegrees(latLngInRadians.longitude)
                };
```
but its not obvious to me how to redactor this to use common code more cleanly then the current implementation.


lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -175,6 +175,16 @@ const FeatureInfoSection = createReactClass({
 
     render() {
         const catalogItemName = (this.props.catalogItem && this.props.catalogItem.name) || '';
+        let baseFilename = catalogItemName;
+        if (this.props.position) {
+            const position = Ellipsoid.WGS84.cartesianToCartographic(this.props.position);
+            const latitude = CesiumMath.toDegrees(position.latitude).toFixed(5);
+            const longitude = CesiumMath.toDegrees(position.longitude).toFixed(5);
+            // Check that the catalogItemName doesn't already contain the lat lon with the same or more precision.
+            if ((typeof baseFilename !== 'string') || (baseFilename.indexOf(latitude) === -1) || (baseFilename.indexOf(longitude) === -1)) {
```
kring: This test may not be completely reliable, because `toFixed` will round the last digit.  e.g. `(11.111116).toFixed(5) === '11.11112'`.

lib/ReactViews/FeatureInfo/FeatureInfoDownload.jsx
```diff
@@ -14,7 +14,7 @@ import Styles from './feature-info-download.scss';
 const FeatureInfoDownload = createReactClass({
     propTypes: {
         data: PropTypes.object.isRequired,
-        name: PropTypes.string.isRequired,
+        baseFilename: PropTypes.string.isRequired,
```
kring: Technically, this is a breaking change.  If someone was previously using `FeatureInfoDownload` from their own code, it will no longer work after this change.  Admittedly this is a bit unlikely, so I'm not going to get too hung up on it.  But if the old name (`name`) doesn't seem too misleading, keeping it will avoid the problem.  Or, if you'd prefer to go with `baseFilename`, we need to mention the breaking change in CHANGES.md.

kring:
Looks good other than the couple of comments.  Can you also update CHANGES.md?

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
a-stacey: Well spotted, thanks for pointing this out. 

Have a look at the latest change and let me know what you think.

I came up with two alternate solutions for this , either checking for both rounded up and down values or could have had the check perform just to number of decimal places -1. Neither are ideal, because checking for both rounding seems overly complex for the situation and reducing the precision with which we check and using floor means that the check vs the result and effect is somewhat imbalanced. Have gone with the more complex solution for now.

lib/ReactViews/FeatureInfo/FeatureInfoDownload.jsx
a-stacey: Ah thanks, had not realised this could create a breaking change for upstream projects, have reverted.

PR #2570, opened 2017-06-19
Make MyLocation button optional

tephenavies:
Move MyLocation button from within TerriaJS to inside the UserInterface of each TerriaJS map that wants a location button, so that instances can choose not to have that button.

What do you think?

This has been requested by the Parliamentary Library, because the location is sometimes inaccurate and they would rather not have the button.


kring:
This is a breaking change, right?  The button will disappear in existing maps.  Can you make it a property set in config.json (`terria.configParameters`) instead maybe?

tephenavies:
Yes. I had thought of adding a configuration parameter, but most of the other optional buttons work by being included in the UserInterface module of Terria maps.

I'm happy to add the config parameter to avoid a breaking change for now.

PR #2572, opened 2017-06-20
Added the ability to share the timelines current position.

a-stacey:
Please comment on whether any tests are required for the new feature or against the patched bug. Looking at the style of test cases existing in the codebase there doesn't seem to be an tests for equivalent init features and the style of the tests for the bug look aimed at higher level rather then confused references.


kring:
Looks great!  I tweaked a few small things.

Issue #2557, opened 2017-06-14
Time UI control to select a specific time instant etc


meh9:
For a few projects we have a need for a UI control for timeseries data that allows the user to select a specific time instant from the ones being made available by the dataset.

For example, in GEOGLAM we have time series data that spans 17 years with up to 4 time instants per month (776 instants in total), in AREMI we have the Solar Satellite data (`http://gis.aremi.nationalmap.gov.au/bom/wms`) with hourly coverage of 1 year (8760 time instants). It becomes difficult to impossible to select a specific time instant using the time slider at the bottom in these cases, so we need a better way to control it.

My naive idea of what we need is a drop down in the workbench, by the layer legend, that has the time instants listed in them. The user can use the dropdown to select an instant and the time slider then jumps to that time and the layer is updated on the screen.

However, 8760 items in a dropdown is perhaps not the best user experience, either for usability or performance, so maybe we can come up with something better?

Note that this will most likely feed into the TerriaCube project, where we need to be able to select 2 time instants so that we can compare the data between them.

RacingTadpole:
I've had a dream of optionally showing only discrete times on the timeline, eg. when only yearly or monthly data is available - see https://github.com/TerriaJS/terriajs/issues/1955 (see the example link there too).  

Perhaps solving that issue could help solve this one: once you zoom into the timeline far enough that there are less than X discrete times to show (which might be no zooming at all), the timeline switches to the discrete version.

For that to be a real solution, would probably need to make the timeline zooming feature more obvious as well.

RacingTadpole:
Actually looking at how the zooming into the timeline works, I think the biggest problem is this: the labels on the timeline don't become any more fine-grained than days as you zoom in.  If they did, you'd have much less trouble realising you are now looking at hours, or minutes, or even seconds, and picking exactly which time point you want.  Eg.  the timeline starts at:
![image](https://user-images.githubusercontent.com/1757858/27467437-d9d47926-5825-11e7-8f97-a5de0a0cb158.png)
and zooming in can get you to this impossible-to-decipher situation:
![image](https://user-images.githubusercontent.com/1757858/27467458-01f4c6cc-5826-11e7-893d-8a1141a97d66.png)

Note if the original scale of the csv file is only hours, it displays sensibly:
![image](https://user-images.githubusercontent.com/1757858/27467480-204fd026-5826-11e7-898c-2401a881b0a3.png)
zooming to
[...]

PR #2576, opened 2017-06-23
Add the time to the timeline labels when zoomed in

RacingTadpole:
Fixes one big problem behind #2557.  Try this with any time-varying dataset that covers >14 days.


meh9:
This works, and I think it's an improvement.

PR #2581, opened 2017-06-26
Allow instance to disable conversion service so no data is sent outside client

tephenavies:
Conversion service can be disabled by setting the `conversionServiceBaseUrl` to `false` instead of `"convert/"`.

The message (at least at first glance) needs to be duplicated since it is possible to create an `OgrCatalogItem` without `createCatalogItemFromFileOrUrl` and the new error should be shown instead of the other file incompatible error when using `createCatalogItemFromFileOrUrl`.

### DIFF
--- a/lib/Models/OgrCatalogItem.js
+++ b/lib/Models/OgrCatalogItem.js
@@ -118,6 +118,14 @@ We recommend you upgrade to the latest version of <a href="http://www.google.com
 <a href="http://www.microsoft.com/ie" target="_blank">Microsoft Internet Explorer</a>.'
         });
     }
+    if (this.terria.configParameters.conversionServiceBaseUrl === false) {
+        // Don't allow conversion service. Duplicated in createCatalogItemFromFileOrUrl.js
+        throw new TerriaError({
+            title: 'Unsupported file type',
+            message: 'This file format is not supported by ' + this.terria.appName + '. Supported file formats include: ' +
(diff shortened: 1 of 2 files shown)


lib/Models/OgrCatalogItem.js
```diff
@@ -118,6 +118,14 @@ We recommend you upgrade to the latest version of <a href="http://www.google.com
 <a href="http://www.microsoft.com/ie" target="_blank">Microsoft Internet Explorer</a>.'
         });
     }
+    if (this.terria.configParameters.conversionServiceBaseUrl === false) {
+        // Don't allow conversion service. Duplicated in createCatalogItemFromFileOrUrl.js
+        throw new TerriaError({
+            title: 'Unsupported file type',
+            message: 'This file format is not supported by ' + this.terria.appName + '. Supported file formats include: ' +
+            '<ul><li>.geojson</li><li>.kml, .kmz</li><li>.csv (in <a href="https://github.com/NICTA/nationalmap/wiki/csv-geo-au">csv-geo-au format</a>)</li></ul>'
```
kring, line 126: I'm guessing you copied this text and URL from somewhere, but it should be https://github.com/TerriaJS/nationalmap/wiki/csv-geo-au instead of that NICTA one.

kring:
Looks good other than the one comment about the URL.  I'll merge this as-is, but if you could update that URL and also updated CHANGES.md in master, that'd be great!

PR #2582, opened 2017-06-26
Fix IE11 bug with colouring vector tiled regions

### LINKED ISSUE #2575: IE11 error with custom enum coloured csv
IE11 error with custom enum coloured csv at this line: https://github.com/TerriaJS/terriajs/blob/ba20899/lib/Models/RegionMapping.js#L690

tephenavies:
In certain cases, `color` can be a typed array which in IE11 don't have the `join` method.

Fixes #2575


kring:
Since this code looks a bit strange, it'd be helpful to add a comment saying why it was done.  Also, please update CHANGES.md.

kring:
Don't forget to update CHANGES.md.

PR #2585, opened 2017-06-28
Fix share in 2D view

### LINKED ISSUE #2529: Share screenshot does not show up when using 2D map
Works fine from the 3D map.

I'm guessing this broke with the Leaflet 1.0.3 upgrade, but I'm not sure.

image

tephenavies:
Fixes #2529.


kring:
Please add this to CHANGES.md in master.

PR #2594, opened 2017-07-11
TerriaJS/aremi-natmap#72 Update share panel with the aim of making it more user friendly.

a-stacey:
Please pay particular attention the the CSS and style of the CSS code in this PR as I have no real knowledge or experience with CSS.

### DIFF
--- a/lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
+++ b/lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
@@ -42,6 +42,24 @@ const SharePanel = createReactClass({
         };
     },
 
+    advancedOptions() {
+        return this.state.advancedIsOpen;
+    },
+
+    toggleAdvancedOptions(e) {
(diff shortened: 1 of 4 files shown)


lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -40,6 +42,24 @@ const SharePanel = createReactClass({
...
+            this.setState({
+                advancedIsOpen: false
+            });
+        } else {
+            this.setState({
+                advancedIsOpen: true
+            });
+        }
+
+        this.forceUpdate();
```
tephenavies:
```js
this.setState((prevState) => ({ 
  advancedIsOpen: !prevState.advancedIsOpen 
});
```
is the recommended way to set new state based on previous state. Also, `setState` will cause an update unless your state values are mutable, so `forceUpdate` isn't needed.

PR #2603, opened 2017-07-24
Improve Esri polygon to GeoJSON conversion

kring:
It now handles polygons with holes and with multiple outer rings.

This is for TerriaJS/geoglam-nm#16.


meh9:
Thanks @kring! :D

PR #2607, opened 2017-07-27
Use MVT for region mapping and WMS for analytics

tephenavies:
This will facilitate finally rolling vector tiles out for region mapping across all instances, yet still have region picking analytics functionality available (and more clearly indicate the dependency of analytics on region mapping definitions).

The better solution would be to completely replace WMS/WFS usage in analytics, but that would require serving regions as GeoJSON somehow from the vector tile server.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### next release (5.2.11)
+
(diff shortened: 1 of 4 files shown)


tephenavies:
Note: this PR leaves analytics region picking no more broken than it is currently.

PR #2611, opened 2017-07-31
Fix a race condition in Leaflet layer initialization

kring:
A user at [Intech Solutions](http://www.intechiq.com) reported a problem where the base map would sometimes be missing from a NationalMap instance embedded in an iframe when the map was launched in 2D mode (e.g. because it was running in a VM without WebGL support).  The cause turned out to be a race condition in Leaflet layer intialization with layers that are not immediately ready upon construction (such as Bing Maps layers).  Here's the sequence of events leading to the bug:

1. CesiumTileLayer._update is called, but the imagery provider is not yet `ready`, so it sets a timeout to call `_update` again in 100ms.
2. Before the timeout expires, `_update` is called again, and this time the imagery provider _is_ `ready`.
3. Because the imagery provider is now ready and the CesiumTileLayer is not yet `initialized`, initialization begins.
4. Initialization entails first setting up some processing to happen 100ms later.  But it first checks to see if a delayed update is already installed.  If it is, it doesn't install another one.
5. When the original timeout installed in step 1 expires, the CesiumTileLayer thinks that initialization has already happened (even though it hasn't), and the layer never ends up getting properly initialized.

This PR rearranges the initialization sequence to avoid this problem.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Change Log
 ### 5.2.10
 
 * Improved the conversion of Esri polygons to GeoJSON by `featureDataToGeoJson`.  It now correctly handles polygons with holes and with multiple outer rings.
+* Fixed a bug that could cause some layers, especially the Bing Maps basemap, to occasionally be missing from the 2D map.
 
 ### 5.2.9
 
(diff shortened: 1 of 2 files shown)


kring:
This is needed for today's NM release today, please review @steve9164. :)

PR #2613, opened 2017-08-01
Add watercolor base map

RacingTadpole:
Just having some fun.

image

image

I don't think we should really put this in all our instances - especially since it only seems to work in 2D. In 3D you get errors like ```TileProviderError: Failed to obtain image tile X: 59 Y: 38 Level: 6.```

But it does look cool. :-)

[Map tiles by <a href="http://stamen.com">Stamen Design</a>, under <a href="http://creativecommons.org/licenses/by/3.0">CC BY 3.0</a>. Data by <a href="http://openstreetmap.org">OpenStreetMap</a>, under <a href="http://creativecommons.org/licenses/by-sa/3.0">CC BY SA</a>.]


kring:
I don't mind adding this.  And it seems to be working pretty well in Cesium too at this point.  I did see it return a bunch of 503s that disabled the layer at one point, but then I couldn't reproduce it.  We can make it resistant to that sort of error by setting `ignoreUnknownTileErrors` to true, which I've just done.  So, even if this PR was opened as a joke, any objections to merging it?

rowanwins:
Nice to have but not really worth adding to the core, should be added on an instance by instance basis if required.

PR #2623, opened 2017-08-09
[TerriaJS/geoglam-nm#37] OgrCatalogItem: Update the client file sizevalidation so that it uses the server property if available.

a-stacey:


lib/Models/OgrCatalogItem.js
```diff
@@ -183,8 +183,12 @@ function loadOgrData(ogrItem, file, url) {
     // generate form to submit file for conversion
     var formData = new FormData();
     if (defined(file)) {
-        if (file.size > 1000000) {
-            errorLoading(ogrItem, 'The file size is greater than the 1Mb limit of the '+terria.appName+' conversion service.');
+        if (defined(terria.serverConfig) &&
+            defined(terria.serverConfig.config) &&
+            defined(terria.serverConfig.config.maxConversionSize) &&
+            file.size > terria.serverConfig.config.maxConversionSize) {
```
kring: If the value isn't in `serverconfig`, we should use the default value rather than skipping the check entirely, right?

PR #2627, opened 2017-08-14
Splitter improvements

kring:
@RacingTadpole this is a pull request into your branch.

What changed:
* We're no longer using `leaflet-side-by-side` at all.  Instead, we use the same React splitter component in both Cesium and Leaflet.
* The slider thumb is now a `div` instead of an `input` with `type=range`.
* Upgraded Cesium to fix a bug with the splitter on high-DPI displays.

What works:
* The latest versions of Chrome, Firefox, Internet Explorer, and Edge on desktop
* Chrome on Android
* Touch input on desktop and mobile in all the browsers above

What doesn't work:
* I'm currently using a unicode character for the face of the splitter thumb that is apparently not available on Android, so it looks like a box with an X through it.  We could swap this for an SVG, or maybe use a different font?
* Splitting ABS layers.  It doesn't provide the option.  I haven't looked into why.

No idea if this works:
* Safari on desktop and iOS
* Sharing (both per-layer split side and split position should be shared I'd say)


RacingTadpole:
Awesome - thanks @kring - looks like heaps of work in there.
@chloeleichen - re Kevin's question:
> I'm currently using a unicode character for the face of the splitter thumb that is apparently not available on Android, so it looks like a box with an X through it. We could swap this for an SVG, or maybe use a different font?

Any chance you could take that on? pretty please? :-)  I'm going to merge this into `2609_2608`, so just do straight into there (or as a PR into that).

RacingTadpole:
btw - this works on Safari.

PR #2628, opened 2017-08-15
Use terriajs-cesium 1.33.2

### LINKED ISSUE #2580: `crs` is not transferred into Feature when selected
Raised in https://github.com/TerriaJS/nationalmap/issues/517.

When the FeatureInfo panel is opened and a boundary is highlighted, a temporary geoJsonCatalogItem is created for the feature corresponding to that boundary, however its `crs` property is not copied in.  This leads to some bad boundaries if using an usual coordinate reference system, eg. EPSG:3577.

kring:
Fixes #2580 
Fixes TerriaJS/nationalmap#544
Fixes TerriaJS/nationalmap#545

Here's what's different in this Cesium version:
https://github.com/TerriaJS/cesium/compare/terriajs-1.33.0...terriajs-1.33.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "simple-statistics": "^4.1.0",
     "string-replace-webpack-plugin": "^0.1.3",
     "style-loader": "^0.18.0",
-    "terriajs-cesium": "1.33.0",
+    "terriajs-cesium": "1.33.2",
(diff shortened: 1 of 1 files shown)


kring:
@steve9164 can you merge this, since it's needed for the NM release today?

tephenavies:
Sure

PR #2641, opened 2017-08-24
Implement a splitter and "split" ability on imagery layers, with time-freeze

### LINKED ISSUE #2609: Splitter control to compare data
See the latter part of #2557 for our discussion so far.

RacingTadpole:
Don't merge this yet. I'm going to use a PR to organise the tasks remaining here:

- [x] The latest versions of Chrome, Firefox, Internet Explorer, and Edge on desktop Chrome on Android.
- [x] Touch input on desktop and mobile in all the browsers above.
- [ ] Currently using a unicode character for the face of the splitter thumb that is apparently not available on Android, so it looks like a box with an X through it. We could swap this for an SVG, or maybe use a different font? @chloeleichen?
- [ ] Splitting ABS layers. It doesn't provide the option. I haven't looked into why.
- [x] Check Safari on desktop.
- [ ] Check on iOS.
- [ ] Check sharing (both per-layer split side and split position should be shared I'd say).
- [ ] Fix sharing mini-preview window to also clip layers (currently ignores the splitter).
- [ ] Merge in Chloe's new date-time picker
- [ ] Check how stable item cloning really is when you click "split".
- [ ] Add a checkbox to unfreeze the datetime.
- [ ] Improve the toggle-splitter icon (currently just a checkbox) @chloeleichen?
- [ ] Review.

(Some tasks courtesy of @kring in https://github.com/TerriaJS/terriajs/pull/2627.)

Fixes #2609.


RacingTadpole:
OK, so the splitter only appears on ImageryLayerCatalogItems, not region-mapped items.

Why? Region-mapped data (from csvs, SDMX-JSON, SOS2 etc) is a TableCatalogItem, with a RegionMapping property (`_regionMapping`).
RegionMapping is very similar to ImageryCatalogItem. It even calls some of ImageryCatalogItem's functions, which are awkwardly defined on the class (with first argument being the instance) rather than the prototype.

It would be great to either:
- Pull out the essential "imageLayer" code into a separate class/mixin. And perhaps separate the essential "time-varying" code too, if that makes sense.  Then both `ImageryCatalogItem` and `RegionMapping` could extend from them.
Or:
- Instead of `TableCatalogItem` having `_regionMapping` and `_dataSource` (for lat/lon tables) properties, make it _be_ both, so that it is in fact an ImageryCatalogItem - but one that doesn't always have an imagery layer.

Not sure if this is worth doing, or which path is better.  @kring?

This would also be useful for some work @steve9164 is doing too.

RacingTadpole:
Don't comment in this PR, we'll never find it again - I'll copy this into its own issue.

RacingTadpole:
Closing this in favour of #2645 , which incorporates the datetime picker as well.

PR #2646, opened 2017-08-29
Added `initialTimeSource` property to `CsvCatalogItem` so it is possible to specify the value of the animation timeline at start from init files.

a-stacey:
I have factored by breaking setClockCurrentTime out of ImageryLayerCatalogItem. I've put this in a file along side ImageryLayerCatalogItem but I'm not sure if this is the best location for this reused function because I don't exactly understand the distinction in the directory layout so please review this and let me know where this should belong.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,11 @@
 Change Log
 ==========
 
+
+### 5.x.x
+
+* Added `initialTimeSource` property to `CsvCatalogItem` so it is possible to specify the value of the animation timeline at start from init files.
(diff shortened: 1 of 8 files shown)


tephenavies:
Travis tests are broken but these changes look fine, and the tests pass locally on Chrome, Safari and Firefox.

RacingTadpole:
Thanks guys. FYI @steve9164 - the tests failed because there were some lint errors. I've fixed it up on master now (#2655).

PR #2647, opened 2017-08-29
initialView is a Rectangle, do not try to take sub-property

mattfullerton:
Came across this while testing https://github.com/TerriaJS/terriajs/pull/2637 - it bumps into this code and gets an undefined because the extents are already in initialView and not initialView.rectangle.


kring:
Thanks @mattfullerton!  Can you please add yourself to `CONTRIBUTORS.md`?  Also, while I'm ok with this change, I'm curious in what situation it came up, just in case it indicates a deeper problem.  Can you describe some steps to reproduce the crash you saw?

kring:
I added you to CONTRIBUTORS.md.  Thanks again @mattfullerton!

mattfullerton:
Sorry, this dropped off my radar. It came up when we were testing https://github.com/TerriaJS/terriajs/pull/2637 - adding the setting caused a client-side crash.

PR #2664, opened 2017-09-20
Support splitting and date/time picking for region mapped layers

kring:


kring:
This is a PR into `2609_2608_all`.  I'm going to hit merge, but @RacingTadpole you should take a quick look at this.

Issue #2524, opened 2017-05-15
Clearing the location search does not repaint map


kring:
1. Make sure you're in 3D mode.
2. Search for a location and click one to show the map marker.
3. Wait a few seconds without moving the cursor over the map window.
4. Click the X next to the search box (again without moving the mouse over the map window).

The marker will not disappear until you move the mouse in the map window again.  This is caused by a missing call to  `notifyRepaintRequired`.

tephenavies:
This bug is no longer present

PR #2678, opened 2017-10-03
Fixed issue #2524

kennib:


kring:
Thanks @kennib!  There are a few places that use `SearchBox`, though, and none of them pass in a `currentViewer` prop.  Rather than teaching `SearchBox` - a low-level component - about viewers, I think it would make sense to use the existing `onClear` prop function.  Some of the places that use `SearchBox`, like `MobileHeader`, already pass in an `onClear` function and you only need to add the `notifyRepaintRequired` to the existing callback (which is easy in `MobileHeader` because it has a `terria` prop from which you can get the `currentViewer`).  In other places (`SidePanel` for example) you'll need to add a new `onClear` attribute, but again it'll be easy to call `notifyRepaintRequired` because you have access to `this.props.terria`.

kennib:
@kring is this better?

kring:
Yeah a bit better @kennib.  But the method is `notifyRepaintRequired` not `notifyRepaint`, and it's on `terria.currentViewer` not directly on `terria`.  And you're going to have problems with `this` being the wrong thing if you pass `this.props.terria.currentViewer.notifyRepaintRequired` directly as the onClear attribute without a `bind`.

It looks like maybe you haven't been able to get set up to the point where you can run TerriaMap and try it out?  Can I help with that somehow?

kennib:
I think the issues I was having are with parts of the build persisting. So it works on my version of the map, but only because it was working before.

kring:
That's strange.  Why don't you put into slack when you get a chance and we can figure it out.

kennib:
@kring I've fixed my build and the changes are hopefully good now.

kring:
This looks good, thanks @kennib!  Would you mind fixing your git history so that it's associated with your GitHub account (or just add the email address in the commit to your GitHub account) and then sign the CLA?

PR #2684, opened 2017-10-10
2588 Request tokens from terriajs-server and plumb them into ArcGisMapServerImageryProvider.

a-stacey:
There is one outstanding piece of lint associated with this code. `function requestMetadata(token) {` is inline. I have left this as is for now as I wonder whether there is a better solution to this problem then to use an intermediate function. The problem that this inline function solves is to prepend the token request in the initalisation process, and so I wonder whether there is a better solution to this problem then using the if -> inline function call.

The inline function presents particular nastiness in that to comply with the lint checks this code needs to be made top level function, however this in its currently implementation is not particularly clean in that doing in as a straight copy-paste means that the checks are separated from the implementation. To clean this up the checks can be moved into the separate function, but then the token request may happen when we know the other checks will fail. So this is not particularity nice overall and so I'm hoping that there is a cleaner higher level solution so that all of this complexity can be avoided.

### DIFF
--- a/lib/Models/ArcGisMapServerCatalogItem.js
+++ b/lib/Models/ArcGisMapServerCatalogItem.js
@@ -206,29 +206,34 @@ ArcGisMapServerCatalogItem.prototype._load = function() {
             layers = uri.segment(-1);
             this.layers = layers; // ## is this ok to do?
         }
-        var serviceUri = getBaseURI(this);
-        var layersUri = getBaseURI(this).segment(layers); // either 'layers' or a number
-        var legendUri = getBaseURI(this).segment('legend');
 
-        var serviceMetadata = this._mapServerData || getJson(this, serviceUri);
-        var layersMetadata = this._layersData || getJson(this, layersUri);
-        var legendMetadata = this._legendData || getJson(this, legendUri);
+        function requestMetadata() {
+            var serviceUri = getBaseURI(that);
+            var layersUri = getBaseURI(that).segment(layers); // either 'layers' or a number
+            var legendUri = getBaseURI(that).segment('legend');
 
-        return when.all([serviceMetadata, layersMetadata, legendMetadata]).then(function(results) {
-            if (defined(results[1].layers)) {
-                that.updateFromMetadata(results[0], results[1], results[2], false);
-            } else if (defined(results[1].id)) {
-                // Results of a single layer query. Make it look like a multi layer query result.
-                that.updateFromMetadata(results[0], {"layers": [results[1]]}, results[2], false, results[1]);
-            } else {
-                var message = defined(results[0].error) ? results[0].error.message : 'This dataset returned unusable metadata.';
-                throw new TerriaError({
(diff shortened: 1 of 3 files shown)


lib/Models/ArcGisMapServerCatalogItem.js
```diff
@@ -84,7 +86,13 @@ var ArcGisMapServerCatalogItem = function(terria) {
      */
     this.allowFeaturePicking = true;
 
-    knockout.track(this, ['layers', 'maximumScale', '_legendUrl', '_generatedLegendUrl', 'maximumScaleBeforeMessage', 'showTilesAfterMessage', 'allowFeaturePicking']);
+    /**
+     * Gets or sets the URL to use for requesting tokens.
+     * @type {String}
+     */
+    this.tokenSource = undefined;
```
kring: I think `tokenUrl` would be a better name than `tokenSource`.

lib/Models/ArcGisMapServerCatalogItem.js
```diff
@@ -195,6 +203,43 @@ function getJson(item, uri) {
...
+
+        item._lastToken = undefined;
+        throw new RuntimeError('The token server responded with an invalid token.');
+    }).otherwise(() => {
+        item._lastToken = undefined;
+        throw new RuntimeError('Unable to request a token from the token server.');
+    });
+}
+
+function appendQuery(query, addition) {
```
kring: `urijs` has a handy method to do this: http://medialize.github.io/URI.js/docs.html#search-set

lib/Models/ArcGisMapServerCatalogItem.js
```diff
@@ -195,6 +203,30 @@ function getJson(item, uri) {
...
+        if (defined(token.token)) {
+            item._lastToken = token.token;
+            return token.token;
+        }
+
+        item._lastToken = undefined;
+        throw new RuntimeError('The token server responded with an invalid token.');
+    }).otherwise(() => {
+        item._lastToken = undefined;
+        throw new RuntimeError('Unable to request a token from the token server.');
```
kring: Probably better to throw `TerriaError` here (and above).

lib/Models/ArcGisMapServerCatalogItem.js
a-stacey: Thanks for this.

f8b72f88960ec67655cc1b4805409459b3c0b78d

lib/Models/ArcGisMapServerCatalogItem.js
a-stacey: Thanks much better.

8c83029b95bfccb2748606190c2018035b1bf6ed

kring:
Just that one last comment!

lib/Models/ArcGisMapServerCatalogItem.js
a-stacey: Thanks.

5bcbe3e2fe6c8272d7239ccd0169b41a30f4e581

PR #2692, opened 2017-10-16
Multiple catalogue tabs

tephenavies:
Splits the catalog at the top level into tabs.
Used in Northern Australia map prototype. (See http://prototype-northernaustralia.terria.io/ for example)

Some choices I made:
- Display
  ```
  Select a dataset to see a preview
  
  - OR -
  
  Go to the map
  ```
  only when you first open the catalog (like currently)
- Show catalog group description page when selecting a tab (just as if it was a normal group)
- Use a tab `category` and `idWithinCategory` to store the current tab
    + This should allow easily adding other tab categories without needing to change ViewState (changes were required to ViewState here only for popular helper functions)


tephenavies:
This PR is ready now.

Travis fail seems to be a CI fail - couldn't launch an IE11 VM: 
> 18 10 2017 00:01:18.193:ERROR [launcher.sauce]: Can not start internet explorer 11.0 (Windows 7)

lib/ReactViewModels/ViewState.js
```diff
@@ -33,7 +34,8 @@ export default class ViewState {
         this.previewedItem = undefined;
         this.userDataPreviewedItem = undefined;
         this.explorerPanelIsVisible = false;
-        this.modalTabIndex = 0;
+        this.activeTabCategory = 'data-catalogue';
```
kring: Should be `data-catalog` I think.

kring:
This looks great to me!  The only thing I'd suggest (other than that "misspelling") is that, when a tab/group needs to load, that the loading be indicated in the UI somehow.  Currently the tab just looks empty while the loading is happening.  Also please add docs for the new config option to https://github.com/TerriaJS/terriajs/blob/master/doc/customizing/client-side-config.md.

kring:
I'll write a separate issue for the loading indicator and merge this.

PR #2696, opened 2017-10-19
Vector tile feature info & flickering

### LINKED ISSUE #2724: New Splitter Datepicker month layout problem
A problem with the layout in the month view of the new splitter date picker has occurred:

screen shot 2017-11-03 at 15 53 21

tephenavies:
Encompasses multiple changes:
- Feature info changes:
  + Pass vector tile properties through to feature info templates
  + JSON parse feature info template if they look like JSON (start with `[` or `{`)
- Time varying region mapping flickering:
  + Make TableStructure's time intervals full length (don't trim 1s or 5%) but half open (like `[1am, 2am)`)
  + Pre-render for the next time interval in a separate layer, then swap at the right time to reduce flicker
  + A couple of small performance improvements with time-varying csvs

Fixes #2724


kring:
1. Run NationalMap
2. Click data.gov.au -> Department of Industry, Innovation and Science.
3. Notice that the preview map is blank (problem 1)
4. Click Add to Map
5. Notice that the main map is blank too, and the Time field on the workbench is blank (problem 2)
image

It looks like the problem might be that the time starts past the end of the data.

kring:
In the same dataset above, when the time moves past 01/01/2015, the map goes blank even though the timeline shows another year.

lib/Models/RegionMapping.js
```diff
@@ -590,14 +594,16 @@ function addDescriptionAndProperties(regionMapping, regionIndices, regionImagery
                 if (defined(cRowObject)) {
                     cRowObject.string._terria_columnAliases = columnAliases;
                     cRowObject.string._terria_numericalProperties = cRowObject.number;
-                    imageryLayerFeatureInfo.properties = cRowObject.string;
+                    imageryLayerFeatureInfo.properties = Object.assign({}, feature.properties, cRowObject.string);
```
kring: We can't use `Object.assign` because it's not available in all browsers and it's inappropriate for a library to globally polyfill something.  Use Cesium's `clone` instead.

kring: or Cesium's `combine`.

lib/Models/RegionMapping.js
```diff
@@ -639,22 +645,95 @@ function addDescriptionAndProperties(regionMapping, regionIndices, regionImagery
...
+    // Before and after are defined according to animation direction
+    var isBefore = animateForwards ? JulianDate.lessThan : JulianDate.greaterThan;
+    var isAfter = animateForwards ? JulianDate.greaterThan : JulianDate.lessThan;
+    // Loop over time intervals and generate a time interval for the current and next imagery layer
+
+    // Need to keep track of 3 times:
+    // - past: The closest transition time in the past (for current interval)
+    // - future1: The closest transition time in the future (for current & next intervals)
+    // - future2: The next closest transition time in the future (for next interval)
+    var past = null;
```
kring, line 646: Since you're using for / of, lambdas, and other fancy features in this function, you might as well use `const` and `let` too right?

lib/Models/RegionMapping.js
```diff
@@ -639,22 +645,95 @@ function addDescriptionAndProperties(regionMapping, regionIndices, regionImagery
...
+    var catalogItem = regionMapping._catalogItem;
+    var currentTime = catalogItem.terria.clock.currentTime;
+    if (regionMapping._nextImageryLayerInterval && TimeInterval.contains(regionMapping._nextImageryLayerInterval, currentTime)) {
+        // Potential race condition. If someone changes variables after the last clock tick of the current imagery,
+        //  the old nextImageryLayer will be shown. Could store _nextILRegionIndices and compare before using that IL
+        setOpacity(catalogItem, regionMapping._nextImageryLayer, catalogItem.opacity);
+        fixNextLayerOrder(catalogItem, regionMapping._imageryLayer, regionMapping._nextImageryLayer);
+        ImageryLayerCatalogItem.disableLayer(catalogItem, regionMapping._imageryLayer);
+        regionMapping._imageryLayer = regionMapping._nextImageryLayer;
+        console.log(`Using prerendered layer made for: ${regionMapping._nextILTime}`);
```
kring: Should probably remove this console.log.

lib/Models/RegionMapping.js
```diff
@@ -639,22 +645,95 @@ function addDescriptionAndProperties(regionMapping, regionIndices, regionImagery
...
  * @param {Array} [regionIndices] An array the same length as regionProvider.regions, mapping each region into the relevant index into the table data source.
  *                  If not provided, it is calculated, and failed/ambiguous warnings are displayed to the user.
  */
 function setNewRegionImageryLayer(regionMapping, layerIndex, regionIndices) {
-    regionMapping._imageryLayer = createNewRegionImageryLayer(regionMapping, layerIndex, regionIndices);
+    var catalogItem = regionMapping._catalogItem;
+    var currentTime = catalogItem.terria.clock.currentTime;
+    if (regionMapping._nextImageryLayerInterval && TimeInterval.contains(regionMapping._nextImageryLayerInterval, currentTime)) {
+        // Potential race condition. If someone changes variables after the last clock tick of the current imagery,
+        //  the old nextImageryLayer will be shown. Could store _nextILRegionIndices and compare before using that IL
```
kring: How big a problem is this?  Couldn't you fix it by setting `_nextImageryLayerInterval` to undefined when the variable changes?

lib/Models/RegionMapping.js
tephenavies: I guess I just followed what the rest of the file was using for variable assignment. I tend to follow `require` vs `import` and `let`/`const` vs `var`, but interchange `function() { }` with `() => {}`, though I do agree it would make more sense to use `const` and `let`. (In this specific case block scope = function scope, so there's no advantage with `let`)

lib/Models/RegionMapping.js
tephenavies: I think I may have fixed this race condition elsewhere, but I'm not sure

lib/Models/RegionMapping.js
kring: yep that makes sense, I don't have any problem with that.

I guess our long term goal should be to move our code toward newer ES features; we just don't want to create a mixed mess in the meantime.  So with that in mind, I think it's fine to write new functions with newer style even if the rest of the file uses an older style.  Also upgrading a whole function is fine.  And if you're really ambitious, upgrading the whole file is great.

If you're making major-but-trivial changes to existing code, though, it'd be good to keep those changes separate for sanity of PR review.  So say you want to replace var with const/let in a whole file.  Ideally you'd do it in a separate branch, PR it, insta-merge it, and then pull it into the branch you were originally working in to make further changes.

TL;DR: you don't need to do anything.  I'm just laying out how I think we should approach this in general.

tephenavies:
It seems wrong to replace the unparsed string with the object if a string if JSON. Maybe instead it should go into some `_terria_jsonProperties` namespace (much like `_terria_numericalProperties`)?

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -306,13 +306,32 @@ function getPropertyValuesForFeature(feature, clock, formats) {
...
+    // Try JSON.parse on values that look like JSON arrays or objects
+    let result = parseValues(properties);
+    result = replaceBadKeyCharacters(result);
     if (defined(formats)) {
         applyFormatsInPlace(result, formats);
     }
     return result;
 }
 
+function parseValues(properties) {
```
tephenavies, line 324: This is conceptually the wrong place to be modifying feature info properties (it's in the view layer, not the model layer), but I'm not sure if there's anything in the model layer that reads all feature info properties.

kring: I think it's ok.  Or maybe it's a symptom of a bigger problem.

Currently, the whole "feature info template" thing is something the UI does, for better or worse (that's the potentially "bigger problem" I'm referring to above).  So we can think of this `parseValues` function as just a step in the process of presenting data to the feature info template.  It doesn't actually modify the properties of the feature; clearly it would be pretty dodgy if the UI were quietly modifying data in the Models layer.

So I don't think it's a major problem.  Is there an aspect to it that I'm missing?

kring:
> It seems wrong to replace the unparsed string with the object if a string if JSON. Maybe instead it should go into some` _terria_jsonProperties` namespace (much like `_terria_numericalProperties`)?

Yeah it's not ideal.  I'm not sure what to make of the `_terria_numericalProperties` though.  It looks like it's only used for downloadable data and isn't really accessed by a feature info template at all.  Maybe it's possible to access from a feature info template, but it's certainly not documented.

I feel like we might have discussed this before, but remind me why this is needed at all.  The problem is that the MVTs provide a JSON property as a string rather than proper JSON, right?  Can we just fix it at that level?  e.g. in MapboxVectorTileImageryProvider.js, pickFeatures, parse out any JSON fields we find?  It might be basically the same logic, but at least its impact is more limited.

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
tephenavies: That makes sense. There's no problem with leaving it in the view layer.

tephenavies:
Replying to your first comments about blank maps, I think this is because I changed intervals to not include their endpoint, and the map jumps to the endpoint of the last interval. I'm not sure how to best fix that.

The previous method of not overlapping intervals was: trim intervals by 1s or 5% unless it's the last interval, and include the endpoint. I guess I could change it to include the endpoint if no new interval starts there?

I don't think the empty workbench time selector is caused by my changes (there were very few merge conflicts with the time slider at least).

kring:
> I guess I could change it to include the endpoint if no new interval starts there?

I think this makes sense.  In fact, it should just happen by defualt.  Make the interval closed, and Cesium's TimeIntervalCollection will automatically truncate the earlier interval (make its end open) when you add the next interval.

> I don't think the empty workbench time selector is caused by my changes (there were very few merge conflicts with the time slider at least).

I think it's the same problem.  The selected time shown in the workbench works by finding the current time in the list of intervals.  Currently it will come up empty so it displays nothing.

tephenavies:
I think the problem is that we're not using Cesium's `TimeIntervalCollection`. Should there be a `TimeIntervalCollection` per id (in this case region)?

kring:
> I think the problem is that we're not using Cesium's TimeIntervalCollection. Should there be a TimeIntervalCollection per id (in this case region)?

Oh right.  There might be a good reason for that.  @RacingTadpole do you remember?

If we can't use `TimeIntervalCollection`, just closing the end of the last TimeInterval makes sense.

PR #2712, opened 2017-10-26
use copy button to copy url to clipboard

chloeleichen:
Mobile:
screen shot 2017-10-26 at 4 05 44 pm
Desktop:
screen shot 2017-10-26 at 4 05 53 pm


lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -143,39 +148,55 @@ const SharePanel = createReactClass({
                        onOpenChanged={this.onOpenChanged}
                        smallScreen={this.props.viewState.useSmallScreenInterface}>
                 <If condition={this.state.isOpen}>
+                        <If condition={this.props.viewState.useSmallScreenInterface}>
```
tephenavies: Probably would be better to break into renderSmallScreen and renderNormal functions because the layouts are so different

tephenavies:
Please add a changelog entry

tephenavies:
Test fail is a false positive.

Works on Chrome & Safari and Chrome Android.

PR #2718, opened 2017-10-31
Only show "zoom to extent" if rectangle is available

### LINKED ISSUE #2579: "Zoom to extent" doesn't work for SDMX-JSON
See https://github.com/TerriaJS/nationalmap/issues/520

RacingTadpole:
Fixes #2579.
Unfortunately I think this makes use of some of the changes introduced to make "split" show nicely, so may be hard to retrofit.

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -764,6 +764,17 @@ var scratchRectangle = new Rectangle();
 };
 
 /**
+ * Returns true if this item currently has a rectangle to zoom to. Depends on observable properties, and so updates once loaded.
+ * @returns {Boolean} A flag for whether this catalog item supports "zoom to".
+ */
+CatalogItem.prototype.canZoomTo = function() {
+    if (defined(this.nowViewingCatalogItem)) {
+        return defined(this.nowViewingCatalogItem.rectangle);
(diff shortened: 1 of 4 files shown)


lib/Models/CatalogItem.js
```diff
@@ -764,6 +764,17 @@ var scratchRectangle = new Rectangle();
 };
 
 /**
+ * Returns true if this item currently has a rectangle to zoom to. Depends on observable properties, and so updates once loaded.
+ * @returns {Boolean} A flag for whether this catalog item supports "zoom to".
+ */
+CatalogItem.prototype.canZoomTo = function() {
+    if (defined(this.nowViewingCatalogItem)) {
+        return defined(this.nowViewingCatalogItem.rectangle);
```
kring, line 772: You should call `this.nowViewingCatalogItem.canZoomTo` instead of checking the `rectangle` property directly.  That way a derived catalog item can override the behavior.

kring: Also, should this bit a property on the prototype instead of a function?

kring:
These changes were cherry-picked into the `5.4.0-upgrade` branch and then merged back into master.  Closing.

Issue #2706, opened 2017-10-24
Sharing a point from the Search


meh9:
It doesn't seem to be possible to share a location you've found in the Search - it's there in the mini screenshot we show in the Share URL function, but when you go to the link the Search is empty and we don't show a marker on the map.

E.g. search for Canberra, click on one of the results to zoom to the location and show a marker, click Share and go to the new Share URL produced. Note the lack of Search results and no marker.

This is also related to https://github.com/TerriaJS/terriajs/issues/2697

a-stacey:
This was resolved via 620d35e326941251c1fe7155cd21707f1ecad638.

Issue #2697, opened 2017-10-19
Lat / Long Info + Share


a-stacey:
@meh9 and I were discussing this morning how to better handle sharing of a specific point.

- For every point that the user clicks the lat/long should be displayed in the feature information. This is both in the case where there is already feature information displayed at the point (in this case the lat long should appear at the bottom, and when there is not - just the lat/lon should appear) and where there is no other feature info (in this case it should be the only item displayed.
- The cross hair should remain visible. (Currently when there is no feature info the cross hair fades out immediately).

We also think that it would be nice to make this share more easily. (From the above implementation you should then be able to click on the share link and share the point with the point highlighted, but this may require more intuition then what could be expected from a casual user). We think that if the lat/lon were linked to the share button (lan/lon click is a synonym for the share button) so that when clicked on the lat lon the share panel opens up (originally I suggested that it was a link to the share link, but @meh9 pointed out that in order to do that a share link would need to be created each time, with would pollute the database).

(Note: Add crosshair to what is displayed in this image).
![no other info](https://user-images.githubusercontent.com/4453477/31797666-3a89e398-b57b-11e7-8417-2f9253e4adcf.png)

![where other info present](https://user-images.githubusercontent.com/4453477/31797295-48e6bcce-b579-11e7-8a23-09bdd7cfdb35.png)

meh9:
This is basically fairly similar to being able to click anywhere to place a pin, and share the location of the pin. It's much more about sharing/highlighting the location of something than the data available at the point.

meh9:
@philipgrimmett what do you think? :)

meh9:
In addition to the lat/lon we should also show elevation.

philipgrimmett:
@meh9 @a-stacey Re: lat/lon being included in feature info: yep, it's been on the wish list for a while! My bad for not creating an issue earlier.

As for sharing a point/feature location: might need to think this one over. Clicking on lat/lon in the feature info panel to launch Share feels a bit odd, but maybe we could clarify by including the share icon?

![screenshot 2017-10-23 12 29 35](https://user-images.githubusercontent.com/5292858/31868709-d94ad072-b7ed-11e7-88e4-69f6c64cb4a2.png)

a-stacey:
Just adding a screenshot of as implemented as I needed to do a build to double check something else anyway.

[...]

PR #2722, opened 2017-11-01
2697 2706 Share Marker Location / Add Lat+Lon to Feature Info Panel

a-stacey:
Please pay particular attention to the MouseCoords commits [1109a9ac515802f848cdc3271b49201cbbc0ba49, 4eb67cd94da677a5d642c4e5764ca863baf3b5e3], as I'm not really sure that I have the terminology / directories correct here.

May also want to pay particular attention to the styling of the FeatureInfoPanel as suspect that there might be cleaner ways of doing this.

Can create a PR for #2706 separately if the PR for #2697 gets bogged down.

I have put the marker update location code in the FeatureInfoPanel (locationItem):
```
+ this.locationUpdated(latitude, longitude);
```
since you really only want to update the location of the marker if the user has the feature info panel open (and has mostly likely toggled the location marker on manually), where as if it updating the location was a generic concept and put as part of the marker location it would happen if the user has searched for and selected a location manually (which seems really counter intuitive to me). Originally I had the marker location only update when the user explicitly selected to update the location (re-clicked the marker pin), but this gave the interface a dead feel (its broken and I have to manually click the location pin to get it to update), and also this means that you have to have another UI interaction to remove the pin (because the button represents update rather then toggle). So I think that the implementation that I have included here is the best of these solutions.


lib/Models/prettyCoordinates.js
```diff
@@ -0,0 +1,43 @@
+import defined from 'terriajs-cesium/Source/Core/defined';
+import proj4 from 'proj4/lib/index.js';
+
+/**
+ * Turns the latitude / longitude in degrees into a human readable pretty strings.
+ */
+export function prettifyCoordinates(latitude, longitude, height, errorBar) {
```
kring: Make the function name match the filename (probably by renaming the file in this case).

lib/Models/prettyCoordinates.js
```diff
@@ -0,0 +1,43 @@
...
+        result.elevation = undefined;
+    }
+
+    return result;
+}
+
+/**
+ * Turns the latitude / longitude in degrees into a human readable pretty UTM zone representation.
+ */
+export function prettifyProjection(latitude, longitude, proj4Projection, proj4longlat, projectionUnits) {
```
kring: Probably better to put these in separate files.

lib/Models/prettyCoordinates.js
a-stacey: Done

lib/Models/prettyCoordinates.js
a-stacey: Done 558bda2.

a-stacey:
Identified a bug with this implementation. When you click on the feature marker (blue) the marker location updates to the location as if you clicked through the marker, but the cross hair highlights the location at the location the marker was pointing at. Futher in the feature info the feature section shows the original lat/lon, but the added feature info shows the new marker location.

See below:
image

Steps to reproduce:
1) Select a known location (in the example above the corner of the SA / QLD / NT boundary).
2) Select the location marker (in the example above clicked on the center of the target of the pin).

a-stacey:
7f4979f should fix the bug listed above and hopefully mean this PR can be reviewed again.

lib/Models/prettifyCoordinates.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+import defined from 'terriajs-cesium/Source/Core/defined';
+import proj4 from 'proj4/lib/index.js';
+
+/**
+ * Turns the latitude / longitude in degrees into a human readable pretty strings.
+ */
+function prettifyCoordinates(latitude, longitude, height, errorBar) {
```
kring: In the code, _always_ put longitude first, please.

lib/Models/prettifyCoordinates.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+import defined from 'terriajs-cesium/Source/Core/defined';
+import proj4 from 'proj4/lib/index.js';
```
kring: This file (and prettyifyProjection) should probably go in Maps instead of Models.  It doesn't depend on `Terria`, knockout, or any of the other model-y stuff.

lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
```diff
@@ -119,6 +122,58 @@ const FeatureInfoPanel = createReactClass({
         }
     },
 
+    addManualMarker(latitude, longitude) {
```
kring: Flip the parameter order here too.

lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
```diff
@@ -119,6 +122,58 @@ const FeatureInfoPanel = createReactClass({
...
+    },
+
+    locationUpdated(latitude, longitude) {
+        if (defined(latitude) && defined (longitude) && markerVisible(this.props.terria)) {
+            removeMarker(this.props.terria);
+            this.addManualMarker(latitude, longitude);
+        }
+    },
+
+    locationItem(cartesianPosition) {
```
kring: How about `renderLocationItem`?  In general, functions should be named as verbs.

lib/Models/LocationMarkerUtils.js
```diff
@@ -44,7 +45,7 @@ export function addMarker(terria, viewState, result) {
         correctEntityHeight(firstPointEntity, cartographicPosition, terria.cesium.scene.globe.terrainProvider, 15);
```
kring, line 48: Not something you changed, but it'd be great to use `prettifyCoordinates` to generate the entity's description, too.  Currently it uses `${result.location.latitude}, ${result.location.longitude}`.

lib/ReactViews/FeatureInfo/feature-info-panel.scss
```diff
@@ -114,3 +114,35 @@
     max-height: none;
   }
 }
+
+.location {
+  margin: 10px;
+  display: flex;
```
kring, line 134: `flex` isn't supported in IE9, but it seems to look passable anyway:
image

kring: Just FYI, no need to change anything.

kring:
Strange goings-on after sharing a location.  Try this:
1. Add Broadband ADSL Availability.
2. Click a location on the map.
3. Click the pin at the bottom of the feature info to make the location persistent.
4. Click share, copy the URL, and open it in a new tab.

You'll see a few strange things in the new window:
* The location pin is missing.
* Collapsing the feature info by clicking the "Feature Information" header at the top doesn't work.
* Closing the feature info doesn't work. It reappears immediately.

kring:
Fairly minor, but there's no space between the `Lat / Lon` label and the value when the contents of the feature info panel are small, such as during loading:

image

a-stacey:
Have just put a single space, let me know if you think it needs more:
image

lib/Models/LocationMarkerUtils.js
a-stacey: Have a look at  88ed3b3538ece4d174f8d0f2ea302407e015e658 and let me know if this is what you were thinking. I've preserved 5 dp since it seems nice to preserve the additional resolution in the description (testing revealed that about 5 dp was detailed enough precision compared with the visual base map without providing more precision then is apparent).
image

a-stacey:
Thanks for picking up the share link problem (for my reference it was introduced in a0717f9624ea5f7573f1887c8bd6d021367490ff). Unfortunately I think that I may have fixed the symptom rather then the cause, but I'm not really sure why the position becomes NaN's and even if we do find the cause there might be other scenarios that cause this same issue. I also found another issue that I thought was related https://github.com/TerriaJS/terriajs/issues/2785, but it seems to also be present in NationalMap (had a look into the cause of this as well, but again it wasn't immediately clear so placed it in a separate issue for now).

Think that this addresses all of the issues mentioned in code review.

a-stacey:
This is ready for review / merge.

lib/Map/prettifyCoordinates.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {Object} options Object with the following properties:
+ * @param {Number} options.height The height.
+ * @param {Number} options.errorBar The error +/- for the height.
+ * @param {Number} options.digits The number of digits to fix the lat / lon to.
+ */
+function prettifyCoordinates(longitude, latitude, options) {
+    var result = {};
+
+    options = defaultValue(options, {});
+    options.digits = defaultValue(options.digits, 3);
```
kring: Better to assign this to a local variable instead of putting it in `options.digits`.  It's impolite to change an object passed in by the user without a very good reason.

a-stacey: Thanks for this, absolutely right...wish we had const correctness so that we could get the complier to detect this ;) Have fixed this for both options and options.digits...don't love the variable name I've used but out of better inspiration atm.

kring: I'm not too fussed about `options` itself, since assigning a new value to that won't be visible outside the function. I know lots of people are of the mind that you shouldn't modify parameters, though, and I'm ok with that.

PR #2745, opened 2017-11-15
Add option to SDMX-JSON to better handle rates and averages

RacingTadpole:


kring:
I'm not so sure about this approach.  For one thing, it loses the user's selections as soon as they choose a rate, which seems a bit rude.  For another, setting each `isActive` to false in a separate `runLater` means it might take awhile for things to settle down (because each state change might trigger a new load, right?) and meanwhile the layer is in an inconsistent state.  And finally, it's sort of useful to be able to see the cartesian product of all the selections in the feature info even for a rate, even though it doesn't make sense to display a total or even to map it.

So without know how hard this would be, what do you think of this: when the user selects something that would require us to sum rates:
* show a blank map (just the regions, no colors)
* show a blank legend
* show a message on the workbench saying that the selections cannot be totaled, so no map is shown.
* don't show the Total selected in the feature info (or leave it blank), but do show the other properties.

Does that seem sensible?  Is it difficult?

RacingTadpole:
Interesting... I didn't think of just not showing it on the map. What you're saying sounds reasonable, and probably not too hard. It might be worth running past @philipgrimmett too.

kring:
A little background on this issue:
If you go to https://map.terria.io/#marriage-sdmx and turn on the second dataset (Marriage Law Postal Survey Participation), you'll see the problem this is trying to solve.
You're looking at a participation rate, which is a percentage, but the values on the legend go way above 100%, which makes no sense.
It happens because we're adding up the male and female participation rates.
Unfortunately it's a mathematical impossibility (given the data we have) to compute the actual total participation rate across males and females.  So the question is: what do we do instead?

@RacingTadpole's original implementation was to tell the user:
> Your selection had multiple values for "Sex", but the measure you now have chosen cannot be totalled across them. As a result, your "Sex" selections have been trimmed back to just one value.
And then deselect all but the first measure in each category (so you'd end up with just Males selected).  

My suggestion was to instead leave the user's choice alone (leave Males and Females both selected), but blank out the map and legend and put a message on the workbench saying something like:
> You have multiple values selected for "Sex" but the measure cannot be totaled across them.  Please select only one value or select a different measure.

RacingTadpole:
ok, ready for another look @kring , thanks!

RacingTadpole:
btw, I've updated the tests because they were failing when run by themselves (though succeeding in the full suite). 

The problem was finding the region ids: it was depending on eg. “STE” defaulting to 2011 values, but we’ve now updated them to default to 2016 values. So I’ve fixed it so future changes to the default year won’t break the tests.

However I did not look through earlier tests to see if one of them is somehow leaking the 2016 STE codes into later tests.

kring:
Thanks @RacingTadpole.  It would be better to show a message on the workbench because the popup gets annoying fast if you need to make multiple changes.  But I'll write a separate issue for that.

PR #2746, opened 2017-11-15
Use app clock if item clock is undefined.

### LINKED ISSUE #2743: Using "Split" on a non-time-varying workbench item causes an error
Error: 
```
TypeError: Cannot read property 'currentTime' of undefined
```
[Example](https://nationalmap.gov.au/northernaustralia/#share=s-ol6orTj5Eg7UWkV1RDr1fWM898k) where I've just clicked split.
[Second example](https://nationalmap.gov.au/northernaustralia/#share=s-yAmlQWaBWDHScbpuZWCNTxzlAn5) where you can click split to see the error.

kring:
Fixes #2743

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 5.5.2
+
+* Fixed a bug that caused an error when Splitting a layer that does not have a `clock`.
+
 ### 5.5.1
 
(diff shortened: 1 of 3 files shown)


RacingTadpole:
Yep, tested this with SDMX-JSON items too, and it solves the problem.

PR #2765, opened 2017-12-05
feature info overflow header

chloeleichen:
A small fix on colour and style for feature info panel message when more feature to display than allowed. 
screen shot 2017-12-05 at 1 12 07 pm
Also made the max width smaller (to show part of map) on desktop as discussed with @philipgrimmett on slack.


lib/ReactViews/FeatureInfo/feature-info-panel.scss
```diff
@@ -22,9 +22,10 @@
   a{
     @include link($feature-info-bg);
   }
-  @media screen and (min-width : $md){
+
+  @media screen and (min-width : $sm){
     //temp
-    max-width: calc(100vw - #{$work-bench-width} - 150px);
+    max-width: 432px;
```
kring: Is this change intentional?  It's not mentioned in the description of the PR.

chloeleichen: Sorry, updated PR description

kring:
I think the feature info sizing needs a bit more thought @philipgrimmett @chloeleichen.  On a large screen (not even that large) and for some content, it feels too small.  For example:
image

In fact, it's smaller on desktop than on a wide mobile device!
image

kring:
The "too many features" change looks great, though.

PR #2766, opened 2017-12-05
Include "Chart Data" and dragged catalog file groups as catalog tabs

### LINKED ISSUE #2751: New Tabbed Catalogue breaks drag and drop of init files
When enabling the new `tabbedCatalog` option for a TerriaJS site it break dragging and dropping (as well as loading from a `#file` etc) of other init files into the catalogue. It seems to process them, but then the layers contained in the init files do not appear in the catalogue.

tephenavies:
Fixes #2751.

### DIFF
--- a/lib/ReactViews/ExplorerWindow/Tabs.jsx
+++ b/lib/ReactViews/ExplorerWindow/Tabs.jsx
@@ -36,7 +36,7 @@ const Tabs = createReactClass({
         };
 
         if (this.props.terria.configParameters.tabbedCatalog) {
-            return [].concat(this.props.terria.catalog.group.items.filter(member => !member.isUserSupplied).map((member, i) => ({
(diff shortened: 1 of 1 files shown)


meh9:
@steve9164 this seems to fix the drag and drop case and the Chart Data case when we load e.g. a Current Generation CSV timeseries in AREMI, however it doesn't work when adding chart data from a WPS response like we do in GEOGLAM. As you can see in the screenshot below although we have data in the chart no Chart Data tab has been added for it:

screen shot 2017-12-06 at 09 20 41

tephenavies:
Looks like standard TerriaJS behaviour is to make a Chart Data top-level catalog group only when the chart is expanded from feature info (both in tabbed and non-tabbed catalog). If that is wrong it'd best to put that in a new issue.

PR #2775, opened 2017-12-21
fix #2774

### LINKED ISSUE #2774: Clicking the location search marker breaks the feature info
1. Go to https://nationalmap.gov.au
2. Add Communications -> Broadband ADSL Availability to the map
3. Click a colored region on the map and observe that the feature info is displayed as expected.
4. Search for a location, e.g. `13 Garden Street, Eveleigh` and click the first search result.
5. Click a colored region again and see that the feature info still works.
6. Click the blue marker indicating your searched location. The feature info will either get stuck at "Loading..." or display "No data is available here".
7. Click on a colored region again and the feature info panel will not update.  Close it and click again and you'll see the feature name with the details apparently collapsed.  Clicking the name in attempt to expand the details will do nothing.
image

The immediate cause is an exception thrown by `descriptionFromFeature` when clicking on the marker.

chloeleichen:
Asking for `clock` when there isn't (eg, when click on location marker) breaks react

Fixes #2774

### DIFF
--- a/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
@@ -12,6 +12,7 @@ import classNames from 'classnames';
 import defined from 'terriajs-cesium/Source/Core/defined';
 import Ellipsoid from 'terriajs-cesium/Source/Core/Ellipsoid';
 import isArray from 'terriajs-cesium/Source/Core/isArray';
+import JulianDate from 'terriajs-cesium/Source/Core/JulianDate';
 
 import CustomComponents from '../Custom/CustomComponents';
 import FeatureInfoDownload from './FeatureInfoDownload';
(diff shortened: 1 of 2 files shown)


lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -132,9 +132,12 @@ const FeatureInfoSection = createReactClass({
         //     markdownToHtml (which applies MarkdownIt.render and DOMPurify.sanitize), and then
         //     parseCustomHtmlToReact (which calls htmlToReactParser).
         // Note that there is an unnecessary HTML encoding and decoding in this combination which would be good to remove.
-        let description = feature.currentDescription || getCurrentDescription(feature, this.props.clock.currentTime);
-        if (!defined(description) && defined(feature.properties)) {
-            description = describeFromProperties(feature.properties, this.props.clock.currentTime);
+        let description = null;
+        if (this.props.clock) {
+          description = feature.currentDescription || getCurrentDescription(feature, this.props.clock.currentTime);
```
kring: I think this `feature.currentDescription` should be moved outside the if, cause it's possible to use that description even without a clock.

PR #2780, opened 2018-01-02
Splitter: Allow the splitter to move vertically as well as horizontally.

a-stacey:
The PR modifies the spiltter thumb tab so that it moves vertically as well as horizontally. This allows the user to move the splitter position if it vertically happens to occlude a significant region of the map. Note the position is not stored in the state for share links as this seems overkill for now (particularly since the location is relative to the view port and so will visibly differ anyway).

Have not reviewed with UX as it was trivial enough that its easier to implement and then evaluate the usability then to discuss previously. One slight issue identified is that it is possible to slip it under other UI elements, but it seems that this is not too much of a problem in practice in that most all cases I was able to retrieve it (i.e. it can look a bit ugly - but is at user action, but functionally the worst I could find was hiding behind feature info, which can be retrieved by closing the feature info).

@meh9 check this out as you may recall that we briefly discussed this last year.


kring:
It looks like the splitter doesn't move when the cursor is directly on the dividing line.  That makes the movement seem very jumpy sometimes.  Here's a video that sort of captures it:
vertical-splitter

kring:
This is cool, but I guess I'm wondering why users wouldn't avoid the occlusion just by moving the map slightly instead?  I'm happy to merge this if @philipgrimmett is happy with it though.  What do you think, Phil?

philipgrimmett:
Seems fine, but worth considering the following:
 
- Users might drag the scrubber right to the top/bottom of the screen and then have trouble finding it later in the session
- In the above case users might forgot that the splitter is turned on, leading to unexpected behaviour

We could snap the scrubber back to vertical centre on mouse up? Solves the above issues, but might be annoying in practice. Should be pretty easy to prototype/test.

a-stacey:
> Users might drag the scrubber right to the top/bottom of the screen and then have trouble finding it later in the session.

So yes this is potentially a problem, but really need to do some testing with non experienced users (i.e. not me) to see how problematic that is in the real world. I did some quick testing when I made this PR because I was concerned it might get lost behind other elements as well...but that didn't seem to be a problem in practice that was to much to work around and seemingly not different to the current state. 

> We could snap the scrubber back to vertical centre on mouse up? Solves the above issues, but might be annoying in practice. Should be pretty easy to prototype/test.

I feel that this somewhat defeats the purpose of this feature. The problem that I was really trying to solve was when the slider was above a specific critical point that you wanted to see directly what was under the split. As Kevin points out the users could move the map...but this is potentially also irritating. (I've run up against this testing other features). One alternative that could be worth considering is restricting the movement to within say +/-25% of the central location, that way we enable some flexibility of movement, but not so much that it moves completely outside the focal region of the map? How would you feel about that?

a-stacey:
This PR is ready for review (both code @kring and UX @philipgrimmett ) and merge.

lib/Models/Terria.js
```diff
@@ -386,13 +386,18 @@ var Terria = function(options) {
      * 0.0 is on the left, 0.5 is in the center, and 1.0 is on the right.  This property is observable.
      */
     this.splitPosition = 0.5;
+    /**
+     * Gets or sets the current vertical position of the splitter (if {@link Terria#showSplitter} is true) as a fraction of the map window.
+     * 0.0 is on the top, 0.5 is in the center, and 1.0 is on the bottom.  This property is observable.
+     */
+    this.splitPosition = 0.5;
```
kring: should this be `this.splitPositionVertical`?

a-stacey:
This should now be ready for review / merge.

I believe that this is the implementation that @philipgrimmett and I worked out was the best a couple of months ago. Seems that these constraints work ok for both desktop and mobile.

Note: I have made the explicit design decision not to include the vertical position in the share link so that it will always default the tab back to the more salient center position when a share link is generated, prompting the user as to the sliders intention rather then maxing the tab if the user hides it below. Its clear that there are multiple sensible options here, since in this configuration the tab could appear over an important region, but my feeling is that its more important to have the usability/discoverability of the splitter in this situation and we can always change this fairly trivially down the track if we bug reports against it.

Issue #2736, opened 2017-11-08
Feature info doesn't always update when time changes


kring:
Ideally, anything displayed in the feature info would update whenever the layer's time changes, and sometimes it does.  But there are cases where it doesn't:
* Feature info for imagery layers (e.g. WMS) _never_ updates when the time changes.  Doing so would require that the pick (GetFeatureInfo for WMS) be run again whenever we move to a new interval.
* ~Feature info for `TableCatalogItem` updates when the time is changed using the timeline (including the date picker on the timeline), but it does _not_ update when using the date picker on the workbench.  Fixing this is tricky because there's currently no event that gets raised when the time changes via that mechanism, so the `FeatureInfoSection` can't subscribe to it.~ ** this is fixed in v8 (@nf-s )

kring:
See comment in `FeatureInfoSection.jsx`, `setSubscriptionsAndTimeouts` function.

```js
// This is probably a DataSourceClock because the catalog item is using its own clock.
// But we currently have no way of subscribing to changes to it.
// See https://github.com/TerriaJS/terriajs/issues/2736
```

Issue #2658, opened 2017-09-08
Add ability to get and format the current timeline time in feature info template


meh9:
Since we update the values in the feature info when the time changes on the timeline, it would be very useful to be able to show at which time we are actually showing the values for. We should add access to the current time of the timeline (or the current time of the actual data from the layer if possible?) as well as a formatting function for it similar to the mustacheFormatNumberFunction or mustacheURLEncodeTextComponent.

It's been suggested it should be relatively easy to add the data in here:

https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx#L76-L95

a-stacey:
>  Since we update the values in the feature info when the time changes on the timeline, it would be very useful to be able to show at which time we are actually showing the values for.

Talking this through with @axman6 and we realised that the value only updates with the date when you select a new point after you have updated the position on the time line. For now I will just assume that we want to show the time for the info shown.

a-stacey:
Update to the above the FeatureInfo also updates when you hit play on the timeline. So it just doesn't update when you manually select a new time...maybe it should?

a-stacey:
@meh9 check out  PR #2782 and see if it meets your requirements in terms of formatting. It should be effectively formatting as per https://www.npmjs.com/package/dateformat and the format for the mustache feature info section will look like: `{{#terria.formatDateTime}}{"dateFormatString": "dd-mm-yyyy HH:MM:ss"}{{terria.currentTime}}{{/terria.formatDateTime}}`.

a-stacey:
This was fixed and merged as part of this PR: https://github.com/TerriaJS/terriajs/pull/2782

PR #2782, opened 2018-01-08
[#2658] Add ability to get and format the current timeline time in feature info template. / Partial [#2736] Fix feature info updating when clock updates for TableCatalogItems

a-stacey:
I used `dateFormat` since it already seemed to have precedent in the codebase and the formatting seems sane enough compared to what you would implicitly expect.

---
From #2831:
This PR is only aimed at fixing part of #2736.

>    Feature info for TableCatalogItem updates when the time is changed using the timeline (including the date picker on the timeline), but it does not update when using the date picker on the workbench. Fixing this is tricky because there's currently no event that gets raised when the time changes via that mechanism, so the FeatureInfoSection can't subscribe to it.


lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -89,6 +91,9 @@ const FeatureInfoSection = createReactClass({
                     longitude: CesiumMath.toDegrees(latLngInRadians.longitude)
                 };
             }
+            if (this.props.clock.currentTime) {
```
kring: Check that `this.props.clock` is defined, too.  In at least one case (when the user clicks on a location search result marker), it won't be.

kring:
For consistency with number formatting, it'd be nice to be able to specify the date/time format with the `formats` property too (http://terria.io/Documentation/guide/connecting-to-data/customizing-data-appearance/feature-info-template/#formatting-numbers).

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -440,16 +446,52 @@ function mustacheFormatNumberFunction() {
...
+ */
+function mustacheFormatDateTime() {
+    return mustacheJsonSubOptions((text, options) => {
+        var date = Date.parse(text);
+
+        if (!defined(date) || isNaN(date)){
+            return text;
+        }
+
+        if (defined(options) && defined(options.dateFormatString)) {
```
kring: I think simply `format` might be better than `dateFormatString`. The `date` bit is implied by the fact that you're passing this it a thing called `formatDateTime` (and perhaps it's even a bit misleading because it's also a time format, not just a date format). And it's not really necessary to state the type (`string`).

lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -442,16 +448,52 @@ function mustacheFormatNumberFunction() {
...
+ */
+function mustacheFormatDateTime() {
+    return mustacheJsonSubOptions((text, options) => {
+        var date = Date.parse(text);
+
+        if (!defined(date) || isNaN(date)) {
+            return text;
+        }
+
+        if (defined(options) && defined(options.dateFormatString)) {
```
kring: (re-adding this comment because GitHub thinks the original is outdated)
I think simply `format` might be better than `dateFormatString`. The `date` bit is implied by the fact that you're passing this it a thing called `formatDateTime` (and perhaps it's even a bit misleading because it's also a time format, not just a date format). And it's not really necessary to state the type (`string`).

kring:
Please update the documentation in `doc/connecting-to-data/customizing-data-appearance.md`  with this new feature.

kring:
#2658 mentions providing a way to get the current time of the layer (rather than the current time on the timeline, which may be different). I think that makes sense, and you can find code to do that in the `render` method of `DateTimeSelectorSection.jsx` if you want to do it.

kring:
That's all I have, looks good!

a-stacey:
I think that addresses everything outstanding so should be ready for review.

a-stacey:
So I just came back through this code looking at another issue related to the clock. I think that I may have been confused when I read through this comment:

> #2658 mentions providing a way to get the current time of the layer (rather than the current time on the timeline, which may be different). I think that makes sense, and you can find code to do that in the render method of DateTimeSelectorSection.jsx if you want to do it.

Looking back through the code now,  I see this:
https://github.com/TerriaJS/terriajs/blob/1259e72c1a64ac754e84e6e0cea65f5214b927e0/lib/ReactViews/FeatureInfo/FeatureInfoCatalogItem.jsx#L51
so by inspection it seems that was already just the case anyway and just needed the additional check???...which does also beg the question of why even pass the clock through, since you can always obtain it from the catalogItem? I searched through the code for other FeatureInfoSection references but couldn't find any and so its probably nicer also to just cleanup an remove clock completely.

If there is a reason to keep the clock construct explicit, we could go to: 
```
            if (defined(this.props.clock) && defined(this.props.clock.currentTime)) {
                            propertyData.terria.currentTime = this.props.clock.currentTime;
```
instead of 6b20ef95b6bf59f8bcb374fc4f9ab2f0f81d5b3b. In this case the test probably needs to be updated as well since I explicitly test that the clock from the catalogItem is used and not that passed through .clock. 

What do you think @kring?

kring:
> I searched through the code for other FeatureInfoSection references but couldn't find any and so its probably nicer also to just cleanup an remove clock completely.

yeah I can't think of any reason for the `clock` prop to exist on `FeatureInfoSection`, either.

But displaying `clockForDisplay` is not _quite_ the same as displaying the layer's time.  Specifically, when the layer is using the timeline clock, the timeline time may be, say, `2018-01-17T17:07:12+11:00`. But that may mean that the layer time is actually `2018-01-17` or something like that, becaues the timeline time is inside an interval covering the entire day.  Does that make any sense?

a-stacey:
Right thanks, I only saw the line https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Workbench/Controls/DateTimeSelectorSection.js#L74, and completely missed https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Workbench/Controls/DateTimeSelectorSection.js#L75...silly. I think I get what you mean now. Take a look at 86d169d6dec2a3d9f4dbf589c761f2b53ccbff47 which is hopefully right now.

Have also removed the clock in 4062d286a8a59cb2e2e7844fe2ed2f5af2cfb8e0. One concern with this. There are a couple of paths where its not explicit that clock is defined (well seems at least not immediately obvious this is the case anyway)...this is more or less the same as the case was previously, but somewhat bothers me. Also not sure if you would normally log a changelog entry for this or not.

a-stacey:
This is ready for review / merge.

a-stacey:
Discussed offline and have merged this PR with https://github.com/TerriaJS/terriajs/pull/2831 both into this one PR now.

Copying the one outstanding comment from https://github.com/TerriaJS/terriajs/pull/2831#pullrequestreview-98436590:
```
In lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx:

>              featureInfoSection.setState({
-                removeClockSubscription: featureInfoSection.props.clock.onTick.addEventListener(function(clock) {
+                removeClockSubscription: featureInfoSection.props.clock.definitionChanged.addEventListener(function(clock) {

-----

Is this subscription needed anymore? I think you might just be able to change the UI to access catalogItem.currentTime and the observability will make the UI automatically refresh when it changes.
```

a-stacey:
So in response to the above question. This was my initial feeling to that I noticed when evaluated merging the code. 

Testing this seems to be fine (to remove) where the template subs in the current time, but not when you aren't explicitly using the time and the time does indirectly effect the values through: `setCurrentFeatureValues` -> `propertyGetTimeValues(feature.properties, clock) / getCurrentDescription(feature, currentTime)`

a-stacey:
Have addressed the above issue.

Please pay particular attention to changes in: 5dc356cac8275857055bd9b9cc502b183690ae10, these dont directly relate to the issues addressed by this PR, but when I modified the code around these lines I believe that this simplification became possible and also the other component addresses an issue that I believe was present.

PR #2801, opened 2018-01-29
Use terriajs-cesium v1.41.0

kring:


kring:
There's a problem with the bird migration example after this upgrade.  In TerriaMap:

1. Click Add Data.
2. Click the Bird migration paths catalog item and wait for the preview to load.
3. Click Add to map

The yellow bird dots may appear briefly, but then they'll disappear.

kring:
The above problem is fixed.  Next problem: Bing Maps basemaps don't work in IE9.

PR #2806, opened 2018-02-05
Upgrade to Leaflet 1.3.1.

kring:


kring:
@steve9164 do you mind taking a look at this?

tephenavies (commented):
I've updated my _onClick monkey-patch (it was missing feature ordering logic and was also throwing exceptions)

tephenavies (approved):

tephenavies:
Merge this when/if you're happy with my changes to the _onClick function. (It's not a great way to keep it updated, but I don't think there's another way to do it other than a fork of Leaflet)

kring:
Thanks @steve9164!

PR #2810, opened 2018-02-06
[#2748] DateTimeSelectorSelection: Add previous and next buttons.

### LINKED ISSUE #2748: Back/Forward functions for the new Date Picker
As we are starting to use the new Date Picker for things like temporally granular satellite image passes we are finding that it needs Back/Forward functions to step the the previous or next available imagery.

E.g., we are looking at satellite imagery for an area in Tasmania over the winter, and many of the images consists mainly of cloud. When this happens we start skipping backwards or forwards in the timeseries of the available imagery until we find one that is free of cloud. 

Note that using the time slider at the bottom isn't appropriate because we can't see when the previous/next time is.

Repeatedly opening the new Date Picker and navigating through the list of imagery is time consuming, and it would be better to have "step" buttons, maybe something like this (but please use better arrows):

screen shot 2017-11-16 at 18 30 25

a-stacey:
Fixes #2748

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### 5.#.#
+
+* Add previous and next buttons to workbench catalog item time picker.
+
 ### 5.6.0
 
 * Upgraded to Cesium 1.41.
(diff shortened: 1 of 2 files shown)


kring:
@chloe FYI Phil has a design for this in #2748.

kring:
#2815 does the same thing, closing this.

PR #2819, opened 2018-02-14
Refactor CatalogItem clock handling and fix split layer time bug.

a-stacey:
This is a fairly extensive refactor of how current time is handled in CatalogItems with the hope of simplifying future code and reducing bugs that were caused through duplicated state. The is also probably some legacy code that could be further removed or simplified and refactored by this, but I have not gone further to keep the size and scope of this change from blowing out completely (this could always be done subsequently or add hock as time goes on).

Please pay particular attention to the lines modified in e7c0cac8fccb6254ddc65ba45a36756597bd51bf as I feel unsatisfied that I have been able to exercise all cases to know whether it is reasonable to set currentTime to undefined when the layer does not have its own clock.


lib/Models/CatalogItem.js
```diff
@@ -320,13 +325,80 @@ var CatalogItem = function(terria) {
...
+        // This is defined as a property rather then a member to restrict acess to set. Clients can always
+        // `item.clock.currentTime = value` if they need to set, but since this is a copy set would need to copy the
+        // change to item.clock.currentTime.
+        get : function() {
+            return this._currentTime;
+        },
+    });
+
+    /**
+     * Gets the CatalogItems current time as the discrete time that time the catalogItem has information for.
```
kring: Extra word?  `time that time the`.

lib/Models/CatalogItem.js
```diff
@@ -1167,4 +1283,104 @@ function isShownChanged(catalogItem) {
...
+            catalogItem.terria.clock.currentTime = JulianDate.clone(catalogItem.clock.currentTime);
+        }
+    }
+
+    catalogItem._lastUseOwnClock = catalogItem.useOwnClock;
+}
+
+function timeAtIndex(catalogItem, index) {
+    if (defined(index) && defined(catalogItem.availableDates) && (index >= 0) && (index < catalogItem.availableDates.length)) {
+        return catalogItem.availableDates[index];
```
kring: According to the docs for the `availableDates` property, elements in `availableDates` can be either a Date or an interval.  I'm not sure offhand if it's true.

lib/Models/CatalogItem.js
```diff
@@ -320,13 +325,80 @@ var CatalogItem = function(terria) {
...
+        // `item.clock.currentTime = value` if they need to set, but since this is a copy set would need to copy the
+        // change to item.clock.currentTime.
+        get : function() {
+            return this._currentTime;
+        },
+    });
+
+    /**
+     * Gets the CatalogItems current time as the discrete time that time the catalogItem has information for.
+     * Returns undefined if it is not possible to query the time (i.e. the item doesn't have a clock or canUseOwnClock is false).
```
kring: Probably worthwhile to mention that this property also requires both `availableDates` and `intervals` to be available.

lib/Models/CatalogItem.js
```diff
@@ -809,6 +875,36 @@ CatalogItem.prototype.useClock = function() {
     });
 };
 
+/**
+ * Move the current time to the next time interval.
+ */
+CatalogItem.prototype.nextTime = function() {
```
kring: Maybe something like `moveToNextTime`? Function names should be verbs.

lib/Models/CatalogItem.js
```diff
@@ -809,6 +875,36 @@ CatalogItem.prototype.useClock = function() {
...
+ */
+CatalogItem.prototype.previousTime = function() {
+    updateIndex(this, previousIndex(this));
+};
+
+/**
+ * Whether it is possible to move to a time interval after the current time.
+ * @returns {Boolean} True if it is possible.
+ */
+CatalogItem.prototype.nextTimeAvaliable = function() {
```
kring: `isNextTimeAvailable`?

lib/Models/CatalogItem.js
```diff
@@ -320,13 +325,80 @@ var CatalogItem = function(terria) {
...
+            return timeAtIndex(this, getCurrentIndex(this));
+        },
+    });
+
+    // A property which defines when we are using the terria clock.
+    // We define this as a knockout property so that we can watch for changes to the propery triggered by observables
+    // that it depends on and update the subscription (watcher) on the terria.clock when the state changes.
+    knockout.defineProperty(this, '_mirroringTerriaClock', {
+        get: function() {
+          return this.isEnabled && this.isShown && !this.useOwnClock && defined(this.clock);
```
kring: Consider removing the check for `isShown`. That way the workbench controls will continue to update even while it's hidden from the map.

lib/Models/CatalogItem.js
```diff
@@ -1167,4 +1283,104 @@ function isShownChanged(catalogItem) {
...
+    }
+
+    return undefined;
+}
+
+function getCurrentIndex(catalogItem) {
+    return catalogItem.intervals.indexOf(catalogItem.currentTime);
+}
+
+function updateIndex (catalogItem, index) {
```
kring: super nit: extra space before open parenthesis.

lib/Models/CatalogItem.js
```diff
@@ -1167,4 +1283,104 @@ function isShownChanged(catalogItem) {
...
+    if (defined(time)) {
+        catalogItem.useOwnClock = true;
+        updateCurrentTime(catalogItem, JulianDate.fromDate(new Date(time)));
+        catalogItem.useClock();
+    }
+}
+
+function nextIndex(catalogItem) {
+    const index = getCurrentIndex(catalogItem);
+    if(defined(index) && (index < (catalogItem.intervals.length-1))) {
```
kring: more nits: missing space after if, and all the parenthesis are unnecessary except the outer ones.

lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -452,7 +433,7 @@ ImageryLayerCatalogItem.prototype._enable = function(layerIndex) {
     if (defined(this.intervals) && this.intervals.length > 0) {
         isTimeDynamic = true;
 
-        var clock = this.clockForDisplay;
+        var clock = this.terria.clock;
```
kring: Shouldn't this be `this.clock` instead of `this.terria.clock`?

kring:
This is looking really good.  Much cleaner! I'm going to spend a bit more time tomorrow testing a few things.

lib/Models/CatalogItem.js
a-stacey: Done 033d2275b31386248dc665e528ab15575cf2232c

lib/Models/CatalogItem.js
a-stacey: Done 033d2275b31386248dc665e528ab15575cf2232c

lib/Models/CatalogItem.js
a-stacey: Done 8920ba9db1d99cd9c0d16f9f24b340201cc6d107

lib/Models/CatalogItem.js
a-stacey: Done 8920ba9db1d99cd9c0d16f9f24b340201cc6d107

lib/Models/CatalogItem.js
a-stacey: Done 19cc3c1f3cbd09c81f2de94c34a560c6c7bbde87

lib/Models/CatalogItem.js
a-stacey: Done 19cc3c1f3cbd09c81f2de94c34a560c6c7bbde87

Note: I only removed one set of parenthesis since I consider the others reasonable practice since it somewhat stops people falling victim to wondering about operator precedence and makes the intention explicit.

lib/Models/CatalogItem.js
a-stacey: Thanks, discussed offline, have removed isShown, see git commit comment for more details c4e72dd40dc563d74e7918474586417bca0e59ae.

lib/Models/ImageryLayerCatalogItem.js
a-stacey: Thanks for picking this up. Offline we sorted out that multiplier should be terria.clock and currenttime should be this.clock.

This and another issue that I discovered in the process are addressed in : bbf0363bb1a3fc49f0165c90579cf47d48571628, ca085e492a4de6d320ee227f82ad830fc0ae9d27, 49403388f258d8c35c86004773d60cdd5366767e.

lib/Models/CatalogItem.js
a-stacey: Types would make it much easier to be sure. Looking at the code this doesn't seem to be the case:
```
                for (let i = 0; i < this.intervals.length; i++) {
                    datetimes.push(JulianDate.toDate(this.intervals.get(i).start, 3));
                }
```
https://github.com/TerriaJS/terriajs/blob/5212564af3c38f09e4e8327f241ff2849644be30/lib/Models/ImageryLayerCatalogItem.js#L187
https://github.com/TerriaJS/terriajs/blob/5212564af3c38f09e4e8327f241ff2849644be30/lib/Models/TableCatalogItem.js#L216

But this line has somewhat bothered me anyway since it was just refactored into that from the original line that it replaced, (https://github.com/TerriaJS/terriajs/pull/2819/files#diff-a8595cd3f06d548e02a70b13e4df2e6cL69, https://github.com/TerriaJS/terriajs/pull/2819/files#diff-a8595cd3f06d548e02a70b13e4df2e6cL74 and so on), but has bothered me since its not really symmetric where the value comes from. Have updated to fix the code now that I've read the docco around available dates since there shouldn't be much of a performance difference here anyway and this more symmetric implementation seems cleaner.

5212564af3c38f09e4e8327f241ff2849644be30

a-stacey:
As far as I know I've now addressed all the issues mentioned so this should be ready for another review.

lib/Models/CatalogItem.js
```diff
@@ -320,13 +325,81 @@ var CatalogItem = function(terria) {
...
 
+    /**
+     * Gets the CatalogItems current time.
+     *
+     * This property is an observable version of clock.currentTime, they will always have the same value.
+     *
+     * This property is effectively an interface adapter for clock.definitionChanged which changes the structure from
+     * an Event when the current time changes to a knockout property which can be observed.
+     *
+     * @member {JulianDate} The CatalogItems current time.
```
kring: The value in the position of `The` here is meant to be the name of the property. The reference doc currently looks like this as a result:
image

kring: Actually that image is for the property below, not this one, but same problem.

kring:
This looks great other than that one doc nit, so I'm going to fix that myself and then merge this.

PR #2831, opened 2018-02-20
Fix feature info updating when clock updates for TableCatalogItems

a-stacey:
This PR should be merged first #2819 as it is an upstream change that is needed for this work.

This PR is only aimed at fixing part of #2736.
>Feature info for TableCatalogItem updates when the time is changed using the timeline (including the date picker on the timeline), but it does not update when using the date picker on the workbench. Fixing this is tricky because there's currently no event that gets raised when the time changes via that mechanism, so the FeatureInfoSection can't subscribe to it.


lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -272,16 +272,12 @@ function setSubscriptionsAndTimeouts(featureInfoSection, feature) {
         setCurrentFeatureValues(changedFeature, featureInfoSection.props.clock);
     });
     if (featureInfoSection.isFeatureTimeVarying(feature)) {
-        if (defined(featureInfoSection.props.clock) && defined(featureInfoSection.props.clock.onTick)) {
+        if (defined(featureInfoSection.props.clock) && defined(featureInfoSection.props.clock.definitionChanged)) {
             featureInfoSection.setState({
-                removeClockSubscription: featureInfoSection.props.clock.onTick.addEventListener(function(clock) {
+                removeClockSubscription: featureInfoSection.props.clock.definitionChanged.addEventListener(function(clock) {
```
kring: Is this subscription needed anymore? I think you might just be able to change the UI to access `catalogItem.currentTime` and the observability will make the UI automatically refresh when it changes.

PR #2833, opened 2018-02-21
CatalogItem / DateTimeSelectorSection: Leave the timeslider visible when updating the current time.

a-stacey:
Updates the workbench controls so they no longer disable the timeslider.

I don't love that there are now two `updateCurrentTime()` functions in lib/Models/CatalogItem.js, but I can't think of a better name for either one (did try `setCurrentTime`, but didn't really feel that it differentiated between them enough or add any real value over direct name confusion).

Also not directly related to this commit, but something that I observed in the process was that I don't love the name of `useClock()`...doesn't really describe the functionality well...but again can't think of a better name right now.

### DIFF
--- a/lib/Models/CatalogItem.js
+++ b/lib/Models/CatalogItem.js
@@ -877,6 +877,22 @@ CatalogItem.prototype.useClock = function() {
 };
 
 /**
+ * Updates the current time for the catalog item.
+ * This function updates the correct clock (terria.clock or this.clock) depending on whether .useOwnClock is true or
+ * false so that the catalog items state will reflect the new time correctly.
+ *
+ * @param {JulianDate} The new time to update the current time to.
+ */
+CatalogItem.prototype.updateCurrentTime = function(newTime) {
+    if (this.useOwnClock) {
+        updateCurrentTime(this, newTime);
+    } else {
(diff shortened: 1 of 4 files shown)


lib/Models/CatalogItem.js
```diff
@@ -876,6 +876,22 @@ CatalogItem.prototype.useClock = function() {
...
 };
 
+/**
+ * Updates the current time for the catalog item.
+ * This function updates the correct clock (terria.clock or this.clock) depending on whether .useOwnClock is true or
+ * false so that the catalog items state will reflect the new time correctly.
+ *
+ * @param {JulianDate} The new time to update the current time to.
+ */
+CatalogItem.prototype.updateCurrentTime = function(newTime) {
```
kring: Could this be the setter on the `currentTime` property, rather than a function?

kring:
Looks good other than the one question.

lib/Models/CatalogItem.js
a-stacey: Oh yeah of course, I completely forgot about that when I wrote this, I will fix this now. Thanks for picking this up.

a-stacey:
This PR is ready for review merge.

PR #2878, opened 2018-03-05
Use Mapbox points in computeRingWindingOrder

### LINKED ISSUE #2740: computeRingWindingOrder always returns clockwise with vector tile {x,y} points
This is due to a difference in expressing points between `computeRingWindingOrder` and those used by `MapboxVectorTileImageryProvider`, so the below calculation ends up being full of NaNs.

https://github.com/TerriaJS/terriajs/blob/ec0570f75572df64c3cf91c28641ac36c62ed157/lib/Map/computeRingWindingOrder.js#L8-L11

`MapboxVectorTileImageryProvider` uses `{x,y}` (`@mapbox/simple-geometry` Points).

The end result is that, in a csv region mapped by state or similar, clicking ACT selects NSW as well as ACT (when zoomed out enough such that ACT is enclosed in NSW in one tile).
image

tephenavies:
Fixes #2740.

Could instead change Mapbox points into array-style points from `MapboxVectorTileImageryProvider`.


kring:
Reviewing this now.

Issue #2135, opened 2016-09-28
Change default autoPlay to False?


rsignell-usgs:
Currently the default is `autoPlay: True`
https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/BottomDock/Timeline/Timeline.jsx#L23
but in every instance of TerriaJS we have deployed, we always set `autoPlay: False`. 

In my experience, other GUIs that deal with geospatial information (e.g. Godiva2, Godiva3, IDV, Google Earth) require you to click the play button to animate.   I think this works well because most folks are interested in looking at information at specific time steps rather than watching an animation, and for many dynamic WMS source with time dependence, the loading is slow enough that animation is not very satisfying. 

What do other folks think?

kring:
I've marked this "Next Major Version" because, while it's simple and almost certainly a good idea, it's also a breaking change.

rsignell-usgs:
Okay!

RacingTadpole:
Yes, this was also requested (and set to off) for the NEII viewer.

kring:
At this point we've made this change in NationalMap, too. We should just go ahead and change the default in TerriaMap, no need to wait for the next major version.

rsignell-usgs:
Nice!

kring:
Fixed by #2886.

PR #2886, opened 2018-03-05
Issue #2135:  changed auto-play of time-series data to false

t83714:
Changes for issue #2135


kring:
I think this won't actually affect anything because the `autoPlay` property in `Terria.js` overrides the one in the UI. Am I mistaken?

t83714:
I was testing it using TerriaMap and this data file:
https://gist.githubusercontent.com/RacingTadpole/f308c8d01630b3d0b15d946e259b0cff/raw/266b62cab3716d76fe4c24f9788489ebfd8c7efa/bird_migration.csv
And it did make the differences. 
Will look into it and find out more.

t83714:
@kring I think there is an issue (https://github.com/TerriaJS/terriajs/blob/master/lib/Models/Terria.js#L460) in `Terria.js` to prevent its `autoPlay` property default value to be used. Have corrected this issue and changed `autoPlay` to false

PR #2887, opened 2018-03-05
Fix 2d rendering when latitude or longitude have bad characters

### LINKED ISSUE #2753: Dodgy characters in lat/lon column break 2D and are unreported
Rename this file to `dodgycharacters.csv` and drop it onto the map.
[dodgycharacters.csv.txt](https://github.com/TerriaJS/terriajs/files/1497756/dodgycharacters.csv.txt)

The CSV has three rows, but in 3D, only two points appear.  Switch to 2D, and only one point appears. The problem is that one of the rows has an extra dodgy character in its latitude column.  We should:
1) tell the user that some of the rows have an invalid value, and
2) show all the valid points in 2D just like we do in 3D.

tephenavies:
Fixes #2753.


kring:
My only concern here is that, if you have lots of rows with invalid lat/lon, you'll send the rest of your life dismissing modal dialogs.  Still better than previous, so I'm going to to merge this, but do you mind writing an issue for that?

tephenavies:
Good point. I thought of this, then forgot before I fixed the error. It's just in a loop, so it's pretty easy to only show the message once or twice.

PR #2897, opened 2018-03-07
Fix calculateImageryLayerIntervals properly

a-stacey:
You might have better ideas for how to include the function into the testcase, but that just worked and so left it like that (its a common paradigm that I've used in other languages, but there might be a cleaner way in javascript).

You might have ideas on better overall formatting, but hopefully the code and the comments are sound now. There are a few opportunities for refactoring to simplify still, but it seems that this is a reasonable tradeoff between simplicity compactness and readability.

### DIFF
--- a/lib/Models/RegionMapping.js
+++ b/lib/Models/RegionMapping.js
@@ -632,7 +632,7 @@ function addDescriptionAndProperties(regionMapping, regionIndices, regionImagery
  * @param {Boolean} animateForwards Whether we are currently animating forwards (true) or backwards (false) (this defines which direction .nextInterval will be in).
  * @param {Boolean} cyclic Whether we are animating cyclicly (if nextInterval falls beyond the end of the data, do we return .nextInterval as the first interval). The default for this value if not supplied is true.
  * @return {Object} A structure with a TimeInterval for .currentInterval and .nextInterval defining periods where the values from timeColumn are static.
- *
+ * @private
  */
 function calculateImageryLayerIntervals(timeColumn, currentTime, animateForwards, cyclic) {
     // Note: For this function we consider the case when isStartIncluded/isStopIncluded applies to the time. There is a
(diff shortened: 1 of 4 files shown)


lib/Models/RegionMapping.js
```diff
@@ -384,6 +384,12 @@ RegionMapping.prototype.loadRegionDetails = function() {
     });
 };
 
+// This is a test interface shim to provide test access to calculateImageryLayerIntervals(timeColumn, currentTime, animateForwards, cyclic).
+// This should only ever be called from test cases.
+RegionMapping.prototype._testCalculateImageryLayerIntervals = function(timeColumn, currentTime, animateForwards, cyclic) {
```
kring: I think the cleanest way to do this is to break the `calculateImageryLayerIntervals` function into a separate module, and then require it in here and in the tests.  You can mark it `@private` in the jsdoc comments so it doesn't show up in the documentation.

lib/Models/RegionMapping.js
```diff
@@ -616,49 +622,121 @@ function addDescriptionAndProperties(regionMapping, regionIndices, regionImagery
...
+/**
+ * Calculates intervals (near the current time) over which the values from the timeColumn do not change.
+ *
+ * This function will return .null for .nextInterval if we are in the last interval and cyclic is false.
+ *
+ * @param {*} timeColumn The time column containging the time intervals.
+ * @param {JulianDate} currentTime The current time to compute the current and next intervals for.
+ * @param {Boolean} animateForwards Whether we are currently animating forwards (true) or backwards (false) (this defines which direction .nextInterval will be in).
+ * @param {Boolean} cyclic Whether we are animating cyclicly (if nextInterval falls beyond the end of the data, do we return .nextInterval as the first interval). The default for this value if not supplied is true.
+ * @return {Object} A structure with a TimeInterval for .currentInterval and .nextInterval defining periods where the values from timeColumn are static.
```
kring: This needs an `@private` so it doesn't show up in the docs.

kring:
Looks good other than the two comments.

a-stacey:
@kring thanks have moved to a separate file, should be good to review again.

Issue #2852, opened 2018-03-01
Lat/Long coordinates to have 6 decimals


AnaBelgun:
From user feedback:
Why can the coordinates not be displayed to six, instead of only three decimals?

kring:
For reference, 3 decimal places (0.001 degrees) is a bit over 100 meters at the equator.   6 decimal places (0.000001 degrees) is about 0.1 meters. Mostly a UX call here though.

geoffrowland:
I tweaked TerriaMap.js to display 5 decimal places (0.00001 degrees) i.e. about 1m precision. This works with the default interface and is of roughly the same order as, or slightly better than, consumer GPS and UK LiDAR data.
So,
            this.latitude = Math.abs(latitude).toFixed(3) + '°' + (latitude < 0.0 ? 'S' : 'N');
            this.longitude = Math.abs(longitude).toFixed(3) + '°' + (longitude < 0.0 ? 'W' : 'E');
becomes
            this.latitude = Math.abs(latitude).toFixed(5) + '°' + (latitude < 0.0 ? 'S' : 'N');
            this.longitude = Math.abs(longitude).toFixed(5) + '°' + (longitude < 0.0 ? 'W' : 'E');

kring:
Thanks @geoffrowland. We'd be thrilled to have a pull request with that change, if you're up for it.

PR #2914, opened 2018-03-20
Display lat lon to 5 decimal places

geoffrowland:
Follow-up to issue #2852


kring:
Beautiful, thanks @geoffrowland!

PR #2921, opened 2018-03-27
Fixed Issue/2920 Error "This dataset is already shown"

### LINKED ISSUE #2920: `This data source is already shown` Error
`This data source is already shown` Error may sometimes be shown to the user.
See relevant issue in Magda project:
https://github.com/TerriaJS/magda/issues/652

t83714:
Fixed #2920 

Might be a timing issue to trigger this error (seems somewhere `_show` & `_hide` not called in pairs or in the correct sequence).

It's hard to track down to the root cause.

Muted the error from UI (still output to console) could be a reasonable solution for now.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Change Log
 ### 5.6.4
 
 * Fixed a bug causing an error message when adding tabular data to the workbench before it was loaded.
+* Fixed a issue that 'This data source is already shown.' error is shown to user.
 
 ### 5.6.3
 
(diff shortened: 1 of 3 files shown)


kring:
Thanks @t83714. Yeah I'm not sure exactly why you're seeing that error, but `ImageryLayerCatalogItem` already ignores that situation, so it's reasonable to do it for `TableCatalogItem` and `SensorObservationServiceCatalogItem`, too. The extra logging will help us keep an eye on it.

Issue #2933, opened 2018-04-12
Crash when useOwnClock is true and then changed to false before load completes


kring:
Easy to reproduce with this link:
http://ci.terria.io/master/#https://gist.githubusercontent.com/kring/6675e7ae88640a34341927f9fe179fa0/raw/4d8b126076be775958121163301d3e695523ea9b/useOwnClockCrash.json

The problem is that `useOwnClockChanged` in CatalogItem.js assumes that `catalogItem.clock` is defined if `useOwnClock` is true, but this will not be true if the catalog hasn't finished loading yet.

kring:
The example above looks slightly contrived, but reproducing it in normal use is easy too:
1. Add a WMS layer that defaults to `useOwnClock: true`, e.g. the Satellite Imagery -> Landsat layers in NationalMap.
2. Turn off `useOwnClock` by clicking the timeline icon on the workbench.
3. Click Share, then try to visit the share link.

kring:
This was partially addressed by #2935 in that the immediate crash no longer happens. However, the (contrived) init file above still has some weird problems (see #2935) and the whole thing still feels rather fragile.

PR #2935, opened 2018-04-13
[#2933] CatalogItem: Check clocks are defined when syncing time between own clock and terria clock.

a-stacey:


kring:
This definitely stops the crash, but it allows the timeline to get out of sync.  For example, visit:
http://ci.terria.io/as_2933/#https://gist.githubusercontent.com/kring/6675e7ae88640a34341927f9fe179fa0/raw/4d8b126076be775958121163301d3e695523ea9b/useOwnClockCrash.json

Then click somewhere in the middle of the timeline. Hit Share, and open the link in a new tab. The time displayed on the workbench will be incorrect.

Also, if you turn off `useOwnClock` before sharing, the timeline will incorrectly appear when you visit the share link.

This could be tricky to fix.

a-stacey:
Thanks for picking up. On inital inspection I'm also surprised that there is a complete breakdown between the timeline and the workbench time, I was expecting that if this caused issue it would only be during initalisation or switching...without inspecting further I'm a bit stumped as to why this has happened. The inialisation order really does present a bit of a problem here that will need more though to resolve this...

PR #2951, opened 2018-04-30
Share panel map image improvements

### LINKED ISSUE #1327: UI forces map image on share panel to be square no matter the actual aspect ratio
To reproduce:
- Switch map view to 2D
- Collapse data panel
- Zoom in a few levels
- Click Share button to view cropped preview

download

### LINKED ISSUE #2863: Leaflet screenshot in share panel is sometimes partially/fully greyed out
From user feedback: 
Great mapping system, however when you click share the image only generates half way, the rest remains greyed out.

Share URL: http://nationalmap.gov.au/#share=s-moFISXGO1CcTvemogPAIMhoAmfw

kring:
* Show the entire map instead of cropping it to a 4/3 aspect ratio.
* Upgrade to a newer html2canvas with some of our own bug fixes (details here: https://github.com/TerriaJS/html2canvas/tree/terriajs)
* Bonus unrelated feature: disable CI auto-deploy for `greenkeeper/*` branches.

Fixes #2863 
Fixes #1327


buildprocess/ci-deploy.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Don't run for greenkeeper branches; there are too many!
+if [[ $TRAVIS_BRANCH =~ ^greenkeeper/ ]]; then
```
tephenavies, line 4: Does this mean that greenkeeper won't tell us when packages break our build?

tephenavies: My bad. Didn't read the full description :)

kring: It would be nice to auto-deploy the greenkeeper stuff too, but there are so many the k8s cluster was running out of memory and exploding. A bigger cluster would solve the problem, but I don't think it's worth the cost.

PR #2958, opened 2018-05-06
[#2957] Remove fill attribute from splitter icons

michaeladams:
This is an PR for #2957

PR #2960, opened 2018-05-07
Update contributors

michaeladams:
As requested in #2958 I've updated the contributors.md.  We hope we can contribute more as we use Terria!


kring:
Thanks!

PR #2972, opened 2018-05-15
fix(package): update json5 to version 1.0.1

kring:
Closes #2905


kring:
Travis is failing because the latest version of json5 uses [codePointAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt), which is not supported in any version of Internet Explorer.

kring:
https://github.com/json5/json5/issues/170

tephenavies:
It looks like we could update this package to v2 and it should work on IE11

kring:
👍 we're also using/requiring a polyfill library as of TerriaJS v7, so that should help too.

PR #2988, opened 2018-05-30
Allow user entry for WPS point fields.

a-stacey:
At this stage I've only fixed for point parameters as described in the original issue. Given the effort involved its not exactly clear that the effort is worth updating it for all cases and this can serve as a template for other parameters once its in.

Overall I feel that this solution is somewhat over complex but given the UX considerations involved it seems hard to simplify.

One proposed improvement was to have a validation function which returns the validation errors (rather then the boolean isValid which stores no information on why validation failed), whilst I think that this would improve the code its not clear to me at this point it is worth the additional time and can always be refactored later to add this if / when we want to validate more complex parameters.

I used the following catalog item to test effects on other elements and couldn't detect style issues, but I'm also not confident that I covered all cases.
    {
      "name":"WPS",
      "description":"All services from WPS test server",
      "url":"http://sparrow.terria.io:5001/wps",
      "type":"wps-getCapabilities"
    },


lib/ReactViews/Analytics/PointParameterEditor.jsx
```diff
@@ -50,31 +112,6 @@ const PointParameterEditor = createReactClass({
...
-        parameter.value = Cartographic.fromDegrees(parseFloat(coordinates[0]), parseFloat(coordinates[1]));
-    }
-};
-
-/**
- * Given a value, return it in human readable form for display.
- * @param {Object} value Native format of parameter value.
- * @return {String} String for display
- */
-PointParameterEditor.getDisplayValue = function(value) {
```
kring: This function is part of the public API and used in Geoglam:
https://github.com/TerriaJS/geoglam-nm/blob/c945f4fbf6181121c83744018b03c44e5b657a58/lib/Views/GeoJsonParameterEditor.jsx#L106

I don't know if `setValueFromText` is used, but it's part of the public API too.

lib/ReactViews/Analytics/PointParameterEditor.jsx
```diff
@@ -27,21 +27,83 @@ const PointParameterEditor = createReactClass({
...
+                </If>
+                <input className={style}
                        type="text"
-                       onChange={this.setValueFromText}
-                       value={PointParameterEditor.getDisplayValue(this.props.parameter.value)}/>
+                       onChange={this.inputOnChange}
+                       onBlur={this.inputOnBlur}
+                       onFocus={this.inputOnFocus}
+                       value={this.getDisplayValue()}
+                       placeholder="-25.3450, 131.0361"/>
```
kring: Not your doing, but this placeholder is misleading. It has latitude first.

PR #2994, opened 2018-06-12
Upgrade to Cesium v1.46

kring:
This isn't ready to merge yet, because there are test failures (not sure how serious) and it needs more testing in general. This was an extremely painful Cesium upgrade.

With this release, you can use Cesium ion terrain by adding `"useCesiumIonTerrain": true` to the `parameters` section of `config.json`. To avoid a message in the credits at the bottom of the map and to comply with the Cesium ion license terms, you'll also need to set `"cesiumIonAccessToken"` with a token from your ion account.

CC @geoffrowland


geoffrowland:
Thanks @kring. Fully appreciate how much of a monster this upgrade was. Still, hope the greater resolution offered by Cesium Ion Terrain will be worth it!

PR #3037, opened 2018-07-24
Asynchronous HTTP 202 + user cancellation

Jason-DNetto:
Instantiating the cesium Request object in TerriaJsonCatalogItem and passing it down through to cesium Resource objects, so that the user clicking 'Remove' on the ResultPendingCatalogItem cancels the Request in the request object and stops creating xhr calls.


kring:
Thanks @Jason-DNetto! Can I bug you to make this a PR into the `cesium-1.46` branch? Much easier to review that way.

PR #3049, opened 2018-08-16
Fix missing legends for WMS items with isEnabled=true

### LINKED ISSUE #3043: Init file drag and drop of WMS isEnabled layer does not show legend, but does after sharing
When the attached init file is dragged and dropped on a terriajs instance with the layer `isEnabled: true` for some reason the legend is not shown.

However, if a share URL is generated and followed, the URL is then shown successfully.

The legend is also shown if using `isEnabled: false` and then going into the catalogue to enable it manually.

screen shot 2018-08-14 at 21 22 42
screen shot 2018-08-14 at 21 23 09

[aus_drought.json.txt](https://github.com/TerriaJS/terriajs/files/2286692/aus_drought.json.txt)

kring:
Fixes #3043.


kring:
CC @meh9 
Test it at http://ci.terria.io/missingLegends/ (once Travis does its thing)

PR #3067, opened 2018-09-03
Use own clock on split

### LINKED ISSUE #2918: Default timeline to off
At the moment when you enable a timeseries layer, such as the new Landsat imagery, it defaults to having the timeline at the bottom of the screen on, and the Use Timeline button is toggled on. If you now add a second timeseries layer it also defaults to on, and from that point the times of both layers is controlled by either time controllers or the timeline at the bottom, i.e. all the times are controlled in lock-step.

This is very confusing for the user because if they now turn on the splitter and attempt to compare different dates by selecting e.g. winter on the left and summer on the right, they are unable to control the layers individually without discovering that they have to disable the timeline on one or both layers. It is not intuitive how to turn that off.

However, if we had the opposite behaviour that the timeline defaults to off, then they can do the first usecase of controlling the layers in lock-step manually, and if they want to compare different times they can without figuring out how to first disable the timeline.

kring:
Fixes #2918

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 Change Log
 ==========
 
+### Next release
+
+* When clicking a `Split` button on the workbench, the new catalog item will no longer be attached to the timeline even if the original was. This avoids a confusing situation where both catalog items would be locked to the same time.
+
 ### v6.1.2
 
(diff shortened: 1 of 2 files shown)


kring:
Try it here: http://ci.terria.io/use-own-clock-on-split/

meh9:
Perfect.

PR #3070, opened 2018-09-05
Issue/2012 my data UI

### LINKED ISSUE #3096: Workbench collapse button is on top of the catalog UI
image

chloeleichen:
Update mydata ui, aim to 
1. streamline data upload
2. make drag and drop function more conspicuous.

For issue https://github.com/terriajs/terriajs/issues/2012

Also fixes #3096


kring:
Travis is reporting 7 test failures in this branch on every browser, so I think it's legit.

kring:
This is looking really good, such an improvement!

chloeleichen:
@philipgrimmett if a user adds a group, we need to put a bin next to the group name. currently, the whole group title is a button that toggles open/close. Any opinion on where/how the new button fit in this situation?

Here is how it looks now:
screen shot 2018-09-19 at 8 51 00 pm

kring:
IMO only the top-level things (whether they're groups or items) should have a bin icon. For example, in this:
image
"wms" should have a bin, but "ClearObservations" etc shouldn't. Put another way, only groups or items that are directly in the "user-added data" group can be removed.

kring:
Travis is showing some eslint warnings too.

chloeleichen:
Is there a staging url for this one for Phil to have a look? Otherwise it's ready 💃

kring:
> Is there a staging url for this one for Phil to have a look? 

Yep! http://ci.terria.io/issue%2f2012/

AnaBelgun:
@philipgrimmett reminder for feedback

kring:
image

Couple of small problems:
* [x] Showing the top-level "My Data" group isn't really necessary, and I suspect removing it will make the implementation easier (but if I'm wrong, it's not a big deal to leave it).
* [ ] ~The (+) buttons to add the layers are missing in the tree.~ Nevermind, it'll be weird to have an add button only on items that aren't at the top level. Maybe we want to have the bin _and_ the add icons, but that can be a separate issue.

kring:
I pushed a change to restore "User-Added Data" as the internal name of the the group, because renaming it breaks old share links. 😬  I don't think that should cause any problems, just a heads up.

lib/Core/canBeDeleted.js
```diff
@@ -0,0 +1,10 @@
+function canBeDeleted(catalogMember) {
```
kring: This should go in `Models` not `Core`. Also, is it needed? I think on the top-level groups and items that you create on the My Data tab, `removable` is true, and for any nested items and groups inside it `removable` is false. No need to check at every level.

lib/Core/canBeDeleted.js
```diff
@@ -0,0 +1,10 @@
...
+    if(catalogMember.isUserSupplied && catalogMember.parent){
+      if (catalogMember.parent.name === catalogMember.terria.catalog.userAddedDataGroup.name) {
```
kring: No need to check the names, just check the reference directly.
```suggestion
      if (catalogMember.parent === catalogMember.terria.catalog.userAddedDataGroup) {
```

lib/ReactViewModels/ViewState.js
```diff
@@ -170,10 +169,11 @@ export default class ViewState {
...
 
     viewCatalogMember(catalogMember) {
-        if (addedByUser(catalogMember)) {
-            this.userDataPreviewedItem = catalogMember;
-            this.openUserData();
-        } else {
+        if (catalogMember.isUserSupplied === true) {
+          this.userDataPreviewedItem = catalogMember;
+          this.openUserData();
+        }
```
kring: Sorry to nitpick formatting, but Terria code should be indented with four spaces and the `else` should be on the same line as the `}`. The spaces, at least, should come automatically from `.editorconfig` if your editor supports it (there's probably a plugin if not).

lib/ReactViews/DataCatalog/CatalogItem.jsx
```diff
@@ -10,13 +10,15 @@ import Styles from './data-catalog-item.scss';
 const STATE_TO_TITLE = {
     loading: 'Loading...',
     remove: 'Remove',
-    add: 'Add'
+    add: 'Add',
+    trash: 'delete dataset'
```
kring: Casing should be consistent with other titles. Also it might be a group rather than a dataset. And I think Remove makes it more clear that you're removing it from this view rather than deleting it completely from your computer or whatever.

```suggestion
    trash: 'Remove from catalogue'
```

lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
```diff
@@ -61,6 +59,15 @@ const DataCatalogGroup = createReactClass({
         this.props.viewState.viewCatalogMember(this.props.group);
     },
 
+    removeUserAddedGroup() {
+      const parent = this.props.group.parent;
+      // can remove if not root group
+      if(parent && parent.parent) {
```
kring: Do we need the `parent.parent` check? I don't think it'll ever be false currently, and if this function were used differently in the future it might actually get in the way.

Actually, given the name, something like this would be pretty sensible:

```
const itemIndex = this.props.terria.catalog.userAddedDataGroup.items.indexOf(this.props.group);
if (itemIndex >= 0) {
    this.props.terria.catalog.userAddedDataGroup.items.splice(itemIndex, 1);
}
```

lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
```diff
@@ -83,6 +90,8 @@ const DataCatalogGroup = createReactClass({
                 loading={group.isLoading}
                 emptyMessage="This group is empty"
                 onClick={this.clickGroup}
+                removable={canBeDeleted(this.props.group)}
+                removeUserAddedGroup ={this.removeUserAddedGroup}
```
kring: Actually, now that I see how this `removeUserAddedGroup` is used, I might get rid of it entirely. IMO it'd be fine if `CatalogGroup` knows how to remove a group from its parent (whatever it is) when `removable` is true and the user clicks that button.

kring: Rather than passing removeUserAddedGroup down as props, I mean.

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
```diff
@@ -15,7 +12,8 @@ import raiseErrorOnRejectedPromise from '../../Models/raiseErrorOnRejectedPromis
 const STATE_TO_TITLE = {
     loading: 'Loading...',
     remove: 'Remove this item',
```
kring:
```suggestion
    remove: 'Remove from map',
```

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
```diff
@@ -15,7 +12,8 @@ import raiseErrorOnRejectedPromise from '../../Models/raiseErrorOnRejectedPromis
...
-    add: 'Add this item. Hold down "shift" to keep the data catalogue open.'
+    add: 'Add this item. Hold down "shift" to keep the data catalogue open.',
+    trash: 'delete this user-added data.'
```
kring:
```suggestion
    trash: 'Remove from catalogue'
```

kring:
Problems in IE11:
image

chloeleichen:
the `CatalogItem.jsx` is just a pure ui component and does not know anything about `terria`, and given the same logic also happens in `DataCatalogItem` as opposed to `CatalogItem`, I think it's better to keep `removeUserAddedData` inside `DataCatalogGroup.jsx` and pass down as a props?

lib/Core/canBeDeleted.js
chloeleichen: removed `canBeDeleted.js`, i see that's not needed :)

lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
chloeleichen: yup~ put it in a new file

lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
chloeleichen: I think it's better to keep it in `DataCatalogGroup`  rather than passing `terria` and `group` into `CatalogGroup`? All the other actions are in DataCatalogGroup.

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
chloeleichen: done

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
chloeleichen: 👍

kring:
Found a bug:
1. Rename [bug.json.txt](https://github.com/TerriaJS/terriajs/files/2494286/bug.json.txt)
 to bug.json and drag it on to the map.
2. Click the Data Catalogue tab and scroll down to the bottom. You should see a new layer called "Test".
3. Attempt to click on "Test". Instead of displaying its details, Terria switches to the My Data tab.

Also, there's no (+) button in the catalogue view for this dataset.

I _think_ this is a result of your change to use `isUserSupplied` to distinguish between items/groups on My Data versus Data Catalogue. But `isUserSupplied` means something different. The comments on the property in CatalogMember.js say:

```
     * Gets or sets a value indicating whether this member was supplied by the user rather than loaded from one of the
     * {@link Terria#initSources}.  User-supplied members must be serialized completely when, for example,
     * serializing enabled members for sharing.  This property is observable.
```

So a catalog item that comes from dragging/dropping a JSON file gets the `isUserSupplied` property because it comes from JSON that only exists in the user's browser, and so sharing that catalog item requires serializing the entire thing, not just the usual sharable properties.

The correct way to determine if a catalog item/group is on the My Data tab is by walking up its parent chain until you find (or don't find) the `userAddedDataGroup`.

lib/ReactViewModels/ViewState.js
```diff
@@ -170,7 +169,7 @@ export default class ViewState {
     }
 
     viewCatalogMember(catalogMember) {
-        if (addedByUser(catalogMember)) {
+        if (catalogMember.isUserSupplied === true) {
```
kring: I think the original code was correct. Was it causing problems?

chloeleichen: yeah i think that was some old changes sneaking in from a messy merge

chloeleichen:
@kring Fixed! 💃

kring:
Looks great, thanks!

PR #3072, opened 2018-09-07
Add LeafletDataSourceDisplay.

kring:
Instead of trying to use the Cesium one for Leaflet.
See here for discussion: https://github.com/AnalyticalGraphicsInc/cesium/pull/6947

This fixes a bug that caused a crash when switching to 2D with vector data already on the map, including when visiting a share link with vector data when the map ends up being 2D.


kring:
Try it here: http://ci.terria.io/leaflet-data-source-display/

kring:
CC @meh9

meh9:
Yeah that works!

PR #3076, opened 2018-09-14
Support zooming to KML and Cesium 3D Tiles

### LINKED ISSUE #2547: Make Zoom To work with KML
"zoom to extent" not working with kml.
I have also tried it in the past with other upload formats and I think I had the same problem. 
I have just tried with the file below. 

[MurrayLLS VegMachine 30_May_2017.zip](https://github.com/TerriaJS/terriajs/files/1051173/MurrayLLS.VegMachine.30_May_2017.zip)

kring:
Don't merge this yet, it only works in 3D so far.

Fixes #2547 

CC @t83714 @aneesha09


t83714:
👍

aneesha09:
Woohoo!

AnaBelgun:
canZoomTo property? @reginapramesti

lib/Models/Leaflet.js
```diff
@@ -236,34 +236,85 @@ Leaflet.prototype.getContainer = function() {
...
+    return radians * 180 / Math.PI;
+}
+
+function entitiesToLatLngBounds(entities) {
+    var maxBounds;
+
+    for (var i = 0, len = entities.length; i < len; i++) {
+        var latLngs = [];
+
+        if (defined(entities[i].polygon)) {
```
kring: There are other types of graphical elements, other than polygons, that should be supported. You can get a list here:
https://cesiumjs.org/Cesium/Build/Documentation/Entity.html?classFilter=Entity

But I think rather than try to support all those things, it might be wise to delegate this to the `LeafletVisualizer`, in the same way `Cesium.js` delegates it to the `DataSourceDisplay`. The visualizer knows the Leaflet elements it creates for each type of graphical primitive and can get a bounding box for them.

kring:
Here's another KML file you can test with:
https://github.com/TerriaJS/nationalmap/blob/master/wwwroot/test/vic_police.kmz

reginapramesti:
@kring could you have a look to see if this is better? Most of the work is now done in `LeafletVisualizer`

lib/Map/LeafletVisualizer.js
```diff
@@ -185,6 +186,41 @@ LeafletGeomVisualizer.prototype.update = function(time) {
...
+ */
+LeafletGeomVisualizer.prototype.getLatLngBounds = function() {
+    var entities = this._entitiesToVisualize.values;
+    var bounds = [];
+
+    for (var i = 0, len = entities.length; i < len; i++) {
+        var entity = entities[i];
+        var latLngs;
+
+        if (defined(entity._point) || defined(entity._billboard) || defined(entity._label)) {
```
kring: This works, and is nicely written, but I think it would be better to let Leaflet do more of the work. Every Leaflet layer has a `getBounds` method. So `this._entityHash` is a hash of entity IDs to "details" objects. The details object has properties like `point`, `polygon`, etc. with details of that type of graphical primitive associated with that entity. Then each graphical primitive object as a `layer` property which is a Leaflet layer. Call `getBounds` on each layer, `extend` them into a single `LatLngBnds`, and you have your answer! 

You can write code that does this without even knowing the different types of graphical elements. E.g. something like:

```
Object.keys(this._entityHash).forEach(entityId => {
    const entityDetails = this._entityHash[entityId];
    Object.keys(entityDetails).forEach(primitiveId => {
        const primitive = entityDetails[primitiveId];
        if (primitive.layer !== undefined && primitive.layer.getBounds !== undefined) {
            const bounds = primitive.layer.getBounds();
            // combine bounds with overall bounds with extend
        }
    });
});

File: lib/Models/Cesium.js [hunk 3]
```diff
@@ -421,28 +430,84 @@ Cesium.prototype.zoomTo = function(viewOrExtent, flightDurationSeconds) {
...
+                return false;
+            } else if (state !== BoundingSphereState.FAILED) {
+                boundingSpheres.push(BoundingSphere.clone(boundingSphereScratch));
+            }
+        }
+
+        var boundingSphere = BoundingSphere.fromBoundingSpheres(boundingSpheres);
+        cesium.scene.camera.flyToBoundingSphere(boundingSphere, {
+            duration : flightDurationSeconds
+        });
```
kring, line 500: This is my fault, but this `pollToPromise` isn't going to work right because the function never returns `true`! So it will just repeat this thing every 100ms until it times out 5 seconds later. I think it just needs a `return true` after the `flyToBoundingSphere`.

lib/Map/LeafletDataSourceDisplay.js
```diff
@@ -196,6 +197,39 @@ LeafletDataSourceDisplay.prototype.update = function(time) {
...
 };
 
+/**
+ * Computes the rectangular bounds which encloses the visualization
+ * produced for the given entities.
+ *
+ * @param {EntityCollection} entities The collection of entities whose bounds to compute.
+ * @returns {LatLngBounds} The computed bounds.
+ */
+LeafletDataSourceDisplay.prototype.getLatLngBounds = function(entities) {
```
kring: The caller of this function already has a data source (in the `target` variable), so if you passed that into this function instead of `target.entities`, you wouldn't need to find the data source below. You could skip down to `var visualizers = dataSource._visualizersByDisplayID[this._displayID]`.

lib/Map/LeafletDataSourceDisplay.js
```diff
@@ -196,6 +197,39 @@ LeafletDataSourceDisplay.prototype.update = function(time) {
...
+
+        if (entities.id === dataSource._entityCollection.id) {
+            var visualizers = dataSource._visualizersByDisplayID[this._displayID];
+            
+            for (var j = 0, vLength = visualizers.length; j < vLength; j++) {
+                var visualizer = visualizers[j];
+                var bounds = visualizer.getLatLngBounds();
+    
+                if (defined(bounds)) {
+                    latLngBounds.push(bounds);
```
kring: Not a big deal at all, but it would be slightly better (faster, less memory usage) to compute the bounds as you go with `extend`, rather than pushing all the bounds into an array and using `combineLatLngBounds`.

For bonus points, do this as a one-liner (eh, maybe two) with a `reduce` over the visualizers array.

lib/Map/LeafletVisualizer.js
```diff
@@ -185,6 +186,38 @@ LeafletGeomVisualizer.prototype.update = function(time) {
...
+    var bounds = [];
+
+    Object.keys(this._entityHash).forEach(entityId => {
+        const entityDetails = this._entityHash[entityId];
+
+        Object.keys(entityDetails).forEach(primitiveId => {
+            const primitive = entityDetails[primitiveId];
+
+            if (defined(primitive.layer)) {
+                if (defined(primitive.layer.getBounds)) {
```
kring, line 204: It'd be more beneficial here to use `extend` rather than putting all the coordinates and bounds into an array and combining them, cause the array could be big-ish with a lot of points.

kring:
This is look really good! Just a couple suggestions to improve the code slightly.

kring:
I ended up doing the last couple of small things because this turned out to be wrapped up in some other work I was doing. Thanks for all your work on it though @reginapramesti!

PR #3086, opened 2018-09-19
fix firefox drag and drop abandon

chloeleichen:
Fixed the issue that you cannot abandon drag and drop in firefox (see https://github.com/TerriaJS/terriajs/issues/3036).

One of the changes is removing the `screenX` and `screenY` check in `handleDragLeave` function in `DragDropFile.jsx`. I don't see why that line is necessary, and it seems that firefox does not set those values to 0 when the drag is finished. Please let me know if that check is there for other reasons.

### DIFF
--- a/lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
+++ b/lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
@@ -57,15 +57,25 @@ const StandardUserInterface = createReactClass({
     /* eslint-disable-next-line camelcase */
     UNSAFE_componentWillMount() {
         const that = this;
-        this.dragOverListener = e => {
+
+        this.dragEnterListener = e => {
+            if (!e.dataTransfer.types || !arrayContains(e.dataTransfer.types, 'Files')) {
+                return;
+            }
+            e.preventDefault();
+            e.stopPropagation();
+            console.log("drag enter")
+            that.acceptDragDropFile();
+        };
+
+        this.dragOverListener = e =>{
             if (!e.dataTransfer.types || !arrayContains(e.dataTransfer.types, 'Files')) {
                 return;
             }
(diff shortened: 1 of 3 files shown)


lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
```diff
@@ -57,16 +57,26 @@ const StandardUserInterface = createReactClass({
...
         const that = this;
-        this.dragOverListener = e => {
+
+        this.dragEnterListener = e => {
             if (!e.dataTransfer.types || !arrayContains(e.dataTransfer.types, 'Files')) {
                 return;
             }
             e.preventDefault();
             e.stopPropagation();
-            e.dataTransfer.dropEffect = 'copy';
```
kring: Are you sure this isn't needed for IE or something?

lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
```diff
@@ -57,16 +57,26 @@ const StandardUserInterface = createReactClass({
...
+            console.log("drag enter")
```
kring: Please remove this.

lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
kring: Oh I see you just moved it, nevermind.

kring:
This isn't working very well in Chrome on Windows for me. It flickers as I move the mouse around. See the gif below. I suspect because a "dragleave" event is being generated for individual element, not just for the whole window.
drag-drop

chloeleichen:
Found a better fix! @kring ready for another look! 💃 💃 💃

kring:
Thanks @chloeleichen!

PR #3100, opened 2018-10-03
Issue/3094

### LINKED ISSUE #3094: Include terria logo by default on TerriaMaps
Currently the terriaMaps do not display Terria logo or attribution at all. 
some options:

1. To include terria logo next/before Cesium ION logo, at the bottom on the map (in line with attributions etc)

chloeleichen:
Fixes #3094

A bit of a hacky approach, but let us config all options independently and meet styling requirements. Let me know what you think


lib/ViewModels/TerriaViewer.js
```diff
@@ -741,6 +753,15 @@ function initializeTerrainProvider(terriaViewer) {
         terriaViewer._terrainProvider = new EllipsoidTerrainProvider();
         CreditDisplay._cesiumCredit = undefined;
     }
+
+    // add terria logo unless specified not to
+    if (!terriaViewer.terria.configParameters.hideTerriaLogo){
+      const terriaLogo = require('../../wwwroot/images/terria-watermark-onLight.svg');
+      terriaViewer._defaultTerriaCredit = new Credit('<a href="https://terria.io/" target="_blank" rel="noopener noreferrer" ><img src="' + terriaLogo + '" title="Cesium ion"/></a>', true);
+      CreditDisplay._terriaCredit = terriaViewer._defaultTerriaCredit;
```
kring: Why assign this to `CreditDisplay`? Doesn't seem to be used.

lib/ViewModels/TerriaViewer.js
```diff
@@ -338,6 +338,16 @@ us via email at '+that.terria.supportEmail+'.'
 
     scene.globe.depthTestAgainstTerrain = false;
 
+    // if both terriajs default credit and cecium ion credits are required
+    if(defined(this._defaultTerriaCredit) && defined(this._bumpyTerrainCredit)){
+      // This line worked for sometimes, then stopped working. Therefore I also use CSS to hide it in case it doesn't work
+      scene.frameState.creditDisplay.removeDefaultCredit(this._bumpyTerrainCredit);
+      const combinedCredit = new Credit(`<div id='logo-credit'>${this._defaultTerriaCredit._html} ${this._bumpyTerrainCredit._html}</div>`);
```
kring: I think you can add multiple default credits, instead of creating a combined one. Or did that not work?

kring:
Some problems:
* The Cesium Ion logo now shows up even when using 3D Smooth.
* 2D looks weird:
image

lib/ViewModels/TerriaViewer.js
chloeleichen: it is to make them display in the layout we want

lib/ViewModels/TerriaViewer.js
chloeleichen: ah no it's not needed, will remove that

lib/ViewModels/TerriaViewer.js
kring: Is it possible to use CSS to achieve the layout we want, instead of combining the credits? I think we already have some logic to treat the first credit as special, so it would just be a matter of treating the first two specially, right?

chloeleichen:
css hack version, ready for another look @kring!

Issue #3092, opened 2018-09-25
Investigate 503 errors for ArcGIS generated data service


AnaBelgun:
Recent user feedback exposed consistent errors with data services generated from ArcGIS servers (e.g. GA, WA SLIP); with WA users reporting errors more in the afternoon than morning and only recently happening (4 weeks).

AnaBelgun:
Investigated and increased the TerriaJS tolerance in v6.2.0 release
https://github.com/TerriaJS/terriajs/blob/master/CHANGES.md

Issue #2927, opened 2018-03-28
Support RESTful access to WMTS servers


kring:
TerriaJS currently only requests tiles from WMTS servers using KVP (query parameter) encoding. Unfortunately, some servers don't support this encoding and require the use of RESTful (URL) encoding. Here's an example of such a server:
https://tiles.openaerialmap.org/5abae68e65bd8f00110f3e42/0/5abae68e65bd8f00110f3e43/wmts

This _may_ also be the reason for the problem described in #1573.

stevenjh:
I think this is another example of this use case, WMTS served from the https://koordinates.com/ data sharing platform.

[New Zealand LINZ data service](https://data.linz.govt.nz/layer/95497-auckland-0075m-urban-aerial-photos-2017/webservices/) (note you'll need a free api key)
https://data.linz.govt.nz/services;key=YOUR_API_TOKEN/wmts/1.0.0/layer/95497/WMTSCapabilities.xml

returns 

`<ResourceURL format="image/png" resourceType="tile" template="https://tiles-a.data-cdn.linz.govt.nz/services;key=YOUR_API_TOKEN/tiles/v4/layer=95497,{style}/{TileMatrixSet}/{TileMatrix}/{TileCol}/{TileRow}.png"/>`

The service will add, but tile requests fail. Does this suggest approaching WMTS as outlined in issue [1371](https://github.com/TerriaJS/terriajs/issues/1371) is required?

kring:
Definitely looks like an example of this issue, at least. Terria is ignoring the `ResourceURL` and instead constructing KVP GetTile requests manually. So if the server doesn't support that (I haven't checked), it's not going to work. But do check that that's really the issue. It could be something simpler, like the `key` isn't being included in the `GetTile` requests. It looks like Terria doesn't currently support the `parameters` property for WMTS that it does for WMS, so there's not a built-in way to add that `key`. But it'd be easy to add it.

stevenjh:
Thanks for the reply, the `key` was being carried across into the requests just as part of the base url in the KVP requests.

Imagine supporting REST would involve refactoring `/lib/Models/WebMapTileServiceCatalogItem.js` and the cesium `WebMapTileServiceImageryProvider` instances that are created, as well as the leaflet instances which I couldn't spot where that was.
Is there more to this that I'm overlooking?

kring:
[...]

PR #3104, opened 2018-10-04
Improve tile error handling

### LINKED ISSUE #2928: Endless tile requests on certain kinds of errors
Certain types of server errors may cause TerriaJS to repeatedly request the same (failing) tiles over and over again, wasting bandwidth and potentially starving other requests. Specifically, if an error occurs while loading an image but then an XMLHttpRequest to the same URL succeeds, TerriaJS will retry the image and repeat this process endlessly. 

Here's an example of a WMTS server that triggers the problem (at least until we address #2927):
https://tiles.openaerialmap.org/5abae68e65bd8f00110f3e42/0/5abae68e65bd8f00110f3e43/wmts

TerriaJS should limit the number of times it is willing to retry image loads. It should also more carefully validate the XMLHttpRequest response. For example, it should set the `Accepts` header and it should verify that the response is actually an image.

kring:
* Fixed a bug that could cause endless tile requests with certain types incorrect server responses.
* Fixed a bug that could cause endless region tile requests when loading a CSV with a time column where none of the column values could actually be interpreted as a time.
* Added automatic retry with jittered, exponential backoff for tile requests that result in a 5xx HTTP status code. This is especially useful for servers that return 503 or 504 under load. Previously, TerriaJS would frequently disable the layer and hit the user with an error message when accessing such servers.

Fixes #2928 (both the WMTS problem and the unrelated CSV problem)
Closes #3083 
Is probably a reasonable workaround for #3092

Sadly I didn't really make this code any more sane. It's still totally crazy.


meh9:
Can confirm unrelated CSV problem fixed.

meh9:
Can confirm that we are now very much more resistant to 503's at least, tested on AREMI with the following layers plus Australian Topography basemap, and it took a very large number of 503's before the layers started to actually fail. These show the ones that did and did not fail + Australian Topography:

screen shot 2018-10-05 at 10 57 28

I probably got 1000 or so 503's before things started failing, so I'd say we should go ahead and merge.

lib/Map/TableStructure.js
```diff
@@ -451,8 +451,10 @@ function calculateFinishDatesFromStartDates(startJulianDates, localDefaultFinalD
         var n = sortedUniqueJulianDates.length;
         if (n > 1) {
             finalDurationSeconds = JulianDate.secondsDifference(sortedUniqueJulianDates[n - 1], sortedUniqueJulianDates[0]) / (n - 1);
+            endDates.push(JulianDate.addSeconds(sortedUniqueJulianDates[n - 1], finalDurationSeconds, new JulianDate()));
+        } else {
+            endDates.push(undefined);
```
tephenavies, line 456: @kring Is this the right thing to do for `n === 1`?

kring: 😬 I think so. The way `endDates` is used below, it looks like there has to be something added, or the indices will be wrong.

lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -586,84 +588,67 @@ ImageryLayerCatalogItem.enableLayer = function(catalogItem, imageryProvider, opa
             }
 
             if (tileProviderError.timesRetried === 0) {
+                // There was an intervening success, so restart our count of the tile failures.
                 tileFailures = 0;
+                tileRetriesByMap = {};
```
stevage, line 603: retriesPerTile would be a better name.

stevage: Also, is it definitely correct to reset it whenever there is any "intervening success"? If tile X has failed 3 times, then some other random tile succeeds, will this (incorrectly) get reset?

lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -705,9 +690,98 @@ ImageryLayerCatalogItem.enableLayer = function(catalogItem, imageryProvider, opa
...
+                        // could not load it as an image.
+                        failTile({
+                            name: 'Tile Error',
+                            message: 'Got a 200 (OK) response from URL:\n' +
+                                     getUrlForImageryTile(imageryProvider, tileProviderError.x, tileProviderError.y, tileProviderError.level) + '\n' +
+                                     'but it could not be loaded as an image.'
+                        });
+                    } else if (!defined(e.statusCode)) {
+                        // On a modern-ish browser, this is almost certainly a CORS problem.
+                        // Note that ignoreUnknownTileErrors is only for genuinely unknown (no status code) errors.
```
stevage, line 778: This situation also arises on DNS failures (eg, non existent domain)

kring:
@stevage reports that https://tiles.openaerialmap.org/5abae68e65bd8f00110f3e42/0/5abae68e65bd8f00110f3e43/wmts doesn't work on http://ci.terria.io/fail-better/
It does the GetCapabilities successfully, but then doesn't request any tiles for some reason.

stevage:
Actually the last time I looked at the CI, it seemed to be fetching tiles, but then failing silently. I really don't know what's going on. Possibly totally spurious.

kring:
Ok here's what's happening with that openaerialmap server...

The extent of the one layer on that server is a tiny little square near Cape Code, MA, USA. So when you're looking at Australia, Cesium only downloads a couple of tiles. But those tiles all return XML, probably because of #2927, which mentions this exact server. Because we only download a couple of tiles, though, it doesn't fail and show the user a message immediately. But if you try to zoom to the layer's extent, you'll get an error like this pretty quickly:

image

So I think everything is working as expected, but let me know if anything still seems off to you @stevage.

lib/Models/ImageryLayerCatalogItem.js
kring: It's definitely not _correct_, but the effect should be that TerriaJS is more forgiving of errors, which is closer to a :feature: than a :bug:.

lib/Models/ImageryLayerCatalogItem.js
kring: Good point!

kring:
I think I've addressed all the review comments @steve9164 and @stevage.

stevage:
>The server denied Cross-Origin Resource Sharing (CORS) access to this URL

More often than not, it will be the browser denying access on the basis that the server didn't provide any CORS instructions at all, right?

stevage:
(I thought I had already pushed these commits, but I pushed them to the wrong branch...)

kring:
> More often than not, it will be the browser denying access on the basis that the server didn't provide any CORS instructions at all, right?

Very true, but I think explaining that just muddies the water. The browser can't be changed, but the server potentially can be, so it's reasonable to explain it as something that has to change on the server. From the browser's perspective, a lack of explicit approval is the same as denial.

PR #3119, opened 2018-10-12
Use Ion for Bing imagery

kring:
With this PR, if a `bingMapsKey` is provided in config.json, nothing changes.

If a `bingMapsKey` is _not_ provided, we access the Bing Maps basemaps via Cesium Ion, using Cesium's default Ion key if none is provided. Because Ion can't be used for Bing geocoding (it has its own geocoder), failing to provide a `bingMapsKey` will also mean that the Bing geocoder will always return no results.

You can tweak this behavior with `useCesiumIonBingImagery` in config.json. If it's set to true, we'll use Ion for the Bing basemaps even if a `bingMapsKey` is provided. If it's set to false, and a `bingMapsKey` is not provided, the Bing basemaps will disappear completely.

Note that if you're using Bing via Ion, you'll get an Ion logo on the map even in 2D or 3D Smooth.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### Next Release
+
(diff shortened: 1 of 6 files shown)


rsignell-usgs:
Will TerriaJS automagically use the Ion geocoder if `bingMapsKey` is not provided?

kring:
Unfortunately no, not currently. Ion uses a totally different API for its geocoder, so we'll have to write some code to support it.

PR #3130, opened 2018-10-22
Remove sos:procedure entries from GetFeatureOfInterest request.

### LINKED ISSUE #3118: Remove sos:procedure from initial SOS calls
The BOM SOS services aren't working, because the server times out and falls over. Apparently our request is too onerous. According to Rudi @ BOM:

>As I understood from your emails, all you're after is the list of sites with "water course level" data. You should effectively be able to remove all the "<sos: procedure>" lines out of your list the "GetFeatureOfInterest" call will filter on as the call "<sos:ObservedProperty>" will give you this list.

>In fact, the way our system is set up, if there is an observed parameter for a station, it automatically has all these time series types rolled out, regardless of whether there is any data. It's also pretty rare to have an observed parameter but no data as the system setup is driven by the incoming data.

>Also, the "<sos:temporalFilter>" is not an option for the "GetFeatureOfInterest" call and gets ignored in the query.

>If you do want this feature (filtering for data since 1980), then you'll have to use the "GetDataAvailability" call, use the "ObservedProperty" to extract all water course level information and do some postprocessing to filter the list on "beginPosition" in the "phenominonTime – TimePeriod" output.

stevage:
Fixes #3118.
Not entirely sure why they were there before. BoM's services interpret their
presence in a way which means vast quantities of data are being processed,
which makes the request fail.

### DIFF
--- a/lib/Models/SensorObservationServiceCatalogItem.js
+++ b/lib/Models/SensorObservationServiceCatalogItem.js
@@ -642,7 +642,7 @@ SensorObservationServiceCatalogItem.prototype.showOnSeparateMap = function(globe
 function loadFeaturesOfInterest(item) {
     const filter = {
         observedProperty:  item.observableProperties.map(observable => observable.identifier)  // eg. 'http://bom.gov.au/waterdata/services/parameters/Storage Level'
-    }
+    };
     if (item.filterByProcedures) {
         filter.procedure = item.procedures.map(procedure => procedure.identifier); // eg. 'http://bom.gov.au/waterdata/services/tstypes/Pat7_C_B_1_YearlyMean',
     }
--- a/lib/Models/SensorObservationServiceCatalogItem.js
+++ b/lib/Models/SensorObservationServiceCatalogItem.js
@@ -179,6 +179,13 @@ var SensorObservationServiceCatalogItem = function(terria, url) {
      */
     this.representAsGeoJson = false;
 
+    /**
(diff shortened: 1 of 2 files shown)


lib/Models/SensorObservationServiceCatalogItem.js
```diff
@@ -626,18 +633,23 @@ SensorObservationServiceCatalogItem.prototype.showOnSeparateMap = function(globe
     }
 };
 
+/**
+ * Performs the GetFeatureOfInterest request to obtain the locations of sources of data that match the required
```
kring, line 637: Don't use jsdoc comments for private functions, because then they'll show up in the reference docs even though they're inaccessible. You can use regular comments. Or if you really want to use jsdoc comments, add `@private`.

kring (commented):
Looks good to me, other than one super trivial comment. Oh and update CHANGES.md, too, please.
@RacingTadpole can you please take a quick look at this? It's not a big change.

meh9:
@stevage when I try it with NEII Viewer it still times out, do I also need to remove the `procedures` from the init file entry?

https://github.com/TerriaJS/neii-viewer/blob/master/wwwroot/init/neii-water.json#L635-L666

meh9:
```
DeveloperError: Both `procedures` and `observableProperties` arrays must be defined on the catalog item.
```
:'(

stevage:
@meh9 I already submitted the required PR here: https://github.com/TerriaJS/neii-viewer/pull/112

PR #3135, opened 2018-10-24
Checkbox groups prototype

Jason-DNetto:
this performed the diff correctly, last one must have been because too many commits


Jason-DNetto:
I updated this with one that actually has its functionality in place

Jason-DNetto:
init file for testing https://gist.github.com/Jason-DNetto/7ccde078dd517be7b279855fdf241cbc

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -371,10 +371,17 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
     if (!defined(parameter.value) || parameter.value === "") {
       return;
     }
-
-    queryParameters[parameter.id] = parameter.formatForService();
+    // determine if parameter is a group of parameters and 
+    // if parameter has groupIsTransparent in its AvailableFormatters
+    // first working prototype and would appreciate feedback.
+    if (parameter.type.endsWith('-group') && ('groupIsTransparent' in parameter.availableFormatters)) {
+        parameter.formatForService().forEach(function(value) {
```
na9da: I think this will fail if `formatter` is set to `groupIsOpaque` and the return value of `formatForService()` is a `string`.

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -371,10 +371,17 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
...
       return;
     }
-
-    queryParameters[parameter.id] = parameter.formatForService();
+    // determine if parameter is a group of parameters and 
+    // if parameter has groupIsTransparent in its AvailableFormatters
+    // first working prototype and would appreciate feedback.
+    if (parameter.type.endsWith('-group') && ('groupIsTransparent' in parameter.availableFormatters)) {
+        parameter.formatForService().forEach(function(value) { 
+          queryParameters[Object.keys(value)] = value[Object.keys(value)];
```
na9da: How about making this more explicit?

If each value is of type `{id: string, value: string}`, then you could  rewrite it as `queryParameters[param.id] = param.value;`

lib/Models/BooleanParameterGroup.js
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * @param {Object} [options] Object with the following properties:
+ * @param {Terria} options.terria The Terria instance.
+ * @param {String} options.id The unique ID of this parameter.
+ * @param {String} [options.name] The name of this parameter.  If not specified, the ID is used as the name.
+ * @param {String} [options.description] The description of the parameter.
+ * @param {Boolean} [options.allValue] Whether or not to have OneForAll.value in this.value
+ * @param {Boolean} [options.allCascade] Whether OneForAll being true passes all ParameterList.id: true values through to this.value
+ * @param {BooleanParameter} [options.OneForAll] A Boolean Parameter that checks and disables the BooleanParameters in the ParameterList when checked, and unchecks and enables them when unchecked. Used to send one option for the entire category/have one option click/reset all options in the category
+ * @param {[]BooleanParameter} [options.ParameterList] An array of Boolean Parameters that belong to a category
```
na9da: This should be `BooleanParameter[]` instead.

na9da (commented):
Looks good, just a few minor changes needed.

Please also run `prettier` on [these files](https://travis-ci.org/TerriaJS/terriajs/jobs/565313559#L511).

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: the values were not necessarily like that, one example I had was 
value = {"C": "true"} which would become 
queryParameters["C"] = true.
can switch to that.

lib/Models/BooleanParameterGroup.js
```diff
@@ -0,0 +1,186 @@
...
+  groupIsOpaque: formatAsStringOfKeyEqualsBoolean,
+  groupIsTransparent: formatAsArrayOfObjects
+};
+
+function formatAsArrayOfObjects(value) {
+  return value;
+}
+
+function formatAsStringOfKeyEqualsBoolean(value) {
+  return value
```
na9da: This also needs to be fixed to reflect the `{id, value}` change.

na9da (approved):
Looks good now.

PR #3146, opened 2018-11-09
Clean up/refactor showDataForTime()

stevage:


stevage:
I believe there is no functional change here. Mostly it's just breaking a longish function into smaller functions for readability. A couple of bits of repeated code are turned into a function. Some more comments.

lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -831,75 +831,86 @@ ImageryLayerCatalogItem.disableLayer = function(catalogItem, layer, globeOrMap)
...
+    const currentTimeIndex = intervals.indexOf(currentTime);
+    if (currentTimeIndex < 0) {
+        // No interval contains this time, so do not show imagery at this time.
+        ImageryLayerCatalogItem.disableLayer(catalogItem, catalogItem._imageryLayer);
+        clearCurrent();
+        return;
+    } else if (currentTimeIndex !== catalogItem._nextIntervalIndex) {
+        // We have a "next" layer, but it's not the right one, so discard it.
+        ImageryLayerCatalogItem.disableLayer(catalogItem, catalogItem._nextLayer);
+        // clearNext(); // seems redundant
```
stevage, line 891: Oh, yeah. So the two lines above are replaced by `clearNext()`. But it seems redundant, because `replaceCurrentWithNext()` is about to be called, which calles `clearNext()` anyway.

kring, line 891: I don't see anything else setting `_nextIntervalIndex` to -1.

kring: Oh nevermind, `replaceCurrentWithNext` should do it.

kring: In that case just delete this line, or change it to a comment saying it will happen below.

PR #3158, opened 2018-11-23
Drop and drop notification

chloeleichen:
Show a notification when drop and drop successfully if explorer panel is not already open: https://github.com/TerriaJS/terriajs/issues/3090

### DIFF
--- a/lib/ReactViewModels/ViewState.js
+++ b/lib/ReactViewModels/ViewState.js
@@ -84,7 +84,7 @@ export default class ViewState {
         this.explorerPanelAnimating = false;
 
         // recently uploaded files via drag and drop interface
-        this.recentlyUploadedFiles = [];
+        this.lastUploadedFiles = [];
 
         knockout.track(this, [
             'previewedItem',
@@ -105,7 +105,7 @@ export default class ViewState {
             'mobileMenuVisible',
             'panelVisible',
             'explorerPanelAnimating',
-            'recentlyUploadedFiles'
+            'lastUploadedFiles'
         ]);
 
         // Show errors to the user as notifications.
--- a/lib/ReactViewModels/ViewState.js
+++ b/lib/ReactViewModels/ViewState.js
@@ -163,10 +163,6 @@ export default class ViewState {
         this.activeTabCategory = 'my-data';
     }
 
-    setPanelToUserData(){
-        this.activeTabCategory = 'my-data';
-    }
-
     closeCatalog() {
         this.explorerPanelIsVisible = false;
     }
@@ -202,5 +198,4 @@ export default class ViewState {
     hideMapUi() {
         return this.getNextNotification() && this.getNextNotification().hideUi;
     }
-
 }
--- a/lib/ReactViewModels/ViewState.js
+++ b/lib/ReactViewModels/ViewState.js
@@ -83,6 +83,9 @@ export default class ViewState {
 
         this.explorerPanelAnimating = false;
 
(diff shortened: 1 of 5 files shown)


kring:
I'm trying this at http://ci.terria.io/issue3090/ and it's not working well for me:
* The upload/load data button doesn't work.
* There's no notification when I drop bike_racks.geojson on the map, it just appears silently on the map.
* After I drop bike_racks.geojson on the map and it appears, the "About This Data" button on the workbench does nothing.

chloeleichen:
@kring Fixed! 💃

lib/ReactViews/DragDropFile.jsx
```diff
@@ -27,13 +27,27 @@ const DragDropFile = createReactClass({
...
                 }
             });
 
         this.props.viewState.isDraggingDroppingFile = false;
     },
 
+    notifyUpload(addedCatalogItems) {
+        // if explorer panel is not open, we show a notification
+        // slide in right, wait 3 + 2 seconds (plus transition time), then slide out right
+        this.props.viewState.recentlyUploadedFiles = addedCatalogItems.map(item => item.name);
```
kring: Shouldn't the uploaded files be added to the list of recently uploaded files, instead of replacing it? Probably need to cancel the previous setTimeout, too, if any. Consider what happens if the user drags a file right after another.

lib/ReactViewModels/ViewState.js
```diff
@@ -159,6 +163,10 @@ export default class ViewState {
         this.activeTabCategory = 'my-data';
     }
 
+    setPanelToUserData(){
```
kring: I think it'd be ok to just call `openUserData` instead of adding this new function. When you know the explorer panel is already open, setting `explorerPanelIsVisible` to true again won't do any harm.

lib/ReactViews/DragDropFile.jsx
```diff
@@ -27,13 +27,27 @@ const DragDropFile = createReactClass({
...
+        setTimeout(
+            ()=> {
+                this.props.viewState.recentlyUploadedFiles = [];
+            },5000);
```
kring: Would it be hard to keep the notification open if the mouse cursor is over it? That'd be a nice touch I think.

lib/ReactViews/DragDropNotification.jsx
```diff
@@ -0,0 +1,32 @@
...
+      this.props.viewState.explorerPanelIsVisible = true;
+    },
+
+    render() {
+      const fileNames = this.props.viewState.recentlyUploadedFiles.join(', ');
+
+      return (
+        <button className={classNames(Styles.notification, {[Styles.isActive]: fileNames.length > 0})} onClick={this.handleClick}>
+            <div className={Styles.icon}><Icon glyph={Icon.GLYPHS.upload} /></div>
+            <div className={Styles.info}><span className={Styles.filename}>{'"'}{`${fileNames.length > 0 ? fileNames : "your files"}`}{'"'}</span> has been added to <span className={Styles.action}>My data</span></div>
```
kring: Should say `have been added` if there is more than one file.

chloeleichen: 👍

lib/ReactViews/DragDropFile.jsx
chloeleichen: Moved this timeout into `DragDropNotificaiton` to handle hover.

chloeleichen:
@kring Ready for another look! 👍

PR #3163, opened 2018-11-27
loading map gif

chloeleichen:
show loader gif rather than text when loading map, part of fix for https://github.com/TerriaJS/terriajs/issues/3015, the other part is in here: https://github.com/TerriaJS/nationalmap/pull/814


kring:
Can we do anything to move all (or at least more) of the functionality into TerriaJS instead of requiring it to be in each map? Right now the UX with an unmodified map is a bit worse than it used to be (e.g. http://ci.terria.io/loader/).

chloeleichen:
set up a basic loader gif that does not require app css.

kring:
@chloe here's an example of the webpack code splitting I was babbling about:
https://github.com/TerriaJS/terriajs/blob/master/lib/Models/GeoJsonCatalogItem.js#L173

There's no way to do this without significant changes to map (rather than TerriaJS code), though. Here's what I'm thinking...

* Change the map webpack config to make a new file, maybe `entry.js`, the entry point rather than `index.js`.
* `entry.js` does whatever immediate "show the loader" stuff is necessary and uses `require.ensure` on `index.js`.
* When the `require.ensure` callback is called, we can hide the loader.

The nice part about this is that map forkers don't need to change their index.html or index.js to get this feature. But it does mean that `entry.js` will need to do direct DOM manipulation and embed CSS or styles.

Do you think that could work?

chloeleichen:
in this case we still need to change the webpack config for all Terriamap instances, right?  the `require.ensure` looks promising!

chloeleichen:
This PR is going to be replaced by https://github.com/TerriaJS/TerriaMap/pull/291

kring:
Replaced by TerriaJS/TerriaMap#291.

PR #3164, opened 2018-11-27
rc-slider to rule them all

chloeleichen:
Fixes https://github.com/TerriaJS/terriajs/issues/3162

This PR will also fix https://github.com/TerriaJS/terriajs/issues/2781


chloeleichen:
Re the compiling warning: 
```
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass
```
it seems that it was a proposed implementation (to disallow import xxx.css into sass file), but it has been reverted in the latest node sass: https://github.com/sass/node-sass/issues/2362, and the warning is gone if I ran `npm install node-sass@beta` in the app directory. 

@kring ready for another look

PR #3170, opened 2018-11-29
added support for wcs styles

edisonguo:
When a user selects a region in TerriaJS to export data (aka. "clip-n-ship"), they would conceivably expect to obtain data for the bands used in the style of the layer they are currently viewing. For example, if a user has selected "False colour" from the style selection dropdown for a Landsat 8 layer, when using the export feature they should receive data for the "swir1", "nir" and "green" bands. If they then switched to the "True colour" style instead, the data should contain the "red", "green" and "blue" bands. In order to facilitate this behaviour, the WCS service provider would need to know both the currently selected WMS layer and style. Currently, when TerriaJS sends a WCS "GetCoverage" request it only includes the layer name (via the "coverage" parameter) and not the style. Note that a standard WCS request would explicitly specify what bands to include from an "AxisDescription" element (eg. "measurements" parameter), however in this scenario the user is attempting to export data corresponding to a WMS layer/style they are viewing and TerriaJS doesn't know what bands they correspond to as they are configured on the WMS service provider side.

### DIFF
--- a/lib/Models/callWebCoverageService.js
+++ b/lib/Models/callWebCoverageService.js
@@ -69,6 +69,12 @@ function launch(wmsCatalogItem, bbox) {
     if (defined(wmsCatalogItem.discreteTime)) {
         query.time = wmsCatalogItem.discreteTime.toISOString();
     }
+
+    const style = getStyle(wmsCatalogItem);
+    if(defined(style)) {
+        query.styles = style;
+    }
+
     var uri = new URI(wmsCatalogItem.linkedWcsUrl).query(query);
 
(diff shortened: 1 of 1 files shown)


edisonguo:
@kring I added an email address to my github profile and now the CLA appears to be signed.

kring:
Thanks @edisonguo!

kring:
@edisonguo could you please open a new pull request adding yourself to https://github.com/TerriaJS/terriajs/blob/master/CONTRIBUTORS.md?

edisonguo:
@kring Here is the contributors PR: https://github.com/TerriaJS/terriajs/pull/3197

PR #3173, opened 2018-11-30
external link icon for external link in menu

chloeleichen:
Fixes https://github.com/TerriaJS/terriajs/issues/2882


kring:
@philipgrimmett do you want to take a look at this one?
http://ci.terria.io/issue2882/

image

I think it's a bit strange to have the icon on the right when the rest are on the left, and also the new icon is a bit bigger and bolder than the existing ones. But maybe it's fine.

philipgrimmett:
@kring yep, agreed, strange. Probably a careless decision on my part (sorry!)

It would be great if we could implement the following:
screen shot 2018-12-10 at 2 29 33 pm

**Icon assets:**
[terria-menu-icons.zip](https://github.com/TerriaJS/terriajs/files/2661424/terria-menu-icons.zip)

Buttons: The print options are basically hidden at the moment, so I'd suggest with change the "Share" button text to "Share / Print"

Icons: The globe icon has never rendered well at small sizes, so I've switched it out for a control panel. Re. related maps, it feels a bit more balance (to me) if all the buttons have prepended icons (now represented as a gallery).

chloeleichen:
Need https://github.com/TerriaJS/TerriaMap/pull/282 to be merged for related map icon,
and @philipgrimmett said it was perfect. 
@kring ready for another look

PR #3178, opened 2018-12-18
Bring TerriaMap's regionmapping.json into TerriaJS

### LINKED ISSUE #1227: Add regionmapping.json to TerriaJS
Might as well make it a Terria functionality, not just an NM one.

stevage:
and use it by default.
Fixes #1227.


kring:
There are three test failures in this branch, possibly cause by the tests expecting something weird from regionMapping.json.

#3198 would ideally also be addressed before we merge this.

stevage:
The latest test failure seems to just be a Sauce thing? I can't really tell. To the best of my knowledge this is safe to merge now.

kring:
All browsers are seeing the same three test failures. Looks like a legit problem to me.

PR #3189, opened 2019-01-07
draggable feature info

chloeleichen:
Demo: http://ci.terria.io/feature-info-drag-constrained/#share=s-rHm2fd8qr5xDtoBUkAixG8zFpwI


philipgrimmett:
@chloeleichen Looks good! Comments:

- Feature info collapse functionality needs to be more discoverable. Should we fix that here, or should I create a new issue (if there isn't one already)?
- We might need a drag exclusion zone to avoid UI overlap (see concept image)

drag exclusion zone

kring:
I find it very hard to click on the expand/collapse button, because if the mouse moves at all between up and down it goes into drag mode instead of counting as a click. This is true of the existing feature info section collapse buttons, too.

kring:
It's possible to get the settings dropdown stuck underneath the chart:

image

Just click the Chart to bring it to the front and then click the Map button.

chloeleichen:
Fixed above issues

lib/ReactViews/DragWrapper.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+      const onend = ()=> {
+          setTimeout(()=> {
+              this.setState({
+                  isDragging: 0
+              });
+          }, 2000);
+      };
+      const node = this.node;
+      interact(node)
```
kring, line 46: Is it necessary to `unset` this when the component unmounts?
http://interactjs.io/docs/faq/#removedestroyrelease-an-interactable

lib/ReactViews/DragWrapper.jsx
```diff
@@ -0,0 +1,64 @@
...
+              restriction: 'parent',
+              endOnly: true,
+              elementRect: { left: 0, right: 1, top: 0, bottom: 1 }
+          },
+          onend: onend
+      });
+  }
+
+  render() {
+      return <div data-is-dragging={this.state.isDragging} ref={node => this.node = node} >{this.props.children}</div>;
```
kring, line 75: What does `data-is-dragging` do? I can't find any mention of it in the interact.js docs or elsewhere.

lib/ReactViews/FeatureInfo/FeatureInfoPanel.jsx
```diff
@@ -68,6 +80,11 @@ const FeatureInfoPanel = createReactClass({
         }
     },
 
+    checkDragging() {
+        console.log(this.ref.current.node.getAttribute('data-is-dragging'));
```
kring: ohh I see what `data-is-dragging` is used for now. 
Two minor things: remove the `console.log`, and consider renaming this function to `isDragging` cause that makes it clearer what the true/false means (not that it's hard to figure out).

PR #3202, opened 2019-01-25
fixed single timestamp issue

### LINKED ISSUE #2956: TerriaJS can't handle timeseries layers (at least WMS) with a single timestamp
When trying to view a timeseries WMS with only a single timestamp, TerriaJS does not expose the date picker, and does not send a `time` parameter with the WMS requests. This causes the service to return a HTTP 400 `Bad request` as a response.

Once a second timestamp was added to the layer then everything works fine.

edisonguo:
@kring This fixes https://github.com/TerriaJS/terriajs/issues/2956
If the time parameter is not set in the WMS request, the server side will have to support default timestamp to avoid 400 bad request.

Also from UX perspective, I think it makes sense to display the single timestamp. If no timestamp shows up in the catalog item, user might feel something went wrong.


edisonguo:
@kring I updated the comments. Here is the procedure you can test this fix:
1) Add this endpoint: http://gsky.nci.org.au/ows
2) Load this layer:  `DEA High Tide Composite 25m v2.0 true colour`. This layer has a single timestamp.

PR #3219, opened 2019-02-15
Fix double-quotes in FeatureInfoTemplate example

andrewdhicks:
The double quotes around the date format string parameter need to be escaped when passing to an inline function

See: https://github.com/TerriaJS/terriajs/issues/3218


kring:
Thanks @andrewdhicks! Mind also adding yourself to `CONTRIBUTORS.md`?

andrewdhicks:
Sure, thanks! Done!

PR #3220, opened 2019-02-19
`npm audit fix` and other dependency cleanup

### LINKED ISSUE #3216: Change terriajs-server version to require post-security release version
We should upgrade the required terriajs-server version to be `^2.9.1` to try to make sure everyone upgrades.

kring:
* Removed Urthecast support. It was based on an npm package that hasn't been updated in 3 years and that was reported to have security vulnerabilities.
* Fixes #3216.
* TerriaMap now reports 0 vulnerabilities when installed (previously was hundreds, even though I don't believe any were actually exploitable).
* Node.js 8.0 or later is now required.
* Upgraded to gulp 4.
* Upgraded to webpack 4.
* Adds a `deployment` status indicator (click "Show all checks" at the bottom of this PR page) that links to the build log while deployment is progress, and to the deployed instance afterward.
* Stopped using Enzyme in the one place we used it in the tests. IE11 is not a fan.

There are definitely breaking changes here, so this must be paired with the `audit-fix` branch of TerriaMap: https://github.com/TerriaJS/TerriaMap/tree/audit-fix


CHANGES.md
```diff
@@ -1,6 +1,13 @@
 Change Log
 ==========
 
+### Next Release
+
+* Breaking Changes:
+  * TerriaJS now requires Node.js v8.0 or later.
+  * TerriaJS now requires Webpack v4.0 or later.
+  * Removed `UrthecastCatalogItem`, `UrthecastCatalogGroup`, and `registerUrthcastCatalogItems`. The Urthecast functionality was dependent on an npm package that hadn't been updated in three years and had potential security vulnerabilities.
+
```
chloeleichen, line 11: Recommend adding:
`TerriaJS build now requires gulp.js v8.0 or later.`

PR #3227, opened 2019-02-25
DO NOT MERGE Resize f info

### LINKED ISSUE #3222: resize featureInfo
Experiment with making feature info panel resizable, as per discussion with Phill,

chloeleichen:
Fix #3222 
TODO:  browser testing


kring:
This is pretty cool, but it loses the automatic dynamic resizing of the feature info that we have now. Like if you resize for a big feature info and then click a small one, the window is unnecessarily huge. And similarly if you resize for a small one and then click a big one, the panel looks all squished and terrible. I'm not totally sure what to do about that, maybe @philipgrimmett has some ideas. You can fix an important half of the problem by making the resize control the max size instead of the actual size. At least that way a small feature info will still be small even if you previously resized for a big one.

philipgrimmett:
@chloeleichen is there a ci link I can look at?

kring:
CI links are found at the bottom of every PR, click details next to deployment.

PR #3228, opened 2019-02-26
Story beautify

chloeleichen:
ready to be reviewed!

- [x] Browser testing (firefox svg icon sizing issue to be confirmed)
- [x] mobile testing
- [x] Create Story
- [x] Edit Story
- [x] Remove story
- [x] View Story
- [x] Security (HTML Injection)
- [x] Too many stories 
- [x] Also https://github.com/TerriaJS/terriajs/issues/3013

ISSUE:

a modified/recaptured story doesn't modify existing url, require new share url.


philipgrimmett:
## Blue story prompt
- [x] Centre to map (currently centred to viewport)
- [x] Reword title to "**This** map contains a story"
- [x] Capitalise W + add missing question mark to "Would you like to view it now?"
- [x] Set action order to **Maybe later | Yes**

## Story component
- [x] Animate in/out from bottom of screen (match data catalogue)
- [x] Retain square aspect ratio for 'zoom to/re-centre' and 'close story' buttons
- [x] Apply 4px corner radius to component (match data catalogue)
- [x] Restrict clickable area to total width of scene dots (currently extends to end of scene title)
- [x] Make story images non-draggable to avoid drag/drop issue (if possible?)
- [x] Apply min. 16px padding to story text
- [x] Allow error/other dialogues that require user action to appear above story component
- [x] Allow feature info to appear above story component
- [x] Update previous/next button title text to "Go to previous/next scene" (currently "go to previous/next story")
- [x] Bug: story component appears to invisibly block mouse interaction with Add data button/workbench on shorter screens
- [x] To discuss: expand previous/next button to share total width of story component if body text isn't provided
- [x] To discuss: set story component baseline min. 16px above timeline/charts (when active)
- [x] To discuss: temporarily animate out story component when data catalogue is launched (simultaneous)
- [x] To discuss: animation options for scene transitions
- [x] To discuss: add hover states
- [x] To discuss: add action confirmation hints (eg. recapture)
- [x] To discuss: Show hint element below 'Story' button (top right on map) with text "You can access existing stories here, or create your own. Action:Got it, thanks!" (trigger when 1. blue story prompt is dismissed, and 2. Story is exited for the first time)

## Story editor panel
- [x] Animate in/out from right of screen (match data workbench)

## Text editor modal
- [x] Animate in/out from bottom of screen (match data catalogue)

Ran out of time, more to come!

chloeleichen:
after discussion:

- [x] 1. strong panel button hover state
- [x] 2. can only only one dropdown menu each time in story builder
~- [ ] 3. chart should be below story panel~
- [x] 4. no need for navigation dots if there are less than 2 story
- [x] 5. chart should push up story 
- [x] 6. when explorer window open , story should be hidden
- [x] 7 Make feature info window opaque when explorer window open

meh9:
Some issues:

- [x] 1. The Right button actually moves left.
- [x] 2. The hover tooltip of the individual scenes is `go to story ${i}`:
Screen Shot 2019-04-01 at 14 32 18

- [x] 3. You can't add layers to a scene that has been captured. This makes it impossible to edit the layers and Recapture a scene, as well as adding useful layers when viewing a scene.

meh9:
When clicking on Next, it briefly removes all layers, even if the same layers are enabled on the next view. This has a distracting "flickering" effect.

It would be great if it worked like the following, and if it could avoid doing a "Remove All" as part of the workflow:
1. When clicking Next, enable new layers for next story point, but do not remove any datasets yet.
2. ZoomTo new view.
3. Remove layers not present in new story point.

See this example for "flickering": http://ci.terria.io/story-beautify/#share=s-lniO4ixrHX54Q48lQhmsIwp7ZBz

AnaBelgun:
@chloeleichen @philipgrimmett 
below some suggestions from users:

- [x] Enable moving/reordering story scenes in the scenes stack otherwise it is a lot of work to rearrange scenes once they are created

- [x] Would it be possible to – in addition to View, Edit, Recapture, Delete controls – also add Move Up and Move Down?

 ### To park this until further info available :

> ### It would be useful to have a time-stamp reference on when the story is created and for each scene, and who the author/contact is. Data sources may change over time and story may lose context as what is in description may no longer correspond to what user sees on screen. Similarly, author may want to provide his/her contact details.

meh9:
Some bugs from user-feedback for this story: http://ci.terria.io/story-beautify/#share=s-pYXy1G7FOudOkzmkjuqOgBlYNOD

> I found a couple of more bugs in the story map link that you have sent:
>  

- [x] > ·         If I ran the story map and come to “Bird migration” scene, I can no longer progress to the story as window controls extend beyond screen and the window is immovable.
- [x] > ·         Then if I go to “Story” and select next scene and “View” to progress this way to next scene, it does load next scene but the description resets to description of the first scene.

meh9:
In Mobile view:

- [x] there is no story-point visualisation so we don't know where in the story we are
- [x] there are no left/right nav buttons (possibly intentional due to screen space?)
- [x] to navigate the story you can only swipe left, which actually moves you backwards through the story

AnaBelgun:
@chloeleichen @philipgrimmett one more feedback:

- [x] clear option to format the text and add the image - currently the user needs to highlight the text to get the formatting bar and when you get there, for images is not clear that it only deals with images on the web / urls + select text + click image

AnaBelgun:
there are some issues submitted: https://github.com/TerriaJS/terriajs/issues/3283 and https://github.com/TerriaJS/terriajs/issues/3284

For the team to decide if they go into v2 or need to be fixed in this version

kring:
A few problems I noticed working on the NSW Digital Twin demo story:
* [x] Terrain split direction is not preserved by stories or share links. Note that only the `digital-twin` branch of TerriaJS supports splitting terrain currently. Probably just need to do #3251 for the terrain catalog item too. So outside the scope of this PR, but I wanted to note it somewhere. 


chloe add:
> and https://github.com/TerriaJS/terriajs/pull/3333


* [x] Animation play/pause state and animation speed are not preserved in stories.
* [x] Editing the text of a story page also recaptures the map state. (pretty sure that's what I saw)

kring:
By the way, just want to say publicly that this feature is awesome. Great work @steve9164 @philipgrimmett @chloeleichen! It was a big hit at Locate19.

chloeleichen:
Fixed https://github.com/TerriaJS/terriajs/issues/3284

chloeleichen:
Fixes #3283

kring:
> IE11 need polyfill findIndex

polyfill.io is _supposed_ to be handling this for us, but I noticed the other day in the mobx branch that it wasn't for `find`. I don't know if it never did, and I just misunderstood, or if something has changed.

chloeleichen:
> polyfill.io is _supposed_ to be handling this for us, but I noticed the other day in the mobx branch that it wasn't for `find`. I don't know if it never did, and I just misunderstood, or if something has changed.

Yeah, so in terria there's already function `findIndex` that polyfill it for us, the actual `polyfill.in.js` seems broken when it tries to do this (looks like related to this: https://github.com/Financial-Times/polyfill-library/issues/115)

kring:
On http://ci.terria.io/story-beautify/:

1. Click Story
2. Click Capture Scene, give it a title, click Save.
3. Hit the hamburger next to your new scene and click Recapture

A duplicate copy of your scene is created, rather than updating the existing one.

kring:
1. Enable the splitter, click Capture Scene and save it.
2. Disable the splitter, click Capture Scene and save it.
3. Click Play Story.

The splitter is correctly enabled in the first scene, but it is not disabled in the second one.

kring:
- [ ] It seems like the basemap is being reinitialized the first time a story is played. The globe blinks to blue and the network tab shows a new Bing Maps Imagery/Metadata request. The extra Imagery/Metadata request is especially bad because our Bing Maps usage is measured by how many times we invoke this service.

kring:
1. In 3D terrain mode, zoom in to Perth and capture a scene.
2. Switch to 2D. Zoom in to Brisbane. Capture a scene.
3. Play the Story

The first scene will be 3D as expected, but it will be looking at Brisbane instead of Perth.

lib/ReactViews/Story/StoryPanel.jsx
```diff
@@ -0,0 +1,179 @@
...
+import ObserveModelMixin from '../ObserveModelMixin';
+import parseCustomHtmlToReact from '../Custom/parseCustomHtmlToReact';
+import { Small, Medium } from '../Generic/Responsive';
+import Icon from "../Icon.jsx";
+import { Swipeable } from 'react-swipeable';
+import Styles from './story-panel.scss';
+
+// given a nowviewing item, build its path in the data catalog
+export function buildPath(item) {
+      if(!item.parent) {
```
kring: Use the catalog item's `uniqueId` property, which is basically the same thing but more resistant to catalog items being renamed and moved.

lib/ViewModels/updateApplicationOnMessageFromParentWindow.js
```diff
@@ -26,12 +26,12 @@ var updateApplicationOnMessageFromParentWindow = function(terria, window) {
...
         }
 
-        // Ignore react devtools
+       // Ignore react devtools
         if((/^react-devtools/gi).test(event.data.source)) {
             return;
         }
 
-        terria.updateFromStartData(event.data).otherwise(function(e) {
+        terria.updateFromStartData(event.data) && terria.updateFromStartData(event.data).otherwise(function(e) {
```
kring: This is going to do it twice 😱.

package.json
```diff
@@ -77,6 +79,7 @@
     "svg-sprite-loader": "^4.1.3",
     "terriajs-cesium": "1.55.1",
     "terriajs-html2canvas": "1.0.0-alpha.12-terriajs-1",
+    "uniqid": "^5.0.3",
```
kring: Cesium has a function `createGuid` that could probably be used instead of this.

wwwroot/test/csv/lat_lon_name_url_col.csv
```diff
@@ -1,5 +1,5 @@
 lat,lon,name,url,col
 -30,128,Ganymede,build/TerriaJS/test/csv_nongeo/A,1
--23,136,Betelgeuse,build/TerriaJS/test/csv_nongeo/B,2
+-23,136,Bete,lgeuse,build/TerriaJS/test/csv_nongeo/B,2
```
kring: Accidentally committed?

kring (changes requested):

lib/Models/Terria.js
```diff
@@ -561,12 +567,14 @@ Terria.prototype.updateApplicationUrl = function(newUrl) {
...
+        const stories = initSources.filter(init => defined(init.stories))[0];
+        if(stories) that.stories = stories.stories;
         return loadInitSources(that, initSources);
     });
 };
 
-Terria.prototype.updateFromStartData = function(startData) {
-    var initSources = this.initSources.slice();
+Terria.prototype.updateFromStartData = function(startData, includePersistantInitSources = true) {
+    var initSources = includePersistantInitSources ? this.initSources.slice() : this.initSources.filter(init => !(init.initFragment && init.initFragment === 'terria'));
```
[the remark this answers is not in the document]
chloeleichen: What about we don't include any init fragments at all for story? 
`var initSources = includePersistentInitSources  ? this.initSources.slice() : []; `

lib/ReactViews/Story/StoryBuilder.jsx
```diff
@@ -0,0 +1,160 @@
...
+      }
+
+      this.setState({
+          editingMode: false
+      });
+  },
+
+    captureStory(story) {
+        story.shareData = JSON.parse(JSON.stringify(getShareData(this.props.terria, false)));
+        this.props.terria.stories = [...(this.props.terria.stories || []), story];
```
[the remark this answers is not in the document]
chloeleichen: ok `push` works but i'm confused here, since `stories` is tracked by knockout, i thought it needs to be replaced (rather than modified) for knockout to notice the change?

philipgrimmett:
@chloeleichen suggested updates:

- [x] While editing scene text, continue to display exisiting scenes in story editor in a disabled state (currently hidden)
- [x] Return 'centre scene' button to default state when scene is no longer centred
- [ ] ~Update icons to 'official' set [Terria icons - April 19.zip]~(https://github.com/TerriaJS/terriajs/files/3129685/Terria.icons.-.April.19.zip) moved to https://github.com/TerriaJS/terriajs/issues/3347

- [x] centre scene' icon, previous/next arrows, close/cross icon in story

- [ ] text editor icon update -> **V2**

- [x] Remove 'Move up/down' options from scene menus (now draggable)
- [x] Slide in/out Story panel (to match data workbench animation)
- [ ] If story is "playing": centre story box when data workbench is hidden
- [x] Bug: multiple scene menus can be opened at the same time (testing in Chrome 73, screenshot below)

- [x] Add scene group header to story panel (# of scenes, remove all): see https://invis.io/43RSPX2U5JD#/360676208_story_-_Panel_Open (late addition, sorry @chloeleichen!)
- [x] Display "you can create and view stories at any time by clicking here" flag below Story button ~1 second after: 1. user dismisses "this map contains a story" prompt OR 2. user closes/stops "playing" a story accessed via prompt "this map contains a story" . Display once only for either condition. Invision link: https://invis.io/43RSPX2U5JD#/360663465_tour_-_Story
- [x] Fade in "Introducing Data Stories" flag below Story button ~2 seconds after page load (we'll need to experiment with exact timing). Display once only. We'll probably need to discuss with the team how long after release this flag should be active, and how future 'new feature' flags are managed. Invision link: https://invis.io/43RSPX2U5JD#/360663466_intro_-_Data_Stories

Screen Shot 2019-04-30 at 11 40 25 am

kring:
If the Cesium render looped is stopped (i.e. you haven't moved the mouse over the map in the past couple seconds), the 3D view doesn't resize properly when the story is expanded/collapsed:
image
It needs a `notifyRepaintRequired`. Check out how the workbench expand/collapse does it.

kring:
I'm still seeing the globe flash blue the first time a story is played.

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
```diff
@@ -29,6 +29,7 @@ const DataCatalogItem = createReactClass({
     },
 
     onBtnClicked(event) {
+        this.props.viewState.storyShown = false;
```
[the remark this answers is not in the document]
chloeleichen: Yes the inconsistency is a bug. 

The reason to end story "play" mode when adding data is to make it clear when a story is being edited, when it's not. If user is playing a scene, then add a dataset, the layers they are looking at is no longer the story, yet the story is still "playing". 

Instead, we make it if user wants to add a layer to a story, they need to quit "play" mode, then they can add dataset either in normal mode or editing mode, and save in editing mode.

philipgrimmett:
@chloeleichen suggested updates:

- [x] Change "Stories [ # ]" in scene list header to "Scenes [ # ]"
- [x] Resize "Scenes [ # ]" and "Remove all" text to match (currently 0.8125rem / 0.7rem, 12px in Invision)
- [ ] ~Double check component colours (palette attached, should match data workbench)~
- [x] Apply min. 8px padding to scene menu text
- [x] Apply hover style to scene menu list items (10% brighter, see palette)
- [ ] ~While dragging/updating scene position, indent scene block by 8px~
- [x] Recapture scene: when recapture is clicked, briefly (~1.8?) display 'recapture-20' icon to the left of the menu (slideFade in/out)
- [x] Scene menu: make entire 'view' list item clickable (clicking in the right hand corner currently dismisses menu)

- [x] Resize icons to 24*24: Play, Capture
- [x] Set icon padding-left to 8px: Play, Capture
- [x] Update icon: close-20, link-20, image-24

Icons: 
[Terria icons - May 19.zip](https://github.com/TerriaJS/terriajs/files/3146450/Terria.icons.-.May.19.zip)

Palette:
story colours

kring:
The blue flash when starting a story the first time is fixed, but now it seems scenes can no longer change basemaps at all.

chloeleichen:
> The blue flash when starting a story the first time is fixed, but now it seems scenes can no longer change basemaps at all.

Fixed @kring

kring (changes requested):

lib/Models/Terria.js
```diff
@@ -669,20 +691,45 @@ Terria.prototype.addInitSource = function(initSource) {
...
+    that.showSplitter = false;
+  }
+
+  if (fromStory === true) {
+    viewerChangeListener = this.afterViewerChanged.addEventListener(function() {
+      if (that.viewerMode === ViewerMode.Leaflet) {
+        that.leaflet.zoomTo(that.initialView, 0.0);
+      } else {
+        that.cesium.zoomTo(that.initialView, 0.0);
+      }
```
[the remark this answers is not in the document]
kring: Also, this event handler, and the fact that it isn't removed, causes a bug:

1. Switch to 2D
2. Zoom in somewhere memorable
2. Capture a scene
3. Switch back to 3D
4. Click Play Story, viewer switches to 2D as expected.
5. Click X to close the story.
6. Switch to 3D
7. Zoom somewhere else, or hit the Reset view button.
8. Switch to 2D

The map will unexpectedly zoom back to your memorable location.

lib/Models/Terria.js
```diff
@@ -669,20 +691,45 @@ Terria.prototype.addInitSource = function(initSource) {
...
 
   if (defined(initSource.viewerMode) && !defined(this.userProperties.map)) {
     var desiredMode = (initSource.viewerMode || "").toLowerCase();
+
     if (desiredMode === "2d") {
+      if (fromStory && this.viewerMode === ViewerMode.Leaflet) {
+        return;
+      }
       this.viewerMode = ViewerMode.Leaflet;
+      if (defined(viewerChangeListener)) viewerChangeListener();
```
kring: I'm not sure why the `afterViewerChanged` listener is being removed here (that's what calling `viewChangeListener()` does, right?). It seems backwards. If the viewer is already the right one, you don't need that `afterViewerChanged` listener, so removing it (or never installing it in the first place) makes sense. But if the viewer is changing, you need the listener in order to set the view after the change. And then I think you'd want the handler to remove itself from the Event, to avoid the bug I described above.

kring (changes requested):

lib/ReactViews/Generic/Prompt.jsx
```diff
@@ -0,0 +1,61 @@
...
+  componentWillUnmount() {
+    clearTimeout(this.fadeInTimer);
+  }
+  componentDidMount() {
+    this.fadeIn();
+  }
+  dismissPrompt() {
+    this.props.dismissAction();
+  }
+  createMarkup() {
```
kring: Looks like this isn't used.

lib/Models/Terria.js
```diff
@@ -659,8 +668,21 @@ Terria.prototype.addInitSource = function(initSource) {
 
   // Extract the init source properties that require no deserialization.
   directInitSourceProperties.forEach(function(propertyName) {
-    if (defined(initSource[propertyName])) {
-      that[propertyName] = initSource[propertyName];
+    // a special case for basemap
+    if (propertyName === "baseMapName" && defined(initSource[propertyName])) {
```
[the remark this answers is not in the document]
chloeleichen: yes indeed :D

lib/Models/Terria.js
```diff
@@ -669,20 +691,45 @@ Terria.prototype.addInitSource = function(initSource) {
     runLater(function() {
       that.showSplitter = initSource.showSplitter;
     });
+  } else {
+    that.showSplitter = false;
```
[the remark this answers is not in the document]
chloeleichen: I added `showSplitter` to scene data, i think it's clearer that way. Thanks!

lib/Models/Terria.js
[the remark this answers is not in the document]
chloeleichen: Removed event handler after zoom inside the event handler

lib/Models/Terria.js
chloeleichen: I'm a bit confused. I'm not removing `viewerChangeListener` here?

kring: `viewerChangeListener` is assigned the return value of a Cesium `addEventListener` call. So `viewerChangeListener()` removes the listener. Probably not what you expected. :)

lib/Models/Terria.js
```diff
@@ -616,10 +624,10 @@ Terria.prototype.updateApplicationUrl = function(newUrl) {
   });
 };
 
-Terria.prototype.updateFromStartData = function(startData) {
-  var initSources = this.initSources.slice();
+Terria.prototype.updateFromStartData = function(startData, fromStory = false) {
```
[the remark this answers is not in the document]
chloeleichen: Yup updated!

kring:
A call for @philipgrimmett, but I think the dismiss behavior of the story-related popups is backwards.

When you visit a share link with a story and you say "Maybe later", you get this:
image

But if you dismiss it, you'll never see it again, even if you reload the page or visit an entirely different story.

On the other hand, every time you load the app without a story already active, you get this:
image

And even if you dismiss it, you'll get the same popup the next time you load the app. I think these should be flipped. Always show the first, only show the second once.

kring:
We also need to think about how the "Introducing Data Stories" popup interacts with the previewer use case used in e.g. Magda and CKAN. At a minimum, we shouldn't show it if the app is launched with `http://ci.terria.io/story-beautify/#hideWorkbench=1`. And maybe we should have a URL parameter to launch the story directly without asking?

lib/ReactViews/Generic/Prompt.jsx
kring: Not sure why this was resolved. I still don't see it used anywhere.

kring (changes requested):
Getting really close now!

philipgrimmett:
@kring @chloeleichen I should probably flesh out the event logic a bit:

- If there is >=1 scene in the story editor, show story prompt
- If story prompt is dismissed, show story button flag

- If there are no scenes in the story editor, do not show story prompt
- If there are no scenes in the story editor,  show "Introducing Data Stories" flag
- If user dismisses "Introducing Data Stories" flag, remember this and do not show again*

*We'll need to think of other ways to promote/prompt users to explore the story feature, and new Terria features in general. A fairly common pattern is to display a short notification in the bottom right-hand corner of the screen that links through to blog posts.

kring:
Sounds good to me @philipgrimmett. I think what you described is mostly what we have except that the story story prompt flag can be dismissed permanently and the "Introducing Data Stories" flag cannot.

philipgrimmett:
@kring yep, hidden stories and infinite introductions could get preeeetty annoying :)

philipgrimmett:
@chloeleichen I know I promised a freeze, but...

If a story author includes a link in their scene text, say "google.com" (without http://) that link will be broken. Are we able to check and prepend "http://" to links like the above and launch them in a new tab by default?

chloeleichen:
@philipgrimmett
@kring 

Yeah I made it that way because I worry that users are going to dismiss the introducing data stories without paying attention to it (therefore I display it everytime page load), meanwhile, the story button only shows when 

1. users very consciously refused to look at an existing story (after they followed a share URL with story, so they should be expecting stories) or,

2. When they closed a story after playing it. 

In both above cases users have had some level of interaction with story and do not need to be “recommended” again to check story out, therefore the all prompts are disabled after this stage. 

So the general idea is to keep showing prompt until user had certain level of interaction with story.

lib/Models/Terria.js
chloeleichen: oh yeah i missed that. Thanks!

kring:
> I worry that users are going to dismiss the introducing data stories without paying attention to it (therefore I display it everytime page load)

But even once they interact with a story, the next time they load a page without a story they'll be hit with the "Introducing Stories" button again. Maybe once the user opens the Story panel, we can disable both notifications? I don't want to bike shed this, but I personally find it annoying that I get introduced to stories every time I run a Terria app now. :)

lib/ReactViews/Custom/Chart/ChartPanel.jsx
```diff
@@ -50,6 +51,7 @@ const ChartPanel = createReactClass({
   render() {
     const chartableItems = this.props.terria.catalog.chartableItems;
     if (findIndex(chartableItems, ci => !ci.dontChartAlone) < 0) {
+      this.props.viewState.chartIsOpen = false;
```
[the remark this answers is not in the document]
chloeleichen: this file is committed by mistake

chloeleichen:
>  once the user opens the Story panel, we can disable both notifications?

Good idea, will do that!

philipgrimmett:
I think disabling both notifications once the user opens the story panel our best bet, but it still feels more like a stick than a carrot. Having a persistant notification system applied across multiple Terria instances to call out new features, datasets, versions, events etc. would help to capture a wider audience, but that's for another PR...

lib/ReactViewModels/ViewState.js
```diff
@@ -213,4 +244,16 @@ export default class ViewState {
   hideMapUi() {
     return this.getNextNotification() && this.getNextNotification().hideUi;
   }
+
+  toggleFeaturePrompt(feature, state) {
+    const featureIndexInPrompts = this.featurePrompts.indexOf(feature);
+    if (state === true) {
+      // add feature prompt
+      featureIndexInPrompts < 0 && this.featurePrompts.push(feature);
```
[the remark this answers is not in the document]
chloeleichen: Done!

lib/ReactViews/Custom/Chart/ChartPanel.jsx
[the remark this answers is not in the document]
chloeleichen: Fixed!

chloeleichen:
Added support for URL query: 
`#showStories=1` will automatically play story if there's any on first load
`#hideWorkbench=1` will automatically play story if there's any on first load

kring:
> #showStories=1 will automatically play story if there's any on first load

Great! Maybe just rename it to `playStory=1`? (consistent with the naming in the UI)

lib/ReactViews/Custom/Chart/ChartPanel.jsx
```diff
@@ -30,15 +29,7 @@ const ChartPanel = createReactClass({
...
-    const chartableItems = this.props.terria.catalog.chartableItems;
-    for (let i = chartableItems.length - 1; i >= 0; i--) {
-      const item = chartableItems[i];
-      if (item.isEnabled && defined(item.tableStructure)) {
-        item.tableStructure.columns
-          .filter(column => column.isActive)
-          .forEach(column => column.toggleActive());
-      }
-    }
+    this.props.viewState.chartIsOpen(false);
```
kring: Took me ages to figure out why `chartIsOpen` is a function instead of a property. It's defined in `ViewState` using `defineProperty`, right? And usually that would create a property, so you'd write code like `this.props.viewState.chartIsOpen = false`. I _think_ the reason you're not getting a property as expected is because `chartIsOpen` is being passed to `knockout.track` prior to the call to `knockout.defineProperty`. You shouldn't do both, and apparently if you do Knockout does some kinda weird stuff. TL;DR: remove `chartIsOpen` from `knockout.track` and change all the places you're using `chartIsOpen` to treat it like a property instead of a function.

chloeleichen: Fixed

kring (approved):

PR #3234, opened 2019-03-01
Model Layer version 3, now with TypeScript, MobX

kring:
This is a work in progress, not at all ready to be merged or even used. I'm just opening it now for a place to jot down some notes.


lib/Models/Cesium.ts
```diff
@@ -0,0 +1,1221 @@
...
+const createImageryLayer: (ip: Cesium.ImageryProvider) => Cesium.ImageryLayer = createTransformer((ip: Cesium.ImageryProvider) => {
+    console.log('Creating a new ImageryLayer');
+    return new ImageryLayer(ip);
+});
+
+function makeImageryLayerFromParts(parts: ImageryParts): Cesium.ImageryLayer {
+    const layer = createImageryLayer(parts.imageryProvider);
+
+    // This should only be called on ImageryParts that are observable. I think we can assume
+    autorun(() => {
```
kring: I don't think this autorun is needed; we can just assign the properties once here. If the alpha changes, this function will automatically be invoked again and update the Cesium layer's `alpha` property.

kring: If I'm right, that's a really good thing because currently this autorun is a leak.

tephenavies: That'll run the entire map items <-> Cesium ImageryLayerCollection every time you change opacity

kring: As discussed on Slack, yeah, that mapping re-runs every time, but it's not as expensive as it sounds, and it runs every time even with this autorun in place. Might be room for improvement, but I think it's pretty much fine as is.

tephenavies:
Closing as the comments here are no longer relevant, the `mobx` branch is old and we will open a new PR from `next` to `master` when we are ready.

PR #3241, opened 2019-03-04
Replaces `getUniqueValues` with `lodash.uniq`

### LINKED ISSUE #3231: Replace `getUniqueValues` with `lodash.uniq`
We're already using `lodash.uniq` in the `DateTimePicker`. We should be able to use it instead of our own `getUniqueValues` everywhere else, too.

na9da:
Fixes #3231


lib/Map/TableColumn.js
```diff
@@ -8,9 +8,9 @@ var destroyObject = require('terriajs-cesium/Source/Core/destroyObject');
 var DeveloperError = require('terriajs-cesium/Source/Core/DeveloperError');
 var JulianDate = require('terriajs-cesium/Source/Core/JulianDate');
 var knockout = require('terriajs-cesium/Source/ThirdParty/knockout');
+var uniq = require('lodash').uniq;
```
kring: I think this should be `var uniq = require('lodash.uniq');` cause otherwise webpack will build all of lodash into our app (no tree shaking here, unfortunately).

kring:
Just that one comment, looks great otherwise!

kring:
Thanks @na9da!

PR #3242, opened 2019-03-05
Repaint when toggling between viewer modes

### LINKED ISSUE #2952: Toggling between 3D Terrain and 3D Smooth does not notifyRepaintRequired
1. Switch to 3D Smooth.
2. Keep the mouse inside the Map panel, but don't move it for a few seconds.
3. Click 3D Terrain.

The UI will update to show that you've selected 3D Terrain, but the globe will not redraw until you move the mouse over the globe. The problem is a missing call to `notifyRepaintRequired`.

na9da:
Fixes #2952

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### Next Release
 
 * Replace `getUniqueValues` with `lodash.uniq`
+* Call `notifyRepaintRequired` when toggling between viewer modes.
 
 ### v7.0.2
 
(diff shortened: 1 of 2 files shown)


kring:
@na9da please add a mention of this fix to CHANGES.md.

kring:
Thanks @na9da! I hope you don't mind I modified your CHANGES.md. The idea with that file is to describe to a user (meaning a developer integrating TerriaJS in their application, or maybe even an end-user of a TerriaJS app when appropriate) what changed in this release. We should avoid assuming the reader knows much about the internal implementation details of TerriaJS, and instead focus on what positive effect they'll see.

PR #3247, opened 2019-03-12
Make sure getlegendgraphic url is cleared for each layer

sppigot:
GetLegendGraphic from catalog was being carried over into layers for which it was not applicable.


kring:
Thanks @sppigot, good catch!

PR #3249, opened 2019-03-13
Show only features from the selected side of the splitter

### LINKED ISSUE #2791: Feature Info ignores splitter
If you have a layer that is only visible on one side of the splitter, but that covers the region that is selected which is on the other side of the splitter it will appear in the feature info:
image

http://nationalmap.gov.au/#share=s-kNqakV9T5zNbR4KrUjhqsA8MINB

Steps to reproduce:
1) Load nationalmap.
2) Turn on two layers that cover the whole continent (for instance gravity anonoly and land cover).
3) Turn on the splitter.
4) Set one layer to be visible on the left and one layer to be visible on the right.
5) Select a point on the contnent on the right of the splitter.
6) Observe that both layers appear in the feature info.

This bug is present in both 2D and 3D modes.

na9da:
Fixes #2791

### DIFF
--- a/lib/Models/Cesium.js
+++ b/lib/Models/Cesium.js
@@ -791,18 +791,6 @@ Cesium.prototype.pickFromScreenPosition = function(screenPosition) {
     var pickRasterPromise = this.scene.imageryLayers.pickImageryLayerFeatures(pickRay, this.scene);
 
     var result = this._buildPickedFeatures(providerCoords, pickPosition, vectorFeatures, [pickRasterPromise], undefined, pickPositionCartographic.height);
-
-    if (this.terria.showSplitter) {
-        // Select only features from the same side or both sides of the splitter
-        var side = this.terria.getSplitterSideForScreenPosition(screenPosition);
-        result.allFeaturesAvailablePromise = result.allFeaturesAvailablePromise.then(function() {
-            result.features = result.features.filter(function(feature) {
-                return ( feature.imageryLayer.splitDirection === ImagerySplitDirection.NONE ||
-                         feature.imageryLayer.splitDirection === side );
-            });
-        });
-    }
-  
     var mapInteractionModeStack = this.terria.mapInteractionModeStack;
     if (defined(mapInteractionModeStack) && mapInteractionModeStack.length > 0) {
         mapInteractionModeStack[mapInteractionModeStack.length - 1].pickedFeatures = result;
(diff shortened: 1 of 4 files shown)


lib/Models/Cesium.js
```diff
@@ -792,6 +792,17 @@ Cesium.prototype.pickFromScreenPosition = function(screenPosition) {
 
     var result = this._buildPickedFeatures(providerCoords, pickPosition, vectorFeatures, [pickRasterPromise], undefined, pickPositionCartographic.height);
 
+    if (this.terria.showSplitter) {
+        // Select only features from the same side or both sides of the splitter
+        var side = this.terria.getSplitterSideForScreenPosition(screenPosition);
+        result.allFeaturesAvailablePromise = result.allFeaturesAvailablePromise.then(function() {
+            result.features = result.features.filter(function(feature) {
+                return ( feature.imageryLayer.splitDirection === ImagerySplitDirection.NONE ||
```
kring: Not all features have an `imageryLayer` property. If you add the "Bird Migration" layer in TerriaMap, and click on one of the yellow dots, you'll get an exception here.

Because the splitter only applies to imagery layers, I think it might make sense to filter right after the `pickFeatures` call in the `pickFromLocation` function, rather than filtering all results here. Unless you've considered that and it won't work for some reason that I haven't thought of.

kring:
Looks good @na9da! Just the one comment.

lib/Models/Leaflet.js
```diff
@@ -800,6 +804,22 @@ function pickFeatures(leaflet, latlng, tileCoordinates, existingFeatures) {
                 return leaflet._createFeatureFromImageryLayerFeature(feature);
             }));
         }, pickedFeatures.features);
+
+        if (leaflet.terria.showSplitter) {
+            // Select only features from the same side or both sides of the splitter
+            var screenPosition = leaflet.computePositionOnScreen(pickedFeatures.pickPosition);
+            var pickedSide = leaflet.terria.getSplitterSideForScreenPosition(screenPosition);
+
+            pickedFeatures.features = pickedFeatures.features.filter(function(feature) {
```
kring: Sorry Nanda, I'm going to nitpick this a bit more. `pickedFeatures.features` is an observable property (you can tell because it is `knockout.track`'d in `PickedFeatures.js`), which means when it is _set_ on line 795 above, other parts of Terria will potentially be listening and will react to the change. So those other parts of Terria will briefly see all the features, and then a moment later all the ones on the wrong side of the splitter will disappear. It would be better to set it just once with the filtered list of features. And I think in this case that will actually simplify the code a bit, as you can do the filtering in the `filteredResults.reduce` (if `result.imageryLayer` is on the wrong side, return `allFeatures` instead of `allFeatures.concat(...)`). A similar approach should work in the Cesium version as well.

kring:
Awesome thanks @na9da!

PR #3252, opened 2019-03-19
Show axes for unitless quantities, other charting fixes

kring:
This fixes the issue Mats described here:
https://github.com/TerriaJS/terriajs/pull/3151#issuecomment-472380328

Remaining problem: if you drag https://droughtreliefpublic.blob.core.windows.net/nationalmapdatapublic/Farm-Businesses/FMD%20industry.csv onto the map and click on Victoria, and then Expand, the 9-digit numbers on the Y-axis are cut off because they're too long. This was true previously, too, but it wasn't as bad because the Y-axis area was bigger. @chloeleichen is there a way to dynamically size the Y-axis area based on the length of the labels? That would be ideal. If not, we should at least make the area as big as it was before so we don't have a regression.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Change Log
 * Highlight polygon and polyline features.
 * Replace `getUniqueValues` with `lodash.uniq`
 * Fixed a bug where the 3D globe would not immediately refresh when toggling between the "Terrain" and "Smooth" viewer modes.
+* Fixed a bug that could cause the chart panel at the bottom to flicker on and off rapidly when there is an error loading chart data.
 
 ### v7.0.2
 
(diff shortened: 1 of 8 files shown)


chloeleichen:
Something not quite relevant, when there's no data to download, the expand/download button in feature info panel are still active, and when click download, the app goes to 404. If there's no chart, those button should be hidden/disabled?

Screen Shot 2019-03-22 at 1 54 07 pm

kring:
> If there's no chart, those button should be hidden/disabled?

The chart data URLs can be separate from the download URLs, so we can't necessarily tell whether the download will work based on the presence of a chart. So I just changed the `a href` to use `target='_blank'` so that the app won't be navigated away from the page when there's a 404.

PR #3258, opened 2019-03-22
Add GeoJsonCatalogItem to the new model layer

### LINKED ISSUE #3255: Model architecture: Flesh out GeoJsonCatalogItem

na9da:
Fixes #3255


lib/Core/TerriaError.d.ts
```diff
@@ -0,0 +1,10 @@
+export = TerriaError;
+
+declare class TerriaError {
```
kring: Would it be reasonable to move `TerriaError` over to TypeScript instead of writing a separate declaration file for it?

lib/Models/Cesium.ts
```diff
@@ -78,13 +79,31 @@ export default class Cesium implements GlobeOrMap {
             );
             // TODO: Look up the type in a map and call the associated function.
             //       That way the supported types of map items is extensible.
+
+            const allDataSources = allMapItems.filter(isDataSource);
+
+            // Remove deleted data sources
+            let dataSources = this.terria.dataSources;
```
kring: Probably a separate issue from this PR, but I think we want to move `dataSources` off `Terria` in this new architecture. It should just be owned by Cesium and Leaflet as necessary.

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,195 @@
...
+        return [this.dataSource];
+    }
+
+    loadMetadata(): Promise<void> {
+        return Promise.resolve();
+    }
+
+    loadData(): Promise<void> {
+        let that = this;
+        this.dataSource = new GeoJsonDataSource(this.name);
```
kring: We (eventually) need to guard against repeated invocations of `loadData` (i.e. it needs to be harmless to call it extra times). I just looked at WMSCatalogItem and it seems to have the same problem, though, so don't worry about it for now.

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,195 @@
...
+                    promise = readJson(data);
+                } else if (data instanceof String || typeof data === "string") {
+                    try {
+                        promise = JSON.parse(data.toString());
+                    } catch (e) {
+                        throw new TerriaError({
+                            sender: that,
+                            title: "Error loading GeoJSON",
+                            message:
+                                "\
```
kring: Let's move away from using `\` for multi-line strings. It's easy to accidentally put extra spaces at the start of each line, which our markdown formatter will then interpret as a `<pre>` block. Just concatenating each line with + is better I think.

lib/Models/loadGeoJson.js
```diff
@@ -0,0 +1,470 @@
...
+        name = propertyName;
+        geoJson = geoJson[propertyName];
+
+        // If we don't already have a name, or our name is just derived from our URL, update the name.
+        if (
+            !defined(geoJsonItem.name) ||
+            geoJsonItem.name.length === 0 ||
+            nameIsDerivedFromUrl(geoJsonItem.name, geoJsonItem.url)
+        ) {
+            geoJsonItem.name = name;
```
kring: You can't directly set traits on a model in the new architecture. The preferred way is to create a loadable stratum for any traits that can be derived from the data or metadata.

lib/Core/TerriaError.d.ts
na9da: Yes. Can do that as another PR.

kring:
Looks good @na9da. Regarding `loadGeoJson`, I do think it would be better, if at all possible, to move this code into TypeScript. TS would have flagged that catalogItem.name isn't settable, for instance. I don't mind doing it in phases, though.

lib/Core/TerriaError.d.ts
na9da: Removed the ambient typing for now.

lib/Models/Cesium.ts
na9da: `.dataSources` is now part of Cesium. Same needs to be done for Leaflet in a later PR.

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,674 @@
...
+                    break; // no need to count past 2 properties.
+                }
+            }
+        }
+
+        var name;
+        if (numProperties === 1 && isDefined(propertyName)) {
+            name = propertyName;
+            geoJson = geoJson[propertyName];
+            this.name = name;
```
kring: Wherever possible, we should think in terms of "this property is a function of this one/these" instead of "when that happens, this happens." So I'd try something like this:

* `LoadGeoJsonStratum` has a private, observable property called `_loadedGeoJson`, which is asynchronously set by `loadGeoJson`. This is the cause/effect bit, but we make it as small as possible.
* `name` is a computed property that reads `_loadedGeoJson` and uses the logic in this function to either return undefined (if there is more than one property), or the GeoJSON's outer property name (if there is exactly one outer property).
* `geoJsonData` is a computed property that returns `_loadedGeoJson` `geoJson[propertyName]` if there is exactly one outer property in `_loadedGeoJson`, and `_loadedGeoJson` itself otherwise.

kring: The common logic in the last two bullets can even be moved to a separate private computed. Values of computed observables are cached and only recomputed when something they depend on changes (not that that should make much difference here).

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,674 @@
...
+                    }
+                });
+        }
+    }
+
+    setTraitsFromData(geoJson: any): Promise<void> {
+        // If this GeoJSON data is an object literal with a single property, treat that
+        // property as the name of the data source, and the property's value as the
+        // actual GeoJSON.
+        var numProperties = 0;
```
kring: Stick to `const` where possible, `let` where necessary.

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,674 @@
...
+    }
+}
+
+export default class GeoJsonCatalogItem
+    extends LoadGeoJsonMixin(
+        UrlMixin(CatalogMemberMixin(Model(GeoJsonCatalogItemTraits)))
+    )
+    implements Mappable {
+    @observable
+    show: boolean = true;
```
kring: This should be a trait instead of defined here.

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,674 @@
...
+
+    @observable
+    dataSource?: GeoJsonDataSource;
```
kring: I think this can be private.

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,674 @@
...
+        return Promise.resolve();
+    }
+
+    loadData(): Promise<void> {
+        const loadGeoJsonStratum = <LoadGeoJsonStratum>(
+            this.strata.get(LoadGeoJsonMixin.loadGeoJsonStratumName)
+        );
+        return loadGeoJsonStratum
+            .loadGeoJson()
+            .then(() => this.buildDataSource(this.geoJsonData))
```
kring: I think the creation of the `DataSource` can be deferred until `mapItems` is accessed. Doing it here will not be reactive when styles are changed after the GeoJSON is loaded.

kring: Again, think "my mapItems are a function of my GeoJSON data", rather than "when I load GeoJSON, I need to create a data source".

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -0,0 +1,674 @@
...
+
+                    entity.polygon = (undefined as unknown) as PolygonGraphics;
+                }
+            }
+            return dataSource;
+        });
+    }
+}
+
+function nameIsDerivedFromUrl(name: string, url?: string) {
```
kring: I don't think this is used anymore, because the strata system eliminates the need for it 🎉.

kring (commented):
Looking good @na9da! This is going to be our exemplar for vector data sources, so pardon my nitpicking.

lib/Models/GeoJsonCatalogItem.ts
na9da: `buildDataSource()` is asynchronous, so this will need `mapItems` to return a `DataSource | Promise<DataSource>`. This is doable since, `DataSourceCollection.add()` can also accept a Promise. 

Does the whole change sound ok?

lib/Models/GeoJsonCatalogItem.ts
na9da: How about making it part of `CatalogMemberTraits`?

lib/Models/GeoJsonCatalogItem.ts
na9da: Understood the gist of the comment. But the suggested change to `@compute geoJsonData` from the unwrapped `_loadedGeoJson` might not be doable since we also need to `reprojectToGeographic()` which is an asynchronous call that will turn `geoJsonData` into a `Promise`.

lib/Models/GeoJsonCatalogItem.ts
kring: Probably not `CatalogMemberTraits` because groups will have that, too, and it doesn't make sense to show/hide a group. Maybe a new `MappableTraits`?

lib/Models/GeoJsonCatalogItem.ts
kring: It's possible to construct a `GeoJsonDataSource` and add it to the `DataSourceCollection` and _then_ call its `load` method. I think that would be better because it avoids dealing with the possibility that a `DataSource` is actually a `Promise<DataSource>` all over the place.

lib/Models/GeoJsonCatalogItem.ts
kring: Ah I see, that's a good point. I _think_ it can be handled again by deferring reprojection until `mapItems` is needed, and then constructing and returning the `GeoJsonDataSource` immediately, then doing the reprojection, and when the reproject promise resolves call `load` with the reprojected data. Does that make sense?

kring:
I'm merging this - even though it's not 100% done - because I realized (belatedly) we're making our lives more difficult than necessary having long-running branches in the fast-changing `mobx` world. Will discuss more at sprint planning tomorrow. Great work so far @na9da!

PR #3273, opened 2019-04-01
Update button positioning for mobile time series

soyarsauce:
When viewing time series data on mobile, the map
tool buttons overlap with the BottomDock items

Fixes https://github.com/TerriaJS/terriajs/issues/3267


soyarsauce:
Cons: still relies on absolute positioning -relative to a container (the map container) that doesn't know about the dock*, and has excess buffer space when a typical multi-line-timeline-titles only spans over 1 line

PR #3275, opened 2019-04-02
Blank catalog groups

### LINKED ISSUE #3224: My Data: parent list item squashed
### Steps to reproduce
1. My Data catalogue > Add web data
2. Add 'https://ows.services.dea.ga.gov.au/'
3. View untitled(?) parent list item/folder

### Screenshot
screen shot 2019-02-25 at 2 28 21 pm

soyarsauce:
- Stop overwriting with empty auto calculated names

When loading in a URL that doesn't have anything
appended after the last forward slash, it will
prevent the name being set once the server responds
with a title to use.

- Fix trash icon positioning in "My Data" catalog

- Add minimum height to empty `DataCatalogGroup`s

Resolves https://github.com/TerriaJS/terriajs/issues/3224


tephenavies:
Thanks @soyarsauce. This works but isn't very consistent with the same URL being entered without a trailing slash. I think if there is a slash is at the end it should use everything since the second last slash as the name.

Then https://ows.services.dea.ga.gov.au would give `ows.services.dea.ga.gov.au`
and https://ows.services.dea.ga.gov.au/ would give `ows.services.dea.ga.gov.au/`

soyarsauce:
Thanks for pointing out the inconsistency @steve9164 - I had a play around and it made me wonder why we use the URL instead of whatever nice title gets returned from the server?

image

kring:
> I had a play around and it made me wonder why we use the URL instead of whatever nice title gets returned from the server?

I don't think there's any good reason. Watch out for the possibility that the server doesn't return a nice name at all, but other than that what you're describing definitely sounds like an improvement!

soyarsauce:
> > I had a play around and it made me wonder why we use the URL instead of whatever nice title gets returned from the server?
> 
> I don't think there's any good reason. Watch out for the possibility that the server doesn't return a nice name at all, but other than that what you're describing definitely sounds like an improvement!

So, via https://github.com/TerriaJS/terriajs/pull/3275/commits/ed092935515c803f68299baf7ca93f9705efffd8 I thought about doing this on `CatalogMember.js` but I think we should leave the title blank on the model if that's what the server has decided, and do this nicer-name as part of rendering/the UI component instead. (And if the server returns a not-nice-name, well, we're currently rendering whatever it tells us anyway)

soyarsauce:
Just saw this issue again in passing, @steve9164 - are you happy with the behaviour on http://ci.terria.io/3224-blank-catalog-groups/ now?

lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
```diff
@@ -72,11 +72,32 @@ const DataCatalogGroup = createReactClass({
...
+        // Grab a name via nameInCatalog, if it's a blank string, try and generate one from the url
+        const group = this.props.group;
+        const nameInCatalog = group.nameInCatalog || '';
+        if (nameInCatalog !== '') {
+            return nameInCatalog;
+        }
+
+        const url = group.url || '';
+        // strip protocol
+        const httpProtocol = url.indexOf('http');
```
tephenavies: This could be simplified to a regular expression.
```suggestion
        return url.replace(/^https?:\/\//, '')
```

soyarsauce: That it could, thanks!

tephenavies (approved):

PR #3276, opened 2019-04-03
More principled portrayal of traits in various places, and more

### LINKED ISSUE #3291: Model architecture: Don't let explicit load calls reload unnecessarily
Currently if you call `loadMetadata` or `loadData` on a model, it will reload stuff, whether it needs to or not. We should make these function like they do in the current architecture where additional calls when loading is already in progress or done should do nothing. Unlike the old architecture, however, calling a `load` method after a relevant property changes (e.g. the `url`) should actually start a new load.

Started in the `kr-mobx-metadata` branch.

### LINKED ISSUE #3292: Model architecture: Improve intellisense for types-created-from-traits
We use `StratumFromTraits`, `FlattenedFromTraits`, and `ModelPropertiesFromTraits` to create these three different types, with subtly different characteristics, from traits classes. This process works, but because it uses a TypeScript `Pick` to remove the `____isModelTraits` property from the traits class, the type shown e.g. on hover in VSCode is really ugly. One solution is to just get rid of `__isModelTraits` entirely.

kring:
* Added `TypeConditionals` and `TypeModifiers` with some utilities for manipulating types and making queries about them. Among other uses, these can be combined with `TypeChecks.ts` to "test" types; a compilation failure is a test failure. See `test/Types/ModelPropertiesFromTraitsSpec.ts` for an example of type tests.
* Removed `getOrCreateStratum` because TypeScript currently does not intersect return types in functions in intersected types (Microsoft/TypeScript#30520). Instead, there are new `setTrait` and `getTrait` methods on `Model`.
* Added `hasTraits`, which serves as both a runtime check that a Model has a particular trait and a TypeScript type guard that allowing type-safe access to those traits. This eliminates the need to write a bunch of little interfaces like `AdjustableOpacity`.
* Re-added support for default values for traits.
* Added `StratumFromTraits`, `FlattenedFromTraits`, and `ModelPropertiesFromTraits`. These all take a traits class and produce a new type describing the interface to a single stratum, the `flattened` property, and the `Model` itself, respectively. The portrayal of traits in these three scenarios are different in terms of default values (only model properties have a default) and writability (only stratum traits are writable). These three types are tricky recursive types, and I kept getting them subtly wrong. So I've added tests for them and now I'm pretty sure they're right.
* Other minor odds and ends and cleanup.

Fixes #3291 
Fixes #3292


kring:
@steve9164 the TypeScript 3.4 error you reported on slack is now fixed.

PR #3277, opened 2019-04-04
Add AEC electoral boundaries 2019.

### LINKED ISSUE #3203: Add Commonwealth Electoral Divisions 2018 - ABS + AEC
Boundary files available here: http://www.abs.gov.au/AUSSTATS/abs@.nsf/DetailsPage/1270.0.55.003July%202018?OpenDocument

* [x] Add catalog item to NationalMap
* [x] Generate and host vector tiles
* [x] Configure in region mapping
* [x] Update csv-geo-au

stevage:
Fixes #3203.


kring:
This looks good to me. @steve9164 do you have any concerns?

tephenavies:
My only concern is that I haven’t reviewed the new s3 powered tile store and associated scripts and merging this would be a promise to support that system

tephenavies:
From reading associated issues it looks like these layers are needed now and should get into this month’s release. Let’s merge this now. That’ll also give us a test of the new system before porting all layers

PR #3278, opened 2019-04-04
3208 share from catalog

soyarsauce:
~draft for ci to resolve 3208~
Ready for review


soyarsauce:
See comment @ https://github.com/TerriaJS/terriajs/issues/3208#issuecomment-485657047 for the todo overhaul comments

lib/ReactViews/ExplorerWindow/ExplorerWindow.jsx
```diff
@@ -33,15 +34,25 @@ const ExplorerWindow = createReactClass({
...
     /* eslint-disable-next-line camelcase */
     UNSAFE_componentWillMount() {
+        const { terria, viewState } = this.props;
+
         this.props.viewState.explorerPanelAnimating = true;
 
         this._pickedFeaturesSubscription = ko.pureComputed(this.isVisible, this).subscribe(this.onVisibilityChange);
+
+        if (terria.sharedFromExplorerPanel) {
+            viewState.openAddData();
```
chloeleichen: we don't actually need this coz the `viewCatalogMember` function in `viewState.js` actually check if the `previewed`  item is added by user or from the default data explorer

lib/ReactViews/ExplorerWindow/ExplorerWindow.jsx
```diff
@@ -33,15 +34,25 @@ const ExplorerWindow = createReactClass({
...
+            viewState.viewCatalogMember(terria.catalog.shareKeyIndex[terria.previewedItemId]);
+        }
         this.onVisibilityChange(this.isVisible());
     },
 
     componentDidMount() {
         this.escKeyListener = e => {
-            if (e.keyCode === 27) {
+            // Only explicitly check share modal state, move to levels/"layers of modals" logic if we need to go any deeper
+            if (e.keyCode === 27 && !this.props.viewState.shareModalIsVisible) {
```
chloeleichen: we have a problem here. If share panel on the top right hand corner is open and explorer window is open, esc key close the share panel, and further esc key press does nothing. I think this behaviour is problematic. Ideally the order should be: close catalog share -> close explorer window -> close map share panel

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
```diff
@@ -16,12 +16,13 @@ const userPropWhiteList = ['hideExplorerPanel', 'activeTabId'];
  * Builds a share link that reflects the state of the passed Terria instance.
  *
  * @param terria The terria instance to serialize.
+ * @param viewState the viewState to read whether we're viewing the catalog or not
  * @returns {String} A URI that will rebuild the current state when viewed in a browser.
  */
-export function buildShareLink(terria) {
+export function buildShareLink(terria, viewState) {
```
chloeleichen: Not super convinced that we need to introduce `viewState` into BuildShareLink, though i guess non of these will be problem once we are in mobx land

chloeleichen (changes requested):
Thanks @soyarsauce

lib/Models/Terria.js
```diff
@@ -389,6 +408,7 @@ var Terria = function(options) {
         'configParameters', 'catalog', 'selectBox',
         'currentViewer', 'showSplitter', 'splitPosition', 'splitPositionVertical',
         'baseMapContrastColor'
+        // 'baseMapContrastColor', 'hashInterpreted'
```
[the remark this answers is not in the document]
soyarsauce: 👍

lib/Models/Terria.js
```diff
@@ -371,6 +371,25 @@ var Terria = function(options) {
      */
     this.showSplitter = false;
 
+    // /**
```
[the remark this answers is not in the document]
soyarsauce: 👍

lib/ReactViews/ExplorerWindow/ExplorerWindow.jsx
soyarsauce: Thanks! This also actually prompted me to check and catalog-shareable-user-added-shares were indeed not working, so I've fixed that w/ https://github.com/TerriaJS/terriajs/pull/3278/commits/19128b5edec86e58e35c86c5d741951872c1e6d0 too

lib/ReactViews/ExplorerWindow/ExplorerWindow.jsx
soyarsauce: Thanks for picking up on this too, the top-right-share-panel open + explorer window combination is fixed now but it also gets dismissed upon any further interaction inside the explorer window.

The end result is esc keys will always eventually bring you back to the map without modals open, but because we don't have any modal layer logic yet are you happy to leave it like this, then we raise an issue to create the new modal system post-mobx?

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
soyarsauce: We need some way to tell if the explorer window is open via `explorerPanelIsVisible` so I can't think of any other way since there's no reference from `viewState` -> `terria`? So yes hopefully mobx land simplifies this

lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
```diff
@@ -52,24 +83,32 @@
...
 }
-.clipboard{
-  margin: 10px 0 0 0;
+.clipboard-for-catalog-share {
+  margin: 15px;
 }
 
 .shareUrlfield{
   composes: field from '../../../../Sass/common/_form.scss';
-  background: $overlay;
```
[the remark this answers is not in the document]
soyarsauce: I've reverted the original share panel back to a dark input & fixed up the contrast for that embed-text-color w/ Phil's OK as well.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -227,18 +234,50 @@ const SharePanel = createReactClass({
...
+                <Otherwise>
+                    <div className={Styles.clipboardForCatalogShare}>
+                        <Clipboard theme="light" text={this.state.shareUrl} source={this.getShareUrlInput('light')} id='share-url' />
+                    </div>
+                </Otherwise>
+            </Choose>
+        );
+    },
+
+    renderContentForPrint() {
```
kring: Why "for print"?

lib/ReactViews/Preview/GroupPreview.jsx
```diff
@@ -64,5 +76,5 @@ const GroupPreview = createReactClass({
     },
 });
 
-export default GroupPreview;
+export default measureWidth(GroupPreview);
```
kring: Fancy!

lib/ReactViews/Preview/measureWidth.js
```diff
@@ -0,0 +1,76 @@
...
+import debounce from 'lodash.debounce';
+
+const getDisplayName = WrappedComponent => {
+    return WrappedComponent.displayName || WrappedComponent.name || 'Component';
+};
+
+/*
+    HOC to check component width & supply updated widths as a prop (stored interally via state)
+    Ensure the wrapped component contains a reference to the element you wish to measure the width of,
+    typically the root element. Do this via a callback instead of `React.createRef()` as it was written for pre-react-16.3
```
kring: Sounds like this originally came from another project. Are we definitely ok to use it?

test/ReactViews/Map/Panels/SharePanel/BuildShareLinkSpec.js
```diff
@@ -0,0 +1,94 @@
...
+            // close the catalog & the share link should reflect that
+            viewState.closeCatalog();
+            const params = decodeAndParseStartHash(buildShareLink(terria, viewState));
+            const initSources = flattenInitSources(params.initSources);
+
+            expect(initSources.previewedItemId).toBeUndefined();
+            expect(initSources.sharedFromExplorerPanel).toBeUndefined();
+
+            done();
+        }).otherwise(fail);
```
kring: I think you may need to use `done.fail` instead of just `fail` here, otherwise you'll get a timeout rather than a quick failure if the promise rejects.

Thanks for writing tests for this by the way! :)

kring (changes requested):
Looking good! Just a few things to fix up.

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
```diff
@@ -189,13 +194,21 @@ function addViewSettings(terria, initSources) {
             break;
     }
 
+    const itemIdToUse = viewState.viewingUserData() ?
+        defined(viewState.userDataPreviewedItem) && (viewState.userDataPreviewedItem.id || viewState.userDataPreviewedItem.uniqueId)
```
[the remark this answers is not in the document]
soyarsauce: Thanks - have re-written the logic to work with this case as well now, & another test to really makes sure in these conditions that `previewedItemId` doesn't get added to the share url @ https://github.com/TerriaJS/terriajs/pull/3278/commits/36cae5343a4eb1558e4db2914c3f88530d34e2c5#diff-ad2a0224e6b8153d6618d26e64308cf9R59

lib/Models/Terria.js
```diff
@@ -371,6 +371,18 @@ var Terria = function(options) {
      */
     this.showSplitter = false;
 
+    /**
+     * Gets or sets whether to toggle viewState's sharedFromExplorerPanel after loading start data. Default false.
+     * @type {Boolean}
+     */
+    this.sharedFromExplorerPanel = false;
```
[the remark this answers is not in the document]
soyarsauce: Thanks - it was used, then not, and now finally used again, almost not required but in the earlier mentioned case it will come in handy

lib/Models/Terria.js
```diff
@@ -371,6 +371,18 @@ var Terria = function(options) {
...
+
+    /**
+     * Gets or sets id of previewed item
+     * @type {string}
+     */
+    this.previewedItemId = undefined;
```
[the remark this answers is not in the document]
soyarsauce: This is great - moved everything over to the `ViewState` & removed all changes to ExplorerWindow @ https://github.com/TerriaJS/terriajs/pull/3278/commits/9db8f6159881d2d3d97251a8e0089384b17f777d

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
soyarsauce: SharePanel still gets used with the print/embed options via the top right icons on the map, "for print" probably wasn't the best name 😃 https://github.com/TerriaJS/terriajs/pull/3278/commits/45d2c3697a184b16f07d9ac88c0627b8f3b67742

lib/ReactViews/Preview/measureWidth.js
soyarsauce: I originally wrote it for another project and tied it in quite heavily to redux - so it's been adapted to what we have here, but I will confirm the definitely part & report back - in the meantime I may just make it work with both styles of refs as it's not like the callback ref has been deprecated.

test/ReactViews/Map/Panels/SharePanel/BuildShareLinkSpec.js
soyarsauce: Thanks for the pickup 👍

soyarsauce:
Addressed everything now but just waiting for the HoC-from-another-project confirmation

lib/ReactViews/Preview/measureWidth.js
kring: I'm not too fussed about the callback versus createRef, just wanted to make sure there are no IP issues.

soyarsauce: Yes hence the confirming the definitely part :) Should get an answer tomorrow.

soyarsauce: "of course!"
No issues here!

lib/ReactViewModels/ViewState.js
```diff
@@ -122,6 +127,28 @@ export default class ViewState {
...
+          this.explorerPanelIsVisible = true;
+        }
+      }, this);
+
+    this._previewedItemIdSubscription = knockout
+      .getObservable(terria, "previewedItemId")
+      .subscribe(previewedItemId => {
+        if (
+          defined(
+            previewedItemId && terria.catalog.shareKeyIndex[previewedItemId]
```
kring: It may be equivalent in this case, but this code is unusual. I'd think you would either want `if (defined(previewedItemId) && defined(terria.catalog.shareKeyIndex[previewedItemId]))` or `if (previewedItemId && terria.catalog.shareKeyIndex[previewedItemId])`. Using `defined` on the result of a logical `and` is not typically done.

lib/ReactViewModels/ViewState.js
```diff
@@ -122,6 +127,28 @@ export default class ViewState {
       }
     });
 
+    // Reflect preview id from terria when loaded from an init source
+
+    this._sharedFromExplorerPanelSubscription = knockout
+      .getObservable(terria, "sharedFromExplorerPanel")
+      .subscribe(sharedFromExplorerPanel => {
+        if (defined(sharedFromExplorerPanel)) {
```
kring: We don't care if it's true? Only that it's defined?

lib/ReactViewModels/ViewState.js
soyarsauce: Oops this certainly wasn't intended & was going for your latter, I just left off a closing paren.

lib/ReactViewModels/ViewState.js
soyarsauce: Thanks good pickup, I think my idea was the property only ever gets set if it is shared, but I've also now updated to check it's also true

soyarsauce:
I need to do some changes for mobile.

soyarsauce:
@philipgrimmett is happy for it to be disabled until he thinks of a mobile solution, so it has been disabled via `useSmallScreenInterface`.

kring:
This looks good to me. Do you want to review again before we merge @chloeleichen?

kring (approved):

chloeleichen (approved):
It works/looks awesome!

PR #3287, opened 2019-04-11
Limit x offset for chart to the gap available.

### LINKED ISSUE #3285: Moment lines hidden in chart
Steps to reproduce:

* Go to http://geoglam.d61.io/
* Click on the chart icon for the default catalog item (Total vegetation cover)
* Only lines up to June 2018 are shown, the rest are hidden

na9da:
Fixes  #3285


kring:
Thanks @na9da!
Quick nitpick, seems your editor mixing tabs and spaces which makes the formatting pretty funky in github diffs. We have an editorconfig file that tells many editor to use spaces and how many, but others need a plugin for it. Looks like you can add support to vim with a plugin: https://editorconfig.org/#download

PR #3290, opened 2019-04-12
Add support for async map items.

kring:
This is a PR into @na9da's `mobx-geojson` branch.

I noticed while reviewing #3258 that a GeoJSON load/reproject error would go unreported because no one was handling the promise rejection. After discussing with @steve9164, this is our attempt to make that less error prone by having an official `asyncMapItems` (via a mixin) that handles the details of turning promises for map items into real map items. I'm not 100% sure this is the right approach, but it's a start.


kring:
On further consideration, I don't think this is the right approach. Closing.

Issue #3226, opened 2019-02-25
Add web data: URL placeholder looks like a pre-fill value


philipgrimmett:
### Problem
Currently input placeholder text is too dark and looks like a pre-fill value. This makes it difficult for users to quickly determine which fields still require input.

### Screenshot
![screen shot 2019-02-25 at 3 21 01 pm](https://user-images.githubusercontent.com/5292858/53314638-98aac100-3913-11e9-9f7b-9852c6852171.png)

### Proposed solution
Body text: #606D80
Placeholder text: 0.5 opacity

kring:
Good early issue for Emma.

PR #3313, opened 2019-04-16
3226 lighter placeholders

KeyboardSounds:
Decreased opacity of placeholder text as suggested in #3226.

@philipgrimmett said to fix it for the entire site, and so I have. 😬

### DIFF
--- a/lib/Sass/common/_mixins.scss
+++ b/lib/Sass/common/_mixins.scss
@@ -7,14 +7,6 @@
     transition:         $transition;
 }
 
-@mixin placeholder {
-  &::-webkit-input-placeholder {@content}
-  &:-moz-placeholder {@content}
-  &::-moz-placeholder {@content}
-  &:-ms-input-placeholder {@content}
-}
-
-
 @mixin transform($transforms) {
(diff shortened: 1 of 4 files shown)


lib/Sass/common/_mixins.scss
```diff
@@ -7,14 +7,6 @@
     transition:         $transition;
 }
 
-@mixin placeholder {
```
KeyboardSounds, line 10: Mine is (marginally) prettier.

kring:
Looks good to me! @philipgrimmett do you want to just take a quick glance at http://ci.terria.io/3226-lighter-placeholders/ and make sure it looks ok to you? In particular, the location and data search boxes are lighter now.

philipgrimmett:
Hmm, 0.5 is pretty hard to read (sorry, I must have sketched it with a darker body colour). 0.6 looks a bit better to my eye.

Weird: looks like the two 'Add web data' input fields have slightly different font styles (colour, maybe size or padding?) when you enter a URL value.
Screen Shot 2019-04-16 at 11 55 20 am

PR #3324, opened 2019-04-17
Mobx Zoom

na9da:
WIP


lib/ModelMixins/ZoomableMixin.ts
```diff
@@ -0,0 +1,56 @@
+import CesiumMath from "terriajs-cesium/Source/Core/Math";
+import DataSource from "terriajs-cesium/Source/DataSources/DataSource";
+import Rectangle from "terriajs-cesium/Source/Core/Rectangle";
+import Constructor from "../Core/Constructor";
+import isDefined from "../Core/isDefined";
+import Model from "../Models/Model";
+import MappableTraits from "../Traits/MappableTraits";
+import Mappable, { isDataSource, ImageryParts } from "../Models/Mappable";
+
+type ZoomableModel = Model<MappableTraits>;
```
na9da: Zoom depends on the `rectangle` trait which is currently defined in `MappableTraits`. Move that to `ZoomableTraits`?

PR #3326, opened 2019-04-18
Implements zoomTo feature for mobx

### LINKED ISSUE #3293: Model architecture: Make "Zoom To" on the workbench work

na9da:
Fixes #3293

### DIFF
--- a/lib/Models/GeoJsonCatalogItem.ts
+++ b/lib/Models/GeoJsonCatalogItem.ts
@@ -58,6 +58,8 @@ class GeoJsonCatalogItem
 
     private _dataSource: GeoJsonDataSource | undefined;
 
+    readonly canZoomTo = true;
+
     constructor(id: string, terria: Terria) {
         super(id, terria);
     }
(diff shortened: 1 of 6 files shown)


lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -58,6 +58,8 @@ class GeoJsonCatalogItem
 
     private _dataSource: GeoJsonDataSource | undefined;
 
+    readonly canZoomTo = true;
```
kring, line 61: It's not immediately obvious what this is needed for. Can it be removed?

na9da: The UI needs it to test if the item supports zooming  [here](https://github.com/TerriaJS/terriajs/blob/mobx/lib/ReactViews/Workbench/Controls/ViewingControls.jsx#L104).

There is a similar field in [`WebMapServiceCatalogItem`](https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/WebMapServiceCatalogItem.ts)

kring: Oh right. Probably want to change how that works eventually, but all good for now.

PR #3333, opened 2019-04-23
fix not able to apply terrain catalog item splitter direction when from sharedCatalog

chloeleichen:
When an item comes from `sharedCatalog`, this change allow its `splitterDirection` be applied to the now viewing item


AnaBelgun:
with DT moving to mobx, this issue / pr will be redundant

PR #3336, opened 2019-04-23
Fix themeing

KeyboardSounds:
- Chart panel header and background now themable
- Fixed Feature Info "Download this Table" button icon
- "Download this Table" dropdown list now vertical instead of horizontal as requested by @philipgrimmett

### DIFF
--- a/lib/ReactViews/FeatureInfo/feature-info-download.scss
+++ b/lib/ReactViews/FeatureInfo/feature-info-download.scss
@@ -27,7 +27,11 @@
 .dropdown__list {
   display: flex;
   flex-direction: column;
+  // this makes sure the hover highlight fills the full width of the item
   li a {
+    display: block;
+    padding-top: 0;
+    padding-bottom: 0;
     @if variable-exists(text-link) {
       color: $text-link;
     }
--- a/lib/ReactViews/FeatureInfo/feature-info-download.scss
+++ b/lib/ReactViews/FeatureInfo/feature-info-download.scss
@@ -20,7 +20,7 @@
     right: 5px;
     height: 20px;
     width: 20px;
-    fill: $text-light;
+    fill: currentColor;
   }
 }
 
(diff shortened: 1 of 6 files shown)


lib/Sass/common/_variables.scss
```diff
@@ -149,6 +149,8 @@ $chart-text-color: #fff;
 $chart-line-color: #ccc;
 $chart-preview-line-color: $chart-line-color;
 $chart-darker: darken($dark, 8%);
+$chart-panel-background: $dark;
+$chart-panel-header: $dark;
```
chloeleichen: is this new design from @philipgrimmett ? otherwise the header should have a slight different shade of color from chart body
Screen Shot 2019-04-24 at 10 39 27 am

lib/ReactViews/FeatureInfo/feature-info-download.scss
```diff
@@ -12,17 +12,21 @@
...
     height: 20px;
     width: 20px;
-    fill: $text-light;
+    fill: currentColor;
   }
 }
 
 .dropdown__list {
   display: flex;
+  flex-direction: column;
```
chloeleichen, line 29: Screen Shot 2019-04-24 at 10 47 00 am
I can't seem to see the change? Am I looking at the wrong place? If so, these drop-downs should probably have the same styles too :).

chloeleichen (changes requested):
Nice Emma, thank you!

lib/ReactViews/FeatureInfo/feature-info-download.scss
KeyboardSounds: Different dropdown. Nice catch :) I'll fix this one too.

The one I changed was the one that says "Download this Table".

lib/Sass/common/_variables.scss
KeyboardSounds: Fixed :)

chloeleichen:
One more nitpicking, the dropdown element looks a bit strange when the container is so wide but the inner element is so narrow. Maybe we can make the highlight go all the way? 

Screen Shot 2019-04-26 at 10 15 07 AM

chloeleichen (approved):

PR #3338, opened 2019-04-28
Hook up My Data tab.

### LINKED ISSUE #3297: Model architecture: Hook up Add Data tab

na9da:
WIP for #3297 

TODOs

- [x] Fix behavior when adding same source multiple times
- [ ] Decide on interface for `createCatalogItemFromUrl.register()`. Should it simply take an item `type` or should it take a constructor?
- [x] Fix the display names
- [x] Fix zoom behavior
- [x] Support loading initialization files


lib/Models/CatalogNew.ts
```diff
@@ -1,12 +1,67 @@
...
+import { computed } from "mobx";
+import { USER_ADDED_CATEGORY_NAME } from "../Core/addedByUser";
+import isDefined from "../Core/isDefined";
+import hasTraits from "./hasTraits";
+import CatalogMemberTraits from "../Traits/CatalogMemberTraits";
+import Model, { BaseModel } from "./Model";
+import CommonStrata from "./CommonStrata";
+import ModelReference from "../Traits/ModelReference";
+
+class UserAddedDataGroup extends CatalogGroup {
```
kring: Do we need a type for this? I think these two methods could be added to `GroupMixin`. Should change them to take a `stratumId` as a parameter too, rather than assuming `CommonStrata.user`. And mark them `@action`. And it should be ok to `push` directly into the `members` array (creating it first if necessary) rather than using concat.

lib/Models/CatalogNew.ts
```diff
@@ -1,12 +1,67 @@
...
+    get userAddedDataGroup(): UserAddedDataGroup {
+        let group = <UserAddedDataGroup>(
+            this.group.memberModels.find(m => m.id === USER_ADDED_CATEGORY_NAME)
+        );
+        if (isDefined(group)) {
+            return group;
+        }
+
+        group = new UserAddedDataGroup(USER_ADDED_CATEGORY_NAME, this.terria);
+        group.setTrait(CommonStrata.user, "name", USER_ADDED_CATEGORY_NAME);
```
kring: This should be `CommonStrata.definition` rather than `user`.

lib/Models/CatalogNew.ts
```diff
@@ -1,12 +1,67 @@
...
+        group.setTrait(
+            CommonStrata.user,
```
kring: Same here.

lib/Models/CatalogNew.ts
```diff
@@ -1,12 +1,67 @@
...
+            "description",
+            "The group for data that was added by the user via the Add Data panel."
+        );
+        this.terria.addModel(group);
+        this.group.setTrait(CommonStrata.user, "members", [group.id]);
```
kring: It does make sense for this one to be `user` though.

kring: Actually, sorry, this should be `definition` too.

kring: However, the _contents_ of the user-added data group will be in the user stratum. A good rule of thumb to decide this is, if I needed to share the state of my app with others, what would I need to send them? The structure of the User-added data group is inherent to the app, but its contents are user-defined.

lib/Models/CatalogGroupNew.ts
```diff
@@ -2,6 +2,8 @@ import CatalogMemberMixin from '../ModelMixins/CatalogMemberMixin';
 import GroupMixin from '../ModelMixins/GroupMixin';
 import CatalogGroupTraits from '../Traits/CatalogGroupTraits';
 import CreateModel from './CreateModel';
+import { BaseModel } from './Model';
```
kring: Are these imports still needed?

lib/Models/addUserCatalogMember.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Adds a user's catalog item or group to the catalog.
+ *
+ */
+export default function addUserCatalogMember(
+    terria: Terria,
+    newCatalogMemberOrPromise: BaseModel | Promise<BaseModel | undefined>,
+    optionsArg?: AddUserCatalogMemberOptions
+) {
+    const options: AddUserCatalogMemberOptions = defaultValue(optionsArg, {});
```
kring: It's ok to just call `optionsArg` `options` and then reassign it with the default value:
```
options = defaultValue(options, {});
```

Sometimes considered a bit dodgy to reassign a parameter like this, but it's IMO ok in this case and it's very commonly done in Cesium and TerriaJS.

kring: I guess TypeScript might fight you on this though eh? hmm

lib/Models/addUserCatalogMember.ts
```diff
@@ -0,0 +1,71 @@
...
+    return when(newCatalogMemberOrPromise, function(newCatalogItem: BaseModel) {
```
kring: Different from how we've done things in the old architecture, we should use Cesium's when.js promises as little as possible. Whenever you get a promise back from Cesium, immediately convert it to a regular promise by calling `makeRealPromise`. Assume that promises everywhere in Terria are real native promises.

lib/Models/addUserCatalogMember.ts
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        terria.catalog.userAddedDataGroup.add(newCatalogItem);
+        terria.catalog.userAddedDataGroup.setTrait(CommonStrata.definition, "isOpen", true);
+
+        if (
+            isDefined(options.open) &&
+            hasTraits(newCatalogItem, GroupTraits, "isOpen")
+        ) {
+            newCatalogItem.setTrait(CommonStrata.definition, "isOpen", true);
```
kring: This should be `CommonStrata.user`.

lib/Models/addUserFiles.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    function loadInitData(initData: {catalog: any}) {
+        updateModelFromJson(terria.catalog.group, CommonStrata.definition, {members: initData.catalog});
+    }
+
+    for (let i = 0; i < files.length; i++) {
+        const file = files[i];
+        const tempCatalogItem = new ResultPendingCatalogItem(file.name, terria);
+        tempCatalogItem.setTrait(CommonStrata.definition, "name", file.name);
```
kring: All the traits set here should be in `CommonStrata.user`.

kring:
@na9da I don't see anything here that should stop you from merging this into `mobx`.

lib/Models/addUserCatalogMember.ts
na9da: Yeah, TS gives an error because the references are inside a closure. Can fix it though, will have to do an `isDefined(options)` test.

PR #3343, opened 2019-04-29
Adds the layer style to the GetLegendGraphic request for a WMS

### LINKED ISSUE #3330: Unclear desired behavior when some WMS styles don't define an LegendURL
There is some undesirable interaction between `TerriaJS` and `datacube-ows` when a layer has styles where only some styles have LegendURL defined. (I think it stems from both packages trying to support a variety of clients or servers that don't always follow the specs.)

`datacube-ows` has a WMS layer with several styles, some of them have legends, some of them do not. When a style is selected in Terria that has `LegendURL` defined, that legend is displayed. When a style without a LegendURL is selected, Terria makes a call to `GetLegendGraphic`, (without the `&styles=...` parameter set).
Although `datacube-ows` doesn't advertise in its `GetCaps` that is supports `GetLegendGraphic`, it returns with an image combining all the available styles for the layer.  Terria shows this image as the legend, which is not relevant to the selected style.

Example: https://terria-cube.terria.io/#share=s-hwhdgBTgp5pAPDHWJxqQtJ4d49R

I'm not sure what the best path forward is maintain support for all the non-standard?

Possible solutions for `datacube-ows`:
 * set `Style.LegendURL` to `''`: this results in the desired behavior, but breaks OGC compliance checks (Example: https://terria-cube.terria.io/#share=s-q8euAv8waNCWmFaEpg8yV74ePNV )
* set `Style.LegendURL` to a transparent 1x1 pixel: Terria displays it, but with a white background
* don't return `GetLegendGraphic` when some layers don't have legends: Not ideal for WMS clients that make no use of `Style.LegendURL` and rely on `GetLegendGraphic`. **Not sure what these other WMS clients are, or what we are assuming**

Ideas for possible solutions for `TerriaJS`:
* call `GetLegendGraphic` with `&styles=...`: does this break ncWMS support etc?
* don't call `GetLegendGraphic` if some styles have `LegendURL`: very edge-case specific solution?
* ?

rowanwins:
Resolves #3330

Currently if a WMS layer does not have a `legendUri` defined in the style Terria manually constructs a `GetLegendGraphic` URI, this occasionally leads to anomalies.

For example this is the url that Terria currently generates when trying to create a legend for a simple RGB layer that has no legend
[https://ows.services.dea.ga.gov.au/?service=WMS&version=1.1.0&request=GetLegendGraphic&format=image%2Fpng&transparent=True&layer=s2a_nrt_granule_nbar_t&srs=EPSG%3A3857](https://ows.services.dea.ga.gov.au/?service=WMS&version=1.1.0&request=GetLegendGraphic&format=image%2Fpng&transparent=True&layer=s2a_nrt_granule_nbar_t&srs=EPSG%3A3857) 
It returns all the available legend styles for the layer which means the UI shows some quirky things.

This pull request results in Terria appending the appropriate style name as part of the `GetLegendGraphic` request, in this case appending `styles=simple_rgb` for the layer even through the style doesn't have a `LegendURL`
[https://ows.services.dea.ga.gov.au/?service=WMS&version=1.1.0&request=GetLegendGraphic&styles=simple_rgb&format=image%2Fpng&transparent=True&layer=s2a_nrt_granule_nbar_t&srs=EPSG%3A3857](https://ows.services.dea.ga.gov.au/?service=WMS&version=1.1.0&request=GetLegendGraphic&styles=simple_rgb&format=image%2Fpng&transparent=True&layer=s2a_nrt_granule_nbar_t&srs=EPSG%3A3857)

What happens in this case is that the server sends `500` response because it can't find an appropriate response (not ideal perhaps) and the Terria UI shows nothing in the legend (which is the desired outcome).

This doesn't appear to have any adverse impacts on other WMS types (eg those coming from THREDDS)

cc @andrewdhicks


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1681,7 +1681,7 @@ function computeLegendForLayer(catalogItem, thisLayer, styleName) {
     if (!defined(legendUri)) {
         // Construct a GetLegendGraphic request.
         legendUri = new URI(cleanUrl(catalogItem.url) +
-            '?service=WMS&version=1.1.0&request=GetLegendGraphic&format=image/png&transparent=True&layer=' + encodeURIComponent(thisLayer.Name));
+            '?service=WMS&version=1.1.0&request=GetLegendGraphic&format=image/png&transparent=True&layer=' + encodeURIComponent(thisLayer.Name) + '&styles=' + encodeURIComponent(style.Name));
```
kring: Need to handle the possibility that `style` is undefined.

kring (changes requested):
Looks good, just that one comment. Please update `CHANGES.md` as well.

lib/Models/WebMapServiceCatalogItem.js
kring: I think we still want to generate a legend even if `style` is undefined, we just won't include the `style=` parameter.

rowanwins: Have added `if (!defined(legendUri) && defined(style) && defined(style.Name)) {` similar to what is done above

kring: But now if there's no `style`, there won't be a `legendUri` at all, right?

rowanwins: Ok yeah I see... so we just pass an empty string in that case? eg
````
if (!defined(legendUri)) {
   const styleToUse = defined(style) && defined(style.Name) ? style.Name : '';
````

kring: It'd be slightly better to leave out the `style=` entirely I think.

kring (approved):

kring:
Thanks @rowanwins!

PR #3346, opened 2019-04-30
Pass available layer styles to creation of WMS item when within group

### LINKED ISSUE #3329: Investigate - WMS with many layers/sub-layers causes browser to crash
## Issue
Currently using NationalMap if a user attempts to load a WMS with a large GetCapabilities document the browser crashes. 
- For example https://geo.weather.gc.ca/geomet/?service=WMS&request=GetCapabilities which is 9MB

## Options
~~1. Handle the parsing of XML files in [terriajs-server](https://github.com/TerriaJS/terriajs-server)~~
~~2. Improving the browser parsing~~
    ~~- There are 1001 XML parsing libs. One option may be a lib like [parse-xml](https://github.com/rgrove/parse-xml), I've picked this because it's small and browser friendly. Here is [an example](https://gist.github.com/rowanwins/7fcfc9ac41e5e00f111c62e7fa9c0dba) parsing the problematic GetCapabilities doc listed above in the browser which works fine. parse-xml weighs 9kb minified so it adds a bit of overhead compared to the current approach but it doesn't crash so that's the big plus.~~

~~My preference would be to keep the parsing browser side.~~

~~Any other thoughts from anyone?~~

Note: See additional diagnosis [below](https://github.com/TerriaJS/terriajs/issues/3329#issuecomment-485650776)

rowanwins:
Resolves #3329 by ensuring available layer styles are passed to the creation of a WMS when looping through a group of WMS layers.

A WMS with 2400 layers now loads in a couple of seconds quite happily without running out of memory, eg https://geo.weather.gc.ca/geomet/?service=WMS&request=GetCapabilities

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Change Log
 ==========
 
+## Next Release
+
+* Fixed a bug where all available styles were being retrieved from a `GetCapabilities` for each layer within a WMS Group resulting in memory crashes on WMS's with many layers.
+
+
 ### v7.3.0
 
 * Added `GltfCatalogItem` for displaying [glTF](https://www.khronos.org/gltf/) models on the 3D scene.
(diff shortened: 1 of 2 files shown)


kring:
Excellent! Mind updating `CHANGES.md`? You should create a new `## Next Release` section at the top for it.

kring (approved):

Issue #3350, opened 2019-04-30
Investigate removing moment.js dependency


rowanwins:
While poking around in the `WebMapServiceCatalogItem.js` I noticed we're using [moment.js](https://momentjs.com/) for some of the handling around dates & time for time-enabled WMS's. I did a quick search and I can't find any other references for moment across the terriajs code base.

Moment is ~50kb minified (16kb gzipped).

There are bunch of other date & time libraries now in the js ecosystem which are more modular and could possibly be used to reduce the overall bundle size (eg [luxon](https://moment.github.io/luxon/) which was born out of moment project, or [date-fns](https://date-fns.org/v2.0.0-alpha.27/docs/Getting-Started)).

It might be worth a quick look at how achievable it would be and what the overall gain would be.

nf-s:
moment is needed for `react-datepicker` (in `lib/ReactViews/BottomDock/Timeline/DateTimePicker.jsx`)

PR #3352, opened 2019-05-01
Removing moment.js dependency

rowanwins:
Investigation related to #3350

Momentjs was used in `WebMapServiceCatalogItem`, and I also discovered some use in the `BottmomDock/Timeline/DateTimePicker.jsx`which used an old version of `react-datepicker` which had a dependency on `moment.js`.

I used [`date-fns@2.0.0-alpha.27`](https://date-fns.org/v2.0.0-alpha.27/docs/Getting-Started) and [`iso8601-duration`](https://www.npmjs.com/package/iso8601-duration) to achieve the same functionality. 

I've had trouble noticing any real change to the overall build size when running `gulp release-app` on TerriaMap, however if I run a mocked up import of the required `date-fns`& `iso8601-duration` functions that I've used and compile them with `rollup` I end up with a package that weighs 5kb minified (compared to the ~50kb of moment.js) so I'm fairly confident that the benefit is there size-wise (unless the `webpack` config we've got is doing something funky).

I wouldn't mind a second pair of eyes to run over some of the date parsing logic. All the tests are passing but I'm not confident that I've entirely replicated the behaviour that was there.


rowanwins:
Not sure why travis is breaking...

kring:
Looks like a part of the process is crashing: https://travis-ci.org/TerriaJS/terriajs/jobs/527087718#L588

kring:
It's `generateSchema`, which tries to parse the code to extract doc comments. Unfortunately it's not ES2015-aware and probably bombing on the `import`. Use `require` for now.

soyarsauce:
Nice - the datepicker functionality still works, but should we use the `1.30.1` version of `date-fns` until their version 2 is out of alpha? The api at least looks the same at a quick glance

rowanwins:
Thanks for the suggestion @kring re `import` vs `require`

@soyarsauce - `date-fns` has been in a v2 alpha for 2 years :) But happy to drop back to `1.30.1` if that's the consensus.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1236,37 +1240,38 @@ function updateIntervalsFromIsoSegments(intervals, isoSegments, time, wmsItem) {
...
-             duration.hours() > 0 || duration.days() > 0 || duration.weeks() > 0 ||
-             duration.months() > 0 || duration.years() > 0)) {
-
-            var thisStop = start.clone();
+        var duration = isoDuration.parse(isoSegments[2]);
+        if (duration.seconds > 0 || duration.minutes > 0 ||
+            duration.hours > 0 || duration.days > 0 || duration.weeks > 0 ||
+            duration.months > 0 || duration.years > 0) {
+
+            var durationAsSeconds = isoDuration.toSeconds(duration);
```
kring: We can't just convert the duration to seconds, because years and months, at least, are different numbers of seconds. 1 month means 1 month, not 30 days.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1236,37 +1240,38 @@ function updateIntervalsFromIsoSegments(intervals, isoSegments, time, wmsItem) {
...
-                thisStop.add(duration);
+            while (thisStop && (isEqual(prevStop, stopDate) || isBefore(prevStop, stopDate)) && count < wmsItem.maxRefreshIntervals) {
+                thisStop = addSeconds(thisStop, durationAsSeconds);
                 intervals.addInterval(new TimeInterval({
-                    start: JulianDate.fromIso8601(prevStop.format()),
-                    stop: JulianDate.fromIso8601(thisStop.format()),
-                    data: formatMomentForWms(prevStop, duration) // used to form the web request
+                    start: JulianDate.fromDate(prevStop),
+                    stop: JulianDate.fromDate(thisStop),
+                    data: JulianDate.fromDate(prevStop) // used to form the web request
```
kring: This used to be a string, and now it's a JulianDate. The `data` property is used directly as the `time` parameter in the `GetMap`, so I'm pretty sure this will break WMSs that actually have a periodicity in their time dimension.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1279,20 +1284,6 @@ function updateIntervalsFromIsoSegments(intervals, isoSegments, time, wmsItem) {
     }
 }
 
-function formatMomentForWms(m, duration) {
-    // If the original moment only contained a date (not a time), and the
-    // duration doesn't include hours, minutes, or seconds, format as a date
-    // only instead of a date+time.  Some WMS servers get confused when
-    // you add a time on them.
-    if (duration.hours() > 0 || duration.minutes() > 0 || duration.seconds() > 0 || duration.milliseconds() > 0) {
```
kring: This logic is important and needs to be preserved.

lib/ReactViews/BottomDock/Timeline/DateTimePicker.jsx
```diff
@@ -175,12 +175,10 @@ const DateTimePicker = createReactClass({
...
     renderDayView(datesObject) {
       if (datesObject[this.state.year][this.state.month].dates && datesObject[this.state.year][this.state.month].dates.length > 31) {
-        // Create one date object per day, using an arbitrary time. This does it via Object.keys and moment().
+        // Create one date object per day, using an arbitrary time. This does it via Object.keys.
         const days = datesObject[this.state.year][this.state.month].indice;
-        const daysToDisplay = days.map(d => moment().date(d).month(this.state.month).year(this.state.year));
-        const selected = defined(this.state.day) ? moment().date(this.state.day).month(this.state.month).year(this.state.year) : null;
-        // Aside: You might think this implementation is clearer - use the first date available on each day.
-        // However it fails because react-datepicker actually requires a moment() object for selected, not a Date object.
+        const daysToDisplay = days.map(d => parse(`${d}/${this.state.month}/${this.state.year}`, 'dd/MM/yyyy', new Date()));
```
kring: Consider using `new Date(this.state.year, this.state.month - 1, d)` rather than parsing strings.

lib/Models/WebMapServiceCatalogItem.js
rowanwins: 🤦 tried to be smart but failed!

lib/Models/WebMapServiceCatalogItem.js
rowanwins: noted, will refactor with [this associated comment](https://github.com/TerriaJS/terriajs/pull/3352#pullrequestreview-233934328)

lib/Models/WebMapServiceCatalogItem.js
rowanwins: noted. Think it would be good to add a test around this somehow

rowanwins:
Thanks for the review @kring , I've made the bulk of those change.

Probably my primary area of concern I've got is that `moment` would provide you some information back on how it parsed the date that you gave it, for example you could do something like 
````
const m = moment.parseZone("2013-01-01")
m.creationData().format
// "YYYY-MM-DD"
````
We were using that [format information here](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/WebMapServiceCatalogItem.js#L1676). My understanding is that with the `formatMomentForWms` function we were basically trying to return a datestring either with time information, or without the time info. So with my current approach I'm just passing in a static date-time format, or a static date format, however these are just guess-ti-mated based on traditional ISO formats rather than what the server originally sent so I'm not sure how risky that is.

kring:
Thanks @rowanwins. Does the WMS spec require ISO8601 with `T` as the separator between the date and time portion? If so, and I think it does, then what you have should be fine. If not, we'd ideally detect all the other ways a time might be represented.

rowanwins:
[Sounds like](https://stackoverflow.com/a/9532375) the T is mandatory (unless specified). The WMS spec itself doesn't provide any direct info although all the examples use the T so presumably it is.

Part of the concern (which I should've mentioned above) was around things like hyphens (eg `2000-08-03` vs `20000803`) which I gather are both legit ISO8601 dates. However the WMS spec does specify to use the "extended" format which is the hyphened version.

kring:
Annex D of the WMS 1.3.0 spec seems pretty clear to be that there must be a `T`, so I'm happy. I don't think I understand your point about hyphens. Even if they were omitted (in spite of the WMS spec saying they shouldn't be), the ISO8601 parsing library we're using should be able to parse the date, right? Since it's valid and common-ish ISO8601.

rowanwins:
Yeah it was more if the server freaked because we'd changed date formats on it eg if the GetCapabilities sent a non-hyphenated date, and we sent it back a hyphenated date, I suspect it's a non-issue.

kring:
Oh ok, I see what you mean. Yeah hopefully not an issue. I definitely haven't seen any servers with that problem, at least.

rowanwins:
So I've updated this to run against the updated tests in #3373 

I do however have a failing test - `ignores leap seconds when evaluating period`

I've narrowed this down to some ISO datestring formatting that is being passed into a `Cesium.JulianDate.fromIso8601`

````
// Used to work and pass the leap second test but fails the new tests
formatDateToString(date, "yyyy-MM-dd'T'HH:mm:ss.SSSxxx")
=> 2015-07-01T11:15:00.000+10:00

// Passes the new tests except fails the leap second test
formatDateToString(date, "yyyy-MM-dd'T'HH:mm:ss'Z'")` 
=> 2015-07-01T11:15:00Z
````
The primary difference is that the prior one used to use milliseconds and a timezone

I've tried a bunch of variations to no avail, nothing makes all the tests happy .

kring:
@rowanwins the leap second test is failing because of a time zone problem. The first date/time in the WMS time dimension is: `2015-06-30T23:00:00Z`. The first date/time in the `intervals` collection is: `2015-07-01T09:00:00Z`, or 10 hours later. My time zone is UTC+10 in July, so something is handling time zones incorrectly.

Digging a little deeper, it looks like `formatDateToString(date, "yyyy-MM-dd'T'HH:mm:ss'Z'")` is the culprit. That `'Z'` on the end isn't going to make it convert the date to UTC, it's just going to slap a `Z` on the end. So we're taking a Date expressed in local time, formatting it as a string by extracting its parts, and then labeling it UTC even though it's not.

`formatDateToString(date, "yyyy-MM-dd'T'HH:mm:ss.SSSxxx")` will at least create an accurate representation of the time, even though it'll be expressed in local time and maybe something (the new tests?) expects it to be UTC. So that's closer. If the old code was always formatting the times as UTC, you'll probably need to convert the Date to UTC before calling `formatDateToString` in order to make all the tests pass. I haven't looked at the new tests though to confirm this.

rowanwins:
This is all now passing however the bundle size has gone up due to something in `date-fns-tz` so I suggest holding off merging until that is resolved. Will post a link to the issue I raise in `date-fns-tz`.

PR #3354, opened 2019-05-02
Use terriajs-cesium v1.57.0.

kring:


kring:
I'm insta-merging this because we don't have a globe in dev until this is released.

PR #3355, opened 2019-05-02
Mobx BingMapsCatalogItem

### LINKED ISSUE #3296: Model architecture: Implement BingMapsCatalogItem
And make sure it works as a base map.

na9da:
Fixes #3296

### DIFF
--- a/lib/Models/BingMapsCatalogItem.ts
+++ b/lib/Models/BingMapsCatalogItem.ts
@@ -1,0 +1,40 @@
+import Mappable from "./Mappable";
+import CreateModel from "./CreateModel";
+import BingMapsCatalaogItemTraits from "../Traits/BingMapsCatalogItemTraits";
+import BingMapsImageryProvider from "terriajs-cesium/Source/Scene/BingMapsImageryProvider";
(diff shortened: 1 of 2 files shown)


lib/Models/BingMapsCatalogItem.ts
```diff
@@ -0,0 +1,40 @@
+import Mappable from "./Mappable";
+import CreateModel from "./CreateModel";
+import BingMapsCatalaogItemTraits from "../Traits/BingMapsCatalogItemTraits";
+import BingMapsImageryProvider from "terriajs-cesium/Source/Scene/BingMapsImageryProvider";
+import Credit from "terriajs-cesium/Source/Core/Credit";
+import { computed } from "mobx";
+
+export default class BingMapsCatalaogItem
+    extends CreateModel(BingMapsCatalaogItemTraits)
```
kring, line 9: Not strictly needed for use as a base map, but because this is named as a catalog item it should really mix in `CatalogMemberMixin`. Also it should have a static and instance `type` property (see `WebMapServiceCatalogItem`).

Also, small typo: BingMapsCatalaogItem

PR #3361, opened 2019-05-05
Run prettier

kring:
Because this rewrites pretty much every source file, merging this into other branches will be a hassle. Here's how to do it (relatively) painlessly:
* Create a new branch for your merge: `git checkout -b whatever-merge`
* Merge the last commit _before_ prettier into your branch: `git merge pre-prettier`. Resolve any conflicts as usual.
* Commit the merge above.
* Merge the commit that ran prettier on all the source files into your branch. This is likely to cause heaps of conflicts, but because you've already merged the last commit before prettier, and because the prettier commit _only_ changes formatting, you can safely accept your version any time there is a conflict. ` git merge post-prettier --strategy=ours --no-commit`.
* Run prettier on the result of the merge before committing it: `npm run prettier`
* Commit the merged result.
* Merge `master` into your branch in order to pick up any changes in master that happened after the prettiergeddon: `git merge origin/master`. Resolve any conflicts as normal.
* Commit and push your branch.
* Open a pull request of your merge branch into the original one. It should merge cleanly.

If you're merging a branch-that-has-already-followed-this-procedure into your branch, you can follow the procedure above _except_ that you need to use different commits instead of `pre-prettier` and `post-prettier`. Look at the commits in the source branch and you should see two that look like this:

image

Use the commit hash of the commit that merged `pre-prettier` in place of `pre-prettier` above. Use the commit hash of the commit that merged `post-prettier` in place of `post-prettier` above.


chloeleichen (approved):
awesome

Issue #3286, opened 2019-04-11
[EPIC] Search 


juan-guerschman:
apologies if this is already open in some other ticket, if so I couldn't find it.

I want to quickly find the SA2: 201031017  in https://nationalmap.gov.au/#share=s-w89XMVg2poL5Hvm2HUU8IVwoTdV 

There are too many so clicking one by one until finding it is not practical. 
It would be good if the user could open a table and list all elements in the layer and select them in such a table. 

Is such a thing possible?

AnaBelgun:
Hi @juan-guerschman 
to understand the issue better
1. do you mean downloading the Feature Info panel - currently that's possible as csv/json
2. do you mean once you added the boundaries, to search by SA2 numbers, rather than name (which would take you to the location/area)
3. or do you mean something completely different e.g. making possible to download any data (inlcuding WMS, WFS etc) as a CSV/tabular 
we had some discussions about similar issues here: https://github.com/TerriaJS/nationalmap/issues/174
let me know

juan-guerschman:
I mean #2 
being able to find a SA2 element by its name or any other of the elements
![image](https://user-images.githubusercontent.com/8801702/56018329-27995e80-5d45-11e9-89d3-2872b390918d.png)

AnaBelgun:
@kring is this a more generic "search within a WMS/WFS" feature?

kring:
> is this a more generic "search within a WMS/WFS" feature?

Yes, I think that's what it is.

I don't think this is a "this sprint" thing, but I'll assign it to Ana so she can decide.

AnaBelgun:
yep, definitely not this sprint (with 30+ issues on the list). I'll leave it in the product backlog for now.

rowanwins:
Dropping a few thoughts on this @AnaBelgun 

- It would be worthing thinking about whether we just wanted to provision search on just some of the typical admin boundaries layers (eg for SA1, SA2 etc) vs whether we want people to search any data that they add to the map
  - being able to search on a few limited things would be simpler than working out how to support search on any layer that gets added

- If we were going for the later (to support any layer which a user added) we'd need to implement different approaches for different layer types (eg you'd query WMS vs WFS differently) so we'd want to pick some priority data source types.

- The other challenge with the approach of supporting searching on any layer is knowing what attribute/property of the data someone wants to search on as we'd want to avoid having to do a global search on the data... so if we went down this route we'd probably need to introduce some new UI pieces that don't yet exist.
[...]

PR #3370, opened 2019-05-07
[Work In Progress] - Workbench data search

rowanwins:
This pull request outlines some initial concepts around a workbench data search related to #3286.

Some thoughts
- Some of the workbench items with require async vs sync searching (eg sending a WMS request where the browser doesn't yet have the data, vs a a geojson source where the browser does have the data)
- Searching on region-mapped datasets (which is mentioned in a fair few of the related tickets) will be a bit fiddlier and require work on our data-structures. With the current vector tile approach getting the match isn't too bad, however zooming to the match is tricky
  - my initial implementation was targeting region-mapped csv's but I think this is now blocked until we have an effective way to retrieve the bounding box for the matched feature
  - So WFS or WMS may be a better first cab of the rank for getting end-to-end results
- Haven't touched tests yet until the general approach is supported by someone else.

ExampleSEarch

Happy to take whatever feedback people may have

### DIFF
--- a/lib/Models/WebMapServiceCatalogItem.js
+++ b/lib/Models/WebMapServiceCatalogItem.js
@@ -38,6 +38,7 @@ var unionRectangleArray = require('../Map/unionRectangleArray');
 var xml2json = require('../ThirdParty/xml2json');
 var LegendUrl = require('../Map/LegendUrl');
 var callWebCoverageService = require('./callWebCoverageService');
+var WebFeatureServiceCatalogItem = require('./WebFeatureServiceCatalogItem');
 
 /**
  * A {@link ImageryLayerCatalogItem} representing a layer from a Web Map Service (WMS) server.
@@ -85,6 +86,18 @@ var WebMapServiceCatalogItem = function(terria) {
      * @type {String}
      */
     this.linkedWcsCoverage = '';
+
+    /**
+     * Specifies whether a linked WFS has been found for this WMS item.
+     * @type {String}
+     */
+    this.hasLinkedWfs = undefined;
+
+    /**
+     * Gets or sets the URL of a WFS that enables search for this WMS item.
+     * @type {String}
+     */
+    this.linkedWfs = undefined;
 
     /**
      * Gets or sets the comma-separated list of styles to request, one per layer list in {@link WebMapServiceCatalogItem#layers}.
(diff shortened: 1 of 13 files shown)


rowanwins:
Have added support for WFS searching given that that's a bit more predictable than CSV

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -86,6 +87,18 @@ var WebMapServiceCatalogItem = function(terria) {
...
 
+    /**
+     * Specifies whether a linked WFS has been found for this WMS item.
+     * @type {String}
+     */
+    this.hasLinkedWfs = undefined;
+
+    /**
+     * Gets or sets the URL of a WFS that enables search for this WMS item.
+     * @type {String}
```
kring, line 98: It's not a string.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1143,6 +1156,40 @@ WebMapServiceCatalogItem.prototype.exportData = function() {
...
+           resolve(workbenchItemResults);
+        } else if (that.hasLinkedWfs) {
+            that.linkedWfs.searchDataForMatchingFeatures(searchText, regex).then(function (res) {
+                workbenchItemResults = convertWfsSearchResultsToWms(res, that);
+                resolve(workbenchItemResults);
+            });
+        } else if (!defined(that.hasLinkedWfs)) {
+            loadXML(that.getCapabilitiesUrl.replace(/wms/gi, 'wfs'))
+            .then(function(xml) {
+                that.linkedWfs = new WebFeatureServiceCatalogItem(that.terria);
```
kring: We work with a lot of WMS servers that have so much data that loading the whole thing with a single WFS will crash the server, the browser tab, or both. Instead, we should construct a WFS `GetFeature` query to do the search on the server.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1143,6 +1156,40 @@ WebMapServiceCatalogItem.prototype.exportData = function() {
...
+                that.linkedWfs.url = cleanUrl(that.getCapabilitiesUrl.replace(/wms/gi, 'wfs'));
+                that.linkedWfs.typeNames = that.layers;
+                that.linkedWfs.requestGeoJson = true;
+                that.linkedWfs.requestGml = true;
+                that.linkedWfs._load().then(function (result) {
```
kring: Call `load` instead of the private `_load`.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1143,6 +1156,40 @@ WebMapServiceCatalogItem.prototype.exportData = function() {
     callWebCoverageService(this);
 };
 
+WebMapServiceCatalogItem.prototype.searchDataForMatchingFeatures = function(searchText, regex) {
+    const that = this;
+    return new Promise(function(resolve, reject) {
```
kring: It's almost never necessary to explicitly construct promises, definitely not necessary here.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1143,6 +1156,40 @@ WebMapServiceCatalogItem.prototype.exportData = function() {
...
+        let workbenchItemResults = [];
+        if (that.hasLinkedWfs === false) {
+           resolve(workbenchItemResults);
+        } else if (that.hasLinkedWfs) {
+            that.linkedWfs.searchDataForMatchingFeatures(searchText, regex).then(function (res) {
+                workbenchItemResults = convertWfsSearchResultsToWms(res, that);
+                resolve(workbenchItemResults);
+            });
+        } else if (!defined(that.hasLinkedWfs)) {
+            loadXML(that.getCapabilitiesUrl.replace(/wms/gi, 'wfs'))
```
kring: Instead, return this promise. An important rule is: you must either handle (.then _and_ .catch) or return every promise.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -1143,6 +1156,40 @@ WebMapServiceCatalogItem.prototype.exportData = function() {
...
+                that.linkedWfs.requestGeoJson = true;
+                that.linkedWfs.requestGml = true;
+                that.linkedWfs._load().then(function (result) {
+                that.hasLinkedWfs = true;
+                that.linkedWfs.searchDataForMatchingFeatures(searchText, regex).then(function (res) {
+                    workbenchItemResults = convertWfsSearchResultsToWms(res, that);
+                    resolve(workbenchItemResults);
+                });
+                })
+                .catch(function (err) {
```
kring: This is probably crashing, it's called `.otherwise` in when.js promises, which `loadXML` returns.
In the current version of TerriaJS, we use Cesium's when.js promises everywhere. If we happen to get another kind of promise from somewhere, we immediately turn it into a when.js promise.

In the new model architecture (`mobx` branch), the rule is flipped. Native promises everywhere, convert anything else to a native promise immediately.

lib/Models/WebMapServiceCatalogItem.js
kring: Also, we can potentially use `DescribeLayer` to discover WFS services.

lib/Models/workbenchDataSearchUtils.js
```diff
@@ -0,0 +1,31 @@
+const GeoJsonCatalogItem = require('./GeoJsonCatalogItem');
+const defined = require('terriajs-cesium/Source/Core/defined');
+
+/**
+ * Adds a location marker to the map with the position supplied in the result, adding a data source to terria if one hasn't
+ * already been added, and removing all previously added markers in that data source. This data source is stored in
+ * terria.locationMarker.
+ */
+export function addGeoJsonFeatureFromWorkbenchData(terria, feature) {
```
kring: Name this file after this function and export it as the default.

stevenjh:
Hi this is a a feature that I'm really keen to see included, has this pull request become somewhat orphaned due to the migration to mobx?

rowanwins:
Hi @stevenjh 

It's fallen on the backburner for the timebeing as we've had a number of competing priorities but we will come back to it - and yeah we'll probably introduce as a new feature in mobx.

Cheers

PR #3372, opened 2019-05-07
Document making husky work with yarn workspaces

kring:

PR #3373, opened 2019-05-08
Add additional tests around date-time handling for WMS

rowanwins:
To support the work in #3352 to remove momentjs works I've written a bunch more tests to help ensure that all the avenues to interval creation produce legit and equivalent results. 

These tests are running against the `master` branch at moment and so I can then pull them into my branch and ensure I get the same results.

### DIFF
--- a/test/Models/WebMapServiceCatalogItemSpec.js
+++ b/test/Models/WebMapServiceCatalogItemSpec.js
@@ -454,6 +454,20 @@ describe("WebMapServiceCatalogItem", function() {
       .load()
       .then(function() {
         expect(wmsItem.intervals.length).toEqual(13);
+        
+        const firstInterval = wmsItem.intervals.get(0);
+        expect(firstInterval.data).toEqual("2002-01-01T00:00:00.000Z");
+        expect(firstInterval.start.dayNumber).toEqual(2452275);
+        expect(firstInterval.start.secondsOfDay).toEqual(43232);
+        expect(firstInterval.stop.dayNumber).toEqual(2452640);
+        expect(firstInterval.stop.secondsOfDay).toEqual(43232);
+
+        const lastInterval = wmsItem.intervals.get(12);
+        expect(lastInterval.data).toEqual("2014-01-01T00:00:00.000Z");
+        expect(lastInterval.start.dayNumber).toEqual(2456658);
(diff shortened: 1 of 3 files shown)


rowanwins:
I am experiencing one [curious result (which I've commented out)](https://github.com/TerriaJS/terriajs/pull/3373/files#diff-614c18d90e0c1ade4388de9ed2f562a0R491) as a result of [this code](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/WebMapServiceCatalogItem.js#L1569-L1601) where I think we should be passing an iso datestring to the `data` param, but we're passing a momentjs object. I suspect this is an error...

test/Models/WebMapServiceCatalogItemSpec.js
```diff
@@ -472,6 +486,15 @@ describe("WebMapServiceCatalogItem", function() {
       .load()
       .then(function() {
         expect(wmsItem.intervals.length).toEqual(1);
+
+        const firstInterval = wmsItem.intervals.get(0);
+        // expect(firstInterval.data).toEqual("2015-04-27T06:15:00.000Z");
+        // expect(typeof firstInterval.data).toEqual("string");
```
kring, line 492: Yeah this is a weird case. Re-reading Annex D in the WMS spec, I think it might not even be valid. It would have to mean "any time in this interval is acceptable, and consider the layer to be continuously varying over that interval". I'm not really sure what we could do with that, and it's hard to even imagine a scenario where a server would really have such a thing. I guess if it were able to interpolate values at any time it might offer a time dimension like this? In any case, unless you come across an actual server like this, I'm happy to put it aside / not worry about it.

kring (approved):

kring:
Just run prettier in here (see #3372 for how to set up your environment so it happens automatically) and I'm happy to merge this.

test/Models/WebMapServiceCatalogItemSpec.js
rowanwins: Whether there is a WMS server out there that would do it or not is one thing, it may well be dead code... That said if there is a server out there we probably shouldn't be sending a moment object but a datestring.

I suggest we leave in the handling for the timebeing, I'll fix the handling in #3352 so that we have working code, and then we can kill it off as a next step. That way we've at least got something to revert back to just in case...

Issue #3103, opened 2018-10-04
Display refresh rate timer for real-time datasets


philipgrimmett:
### Problem description
Real-time data is neat, but it can be frustrating waiting for a feature to update if you don't have an indication of the expected refresh rate.

### Proposed solution
Display an animated update timer in the header of real-time workbench items. This should help to distinguish static data from real-time data, and communicate the expect refresh rate.

### Screenshot / Design / File reference
<img width="650" alt="screen shot 2018-10-04 at 9 55 52 am" src="https://user-images.githubusercontent.com/5292858/46449952-070aad80-c7d2-11e8-80b4-5eb052a3ce5c.png">

Code reference: https://codepen.io/HugoGiraudel/pen/BHEwo

PR #3374, opened 2019-05-08
3103 data refresh timer

KeyboardSounds:
Adds data refresh timer for CSV workbench items that poll for changes.

The only example I could find is [here](http://ci.terria.io/3103-data-refresh-timer/#share=s-ifNc77hfWaiBudJKURtdcxVPui7).

This doesn't finish #3103, I still need to get the refresh timer showing for other kinds of datasets. Also, I haven't written any tests yet. It would be good to get peoples' thoughts on what sort of testing is necessary here.


lib/Models/CsvCatalogItem.js
```diff
@@ -174,14 +174,30 @@ CsvCatalogItem.prototype.loadIntoTableStructure = function(url) {
  */
 CsvCatalogItem.prototype.startPolling = function() {
   const polling = this.polling;
+
   if (defined(polling.seconds) && polling.seconds > 0) {
     var item = this;
+
+    // Initialise polling and timer variables, because they might not be set yet
+    polling.isPolling = item.isEnabled;
```
KeyboardSounds, line 182: A separate `isPolling` variable is not strictly necessary, you can check if there's a nextScheduledUpdateTime. But this way someone using this (albeit internal) API doesn't need to know that nextScheduledUpdateTime = isPolling, which might also not always be the case? Maybe? They're technically two separate pieces of information.

KeyboardSounds:
Whooops found a bug. Will reopen once it's fixed.

KeyboardSounds:
Bugs were squashed.

chloeleichen (changes requested):
A problem with using d3 transition. D3 uses requestAnimationFrame, which pauses when the browser tab is minimised or hidden (https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). The problem it causes us is that the timer tooltip and timer icon animation goes out of synch. 

A work around be to synch the animation again when tabs come into view again, with the visibility API perhaps: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
Screen Shot 2019-05-13 at 2 22 37 PM

lib/ReactViews/Generic/Timer/Timer.jsx
```diff
@@ -0,0 +1,102 @@
...
+  },
+
+  componentDidUpdate() {
+    this.startTimer();
+  },
+
+  componentWillMount() {
+    // We need a unique selector for the timer container.
+    // If we use a class and there are multiple timers, our drawTimer functions don't know which one to draw to.
+    this.containerId = "timer-container-" + new Date().getTime().toString();
```
chloeleichen: we can use Cesium `createGrid` to generate unique id here

lib/Models/CsvCatalogItem.js
```diff
@@ -210,6 +226,9 @@ CsvCatalogItem.prototype.startPolling = function() {
             }
           });
       }
+      // update isPolling - if the item is disabled then we are not polling
+      polling.isPolling = false;
+
       // Note this means the timer keeps going even when you remove (disable) the item,
       // but it doesn't actually request new data any more.
       // If the item is re-enabled, the same timer just starts picking it up again.
```
chloeleichen, line 234: Is it that important that the timer picks up from the last one when item is disabled and enabled again? I don't see much point of that here since data is not being updated in the background anyway. Instead, when an item is disabled, we can clear the timeout and only keep polling as long as the item is enabled? Especially if we have a lot of datasets with polling and each have a timer constantly going in the background even after they have been disabled

lib/ReactViews/Generic/Timer/drawTimer.js
```diff
@@ -0,0 +1,186 @@
...
+  // The arc representing the elapsed time should be filled up to the current time.
+  // We find the elapsed time as a percentage of the total duration, and then get the angle interpolator to calculate
+  // the corresponding angle.
+  const startAngle = angleInterpolator(options.elapsed / interval);
+  elapsedTimeElement
+    .datum({ endAngle: angleInterpolator(startAngle) })
+    .transition("arc" + new Date().getTime().toString())
+    .duration((interval - options.elapsed) * 1000) // d3 uses milliseconds
+    .ease(d3EaseLinear)
+    // attrTween requires a generator function A that returns a function B
```
chloeleichen: a bit confusing to call the function passed into `attrTween` a "generator function" in the comment as the generator function has some specific meaning. In this case it is just a function that returns another function? Maybe currying is a more accurate description?

lib/ReactViews/Workbench/Controls/TimerSection.jsx
```diff
@@ -0,0 +1,136 @@
...
+    this.startCountdown();
+  },
+
+  componentWillUnmount() {
+    clearInterval(this.interval);
+  },
+
+  render() {
+    return (
+      <div>
```
chloeleichen: Since this div is not used anywhere, we can use `React.Fragment` instead so we don't have an extra node in the DOM

chloeleichen (changes requested):

lib/Models/CsvCatalogItem.js
KeyboardSounds: The actual model layer code in this file for performing the polling was here already, I just added a few variables. It's definitely less efficient than it could be, but I decided to focus on the UI stuff rather than rewriting this bit since it's most likely going to be rewritten in less than 6 months anyway.

KeyboardSounds: We should absolutely do what you're suggesting, I'm just not sure if we should do it now.

chloeleichen: It might be something that should go into a different issue, but for the sake of trying, here's just an idea, what about subscribe to `isEnabled` in `CSVCatalogItem` and startPolling every time `isEnabled` is true? I do not know what it was done the original way, maybe you know the big picture better than me. 

If not we should create a seperate issue to document this & address it in near future.

chloeleichen:
Looks awesome. One last comment for @philipgrimmett : is it necessary to have the fade in/fade out (opacity) animation? Personally I find it a tad distracting. I'm thinking of the refresher on trip view, which is the same with ours but doesn't have the "blinking effect".

lib/Models/CsvCatalogItem.js
KeyboardSounds: @kring any idea why it might have been originally written this way? because I don't know why either haha

philipgrimmett:
@chloeleichen re. the pulse animation: when the update frequency of a dataset is >5min it can be difficult to tell whether the timer is actually moving, and the dataset appears more static than dynamic (probably not the correct terminology, but eh). The aim of the pulse is to communicate a 'live' quality, while the timer and text are there to help users decide whether they want to wait around for the next update. 

That said, users might still find it distracting. We should keep an eye out for similar feedback and make changes if they stack up.

chloeleichen (approved):
Awesome!

PR #3377, opened 2019-05-09
Now uses newer Bing Maps styles.

### LINKED ISSUE #3113: Bing Maps AerialWithLabels is deprecated
https://github.com/AnalyticalGraphicsInc/cesium/issues/7128

KeyboardSounds:
Also, updates terriajs to use terriajs-cesium v1.57.1

Fixes #3113

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "string-replace-webpack-plugin": "^0.1.3",
     "style-loader": "^0.23.1",
     "svg-sprite-loader": "4.1.3",
-    "terriajs-cesium": "1.57.0",
+    "terriajs-cesium": "1.57.1",
     "terriajs-html2canvas": "1.0.0-alpha.12-terriajs-1",
     "urijs": "^1.18.12",
     "url-loader": "^1.1.2",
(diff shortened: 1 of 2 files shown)


kring:
Need to run prettier on these changes (sorry).

KeyboardSounds:
> Need to run prettier on these changes (sorry).

Done :)

soyarsauce:
The only thing to discuss here is whether we continue to lock the patch version of our own terriajs-cesium - even though we do the updates, I think from a package point of view it makes sense to do at least tilde / `x.y.*`, so that if for some reason another project using terriajs depends on terriajs-cesium they don't unnecessarily get two copies bundled in?

package.json
```diff
@@ -75,7 +75,7 @@
     "string-replace-webpack-plugin": "^0.1.3",
     "style-loader": "^0.23.1",
     "svg-sprite-loader": "4.1.3",
-    "terriajs-cesium": "1.57.0",
+    "terriajs-cesium": "1.57.1",
```
soyarsauce, line 78: The only thing to discuss here is whether we continue to lock the patch version of our own terriajs-cesium - even though we do the updates, I think from a package point of view it makes sense to do at least tilde / `x.y.*`, so that if for some reason another project using terriajs depends on terriajs-cesium they don't unnecessarily get two copies bundled in?

kring: No, we need an exact version. Cesium unfortunately doesn't follow semver. And our version number tracks Cesium's except that we increment the patch version each time we need a new release. But these patch releases can and do introduce new features and break things. In short, the only safe way to depend on terriajs-cesium is to depend on an exact version of terriajs-cesium.

kring (approved):

PR #3380, opened 2019-05-10
Add support for overflow scrolling on dropdown component

### LINKED ISSUE #3337: Problem scrolling down when manually selecting web service type
If I try to add my own data from a web service and I want to see what types are available I can't scroll the select box.

AddData

rowanwins:
See context in #3337

Currently dropdown boxes like the 'Select file type' within 'Add Data' gets chopped off and is somewhat unusable because it has many options, most of which you can't see because the dropdown is unscrollable.

This pull request adds support for overflow scrolling on the dropdown component. I've added a `height` prop on the `Dropdown` component which defaults to `auto`. Where intervention is required though (eg when there are many options and space is constrained) we can now pass through a height prop which gets add as an inline style.

My screenshot wasn't can't quick enough to capture the scrollbar but it does appear :)
Scrollable

Resolves #3337

### DIFF
--- a/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
+++ b/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
@@ -150,7 +150,6 @@ const AddData = createReactClass({
               selectOption={this.selectLocalOption}
               matchWidth={true}
               theme={dropdownTheme}
-              height="150px"
             />
             <label className={Styles.label}>
               <strong>Step 2:</strong> Select file
@@ -174,7 +173,6 @@ const AddData = createReactClass({
               selectOption={this.selectRemoteOption}
               matchWidth={true}
               theme={dropdownTheme}
-              height="150px"
             />
             <label className={Styles.label}>
               <strong>Step 2:</strong> Enter the URL of the data file or web
--- a/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
+++ b/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
@@ -150,6 +150,7 @@ const AddData = createReactClass({
               selectOption={this.selectLocalOption}
               matchWidth={true}
               theme={dropdownTheme}
+              height="150px"
             />
             <label className={Styles.label}>
               <strong>Step 2:</strong> Select file
@@ -173,6 +174,7 @@ const AddData = createReactClass({
               selectOption={this.selectRemoteOption}
               matchWidth={true}
               theme={dropdownTheme}
+              height="150px"
             />
             <label className={Styles.label}>
               <strong>Step 2:</strong> Enter the URL of the data file or web
--- a/lib/ReactViews/Generic/dropdown.scss
+++ b/lib/ReactViews/Generic/dropdown.scss
@@ -21,6 +21,7 @@
 
   &.is-open {
     transform: scaleY(1);
+    overflow-y: scroll;
   }
 }
 
--- a/lib/ReactViews/Generic/dropdown.scss
+++ b/lib/ReactViews/Generic/dropdown.scss
@@ -21,7 +21,7 @@
 
   &.is-open {
     transform: scaleY(1);
-    overflow-y: scroll;
+    overflow-y: auto;
   }
 }
 
--- a/lib/ReactViews/Generic/dropdown.scss
+++ b/lib/ReactViews/Generic/dropdown.scss
@@ -18,7 +18,8 @@
   box-sizing: border-box;
   background: #fff;
   border-radius: $radius-small;
-
+  border: 1px solid rgba($grey, 0.4);
+  box-shadow: 0px 3px 15px rgba($dark, 0.15);
   &.is-open {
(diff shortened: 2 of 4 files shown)


lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
```diff
@@ -150,6 +150,7 @@ const AddData = createReactClass({
               selectOption={this.selectLocalOption}
               matchWidth={true}
               theme={dropdownTheme}
+              height="150px"
```
KeyboardSounds: Should this be part of the theme, rather than a separate prop? (sorry, nitpicking. it probably doesn't matter)

lib/ReactViews/Generic/dropdown.scss
```diff
@@ -21,6 +21,7 @@
 
   &.is-open {
     transform: scaleY(1);
+    overflow-y: scroll;
```
KeyboardSounds: Is there any particular reason you picked `overflow-y: scroll` over `overflow-y: auto`?

KeyboardSounds:
Totally optional, but if you have time it'd be really great if you could do a quick search through the repo for other dropdowns and set heights for those too. For example, the download dropdown button in FeatureInfoDownload.jsx.

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
rowanwins: I hadn't thought about that to be honest, I'm not sure what the rationale is between stuff that goes into the theme prop vs what gets passed down as its own prop.

lib/ReactViews/Generic/dropdown.scss
rowanwins: Short answer no. 

In both chrome & firefox on mac the scrollbar disappears when not hovered over the element which is the same behaviour as auto I think. IE11 might be more finicky and apply a scrollbar even if it's not required.

rowanwins:
Thanks for the review @KeyboardSounds . 
> Totally optional, but if you have time it'd be really great if you could do a quick search through the repo for other dropdowns and set heights for those too. For example, the download dropdown button in FeatureInfoDownload.jsx.

I did do a quick visual check for other uses of dropdowns to ensure no adverse affects. My thought was that you'd only apply height where required, hence why I've set the default height to `auto`.

lib/ReactViews/Generic/dropdown.scss
chloeleichen: should use auto, no need to constantly have a scroll bar

rowanwins: Updated

soyarsauce:
I think we should allow for the dropdown to not be constrained to the fixed height if there's enough space to view more items at a time, maybe @philipgrimmett can chime in on whether it's actually better?

The path to doing this would be a bit tricky though, involving resizing elements to fill its containers, and calculating the height. There should be a helper component to do this ready from another PR, I can repurpose to also measure heights and you can copy the component into your pr here.

soyarsauce:
Rowan if it's useful you can pull in this to use:  ~https://github.com/TerriaJS/terriajs/blob/64ce7f62421ea210aaffbfcb7a6189dc6d7741f0/lib/ReactViews/measureComponent.js~
Sorry renamed it!
https://github.com/TerriaJS/terriajs/blob/1ef62f8c278b64f8e33f88a66f472d49ab4d3bd0/lib/ReactViews/measureElement.js

rowanwins:
Thanks for the suggestion @soyarsauce - I had wondered about that originally but wasn't sure how feasible it would be without a massive investment of time, fingers crossed with what you've provided it should be pretty feasible!

rowanwins:
So I managed to hook up @soyarsauce 's [HOC](https://reactjs.org/docs/higher-order-components.html) locally for getting the size of a parent component. 
However, at least in the case of the 'Add Data' panel, it would still need a fair bit of extra work to work out how much space is actually available for the dropdown to fill because of how the elements are structured. I'm sure it's doable but I'm wondering if it's more effort than it's worth in this case.

chloeleichen:
It is ok to give dropdown fixed height, check with @philipgrimmett on an appropriate value and it should vary based on device size. After all this is how the default browser dropdown works and I don't think it's worth our time to overcomplicate things

rowanwins:
@chloeleichen - currently users don't have the 'Add data' panel available if they are on devices with small screens.

@philipgrimmett - the current config for the 'Add Data' panel is set for 150px, it can be viewed here 
http://ci.terria.io/dropdown-scroll/

philipgrimmett:
150px seems fine, but don't default browser select dropdowns normally have either a very light 1px border or shadow? https://developer.mozilla.org/en-US/docs/Learn/HTML/Forms/The_native_form_widgets#Drop-down_content

soyarsauce (dismissed):
@rowanwins let's add a few fixed heights & see Phil's comment about styling a bit nicer, then we'll get it in

rowanwins:
Added a outline & shadow
Dropdown

soyarsauce (approved):

PR #3400, opened 2019-05-17
Add coverage reports

na9da:
Adds coverage reporting to `mobx` branch.


buildprocess/configureWebpack.js
```diff
@@ -107,6 +107,15 @@ function configureWebpack(terriaJSBasePath, config, devMode, hot, MiniCssExtract
         ]
     });
 
+    config.module.rules.push({
+      test: /\.(ts|js)x?$/,
+      include: [path.resolve(terriaJSBasePath, "lib")],
+      use: {
+        loader: "istanbul-instrumenter-loader"
```
kring: `configureWebpack` is used to build TerriaMap, too, so won't this add instrumentation to apps, too? Or is that not a problem?

kring: Actually at the moment it's breaking the app build completely. See the Deployment log.

na9da: I have made the configuration package local. Currently, it only reports classes referenced via the tests!

kring:
Looks good and I'm merging it. Just wondering, though: can/should we add the istanbul reporter to karma-saucelabs.conf.js too so that we get coverage results from Travis?

Issue #3362, opened 2019-05-06
User help tutorial - bring back to life from SoE


AnaBelgun:
State of Environment (https://github.com/TerriaJS/soe) instance had a help guide for users/prompts which is not available anymore in the app. 
This ticket is to:
- find the guide development in the https://github.com/TerriaJS/soe code base
- improve it and implement it in TerriaJS master to be available as part of the story feature

soyarsauce:
At least for SoE, the broken SoE help tutorial will be fixed if this gets merged:
https://github.com/TerriaJS/soe/pull/8

@philipgrimmett wants to tidy some stuff up but I think that can happen concurrently to the help-currently-unavailable-state of SoE

AnaBelgun:
we need to re-release

AnaBelgun:
release approval pending; issue resolved

PR #3401, opened 2019-05-17
Fix HelpMenuPanel open state props

soyarsauce:


chloeleichen:
@soyarsauce I don't see `HelpMenuPanel` being used anywhere, is there a specific instance that i should look at?

soyarsauce:
@chloeleichen https://github.com/TerriaJS/soe/blob/9b900f9d4102ea2ef38b793db4214f8f9247531e/lib/Views/UserInterface.jsx 👍

kring:
@soyarsauce this is to fix the SoE help menus not working, right? It would be good to mention that in the issue description. Can we even say it fixes #3362?

soyarsauce:
Yes, this alone would actually address the "bring back to life" aspect of #3362 but the other parts need a bit more defining, separately https://github.com/TerriaJS/soe/pull/7 should fix SoE help

chloeleichen (approved):
Thanks @soyarsauce

PR #3411, opened 2019-05-21
fix color range not updating from share data

### LINKED ISSUE #3404: Story scenes are not capturing WMS colour scale range
### Steps to reproduce
1. Add _Annual climatology of daily exposure - Global Horizontal Exposure_ to the map (http://ci.terria.io/master/#share=s-h70gpchtxlo1Wv9cBw56E4iRf3G)
2. Capture scene with default colour scale range (Max 23.5, Min 11)
3. Set colour scale range to Max 25, Min 21.5
4. Capture scene
5. Play story, noting that scale range does not change on scene transition.

Ping @chloeleichen

chloeleichen:
Fixes: https://github.com/TerriaJS/terriajs/issues/3404

For the color range field to be updated from share data, it needs to be a controlled element, hence not `defaultValue` but `value` (surprising react was not giving us a warning), and we need to copy props to state when props change.

However, when the parent component is passing in `item`, any change in `item.colorMax` unfortunately does not trigger `componentWillMount`. Therefore we add the color range values directly as props to `ColorScaleRangeSection` so we can correctly copy the latest props into state.

Since we are using mixin and pre ess6 syntax, we are now using the `UNSAFE_componentWillMount` and `UNSAFE_componentWillReceiveProps`. After we update to es6 syntax we should use `static getDerivedStateFromProps()` and do the update in one go. 

Update:
The item refresh from the shared data should not rely on Ui updates, therefore max and min color range is subscribed and refresh happens when there's a change in either

### DIFF
--- a/lib/Models/WebMapServiceCatalogItem.js
+++ b/lib/Models/WebMapServiceCatalogItem.js
@@ -333,17 +333,18 @@ var WebMapServiceCatalogItem = function(terria) {
     "_allLayersInRawMetadata"
   ]);
 
-  knockout.defineProperty(this, "_colorRange", {
+  knockout.defineProperty(this, "_colorScaleRange", {
     get: function() {
-      return this.colorScaleMaximum + "  " + this.colorScaleMinimum;
+      return [this.colorScaleMinimum, this.colorScaleMaximum];
     }
   });
 
-  knockout.getObservable(this, "_colorRange").subscribe(function(_colorRange) {
+  this._refreshInProgress = undefined;
+  knockout.getObservable(this, "_colorScaleRange").subscribe(function() {
     if (this.isEnabled && !this._refreshInProgress) {
       this._refreshInProgress = runLater(() => {
         this.refresh();
-        this._refreshInProgress = null;
+        this._refreshInProgress = undefined;
       });
     }
   }, this);
--- a/lib/Models/WebMapServiceCatalogItem.js
+++ b/lib/Models/WebMapServiceCatalogItem.js
@@ -333,15 +333,15 @@ var WebMapServiceCatalogItem = function(terria) {
     "_allLayersInRawMetadata"
   ]);
 
-  knockout.defineProperty(this, "colorRange", {
+  knockout.defineProperty(this, "_colorRange", {
     get: function() {
       return this.colorScaleMaximum + "  " + this.colorScaleMinimum;
     }
   });
 
   knockout
-    .getObservable(this, "colorRange")
-    .subscribe(function(colorScaleMaximum) {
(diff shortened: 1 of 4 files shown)


lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -332,6 +333,21 @@ var WebMapServiceCatalogItem = function(terria) {
     "_allLayersInRawMetadata"
   ]);
 
+  knockout.defineProperty(this, "_colorRange", {
+    get: function() {
+      return this.colorScaleMaximum + "  " + this.colorScaleMinimum;
```
kring: How about putting them in an array rather than concatenating strings?

```suggestion
      return [this.colorScaleMinimum, this.colorScaleMaximum];
```

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -332,6 +333,21 @@ var WebMapServiceCatalogItem = function(terria) {
     "_allLayersInRawMetadata"
   ]);
 
+  knockout.defineProperty(this, "_colorRange", {
```
kring: More precise name: `_colorScaleRange`.

lib/Models/WebMapServiceCatalogItem.js
```diff
@@ -332,6 +333,21 @@ var WebMapServiceCatalogItem = function(terria) {
...
+    get: function() {
+      return this.colorScaleMaximum + "  " + this.colorScaleMinimum;
+    }
+  });
+
+  knockout.getObservable(this, "_colorRange").subscribe(function(_colorRange) {
```
kring: This isn't ideal because every WMS catalog ever, even ones that are just in the catalog, will subscribe to this property. But I guess fixing this is more trouble than it's worth. In the new model architecture we won't have to subscribe at all, it'll just work.

kring:
Super minor comments, looks good!

lib/Models/WebMapServiceCatalogItem.js
chloeleichen: What would be the benefit of putting them in an array than concat into strings?

kring: Oh sorry it's pretty much just a nitpick. But the (small) advantages are:
* it's slightly faster (not that anyone cares here)
* the value is actually usable. We could theoretically get that property and use the elements of the array for something. The string is not likely to be useful to anyone. You'd have to parse it to get the individual values, and even if you just wanted to display them together it's unlikely that the formatting used here would be appropriate for that.

kring (approved):

PR #3413, opened 2019-05-21
Set `uniqueId` of WPS result items to GUID.

### LINKED ISSUE #3412: Running (WPS) analysis tool multiple times breaks the workbench
Steps to reproduce bug:
 1) Visit [GEOGLAM Map](http://map.geo-rapp.org/)
 2) Open 'Analysis tools'
 3) Select any point & run analysis, then wait for new catalog items to added.
 4) Reopen 'Analysis tools' and re-run the analysis, wait for catalog items to be added.

You can see multiple catalog items with the same name and some of them cannot be removed.

This happens because different catalog items have the same `name` (which comes from the server), hence the same `uniqueId`. This `uniqueId` is also used as a react key. Having duplicate keys messes with the react state.

na9da:
Fixes #3412

### DIFF
--- a/lib/Models/WebProcessingServiceCatalogItem.js
+++ b/lib/Models/WebProcessingServiceCatalogItem.js
@@ -192,11 +192,9 @@ WebProcessingServiceCatalogItem.prototype._load = function() {
               that.terria
             );
             catalogItem.updateFromJson(jsonCatalogItem);
-            // If the server returns duplicate names, we end up having duplicate uniqueIds.
-            // To avoid this we force uniqueId to a GUID.
-            Object.defineProperty(catalogItem, "uniqueId", {
-              value: createGuid()
-            });
+            // The server might return duplicate names and IDs derived from it will not be unique.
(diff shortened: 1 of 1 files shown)


lib/Models/WebProcessingServiceCatalogItem.js
```diff
@@ -191,6 +192,11 @@ WebProcessingServiceCatalogItem.prototype._load = function() {
               that.terria
             );
             catalogItem.updateFromJson(jsonCatalogItem);
+            // If the server returns duplicate names, we end up having duplicate uniqueIds.
+            // To avoid this we force uniqueId to a GUID.
+            Object.defineProperty(catalogItem, "uniqueId", {
```
kring: You should be able to just set the `id` property rather than clobbering the `uniqueId` property.

na9da: Ah yes! I didn't notice that bit :man_facepalming:

kring (approved):

PR #3415, opened 2019-05-21
icon size regression

chloeleichen:
we had some icons that are too big after latest svg icon update. This Pr fixed those. 
eg: previously:
image
now:
image


kring (approved):

soyarsauce:
Looks good

PR #3439, opened 2019-05-28
Generate attributions section of docs.

stevage:
Ok, I went a different way with this, using yarn's included `licenses generate-disclaimer` feature.


reginapramesti (changes requested):
https://github.com/TerriaJS/terriajs/blob/a3027743cef329b1249d8cfab6d7b8617ca1db1e/gulpfile.js#L143

Not a major issue, but since there is a global import of spawnSync at the beginning of this file maybe this local import could be removed for consistency? Or alternatively the global import could be made local to the `code-attribution` task.

reginapramesti (approved):

PR #3445, opened 2019-05-28
Move prettier paths to .prettierignore file

tephenavies:
Use a more standard way of including only certain paths for prettier. This allows using prettier IDE features like format on save without prettier formatting other files that shouldn't have prettier applied (like CHANGES.md or gulpfile.js etc.)

Prettier commands should work identically, and the pre-commit hook still works.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,18 @@
+# Ignore everything
+*
+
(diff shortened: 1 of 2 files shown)


soyarsauce (approved):
💯✨

PR #3449, opened 2019-06-03
Add prettier resolution docs

soyarsauce:
Should we add something like this?


soyarsauce:
Adds instructions from https://github.com/TerriaJS/terriajs/pull/3361 with only change being the intro & an additional package-lock/yarn-lock step

kring:
The instructions look good, but we need to think about this a bit. The majority of people impacted by this will not have ever set up a TerriaJS development environment. They'll have just forked TerriaMap and made some customizations to it. They'll be surprised and annoyed when there are suddenly heaps of conflicts in their project.

So perhaps the instructions should go somewhere near here instead: https://docs.terria.io/guide/getting-started/#keeping-up-with-updates

But people are still going to miss it. We can get the word out Twitter and Gitter and Medium and wherever else, that'll help some. We could defer prettier'ing existing .js/.jsx files in master, and only do it in mobx. When mobx lands, users will need to make significant changes to their index.js files anyway, so conflicts are inevitable. But managing both major changes (prettier and mobx) at the same time might just make it harder.

I'm mostly thinking out loud and interested in any ideas, but I think what I'm coming around to now is to move these instructions to the getting started page and point people to the instructions through all the channels we have for reaching our users, as soon as it all hits master.

kring (approved):

PR #3453, opened 2019-06-05
Add treat404AsError property to region mapped CSV item

rowanwins:
Currently when a `RegionMapping` provider crosses the anti-meridian (for example countries in the south pacific) the bounding box property the regionMapped provided means that tiles are requested outside the real area of interest, this results in many 404 errors, which in turn [triggers an error on the ImageryLayerCatalogItem](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/ImageryLayerCatalogItem.js#L1078-L1113) because the associated catalogItem (a `CsvCatalogItem`) does not have a `treat404AsError=false`.

### DIFF
--- a/lib/Models/RegionMapping.js
+++ b/lib/Models/RegionMapping.js
@@ -904,6 +904,11 @@ function createNewRegionImageryLayer(
   var regionImageryProvider;
   var layer;
 
+  // Handle the case where a region mapped dataset crosses the antimeridian
+  // and we get 404's retrieving tiles, this prevents the catalog item
+  // from crashing
+  catalogItem.treat404AsError = false;
+
   switch (regionDetail.regionProvider.serverType) {
     case "MVT":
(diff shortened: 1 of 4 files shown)


kring (changes requested):
The code change looks good, but this PR also adds the SPC regions. We don't want to do that, do we?

rowanwins:
Yeah I did wonder about that, the reason I added them in was that so someone might be able to see a troublesome tileset.

kring:
I'm ok with adding the new region types to our common regionMapping.json, but they can't use generic aliases like "COUNTRY". That alias is already assigned to a different region type, for one thing.

kring:
I saw the 404 problem this morning with the SED_2018 region type, so finding a troublesome tileset isn't a problem. :)

kring (approved):

PR #3494, opened 2019-06-12
1757 new routing

### LINKED ISSUE #1757: [EPIC] SEO: Make datasets findable on search engines
At the moment if you search for some specific dataset that you are interested in using some internet search engine, no links to that dataset on any TerriaJS instances will be found unless maybe if somebody has made a share URL and put that into a static document which was then crawled by a search engine. This is a shame because this would probably lead to a lot more users finding useful data.

The problem is that search engine crawlers can probably not "run" the TerriaJS instance JavaScript in order to meaningfully extract the data.

We should implement some feature that makes it possible for search engines to deep link to the layers provided on TerriaJS instances. 

This might take a number of forms:
- A dynamically generated page that lists all datasets with deep links (share URLs?) into the TerriaJS instance that a search engine can find and crawl.
- Some form of sitemap? I know Google can accept that, but even if we made one with the share URL's of all the available layers I don't think it would be able to extract the name/metadata.
- Detect when the client is a crawler and then present a "HTML only" version of the catalogue with all layers available together with their metadata.
- Some other as-yet-unthought-of clever thing.

soyarsauce:
Adds routing such that catalog navigation can be traversed with browser history, and subsequently telling terriajshow to resolve URLs for a given catalog link

A demo of this in action with the 3 separate versions on one deployment is at (VPN required)
http://152.83.140.90:3001/catalog/54a553b4
>, if you click into any of the catalog items and hit refresh you’ll see the metadata directly from the initial HTML served up, before javascript takes over and makes the rest of the UI interactive

Sitemap now autogenerated (after specifying appBaseUrl under config.json's `parameters` object)
http://152.83.140.90:3001/sitemap.xml

Attempts to resolve https://github.com/TerriaJS/terriajs/issues/1757

Also see:
https://github.com/TerriaJS/terriajs-server/pull/106
https://github.com/TerriaJS/TerriaMap/pull/342


lib/ReactViews/DataCatalog/DataCatalogItem.jsx
```diff
@@ -114,4 +119,8 @@ const DataCatalogItem = createReactClass({
   }
 });
 
-module.exports = DataCatalogItem;
+module.exports = {
+  default: withRouter(DataCatalogItem),
+  DataCatalogItem: withRouter(DataCatalogItem),
```
kring: Probably not a huge problem, but if I'm not mistaken calling `withRouter` twice will create two separate but identical functions. Better to avoid that I think.

lib/ReactViews/DataCatalog/DataCatalogMember.jsx
```diff
@@ -4,7 +4,7 @@ import React from "react";
 import createReactClass from "create-react-class";
 import PropTypes from "prop-types";
 import ObserveModelMixin from "../ObserveModelMixin";
-import DataCatalogItem from "./DataCatalogItem.jsx";
+import { DataCatalogItem } from "./DataCatalogItem.jsx";
```
kring: Not sure why you had to change this, since we're also exporting it as default. Maybe use ES exports for DataCatalogItem.jsx instead of mixing CJS and ES?

lib/ReactViews/Preview/DataPreviewMap.jsx
```diff
@@ -102,6 +102,10 @@ const DataPreviewMap = createReactClass({
 
   /* eslint-disable-next-line camelcase */
   UNSAFE_componentWillReceiveProps(newProps) {
+    // if (this.props.previewedCatalogItem !== newProps.previewedCatalogItem) {
```
kring: Remove commented out code.

lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
```diff
@@ -60,6 +68,7 @@ const StandardUserInterface = createReactClass({
   /* eslint-disable-next-line camelcase */
   UNSAFE_componentWillMount() {
     const that = this;
+    // this.props.viewState.history = this.props.history;
```
kring: Here too.

kring (changes requested):
This is really cool! Mostly nitpicky comments below, but there are some bigger-picture things to discuss as well:

1. It has long been possible to deploy TerriaMap to a static web server, and we don't want to lose that ability. Does that mean providing an option to turn off the routing? Or making it (optionally?) use the hash portion of the URL instead of the path portion?
2. The need to specify an `appBaseUrl` isn't great, and is a breaking change. The Investor Map staging site, for instance, is available at both https://inv-staging.nm.terria.io/ and https://inv-staging.nm.terria.io/investormap/. NEII has something similar happening. Why do we need to know the base URL? I thought it was for the sitemap, but a quick google seems to indicate sitemaps can use relative URLs too.
3. As mentioned briefly in the comments below, I don't want to see the routing become yet another "source of truth". We need to be clear about where the state (whether catalog is open, which item is selected) comes from. Maybe you are, but it wasn't clear to me while looking at the code.
4. The `mobx` branch 😱. There will be conflicts. Some of this will be different there.

soyarsauce:
Thanks for bringing these up @kring 
> * It has long been possible to deploy TerriaMap to a static web server, and we don't want to lose that ability. Does that mean providing an option to turn off the routing? Or making it (optionally?) use the hash portion of the URL instead of the path portion?

We have a couple of options. In fact with the PRs as they are, it is still possible to deploy this new configuration of TerriaMap to a static web server & have it work for basic catalogs that don't query a server for layers - it gets tricky once we start to support that too. I see a few options:
- extend TerriaMap's new `generate-init-routes.js` rely on getting a response from the remote services, so that it can query its layers, so we know what other routes to serve up an index.html - but this might be more effort than its worth at the moment given the upcoming changes + seeing how it plays out with magda, & even then it is still kind of separate to the "backendless", "static TerriaMap"
- depending on how much control a given user has over their "static web server", we simply provide configuration examples for rewriting URLs on common static web servers (apache+/.htaccess, nginx, other)
- as you said, do some sort of detection or build env variable for routing on browserhistory vs hash, but this is the least ideal situation as we want proper-path-URLs for discoverability (through crawling)



> * The need to specify an `appBaseUrl` isn't great, and is a breaking change. The Investor Map staging site, for instance, is available at both https://inv-staging.nm.terria.io/ and https://inv-staging.nm.terria.io/investormap/. NEII has something similar happening. Why do we need to know the base URL? I thought it was for the sitemap, but a quick google seems to indicate sitemaps can use relative URLs too.

The `appBaseUrl` is actually optional & not required - and serves the purpose of
- the sitemap as you say
- cross-domain-canonical-references
Sitemaps can in fact be relative, I thought it wouldn't hurt to help crawlers determine which version of a duplicate deployment to use as the canonical reference. I thought about the case of same deployment to 'different hosts' but not the 'same host+different' path - are those two Investor Map staging-deployments the exact same? What would the benefit of that be?

The new default added to `devserverconfig.json` for `baseHref` of "`/`" is if we served up a TerriaMap that wasn't on root (e.g. on `/renewables/`), without knowing the baseHref we can't tell the browser the correct place to load `build/TerriaMap.js` from, and any subsequent relative links (e.g. `/renewables/catalog/xyz`)



> * As mentioned briefly in the comments below, I don't want to see the routing become yet another "source of truth". We need to be clear about where the state (whether catalog is open, which item is selected) comes from. Maybe you are, but it wasn't clear to me while looking at the code.

I haven't had much success without making the URL the absolute truth for these sort of things, & I tried to 'subscribe' to routing changes via `RoutingListener.jsx` & update viewState that way, there's certainly more room to make it clearer & I'll revisit that 



> * The `mobx` branch 😱. There will be conflicts. Some of this will be different there.

......yes 😵

soyarsauce:
At present time this is postponed indefinitely for pre-mobx(mastere), and will be fully re-implemented in mobx.

soyarsauce:
(Barring merging it with even safer defaults of everything-off-by-default)

tephenavies:
Won't be merged/implemented in version 7. PR to implement functionality in version 8 is #4961.

PR #3495, opened 2019-06-13
Adding ArcGisMapServerCatalogItem.

na9da:


lib/Models/getToken.js
```diff
@@ -31,7 +31,7 @@ function getToken(terria, tokenUrl, url) {
 
       return tokenResponse.token;
     })
-    .otherwise(() => {
+    .catch(() => {
```
na9da, line 34: Thanks! Not sure how that went through, maybe I forgot to commit the file.

tephenavies: I think it was needed because of new changes on `mobx` that I merged in.

lib/Models/ArcGisMapServerCatalogItem.ts
```diff
@@ -0,0 +1,509 @@
+import lodash from "lodash";
```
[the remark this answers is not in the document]
tephenavies: Whoops. I meant to send that comment much earlier. I've found that importing with `import func from "lodash-es/func"` seems to be the best way to use a few functions of lodash without adding the full lodash to chunks that use it.

PR #3498, opened 2019-06-13
Fix feedback

kring:
It was broken by some of the recent sharing changes.

Also enabled feedback on ci.terria.io so we can more easily test this, and so that we're less likely to miss it in the future.


na9da (approved):
Looks good to me!

PR #3507, opened 2019-06-17
Wrap map click coordinates in leaflet

### LINKED ISSUE #3505: Clicking on map in 2D mode across anti-meridian can miss features
Working with a region mapped csv whose regions cross the anti-meridian we're noticing some issues picking features when in 2D mode when zoomed out.

If I click on the left (east) of the anti-meridian I can successfully pick features
Working

If I click on the right (west) of the anti-meridian things I'm unable to pick features
AntiMeridian

If I zoom in a few zoom levels and pan across the anti-meridian then clicking works again.

Clicking appears to work fine in 3D mode.

rowanwins:
This PR adds a check on a leaflet map click event to make sure the `latlng` is between -180 and +180., if it's not we use [latlng.wrap()](https://leafletjs.com/reference-1.3.4.html#latlng-wrap) to wrap coordinates to within that range.

This prevents errors when trying to pick features on a zoomed out 2D map that cross the anti-meridian.
Working

Resolves #3505

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### v7.6.3
+
+* Fixed a bug with picking features that cross the anti-meridian in 2D mode 
+
 ### v7.6.2
 
 * Fixed a bug that made some input boxes unreadable in some web browsers.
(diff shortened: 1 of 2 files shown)


soyarsauce:
Nice 💯

soyarsauce (approved):

soyarsauce:
For reference, reproduce steps @ https://github.com/TerriaJS/terriajs/issues/3505#issuecomment-503838562

Issue #3399, opened 2019-05-16
DragDrop CSVs appear as 'unnamed item'


philipgrimmett:
**Steps to reproduce**
1. Navigate to http://ci.terria.io/master/
2. Drag CSV onto map (eg. "Population Estimates 2017-18.csv")
3. See "unnamed item" in workbench

rowanwins:
The same thing occurs with geojson files that get dropped on.

If this were to be fixed @philipgrimmett do you think we should display the filename with or without the extension?

![Filename](https://user-images.githubusercontent.com/6735870/60552745-6f8fb680-9d73-11e9-9a47-1c4e82b092cd.png)

philipgrimmett:
@rowanwins Hmm. Not sure. My feeling is that extensions should be shown where the user might reasonably expect to re-export/download their file after drag-dropping it into Terria. Where Terria is the endpoint for that data, I'd be less concerned about displaying file extensions.

That said: I think we should retain them for the time being, if only for the edge case where a user drags the same dataset in two different formats onto the map and has to distinguish between the two (eg. _publicParks.csv_ and _publicParks.geojson_)

PR #3518, opened 2019-06-19
Add warning in catalogue sharing

### LINKED ISSUE #1367: Missing list of catalog items that will not be shared
I find it very easy to miss the warning that you actually will not be sharing your data here:

image

Can we make it stand out more, perhaps show it in red? I'd also suggest adding some space before the dot point.

reginapramesti:
Fixes #1367

### DIFF
--- a/lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
+++ b/lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
@@ -124,3 +124,23 @@
 .print-view-loader {
   fill: $text-light;
 }
+
+.warning {
+  margin: $padding 0;
+  border-radius: $radius-small;
+  background: #feb938;
+  color: #552800;
+  padding: $padding;
+
+  ul {
+    padding-top: $padding;
+    padding-inline-start: 0;
+    list-style: none;
+  }
+
+  a {
+    color: #552800 !important;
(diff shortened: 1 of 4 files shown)


reginapramesti:
Some things to note:

* The list of CSVs that won't be shared currently just lists a bunch of Unnamed Items. This should in theory be fixed by fixing #3399
* Clicking `Remove` from the workbench doesn't get rid of the warning as it doesn't actually remove the data from the catalogue but simply removes it from the map. Though removing it from the `My Data` tab would get rid of the warning

tephenavies:
Please run `gulp lint` and fix the lint issues.

lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
```diff
@@ -124,3 +124,23 @@
...
+  padding: $padding;
+
+  ul {
+    padding-top: $padding;
+    padding-inline-start: 0;
+    list-style: none;
+  }
+
+  a {
+    color: #552800 !important;
```
tephenavies: Why do these have to be `!important`? Is there any way to avoid it?

reginapramesti:
@steve9164 could you have a look at it now please?

tephenavies (approved):

PR #3519, opened 2019-06-19
Model architecture: Nested traits are models, etc

kring:
* Added `AsyncLoader`, which encapsulates the logic that was repeated for `loadMapItems`, `loadMetadata`, `loadReference`, etc.
* Renamed model's `id` to `uniqueId`.
* `uniqueId` is now optional. Not all models need IDs, only the ones that are in `terria.models` do.
* Nested traits (e.g. `objectTrait` and `objectArrayTrait`) are now represented in the model layer as models. For example, the type of the `legend` property on WMSCatalogItem is now `Model<LegendTraits>` instead of `ModelPropertiesFromTraits<LegendTraits>`. That means you can get/set traits from individual layers:
```ts
const wms: WebMapServiceCatalogItem = ...;
wms.legend.setTrait(CommonStrata.user, "url", "http://example.com/legend");
```
* Started hooking up sharing, but this is WIP so far.


kring:
This branch of terriajs requires the `mobx-nested-traits-are-models` branch of TerriaMap.

PR #3527, opened 2019-06-24
Add support for SDMX v2.1 to SdmxJsonCatalogItem

rowanwins:
This PR adds support for SDMX v2.1 spec

Significant things include
- [x] Add two new request types to the `Core` directory which use specific request `Accept` headers for retrieving data
  - It appears that if you use the standard `Accept: 'application/json'` you get a partial response from the server but not the full response
- [x] Many of the response json objects in v2.1 have an internationalised structure like
````
dimension: {
    name: {
         en: 'something'  <--- note the nested 'en' key
    }
}
// compared to v2.0
dimension: {
      name: 'something'
}
````
  - It might be worth adding a property to the `SdmxJsonCatalogItem` to specify what language key to use. At the moment the assumption is we take the `en` key

One thing I'd like feedback on is that there is some specific handling for v2.1 services compared to v2.0 services. I'm not sure if there is a preference for which way around I should be making my equality statements. 
````
  this._getData = this.sdmxVersionNumber === 2.1 ? loadSdmxDataJson : loadJson;
  // vs 
  this._getData = this.sdmxVersionNumber === 2.0 ? loadJson : loadSdmxDataJson;
  // vs
  this._getData = this.sdmxVersionNumber !== 2.0 ? loadSdmxDataJson : loadJson;
````

I've not added any tests yet but happy to take advice on the level required.

### CODE REVIEW

**Comment by rowanwins** [comment]
In terms of the test that I modified
[`is less than 2000 characters when serialised to JSON then URLEncoded`](https://github.com/TerriaJS/terriajs/pull/3527/files#diff-90606368cfdd1711bde0b1398a3b8871R693)
Prior to the change the url length was 1988, and it's now 2225. I presume this is because I added a number of additional keys to the `SdmxJsonCatalogItem` such as `sdmxVersionNumber` and `measureDimensionId`

File: lib/Core/loadSdmxDataJson.js [hunk 1]
```diff
@@ -0,0 +1,13 @@
+const Resource = require("terriajs-cesium/Source/Core/Resource");
+
+function loadSdmxDataJson(urlOrResource) {
```
kring, line 3: Because this is so specific to SDMX and is only used it one place, it'd be fine to just use this code inline where it's needed rather than creating a separate function. Really even the `loadJson` function exists primarily because a similar function used to exist in Cesium and it was deprecated, and it was easier to add our own version of it than to change all our call sites. No need to go back and change this unless you feel like it, just an FYI.

kring (approved):
Looks good, thanks @rowanwins!

kring:
Can you please update CHANGES.md in master?

PR #3544, opened 2019-06-26
Story: v1.1 - disable autosuggest on form fields

chloeleichen:
https://github.com/TerriaJS/terriajs/issues/3536


rowanwins (approved):
Too easy!

Issue #3533, opened 2019-06-25
Story: v1.1 - Feedback button covers arrow in story play mode 


AnaBelgun:
When in mobile mode, the feedback button covers the arrow when story in play mode

example here: https://www.nationalmap.gov.au/investormap/#share=s-5qPAfERuEQ1UKsWOw0ggcGu376k

![image](https://user-images.githubusercontent.com/18452161/60068994-7ce5e900-9754-11e9-81a9-f48926798696.png)

PR #3547, opened 2019-06-27
allow story panel to be ordered like other ui components

chloeleichen:
gif

Part of solution addressing issue #3533
Another Investor-map specific pr is here:


soyarsauce:
We should also toggle top elements on focus/blur events? (We have our very own keyboard user @na9da!)

chloeleichen:
@soyarsauce good point but that should be a different issue (focus management for ui components) as most of our ui needs some work on that front.

soyarsauce (approved):

PR #3555, opened 2019-06-28
Add quality slider

### LINKED ISSUE #3280: Add a performance/quality slider to Map panel
We should add a Performance/Quality slider to the Terria map panel like similar to the screenshot mockup:

Screen Shot 2019-04-08 at 10 13 46

On the left everything is default settings, then for each of the sliders steps towards quality we enable first this:

```
terria.cesium.viewer.scene.globe.maximumScreenSpaceError = 4.0 / 3.0;
```

Then this:

```
terria.cesium.viewer.resolutionScale = devicePixelRatio
```

soyarsauce:
Addresses https://github.com/TerriaJS/terriajs/issues/3280

Resolves: https://github.com/TerriaJS/terriajs/issues/3280


lib/ViewModels/TerriaViewer.js
```diff
@@ -303,6 +309,29 @@ function changeFogSettings(viewer) {
...
+  if (defined(viewer.terria.cesium)) {
+    const terria = viewer.terria;
+
+    // 0 - defaults
+    // 1 - devicePixelRatio as resolutionScale
+    // 2 - increase maximumScreenSpaceError
+    switch (viewer.terria.quality) {
+      case 1:
+        terria.cesium.viewer.resolutionScale = devicePixelRatio;
+        terria.cesium.viewer.scene.globe.maximumScreenSpaceError = 1;
```
kring: The default is 2, not 1.

lib/ViewModels/TerriaViewer.js
```diff
@@ -303,6 +309,29 @@ function changeFogSettings(viewer) {
...
+function changeQuality(viewer) {
+  if (defined(viewer.terria.cesium)) {
+    const terria = viewer.terria;
+
+    // 0 - defaults
+    // 1 - devicePixelRatio as resolutionScale
+    // 2 - increase maximumScreenSpaceError
+    switch (viewer.terria.quality) {
+      case 1:
+        terria.cesium.viewer.resolutionScale = devicePixelRatio;
```
kring: After changing `resolutionScale`, you need to call `terria.cesium.viewer.resize()` for it to take effect.

lib/ViewModels/TerriaViewer.js
```diff
@@ -303,6 +309,29 @@ function changeFogSettings(viewer) {
...
+        terria.cesium.viewer.scene.globe.maximumScreenSpaceError = 1;
+        break;
+      case 2:
+        terria.cesium.viewer.scene.globe.maximumScreenSpaceError = 4.0 / 3.0;
```
kring: Set `resolutionScale` here too, otherwise it won't happen if the user clicked on the right directly rather than dragging.

kring (changes requested):
At the risk of doing armchair UX, I think this should be a dropdown instead of a slider. The slider implies continuity that just doesn't exist.

Or maybe even better is to separate the settings. We can have a slider for `maximumScreenSpaceError`, which really can be continuous. Call it "Raster Map Quality" and let it vary between 1.0 and 3.0 or so. And the other can be a checkbox, only shown when `devicePixelRatio` is greater than 1, to "Use device native resolution" or something like that?

By the way there is a similar setting for Cesium 3D Tiles. The default is larger there, 16.0 versus 2.0 if I remember correctly.

But if we want to keep it simple and have three settings, a dropdown seems a lot better to me. Actually, it should show only two settings when `devicePixelRatio=1`.

AnaBelgun:
@philipgrimmett to refine the design (and decide between sliver vs drop down)

lib/ViewModels/TerriaViewer.js
soyarsauce: Thanks resolved

lib/ViewModels/TerriaViewer.js
soyarsauce: Done for any change now

soyarsauce:
https://github.com/TerriaJS/terriajs/pull/3555/commits/344038d27108b1530a75b5676bda80d873bc9ce4
>Update quality slider to new cesium behaviour

>New cesium behaviour interprets the resolutionScale
differently, and we also update 3d-tiles if they
don't have any overrides specified in its definition
json

soyarsauce:
Actually @philipgrimmett I thought about it a little more and agree w/ what Kevin is saying - what do you think with at the very least splitting it out to a checkbox + slider, checkbox for "using native device resolution", slider for "Raster Map Quality"?

soyarsauce:
06973a4
>Split "image optimisation" options independently

>Lets user toggle native resolution vs raster quality
without being locked into "slider presets"

philipgrimmett:
> Actually @philipgrimmett I thought about it a little more and agree w/ what Kevin is saying - what do you think with at the very least splitting it out to a checkbox + slider, checkbox for "using native device resolution", slider for "Raster Map Quality"?

@soyarsauce sure, sounds like a good middle ground. When 'native device resolution' is enabled, should the slider be disabled? If not, what does it mean to select 'performance'?

soyarsauce:
@philipgrimmett they are independent of each other, dragging towards "performance" will make parts of the map sharper/"crisper"

kring (approved):

PR #3563, opened 2019-07-03
fix #3511, show prompt once

### LINKED ISSUE #3511: Story: v1.1 - disable pop up after clicking"got it, thanks!"
For return visits, users would like to see the story pop up not appearing again after you click "got it, thanks!"

chloeleichen:
Fix https://github.com/TerriaJS/terriajs/issues/3511, show the prompt only once

### DIFF
--- a/lib/Models/test.ts
+++ b/lib/Models/test.ts
@@ -1,5 +1,0 @@
-let isDone = false;
-
-isDone = true;
-
-console.log(isDone);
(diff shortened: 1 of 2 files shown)


lib/Models/test.ts
```diff
@@ -0,0 +1,5 @@
+let isDone = false;
```
kring: Looks like you committed this file accidentally, can you please remove it?

kring (changes requested):

PR #3581, opened 2019-07-08
Model architecture: Use core-js for polyfilling instead of polyfill.io

kring:
The mobx branch is broken right now because babel started using `getOwnPropertyDescriptors`, which is not available in IE11 or from polyfill.io. So this PR switches to using core-js, which is the more typical solution used with Babel. The main downside is that Terria includes a full polyfill suite for all its target browsers, rather than benefitting from polyfill.io's nifty tricky of serving only the polyfills that are needed for the user's current browser.


kring:
Looks like core-js adds about 64k to our release build currently. We're only including the bits we actually use (via `@babel/preset-env`), so that'll grow if we use more features that aren't supported in our target browsers, but it doesn't seem too bad.

PR #3584, opened 2019-07-09
Give a drag'n'dropped file a name in the workbench

### LINKED ISSUE #3399: DragDrop CSVs appear as 'unnamed item'

rowanwins:
Resolves #3399 by giving a drag'n'dropped item it's filename as the name in the workbench

### DIFF
--- a/lib/Models/createCatalogItemFromFileOrUrl.js
+++ b/lib/Models/createCatalogItemFromFileOrUrl.js
@@ -156,6 +156,7 @@ function loadItem(newCatalogItem, name, fileOrUrl) {
     newCatalogItem.data = fileOrUrl;
     newCatalogItem.dataSourceUrl = fileOrUrl.name;
     newCatalogItem.dataUrlType = "local";
+    newCatalogItem.name = fileOrUrl.name;
   }
 
   return when(newCatalogItem.load()).yield(newCatalogItem);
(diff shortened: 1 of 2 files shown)


lib/Models/createCatalogItemFromFileOrUrl.js
```diff
@@ -156,6 +156,7 @@ function loadItem(newCatalogItem, name, fileOrUrl) {
     newCatalogItem.data = fileOrUrl;
     newCatalogItem.dataSourceUrl = fileOrUrl.name;
     newCatalogItem.dataUrlType = "local";
+    newCatalogItem.name = fileOrUrl.name;
```
kring, line 159: Does this work if `fileOrUrl` is a URL? I wouldn't think it would have a name property.

kring:
Please update CHANGES.md as well.

lib/Models/createCatalogItemFromFileOrUrl.js
rowanwins: This change only effects files - the [if/else statement above](https://github.com/TerriaJS/terriajs/pull/3584/files/44c73e6560d91f41a3f5d277014489c0095007e7#diff-45581fa26beec87f4e76b259504095bcL153) basically means this block only gets applied to files.

tephenavies (approved):

PR #3585, opened 2019-07-09
Prevent terria from crashing when clicking on a region with no region-mapped poly and a chart

### LINKED ISSUE #3582: Unexpected crash on region mapped dataset with chart
I made Aremi crash by clicking on an area where there wasn't a poly with a region mapped dataset

Add "Electricity Infrastructure" -> "Network Opportunities" -> "Peak Day Available Capacity"
https://www.nationalmap.gov.au/renewables/#share=s-8M7oj8LSeEekdZh8

I notice in the console there are lots of messages like `Unexpected geometry type: 1`
CrashTerria

This error is in the console 
````
react-dom.production.min.js:198 TypeError: Cannot read property 'then' of undefined
    at t.componentDidMount (Chart.jsx:133)
````
I'm guessing somewhere around [here](https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Custom/Chart/Chart.jsx#L130-L139)

rowanwins:
Resolves #3582 

See the issue listed above for details on how to replicate crash. This PR adds some light-weight error handling to the chart component.

### DIFF
--- a/lib/ReactViews/Custom/Chart/Chart.jsx
+++ b/lib/ReactViews/Custom/Chart/Chart.jsx
@@ -132,6 +132,7 @@ const Chart = createReactClass({
       that.props.url,
       that.props.catalogItem
     );
+    if (!defined(promise)) return;
     promise.then(function(data) {
       chartParameters.data = data;
       ChartRenderer.create(that._element, chartParameters);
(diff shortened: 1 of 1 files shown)


kring (approved):

kring:
I'll update CHANGES.md in master after merging this.

PR #3592, opened 2019-07-12
Add WPS catalog function & item.

### LINKED ISSUE #3478: Model architecture: Catalog functions
i.e. WPS

na9da:
Fixes #3478.

* Adds `WebProccessingServiceCatalogFunction`
* Adds `WebProccessingServiceCatalogItem`
* Converts `LineParameter`, `PointParameter` etc to TS.
* Ports UserDrawing.
* Re-enables `mapInteraction`


lib/Models/UserDrawing.ts
```diff
@@ -0,0 +1,434 @@
...
+      message: this.getDialogMessage(),
+      buttonText: this.getButtonText(),
+      onCancel: () => {
+        this.terria.mapInteractionModeStack.pop();
+        this.cleanUp();
+      }
+    });
+
+    this.terria.mapInteractionModeStack.push(pickPointMode);
+    autorun(async reaction => {
```
tephenavies: Is it possible to cancel picking without having this reaction disposed?

lib/Models/UserDrawing.ts
```diff
@@ -0,0 +1,434 @@
...
+  }
+
+  /**
+   * Called after a point has been added, prepares to add and draw another point, as well as updating the dialog.
+   */
+  private prepareToAddNewPoint() {
+    const pickPointMode = this.mapInteractionModeUpdate();
+
+    autorun(async reaction => {
+      if (isDefined(pickPointMode.pickedFeatures)) {
```
tephenavies: Again, I think this could be a memory leak.

tephenavies (changes requested):
I'm about halfway through, but I think there's a few comments that need either clarification or actioning.

lib/ModelMixins/XmlRequestMixin.ts
```diff
@@ -0,0 +1,28 @@
+import Constructor from "../Core/Constructor";
+import isDefined from "../Core/isDefined";
+
+const loadXML = require("../Core/loadXML");
+const loadWithXhr = require("../Core/loadWithXhr");
+
+export default function XmlRequestMixin<T extends Constructor<any>>(Base: T) {
+  class XmlRequestMixin extends Base {
```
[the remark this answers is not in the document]
na9da: I did this because I wanted an easy way to stub network calls in specs. I had used it in a couple of places, so I thought of extracting to a mixin. But yeah, it doesn't feel right. Ideally we need a `Network` namespace which includes all the `loadX` functions, then we can stub them easily in the specs.

I'll just remove the `mixin` for now.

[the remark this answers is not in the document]
tephenavies: Interesting. How does the mixin make it easier to stub network calls?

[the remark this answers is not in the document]
na9da: That's simple, nothing complicated: https://github.com/TerriaJS/terriajs/blob/mobx-wps/test/Models/WebProcessingServiceCatalogFunctionSpec.ts#L116

lib/Models/ResultPendingCatalogItem.ts
```diff
@@ -5,6 +5,9 @@ import ResultPendingCatalogItemTraits from "../Traits/ResultPendingCatalogItemTr
 export default class ResultPendingCatalogItem extends CatalogMemberMixin(
   CreateModel(ResultPendingCatalogItemTraits)
 ) {
+  readonly showsInfo = true;
+  readonly isMappable = true;
```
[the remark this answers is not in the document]
na9da: I think you are right about `isMappable`, we don't need to set it. 

We need `showsInfo` because in this case we are showing `parameters` for the request we are making and an also the error in case it fails. But I do think we can make it false by default and just turn it on if needed.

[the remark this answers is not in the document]
na9da: image

This is the info panel for pending item from master that I am trying to recreate. It requires `isMappable` to be `true` due to the [check here](https://github.com/TerriaJS/terriajs/blob/mobx/lib/ReactViews/Preview/DataPreview.jsx#L43).

Anyway, I have set the defaults to `false`.

lib/Models/UserDrawing.ts
na9da: Thanks!

I have modified the code to slightly improve clarity & dispose the reaction when the interaction is cancelled.

lib/Models/UserDrawing.ts
```diff
@@ -0,0 +1,438 @@
...
+  onPointClicked?: (dataSource: DataSource) => void;
+  onPointMoved?: (dataSource: DataSource) => void;
+  onCleanUp?: () => void;
+}
+
+class EmptyTraits extends ModelTraits {
+  static traits = {};
+}
+
+export default class UserDrawing extends CreateModel(EmptyTraits) {
```
tephenavies, line 40: Why extend Model without any traits? I think it'd at least be useful to implement Mappable and have Mappable-associated traits. Maybe because this shouldn't be shown in the workbench?

Maybe this should be a map overlay instead of a model? Currently there's no distinction, but maybe that'd be useful? I think we should talk about this.

na9da: It has to be a `BaseModel` because `terria.overlays` is a `Workbench` and it references some non-Trait things like `uniqueId`. I made the traits empty because I couldn't think of any useful traits for overlays.

cc @kring

tephenavies: Interesting. I didn't catch that there already was an overlays list

lib/Models/FunctionParameter.ts
```diff
@@ -0,0 +1,41 @@
...
+  abstract readonly type: string;
+  readonly id: string;
+  readonly name: string;
+  readonly description: string;
+  readonly isRequired: boolean;
+  readonly converter: unknown;
+
+  @observable value?: unknown;
+
+  geoJsonFeature?: Promise<Feature> | Feature | undefined;
```
tephenavies: What is this for? I think it should be private, const, or observable

lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -70,6 +70,9 @@ class GeoJsonCatalogItem extends AsyncMappableMixin(
 
   readonly canZoomTo = true;
 
+  // The final data after all transformations are applied
+  readyData?: JsonObject;
```
tephenavies: This might be better as a computed, but I'm not sure what that would involve.

lib/Models/FunctionParameter.ts
na9da: This is for sub classes to return a geojson feature representing them. Ideally it should be an optional `abstract`, but there is no such thing in TS. I'll make it `readonly`.

PR #3609, opened 2019-07-22
Cesium fetch xhr http 202 terria json catalog function

Jason-DNetto:
based off the feedback from the last attempt at integrating longer requests, the headers and status need to be passed back to be handled in the catalog function.
if the function receives JSON instead of a 202:Accepted, the new behavior returns the JSON immediately

### DIFF
--- a/lib/Models/TerriaJsonCatalogFunction.js
+++ b/lib/Models/TerriaJsonCatalogFunction.js
@@ -188,14 +188,14 @@ function TerriaJsonCatalogFunction(terria) {
 
   /*
    * Gets or sets the part of a HTTP 202 response to poll again for content.
-   * A HTTP 202 response is returned as an object with 
+   * A HTTP 202 response is returned as an object with
    * {
    *   status: 202
    *   headers: responseHeaders
    *   respone: responseBody
    * }
    * where
-   * responseHeaders is an object with key-value pairs of all HTTP response 
+   * responseHeaders is an object with key-value pairs of all HTTP response
    * 	headers
    * responseBody is the unaltered XMLHttpRequest.response
    * If you were to supply a 202 response with the url as a location header,
@@ -210,9 +210,9 @@ function TerriaJsonCatalogFunction(terria) {
    *   }
    * }
    * acceptedUrl can be specified as response.prop1.prop3
-   * If you were to supply a 202 response with the url as plain text in the 
+   * If you were to supply a 202 response with the url as plain text in the
    * response body, acceptedUrl can be specified as response
-   * If you were to supply a 202 response with the url as a property of some 
+   * If you were to supply a 202 response with the url as a property of some
    * other mime type/format, eg. XML, YAML
    * This is open source software, add the code yourself.
    * Leave undefined for new functionality to be off by default
@@ -221,7 +221,7 @@ function TerriaJsonCatalogFunction(terria) {
   this.acceptedUrl = undefined;
 
   /*
-   * Gets or sets the time in milliseconds to wait before polling a 
+   * Gets or sets the time in milliseconds to wait before polling a
    * HTTP 202 response acceptedUrl
    * default 10 seconds
    * is unused if acceptUrl is not defined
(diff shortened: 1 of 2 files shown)


KeyboardSounds (commented):
Thank you for submitting this PR. We appreciate the time and effort you put into contributing to Terria and keeping the open source dream alive :)

This is functionality that we really want to add to Terria, and so I'm keen to work with you to get it into `master`. Just a heads up though; at the moment we're in the process of refactoring much of our model layer to use TypeScript and MobX, and these changes might overwrite yours. The new Terria probably won't be ready for a few more months, but when it is, we might have to reimplement some of this functionality. If you'd like wait for the refactor to be finished first, let us know and we'll come back to this then. :)

I'm about to go through the diff and leave some more detailed feedback as line comments, but I think overall this is on the right track.

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -185,13 +186,54 @@ function TerriaJsonCatalogFunction(terria) {
...
+  this.acceptedDelay = 10000;
+
   /**
    * Gets or sets the input parameters to the service.
    * @type {FunctionParameter[]}
    */
   this.inputs = [];
 
-  knockout.track(this, ["url", "inputs"]);
+  knockout.track(this, ["url", "acceptedUrl", "aceptedDelay", "inputs"]);
```
KeyboardSounds: As far as I can tell, these don't need to be tracked as observables by knockout - nothing needs to react to them changing.

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -443,4 +491,102 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
...
+      // needs to be called to re-initialise the first next value,
+      var referenceIterator = this.acceptedUrl.split('.')[Symbol.iterator]();
+      var iteratorCurrent = referenceIterator.next().value;
+      // want to copy the all object because the copy will get modified
+      var iteratorAll = all;
+      while (iteratorCurrent !== undefined) {
+        iteratorAll = iteratorAll[iteratorCurrent];
+        iteratorCurrent = referenceIterator.next().value;
+      }
+      // is proxying this a good idea?
```
KeyboardSounds: All outgoing requests have to be proxied, so yes :)

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -443,4 +491,102 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
...
+     
+      // JSON-ify the body here, used Resorce.fetch, so body must be text
+      var jsonBody;
+      if (all.response !== undefined && typeof all.response === "string" &&
+          all.response !== "") {
+          jsonBody = JSON.parse(all.response);
+      }
+      if (typeof jsonBody === "object") {
+          all.response = jsonBody;
+      } else {
```
KeyboardSounds: We won't include support for non JSON responses for now, so please remove the empty else clause :)

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -374,18 +416,24 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
...
-
-  const promise = loadJson(proxiedUrl).then(json => {
+  const promise = Resource.fetchXHR({
+      url: proxiedUrl,
+      responseType: "text",
+      headers: {
+        Accept: "application/json,*/*;q=0.01"
+      },
+      returnType: "XHRJSONHEADERS"
+    }).then(all => this._handleHttp202(all)).then(json => {
```
KeyboardSounds: I'm not sure we should default to handling a 202 response like this. I think we might be better leaving this functionality "off" by default and having it selectively enabled in the constructor or something? That way consumers of this API can choose what they want to happen, and there isn't unexpected "magic" happening behind the scenes. @Jason-DNetto thoughts?

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -443,4 +491,102 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
   return promise;
 };
 
+// private method split into function to make it recursive
+TerriaJsonCatalogFunction.prototype._handleHttp202 = function(all) {
```
KeyboardSounds: This is purely a matter of opinion, but I'm not sure what `all` as a variable name really means here? `response` might be a better choice, maybe.

lib/Models/TerriaJsonCatalogFunction.js
```diff
@@ -443,4 +491,102 @@ TerriaJsonCatalogFunction.prototype.invoke = function() {
...
+        headers: {
+          Accept: "application/json,*/*;q=0.01"
+        },
+        returnType: "XHRJSONHEADERS"
+      };
+      // ability to handle Accepted URL locations like first.second
+      // or Headers.location or Body.prop1.prop2.whatever.etc
+      // cant be a static input and 
+      // needs to be called to re-initialise the first next value,
+      var referenceIterator = this.acceptedUrl.split('.')[Symbol.iterator]();
```
KeyboardSounds: To better fit with the code style that the rest of Terria uses, would it be possible for you to use a for loop here instead of directly calling next on the iterators?

KeyboardSounds (changes requested):

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: I thought that was how to get them from the init file. changed.

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: will remove comment on next commit

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: I have a re-write that i will commit soon, which involves changing a few more lines around this. into 
try{all.response = JSON.parse(all.response);} catch {}
I have done this because the response body might be empty, which is a syntax error for JSON.parse, and i thought empty clauses were javascript's no-op

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: isn't off by default returning a response with status code HTTP 200: OK?
sometimes protocols/servers return HTTP status code 200 with an XML document that has a status element with 202 in it. 
when i wrote this i had intended for consumers to control what happens when they assign acceptedUrl in the init file, and have been treating acceptedUrl === undefined as an error, but maybe this can be the off by default switch.

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: Ok

lib/Models/TerriaJsonCatalogFunction.js
Jason-DNetto: started out with response = {Status:202,Headers:{Object ...},Body:""}, but then switched to mimicking the XMLHTTPResponse object, which then had response.response, so i switched to all. 
all as a variable name was the entire xhr object, as opposed to the response part of the xhr call from Resource options method=GET, or the status and headers from Resource options method=OPTIONS or method=HEAD. 
might make more sense to rename all to xhr?

Jason-DNetto:
test init file for this and cesium TerriaJS pull request
https://gist.github.com/Jason-DNetto/ffdaa1b838c5429ee02931692a1416d4

KeyboardSounds (commented):
Sorry for taking so long to get back to you, we've got a couple of people away at the moment and so are a bit busier than usual.

Thanks for making those changes, this looks ready to be merged now! 🎉 Once I release a new terriajs-cesium, I'll hit the merge button.

Jason-DNetto:
cool!

Jason-DNetto:
found a case sensitivity issue that I am solving now

KeyboardSounds (approved):

PR #3610, opened 2019-07-22
Load basemaps before showing them

### LINKED ISSUE #3606: ArcGIS MapServer basemaps don't show
Regression bisected to this commit TerriaJS/terriajs@f82180b933f2c81.

ArcGIS MapServer items are only loaded when metadata or legends are needed. Since legends are now built only when needed, the item no longer loads, so we get no base map.
This doesn't affect non-basemap layers (since legends will be loaded).

See http://ci.terria.io/master/#share=s-lhUesl7qwTOjelvddXEmUMYAbhJ

tephenavies:
Fixes #3606.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### Next Release
+* Basemaps are now loaded before being enabled & showed
+
+
 ### v7.6.5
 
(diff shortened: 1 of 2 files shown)


tephenavies:
CI failed because it ran out of memory.

soyarsauce (approved):
Nice 💯
Ran tests locally and all good, but also restarted CI & works now

PR #3613, opened 2019-07-23
Beautiful fonts

soyarsauce:
Make fonts match Phil's sketchs on MacOS

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@ Change Log
 
 ### Next Release
 * Basemaps are now loaded before being enabled & showed
-
+* Enable css anti aliasing for MacOS
 
 ### v7.6.5
 
--- a/lib/Sass/global/_terria.scss
+++ b/lib/Sass/global/_terria.scss
@@ -1,0 +1,4 @@
+html {
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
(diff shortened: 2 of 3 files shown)


soyarsauce:
DO NOT MERGE YET! requires further discussion

soyarsauce:
>I’m going to leave that for next month. It’s a change that will require discussion and reworking some things

soyarsauce:
(Some other stuff I forgot to add here)
we haven't added it yet or will probably never, so we need to decide how we marry the designs vs web implementation

https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/
https://www.zachleat.com/web/font-smooth/
https://github.com/google/fonts/issues/1170

soyarsauce:
Re: https://github.com/TerriaJS/nsw-digital-twin/issues/244#issuecomment-599874120 bringing it up again

Myndex:
Hi @soyarsauce 

This thread was linked to a related one, so to chime in on the topic:

There are several reasons that "-webkit-font-smoothing: antialiased;" and other font smoothing properties **do _NOT_ show fonts off at their best**. What does show fonts at their best is the _DEFAULT_ for the given device, and this is nearly always subpixel antialiasing. The font-smoothing "antialiasing" property is inferior to the default subpixel antialiasing both in quality and by reducing contrast needlessly — in fact the contrast reduction alone can make a font unreadable in some cases.

IMO, _font-smooth_ is a CSS property that should be depreciated. Best results are to use the device's default antialias which is optimized for that device. That provides the best and most consistent results.

See also:  http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

Thank you,

Andrew Somers
_Myndex Color Science Researcher
W3C WAI/WCAG Invited Expert_

PR #3622, opened 2019-07-26
Model architecture: Charts

### LINKED ISSUE #3466: Model architecture: Charts

kring:
Gets charts working in feature info and in the panel at the bottom, but only just. There are heaps of problems. Mostly the problems boil down to `CsvCatalogItem` isn't putting all the necessary information into the `ChartData` instances returned by the `chartItems` property, and the D3 code is dodgy.

Fixes #3466


kring:
Chart panel is always open and cannot be closed.

PR #3634, opened 2019-08-06
Use existing share-building functions to work out what items can't be shared

tephenavies:


na9da:
Build passes but deployment fails due to error in [sync-dependencies](https://github.com/TerriaJS/sync-dependencies/pull/1).

Merging!

na9da (approved):

PR #3645, opened 2019-08-09
WIP - Chart fixes

rowanwins:
### Problem - Terria can crash when there is no data to chart
https://www.nationalmap.gov.au/renewables/#share=s-7ozcBfhZiUTnlQl8
Click on one of the markers where there is no data data (one of the white points)
- A blank chart is shown with no message
- Click on the on 'Download' or 'Expand' buttons above the chart and Terria will crash

#### Summation of issues encountered
- The generic dropdown component did not have a disabled state, I've now added one
- The chart can get confused if headers exist but there is no row data, for example 
  - it wouldn't know what type of chart to render (eg line chart vs bar chart)
  - it can't work out the axis

#### Things implemented to help
  - [x] assigning a default chart type when there is no data which helps prevent downstream crashes
  - [x] not attempting to render the chart when there is no axis 
  - [x] the 'No preview available' message is now properly displayed when there is nothing to chart

-------------------------------

### Problem - Terria can't plot imagery availability by itself
https://www.nationalmap.gov.au/#share=s-qfULVIswLao2JkrOzqlWoc6YnCg
Click on the chart button next to the date picker and nothing happens

#### Things implemented to help
- [x] allow imagery to be plotted on a chart by itself even though it doesn't have a table structure by modifying the `chartIsOpen` property in `ViewState`

--------
### Additional observations
I think one of the challenges is that there is no useful shared state between `ChartExpandAndDownloadButtons` and the `Chart` component within the `FeatureInfoSection` component
````
// Component Hierarchy
FeatureInfoSection component contains
- ChartExpandAndDownloadButtons component
- Chart component
````
The `FeatureInfoSection` is primarily based around the `Feature` model.

The `Chart` component is responsible for retrieving and rendering the data required for the chart, however it doesn't share that information with any other component which might benefit from it (eg `FeatureInfoSection` or `ChartExpandAndDownloadButtons`).

The `ChartExpandAndDownloadButtons` component has access to a `Feature` that's been selected, however it doesn't know anything about the data being shown on the chart. The `Chart` does not have access to the `Feature` - it just has a url which has some uniqueid.

### CODE REVIEW

**Comment by rowanwins** [comment]
Things I've tested
- Add an aremi layer 
  - http://ci.terria.io/chart-imagery-fix/#share=s-nkvSMv4HaJ3BeUfBZppgjN3UOxZ
  - click a site with data
    - check the chart and downland and expand btns are shown ok
    - check that it refreshes ok with the 5 min marker
    - check that share works ok
  - select a site with no data
    - check the chart shows the no data message
    - check the downland and expand btns are disabled
    - check that nothing breaks when the data refreshes after the 5 mins
    - check that share works ok

- Add an imagery layer that has time available
  - http://ci.terria.io/chart-imagery-fix/#share=s-XeNbAIpeWfYAKAm0ssZm0MbT4Q
  - click the chart button on the workbench item


- Add an imagery layer with time and a chart layer from AREMI
  - http://ci.terria.io/chart-imagery-fix/#share=s-f8XASGSUAI0o71LJIvr3m2k0rHy
  - Check that data from both can be displayed

**Comment by rowanwins** [comment]
Also added a fix where hovering over a data point wouldn't work when it was in the same area as the tooltip. The change means that the y position of the tooltip adjusts based on position of the users cursor.

**Comment by rowanwins** [comment]
@philipgrimmett - do you wanna have a play around and see if you can break things or if there is anything else that needs tidying?

**Comment by kring** [comment]
I haven't looked at this closely yet, but I tried out http://ci.terria.io/chart-imagery-fix/ with the current power generation layer, and the chart zooming/panning on the feature info panel is really cool! Couple problems I noticed though:
* The Expand buttons on the feature info don't work (silent exception appears in the console, instead of the chart appearing at the bottom).
* I've seen the zoom/pan of a chart on the feature info get "stuck". I can't quite see what causes it though. I do something like: click on a power station, zoom in a bit with the mouse wheel, pan all the way to the left, zoom in and out a bit, pan all the way to the left again, do some more zooming, etc., and eventually zooming or panning or both stop working. I can make it happen reliably, but step-by-step instructions are elusive. If you're not able to reproduce it, let me know and I'll see if I can nail it down better.

**Comment by rowanwins** [comment]
> The Expand buttons on the feature info don't work (silent exception appears in the console, instead of the chart appearing at the bottom).

Good pickup - I've mainly been testing with water data and imagery, the last commit added a highlight marker on the map for where you'd clicked on say a stream gauge, this is obviously is a bit brittle!

Re zooming and panning getting stuck I don't think I've seen this to the extent you have although I have noticed a few occasions where it's stopped, when this has occurred its normally resolved itself after a click or shuffle of the mouse to a different part of the chart.
This is the share url most of us have been testing on.
http://ci.terria.io/chart-imagery-fix/#share=s-9KNNerWW3JJZKibZmRpJuu0NaLY

**Comment by rowanwins** [comment]
@kring I think this is ready for another play around test to see if you can break anything. I'm just check on Travis.

- I've added a prop to the chart which specifies if the chart is zoomable or not
  - charts in popups are set to not be zoomable

- One bit I'm unhappy with is that the tracking of the selected imagery date and how that relates to which marker on the chart is highlighted. 
  - I feel like this could be done better but at the time I was battling so many things 
  - There is a few things in [here](https://github.com/TerriaJS/terriajs/pull/3645/files#diff-aa47eb4c54bb21b9570a1ee2ec1c3bdc) and [here](
https://github.com/TerriaJS/terriajs/pull/3645/files#diff-43f84857618fd18627067e59a3816c20)

- The relationship between the x-highlight and the tooltip needs more thought. I'm not able to clearly articulate the expected behaviour although I think this can happen down the track.

- I'm still working on the sharing of markers that are created, but I think that can be finished off separately.

File: lib/Models/ImageryLayerCatalogItem.js [hunk 1]
```diff
@@ -151,6 +151,20 @@ var ImageryLayerCatalogItem = function(terria) {
    */
   this.dontChartAlone = false;
 
+  /**
+   * Gets or sets the color of how the dataset will displayed on the chart
+   * @type {String}
+   * @default '#44C99F'
+   */
+  this.chartColor = this.chartColor ? this.chartColor : "#44C99F";
```
kring: `this.chartColor` can't be defined already on this line because it is created for the first time on this line.

lib/Charts/ChartRenderer.js
```diff
@@ -32,9 +33,13 @@ const defaultNoDataText = "No preview available";
 
 const defaultTransitionDuration = 1000; // milliseconds
 
-const threshold = 1e-8; // Threshold for 'equal' x-values during mouseover.
+// const threshold = 1e-8; // Threshold for 'equal' x-values during mouseover.
```
kring: Just remove this if it's not needed anymore.

lib/Charts/ChartRenderer.js
```diff
@@ -118,23 +158,34 @@ class ChartRenderer {
   }
 
   static render(container, state) {
+    state.dblClickTimer = false;
     if (!defined(state.data)) {
       return;
     }
 
+    // We want points to appear on top of everything else
+    state.data.sort(function(a, b) {
```
kring: This sort isn't guaranteed to the stable. MDN says it's not on IE, Edge, and possibly Safari: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Browser_compatibility
But because you're returning -1 for all non-moment-points, it'll change the relative order even in browsers that do a stable sort.

lib/Charts/ChartRenderer.js
```diff
@@ -516,49 +679,83 @@ function findSelectedData(data, x) {
...
     )
   );
+  if (closestXPoints.length === 0) return [];
+  let selectedPoints = null;
+  let isSelectedArray = null;
+
+  function compareUsingDates(a, b) {
+    const diffTime = Math.abs(a.x.getTime() - b.getTime());
+    const diffDays = Math.ceil(diffTime / (1000 * 60 * 60 * 24));
+    return diffDays < 3;
```
kring: This `3`, and the `1e-8` below, are pretty arbitrary and likely to be give very bad results for some datasets. Is there any way to ask D3 for the X delta that corresponds to, say, 3 pixels? Cause I think something like that would be ideal.

lib/Charts/ChartRenderer.js
```diff
@@ -516,49 +679,83 @@ function findSelectedData(data, x) {
...
-    .data(selectedData);
+    .data(data);
   selection
     .enter()
     .append("circle")
     .merge(selection)
-    .attr("cx", d => scales.x(d.point.x))
-    .attr("cy", d => scales.y[d.units || Scales.unknownUnits](d.point.y))
+    .attr("cx", d => (scales.zoomedX ? scales.zoomedX(d.x) : scales.x(d.x)))
+    // .attr("cy", d => scales.y[d.units || Scales.unknownUnits](d.y))
```
kring: Remove this line if it's not needed.

lib/Charts/ChartRenderer.js
```diff
@@ -575,26 +772,50 @@ function highlightAndTooltip(
...
     chartBoundingRect
   );
 
   if (fireClick) {
-    selectedData.forEach(chartData => {
-      if (chartData.onClick) {
-        chartData.onClick(hoverX); // TODO also pass Y
-      }
-    });
+    // Because double click should be used for zooming the chart out
```
kring, line 825: I don't quite understand the need for this manual double click detection. D3 lets you attach any event to a DOM element with `.on`, right? And browsers have a `dblclick` event. So I don't see why `whatever.on("dblclick", callback)` wouldn't work. One good reason to avoid manually detecting double clicks is that its an accessibility issue. 250ms is ok for some users, but not all. That's why it's configurable at the OS level.

rowanwins:
Hmm interesting @kring - I had noticed that same thing during testing but it always dissapears once you close the popup, I'd assumed that it had always been like that but I think it is something funky associated with this PR so I'll investigate.

lib/Charts/Tooltip.js
```diff
@@ -126,28 +139,37 @@ const Tooltip = {
...
-          line.color,
-          line.categoryName,
-          line.point.y,
-          line.units
-        );
-      });
-      html += "</tbody></table>";
-      return html;
-    }
+    // const dataGroupedByName = d3Nest()
```
kring, line 142: Remove this code if it's not needed.

lib/ReactViews/Custom/Chart/ChartExpandAndDownloadButtons.jsx
```diff
@@ -68,6 +72,26 @@ const ChartExpandAndDownloadButtons = createReactClass({
...
     }
+
+    // This feels very smelly but it works for the AREMI layers
+    // chartData for a catalogItem returns stuff like
+    // {x: "Butlers Gorge Power Station", y: 67.33}
+    // Occassionally y is null so we know we can't plot it
+    // {x: "Some Station", y: null}
+    // But there isn't a good way to get the chartData for a feature
+    // which is what we ideally need
+    // It doesn't work for SensorObservationService data where chartData is undefined
```
kring: Yeah this doesn't seem right to me. Even if the current value is null (which is what we're looking at here I think), there might still be historical data to plot.

Why can't we get the feature's chartData? This component is used alongside a chart on the feature info panel, right? And that chart certainly has the feature's chart data.

kring (changes requested):

rowanwins:
So I think what's happening in the above is that the two charts (the featureinfo and the panel) are being the same size, and because the featureinfo is created first the panel is getting it's size...

rowanwins:
Okay so the missing chart when popup and panel first get opened is now resolved. It was caused by a clash in `id` selector's (eg there being 2 elements with an id of `chartAreaClip`). It would perhaps be handy to have some sort of chartId that could be used to make specific selector.

lib/Charts/ChartRenderer.js
rowanwins: So I think the problem I was trying to solve here was that a `dblclick` was supposed to make the chart zoom out. However it was also triggering the single `click` event which meant it was triggering a change in imagery.

kring: Ah I see. Yeah that makes it awkward. It'd be slightly better to make the dblclick event cancel the delayed action from the click event, rather than letting a second click (whether interpreted as a double click or not) cancel the action. Even better would be to do the click action immediately, and then "undo" it when the dblclick arrives, but that might be hard. And even better than all that is to not rely on double-clicking to zoom out, cause I'm not sure anyone is going to think of that anyway. :)

lib/Charts/ChartRenderer.js
rowanwins: Have replaced with a simple for loop

lib/Charts/ChartRenderer.js
rowanwins: Have made a 3 pixel range

lib/ReactViews/Custom/Chart/ChartExpandAndDownloadButtons.jsx
rowanwins: Because the `chart` and the `chartExpandAndDownloadButtons` are being rendered independently as child components of `FeatureInfoSection` they don't have a shared state of the retrieved chart data I don't think.

I'm wondering if `chartData` ought to be a property, and have a method on the catalogItem called `getChartData`, and then perhaps some sort of `chartDataStatus` property which has a few prescribed values (eg `retrieved`, `retrieving`, `unretrieved` or some way better word).

I think the other challenge in this space is chartData that relates to a feature vs chartData which relates to a whole catalog item. It's a bit unclear in my mind how's that's currently working.

rowanwins: Have removed the disabling of buttons for the timebeing, but needs more thought. Have check that nothing breaks at least and that we show the 'no preview available' message properly.

kring:
1. Zoom in on a chart.
2. Double-click on the chart to zoom back out.
3. Click and drag within the chart and you'll zoom back in unexpected.

kring:
Minor: an `Unnamed Item` appears on the workbench while chart data is loading. This is easy to see with the power generation dataset when expanding 3 months.

kring:
When you have two datasets on the chart with different time periods, and put the cursor before the start of the second one (up to about 1/5 of the total chart width), the tooltip misleadingly shows a value for the second one attributed to an incorrect time.
image

kring:
1. Visit http://ci.terria.io/chart-imagery-fix/#charts
2. Enable "Star Stations" and "Sheep Stations".
3. Click Sheep Station Number 55, which is the northeast point of the group of four just north of Melbourne.
4. Click Expand on the feature info chart.
5. Click Ganymede, which is the the point roughly in the Nullarbor, click Expand -> 1 day in the feature info

Both lines will be cyan:
image

Even though the workbench says one should be orange:
image

If you toggle Ganymede's "Signal at 560GHz" off and back on it'll right itself.

Also, the vertical line / tooltip only snaps to one (the first?) dataset, rather than snapping to any of them.

Master (correct, snaps to every point on either line):
chart-snap

This branch (incorrect, snaps only to points on the cyan line):
chart-snap-broken

rowanwins:
@kring 

I've resolved the tooltip not showing the right thing and [strange snapping behaviour.](https://github.com/TerriaJS/terriajs/pull/3645#issuecomment-539337995)

I've resolved the [colours issue](https://github.com/TerriaJS/terriajs/pull/3645#issuecomment-539347831)

Re the Unnamed Item appearing in the workbench I've tried a few things on this one and can't get it to work. Either the item doesn't show in the workbench and no marker appears on the map, or vice-versa.

kring (approved):

kring:
Thanks @rowanwins. Please update CHANGES.md in master.

PR #3663, opened 2019-08-26
SDMX 2.0 fixes 

### LINKED ISSUE #3659: SDMX doing weird things
https://nationalmap.gov.au/#share=s-o1oGEGcGdINRoG6tiezmIqQvfu5

Add an SDMX dataset and change some of the conditions and it starts freaking.

In the link above if you try and remove the dataset from the map you can't - it's like it keeps holding on to some reference to a past version of the dataset which is can no longer remove

rowanwins:
Resolves #3659 by preventing the double-enabling of items.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### v7.6.10
+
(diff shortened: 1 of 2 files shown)


na9da (approved):
Looks good to me.

Issue #3441, opened 2019-05-28
Model architecture: Add support for continuous color ramps


kring:
E.g. http://ci.terria.io/master/#build/TerriaJS/test/init/csv.json and add CSV Testing -> Colors -> CSV with colorBins=0.
![image](https://user-images.githubusercontent.com/924374/58449802-9abd2f80-814f-11e9-827d-cb25a7e25f55.png)

nf-s:
https://github.com/TerriaJS/terriajs/pull/5426

nf-s:
Related to #3524 SDMX 2021-02-23_Testing feedback:

### Some heat maps don't have high contrast

• Not a major issue but some heat maps aren't displaying the contrast between low and high values on the map:
• E.g. Retail Trade in previous National Map: https://nationalmap.gov.au/#share=s-e1xho9kGoFg0hVO3LvFs0lF4TgW
• Same data in new National Map: http://ci.terria.io/sdmx-fixes-loading/#share=s-4lk6ZdzU1VcnSJIdTQIyCLP2CyU

This can be solved by using a continuous color ramp - i.e. not binning the data (#1585)

nf-s:
From duplicate ticket: #1585 

> Rather than binning the area values, they could be mapped to a continuous interval. For example, we could:
> - 'equalise' the value distribution through a non-linear mapping - eg compute (area rank)/(number of areas) to obtain a value between 0 and 1
> - linearly map the values to the (0,1) interval by computing (y-y_min)/(y_max-y_min). This can be made more robust to data outliers by mapping between the nth and nth last percentiles, rather than the min and max.
> - apply any other transforms such as truncated logarithm
> 
> We then apply the colour map function to each value to compute a RGB colour. There are many choices of such functions in the literature, but a perceptually linear colour map would be a principled choice - Matplotlib implements a bunch of (often piecewise-linear) functions that could be easily ported to javascript:
> http://matplotlib.org/examples/color/colormaps_reference.html
> https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/_cm.py

wibily:
considering clamping. E.g most data between 0-5 and a random 2000 value

nf-s:
Closed by https://github.com/TerriaJS/terriajs/pull/5426

PR #3667, opened 2019-08-27
Add continuous color ramp for table items.

na9da:
PR for issue #3441  

This is how it looks now: 
image

It uses CSS `linear-gradients` to implement the gradient, however I haven't been able to accurately re-create the ticks, that might need reworking.


kring:
@na9da I think there might be a better way to set up the traits here. If I'm understanding correctly, when a legend item has `isGradientItem` set, its color and other fields are ignored and it's only used to tick/label points in the gradient. And then the separate `gradientColorStops` trait specifies the colors of the gradient. I had a hard time wrapping my head around this, so I suspect other people will too.

I think it would be cleaner to say that `items` is always discrete boxes, and add a `title` property to `gradientColorStops` for the gradient case. All `items` would be shown below the (single) gradient. A legend with just a gradient wouldn't need an `items` property at all.

If we want to support multiple gradients or to allow the discrete boxes to be above the gradient, we'll need something more sophisticated, but I don't think either of those are important use cases.

What do you think?

kring:
Also would it help to have @soyarsauce's help with the HTML/CSS to align the legend ticks?

nf-s:
Superseded by https://github.com/TerriaJS/terriajs/pull/5426

PR #3671, opened 2019-08-29
Enable CSV polling.

### LINKED ISSUE #3541: Model architecture: CSV polling
`CsvCatalogItem` in the old architecture has a `polling` property that can be used to periodically refresh the CSV data. This needs to be added to the new architecture.

na9da:
Resolves #3541.

Adds an `AutoRefreshMixin` which can be used to implement polling behavior for any mappable item.


lib/ModelMixins/AutoRefreshingMixin.ts
```diff
@@ -0,0 +1,76 @@
...
+          this._autoRefreshDisposer !== null
+        ) {
+          this._autoRefreshDisposer();
+        }
+      });
+    }
+
+    @computed
+    protected get _pollingTimer(): number | undefined {
+      if (this.refreshInterval !== null && this.refreshInterval !== undefined) {
```
kring: No need to check for null when the type doesn't allow it.

lib/ModelMixins/AutoRefreshingMixin.ts
```diff
@@ -0,0 +1,76 @@
...
+            this.refreshData();
+          }
+        );
+      });
+      onBecomeUnobserved(this, "mapItems", () => {
+        if (
+          this._autoRefreshDisposer !== undefined &&
+          this._autoRefreshDisposer !== null
+        ) {
+          this._autoRefreshDisposer();
```
kring, line 40: I think a simple `if (this._autoRefreshDisposer)` would be fine here. Also it'd be a good idea to set `this._autoRefreshDisposer=undefined` after you call it.

lib/ModelMixins/AutoRefreshingMixin.ts
```diff
@@ -0,0 +1,76 @@
...
+type AutoRefreshing = Model<MappableTraits>;
+
+export default function AutoRefreshingMixin<
+  T extends Constructor<AutoRefreshing>
+>(Base: T) {
+  abstract class AutoRefreshingMixin extends Base {
+    _autoRefreshDisposer: IReactionDisposer | undefined;
+
+    /* Return the interval in seconds to poll for updates. */
+    abstract refreshInterval: number | undefined;
```
kring: No need for this to be settable, right?

lib/ModelMixins/AutoRefreshingMixin.ts
```diff
@@ -0,0 +1,76 @@
...
+          this._autoRefreshDisposer !== undefined &&
+          this._autoRefreshDisposer !== null
+        ) {
+          this._autoRefreshDisposer();
+        }
+      });
+    }
+
+    @computed
+    protected get _pollingTimer(): number | undefined {
```
kring: Can this be private? It looks like it can be. That'd be better because the name is much less important for a private field than for a protected one. And the name isn't quite right. It's not a timer, it's a time. But it's a funny time that only updates every refresh interval. So, hard to name, better if we don't have to sweat it too much.

lib/ModelMixins/AutoRefreshingMixin.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (this.refreshInterval !== null && this.refreshInterval !== undefined) {
+        return now(this.refreshInterval * 1000);
+      }
```
kring, line 52: IMO it's good practice to make the `return undefined` explicit when that is what's intended. Otherwise it looks like an accident that the function can return undefined.

lib/ModelMixins/TableMixin.ts
```diff
@@ -175,6 +176,29 @@ export default function TableMixin<T extends Constructor<Model<TableTraits>>>(
...
+      if (
+        this.dataColumnMajor !== undefined &&
+        this.dataColumnMajor.length !== dataColumnMajor2.length
+      ) {
+        throw new DeveloperError(
+          "Cannot add tables with different numbers of columns."
+        );
+      }
+
+      runInAction(() => {
```
kring: Consider adding the `@action` decorator to this method instead of using `runInAction`.

lib/ModelMixins/TableMixin.ts
```diff
@@ -175,6 +176,29 @@ export default function TableMixin<T extends Constructor<Model<TableTraits>>>(
...
+        const appended = this.dataColumnMajor || [];
+        dataColumnMajor2.forEach((newRows, col) => {
+          appended[col] = (appended[col] || []).concat(newRows);
```
kring: This `concat` is likely to be a _lot_ slower than just pushing the new data.

kring (changes requested):
This looks really good @na9da, just a few nitpicks.

PR #3672, opened 2019-08-29
Don't encode unique ID on WMS groups

soyarsauce:
UniqueID should leave encoding/decoding to the consumer, the base
unique ID isn't encoded, so if we don't change this, we end up
with a mix of non-encoded and encoded parts in a given ID


na9da:
Looks good to me!

na9da (approved):

PR #3676, opened 2019-09-02
Mobx chart changes for NSW air quality

na9da:
* Open/close chart panel
* Increase default point size
* Make table style name available to feature info template
* Fixes error when chart data is empty
* Mobx merge


kring:
@na9da the merge from mobx makes it hard to review this (too many unrelated changes). Also it's weird that 117dae9 isn't a merge commit, what happened there?

na9da:
See #3726

PR #3677, opened 2019-09-02
Retrieve styling for Esri Feature Services

rowanwins:
This PR allows a parameter to be set on Esri Feature Services (currently default to `false`) to retrieve styling information, for example setting fill-color on polygons.

For example on a Feature Service's metadata endpoint you can find information on renderers (see [Esri doco here](https://developers.arcgis.com/documentation/common-data-types/renderer-objects.htm))
````
drawingInfo": {
    "renderer": {
        "type": "simple",
        "symbol": {
            "type": "esriSFS",
            "style": "esriSFSSolid",
            "color": [235, 247, 208, 255],
            "outline": {
                "type": "esriSLS",
                "style": "esriSLSSolid",
                "color": [110, 110, 110, 255],
                "width": 0.4
            }
            }
        },
    "transparency": 0,
    "labelingInfo": null
}
````

There are different types of renderers so at the moment I've just done the most basic stuff for `simple` and `uniqueValue` renderers 

There is also a `classBreaks` renderer that I could also look at supporting - I just need to find a service that actually uses those to test against.

### CODE REVIEW

File: lib/Core/rgbToHex.js [hunk 1]
```diff
@@ -0,0 +1,8 @@
+function componentToHex(c) {
+  var hex = c.toString(16);
+  return hex.length === 1 ? "0" + hex : hex;
+}
+
+export default function rgbToHex(r, g, b) {
```
kring: Cesium's `Color` class can do this: https://cesiumjs.org/Cesium/Build/Documentation/Color.html#toCssColorString

lib/Models/ArcGisFeatureServerCatalogItem.js
```diff
@@ -33,6 +35,15 @@ var ArcGisFeatureServerCatalogItem = function(terria) {
...
   this.layerDef = "1=1";
+
+  /**
+   * If set to true attempts to symbolise the data using the drawingInfo
+   * object available in the service endpoint. Defaults to false.
+   * Currently supports getting fill-color, fill-opacity, stroke-color, stroke-width
+   * from simple and uniqueVals renderers.
+   * @type {Boolean}
+   */
+  this.inheritStyleInformation = false;
```
kring: I think it'd be fine to default this to true, unless there's some cost to it that isn't obvious to me? And maybe `useDrawingInfo` rather than `inheritStyleInformation`?

lib/Models/ArcGisFeatureServerCatalogItem.js
```diff
@@ -146,13 +212,68 @@ ArcGisFeatureServerCatalogItem.prototype.showOnSeparateMap = function(
...
+function setupUniqueValRenderer(renderer) {
+  const out = {};
+  for (var i = 0; i < renderer.uniqueValueInfos.length; i++) {
+    const val = renderer.uniqueValueInfos[i].value;
+    out[val] = renderer.uniqueValueInfos[i];
+  }
+  return out;
+}
+
+function convertEsriRendererToStyle(symbol) {
```
kring: BTW I partially implemented this for the NSW Digital Twin:
https://github.com/TerriaJS/nsw-digital-twin/blob/master/lib/Models/EsriRestVectorCatalogItem.js#L145
Just in case it's helpful.

lib/Models/ArcGisFeatureServerCatalogItem.js
```diff
@@ -108,10 +119,65 @@ ArcGisFeatureServerCatalogItem.prototype._getValuesThatInfluenceLoad = function(
...
+                if (val3) {
+                  symbolName += renderer.fieldDelimiter + val3;
+                }
+              }
+            }
+
+            let rendererStyle = rendererObj[symbolName];
+            if (rendererStyle === null)
+              rendererStyle = that._metadata.drawingInfo.renderer.defaultSymbol;
+            Object.assign(
```
kring: Make sure `Object.assign` is ok in IE11. It used to not be, but it might be ok now. If it's still a problem, you can use Cesium's `combine` function instead.

lib/Models/ArcGisFeatureServerCatalogItem.js
```diff
@@ -146,13 +212,68 @@ ArcGisFeatureServerCatalogItem.prototype.showOnSeparateMap = function(
...
 function loadGeoJson(item) {
   return loadJson(buildGeoJsonUrl(item)).then(function(json) {
     // console.log('Converting first element of', json.layers);
     return featureDataToGeoJson(json.layers[0]);
   });
 }
 
+function loadMetadata(item) {
+  const metaUrl = buildMetdataUrl(item.url);
+  return loadJson(metaUrl).then(function(json) {
```
kring, line 303: It might be necessary to proxy if the server doesn't support CORS, use `proxyCatalogItemUrl`.

kring (dismissed):
👍

lib/Models/ArcGisFeatureServerCatalogItem.js
rowanwins: It sends one extra request to the overarching serving info, eg 
https://sampleserver3.arcgisonline.com/ArcGIS/rest/services/Fire/Sheep/FeatureServer/0?f=json

lib/Models/ArcGisFeatureServerCatalogItem.js
rowanwins: You are right - and I feel like you knew that that wouldn't work!

kring: Haha I really wasn't sure. We've adjusted our polyfills enough times now that I've lost track. I think in mobx it would be ok now, at least.

lib/Core/rgbToHex.js
rowanwins: Cool - I did look at that class and was suprised it couldn't do what I wanted - I read that method as converting a Cesium Color to a css string.

lib/Models/ArcGisFeatureServerCatalogItem.js
rowanwins: Think we've both covered off different parts - I'll steal some of your bits to flesh mine out a bit further.

rowanwins:
I've refactored this a bit @kring and think it's probably a bit more sturdy now, basically I took your idea from the mobx branch of modifying the entities after they've been created, rather than attempting to pass down simple-style properties which was getting very convoluted.

I've got support for 
- plain points, 
- points as billboards that have a url
- polylines
- polygons

I have left the support for the GeoJsonCatalogItem being able to get a `marker-url` but this could be removed now in theory.

Here are some sample catalog items to test 
````
{
  "homeCamera": {
    "north": 50,
    "east": -69.60937499999999,
    "south": 23,
    "west": -130.78125
  },
  "catalog": [    
    {
        "id": "Esri-FS1",
        "name": "Esri styled polygons",
        "type": "esri-featureServer",
        "url": "https://services.arcgis.com/P3ePLMYs2RVChkJx/arcgis/rest/services/World_Regions/FeatureServer/0",
        "clampToGround": true,
        "isEnabled": false,
        "useStyleInformationFromService": true
    },
    {
        "id": "Esri-FS2",
        "name": "Esri styled lines",
        "type": "esri-featureServer",
        "url": "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/ArcGIS/rest/services/Chicago_L_Lines/FeatureServer/1",
        "isEnabled": false,
        "clampToGround": true,
        "useStyleInformationFromService": true
    },
    {
        "id": "Esri-FS3",
        "name": "Esri Markers (in US)",
        "type": "esri-featureServer",
        "url": "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/ArcGIS/rest/services/Point_SmartMapping/FeatureServer/5",
        "isEnabled": false,
        "clampToGround": true,
        "useStyleInformationFromService": true
    },
    {
        "id": "Esri-F4",
        "name": "Esri Colored Points (in US)",
        "type": "esri-featureServer",
        "url": "https://services.arcgis.com/V6ZHFr6zdgNZuVG0/ArcGIS/rest/services/homicide_data/FeatureServer/0",
        "isEnabled": false,
        "clampToGround": true,
        "useStyleInformationFromService": true
    }
   ]
}
````

File: lib/Models/ArcGisFeatureServerCatalogItem.js (hunk 2 above)
Comment (tephenavies) [reply lib/Models/ArcGisFeatureServerCatalogItem.js]: Any more thoughts on this? I think `inheritStyleInformation` could be mistaken with the logic for passing configuration from groups to members.

File: lib/Models/GeoJsonCatalogItem.js [hunk 6]
```diff
@@ -511,10 +513,21 @@ function loadGeoJson(geoJsonItem) {
...
+
+      // If we've got a marker url use that in a billboard
+      if (defined(entity.billboard) && defined(options.markerUrl)) {
+        entity.billboard = {
+          image: proxyCatalogItemUrl(geoJsonItem, options.markerUrl),
+          heightReference: HeightReference.RELATIVE_TO_GROUND,
+          width: style["marker-width"],
+          height: style["marker-height"],
+          rotation: style["marker-angle"]
+        };
```
tephenavies: Why do we need HeightReference here but not elsewhere?

tephenavies (commented):
It looks good. Just a couple of questions to resolve

lib/Models/GeoJsonCatalogItem.js
rowanwins: Because for these markers we're moving to a billboard and we want them to not float on some unknown part of the atmosphere.

rowanwins: re `inheritStyleInformation` is perhaps a better name `useStyleInformationFromService` @steve9164 ?

rowanwins: Looks like we should add the height reference to all entities

lib/Models/ArcGisFeatureServerCatalogItem.js
```diff
@@ -108,10 +126,67 @@ ArcGisFeatureServerCatalogItem.prototype._getValuesThatInfluenceLoad = function(
...
+
+            let rendererStyle = rendererObj[symbolName];
+            if (rendererStyle === null) {
+              rendererStyle = that._metadata.drawingInfo.renderer.defaultSymbol;
+            }
+
+            updateEntityWithEsriStyle(entity, rendererStyle.symbol, that);
+          });
+          // Because we're adding the styles after we've loaded the geojson we need to reload the CatalogItem
+          // that._geoJsonItem.load();
```
tephenavies: Is this no longer needed?

rowanwins: Good pickup

tephenavies (approved):

PR #3678, opened 2019-09-02
Mobx terrace

soyarsauce:
update for terrace-related-requirements

- Magda groups need to update its dereferenced knownContainerUniqueIds 
- Rough "terria-config" loading from magda on init if it exists


na9da (approved):
Looks good to me.

PR #3679, opened 2019-09-03
Support coords crossing anti-meridian in 2D mode with geojson

rowanwins:
When using the Leaflet viewer there are issues viewing data that crosses in the anti-meridian.
AntiMeridian

Leaflet normally handles this by displaying coords that cross the anti-meridian by using coords that are greater than 180 degrees (eg see below).

The challenge currently is that we generate our Leaflet polygon coordinates from a Cesium `Entity` which converts the coords to a traditional range (eg geojson with a coord of `181` becomes `-179`).

This PR attaches the original geojson feature to an entity when it is created via the `GeoJsonCatalogItem`, and the LeafletVisualiser uses the original geojson if it finds it. 

````
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              176.572265625,
              -17.14079039331664
            ],
            [
              176.044921875,
              -19.72534224805787
            ],
            [
              179.82421875,
              -19.808054128088575
            ],
            [
              181.40625,
              -18.562947442888298
            ],
            [
              181.845703125,
              -17.224758206624628
            ],
            [
              179.736328125,
              -15.707662769583505
            ],
            [
              176.572265625,
              -17.14079039331664
            ]
          ]
        ]
      }
    }
  ]
}
````

### DIFF
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -512,9 +512,11 @@ function hierarchyToLatLngs(hierarchy) {
   let lastLongitude = null;
   for (var i = 0; i < carts.length; ++i) {
     let lon = CesiumMath.toDegrees(carts[i].longitude);
-    // some anti-merdian swapping going on
+
     if (lastLongitude - lon > 180) {
       lon = lon + 360;
+    } else if (lastLongitude - lon < -180) {
+      lon = lon - 360;
     }
     latlngs.push(L.latLng(CesiumMath.toDegrees(carts[i].latitude), lon));
     lastLongitude = lon;
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -243,15 +243,22 @@ LeafletGeomVisualizer.prototype.getLatLngBounds = function() {
 
 var cartographicScratch = new Cartographic();
 
-function positionToLatLng(position) {
+function positionToLatLng(position, prevLayer) {
   var cartographic = Ellipsoid.WGS84.cartesianToCartographic(
     position,
     cartographicScratch
   );
-  return L.latLng(
-    CesiumMath.toDegrees(cartographic.latitude),
-    CesiumMath.toDegrees(cartographic.longitude)
-  );
+
+  let lon = CesiumMath.toDegrees(cartographic.longitude);
+  if (prevLayer) {
+    if (prevLayer._latlng.lng - lon > 180) {
+      lon = lon + 360;
+    } else if (prevLayer._latlng.lng - lon < -180) {
+      lon = lon - 360;
+    }
+  }
+
+  return L.latLng(CesiumMath.toDegrees(cartographic.latitude), lon);
 }
 
 LeafletGeomVisualizer.prototype._updatePoint = function(
@@ -322,8 +329,13 @@ LeafletGeomVisualizer.prototype._updatePoint = function(
       opacity: outlineColor.alpha
     };
 
+    let lastPoint = null;
+    featureGroup.eachLayer(function(layer) {
+      lastPoint = layer;
+    });
+
     layer = details.layer = L.circleMarker(
-      positionToLatLng(position),
+      positionToLatLng(position, lastPoint),
       pointOptions
     );
     layer.on("click", featureClicked.bind(undefined, this, entity));
@@ -340,7 +352,7 @@ LeafletGeomVisualizer.prototype._updatePoint = function(
   }
 
   if (!Cartesian3.equals(position, details.lastPosition)) {
-    layer.setLatLng(positionToLatLng(position));
+    layer.setLatLng(positionToLatLng(position, lastPoint));
     Cartesian3.clone(position, details.lastPosition);
   }
 
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -446,17 +446,10 @@ LeafletGeomVisualizer.prototype._updatePolygon = function(
       opacity: outlineColor.alpha
     };
 
-    // By default we use cesium entities to generate coords for our cesium map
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rowanwins** [comment]
I'd also like us to consider setting `worldCopyJump` to `false` in the [TerriaViewer.js](https://github.com/TerriaJS/terriajs/blob/master/lib/ViewModels/TerriaViewer.js#L542) because currently the geojson disappears if you move the center of the map slightly around the anti-meridian. Not sure if it's set to `true` for a specific reason though...

**Comment by rowanwins** [comment]
So this one probably needs some high level thought before a code review - does the proposed way forward make sense, or is there a more cesium/entity native way to do this? Probably looking at you @kring or @steve9164 

If we're happy with the approach I'll need to do a bit more work on tidying up other geometry types.

**Comment by kring** [comment]
> This PR attaches the original geojson feature to an entity

I'm not a fan of this approach, because it will only work with GeoJSON. Other vector formats will still have the same problem. And teaching the Leaflet visualizer about specific formats is a step backwards.

> I'd also like us to consider setting worldCopyJump to false in the TerriaViewer.js because currently the geojson disappears if you move the center of the map slightly around the anti-meridian. Not sure if it's set to true for a specific reason though...

I was under the impression we set `worldCopyJump` to true so that you could pan over the anti-meridian and still see data. But this was like 4 years ago so my information could be old or just plain wrong, I don't know.

Ok so what do we do about the anti-meridian? Well, the GeoJSON spec suggests features should be cut at the anti-meridian: https://tools.ietf.org/html/rfc7946#section-3.1.9
I have mixed feelings on that, but in any case we probably can't rely on every file we see doing that.
Similarly, I can't find anything in the spec that says whether longitude values are or are not allowed to be greater than 180 degrees, but some files do that, too.

So I think the best solution is to always connect up lines and polygons along the shortest path. This is exactly what Cesium does. If the last vertex is at lon,lat 175,0 and the next is at -175,0, the shortest path is clearly over the anti-merdian (10 degrees) rather than around the other way (350 degrees). In other words, we can detect that treating -175 as 185 is better and do so, without preserving that it was 185 to begin with.

As a bonus, the GeoJSON will display correctly even if the coordinates are normalized in the GeoJSON. With the approach in this PR, a GeoJSON that has normalized the anti-meridian-crossing longitudes to the -180 to 180 range will display correctly in Cesium but incorrectly in Leaflet.

**Comment by rowanwins** [comment]
Righto thanks for the input @kring - will have a ponder as to the best way to handle that

**Comment by tephenavies** [comment]
If possible we should modify LeafletVisualizer to match Cesium behaviour, preferably without requiring extra data to be present just for Leaflet.

Looks like Kevin made it in first.

**Comment by rowanwins** [comment]
Hmmm this new approach seems suspiciously easy :)

**Comment by rowanwins** [comment]
So I've revised this PR @kring - if you're happy with the approach then it needs to be applied to lines also, and also points potentially...

**Comment by kring** [comment]
Looks good to me! I can't think of how it would apply to points, but I could be missing something.

**Comment by rowanwins** [comment]
Have added polyline support and refactored a smidge.

Re points the annoying thing is if you've got a bunch of points on fiji half will appear on one side of the 2D map and the other half on the other side. However it looks like each point is managed on it's own rather than as part of a feature collection so it's a bit trickier to compare it to previous points to adjust its longitude. eg the feature collection below.

````
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          174.638671875,
          -14.77488250651626
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          192.48046875,
          -16.04581345375217
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          173.32031249999997,
          -21.861498734372553
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          181.05468749999997,
          -18.895892559415024
        ]
      }
    }
  ]
}
````

**Comment by kring** [comment]
Hmm yeah that's annoying. Does Leaflet maybe have a mode that copies the data to the other side of the anti-meridian anytime it's visible, rather than just when the center of the view crosses the anti-meridian?

**Comment by rowanwins** [comment]
Think I've resolved it for points. It looks like each catalog item (eg a set of points) gets put into a leaflet featureGroup which I can access. If I add another point layer to the map it's points are put into a new featureGroup which means layers won't interfere with each other.

So I think this is done now @kring 🤞

File: lib/Map/LeafletVisualizer.js [hunk 1]
```diff
@@ -243,15 +243,22 @@ LeafletGeomVisualizer.prototype.getLatLngBounds = function() {
...
     cartographicScratch
   );
-  return L.latLng(
-    CesiumMath.toDegrees(cartographic.latitude),
-    CesiumMath.toDegrees(cartographic.longitude)
-  );
+
+  let lon = CesiumMath.toDegrees(cartographic.longitude);
+  if (prevLayer) {
+    if (prevLayer._latlng.lng - lon > 180) {
```
kring, line 254: Can we avoid digging into private fields (`_latLng`)? There's no guarantee the next Leaflet version won't remove or change a private field like that.

kring: I'm going to replace it with `getLatLng()` (and also check to make sure that exists). Let me know if you see any problems with that.

kring: Eh on second thought we seem to be using `_latLng` everywhere already. So nevermind, moving on. :)

kring (approved):

PR #3681, opened 2019-09-05
Enable time varying CSVs.

na9da:
WIP: Trying to add time varying CSV.

TODOs
- [ ] Group rows by `idColumns` when specified

### DIFF
--- a/lib/Table/TableColumn.ts
+++ b/lib/Table/TableColumn.ts
@@ -1,5 +1,8 @@
 import countBy from "lodash-es/countBy";
+import sortedUniqBy from "lodash/sortedUniqBy";
 import { computed } from "mobx";
+import JulianDate from "terriajs-cesium/Source/Core/JulianDate";
+import TimeInterval from "terriajs-cesium/Source/Core/TimeInterval";
 import RegionProvider from "../Map/RegionProvider";
 import RegionProviderList from "../Map/RegionProviderList";
 import createCombinedModel from "../Models/createCombinedModel";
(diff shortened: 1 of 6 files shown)


lib/Table/TableColumn.ts
```diff
@@ -224,6 +227,75 @@ export default class TableColumn {
...
+      }
+    });
+    return julianDates;
+  }
+
+  /**
+   * Returns any array of next higher date for each date in `valuesAsJulianDates`.
+   */
+  @computed
+  get stopDates(): ReadonlyArray<JulianDate | null> {
```
kring, line 256: I think the logic here is not quite right. For CSVs with points (not regions), and without an explicit "end time" column, there are a few different variations on "time varying":

1. Each row represents an instantaneous event at a location, e.g. the auto accidents in `wwwroot/test/incidents.csv`. We want to show each incident as a blip as it occurs and continue to show it for a fixed amount of time afterward. We should _not_ stop showing one incident when the next one starts.
2. Each row represents a new value for a persistent thing at a location. For example, the power output by a particular power station at a particular time. In this case, we should show the row's value until we get to the time for the next row _at that same location_. Other locations may have updates in the meantime, but those won't affect that value at this location. In other words, in this scenario, the "stop time" for the row is the time of the next row at the same location, not the next time at any location.
3. Same as 2 except the persistent thing can move, so different rows may have different locations. We know they're the two rows represent the same thing because they have a common ID (i.e. the same value in the column designated as the ID column). In this case, we allow Cesium to interpolate the position in between the times in adjacent rows so that the point moves smoothly between locations.

Distinguishing case 2 from 3 is easy, it's just case there if there's an ID column, otherwise case 2. Distinguishing case 1 from 2/3 is slightly harder. In case 1, most longitude/latitude pairs are present in only one row, whereas in 2/3 most longitude/latitude pairs have 2+ rows.

Note that finding rows with the same ID/(longitude,latitude) is very similar to finding the rows that apply to a region, as done in `valuesAsRegions`.

lib/Table/TableTimeVaryingStratum.ts
```diff
@@ -0,0 +1,30 @@
+import { computed } from "mobx";
+import CsvCatalogItem from "../Models/CsvCatalogItem";
+import LoadableStratum from "../Models/LoadableStratum";
+import StratumFromTraits from "../Models/StratumFromTraits";
+import CsvCatalogItemTraits from "../Traits/CsvCatalogItemTraits";
+import DiscreteTimeTraits from "../Traits/DiscreteTimeTraits";
+import TableColumnType from "./TableColumnType";
+
+export default class TableTimeVaryingStratum extends LoadableStratum(
+  CsvCatalogItemTraits
```
kring, line 10: Probably shouldn't be `CsvCatalogItemTraits`?

lib/Table/TableTimeVaryingStratum.ts
```diff
@@ -0,0 +1,30 @@
...
+) {
+  constructor(readonly catalogItem: CsvCatalogItem) {
```
kring, line 12: Maybe `CsvCatalogItem` here can be more general too? If not, this stratum should be renamed to indicate it is CSV-specific.

kring (changes requested):
Great start!

tephenavies:
@na9da Should this be closed now that #4226 is here?

na9da:
Yes. Closing.

PR #3717, opened 2019-10-08
Implement Chart component using victory charts

na9da:
TODOs
* [x] Derive axis values from chart items
* ~[ ] Moment charts~ will deal in a different PR.


na9da:
Some behaviors that are different from `master`
 * The cursor line just follows the mouse pointer, unlike in master where it jumps to the nearest point. 
 * Instead, the tooltip behaves like the cursor jumping to the closest point

We can either entirely remove the cursor or we can modify the tooltip component to also include a vertical line that looks like the cursor.

na9da:
Closing. See #3726

PR #3719, opened 2019-10-10
Mobx chart colors

na9da:
Adds coloring for chart lines & axes.

### DIFF
--- a/lib/ModelMixins/TableMixin.ts
+++ b/lib/ModelMixins/TableMixin.ts
@@ -26,6 +26,7 @@ import TableColumnType from "../Table/TableColumnType";
 import TableStyle from "../Table/TableStyle";
 import LegendTraits from "../Traits/LegendTraits";
 import TableTraits from "../Traits/TableTraits";
+import getNextChartColor from "../Charts/getNextChartColor";
 
 // TypeScript 3.6.3 can't tell JSRegionProviderList is a class and reports
 //   Cannot use namespace 'JSRegionProviderList' as a type.ts(2709)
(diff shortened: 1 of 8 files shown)


na9da:
This is what it looks like now for multi-axis charts:

image

I am yet to port the line selectors so all chart lines are drawn by default.

Testing notes:

The `column-units` attribute for `<chart>` component has changed. For the sheep station catalog item, it now looks like: `column-units='Capacity#litres,Power#watts,Scale#zig'`. You must explicitly map column names to units.

I had to use `#` as the separator instead of the more intuitive `:` because the html to react parser completely discards the attribute when using `:`.

tephenavies:
@na9da Do you know why html-to-react discards attributes with a colon? What happens if a csv column has either a # or a : in it's name?

lib/ModelMixins/TableMixin.ts
```diff
@@ -194,12 +195,15 @@ export default function TableMixin<T extends Constructor<Model<TableTraits>>>(
...
+          const colorId = `color-${this.name}-${yColumn.name}`;
           const chartData = new ChartData({
             name: yColumn.name,
             categoryName: this.name,
             points,
             units: yColumn.traits.units,
-            color: line.color
+            getColor: () => {
+              return line.color || getNextChartColor(colorId);
+            }
```
tephenavies, line 206: Does this need to be evaluated later? Couldn't it be set when you construct ChartData?

na9da:
@steve9164 Thanks for the review so far. I have consolidated the changes to a new PR #3726. Still in draft, will let you know when it is ready.

lib/ModelMixins/TableMixin.ts
na9da: I think yes. For instance we use a fixed color (white) to show chart items in feature info panel, eagerly allocating a color in this case will use up the pool.

na9da:
> @na9da Do you know why html-to-react discards attributes with a colon? 

Tracked it down to [this call](https://github.com/TerriaJS/terriajs/blob/mobx/lib/Core/markdownToHtml.js#L46) to `DomPurify.sanitize`. It treats `something:x` as an unknown protocol and discards the attribute. Passing the flag `ALLOW_UNKNOWN_PROTOCOLS` will fix it, but should we do that?

> What happens if a csv column has either a # or a : in it's name?

I'll change it so that it will treat the last part as the unit (assuming the unit wont have a `:`).

PR #3720, opened 2019-10-10
fix zooming on momentchart

rowanwins:
Was testing on terriacube and realised that zooming wasnt working on `moment` charts (not to be confused with `momentPoint`).


na9da (approved):
Looks good to me.

KeyboardSounds (approved):

PR #3725, opened 2019-10-15
Update styles dropdown

### LINKED ISSUE #3709: Fix selecting style - add drop down
Currently when a dataset with different styles is selected, it is not intuitive that multiple styles are available. Only when you click on the text in the Style box, users are presented with the list. 
Users could be presented with a drop/down arrow/other tool to understand there is a selection. 

Check this example: https://nationalmap.gov.au/#share=s-hxVJI9RUbTD7d9nRL1POGSmst6U
image

rowanwins:
Resolves #3709 

The style dropdown on a workbench item is currently plain with no indication that it can be changed. This PR adds some basic styling to the element as well as an icon. 

Unfortunately the select element can't have a `:before` or `:after` added to it which may've been an alternate for the dropdown icon, this PR just add it as another element after the `select`.

StyleDropdown

@philipgrimmett requested that we keep the default browser styling for the actual dropdown state.

### DIFF
--- a/lib/ReactViews/Workbench/Controls/StyleSelectorSection.jsx
+++ b/lib/ReactViews/Workbench/Controls/StyleSelectorSection.jsx
@@ -5,6 +5,7 @@ import React from "react";
 import createReactClass from "create-react-class";
 import PropTypes from "prop-types";
 import ObserveModelMixin from "../../ObserveModelMixin";
+import Icon from "./../../Icon.jsx";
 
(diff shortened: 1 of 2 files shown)


rowanwins:
Also checked 
- [x] responsive-ness
- [x] firefox
- [x] safari

kring:
Something strange happening in IE11:
image

For reference, here's what IE11 looks like in master:
image

tephenavies (approved):

PR #3740, opened 2019-10-23
Pass a custom update interval to leaflet tile layers

rowanwins:
The current `CesiumTileLayer.js`, used to put tiles on our leaflet map had a trigger rate of 100ms when required to refresh, this was causing issues for DEA layers behind cloudfront when changing dates rapidly or panning and zooming quickly.

The PR passes an option down via the catalog item config to the ImageryLayerCatalogItem allowing that rate to be set to something higher.

This sample catalog shows it in action
````
{
  "homeCamera": {
    "north": -8,
    "east": 158,
    "south": -45,
    "west": 109
  },
  "corsDomains": ["ows.dea.ga.gov.au"],
  "catalog": [
    {
       "name": "Sentinel imagery",
       "type": "wms",
       "isEnabled": true,
       "leafletUpdateInterval": 500,
       "layers": "s2a_ard_granule_nbar_t",
       "url": "https://ows.dea.ga.gov.au/",
       "featureTimesProperty": "data_available_for_dates"

     }
  ]
}
````
From my initial testing 500ms seems like a reasonable value to prevent the SPDY errors from appearing, although the higher you set the value the more reduce the risk of them happening.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Change Log
 ==========
 
+### next release
+
+* Add ability to pass `leafletUpdateInterval` to an `ImageryLayerCatalogItem` to throttle the number of requests made to a server.
+
+
 ### v7.7.0
 
 * Added a quality slider for the 3D map to the Map panel, allowing control of Cesium's maximumScreenSpaceError and resolutionScale properties.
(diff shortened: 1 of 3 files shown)


AnaBelgun:
reminder to @kring to review this one

kring (approved):

Issue #3707, opened 2019-09-25
"Add data" reworded to "Explore data" 


AnaBelgun:
Based on discussions here: https://docs.google.com/document/d/1Y7Z1EwwkzuBrOgrqowxauJ0-p4v70oVGyoDztOSwoUY/edit and previous user feedback, we should consider changing 
Add Data to Explore Data and reflect that in the workbench tip text: 

_Click 'Explore data' above to:_

- _browse the Data Catalogue_

- _add your data to the map_

TIP: All your selected data sets will be listed here

AnaBelgun:
@kring The wording change when done needs to go into master so all apps will have it when a new version is released

AnaBelgun:
to start with DE Africa and DE australia

AnaBelgun:
rewording of the workbench underneath Explore data: 
![workbench rewording](https://user-images.githubusercontent.com/18452161/67356357-a6e5d580-f5a5-11e9-817c-a139e6b69eb9.jpg)

rowanwins:
@kring - what's your thoughts on implementing some sort of basic language file approach where we provide sensible defaults but allow overriding through configuration? We could start with very lightweight implementation for just a few key spots in the app and slow expand it out?

rowanwins:
Done in #3742

PR #3742, opened 2019-10-23
Very rough approach for implementing language settings

rowanwins:
Related to #3707 where it's suggested to changing the naming of some buttons are parts of the interface.

This approach would see languageOverrides to be passed in from TerriaMap to via the options to `Terria.js`

The main downside I've seen so far is that some text components occasionally include React components like Icons, for example [empty workbench has a message](https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/SidePanel/SidePanel.jsx#L162
) which includes a lightbulb icon. The current approach doesn't render those react components. Someone much better with React would probably have a few ideas how to handle this.

What I've done in `TerriaMap/index.js` is 
````
import languageOverrides from './lib/Language/overrides';

terriaOptions.languageOverrides = languageOverrides;
````
These overrides then get merged into the defaults.

Anyway this is my very naive approach :)

CustomLanguage

### CODE REVIEW

File: lib/ReactViews/SidePanel/SidePanel.jsx [hunk 1]
```diff
@@ -150,19 +150,11 @@ const SidePanel = createReactClass({
...
-                  <li>Browse the Data Catalogue</li>
-                  <li>Load your own data onto the map</li>
-                </ul>
-                <p>
-                  <Icon glyph={Icon.GLYPHS.bulb} />
-                  <strong>TIP:</strong>{" "}
-                  <em>All your active data sets will be listed here</em>
-                </p>
+                <div
+                  dangerouslySetInnerHTML={{
```
soyarsauce: Even though the defaults are loaded in by terria, we should probably use something like https://github.com/cure53/DOMPurify to sanitise this anyway before we forget (e.g. later down the track it gets loaded in from elsewhere)

lib/Language/defaults.js
```diff
@@ -0,0 +1,21 @@
...
+      <div>Your workbench is empty</div>
+      <p>
+        <strong>Click &apos;Add data&apos; above to:</strong>
+      </p>
+      <ul>
+        <li>Browse the Data Catalogue</li>
+        <li>Load your own data onto the map</li>
+      </ul>
+      <p>
+        <Icon glyph={Icon.GLYPHS.bulb} />
```
soyarsauce: This works???! 😮

soyarsauce: Answering my own question: probably not, based off the bulb missing in the screenshot 😆

soyarsauce: Answering even more, I just read the actual PR description. Nevermind me?

soyarsauce:
So there's a few ways around the icon problem, e.g. feeding in an actual react component as part of the "i18n"/language config, or the simplest way for now is to keep the existing components as "defaults" (with icon), then overrides means no icon-ability atm.

lib/Language/defaults.js
rowanwins: You make me laugh @soyarsauce!

rowanwins:
I'll do a quick audit tomorrow of how many places we use components within regular HTML, if it's only the lightbulb icon in that empty workbench effected then I'm happy to have no icon-ability :)

lib/ReactViews/SidePanel/SidePanel.jsx
rowanwins: So I'm wondering how we differentiate between those things which are just text strings vs those which we'll accept a html string for...
Or perhaps as a first cut we should only do strings?

kring:
@rowanwins @soyarsauce why not use https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Custom/parseCustomMarkdownToReact.js to render the localized "strings". Then you can support icons by defining a custom component (https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Custom/registerCustomComponentTypes.js). It sanitizes the input too.

rowanwins:
Good idea @kring - I'll give it a whirlybird!

rowanwins:
Okay @kring & @soyarsauce - the latest commit adds support for Icon's as customComponents.

~~The other component that comes to mind that might be nice to support via custom text is the Button component, eg I'm thinking about the possibility of buttons in that empty workbench area...~~

~~Before I add support for the Buttons do you want to make sure you're happy with that handling for the Icon component?~~
EDIT: We don't use a fancy button component, just regular old buttons with a few classes work fine!

And are we happy to proceed and get this into `master`?

On a sidenote there is also some similar stuff that we do via the config.json in TerriaMap (eg we have a `feedbackPreamble` option - who knew!! At some stage it would be good to rationalise.

rowanwins:
Theoretically the `Icon.jsx` also takes a `style` object as a prop however this would be a smidge fiddily to support as we'd basically be taking a string and converting it to a object. I reckon we just support styling via the `className` prop.

AnaBelgun:
reminder to @soyarsauce to review this one

soyarsauce:
This is awesome work @rowanwins & looks good, hope you don't mind I just extended so we can also use any component/jsx, if the latest commit looks good to you, let's merge it
https://github.com/TerriaJS/terriajs/pull/3742/commits/a7305b0005277d0fdea4ce9be2cb3f113338c0a9

lib/Models/Terria.js
```diff
@@ -35,6 +35,9 @@ var Services = require("./Services");
 var TimeSeriesStack = require("./TimeSeriesStack");
 var ViewerMode = require("./ViewerMode");
 var findIndex = require("../Core/findIndex.js");
+
+var language = require("../Language/defaults.jsx").language;
```
soyarsauce, line 39: I just switched this back to cjs to be consistent with the rest of th efile

rowanwins:
Looks good to me @soyarsauce - it doesn't seem like I can merge it because I can't review it because I opened the PR, but I've looked over it and tested locally and it all makes sense. 
I'll open a corresponding PR over in the TerriaMap repo.

soyarsauce (approved):

soyarsauce:
Sorry @rowanwins I thought I approved w/ my single comment review

PR #3747, opened 2019-10-25
Always initialise the notification window as the top-element.

### LINKED ISSUE #3716: Error modal hidden by catalog explorer panel
This is especially bad when there's an error adding data by URL. The error message is hidden behind the explorer panel, which shows that the added data item is still loading:

Possibly broken by dynamic Z index changes. The error modal should be above any other elements.

rowanwins:
Sometimes a notification window gets hidden by something else which has already been designated as the top element. This PR assigns the class `top-element` which pushes the element to the top.

Resolves #3716

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Change Log
 ==========
 
+### next release
+
+* Ensure the NotificationWindow is rendered as the top element.
+
+
 ### v7.7.0
 
 * Added a quality slider for the 3D map to the Map panel, allowing control of Cesium's maximumScreenSpaceError and resolutionScale properties.
--- a/lib/ReactViews/Notification/Notification.jsx
+++ b/lib/ReactViews/Notification/Notification.jsx
@@ -58,6 +58,7 @@ const Notification = createReactClass({
           type={defined(notification.type) ? notification.type : "notification"}
           width={notification.width}
           height={notification.height}
+          viewState={this.props.viewState}
         />
       )
     );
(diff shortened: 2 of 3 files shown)


mgcsls:
@rowanwins this seems like it's probably a cleaner solution than what I had, but possibly because the notification window is being set as the top element, after dismissing the notification the previous top element is "forgotten"? e.g. go to http://ci.terria.io/make-notification-top/#share=s-iYcd2nM8MvPWE7OXYwbB83Fb9VB > add data > my data > add web data > add a bad url > cancel, and then notice that the feature info window is now on top of the add data window.

soyarsauce:
The whole "forgetting about the previous top element" is part of a bigger problem, as a stopgap I like both your measures, given they both address different things - with a small change to @rowanwins's one to check if there already is a top element, and if it's undefined, then initialise it as the top one, otherwise the logic from @mgcsls's PR should look after the rest.

https://github.com/TerriaJS/terriajs/issues/3958

soyarsauce:
Thanks @rowanwins but after some thought the PR at https://github.com/TerriaJS/terriajs/pull/3957/files is the better option in every (thought out) way, avoids overriding topelement as well as ensures notifications are always on top of the current TopElement. I initially thought I would have a combination of both but it's not actually required once we have that PR in.

Deferring proper solution to https://github.com/TerriaJS/terriajs/issues/3958

soyarsauce:
(If I've forgotten or missed something, please re-open and tell me I'm wrong 🙏)

PR #3754, opened 2019-10-29
Fix splitting filtered items

tephenavies:
A bit of a dodgy fix to splitting filtered items.


tephenavies:
This doesn't even work at all

PR #3755, opened 2019-10-29
Fix location filter split

tephenavies:
Adds a parameter `ignoreSplitter` to public picking functions of Leaflet and Cesium used to deserialise in the filter by location feature.


rowanwins:
Proposed fix works well @steve9164  

Right up until you attempt to share, and then the filter seems to be lost
http://ci.terria.io/fix-location-filter-split/#share=s-klYNO2lOD6DlRiwPzla9T72kmwA


In current master the location filter is retained in a share
https://map.terria.io/#share=s-3NrwtfaDNDv8MzJ3PtswgLHnFh6

tephenavies:
I've fixed sharing the location filter.

rowanwins:
So when I tested this out locally it retained the filter for the original imagery layer, however the filter for the layer created by the split process wasn't keep. 
Filter

tephenavies:
Sometimes it can take up to 20 seconds to repopulate the filter information (all in requests: 8s for GetCapabilities + 12s for GetFeatureInfo), but I haven't seen a case where it hasn't happened. Maybe a UI change is needed to make it clear that it's processing something? Or we could cache these filter requests somewhere and fetch the old results for the new split layer.

KeyboardSounds (approved):

Issue #3274, opened 2019-04-02
Change detection (delta)


AnaBelgun:
Delta functionality (requested by MDBA/GA)
•	To develop a functionality which allows users to visualise changes for an area based on a calculation performed on (same?) layer at 2 different points in time
•	First stage: Sentinel to be used; later to extend to Landsat if possible
•	Roughly the user interaction:
1.	Select the near real time layer (extended to all Sentinel 2 a). For MVP use a calendar picker to select two dates (precursor to this would be zoom to a particular location, query the data for ‘data available for dates’ so that the user is presented with calendar pick options for which data is available.  Long term objective would be to integrate with hydrograph tool so that user can select ‘before and after’ in the context of the hydrograph rather than using a calendar picker
2.	Select the dates
3.	Turn on functionality (yes)
4.	Select (the) separate layer which shows the difference (using colour and the actual values) (yes)
5.	Export the difference as a full resolution (not just screen resolution) GeoTIFF to be used in own GIS and other systems (yes)

Some Q&A:
-	Does Terria need to perform a calculation per se (run an algorithm) (yes, this would be image array for date 1 – image array for date 2 calculation.  To add to the complexity here, the image differencing would be performed on data that need to be calculated on the fly.  For example the web services provide access to NIR and red band surface reflectance measurements.  So the client would need to a. calculate NDVI (NIR-RED/NIR+RED) for both dates, and then perform an image difference between the two dates (NDVI date 1 – NDVI date 2)
-	Is the difference something like: I see for this location an area coloured in blue on 1 Dec and for the same location, on 1 Jan, same area is coloured in red; a new layer will show that for such an area, there are changes in colours? - this is what the user is expecting

Example: 
NDMI 25/4/19
![NDMI 25-04-19](https://user-images.githubusercontent.com/18452161/57827597-429b3a80-77eb-11e9-9f4e-de632672ff42.png)

 
NDMI 05/05/2019
![NDMI 05-05-19](https://user-images.githubusercontent.com/18452161/57827603-46c75800-77eb-11e9-8283-aac14b5f7d1d.png)

 
NDMI Difference 05/05/2019 minus 25/04/2019
 
NDMI difference > 0.15 as red grid cells converted into 'uncleaned' polygons. Centre pivot polygons 'of interest' selected (light blue)
![NDMI difference uncleaned polygons](https://user-images.githubusercontent.com/18452161/57827710-bb9a9200-77eb-11e9-8317-abe67fc26132.png)

[...]

PR #3762, opened 2019-10-31
Delta feature

na9da:
Implements #3274.

Phils UX designs for the feature - https://projects.invisionapp.com/share/CASND0RE5YT#/screens/370254676

Test catalog file: https://gist.githubusercontent.com/na9da/b41a9a55f6c41a807fed450a3815f84d/raw/05fb17f3f5647caf2cd68be89523f4bde77cc9f9/delta-feature-test-catalog.json


AnaBelgun:
@na9da is this ready for review?

na9da:
@AnaBelgun 

So far, I have implemented most of Phils designs. But, I think before reviewing this we need to integrate the real API and get feedback from GA.

soyarsauce:
Is the delta service accessible outside the network?

edisonguo:
@soyarsauce the demo server has restrictions on IP ranges. Please use the following public endpoints.

The general steps to test delta are as follows:
1) Open a layer from the data panel
2) Select either of the these two styles - `Normalised Difference Vegetation Index (NDVI)` or `Modified Normalised Difference Water Index (MNDVI)`. As requested by Leo, only these two styles have delta enabled.
3) Click on the `delta` link and generate the delta images

```
{
    "catalog": [
        {
            "name": "Delta Blended service",
            "type": "wms",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Sentinel2 delta service",
            "type": "wms",
            "layers": "sentinel2_nbart_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Daily Landsat 8 NBART delta service",
            "type": "wms",
            "layers": "landsat8_nbart_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Daily Landsat 8 NBAR delta service",
            "type": "wms",
            "layers": "landsat8_nbar_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Daily Landsat 7 NBART delta service",
            "type": "wms",
            "layers": "landsat7_nbart_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Daily Landsat 7 NBAR delta service",
            "type": "wms",
            "layers": "landsat7_nbar_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Daily Landsat 5 NBART delta service",
            "type": "wms",
            "layers": "landsat5_nbart_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },
        {
            "name": "Daily Landsat 5 NBAR delta service",
            "type": "wms",
            "layers": "landsat5_nbar_daily",
            "ignoreUnknownTileErrors": true,
            "opacity": 1,
            "featureTimesProperty": "data_available_for_dates",
            "supportsDeltaComparison": true,
        },

    ]
}
```

lib/ReactViews/Tools/DeltaTool/DeltaTool.jsx
```diff
@@ -0,0 +1,221 @@
...
+  );
+}
+
+PrettyLocation.propTypes = {
+  location: PropTypes.object.isRequired
+};
+
+PrettyLocation.displayName = "PrettyLocation";
+
+function duplicateItem(item) {
```
soyarsauce: I know it's for ease but we've got the exact same code so might as well avoid duplication here - let's pull out the `splitItem()` fn from ViewingControls, add some controls via parameters/options on whether to do other side effects (`addUserCatalogMember` etc etc)

lib/ReactViews/Tools/DeltaTool/DeltaTool.jsx
```diff
@@ -0,0 +1,221 @@
...
+
+function dateParamFormat(date) {
+  return dateFormat(date, "yyyy-mm-dd'T'HH:MM:ss.l'Z'", true);
+}
+
+function dateDisplayFormat(date) {
+  return dateFormat(date, "dd-mm-yyyy", true);
+}
+
+module.exports = DeltaTool;
```
soyarsauce: (PS all our JSX files can safely use import/export as it's not used by the doc generation)

lib/ReactViews/Tools/DeltaTool/DeltaTool.jsx
```diff
@@ -0,0 +1,221 @@
...
+  const [item, setItem] = useState(undefined);
+  const [location, setLocation] = useState(undefined);
+  const [primaryDate, setPrimaryDate] = useState(catalogItem.discreteTime);
+  const [secondaryDate, setSecondaryDate] = useState(catalogItem.discreteTime);
+  const [pickerMessage, setPickerMessage] = useState(
+    "Select a point by clicking on the map"
+  );
+
+  // Duplicate the catalog item
+  useLayoutEffect(() => {
```
soyarsauce: Does this need to happen before paint? (e.g. can it be moved into the earlier `useEffect`?) I'd rather use placeholder things and _then_ do any duplication required, than block things with this.

lib/ReactViews/Tools/DeltaTool/delta-tool.scss
```diff
@@ -0,0 +1,112 @@
...
+  padding: $padding;
+}
+
+.buttons {
+  display: flex;
+}
+
+.cancelBtn {
+  composes: btn-grey from "../../../Sass/common/_buttons.scss";
+  width: 10vw;
```
soyarsauce: Can you change this to em please, it doesn't need to change width awkwardly on width-resizing

soyarsauce (changes requested):
@na9da
Works really great Nanda, awesome work!!
A few things
- merge i18n/l10n changes from master & i18-ify the strings
- add the newly created catalog item on top of workbench, so that any 100% opacity layers don't block the view - the user's explicitly asked to look at delta things so it would be confusing if it wasn't showing without yet another action
- clear out the error-state-picker-message when user tries to select another location within bounds
- ... IE11 😩😩😩 I tried to test this in IE for you but the delta service was too slow & wouldn't load in the VM
- disable the delta button when `useSmallScreenInterface` is true - better to do that than have a broken menu where a mobile user would get stuck in that mode:

image

soyarsauce:
Also, was there a decision to not use the "Update Difference Map" in the designs?

na9da:
> Also, was there a decision to not use the "Update Difference Map" in the designs?

@soyarsauce I might have missed it, or Phil updated it - I don't remember!

Anyhow, we need it for v2.

lib/ReactViews/Tools/DeltaTool/DeltaTool.jsx
na9da: Have extracted the common bits to a new shared import. Not very configurable atm, since much of the things like `addUserCatalogMember` is call site specific.

soyarsauce (changes requested):
Thanks @na9da - as discussed
- seems to not be working in IE11, see if you can quickly narrow down why
- chunk it via dynamic imports && `SplitPoint.jsx`

na9da:
@soyarsauce 

1) The IE issue was because I was using `array.find()` to filter the picked features. Have replaced it with a `for` loop.

2) Added dynamic imports to defer loading the delta tool.

soyarsauce:
Awesome work @na9da - delta all loads dynamically on demand as expected. The only additional feedback I have left for a v2 thing, is that you may want to have the WMS as a separate item that only has styles which have the delta enabled - I tripped on that for a bit when it defaulted to false colour, and I hadn't switched it to `Normalised Difference Vegitation Index` again.

soyarsauce (approved):

soyarsauce:
Just saw https://github.com/TerriaJS/terriajs/issues/3274#issuecomment-561988985 - now I see why the default has been switched.

soyarsauce:
delta test for masteR:
http://ci.terria.io/master/#https://gist.githubusercontent.com/na9da/b41a9a55f6c41a807fed450a3815f84d/raw/05fb17f3f5647caf2cd68be89523f4bde77cc9f9/delta-feature-test-catalog.json

soyarsauce:
(Nevermind - delta not served off that IP anymore)

soyarsauce:
http://ci.terria.io/master/#https://gist.githubusercontent.com/soyarsauce/ad7319d399bf6d743bf77e1c48474c5c/raw/ae95ab1e5a40da6354f8ceff554e00ca58aba3c7/delta.json

PR #3763, opened 2019-11-01
Bug fix for expanding chart

rowanwins:
Prevent a bug if a chart was expanding where a lat-lon wasn't available. This PR only create a marker where one can be created.

The error would occur when expanding a chart on a CSV item with lon-lon columns.


soyarsauce:
Restarting tests

soyarsauce:
Restarting CI

na9da (approved):

PR #3790, opened 2019-11-19
More tweaks for running in node.js.

kring:
* Move the language override stuff from Terria.ts to ViewState.ts. Putting it in Terria.ts means the model layer has a dependency on React and the ReactViews layer, which is no good.
* Use terriajs-cesium v1.63.2.
* Add a prepare npm script to build terria when it's installed as a git dependency and when npm publishing it.


soyarsauce:
@kring FYI

> Checking formatting...
> lib/ReactViews/HelpScreens/HelpMenuPanelBasic.jsx
> Code style issues found in the above file(s). Forgot to run Prettier?

soyarsauce:
+ as mentioned, am totally on board with moving the language stuff to viewstate.

soyarsauce (approved):

PR #3796, opened 2019-11-20
Add moment charts.

na9da:
Implements the following:

- [x] Moment lines chart
- [x] Moment points chart


na9da:
Zoom performance is very poor at this point for large datasets (> 300 points). This is a known problem for victory charts. [Victory docs](https://formidable.com/open-source/victory/guides/zoom-on-large-datasets) suggests implementing domain clipping and downsampling as solutions to manage this problem. 

The downsampling algo given in the docs does not faithfully preserve the shape of the original full-data chart. [LTTB](https://datylon.com/blog/sampling-time-series-data-sets) algo looks like a better alternative.

One problem still remains even after implementing LTTB. Zooming will result in the chart line fluctuating a bit because of choosing slightly different points at different zoom levels.

TODOs:
- [x] Domain clipping
~- [ ] LTTB Downsampling~

na9da:
Have improved performance by passing [pre-formatted data](https://github.com/FormidableLabs/victory/pull/1212) to victory. So not implementing downsampling for now.

PR #3797, opened 2019-11-20
Mobx - Keyboard interface v 1.0

KeyboardSounds:


tephenavies (changes requested):
Looks good. The feature selection in 2d worked well for me. There's a few things to fix:
- [ ] slight increase in padding or margin in a lot of buttons, sometimes making button text off-centre or making icons smaller
- [ ] the focus border on the map window shows when you interact with the data explorer (and the focus border also moves the map)
- [ ] feedback doesn't work

Issue #3785, opened 2019-11-18
Internationalization of terria


zoran995:
Starting this so we can track work on internationalization of app, hope it is ok

zoran995:
@rowanwins and @soyarsauce How we should proceed with data in `defaults.jsx`? I am not sure is it possible to keep it as it is with `i18next`

soyarsauce:
We should continue the rest of the internationalisation efforts in the work done through https://github.com/TerriaJS/terriajs/pull/3787, and stop using `defaults.jsx` for i18n - however it could still be useful for defining other ways to extend terria (see https://github.com/TerriaJS/de-australia-map/pull/29/files#diff-e7c237f2b6a10131a063f7b5740c39aaR5)

zoran995:
I will post some questions:
- language switch:

    - Where should we put a button for users to switch languages of the map in the `TerriaMap` or `TerriaJS`? The user should have an option to define languages that he want to enable. The switcher should not be visible by default if there is only one language in the `Language` folder.
    - From the design perspective dropdown or something else, and where to position it? 
    - Should we show the country flag next to language name (if yes then user should also have option to choose that)? 
- should we split `translation.json` file to multiple parts to separate ReactViews translation from the rest (`core`, `models` and `map` which are biger groups in the `translation.json`)

PR #3807, opened 2019-11-25
Extending i18next support

zoran995:
Continuing the work from #3787, tracking of this is done through #3785 
Most of the `ReactViews` components moved to support i18next, there is some problems with use of i18next in `DropDownNotification` so I am not adding it for now.


kring:
Thanks for all your work on this @zoran995! Can you please also add yourself to https://github.com/TerriaJS/terriajs/blob/master/CONTRIBUTORS.md?

zoran995:
Some part of developer errors are also included in `translation.json`. Still need to go through Models folder, there is a lot of data mainly errors that need to be included. Hope to complete that today/tommorow.
@kring will do that with next commit, if it's ok?

lib/Models/Terria.js
```diff
@@ -86,12 +85,6 @@ var Terria = function(options) {
     throw new DeveloperError("options.baseUrl is required.");
   }
 
-  options.languageOverrides = defined(options.languageOverrides)
-    ? options.languageOverrides
-    : {};
-
-  this.language = combine(options.languageOverrides, language);
```
soyarsauce, line 93: As mentioned in https://github.com/TerriaJS/terriajs/issues/3785#issuecomment-557305148 we should still leave this in, as the overrides are still used for function-specific-overrides

lib/ReactViews/WelcomeMessage/WelcomeMessage.jsx
```diff
@@ -54,19 +54,12 @@ export const WelcomeMessagePure = ({
...
-    WelcomeMessagePrimaryBtn,
-    WelcomeMessageSecondaryBtn,
-    WelcomeMessageDissmissText,
-
-    WelcomeMessagePrimaryBtnClick,
-    WelcomeMessageSecondaryBtnClick
-  } = viewState.terria.language;
-
+  const WelcomeMessagePrimaryBtnClick = undefined;
+  const WelcomeMessageSecondaryBtnClick = undefined;
```
soyarsauce: Keep these loaded in from overrides, otherwise it will break expected behaviour for maps using this

soyarsauce (commented):
Awesome work :100: - had a brief look and wanted to mention that, we need to make sure the existing `lib/Language/defaults.jsx` behaviour still exists - moving any of the language translations over out from there, and keep the "function as value" keys like `WelcomeMessagePrimaryBtnClick` - perhaps also rename `lib/Language/defaults.jsx`  to something more appropriate suggesting hooks/overrides of behaviour instead of language

lib/ReactViews/Guide/Guide.jsx
```diff
@@ -77,7 +78,8 @@ export const analyticsSetShowGuide = (
     setCalledFromInside: false
   }
 ) => {
-  const openOrClose = bool ? "Open" : "Close";
+  const { t } = this.props;
```
tephenavies: This throws an error

lib/Models/Terria.js
zoran995: can you please check if this part is OK now, after last commit

lib/ReactViews/Guide/Guide.jsx
zoran995: should be fixed with the last commit

lib/Models/Terria.js
soyarsauce: Thanks @zoran995 ! That's exactly what we need.

soyarsauce:
@zoran995 Thanks for addressing the overrides part, @steve9164 will look over the rest.

tephenavies:
Not sure if it's useful to you but I made [this script](https://gist.github.com/steve9164/0ff1a01ef6ed7cc02d81b2d4240eec0c) to test translation strings being correctly replaced in terria and the UI being redrawn on language change. Modify for different location of [`words`](https://en.wikipedia.org/wiki/Words_(Unix)). It's dumb and screws up translations that use templating or `<Trans>`, but it gives some idea of what is reactive.

tephenavies (approved):

PR #3811, opened 2019-11-26
Add min-height to My Data drag drop box

### LINKED ISSUE #3223: My Data: apply min-height to dragDrop box
### Problem
Currently the dragDrop box that appears in the My Data catalogue scales in height according to how many datasets have been added, but does not have a set min-height. Without a min-height the entire box collapses, providing limited functional value.

### Steps to recreate
1. nationalmap.gov.au > My Data > Add web data > https://ows.services.dea.ga.gov.au/
2. Note squished dragDrop element

### Screenshot
screen shot 2019-02-25 at 2 10 18 pm

mgcsls:
Resolves #3223


soyarsauce:
As discussed, we won't add this in until it doesn't make IE11 worse off, which it does with the newly added min-height

soyarsauce (changes requested):
Hi @mgcsls - the IE11 is fixed!
However the drag drop box has regressed in other areas - 
image

vs current
image

soyarsauce (approved):
Amazing!

PR #3813, opened 2019-11-26
Include full catalog item path in catalog event labels

### LINKED ISSUE #2795: Include full catalog item path in Google Analytics events
For AREMI, we pretty regularly have people asking us for specific usage numbers for their layers or layer groups. At the moment, it's very hard or impossible to provide these with any certainty of their accuracy. It would be helpful if the event labels for catalog items were more descriptive so we can better filter events. One way to improve this would be to make the labels for each event include the catalog path as a prefix.

I'm not familiar with what's possible with GA, so there might be reasons for not doing it this way, and  am open to alternatives which achieve the result we want.

mgcsls:
Resolves #2795


soyarsauce (approved):
Resolves https://github.com/TerriaJS/terriajs/issues/2795

PR #3838, opened 2019-12-02
Remove duplicate vehicle positions from GTFS feed

rowanwins:
When a GTFS update is retrieved it can many records for a single vehicle. This PR slims down the data by ensuring we only use the most recent record for a vehicle based on the timestamp.

Here is a sample catalog item
````
    {
        "name": "Ferries",
        "url": "https://api.transport.nsw.gov.au/v1/gtfs/vehiclepos/ferries",
        "type": "gtfs"
      }
````

### DIFF
--- a/lib/Models/GtfsCatalogItem.ts
+++ b/lib/Models/GtfsCatalogItem.ts
@@ -93,24 +93,24 @@ export default class GtfsCatalogItem extends AsyncMappableMixin(
     // so we'll only display the newest record.
     // Although technically the timestamp property is optional, if none is
     // present we'll show the record.
-    const vehicleMap = new Map()
+    const vehicleMap = new Map();
     for (var i = 0; i < feedEntities.length; ++i) {
-      const entity: FeedEntity = feedEntities[i]
-      const item: VehicleData = this.convertFeedEntityToBillboardData(entity)
+      const entity: FeedEntity = feedEntities[i];
+      const item: VehicleData = this.convertFeedEntityToBillboardData(entity);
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/Models/GtfsCatalogItem.ts [hunk 1]
```diff
@@ -88,15 +88,29 @@ export default class GtfsCatalogItem extends AsyncMappableMixin(
...
+    // Sometimes the feed can contain many records for the same vehicle
+    // so we'll only display the newest record.
+    // Although technically the timestamp property is optional, if none is
+    // present we'll show the record.
+    const vehicleMap = new Map()
+    for (var i = 0; i < feedEntities.length; ++i) {
+      const entity: FeedEntity = feedEntities[i]
+      const item: VehicleData = this.convertFeedEntityToBillboardData(entity)
+
+      if (item && item.position && item.featureInfo) {
```
KeyboardSounds, line 101: I'm not a huge fan of these sorts of undefined checks but that's pure personal preference and you should ignore me if you don't agree
in general, things could evaluate to false and still be defined

KeyboardSounds (approved):

lib/Models/GtfsCatalogItem.ts
rowanwins: Is using the `defined` method from `"terriajs-cesium/Source/Core/defined"` more appropriate @KeyboardSounds ?

rowanwins: Couldn't make typescript happy using the `defined` method :(

PR #3842, opened 2019-12-02
Fix closing of search in mobx

rowanwins:
- use a mobx deployment
- search location (Penrith or any other)
- select one of the search results
- click on Done
- list doesn't close when clicking on Done

A mobx action error was thrown so this resolves it

Related to https://github.com/TerriaJS/nsw-digital-twin/issues/157

### DIFF
--- a/lib/ReactViews/Search/SidebarSearch.jsx
+++ b/lib/ReactViews/Search/SidebarSearch.jsx
@@ -1,5 +1,6 @@
 import createReactClass from "create-react-class";
 import { observer } from "mobx-react";
+import { runInAction } from "mobx";
 import PropTypes from "prop-types";
 import React from "react";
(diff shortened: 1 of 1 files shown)


lib/ReactViews/Search/SidebarSearch.jsx
```diff
@@ -19,7 +20,9 @@ const SidebarSearch = observer(
     },
 
     backToNowViewing() {
```
na9da, line 22: Both are equivalent, but it might be prettier to use the `@action` decorator here which you can import from `mobx-react`, just like `@observer`.

na9da (approved):

kring:
Ah sorry Rowan I didn't see this PR before I did the same last night. Feel free to merge your own PRs into the `mobx` branch, especially when they're simple like this.

rowanwins:
All good :)

PR #3850, opened 2019-12-03
Menu bar fix #3844

### LINKED ISSUE #3844: MenuButton style not changing to inactive
The style of MenuItem doesn't change to notActive in several cases: 

- when MenuItem button is clicked, the button color change to blue ($color-primary) as it is suppose to and panel appears, but on second click on the same button it closes the panel but color of the button remain blue while it should return to color of inactive button ($charcoal-grey).
- Also while the panel is open if click happen somewhere on the map it closes the panel but color again remains like it MenuItem is active.
- In the case the click happen somewhere inside of the panel color of the button turns to inactive ($charcoal-grey) but panel is still open.
- When the panel is closed with `ECS` key, button again has active color.
- When the story panel is open the only way to close it is by clicking the story MenuItem, and it is odd to see that button remain active

Click on the close mark in the panel closes the panel and change the color as it is supposed to do.
story
share-print

zoran995:
Fixing the issues defined in #3844, and additionally fixing the feature info collapse button to have $color-primary only on hover, so it doesn't stay return to original color after click. Focus was making a problem in all cases. I think this is in accordance with your concepts @philipgrimmett, at least I noticed that buttons stay like that in your invision concepts

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,18 @@
 Change Log
 ==========
 
-### next minor
 
+### Next release
+* Fixed a bug where calling `openAddData()` or `closeCatalog()` on ViewState did not correctly apply the relevant `mobileViewOptions` for mobile views.
+* Fixed filter by available dates on ImageryLayerCatalogItem not copying to the clone when the item is split.
 * Fixed an error in `regionMapping.json` that causes some states to be mismatched when using Australian state codes in a column labelled "state". It is still recommended to use "ste", "ste_code" or "ste_code_2016" over "state" for column labels when matching against Australian state codes.
 * Fixed bug where "User data" catalog did not have add-buttons.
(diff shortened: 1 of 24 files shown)


soyarsauce (changes requested):
Hi @zoran995 - thanks for putting this in, understanding it's in the spirit of matching design concepts. However it certainly makes it worse off for keyboard-only-interactions, as focus styles are now not consistent with hover styles, and we aim to make all modes of interaction work well when and where we can.

If you would still like to make this better
- keep the isOpen styling logic (good that it stays in 'active' when open)
- keep the focus styles to match any hover styling

zoran995:
Restored focus options for styling.

soyarsauce (approved):
@zoran995 great! This makes terriajs no worse-off, and improves the open-or-not state of the menu bars.

PR #3853, opened 2019-12-04
Filter out times that cause fromISO8601 to throw

tephenavies:
Primarily for non-standards conforming WMS.


soyarsauce (approved):
I feel like I've made almost exactly the same change on this file for terrace earlier on, but don't see where it is.

PR #3862, opened 2019-12-05
Fix incorrect i18n pathing for welcome message

soyarsauce:
Spatial data made what?!
image


soyarsauce:
There are probably a few more throughout but this one is most pressing as it breaks the welcome message messaging.

tephenavies (approved):

Issue #2856, opened 2018-03-01
Saved views, "Back to previous view", "Forward to next view"


AnaBelgun:
User feedback:

One thing I use NationalMap for a lot is trying to identify topographic features (mostly mountains) in photos I have taken. This requires zooming in and tilting to simulate the view in the photo (worm's eye view), then zooming out and returning to a bird's eye view to read the names of features from a topographic map layer. It would be really useful to have 'back to previous view' and 'forward to next view' buttons to allow quick and easy shifting between perspectives.

Share URL: https://nationalmap.gov.au/#share=s-zMD4AF8ymd0yjvSes4LaZP484y2

rowanwins:
Consensus to move this to a plugin in #3863

PR #3863, opened 2019-12-05
Saved views, "Back to previous view", "Forward to next view" #2856

zoran995:
Creating 'back to previous view' and 'forward to next view' buttons to allow quick and easy shifting between perspectives as defined in #2856. For now this is available only in 2D mode, but it would be nice to hear opinions about the work done.  
Further design is needed, I used dummy approach for design (buttons for zoomIn and zoomOut are used)


zoran995:
This is ready for review

rowanwins:
Here is a screengrab of what the UI looks like associated with this PR.
BackForward

On to the general concept - to be honest I'm not sure I like this idea. Having come from GIS software world (ArcMap & QGIS) where they have these sorts of buttons I think I can count the number of times I've used them on one hand over a decade. As a user I have no idea where the backwards or forwards buttons are going to take me and so inevitably I don't use them. IMHO it clutters the UI for not much gain.

That said I'm only one vote :)

tephenavies:
Hi @zoran995. Thanks for the pull request! We're having some discussions on whether this is a good fit to be placed in the core terriajs library, or if there's somewhere else we should put it for other open source users. I'll keep you updated.

rowanwins:
Hi @zoran995 

Consensus among the team is that we'd prefer if this wasn't in master as it 
1. takes up space on the UI 
2. is a pretty niche requirement, and 
3. the behaviour isn't super obvious for the user 

That said we think this would make a good candidate for a plugin/component that people could use in their terria instances. So with that in mind we are proposing that this lives in a standalone repo that people could use if they wanted. We're thinking the user could inject the componenet via their TerriaMap instance in the [UserInterface.jsx](https://github.com/TerriaJS/TerriaMap/blob/master/lib/Views/UserInterface.jsx). Does that make sense and sound achievable?

Cheers

rowanwins:
Recommendation acknowledged by contributor as a thumbs up response :)

PR #3868, opened 2019-12-09
Apply prettier for translation files

soyarsauce:


soyarsauce:
Resolves https://github.com/TerriaJS/terriajs/issues/3869

zoran995:
.prettierignore should be also updated so prettier doesn't ignore translation files
`!lib/Language/**/*.json`

soyarsauce:
Oops that was actually the original plan, but missed my commit. Thanks for picking up on that

tephenavies (approved):

PR #3886, opened 2019-12-15
Don't Merge - Add model for Senaps locations

rowanwins:
This PR adds a new model for senaps locations

~~One issue I've got is that the senaps API requires you to pass an `apikey` param, eg `apikey=12324`. I spent a while trying to get this working in terriajs-server but didn't have any luck but didn't have any joy, it would be nice to revisit that so that the api-key could be attached by the server.~~

A typical config looks something like
````
    {
      "name": "Senaps Locations",
      "type": "senaps-locations",
      "locationIdFilter": "*boorowa*",
      "streamIdFilter": "*SHT31DIS_ALL*,*environdata*"
    }
````

I've also provided options for
- `locationIdFilter` and `streamIdFilter` follows the senaps id filter syntax (comma seperated for multiple queries, and * for wildcard)
- style which gets passed down to the `GeoJsonCatalogItem` options

### CODE REVIEW

**Comment by rowanwins** [comment]
Have done some work on the apikey issue in terriajs-server (https://github.com/TerriaJS/terriajs-server/pull/113) so once that gets merged I'll revisit this PR

**Comment by rowanwins** [comment]
So with the revised terriajs-server approach the apikey can either be managed via the `serverconfig.json` with something like
```
    "appendParamToQueryString": {
        "senaps.io": [
            {
                "regexPattern": "locations",
                "params": {
                    "apikey": "123"
                }
            }
        ]   
    }
````
Or conceivably via the Authorization header, but I could never get that approach to work

AnaBelgun:
This PR work was used for Senaps data in the CSIRO - BARS project. it will not affect any other projects going forward, because superseded by MobX work. 
The MobX version is here: https://github.com/TerriaJS/terriajs/pull/4246

rowanwins:
Might reopen until serenity is on mobx just in case

rowanwins:
Closing as serenity is now running on mobx

PR #3899, opened 2019-12-19
Mobx: Fix display of projected coordinates in LocationBar

### LINKED ISSUE #3891: Changing coord presentation fails in mobx
Clicking on the coord bar in a mobx app doesn't do anything. Throws an error related to `toggleUseProjection` method.

Coords

rowanwins:
Fixes when you click on the LocationBar - now it correctly shows the projected coordinates.

Resolves #3891


lib/ReactViews/Map/Legend/LocationBar.jsx
```diff
@@ -23,7 +24,9 @@ const LocationBar = observer(
     },
 
     toggleUseProjection() {
-      this.props.mouseCoords.toggleUseProjection();
+      runInAction(() => {
```
na9da: Just a small nitpick - It might be better to wrap `MouseCoords.toggleUseProjection` with an `@action` decorator instead.

PR #3904, opened 2019-12-19
Replace static method in parameter editor components with plain functions

### LINKED ISSUE #3902: Catalog functions broken because `withTranslation` does not re-export static methods.
Following discussion with @soyarsauce and @steve9164 the fix is to rewrite the static methods as plain functions and export them individually.

na9da:
So that we can pass them to higher-order-components like `withTranslation` without breaking code.

Fixes #3902


rowanwins (approved):
Looks good, now works as expected

PR #3905, opened 2019-12-19
Fix infinite recursive loop when processing `PropertyBags`.

### LINKED ISSUE #3903: Cannot select an existing polygon from a catalog function
Steps to recreate the problem:
* Open this geo-rapp share [link](https://map.geo-rapp.org/#share=s-AqeWeWMD4YPVw0pxAfmgUHpXp3p)
* Go to 'Add Data > Analysis Tools'
* Select `Existing Polygon` from the catalog function form
* Pick any polygon on the map
* Nothing happens (exception logged in the console)

This is happening because of an infinite recursion [here](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/GeoJsonCatalogItem.js#L200); when the method receives a `Cesium.PropertyBag` and it has a reference loop.

Branches:
- [x] master
- [ ] mobx

We also need to fix #3902 before we can roll out the fix to geo-rapp.

na9da:
Fixes #3903

### DIFF
--- a/lib/Models/GeoJsonCatalogItem.js
+++ b/lib/Models/GeoJsonCatalogItem.js
@@ -17,6 +17,9 @@ var loadBlob = require("../Core/loadBlob");
 var loadJson = require("../Core/loadJson");
 var PolylineGraphics = require("terriajs-cesium/Source/DataSources/PolylineGraphics")
   .default;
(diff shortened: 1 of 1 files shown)


rowanwins (approved):
👍

PR #3920, opened 2020-01-10
Mobx BuildShareLink, detect when a model is not shareable

### LINKED ISSUE #3798: Model architecture: Add distinction between shareable and non-shareable models
Currently user stratum traits of models created by drag-and-drop or "uploading" will be serialised, but not the data. We should instead not serialise any of it. In the previous model this distinction was made based on whether there was local data detected in the member.

The distinction will also allow us to reintroduce the share menu caveat message about which datasets can't be shared.

mgcsls:
Resolves #3798


mgcsls:
See #3929

PR #3923, opened 2020-01-13
Fix bug with momentPoints chart type 

rowanwins:
Fixes a bug with momentPoint chart when plotting against series with null values, the points would end up in no mans land because they wouldn't know where to put the y value.


na9da (approved):
Looks good to me!

PR #3927, opened 2020-01-13
WFS autodetect format, Name, zoomToExtent

### LINKED ISSUE #3926: WFS autodetect format, Name, zoomToExtent
There is several issues when adding WFS from GeoServer instance ([geosolutions official demo](https://demo.geo-solutions.it/geoserver/web/)): 
1) on autodetect it is detected as WMS, not WFS, this is because GeoServer by default returns link in form of `ows?service=wfs`, [link](https://demo.geo-solutions.it/geoserver/ows?service=wfs&version=1.1.0&request=GetCapabilities), and in terria we detect `/wfs`, check the [`registerCatalogMembers`](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/registerCatalogMembers.js#L175-L179), after that, few fallbacks are added ([link](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/registerCatalogMembers.js#L221)). The first fallback is WMS and it loads like that. Maybe we should also register an option `=wfs`, and also for the WMS (`=wms`), to exclude fallback in this case.
2) if a user explicitly chooses WFS the Title(name) of the loaded group is not the Title of the service but the URL of the service. Currently, we don't try to set the Title of the WFSCatalogGroup, fix me if I am wrong.
3) when the service is added as a WFS, the Zoom To Extent button is not available. The rectangle of WFSCatalogItem is destroyed [here](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/WebFeatureServiceCatalogItem.js#L277).

zoran995:
resolves #3926


lib/Models/registerCatalogMembers.js
```diff
@@ -177,11 +177,21 @@ var registerCatalogMembers = function() {
     WebFeatureServiceCatalogGroup,
     true
   );
+  createCatalogItemFromUrl.register(
```
rowanwins, line 180: Rather than registering the item again could we do improve the regex on the first registration so it captures both options, eg `(\/wfs|=wfs)` - that seems better than registering the item twice...

lib/Models/registerCatalogMembers.js
```diff
@@ -177,11 +177,21 @@ var registerCatalogMembers = function() {
...
+    WebFeatureServiceCatalogGroup,
+    true
+  );
   createCatalogItemFromUrl.register(
     matchesUrl(/\/wms/i),
     WebMapServiceCatalogGroup,
     true
   );
+  createCatalogItemFromUrl.register(
+    matchesUrl(/\=wms/i),
```
rowanwins: Same as above, I think it would be preferable to use a single regex to capture both options.

rowanwins (changes requested):
A couple of minor comments on the registration, but the zoom to and naming looks greats!

rowanwins (approved):

PR #3929, opened 2020-01-14
Mobx sharing user added data

### LINKED ISSUE #3602: Model Architecture Bug: Share with user added data broken
Steps to reproduce
1. Hit http://ci.terria.io/mobx/
2. Click "Add data" in the workbench
3. My Data tab -> add any data (local + web has same effect) (`http://ereeftds.bom.gov.au/ereefs/tds/wms/ereefs/mwq_gridAgg_P1A` for a test case)
4. Close explorer window (Click done)
5. Click Share/Print button
Observe white page

```
SharePanel.jsx:305 Uncaught TypeError: Cannot read property 'map' of undefined
    at Object.renderWarning (SharePanel.jsx:305)
    at Object.renderContentWithPrintAndEmbed (SharePanel.jsx:360)
    at Object.renderContent (SharePanel.jsx:324)
    at Object.SharePanel_render [as render] (SharePanel.jsx:493)
    at ObserveModelMixin.js:34
    at Function.xd (knockout-3.5.0.js:55)
    at Function.yd (knockout-3.5.0.js:55)
    at Function.ha (knockout-3.5.0.js:54)
    at Object.a.o.a.$ (knockout-3.5.0.js:52)
    at ObserveModelMixin.js:25
```

mgcsls:
Resolves #3602 and #3798


soyarsauce:
@steve9164 are you able to look at this this week?

lib/Models/Terria.ts
```diff
@@ -401,15 +402,20 @@ export default class Terria {
...
+          filterOutUndefined(initSources).reduce(
+            (promises: any, initSource) => {
+              return [
+                ...promises,
+                this.applyInitData({
+                  initData: initSource
+                })
+              ];
+            },
+            [Promise.resolve]
```
tephenavies: Probably meant to be `Promise.resolve()`

tephenavies (commented):
Looks good. Needs merge conflicts resolved and some discussion on user/definition strata.

lib/Models/createCatalogItemFromFileOrUrl.ts
```diff
@@ -166,6 +166,7 @@ function getConfirmation(
 function loadItem(newCatalogItem: BaseModel, fileOrUrl: File | string) {
   if (typeof fileOrUrl === "string") {
     newCatalogItem.setTrait(CommonStrata.definition, "url", fileOrUrl);
+    newCatalogItem.setTrait(CommonStrata.user, "url", fileOrUrl);
```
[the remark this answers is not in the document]
mgcsls: Yeah, at the moment only the user stratum is shared so setting the url at the user stratum seemed like the most straight-forward thing to do (otherwise the url would probably need to be shared separate to the user stratum). At the same time, I didn't want to accidentally mess something else up by not setting url in definition stratum which is why I left that there.

CHANGES.md
```diff
@@ -5,7 +5,7 @@ Change Log
 * Added `hasLocalData` property to indicate when a catalog item contains local data. This property is used to determine whether the item can be shared or not.
 * Fixed bug in StratumOrder where `sortBottomToTop` would sort strata in the wrong order.
 * Allow member re-ordering via GroupMixin's `moveMemberToIndex`
-* Fixed bug causing user added data to not be shared.
+* Fixed bug causing user added data to not be shared. Note that catalog item urls are now set at the user stratum rather than the definition stratum.
```
soyarsauce: Perhaps more accurate that "User added catalog item URLs[...]" (given the only use of `createCatalogItemFromFileOrUrl` is from Add Data)

mgcsls:
Added an extra bit to also fix TerriaJS/nsw-digital-twin#252

lib/Models/CatalogNew.ts
```diff
@@ -17,13 +17,28 @@ export default class Catalog {
   constructor(terria: Terria) {
     this.terria = terria;
     this.group = new CatalogGroup("/", this.terria);
+    this.terria.addModel(this.group);
+
+    // create user added data group
+    this.userAddedDataGroup;
```
tephenavies: We should try to avoid side effects from `get` functions. If we need the group to be created straight away it would be better to create and assign it in the constructor (or a function called by the constructor) and remove `userAddedDataGroupIfItExists`.

lib/Models/CatalogNew.ts
```diff
@@ -17,13 +17,28 @@ export default class Catalog {
...
+    // check for the case where user added data group exists but was
+    // removed from this.group.memberModels after catalog members were reset
+    group = <CatalogGroup | undefined>(
+      this.terria.getModelById(BaseModel, USER_ADDED_CATEGORY_ID)
+    );
+    if (isDefined(group)) {
+      this.group.add(CommonStrata.definition, group);
+      return group;
+    }
+
```
tephenavies: Can you explain what the sequence of events that happens for the user added data group to be in the terria model by ID, but not linked to from `catalog`? Is it when that group is serialised and de-serialised?

If so I think it'd be better to either serialise the user-added data group separately (put it into a different property of the `initData`) or deal with this merging during de-serialisation.

mgcsls: Ah yeah so the problem is that when you use updateModelFromJson on members, e.g. in https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/Terria.ts#L558 all the existing members of the catalog group (including the user added data group) are removed. Probably something which needs to be fixed sometime, but in the meantime I added this to make things at least work. Not too sure what the best thing to do is, would it be better to just go and replace all the relevant occurrences of updateModelFromJson?

PR #3937, opened 2020-01-17
Prevent feature info requests when creating user drawing

### LINKED ISSUE #2949: Mr Mustachio measuring tool waits for unnecessary feature info requests to finish before placing points
When using the Mr Mustachio measuring tool, whenever you click to add a point to the line we are doing feature info requests for all enabled layers, wait for them all to return, and then place the point. This leads to significant lag time in adding points.

As we are not using the feature info when using the measuring tool we should not be sending them in the first place.

mgcsls:
Resolves #2949

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Change Log
 * Expanded test cases to ensure WorkbenchItem & Story have the correct order of components composed
 * Fix broken catalog functions when used with translation HOC
 * Make the default `Legend` width a little smaller to account for the workbench scrollbar
+* Add `allowFeatureInfoRequests` property to `Terria` and prevent unnecessary feature info requests when creating `UserDrawing`s.
 
 ### v7.10.0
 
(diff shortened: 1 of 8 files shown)


soyarsauce:
💯

rowanwins:
Hi @mgcsls

I gave this a whirl but I'm not sure it's working
http://ci.terria.io/user-drawing-feature-info/#share=s-dxtqSphIAGIwmTGYfdYTutAznSI

- Open the network tab in devtools
- Start using the measure tool and notice that requests are sent, in this case to an endpoint called `identify?f=json`

How were you testing the scenario?

mgcsls:
Hi @rowanwins, thanks for checking the pr! I haven't actually stopped feature info requests from being sent since I thought the main issue was the measuring tool having to wait for feature info requests, rather than the feature info requests themselves. But rereading the issue it seems like I should probably stop the requests altogether while the measuring tool is being used, so I'll try to do that.

rowanwins (approved):
Looks good @mgcsls !

PR #3947, opened 2020-01-21
Disable welcome message default in mobx

soyarsauce:


soyarsauce:
This PR changes no behaviour now, given the welcome message is about to get axed via the i18next-PR, until overrides is reimplemented on `terria`

But it now has a commit for prettier so we can add that in for CI to work again.
cc @rowanwins

PR #3951, opened 2020-01-21
Mobx auto-detect esri-mapServer

### LINKED ISSUE #3944: Model Architecture: Auto-detect issue for esri-mapServer
You can't auto-detect this layer when adding web data:
http://services.ga.gov.au/gis/rest/services/NEXIS_Agricultural_Exposure/MapServer

But you can add it if you manually select Esri ArcGIS MapServer

Repro:
1. Visit http://ci.terria.io/mobx/
2. Click add data -> my data -> add web data -> http://services.ga.gov.au/gis/rest/services/NEXIS_Agricultural_Exposure/MapServer with auto-detect as file type
3. Observe "Use conversion service?" error message underneath(separate issue!) the explorer window

mgcsls:
Added some extra catalog member type registrations to `registerCatalogMembers` and modified `createCatalogItemFromUrl` so that it only adds a model to Terria if it has been successfully created and is the correct type. Resolves #3944.


lib/Models/registerCatalogMembers.ts
```diff
@@ -55,4 +56,77 @@ export default function registerCatalogMembers() {
...
+  );
+
+  // These items work by trying to match a URL, then loading the data. If it fails, they move on.
+  createCatalogItemFromUrl.register(
+    matchesUrl(/\/wms/i),
+    WebMapServiceCatalogGroup.type,
+    true
+  );
+  createCatalogItemFromUrl.register(
+    matchesUrl(/\/arcgis\/rest\/.*\/MapServer\/\d+\b/i),
```
soyarsauce, line 104: @steve9164 @rowanwins sorry need your help here - I'm not familiar enough to know if all mapservers have arcgis or rest preceding it? Is it enforced?* or at least some sort of default?

test/Models/createCatalogItemFromUrlSpec.ts
```diff
@@ -0,0 +1,54 @@
...
+    );
+    CatalogMemberFactory.register(GeoJsonCatalogItem.type, GeoJsonCatalogItem);
+
+    createCatalogItemFromUrl.register(
+      s => true,
+      WebMapServiceCatalogGroup.type,
+      true
+    );
+    createCatalogItemFromUrl.register(
+      s => Boolean(s.match(new RegExp("\\.geojson$", "i"))),
```
soyarsauce: Just pull in the `registerCatalogMembers` from `registerCatalogMembers.ts` & use that, OR make `matchesExtension` from `createCatalogItemFromUrl.ts` an export & use that

soyarsauce (changes requested):
Works great!! :100: I had a bit of difficulty at the beginning with the server url provided in https://github.com/TerriaJS/terriajs/issues/3944 being flaky and not returning data. Just need Rowan/Crispy to check the regexes and a small change to the specs.

lib/Models/registerCatalogMembers.ts
rowanwins: I can't think of any Esri/ArcGIS map services that don't have `/MapServer`.

We'd want to auto-detect `/MapServer`(which is a group of layers) and also `/MapServer/0` (which is a specific layer).

soyarsauce:
@mgcsls could you post those summary dotpoints of the parts you wanted clarity on, to fix?

mgcsls:
> @mgcsls could you post those summary dotpoints of the parts you wanted clarity on, to fix?

Yep:

- Currently on mobx master, if a type is a potential match, the item is created using that type and added to terria even if it fails to load later on. But then when we later try to create the item using the correct type, nothing happens because an item with the id (which at the moment is just the url) already exists in terria.
- What I have on this branch right now is essentially: if a type is a potential match, create an item using that type but only add it to terria if it loads successfully. However this feels a bit messy and probably means that a lot of logic ends up being duplicated between upsertModelFromJson and createCatalogItemFromUrl, particularly error checks, which I haven't actually included at the moment.
- Possibly a better approach is to use ReferenceMixin so that the item can be created without knowing its type, but I’m not familiar enough with ReferenceMixin to say for sure.

(Also, a different but related issue is how to make sure an item’s id is unique when another item with the same url already exists.)

tephenavies:
It does sound like a good fit for a ReferenceMixin-based model. The best example of the ReferenceMixin and changing the catalog item type is in MagdaReference.ts.

lib/Models/createCatalogItemFromUrl.ts
```diff
@@ -29,65 +30,39 @@ export function createCatalogItemFromUrlWithOptions(
...
   _index?: number
 ): Promise<BaseModel | undefined> {
-  const index = _index || 0;
-  if (index >= mapping.length) {
-    return Promise.resolve(undefined);
+  let id = options.id || url;
+
+  // make sure id is unique
+  while (terria.getModelById(BaseModel, id) !== undefined) {
+    id += "-";
```
soyarsauce: Increment with a number instead, lest we end up with
someId---------

lib/Models/addUserCatalogMember.ts
```diff
@@ -45,31 +46,36 @@ export default function addUserCatalogMember(
       );
       terria.catalog.userAddedDataGroup.add(CommonStrata.user, newCatalogItem);
 
+      const dereferenced =
```
soyarsauce: Maybe it's time we port over the helper I have in the editor to terriajs#mobx
https://github.com/TerriaJS/terrace/blob/995a33cfc588a7d7d7a4cc3f607982dc02d3d083/lib/Helpers/referenceHelpers.ts#L10

PR #3954, opened 2020-01-21
Set the name of ArcGis MapServer CatalogGroup and CatalogItem on load

### LINKED ISSUE #3953: Name of ArcGis MapServer CatalogGroup and CatalogItem not set on load
Name of `ArcGisMapServerCatalogGroup`  and `ArcGisMapServerCatalogItem` is not Name of the service but the URL, a similar thing was happening with WFS #3926

zoran995:
resolves #3953


rowanwins:
**Before**
Before

**After**
After


Tests all happy locally but failing here because the PR is coming from an external account
happy

rowanwins (approved):

PR #3959, opened 2020-01-22
Mobx: fix shared workbench item order

### LINKED ISSUE #3949: Model Architecture: Workbench order not shared correctly
http://ci.terria.io/mobx/#share=s-nT64YfkeRLnx6tqlEL8byMeNH1v

Repro:
1. visit http://ci.terria.io/mobx/
2. click explore data, and repeat to add datasets "geojson test" and "csv test"
3. click share/print, copy share link
4. load newly generated share link, observe different workbench ordering

mgcsls:
Resolves #3949.


rowanwins:
Currently seeing an error when I sort items in the workbench
`this.props.terria.workbench.moveMemberToIndex is not a function` and when I ran the share it didn't share correctly which I'm guessing is because of the above...

rowanwins:
Tip top work @mgcsls !

PR #3968, opened 2020-01-30
Patch 7.11.0 release 

rowanwins:
Patch 7.11.0 release with minor bug fix colour of marker associated with chart items.

Basically when you click on feature and expand the chart via the feature info it creates a marker on the map which should be the same colour as the line in the chart. I accidentally introduced this bug in #3763 

MarkerColors

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@ Change Log
 
 ### Next Release
 
+### v7.11.1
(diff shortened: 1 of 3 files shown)


soyarsauce (approved):

soyarsauce:
Changelog update too

PR #3969, opened 2020-01-31
Mobx vx charts

na9da:
Rewrite charts using `vx` library.

Test catalog file: https://gist.github.com/na9da/1c50754c4b04fcfd424694f1c9e9106d

Please use terriamap branch - https://github.com/TerriaJS/TerriaMap/pull/416

- [x] Bottom dock chart
- [x] Feature info panel chart
   - [x] Line chart
   - [x] Axis
- [x] Remove references to victory
- [ ] IE test :crossed_fingers: 
- [x] Test on Mobile 
- [x] Profile library size (The total vx dependency size is **981.89 KB**)
- [x] Update `<chart>` component parameters in charts.json and simple.json
- [x] Uncomment call to register custom component in terriamap#mobx 
- [x] Handle chartItems with empty points


rowanwins:
- [ ] resolve issue with overlapping labels
overlappingLables

- [x] remove unused chart selector on workbench items
ChartItemsWorkbench

- [x] add axes labels

- [ ] tooltip only shows values from one dataset (eg if you have points on a line)
Screen

na9da:
About the last point 
> tooltip only shows values from one dataset (eg if you have points on a line)

I actually copied the behavior from [master](https://github.com/TerriaJS/terriajs/blob/master/lib/Charts/Tooltip.js#L113) - 

It says: If there is only a moment dataset at the hover position, then show it's x value only (a date will be shown).

eg:
image

na9da:
@rowanwins 
I couldn't replicate the squished overlapping label problem.

lib/Traits/TimeVaryingTraits.ts
```diff
@@ -1,7 +1,12 @@
-import ModelTraits from "./ModelTraits";
 import primitiveTrait from "./primitiveTrait";
+import mixTraits from "./mixTraits";
+import CatalogMemberTraits from "./CatalogMemberTraits";
+import MappableTraits from "./MappableTraits";
 
-export default class TimeVaryingTraits extends ModelTraits {
+export default class TimeVaryingTraits extends mixTraits(
+  CatalogMemberTraits,
+  MappableTraits
```
tephenavies: I'm not sure something with TimeVaryingTraits will always be Mappable

tephenavies (changes requested):
Please also add a changelog entry in CHANGES.md. That entry should describe the change from master to what's introduced here.

lib/Traits/TimeVaryingTraits.ts
na9da: This is because `DiscretelyTimeVaryingMixin` depends on the `.show` trait to decide [whether to show/hide](https://github.com/TerriaJS/terriajs/blob/mobx-vx-charts/lib/ModelMixins/DiscretelyTimeVaryingMixin.ts#L280) a chart item in chart panel (similar to the behavior in master). We can probably move the show trait to `CatalogMemeberTraits` which is more in line with its actual usage?!

tephenavies: Ahh ok. I don't think it's a CatalogMemberTrait, since not all members are "show"able (e.g. a group). It might have to be it's own Trait class `Showable` that's mixed into DiscretelyTimeVaryingTraits & MappableTraits.

lib/ReactViews/Custom/Chart/FeatureInfoPanelChart.jsx
```diff
@@ -38,9 +38,14 @@ class FeatureInfoPanelChart extends React.Component {
     );
   }
 
+  componentDidMount() {
+    runInAction(() => {
```
tephenavies, line 48: This should also run in componentDidUpdate if it’s possible that props.item could change without the component remounting (maybe when there’s multiple feature info results from different items each with a graph? It’s probably still remounting though)

PR #3971, opened 2020-01-31
minNativeZoom

geoffrowland:
Pass minimumLevel, in Cesium, to minNativeZoom, in Leaflet.
As already happens for maximumLevel to maxNativeZoom.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@ Change Log
 
 ### Next Release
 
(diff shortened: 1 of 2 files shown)


soyarsauce (approved):
I can't see why we wouldn't - is there any reason not to? @steve9164 @rowanwins

rowanwins:
seems reasonable - go for it @soyarsauce

PR #3978, opened 2020-02-04
Mobx: Fix the story panel location button

rowanwins:
Roughly from ticket
> Create a story. Play the story, then pan away, then click the location button. The map should recenter to the scene location, but nothing happens. There is a console error which hints that the feature is not fully ported.

Resolves https://github.com/TerriaJS/nsw-digital-twin/issues/247

btn


na9da:
Looks good!

Issue #237, opened 2014-07-18
Add support for csv files with region data


bsimyo:
There's been a lot of interest from govt agencies in visualising data from spreadsheets which have data by region (eg per postcode, lga, state, etc) along the lines of our "data-by-region" govhack project.  This form of csv files should be able to be in a data collection, dragged in, added etc.

PR #3979, opened 2020-02-04
Mobx: restrict location search results to a bounding box

rowanwins:
This PR adds a bounding box restriction to the location search (handy for providers which don't support anything similar via API requests).

In the config.json for a TerriaMap add a `locationSearchBoundingBox` to your params in the form a bounding box In minx, miny, maxx, maxy
````
    parameters: {
        "locationSearchBoundingBox": [109, -45, 158, -8]   <---- roughly australia
    }
````
Defaults to undefined and no results are filtered
**Without Filter**
WithoutFilter

**With Filter**
screenshot-localhost_3001-2020 02 05-09_18_47

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Change Log
 * Fixed bug causing workbench items to be shared in the wrong order.
 * Fix bug where urls in the feature info panel weren't turned into hyperlinks
 * Fix preview map's base map and bounding rectangle size
+* Added the ability to filter location search results by an app-wide bounding box configuration parameter
 
 ### Next Release
 * Fix draggable workbench/story items with translation HOC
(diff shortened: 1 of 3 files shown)


rowanwins:
The only other thought here is that some providers will support querying by a bounding box and we won't need to filter on the client so perhaps we should add some flag to the `SearchProvider` class. Suggest we tackle down the track when we actually have more than 1 search provider in mobx that would actually require this.

tephenavies:
Can you add an entry to the changelog?

tephenavies:
Actually don't worry. I'll add one.

tephenavies (approved):

KeyboardSounds:
Fixed: #237

KeyboardSounds:
Fixes TerriaJS/nsw-digital-twin#237

PR #3980, opened 2020-02-04
Mobx: ArcGisFeatureServer catalog item and group

mgcsls:


lib/Models/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -0,0 +1,621 @@
...
+        if (color) {
+          const item = {
+            title: label,
+            color: convertEsriColorToCesiumColor(color).toCssColorString()
+          };
+          items.push(item);
+        }
+      }
+      const legend = <StratumFromTraits<LegendTraits>>(
+        (<unknown>{ items: items })
```
rowanwins: Could possibly add the legend title here from `drawingInfo.renderer.field` - I think this is how Esri does it

rowanwins:
One thing I've noticed is that if I do something
````
    {
      "type": "esri-featureServer",
      "name": "Esri FS",
      "url": "https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/World_Cup_Qualifiers_2014/FeatureServer"
    },
````
where there are multiple layers Terria only draws the first layer. 
It's possible that the above is a misconfiguration (it should be added with `"type": "esri-featureServer-group"`, but perhaps we should through an error?

File: lib/Models/ArcGisFeatureServerCatalogItem.ts [hunk 2]
```diff
@@ -0,0 +1,621 @@
...
+
+function buildMetadataUrl(catalogItem: ArcGisFeatureServerCatalogItem) {
+  return proxyCatalogItemUrl(
+    catalogItem,
+    new URI(catalogItem.url).addQuery("f", "json").toString()
+  );
+}
+
+function buildGeoJsonUrl(catalogItem: ArcGisFeatureServerCatalogItem) {
+  var url = cleanAndProxyUrl(catalogItem, catalogItem.url || "TODO");
```
rowanwins: The TODO can be `"0d"`

lib/Models/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -0,0 +1,621 @@
...
+    if (
+      !this._item.useStyleInformationFromService ||
+      !this._featureServer.drawingInfo
+    ) {
+      return undefined;
+    }
+    const renderer = this._featureServer.drawingInfo.renderer;
+    const rendererType = renderer.type;
+
+    if (rendererType === "classBreaks") {
```
rowanwins: Probably need to work out how to render legends for other renderer types like `uniqueValue`. This service offers a good example
https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/World_Regions/FeatureServer/0 , and here's an associated [Esri demo map](https://www.arcgis.com/home/webmap/viewer.html?url=https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/World_Regions/FeatureServer/0&source=sd)

lib/Models/ArcGisFeatureServerCatalogItem.ts
mgcsls: I agree that a legend title would be good... not too sure how to actually add one though so I might leave this for another time.

rowanwins:
One minor quibble - it would be good to grab the name from the feature server - it seems to be squirrelled away in the featureServer strata at the moment. Then I think we're good

Name

rowanwins (approved):
Great work @mgcsls - this is excellent!

PR #3983, opened 2020-02-05
Make gtfs hideable

reginapramesti:
Fixes https://github.com/TerriaJS/nsw-digital-twin/issues/255


soyarsauce:
Remember changelog too 👍

lib/Models/GtfsCatalogItem.ts
```diff
@@ -258,7 +259,11 @@ export default class GtfsCatalogItem extends AsyncMappableMixin(
 
   @computed
   get mapItems(): DataSource[] {
-    return [this.dataSource];
+    if (isDefined(this.dataSource)) {
```
na9da: Do we need the check here since the type of `get dataSource` guarantees that it is always a `DataSource`?

lib/Models/GtfsCatalogItem.ts
```diff
@@ -258,7 +259,11 @@ export default class GtfsCatalogItem extends AsyncMappableMixin(
...
+      this._dataSource.show = this.show;
```
na9da: The best place to set `_dataSource.show` might actually be inside the `get dataSource`. That's just my opinion :)

lib/Models/GtfsCatalogItem.ts
reginapramesti: Yeah I guess not, was just basing it off the other catalog items whoops!

lib/Models/GtfsCatalogItem.ts
reginapramesti: That makes sense - will fix that

reginapramesti: Though I'm finding that when I set `_dataSource.show` in `get dataSource` it doesn't behave in the same way. Instead of hiding the images when the button is clicked, it only does that when the timer resets. Not sure why - do you have any thoughts? I might keep it in `mapItems` for now.

na9da: You are right, it seems to show/hide only when setting `_dataSource.show` inside of `mapItems`. Setting `_dataSources.show` inside `get dataSource` is not triggering the autorun `Cesium.ts` for some reason :thinking: 

I'll approve this for now.

na9da (approved):

PR #3986, opened 2020-02-06
Mobx: Support zoom to on gtfs catalog item

rowanwins:
This adds a populated `rectangle` property to the `GtfsCatalogItem`


rowanwins:
resolves https://github.com/TerriaJS/nsw-digital-twin/issues/256

na9da (approved):
Looks great!

In the future, I think, we should refactor and move the fetching/parsing code to the Loadable stratum.

PR #3987, opened 2020-02-06
Mobx: Add a legend title trait

rowanwins:
Makes it possible to put useful titles on top of legends.

Note that there is already a `titleBelow` and `titleAbove` traits on the `LegendItemTraits` although I think this fills a different purpose of providing a single overarching title for a legend.

screenshot-localhost_3001-2020 02 06-13_43_51


soyarsauce:
Remember changelog too 👍

PR #3988, opened 2020-02-06
Fix mobile ui bottom part

reginapramesti:
Fixes https://github.com/TerriaJS/nsw-digital-twin/issues/265

~~Still in the process of testing it out on mobile~~

EDIT: Mostly works now, but still has issues in firefox with having to scroll down first to hide the address bar in order to see the buttons at the bottom. This is apparently the case with various other websites too - any div with fixed position doesn't show until you hide the address bar by scroll down a bit.


soyarsauce:
CI failure is due to prettier

soyarsauce:
Good start! FYI the bottom buttons still get covered by the timeline when it's visible

soyarsauce:
Remember changelog too 👍

reginapramesti:
Works for Chrome, Safari (based on Saucelabs), and normal Firefox. A minor issue described above on Firefox Focus but that might just be an issue with the browser itself.

soyarsauce:
Amazing work on tidying this up @reginapramesti except this assumes the timeline only spans 1 line, which is why things weren't positioned in `fixed` 

image

I'm wondering if we truncate the text, or revert things back to relative. Any thoughts or ideas? I'm leaning towards less-fixed-positioning, on top of truncating to max two lines. That way we have flexibility.

Failing that we simply keep it as is now, truncate, and revisit post launch.

soyarsauce:
(Previous / mobx-master is currently 'responsive' to height of timeline label)
image

soyarsauce:
I'm actually leaning towards my last suggestion,
>Failing that we simply keep it as is now, truncate, and revisit post launch.

soyarsauce:
Hey @reginapramesti I know I said I'll pick this up but I'll give it back to you, can you just truncate the name (only for mobile/small-screen-ui) so that it never wraps onto two lines? Then we can get this merged, and then we'll raise another (low prio issue) for a better idea of how to show it all. Thanks 🙏🙏

reginapramesti:
Text truncated! Hopefully doesn't majorly affect the layout of other buttons/divs. From what I can tell based on testing and the below comparisons it doesn't affect much. Note this layer has the longest name I can find in the catalog 😆 

On desktop - before (top) and after (bottom)
<p align="center">


</p>

On mobile (Samsung Galaxy S9 resolution as it has the smallest width of viewport)- before (left) and after (right)
<p align="center">


</p>

On tablet (iPad) - before (left) and after (right)
<p align="center">


</p>

soyarsauce:
🎉

PR #3992, opened 2020-02-07
Mobx: Prevent app crash when timeline active and switching to 2D

### LINKED ISSUE #3989: App crashes when switching to 2d map with timeline on
* From the map menu select "Always show timeline"
* Then switch to 2d map

App crashes with a white screen.

rowanwins:
I'm not super sure of this fix but it works... 

Previously
- From the map menu select "Always show timeline"
- Then switch to 2d map

Resolves #3989

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Change Log
 * Fix bug where urls in the feature info panel weren't turned into hyperlinks
 * Fix preview map's base map and bounding rectangle size
 * Re-introduce UI elements for search when a catalogSearchProvider is provided
+* Prevent TerriaMap from crashing when timeline is on and changing to 2D
 
 ### Next Release
 * Fix draggable workbench/story items with translation HOC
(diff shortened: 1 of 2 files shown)


soyarsauce:
Remember changelog too 👍

na9da:
I'm not sure either :) Maybe @steve9164 can help.

As far as I understood, the `computed` getter `TerriaViewer.leaflet` sets off a chain of calls (including a cesium raiseEvent) which finally errors out by calling an action that tries to mutate things. This wouldn't have happened if the `raiseEvent` was actually asynchronous!

One fix I could think of is breaking that chain of calls by wrapping this [ticker call](https://github.com/TerriaJS/terriajs/blob/ff7ca079beac6b1f70e3dcee90a9b815a665f5df/lib/Models/Leaflet.ts#L166) in a `window.setTimeout(() => ticker(), 0)` to force it to be asynchronous.

tephenavies:
Thanks @rowanwins. I think I prefer @na9da's solution. I've made the PR #3998 with that alternate fix.

PR #3997, opened 2020-02-10
Set default value for `showsInfo`.

na9da:
We set a default `true` value for `showsInfo` by defining a getter in the `CatalogMemberMixin`.
This can be overridden if required, in the child class, as in ResultPendingCatalogItem.


soyarsauce (approved):

soyarsauce (changes requested):
CI Failures

na9da:
> CI Failures

Fixed.

PR #4000, opened 2020-02-10
Mobx: Support zoom to on CsvCatalogItem

rowanwins:
Works for CSV's with an identified lat lon column in the default style
screenshot-localhost_3001-2020 02 11-08_38_46

No button displayed when 
- region mapping
- CSV has non-geographic data


lib/Models/CsvCatalogItem.ts
```diff
@@ -65,6 +65,18 @@ export default class CsvCatalogItem extends TableMixin(
     return isDefined(this._csvFile);
   }
 
+  // Zooming to tables with lat/lon columns works
+  get canZoomTo() {
```
tephenavies, line 69: This should probably be a computed. That will avoid any unnecessary React redraws of the workbench UI/other parts of the UI that use `canZoomTo`.

PR #4003, opened 2020-02-11
Mobx: fix catalog userAddedDataGroup bug

mgcsls:
Added some checks to make sure the catalog always has a userAddedDataGroup by adding the userAddedDataGroup back whenever it is accidentally removed (e.g. when setting terria.catalog.group or terria.catalog.group.members).


tephenavies (commented):
This looks good. Please change the name of the dispose function to something like I've suggested (in the constructor too). Also I think `runInAction` import is not used and there should be a `destroy` method that calls that disposer (although we don't have any place that needs to call the destroy function at the moment, except maybe in tests where new Terria objects are created frequently).

PR #4005, opened 2020-02-11
remove duplicate text from descriptions

### LINKED ISSUE #4004: Mobx: Duplicate "This is a" in description
screenshot-nsw dt terria io-2020 02 11-13_44_09

rowanwins:
Removes duplicate text that was probably introduced by internationalisation

resolves #4004

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -24,6 +24,7 @@ Change Log
 * Fix bug that prevented live transport data from being hidden
 * Hide opacity control for point-table catalog items.
 * Show About Data for all items by default.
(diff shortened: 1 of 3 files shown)


rowanwins:
After

tephenavies:
Closing in favour of #4011. The problem is the `<0>...</0>` tag should be `<1>...</1>`. Unfortunately there are a few small errors in the translation strings used by the `<Trans>` component. See https://react.i18next.com/latest/trans-component#how-to-get-the-correct-translation-string for how the numbering system works.

PR #4013, opened 2020-02-11
Fix initialisation of terria's shareDataService

mgcsls:
Should fix problem where ci mobx isn't shortening share urls. I also very slightly rewrote the relevant part of terria.ts to avoid confusion with `serverConfig`, which I originally thought was `this.serverConfig` but is actually `this.serverConfig.config`.


na9da:
Looks good to me!

PR #4014, opened 2020-02-12
Mobx: Add trait to hide shadows UI on workbench items for Cesium3dTilesCatalogItem

rowanwins:
This PR hides the shadows UI from the workbench item when it's a Cesium3DTiles dataset that has points

I don't know the 3dTiles style spec at all so there may be a more robust way to achieve the same functionality - asking for a friend in @KeyboardSounds hopefully :)

Example catalog item that shouldn't have shadows
````
{
      "name": "Education3D",
      "type": "3d-tiles",
      "url": "https://nsw.dt.terria.io/i3s-to-3dtiles/https://portal.spatial.nsw.gov.au/i3s/rest/services/Hosted/ASP/Education3D/SceneServer"
    }
````

NoShadows


rowanwins:
partially resolves https://github.com/TerriaJS/nsw-digital-twin/issues/258

KeyboardSounds:
Oh boy. Determining whether a 3d tileset has point data in it is a seemingly trivial thing that is actually really, really complicated.

The **TLDR** version is that you can't reliably do it. Rather than automatically guessing whether we should show the shadow UI or not, let's just make it a new trait on the catalog item-- `showShadowUI` or `allowShadowConfig` or something.

Here's the long version:
3D Tiles Tilesets are made up of many, many tiles. Each tile can contain entirely different data to all of the other tiles in a dataset. One tile could be a bunch of points, the one next to it could be a 3D model of my head. There's no real concept of "points" and "3D volume mesh" at the tileset level. A tileset has no idea what kind of data it contains.

The `tileset.extras.style.pointSize` thing you're doing here is a clever workaround. `extras` is for application specific properties, and there's a Cesium specific property that we add to tilesets that we generate called `style.pointSize` when that tileset contains points. It also gets added when you specify styles for a tileset in the catalog using the 3D tiles styling language. Buuuut someone could hand us a tileset made up entirely of point data that doesn't have that property in it at all. It's not part of the 3D Tiles spec.

Because of all this, I think we should just give up, and make it configurable in the catalog.

KeyboardSounds (changes requested):
As per my comment, just give up doing the clever, user friendly thing 😢 , and make a trait for this so it can be set in the catalog.

rowanwins:
Refactored based on the advice above. Now controlled by catalog config
````
{
      "name": "Education3D",
      "type": "3d-tiles",
      "url": "https://nsw.dt.terria.io/i3s-to-3dtiles/https://portal.spatial.nsw.gov.au/i3s/rest/services/Hosted/ASP/Education3D/SceneServer",
      "showShadowUi": false
}

KeyboardSounds (approved):

PR #4015, opened 2020-02-12
Mobx: add only loaded items to workbench

mgcsls:
Modified `addToWorkbench` so that items that fail to load are removed from the workbench. Also added an error message which is displayed when the item doesn't load. (Note however that this error message currently appears underneath the catalog, but that'll probably be fixed later on.)

Resolves TerriaJS/nsw-digital-twin#274.


rowanwins (approved):
Looks good to me!

PR #4018, opened 2020-02-13
Mobx: support for FeatureInfo on region mapped datasets

rowanwins:
Add support for getting Feature Info results from region mapped datasets

Just missing the feature highlighting...


rowanwins:
Dont merge just yet - might be an adjustment coming now that I've tested another CSV

rowanwins:
So at the moment there is basically no support for time-varying csv's 
eg selecting the year in the workbench dropdown.

At the moment this PR will just grab the first matching year for the region but this needs to be revisited.

PR #4019, opened 2020-02-13
Mobx: fix feature server proxy url and token bugs

mgcsls:
Fixed bug causing `ArcGisFeatureServerCatalogItem` to throw an error when a token is included after its url is proxied.

### DIFF
--- a/lib/Models/proxyCatalogItemUrl.js
+++ b/lib/Models/proxyCatalogItemUrl.js
@@ -31,15 +31,10 @@ var proxyCatalogItemUrl = function(catalogItem, url, cacheDuration) {
     return url;
   }
 
-  const urlOrResource = corsProxy.getURL(
+  return corsProxy.getURL(
     url,
     defaultValue(catalogItem.cacheDuration, cacheDuration)
   );
-  if (urlOrResource instanceof Resource) {
-    return urlOrResource.url;
-  } else {
-    return urlOrResource;
-  }
 };
 
 module.exports = proxyCatalogItemUrl;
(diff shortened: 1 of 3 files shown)


lib/Models/proxyCatalogItemUrl.js
```diff
@@ -30,10 +31,15 @@ var proxyCatalogItemUrl = function(catalogItem, url, cacheDuration) {
     return url;
   }
 
-  return corsProxy.getURL(
+  const urlOrResource = corsProxy.getURL(
     url,
     defaultValue(catalogItem.cacheDuration, cacheDuration)
   );
+  if (urlOrResource instanceof Resource) {
+    return urlOrResource.url;
```
kring: This is going to break token refreshing in the NSWDT.

mgcsls: Not entirely sure I understand what that means or how that works but ok haha I'll change `proxyCatalogItemUrl` back. But there do seem to be places where things seem to be expecting `proxyCatalogItemUrl` to return a string (at least from what I can tell, very possible I'm wrong), e.g. https://github.com/TerriaJS/terriajs/blob/mobx/lib/ModelMixins/FeatureInfoMixin.ts#L42 and https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/CsvCatalogItem.ts#L109, not to mention the (outdated?) javadoc here https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/proxyCatalogItemUrl.js#L12 definitely claims that `proxyCatalogItemUrl` returns a string.

mgcsls: I just realised this wasn't even a necessary change lmao yeah so definitely changing this back to how it was.

PR #4020, opened 2020-02-13
Mobx: fix sharing of added data with auto detected data type

mgcsls:
Fixes bug causing user added data whose type is auto-detected to not be shared properly.


tephenavies:
Hey @mgcsls. Could you please add more of a description about what the original bug is and how this fixes it?

mgcsls:
@steve9164 go to ci.terria.io/mobx/, add any data (e.g. http://ereeftds.bom.gov.au/ereefs/tds/wms/ereefs/mwq_gridAgg_P1M), auto-detect the data type. If you try sharing the data (from anywhere) and then visit the share link, you'll get an error message and the data won't be shared:

image

Basically this pr should fix this by making sure UrlReferences are actually dereferenced properly when init data is being applied.

na9da:
Looks good!

PR #4021, opened 2020-02-13
Mobx: set traits of newly created item at defaults stratum

mgcsls:
Sets name of newly created catalog members to be the url in the default stratum rather than the definition stratum so that the "url" name does not override actual names from a load stratum.


soyarsauce:
I dislike this change, only from my biased-via-terrace-editor view (and that it will severely break the adding-items-feature of the editor without some re-thinking), because you are technically defining(ala definition) a URL here - what problem is it trying to fix?

soyarsauce:
A better approach is to explicitly set the name on default, and remove it from definition - given we set localId already, so no name is required.

mgcsls:
> I dislike this change, only from my biased-via-terrace-editor view (and that it will severely break the adding-items-feature of the editor without some re-thinking), because you are technically defining(ala definition) a URL here - what problem is it trying to fix?

The problem I'm trying to fix is this:

image

You can see how in the second catalog item, the displayed name is the url, rather than the name from the ArcGIS feature server, which is displayed in the first item. This will happen whenever you add an ArcGisFeatureServerCatalogItem (without auto-detecting type), or any other item which has a name at the load stratum. The reason the proper name doesn't appear is because it gets overridden by the name set in the definition stratum, which is just the url.

Actually, I only needed to change the line in AddData.jsx to fix the above problem, but thought it would probably make sense to make things consistent hence also changing the line in createCatalogItemFromUrl.ts. But yeah if it'll mess up editor stuff I'll change that bit to what you suggested.

mgcsls:
> A better approach is to explicitly set the name on default, and remove it from definition - given we set localId already, so no name is required.

Actually I think it isn't even necessary to set the name at all in `createCatalogItemFromUrlWithOptions`? Since I think the name of a `UrlReference` is never used, and the name of the target is set to be the url (at a default stratum) here: https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/UrlReference.ts#L99.

soyarsauce:
I just tried this, first auto detect, second specified the type:

soyarsauce:
I just tried this, first auto detect, second specified the type:
image

soyarsauce:
(on http://ci.terria.io/mobx-new-item-default-stratum/)

mgcsls:
> I just tried this, first auto detect, second specified the type:
> image

Yeah that's because that feature server doesn't actually have a name (there's no documentInfo property in https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/World_Cup_Qualifiers_2014/FeatureServer?f=pjson). Try this one instead: https://sampleserver6.arcgisonline.com/arcgis/rest/services/EmergencyFacilities/FeatureServer)

mgcsls:
(the one I used in the earlier comment was https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/World_Cup_Qualifiers_2014/FeatureServer/1, not https://services.arcgis.com/P3ePLMYs2RVChkJx/ArcGIS/rest/services/World_Cup_Qualifiers_2014/FeatureServer)

soyarsauce:
Aha that makes sense. Works as expected now, however I'll leave the direction of this to @steve9164 - I'm still not sold on setting the type on default (for serialising purposes down the track), I believe it makes most sense as
definition: type
default: name

mgcsls:
> Aha that makes sense. Works as expected now, however I'll leave the direction of this to @steve9164 - I'm still not sold on setting the type on default (for serialising purposes down the track), I believe it makes most sense as
> definition: type
> default: name

Oh type isn't actually a trait (it's just used for creating the catalog item), so it won't matter which stratum type is set in, at least with how things are now. But yes I'll wait to see what Crispy says.

soyarsauce:
Wow how did I forget that. You're right. In which case this is all probably good to go

tephenavies (approved):

PR #4027, opened 2020-02-16
Mobx ui tweaks

soyarsauce:
Small tweaks to map buttons to match styling, bottom half hasn't been implemented due to mobile layout solution not being clear
(Also fixes https://github.com/TerriaJS/terriajs/issues/4026 so you can dev quicker in 2d)


soyarsauce:
Ready(?!?)

KeyboardSounds:
Related maps looks like this:

image


But otherwise it's very beautiful 😍

KeyboardSounds (changes requested):
Fix the Related Maps and you're good to go!

soyarsauce:
Really annoying... on DT map I let that slide because the login button is to the left of related maps and gave it enough space. I'll see what quick fix we can do

rowanwins:
Also slight issue with button wrapping with map width is less than 575px

screenshot-ci terria io-2020 02 18-15_20_37

soyarsauce:
re; button wrapping:
>(ps vic i need an idea for this:) when the UI is small enough that the buttons run out of room, but not small enough to trigger mobile UI.

>If we are running out of time on the project I dont this this is a huge priority. I mean i am all about making sure this doesnt happen, but in the bigger picture, workbench is still way more important. Especially considering that most people will have their browser quite large.

I will try and whip something up. .

soyarsauce:
Wrapping tweaked like this:
image

soyarsauce:
Wrapping tweaked like this, still happens when you have story and really small width, but it's an improvement.
image

soyarsauce:
& `MenuPanel`s have an option to pass `showDropdownInCenter` as a prop, to do as it describes - open from center instead of offset to the right.
image

PR #4028, opened 2020-02-17
Mobx: Fix highlighting of cesium 3D tile features

mgcsls:
Change colour of feature highlight (of cesium 3D tile features) so that it is more visible on lighter colours. The new colour looks like:
image
image
This new highlight colour won't really appear on dark/black objects, but neither does the old highlight colour. In fact, from looking at how cesium feature highlighting colours work, I'm actually not sure if there is a easy way of highlighting dark objects at all.

Aside from changing feature highlight colour, also fixed a bug causing cesium 3D tile feature highlights to stop working after a cesium3DTilesCatalogItem is removed from the workbench while its feature info is displayed.

Resolves TerriaJS/nsw-digital-twin#262.


kring:
I recently learned that there's a fancy, not-too-well-documented feature to fix this problem.

```
tileset.colorBlendMode = Cesium.Cesium3DTileColorBlendMode.REPLACE;
```

With that, the style color will replace the material color instead of being blended with it by multiplying. Apparently there's a bug where it doesn't work when the color is full white (`#ffffff`) so use `#fffffe` instead if you want white.

mgcsls:
> I recently learned that there's a fancy, not-too-well-documented feature to fix this problem.
> 
> ```
> tileset.colorBlendMode = Cesium.Cesium3DTileColorBlendMode.REPLACE;
> ```
> 
> With that, the style color will replace the material color instead of being blended with it by multiplying. Apparently there's a bug where it doesn't work when the color is full white (`#ffffff`) so use `#fffffe` instead if you want white.

Ooh interesting, I'll have a look into that. Thanks :)

soyarsauce:
Only thing to check when going to cesium's 1.66 type definitions is double check the deprecated `createOpenStreetMapImageryProvider` and `createTileMapServiceImageryProvider` items are updated to match in `lib/ThirdParty/terriajs-cesium/index.d.ts` - which it is, hope that's all we need to do for that 🎉

na9da:
Looks great!

PR #4036, opened 2020-02-19
Add terria branding

reginapramesti:
Fixes https://github.com/TerriaJS/nsw-digital-twin/issues/259

Things to note:
* The dimensions of the black bar at the bottom doesn't exactly follow the invision designs mainly because the terria logo becomes tiny, easy to fix if need to change
* Hovering over the lat/long/elev and the scale bar still changes the background colours to the original settings, not sure what to do here


reginapramesti:
I have no idea why prettier is giving that section of the code an error 😢

reginapramesti:
@soyarsauce I can't find a way to revert back to the previous formatting, but I might just be inexperienced and don't know where to look it up exactly :( the format changes came with prettier version 1.19.0, and sadly the versions prior to that would throw an error like it did previously so I might stick with the current updated version for now

PR #4037, opened 2020-02-19
Mobx: add pointSizeMap to TableStyle

mgcsls:
Allow point sizes to be scaled by value, e.g.

image


soyarsauce:
🚀

na9da:
@mgcsls This looks great!

Can you make the default sizes match the one in your screenshot? The points generated using the default values for both ConstantPointSizeMap and ScalePointSizeMap look small to me:

image

PR #4038, opened 2020-02-19
Mobx: Rejig magda feature server handling

rowanwins:
I'm trying to improve how some of the funny esri feature server handling occurs when passed from magda & an esri portal.

Basically i'm attempting to override a service reference like
https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Features_of_Interest_Category/FeatureServer
with
https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Features_of_Interest_Category/FeatureServer/6
because the `distribution.aspects["dcat-distribution-strings"].accessURL` isn't accurate

TS is giving me grief in trying to use async/await and saying something about needing to return a promise, [similar to this](https://github.com/Microsoft/TypeScript/issues/15292), but I can't seem to make it happy.

The PR needs some tidying still but I need some help just to get ts to compile so I can continue :)

PR #4042, opened 2020-02-20
DONT MERGE - Fix handling of feature servers provided by magda

rowanwins:
I'm trying to improve how some of the funny esri feature server handling occurs when passed from magda & an esri portal.

Basically i'm attempting to override a service reference like
`https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Features_of_Interest_Category/FeatureServer`
with
`https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Features_of_Interest_Category/FeatureServer/6`
because the `distribution.aspects["dcat-distribution-strings"].accessURL` isn't accurate

supercedes #4038


rowanwins:
Test with

This will add a single layer
````
{
      "type": "magda",
      "url": "https://nsw.dt.terria.io",
      "name": "Electricity Lines",
      "isMappable": true,
      "recordId": "ds-nsw-portal-deee22e925fd4b9b9e0de1ffd5c8fa62"
  }
````

This one will show a group of layers - the `delivsdm geodb MinesSubsidenceDistrict` layer actually works but most of the others dont for reasons unrelated to this PR
````
{
      "type": "magda",
      "url": "https://nsw.dt.terria.io",
      "name": "Admin boundaries group",
      "isMappable": true,
      "recordId": "ds-nsw-portal-2a74b164882940f9aa40fbed175d446b"
  }
````

rowanwins:
The guts of my work is these [lines](https://github.com/TerriaJS/terriajs/pull/4042/files#diff-4f2931488da152d599604d7674336882R681-R710)

tephenavies:
Why is this a problem for `MagdaReference`s and not for normal Esri layers?

rowanwins:
`MagdaReference` takes a magda record and decides whether it should create `esri-featureServer` or an `esri-featureServer-group`. Ideally we'd have a perfect understanding by the time it reached Terria (eg our portal connector would do a better job) but alas.

rowanwins:
So I'm thinking we perhaps shove this off to the terria minion.

PR #4053, opened 2020-02-21
Mobx: temporary fix for errors while loading init data models

mgcsls:
A temporary fix for TerriaJS/nsw-digital-twin#249.

### DIFF
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -680,7 +680,10 @@ export default class Terria {
             stratumId,
             models,
             replaceStratum
-          );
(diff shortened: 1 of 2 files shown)


lib/Models/Terria.ts
```diff
@@ -680,7 +680,10 @@ export default class Terria {
             stratumId,
             models,
             replaceStratum
-          );
+          ).catch(() => {
+            // TODO: deal with shared models that can't be loaded because, e.g. because they are private
```
soyarsauce, line 684: let's at least show the error in console

PR #4062, opened 2020-02-22
Stop 499 error throwing

tephenavies:


soyarsauce:
would be good if you can link a share link to reproduce, to test that it's all good

PR #4081, opened 2020-02-25
Help ui cleanup

reginapramesti:
Just a bunch of tidying up - fix minor errors, remove console logs, reimplement things that were a bit hacky


soyarsauce:
Restarted CI job

soyarsauce:
Restarted CI job

lib/Styled/Text.jsx
```diff
@@ -48,7 +48,7 @@ export const Text = styled.div`
     font-size: 15px;
   `}
   ${props =>
-    props.title &&
+    props.heading &&
```
soyarsauce, line 51: @reginapramesti have a look at what I've done in
https://github.com/TerriaJS/terrace/blob/master/lib/Styled/Heading.tsx#L20

Maybe it's time we brought that across, it's basically a wrapper around `Text` but with the proper html element for whichever heading we want to use, I'm not sure what the final solution/mishmash is going to look like, let me know what you think 🙂

soyarsauce (approved):

lib/Styled/Text.jsx
reginapramesti: @soyarsauce oooh yeah that would be really good to have I reckon!

PR #4084, opened 2020-02-25
Fix story analytics

tephenavies:
Refactored datasource event logging function out and instead created a general `getPath` used for sending with event data and in the catalogue and workbench. With tests!

Also fixed the order that `getAncestors` returns ancestors in.


soyarsauce:
Restarted CI job

na9da:
Merged in mobx branch to fix the CI heap issue.

lib/ReactViews/Workbench/WorkbenchItem.jsx
```diff
@@ -2,14 +2,18 @@
...
 import React from "react";
 import { sortable } from "react-anything-sortable";
+import { withTranslation } from "react-i18next";
 import defined from "terriajs-cesium/Source/Core/defined";
+import getPath from "../../Core/getPath";
 import CommonStrata from "../../Models/CommonStrata";
-import getAncestors from "../../Models/getAncestors";
+import Box from "../../Styled/Box";
 import Icon from "../Icon";
+import PrivateIndicator from "../PrivateIndicator/PrivateIndicator";
```
soyarsauce, line 16: How is this sorted? # of traversals?

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
```diff
@@ -76,12 +75,12 @@ export const DataCatalogItem = observer(
             this.props.terria.workbench.contains(this.props.item) &&
             !keepCatalogOpen
           ) {
-            logDatasetAnalyticsEvent(
-              this.props.terria,
-              this.props.item,
-              toAdd ? "addFromCatalogue" : "removeFromCatalogue"
-            );
             this.props.viewState.closeCatalog();
+            this.props.terria.analytics?.logEvent(
```
soyarsauce, line 79: Come to think of it, isn't analytics **always** available on terria?

soyarsauce (changes requested):
Really awesome cleaning up, @steve9164! 🚀 
Should we add type definitions for `analytics` on `Terria`, or you want to leave it with the optional chaining for now?

lib/ReactViews/Workbench/WorkbenchItem.jsx
tephenavies: It looks like alphabetically with less local modules first. It's vscode's default when you run `Organise imports`. It's a de facto standard that Kevin and I have used for a couple of years. Just a little neater than having everything everywhere.

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
soyarsauce: (.. except I now realise terria isn't marked as a required prop in its prop types)

lib/ReactViews/Workbench/WorkbenchItem.jsx
tephenavies: It might make merges a bit messy, but in a 3 way merge you can accept both current changes and incoming changes, then run `Organise imports` again and it'll remove all duplicate and unused imports and order them again.

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
soyarsauce: Fix this here:https://github.com/TerriaJS/terriajs/issues/4122

PR #4088, opened 2020-02-26
Mobx fix ar

soyarsauce:


soyarsauce:
Restarted CI job

lib/ReactViews/Map/Navigation/AugmentedVirtualityTool.jsx
```diff
@@ -36,6 +36,33 @@ class AugmentedVirtualityTool extends React.Component {
   handleClickAVTool() {
     // Make the AugmentedVirtuality module avaliable elsewhere.
     this.props.terria.augmentedVirtuality = this.augmentedVirtuality;
+    // feature detect for new ios 13
+    // it seems you don't need to ask for both, but who knows, ios 14 / something
+    // could change again
+    if (
+      DeviceMotionEvent &&
```
na9da: This should probably be `window.DeviceMotionEvent`  My chrome version doesn't have DeviceMotionEvent and it errors out saying DeviceMotionEvent is undefined.

Same with `DeviceOrientationEvent` below.

na9da:
- [x] Works in Firefox
- [x] Works in Chrome for Android
- [ ] Safari
  I couldn't see the AR buttons on Safari (iOS version 13.3.1). Same with Chrome on iOS (80.0.3) but works well in Firefox iOS.

image


There is also a problem with the alignment of buttons in Chrome which is not new. I have logged it here: https://github.com/TerriaJS/terriajs/issues/4090

lib/ReactViews/Map/Navigation/AugmentedVirtualityTool.jsx
soyarsauce: Thanks Nanda this is a good catch, I was too focused on whether `requestPermission` existed!

na9da:
- [x] Confirmed that it works in Safari & Chrome for iOS in portrait mode.

soyarsauce:
https://github.com/TerriaJS/terriajs/issues/4097

PR #4089, opened 2020-02-26
GeoRSS Catalog Item

### LINKED ISSUE #3943: GeoRSS Support
Several Australian government data custodians publish incident data using the GeoRSS format:

Examples:
http://esa.act.gov.au/feeds/currentincidents.xml
https://www.qfes.qld.gov.au/data/alerts/bushfireAlert.xml
http://emergency.vic.gov.au/public/atom-cap.xml

This geospatial format is non-proprietary, well-documented and well-established. I'd like to propose it's inclusion in the list of supported formats when adding Web Data.

zoran995:
support for GeoRSS catalog item, resolves #3943

### DIFF
--- a/lib/Models/GeoRssCatalogItem.js
+++ b/lib/Models/GeoRssCatalogItem.js
@@ -241,21 +241,27 @@ function resolveCatalogMetaData(documentElement, geoRssItem) {
     } else if (child.localName === "copyright") {
       geoRssItem.copyrightText = child.copyright;
     } else if (child.localName === "author") {
-      var author = child.children;
-      //TODO: check if this can be dataCustodian or author of the item
-      if (author.length === 0) {
-        geoRssItem.dataCustodian = child.textContent;
-      } else {
-        var name;
-        var email;
-        for (var authorIndex = 0; authorIndex < author.length; ++authorIndex) {
-          var authorChild = author[authorIndex];
+      var authorNode = child.childNodes;
+      var name;
+      var email;
+      for (
+        var authorIndex = 0;
+        authorIndex < authorNode.length;
+        ++authorIndex
+      ) {
+        var authorChild = authorNode[authorIndex];
+        if (authorChild.nodeType === 1) {
           if (authorChild.localName === "name") {
             name = authorChild.textContent;
           } else if (authorChild.localName === "email") {
             email = authorChild.textContent;
           }
         }
+      }
+      //TODO: check if this can be dataCustodian or author of the item
+      if (name === undefined && email === undefined) {
+        geoRssItem.dataCustodian = child.textContent;
+      } else {
         geoRssItem.dataCustodian = name + ", " + email;
       }
     }
--- a/lib/Models/GeoRssCatalogItem.js
+++ b/lib/Models/GeoRssCatalogItem.js
@@ -231,7 +231,7 @@ function resolveCatalogMetaData(documentElement, geoRssItem) {
     ++childNodeIndex
   ) {
     var child = childNodes[childNodeIndex];
-    if (child.nodeType !== 1 && child.localName === "item") {
+    if (child.nodeType !== 1 || child.localName === "item") {
       continue;
     }
     if (child.localName === "title") {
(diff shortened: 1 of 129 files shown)


lib/Models/GeoRssCatalogItem.js
```diff
@@ -0,0 +1,273 @@
...
+    if (!defined(that.dataSource)) {
+      return;
+    }
+
+    return that.terria.currentViewer.zoomTo(that.dataSource);
+  });
+};
+
+function loadGeoRss(geoRssItem) {
+  return loadXML(proxyCatalogItemUrl(geoRssItem, geoRssItem.url)).then(xml => {
```
tephenavies: In TerriaJS 7 and below we parse catalog item javascript files with an es5 parser to generate documentation for them.  Could you please replace this with the old anonymous function syntax `function (xml) {`?

lib/Models/GeoRssCatalogItem.js
```diff
@@ -0,0 +1,273 @@
...
+      resolveCatalogMetaData(documentElement, geoRssItem);
+    } else {
+      throw new RuntimeError("document is not valid");
+    }
+    return promise;
+  });
+}
+
+function resolveCatalogMetaData(documentElement, geoRssItem) {
+  var children = documentElement.children;
```
tephenavies: This isn't supported in IE11. I think it can be replaced with iteration over `childNodes`. See https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children (the "Polyfill" section shows how `childNodes` can be used similarly).

tephenavies (changes requested):
Hi @zoran995. Thanks for the contribution! As you're aware of already we're focusing more on our `mobx` branch which we'll release as TerriaJS 8 later this year, so we won't be accepting many PRs to `master`. This PR was a nice easy additional catalog item, so I feel it's worth merging to `master`. I've checked this with a GeoRSS dataset on data.gov.au and it worked great (you can see that [here](http://ci.terria.io/zoran995-georss/#share=s-lpHKK2bSWm0FWY0usMB8RvAaB5i)).

I've written a couple of suggested improvements that I'd like you to review. I caught the IE11 incompatibility because of your tests, so thank you especially for those.

tephenavies:
I'm not sure what GitHub has done with the comparison here after I merged latest `master` back into your branch. [This commit](https://github.com/TerriaJS/terriajs/pull/4089/commits/9ff0d0f1fd44a04439157959574e3a55b069f8f2) are the only changes I made to your contributions.

lib/Models/GeoRssCatalogItem.js
zoran995: Fixed

zoran995:
Happens sometimes. IE support should be fixed now. This is on my high-priority list for mobx also

lib/Models/GeoRssCatalogItem.js
```diff
@@ -0,0 +1,276 @@
...
+      continue;
+    }
+    if (child.localName === "title") {
+      geoRssItem.name = child.textContent;
+    } else if (child.localName === "description") {
+      geoRssItem.description = child.textContent;
+    } else if (child.localName === "copyright") {
+      geoRssItem.copyrightText = child.copyright;
+    } else if (child.localName === "author") {
+      var author = child.children;
```
tephenavies: We both missed this other instance of `Element.children` here, and I also forgot to ask if you could please write an entry into CHANGES.md. You should create a new section right at the top and label it something like "next release" and add one line about the GeoRSS catalog item in that section.

Once those 2 issues are resolved I'll merge it.

lib/Models/GeoRssCatalogItem.js
```diff
@@ -0,0 +1,282 @@
...
+function resolveCatalogMetaData(documentElement, geoRssItem) {
+  var childNodes = documentElement.childNodes;
+
+  for (
+    var childNodeIndex = 0;
+    childNodeIndex < childNodes.length;
+    ++childNodeIndex
+  ) {
+    var child = childNodes[childNodeIndex];
+    if (child.nodeType !== 1 && child.localName === "item") {
```
tephenavies: Should this line instead be the following?
```suggestion
    if (child.nodeType !== 1 || child.localName === "item") {
```

I don't think it will change the overall logic of the function as none of the if statements will be true anyway.
Also should this if statement also exclude "entry" from being processed?

zoran995: I missed that you added this as a suggestion, so I updated it on my end, an entry should be there and is now

tephenavies (approved):
Thanks for making those changes!

PR #4091, opened 2020-02-27
Fix overflowing catalog item names

reginapramesti:
Resolves https://github.com/TerriaJS/nsw-digital-twin/issues/321 by keeping the name div a fixed width, and wraps long strings to the next line


reginapramesti:
Check: add the layers with long names under `Marine water quality grids for the Great Barrier Reef region - monthly data`

lib/ReactViews/Workbench/workbench-item.scss
```diff
@@ -94,4 +95,6 @@ $workbench-margin: $padding-small;
 .draggable {
   font-size: $font-size-mid-small;
   cursor: move;
+  max-width: 100%;
+  overflow-wrap: break-word;
```
rowanwins: A bit nit-picky but I'd perhaps add the `overflow-wrap` to the `.nameColumn` class because that's what it's really targeting. The draggable class is more generic.

rowanwins (changes requested):
1 suggested change but looks good otherwise

lib/ReactViews/Workbench/workbench-item.scss
reginapramesti: Yup that makes sense!

rowanwins (approved):
Love it!

PR #4098, opened 2020-03-01
Location search by coordinates

zoran995:
Recently @RussellGrew asked about the possibility to find a location with its coordinates. Currently, this is only possible with the usage of the Bing Maps location search provider, which requires the Bing Maps key defined. It would be nice to have some kind of fallback for this search when the key is not available.


zoran995:
reopened in #4100

PR #4100, opened 2020-03-01
Location search by coordinates

### LINKED ISSUE #4099: Point search by coordinates
Recently @RussellGrew asked about the possibility to find a location with its coordinates. Currently, this is only possible with the usage of the Bing Maps location search provider, which requires the Bing Maps key defined. It would be nice to have some kind of fallback for this search when the key is not available.

zoran995:
Resolves #4099 
- point search is done in format latitude, longitude (i.e. -37.84,144.97);
- check for the existence of the Bing Maps Key is not defined because it is meant to be called from TerriaMap instance (this pull request just creates a possibility for a user to implement this search if needed).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 
 * Pass minimumLevel, in Cesium, to minNativeZoom, in Leaflet.
 * Upgraded to Cesium v1.66.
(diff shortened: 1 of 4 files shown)


zoran995:
If we get further interest in this, it shouldn't be hard to implement this in v8. Closing this as stale

PR #4076, opened 2020-02-24
Mobx: mapserver name

mgcsls:
Try adding https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer and https://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer/0.

Before: 
image

After: 
image

PR #4101, opened 2020-03-02
Mobx: fix blank name on share url when sharing esri mapserver

reginapramesti:
Resolves https://github.com/TerriaJS/nsw-digital-twin/issues/322


rowanwins:
Not sure this completely fixes things
**Pre Share**
PreShare

**Post Share**
PostShare

I think perhaps there is something going on in the serialising that means the name isn't being saved, so rather than trying to guess the name when something comes back in post-share, we should make sure the the name gets saved pre-share

reginapramesti:
That's fair it's definitely a cleaner solution, I'll look into why name isn't added pre-share. 

Though I was thinking of all the numerous share links we've generated for launch-related stories that don't have `name` as part of the share json 😬 wondering if I should keep this rather hacky solution for those cases, or if we should just redo the old stories and re-share them

na9da:
Merged in mobx branch to fix the CI heap issue.

reginapramesti:
@rowanwins so the reason why the name doesn't get saved in the share data is because the name is added as part of the traits in the `default` stratum. The share data only contains traits that are part of the `user` stratum. At the moment, when adding web datasets, the only traits that get saved in the `user` stratum is the `url`. 

https://github.com/TerriaJS/terriajs/blob/a56675e141c449b4e9062aef78bd74b0f9fded7e/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx#L101

I've tried to add another line after the above line that would set the default `name` trait to the url of the data, but this causes problems with the `get name()` function in `ArcGisMapServerCatalogItem`, which changes the name of the layer on the workbench without setting it as a name trait in the `user` stratum. This causes post-share workbench to set the url as the catalog layer name, rather than the layer name that was set in `get name()`.

What I've done in my commit is essentially take the name that was set in `get name()` in `ArcGisMapServerCatalogItem` and add it as part of the `name` trait in the user stratum. I couldn't think of a better way to fix this, but happy to take ideas!

rowanwins:
Think this was mostly resolved by #4076

PR #4106, opened 2020-03-02
Port location filter to mobx

na9da:
Test config:

```
{
  "catalog": [
    {
      "name": "Sentinel - moment lines",
      "type": "wms",
      "layers": "s2_ard_granule_nbar_t",
      "url": "https://ows.dea.ga.gov.au/",
      "chartType": "momentPoints",
      "opacity": 1,
      "timeFilterPropertyName": "data_available_for_dates"
    }
  ]
}
```


lib/Traits/TimeFilterTraits.ts
```diff
@@ -0,0 +1,45 @@
...
+import LatLonHeightTraits from "./LatLonHeightTraits";
+import objectTrait from "./objectTrait";
+import primitiveTrait from "./primitiveTrait";
+import mixTraits from "./mixTraits";
+
+export class TimeFilterCoordinates extends mixTraits(LatLonHeightTraits) {
+  @primitiveTrait({
+    type: "number",
+    name: "x",
+    description: "X coordinate of the tile"
```
rowanwins, line 11: These traits are very specific to how WMS works - I wonder if we should name them accordingly?

rowanwins: Or perhaps make reference to WMS in the description. 

The focus should be on the `LatLonHeightTraits` as that's what we're filtering by, but here it's pretty easy to miss that and get confused by the `x`, `y`, `level` which are really WMS specific.

rowanwins (approved):
This all works as it should, good job @na9da 

My only comment is that some of the trait naming or descriptions could perhaps be tweaked as currently it sounds more generic than it is in reality.  There are some really WMS specific stuff in there.

lib/Traits/TimeFilterTraits.ts
na9da: Have renamed the traits and added a pointer to the cesium doc in the trait description.

rowanwins: As per discussion with Nanda it turns out we're just following the approach of Cesium for pickFeatures

PR #4109, opened 2020-03-02
Set travis heap size to 4gigs for both jobs.

na9da:


soyarsauce:
Will merge once the jobs are ✅ !

soyarsauce (approved):

PR #4118, opened 2020-03-03
Bump CI nodejs to v10.

na9da:
v8 is end of life.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
   - language: node_js
     node_js:
-    - '8'
+    - '10'
     script:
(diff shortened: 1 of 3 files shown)


soyarsauce:
Some doc updates should go with this, as well as the hint in `package.json` which still says v8

soyarsauce (approved):
I'm for the most part, in agreement with getting this in for mobx.
& Pete says if we still need it, can always grab v8 for whichever image we are building in.

>ahhhh but if we build terriajs with one version it means we have that version on dev to build terriajs-server as well, meaning we wouldn't catch out any need-node-8 problems.

PR #4120, opened 2020-03-04
Add support for text and region columns for styling CSV data

rowanwins:
Prior to this PR CSV's would only be styled if they detected a TableColumn of type `enum` or  `scalar`. This would mean that a CSV containing region names wouldn't be visible.

This PR applies enum styling to `region` or `text` `TableColumn`'s which  seems reasonably sensible I think (giving unique colours to a bunch of unique values)

Try loading this CSV via the My data panel
[region.csv.txt](https://github.com/TerriaJS/terriajs/files/4284916/region.csv.txt)


rowanwins:
Addresses https://github.com/TerriaJS/nsw-digital-twin/issues/339

tephenavies:
I believe this will style differently to current master in that all regions will get a different colour. Is this change intentional. Is colouring each region differently better?

rowanwins:
I think it's the same as master

screenshot-ci terria io-2020 03 04-15_41_55

rowanwins:
Although in master it looks like it's only picking up the text column, rather than the region column, so perhaps that's the difference...

tephenavies:
This PR doesn't work with this CSV: http://ci.terria.io/mobx-col-styling/#share=s-6e2jIn5rLy0hLqcWELztKuo3z33

Which works on master: http://ci.terria.io/master/#share=s-aMEdz8IuZi9ES5gVcKfaj1e9P2b

rowanwins:
I'll take a look-see, thanks for the report Steve.

For completeness here is the CSV link, it contains 1 column
https://gist.githack.com/steve9164/dc5bb6cd77ba5121ee3163af5ec467e9/raw/4484792e589bbee0ee8b061bd804fcffd19f5918/sa4.csv

rowanwins:
So the CSV works if using the 'Add local data' approach, but via the URL it gets a 403.

It looks like hits an `OPTIONS` request whereas on the `master` branch it's requested with a GET request.
Screen

If I put the githack gist in a github gist it works fine
https://gist.githubusercontent.com/rowanwins/84045187b48c46408c659b3c97d8be5c/raw/39925d8ace52fc808ac92720f27d0c572a2ebd6b/sa4.csv

And if I add `"githack.com"` to the `allowProxyFor` list in then it works again. Although on master example above it doesn't go via the proxy ¯\_(ツ)_/¯

tephenavies:
Master:
image


This PR:
image

rowanwins:
Yep different behaviour although I think I prefer the proposed...

tephenavies:
Two reasons to keep to `master`'s single colour for all regions are:
- When you have a lot of regions you want to avoid colouring some regions the same colour as other regions when they are completely unrelated
- Keep the same behaviour as point CSVs

See this share for both of those problems: http://ci.terria.io/mobx-col-styling/#share=s-kshGXM5lBySTnFoX2aqMykH7YSy
 
Lastly, whatever decision we make we should document it.

rowanwins:
Have rejigged as per Steve's request

PR #4125, opened 2020-03-06
Mobx: Fallback name for ArcGisServerCatalogItem names

rowanwins:
Related to https://github.com/TerriaJS/nsw-digital-twin/issues/322

Add this MapServer and notice that the name is shown as the url
https://sampleserver6.arcgisonline.com/arcgis/rest/services/911CallsHotspot/MapServer

Example in [mobx branch](http://ci.terria.io/mobx/#share=s-hdcKz6762p9UMh6HiM8HOix1Oqw)

This service doesn't have anything defined in the `documentInfo`, it does however have something in the `mapName` so we can use that instead.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -78,6 +78,7 @@ Change Log
 * Updated upload icon to point upwards
 * Prevent catalog item names from overflowing and pushing the collapse button off the workbench
 * Stopped analytics launch event sending bad label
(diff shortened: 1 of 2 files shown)


reginapramesti (approved):
Looks good! Thanks Rowan for fixing it in a nicer way! Why does MapServer have to have its layer name potentially listed in two different properties 🤷‍♀

PR #4128, opened 2020-03-06
Ignore invalid HTTP 200 responses for tile image requests.

### LINKED ISSUE #3990: `ignoreUnknownTileErrors` should also ignore invalid HTTP 200 responses
The docstring for the config param says:
>   Gets or sets a value indicating whether non-specific (no HTTP status code) tile errors should be ignored. This is a last resort, for dealing with odd cases such as data sources that return non-images (eg XML) with a 200 status code. No error messages will be shown to the user.

But currently invalid 200 responses raise an error and the item is disabled. 
Eg: The `Seasonal fractional cover..` item in https://map.geo-rapp.org/#share=s-MlbU7csc1geW5WRh5PkKnytbUQ 

The reasoning is (from Kevin):

> downloads via an Image DOM element never return http status codes, so they're always unknown
so that's probably the basis for that statement in the doc. but I think these days we retry failed Image loads with XHR, and get a status code that way, so the error probably won't be considered unknown anymore

We need a fix so that `ignoreUnknownTileErrors` will also ignore invalid 200 OK responses.

na9da:
Fixes #3990 
Ignores image responses that are 200 OK but are actually something else (like a service error XML), instead of throwing an error.

### DIFF
--- a/lib/Models/ImageryLayerCatalogItem.js
+++ b/lib/Models/ImageryLayerCatalogItem.js
@@ -1183,17 +1183,21 @@ ImageryLayerCatalogItem.enableLayer = function(
             } else if (!defined(e.statusCode) && defined(e.target)) {
               // This is a failed image element, which means we got a 200 response but
               // could not load it as an image.
-              failTile({
-                name: i18next.t("models.imageryLayer.tileErrorTitle"),
-                message: i18next.t("models.imageryLayer.tileErrorMessageII", {
-                  url: getUrlForImageryTile(
-                    imageryProvider,
-                    tileProviderError.x,
-                    tileProviderError.y,
-                    tileProviderError.level
-                  )
-                })
-              });
+              if (catalogItem.ignoreUnknownTileErrors) {
+                tellMapToSilentlyGiveUp();
+              } else {
+                failTile({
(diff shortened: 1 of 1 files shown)


tephenavies (approved):

lib/Models/ImageryLayerCatalogItem.js
```diff
@@ -1183,17 +1183,21 @@ ImageryLayerCatalogItem.enableLayer = function(
             } else if (!defined(e.statusCode) && defined(e.target)) {
```
tephenavies, line 1183: This condition looks like it's the wrong way around. Is this meant to be `defined(e.statusCode) && !defined(e.target)`? Just from the comment it looks more reasonable to be that.

na9da: Looks alright to me.

>   // This is a failed image element, which means we got a 200 response but
>    // could not load it as an image.

What happens is the result of the xhr is set to an img element as a blob. But the img element errors out because the blob is invalid. The `e.target` is the img element that errored.

https://github.com/CesiumGS/cesium/blob/1.67/Source/Core/Resource.js#L1800

tephenavies: Ok. Thanks for the explanation.

PR #4140, opened 2020-03-12
fixed Cesium3DTilesCatalogItem to work with ionAssetIds

nf-s:
### Fix Cesium3DTilesCatalogItem `ionAssetIds`

You could only create catalog items with the `url` property, but not the `ionAssetIds`. Whereas you should be able to create one with either property.

For example:

```json
{
  "name": "Test",
  "type": "3d-tiles",
  "ionAssetId": 1234,
}
```

```json
{
  "name": "Brisbane 3D city model (aero3Dpro)",
  "type": "3d-tiles",
  "url": "https://sample.aero3dpro.com.au/BrisbaneCBD/Scene/recon_h_3DTiles.json",
  "options": {
    "maximumScreenSpaceError": 1,
    "maximumNumberOfLoadedTiles": 1000
  },
  "rectangle": [
    153.01,
    -27.484,
    153.034,
    -27.46
  ]
}
```

### DIFF
--- a/lib/Models/Cesium3DTilesCatalogItem.ts
+++ b/lib/Models/Cesium3DTilesCatalogItem.ts
@@ -92,12 +92,8 @@ export default class Cesium3DTilesCatalogItem
   }
 
   private loadTileset() {
-    if (!isDefined(this.url)) {
-      return;
-    }
-
     const tileset = this.createNewTileset(
-      proxyCatalogItemUrl(this, this.url),
+      this.url,
       this.ionAssetId,
       this.ionAccessToken,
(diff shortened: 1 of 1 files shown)


rowanwins:
You mean to tell me there is a provider of 3D tiles other than Ion 😂

nf-s:
To clarify, this fixes a bug where you could only create `3d-tiles` layers if they had a `url` property. That is the you **could not** create a layer which had a `ionAssetId` but **no** `url`

KeyboardSounds (approved):

PR #4149, opened 2020-03-17
Setup framework for error provider services

rowanwins:
### What this PR does

First part of https://github.com/TerriaJS/TerriaMap/issues/430

This PR sets up some basic pieces for configuring error handling services like rollbar, or sentry etc. A user would pass down a parameter via `config.json` including the API key for the respective service.

An app will also need some configuration in TerriaMap index.js to specify which ErrorProvider they want to use. This should prevent things like the rollbar library from being bundled in the application if it's not required.

### Checklist


soyarsauce:
https://github.com/TerriaJS/handbook/issues/6

soyarsauce (approved):

soyarsauce:
See https://github.com/TerriaJS/TerriaMap/pull/432#discussion_r398253106 && https://github.com/TerriaJS/TerriaMap/pull/432#issuecomment-604162755

PR #4151, opened 2020-03-18
WIP Mobxify MapboxVectorTileCatalogItem

nf-s:
### Mobxify MapboxVectorTileCatalogItem

It seems to work.

I'm not sure how to handle `legends` set by manual properties like `lineColor` and `fillColor` - see line 40 in `lib/Models/MapboxVectorTileCatalogItem.ts`

### TODO

- [ ] Add sample catalog JSON


lib/Models/MapboxVectorTileCatalogItem.ts
```diff
@@ -0,0 +1,149 @@
...
+    return new MapboxVectorTileLoadableStratum(
+      newModel as MapboxVectorTileCatalogItem
+    ) as this;
+  }
+
+  static async load(item: MapboxVectorTileCatalogItem) {
+    return new MapboxVectorTileLoadableStratum(item);
+  }
+
+  // @computed get legends() {
```
nf-s: As stated in decription:
I'm not sure how to handle `legends` set by manual properties like `lineColor` and `fillColor`

tephenavies: I think a computed on MapboxVectorTileCatalogItem is best. Just check `super.legends` and return that if it exists. See e.g. https://github.com/TerriaJS/terriajs/blob/b131abaa77fa8aa9a6f89238951f33cbf0ac4130/lib/Models/ArcGisMapServerCatalogItem.ts#L435-L446

na9da: I have defined the `legends` inside the `LoadableStratum`. I think extending like above is easy for `primitiveTrait`s as they return a simple type `T`, but it is difficult for more complex traits returning a `Model<T>`.

na9da:
Sample catalog definitions: https://gist.github.com/na9da/aada750e2bb9501f3088df819a0db576

lib/Models/MapboxVectorTileCatalogItem.ts
```diff
@@ -0,0 +1,152 @@
...
+    autorun(() => {
+      if (
+        !isDefined(this.url) ||
+        !isDefined(this.layer) ||
+        !isDefined(this.rectangle)
+      ) {
+        return;
+      }
+
+      this.imageryProvider = new MapboxVectorTileImageryProvider({
```
tephenavies: This should be a computed, not an autorun setting an observable.

lib/Models/MapboxVectorTileCatalogItem.ts
```diff
@@ -0,0 +1,152 @@
...
+        url: this.url,
+        layerName: this.layer,
+        styleFunc: action(() => ({
```
tephenavies: You could avoid this action using a closure. e.g:

```js
styleFunc: ((style) => () => style)({
  fillStyle: this.fillColor, 
  strokeStyle: this.lineColor, 
  lineJoin: "miter", 
  lineWidth: 1
})
```

or:

```js
styleFunc: (({fillColor, lineColor}) => () => ({
  fillStyle: fillColor, 
  strokeStyle: lineColor, 
  lineJoin: "miter", 
  lineWidth: 1
}))({
  fillColor: this.fillColor, 
  lineColor: this.lineColor
})
```
This will evaluate the observables/computed early and the resultant `styleFunc` will no longer use any tracked variables

lib/Models/MapboxVectorTileCatalogItem.ts
```diff
@@ -0,0 +1,152 @@
...
+        imageryProvider: this.imageryProvider,
+        show: this.show,
+        alpha: this.opacity,
+        canZoomTo: this.canZoomTo
+      }
+    ];
+  }
+
+  @action.bound
+  featureInfoFromFeature(feature: VectorTileFeature) {
```
tephenavies, line 131: I don't think this will be called. `featureInfoFromFeature` is a `TableMixin` function and `TableMixin` isn't being used here.

tephenavies (changes requested):

lib/Models/MapboxVectorTileCatalogItem.ts
tephenavies: The functionality here should either be ported to MVT Imagery Provider or deleted.

lib/Models/MapboxVectorTileCatalogItem.ts
na9da: Good point!

lib/Models/MapboxVectorTileCatalogItem.ts
na9da: Good point again!

lib/Models/MapboxVectorTileCatalogItem.ts
na9da: That function is used internally here - https://github.com/TerriaJS/terriajs/blob/mobx-vector-imagery-provider-ts-flat/lib/Models/MapboxVectorTileCatalogItem.ts#L110

tephenavies: Ahh ok. I missed that it's passed into the ImageryProvider.

tephenavies:
Specs passed on Travis CI push build but deployment failed due to what sounds like cluster issues:
```
Error: UPGRADE FAILED: "terriajs-mobx-vector-imagery-provider-ts-flat" has no deployed releases
```

tephenavies (approved):

PR #4154, opened 2020-03-18
Add new menu item in help panel

reginapramesti:
### What this PR does

Fixes https://github.com/TerriaJS/nsw-digital-twin/issues/366

The changes made in this PR only consists of adding a new `HelpPanelItem`, essentially replacing the placeholder that was previously created for this new panel item. Note the scss files were automatically generated, and so far we've just been pushing it until Wing makes some changes.

### Checklist

-   [x] No unit tests exist so far, and the changes are only involving adding a new component with similar properties, so these specific changes shouldn't break anything in theory


lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
```diff
@@ -142,11 +142,11 @@ class HelpPanel extends React.Component {
                 "One of the key features of the NSW Spatial Digital Twin is the ability to add and view 3D data, including 3D imagery, or photo reality mesh. This is an interactive way to explore high resolution, 3D aerial imagery of some areas of NSW.",
```
rowanwins, line 142: >One of the key features of the NSW Spatial Digital Twin

@reginapramesti can we make this text not specific to the NSW Digital Twin? I'm wondering if we could pull this text from the lang file or the config settings for the map name or something...

rowanwins: In fact we need to think about how we handle all of those help items, they all reference NSW Digital Twin...

reginapramesti: That's true, I've been thinking about this for a while but kept forgetting because I didn't log an issue for it 😆 I'll log an issue and work on it, though probably will be a separate PR

rowanwins (approved):
Looks good

PR #4155, opened 2020-03-18
Fix overlaying text bug

reginapramesti:
### What this PR does

Fixes https://github.com/TerriaJS/nsw-digital-twin/issues/355

This PR fixes the bug described in the issue where the contents in the video panel of the help UI overlays the video itself. For some reasons the z-index css property of the video div got removed (I must've accidentally removed it from the code when I was tidying up 😬). Now it's re-added, the properties of the almost-full-screen-video div follows the properties of the video guide used in `story-panel.scss`, in that it has to be the top most component z-index-wise.

### Checklist

-   [x] No unit tests exist that tests the styling (css/scss) properties of the React Components (from what I've looked through so far), but will confirm with Wing first when he's back at work


rowanwins:
Nice work Regina!

PR #4159, opened 2020-03-20
Fix bug where viewers are recreated

tephenavies:
### What this PR does

Fixes a bug where a TerriaMap would crash or reset zoom back to the home camera when changing 3D settings or 3D terrain provider while displaying 3D datasets.

### Checklist

-   [x] ~There are unit tests to verify my changes are correct~ Unit tests removed because they didn't work on release builds 


rowanwins (approved):
Looks good to me

tephenavies:
To summarise my changes I:
- Changed the way Leaflet and Cesium promises worked to simplify:
    + a computed that fetches the chunk and uses an async action to set an observable that the original computed then returns, to 
    + a computed that uses `fromPromise` to download and return the contents of the chunk
- Changed computed `get currentViewer` to use `untracked` for everything except the things that should be able to make it recompute

PR #4185, opened 2020-03-31
Antimeridian handling in LeafletJS

rowanwins:
### What this PR does
Fixes https://github.com/PacificCommunity/Terria-PacificMap/issues/34

Reworks the anti-meridian handling of points in the leaflet viewer so that it only gets applied in certain circumstances. The previous logic kinda worked for small scale datasets but was flawed when dealing with points covering the globe.

Try using this Esri FeatureService for example
https://services1.arcgis.com/0MSEUqKaxRlEPj5g/ArcGIS/rest/services/ncov_cases/FeatureServer/1

In the new logic
 - we track whether we're zoomed in around the anti-meridian, if not points aren't adjusted
 - we keep track of 3 viewing modes, normal (no anti-meridian in play), left-side anti-meridian (when the map bbox northEast is 180+) and right-side anti-meridian (when the map bbox southWest is < 180)
 - if we change viewing mode (because we've zoomed or panned across the anti-meridian area) then we'll double check if our points need tweaking.

### Checklist

### DIFF
--- a/lib/Map/LeafletVisualizer.js
+++ b/lib/Map/LeafletVisualizer.js
@@ -165,6 +165,8 @@ function cleanPolyline(entity, group, details) {
   }
 }
 
+let prevBoundsType = 0;
+
 /**
  * Updates the primitives created by this visualizer to match their
  * Entity counterpart at the given time.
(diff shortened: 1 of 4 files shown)


lib/Map/LeafletVisualizer.js
```diff
@@ -178,22 +180,37 @@ LeafletGeomVisualizer.prototype.update = function(time) {
     throw new DeveloperError(i18next.t("map.leafletVisualizer.devError3"));
   }
   //>>includeEnd('debug');
+  const bounds = this._leafletScene.map.getBounds();
+  let applyLocalisedAntiMeridianFix = false;
+  let currentBoundsType = 0;
+  if (
+    bounds._southWest.lng > 140 &&
+    (bounds._northEast.lng < -140 || bounds._northEast.lng > 180)
```
tephenavies: Needs a comment. Does this detect looking at the antimeridian with the centre being to the east of the antimeridian and being zoomed in?

lib/Map/LeafletVisualizer.js
```diff
@@ -178,22 +180,37 @@ LeafletGeomVisualizer.prototype.update = function(time) {
...
+  ) {
+    applyLocalisedAntiMeridianFix = true;
+    currentBoundsType = 1;
+  } else if (
+    (bounds._southWest.lng > 180 || bounds._southWest.lng < -140) &&
+    bounds._northEast.lng < -140
```
tephenavies: Also needs a comment. Is this one is detecting looking just to the west of the antimeridian, or possibly at the antimeridian and being zoomed in?

tephenavies (changes requested):
Unfortunately I couldn't get your example dataset to load on http://ci.terria.io/master/.

I tested performance with http://ci.terria.io/antimeridian-again/#share=s-oYqgkdCLUF2KSWlGRG4VtEB1880 and it seemed alright. There were some strange cases around browser width.

I'd prefer that we not allow zooming out past a bounds of 360 degrees longitude and then just put features in the right place anywhere on the map. I guess that doesn't work too well with Leaflet's DOM implementation.

Other than fixing it like that, I think with a few changes this should be enough.

rowanwins:
The sample server doesnt work because there is a null geometry and our handling of that doesn't work in master. This PR also resolves that :)

tephenavies (approved):

PR #4213, opened 2020-04-08
Mobx tighten gyro

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4046

(See checklist in issue)

### Checklist


soyarsauce:
FYI @vicborgy this can be viewed at http://ci.terria.io/mobx-tighten-gyro/, please have a look and let me know what other tweaks are needed 🙂 

Noting a few things:
- Tried to keep everything scoped just to the compass/gyro component, to minimise clashes with https://github.com/TerriaJS/terriajs/pull/4165
- I'm very aware the "Controls" modal differs slightly to what's in sketch/invision, but it's because we repurpose the map settings/related maps/share+print modals, and am happy for you to open another ticket for revamping all these later down the track
- the 0.85 opacity applies to the whole modal (meaning the text and icons get affected too), it's possible to do just the background too but along with the need-for-revamped-panel, i've gone with this simpler way in the meantime
- the font family is open sans in designs, but we've redone everything with nunito for now

vicborgy:
Thanks @soyarsauce, looking good. Just a few things that stand out to me:
- [x] Can we make the rollover more of a fade transition rather than just popping in
- [x] The close 'X' button for the controls should be black
- [x] Ive created a design for the actual controls and just updated the NSW DT Invision designs - https://invis.io/6GVY8M6WJ89#/403934755_NSW_Digital_Twin_-_Gyro_Controls_V2a

soyarsauce:
Thanks @vicborgy - give it another ~15 minutes from this comment for CI to redeploy the new updates, and those 3 should be addressed on that same ci link

soyarsauce:
image
image

A few instances where the black "close X button" is a little problematic (top: default bing maps basemap, bottom: dark matter basemap)

vicborgy:
Hi @soyarsauce I appreciate what you're saying but guess the same can be said for when the maps are all white. So int this case I'd much prefer it to be as per my designs so it differs from a top level buttons that use the white background. Cheers

soyarsauce:
Hey @vicborgy no problem, are you happy with the fixes & for this to be merged?

vicborgy:
hey @soyarsauce looks good bro. One other thing, I dont suppose making the hover state less jumpy and more of an animated expand is possible in this ticket? If a smooth animation is too much even just fading it in would look much nicer.

soyarsauce:
Hey @vicborgy - which part do you mean by "hover state"? When you hover over the gyro button and the "close/controls/help" appears? Because that bit is already a fade in/out

vicborgy:
sorry @soyarsauce I mean the main gyro
Screenshot 2020-04-20 12 25 02

lib/Styled/Spacing.jsx
```diff
@@ -9,6 +9,10 @@ import styled from "styled-components";
 //   ${props => props.bottom && `margin-bottom: ${props.bottom * 4}px;`}
 // `;
 export const Spacing = styled.div`
+  // Unsure how to deal with 1px right now
```
reginapramesti, line 12: Maybe we could do it like the way box padding works - there's padding with ratio/multiplier vs normal padding? Alternatively does 0.2 * 5 work if you want it to be scalable? 😆 though I understand it's not very intuitive

reginapramesti:
I think it looks great @soyarsauce 👏 These comments are pretty nitpicky/are just things I'm not sure about 
* Hovering over the buttons shows the same information as hovering over the compass - should these buttons have their own specific descriptions or have it removed entirely?
* Should the dropdown panel hide when opening another panel (eg. Explore data)? Unless we've made a conscious decision not to?

reginapramesti:
PS thanks for writing those tests, I'm gonna follow them from now on 😄

lib/ReactViews/GyroscopeGuidance/GyroscopeGuidance.jsx
```diff
@@ -27,6 +30,10 @@ const CompassWrapper = styled(Box).attrs({
   width: 64px;
   height: 64px;
   margin-right: 10px;
```
reginapramesti: Also maybe this could be turned into a `<Spacing right />` component?

reginapramesti:
And this is from a while ago (launch crunch time), but might be worth changing the padding to props `paddedVertically`

https://github.com/TerriaJS/terriajs/blob/22ed20c1e8ac547d2e6edaa934bf89ea1362c87e/lib/ReactViews/GyroscopeGuidance/GyroscopeGuidance.jsx#L70-L74

lib/Styled/Spacing.jsx
soyarsauce: Yeah wouldn't be too intuitive but more importantly, once/if we change to say, `4`, then you get fractions when you multiply by 0.2 - if it's 1px we just do 1px.. 😬

lib/ReactViews/GyroscopeGuidance/GyroscopeGuidance.jsx
soyarsauce: Good pickup - done

soyarsauce:
>Hovering over the buttons shows the same information as hovering over the compass - should these buttons have their own specific descriptions or have it removed entirely?

fixed the same title-everywhere problem, and will let the inner buttons be labelled via it's inner text

>Should the dropdown panel hide when opening another panel (eg. Explore data)? Unless we've made a conscious decision not to?

Not concious but will leave it & should be addressed by https://github.com/TerriaJS/terriajs/issues/3958

Thanks regina, any more? 😄 @reginapramesti

lib/Sass/common/_variables-export.scss
```diff
@@ -6,13 +6,29 @@
 @import "terriajs-variables";
 
 :export {
+  // TODO: find a better solution for this - some of these we need to provide as
+  // numbers only so we can do math on them and add the unit afterwards 😵
+  // mobile: $mobile;
+  // sm: $sm;
+  // md: $md;
+  // lg: $lg;
+  mobile: 767;
```
KeyboardSounds, line 15: These are widths, yeah? Which is reasonably self evident but it might be worth sticking a comment in here to make it explicit

KeyboardSounds (approved):

lib/Sass/common/_variables-export.scss
soyarsauce: Tidied them a little 🙂

PR #4215, opened 2020-04-09
Upgrade to cesium 1.68 and publish v7.11.4

reginapramesti:
### What this PR does

Partially addresses https://github.com/TerriaJS/terriajs/issues/4198

* Upgraded the version of `terriajs-cesium` on `package.json` to 1.68. 
* Because of this upgrade, had to replace all of `defineProperties` and `freezeObject` that was previously dependent on Cesium's functions to use built in `Object` functions instead, to accommodate the change
* From this change, getters are made to return something within the `Object.defineProperties` function to fix lint errors
* Travis builds apparently need v10+ now, so bumped node version on travis to v10 now
* Also removed running `docs` on `gulp` to fix errors (current gulp command copied from mobx branch)

### Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
   - language: node_js
     node_js:
-    - '8'
+    - '10'
     script:
     - npm run prettier-check
     - pyenv shell 2.7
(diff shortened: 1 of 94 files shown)


tephenavies (approved):
Looks good, and works well.

Issue #4216, opened 2020-04-09
Mobx: User added web data issues


zoran995:
There are some issues in add web data option:

- Single Layers (ESRI ArcGIS MapServer Single Layer),
  - data is automatically added to the map, and there are options "remove from map" and "remove from catalogue", whichever is clicked data is removed from the catalogue, and data stays on the map. - _this is resolved_
  - If the user closes the explorer window and removes dataset from the map using workbench remove option, and again open the explorer data will be removed from the catalogue when he clicks add to the map. - _this is resolved_
- WebMapServer doesn't set the name from capabilities response (for example add: [wms](http://marine.ga.gov.au/geoserver/ows?service=wms&version=1.3.0&request=GetCapabilities) - the name should be Geoscience Australia Marine Data), layer names are fine from what I see - _this is resolved_
- KML Layer 
[...]

PR #4221, opened 2020-04-12
WMS group name, hideSource, description and item strats

zoran995:
### What this PR does

- Added an option to hide a source of catalog members (defaults false), support for it was already defined in view part
- Added `Title`, `ContactInformation` and `Fees` to `WebMapServiceCapabilities`, so they are pulled with getCapabilities request.
- Added `get name()` for `WebMapServiceCatalogGroup`, so it resolves not setting name for user-added layer (described in #4216). There is a small issue with this approach for the catalogGroup defined in user catalogue without a name; it will be changed on metadata load from URL to the name resolved from capabilities response. Does anyone have an idea of what should be done for that part, or to leave it like that?
- Added `info` response for `WebMapServiceCatalogGroup` as it is done for ArcGisFeatureServerCatalogGroup. New attribute introduced is `fees`, unless it is `none`.
- Set the `isOpenInWorkbench`, `isExperiencingIssues`, `hideLegendInWorkbench`, `hideSource` for the catalogItems resolved from getCapabilities response. For me, it is logical if I define any of those properties on group-level that it should be respected on items also (unless defined differently for a specific item)
- Added more test for `WebMapServiceCatalogGroup`

Hope these changes are fine for you. This is my first work with new model architecture, so I hope I did it right.

### Checklist

-   [+] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [+] I've updated CHANGES.md with what I changed.


tephenavies:
Hi @zoran995. Thanks for the pull request. These changes are great. I've fixed a build problem with the tests and refactored the `newInfo` function that was redefined in multiple places and used to create info sections into it's own function (and file).

tephenavies (approved):

Issue #3303, opened 2019-04-15
Model architecture: Enable eslint for .ts files


kring:

kring:
https://github.com/Microsoft/TypeScript/issues/30553
https://github.com/typescript-eslint/typescript-eslint

tephenavies:
Our current eslint config is really outdated. Now `prettier` fills a lot of the requirements we used `eslint` for. We should maybe consider cutting it down a lot, starting with something like:
```json
{
  "extends": [
    "eslint:recommended",
    "prettier"
  ]
}
```
And add config to make it work with TypeScript etc.

We'll probably need some `react` rules too, e.g. see #4565.

soyarsauce:
Related https://github.com/TerriaJS/terriajs/pull/4519 & https://github.com/TerriaJS/terriajs/issues/4565

ljowen:
Blocked by #6991

ljowen:
PR: https://github.com/TerriaJS/terriajs/pull/6985

PR #4227, opened 2020-04-16
Changes of i18next config

zoran995:
### What this PR does

This is just a draft of some updates for i18next configuration, but it would be nice to discuss them.
This introduces
- Ability for the user to define which languages should be used and fallback language in TerriaMap `config` file (if not defined default values used). This should replace the instance of `i18n.js` file (used in `prerequisites`).
- Introduces usage of `i18next-xhr-backend` so language files get pulled from the backend (this will require exposing `Languages` so a browser can access them). More info about it [here](https://github.com/i18next/i18next-xhr-backend).
- Introduces dev-dependency `eslint-plugin-i18next`. This should wait for changes from this [pull request](https://github.com/edvardchen/eslint-plugin-i18next/pull/14). This would be useful to detect hard-coded strings. Needs a lot of work to separate all strings that should not be detected as hard-coded (currently detects over 1k of problems). Currently works only on `.js` and `.jsx` files, to make it work for `.ts` files #3303 is needed.

### Checklist


soyarsauce:
Just a quick note - I had a quick go at implementing a super lightweight usage of something similar to this PR, and saw `i18next-xhr-backend` is deprecated - `i18next-http-backend` (https://github.com/i18next/i18next-http-backend) is the favoured lib now, except it is still early days for it as I ran into some problems with using the options as they are developed out.

zoran995:
loading translations override from backend is moved to #4324, with the usage of `i18next-http-backend`

zoran995:
I think this is suppersed by #4519, while other parts are already implemented

PR #4229, opened 2020-04-17
WIP: Mobxify and extend CatalogFunctions

nf-s:
### Create CatalogFunction Mixin and Traits

Take common `CatalogFunction` functionality out of `WebProcessingServiceCatalogFunction` and put into `CatalogFunctionMixin`.

This has the following abstract properties:

```ts
abstract async invoke(): Promise<void>;
abstract get parameters(): FunctionParameter[];
```

I have also added the following functions:

```ts
createPendingCatalogItem()
setErrorOnPendingItem(errorMessage?: string)
```

Which may be useful for Catalog Functions other than WPS, but this lefts to be seen.

This is a first draft - it is a precursor to creating a YourDataYourRegions catalog function.

### Create `CatalogFunctionJob`

See comment below - https://github.com/TerriaJS/terriajs/pull/4229#issuecomment-624463759

### CatFun Generalities

- [ ] Spinner while running
- [ ] Cancel function (if supported by catalog item)
- [ ] Ability to share with pending catalog item
- [ ] Mobxify DropDown component
- [ ] Replace Select Elements with DropDown component

### Mobx / TypeScriptify FunctionParameters

*Note: I may not complete all of these*

- [x] Boolean Parameter
- [x] Function Parameter
- [ ] InfoParameter (this is empty?)
- [ ] RegionDataParameter
- [ ] BooleanParameterGroup
- [ ] createParameterFromType

### Mobx / TypeScriptify Parameters Editors

*Note: I may not complete all of these either*

- [ ] ParameterEditor
- [x] EnumParamEditor
- [x] BooleanParamEditor
- [ ] BooleanGroupParamEditor
- [x] GenericParamEditor
- [ ] LineParamEditor
- [ ] PointParamEditor
- [ ] RegionParamEditor
- [ ] GeoJsonParamEditor
- [ ] PolygonParamEditor
- [ ] DateTimeParamEditor
- [ ] RectangleParamEditor
- [ ] RegionDataParamEditor
- [ ] RegionTypeParamEditor
- [ ] SelectAPolygonParamEditor


nf-s:
# CatalogFunctionJob

## Initial thoughts

> So I'm thinking about how you can save catalog function state (or rather the invocation of a catalog function's state?) so you can share a terria map with a pending catalog function.
I was thinking maybe I could just extend ResultPendingCatalogItem to do this.  
> 
> So a CatalogFunction would have three things:
>- The function - which defines input parameters and invoke function
>- The pending item (or I was thinking maybe you call this a CatalogFunctionJob) which is created by invoke . This handles polling for job results and then, when finished, it creates...
>- The catalog item(s) for results
>
> And maybe the pending item would hang around after it finishes and it would contain all parameters for a job.

## CatalogFunction parts:

### CatalogFunctionMixin

- `parameters` to render `ParameterEditor` 
- `invoke` function which creates a `CatalogFunctionJob` (could be multiple jobs)
  - What is responsibility of `invoke` funciton?
    - Should it also handle submitting job to server? (for validation etc...)
    - I think not, let the job handle all that. If it fails, add a 'Retry' button or something which will populate `CatalogFunction` with set parameters.
    - Keep the CatalogFunction simple, it only handles input parameters, input logic and creating job

### CatalogFunctionJob 

- Not mixin at this stage.
- Extends `AutoRefreshTraits` - to handle polling for job results
- Has `JobStatus` and handles polling for results (or websocket or whatever)
- Has `parameters` from `CatalogFunction` - but only `id` and `value` - which is enough to load back into `CatalogFunction`
- Ability to see job progress/job logs
- Ability to cancel job (if supported by service)
- Ability to "re-invoke" a job - that is copy parameters back to `CatalogFunction` so to edit and create a new job.
  - If a job fails, this can be used to retry.

### CatalogFunctionItem / Handling result layers

I was thinking of maybe having a separate `CatalogFunctionItem` which handled creating result catalog items, but it might be better to do this in `CatalogFunctionJob`

A `CatalogFunctionJob` can create multiple `CatalogItem`s for results 
  - Is there a way to visually tie a `CatalogFunctionJob` with its child layers? 
  - Could also have a button to hide/show all layers, or to 'reset' layers if one is deleted.
  - A `CatalogFunctionJob` could appear as a CatalogGroup in the Data Catalog (under My Data)
  - Then you could have a history of jobs run, you could re-add previous results to Workbench (which would create all result layers again)

nf-s:
## First steps

- [x] Create `CatalogFunctionJob` mixin
  - Add parameters trait with kv-pair
- [ ] `invoke` creates `CatalogFunctionJob`

nf-s:
Closed - this will be apart of #4406

PR #4230, opened 2020-04-17
[2227] Now is't possible to use Mustache template to define an item e…

### LINKED ISSUE #2227: Why are these CSW record titles repeated?
I'm wondering why certain CSW record titles are repeated, as in this screenshot:

You can explore this by visiting 

I notice this only on WMS endpoints from THREDDS Data Servers.

ccancellieri:
…lement name

### What this PR does

Fixes #2227 

I've seen that it's very hard to define a common way to present items and usually if a *record* has multiple external WMS resources (this also happens in CKAN) all the resources will have the same name.
I've also noticed that we have [Mustache](https://docs.terria.io/guide/connecting-to-data/customizing-data-appearance/feature-info-template/) support for the getFeatureInfo template.
What I did is simply add a new configuration node into the item properties to define a template for the itemName.

*Not sure what to do with tests* (but let's see if the patch can be accepted firts).

I'm open to comments, we really need this functionnality, please.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 ### next release
 
 * Added `GeoRssCatalogItem` for displaying GeoRSS files comming from rss2 and atom feeds.
+* Now is't possible to use Mustache template to define an item element name
 
 ### v7.11.4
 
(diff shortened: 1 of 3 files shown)


ccancellieri:
apologize, problem with squash and lint
going to retry

PR #4232, opened 2020-04-17
[2227] Now it's possible to use Mustache template to define an item name

### LINKED ISSUE #2227: Why are these CSW record titles repeated?

ccancellieri:
…lement name
Fixes #2227

I've seen that it's very hard to define a common way to present items and usually if a record has multiple external WMS resources (this also happens in CKAN) all the resources will have the same name.
I've also noticed that we have Mustache support for the getFeatureInfo template.
What I did is simply add a new configuration node into the item properties to define a template for the itemName.

Implemented test using jasmine match()

I'm open to comments, we really need this functionnality, please.

### Checklist

### DIFF
--- a/lib/Models/CswCatalogGroup.js
+++ b/lib/Models/CswCatalogGroup.js
@@ -1,5 +1,7 @@
 "use strict";
 
+import Mustache from "mustache";
+
 /*global require*/
 var ArcGisMapServerCatalogItem = require("./ArcGisMapServerCatalogItem");
 var CatalogGroup = require("./CatalogGroup");
@@ -53,6 +55,15 @@ var CswCatalogGroup = function(terria) {
    * @type {String}
    */
   this.getRecordsTemplate = undefined;
+
+  /**
+   * Gets or sets a template to display group item names.
+   * May be a string or an object with template, name and/or partials properties.
+   * Passed model is in the following form: {record: record, uri: uri}
+   * example: "{{record.title}} - {{uri.description}}"
+   * @type {String|Object}
+   */
+  this.itemNameTemplate = undefined;
 
   /**
    * True to allow WMS resources to be added to the catalog; otherwise, false.
@@ -709,14 +720,15 @@ function createItemForUri(catalogGroup, record, uri, downloadUrls, legendUrl) {
   });
 
   if (defined(catalogItem)) {
+    catalogItem.name = getCatalogItemName(catalogGroup, record, uri);
+
+    catalogItem.url = uri.toString();
+
     if (catalogItem instanceof WebProcessingServiceCatalogGroup) {
       // only a few things we care about here
-      catalogItem.name = record.title;
-      catalogItem.url = uri.toString();
+      return catalogItem;
     } else {
-      catalogItem.name = record.title;
       catalogItem.description = record.description;
-      catalogItem.url = uri.toString();
       catalogItem.dataCustodian = "";
 
       if (defined(record.contributor)) {
@@ -763,6 +775,24 @@ function createItemForUri(catalogGroup, record, uri, downloadUrls, legendUrl) {
   }
 
   return catalogItem;
+}
+
+function getCatalogItemName(catalogGroup, record, uri) {
+  var templateName = catalogGroup.itemNameTemplate;
+
+  if (defined(templateName)) {
+    var datamodel = { record: record, uri: uri };
+    if (typeof templateName === "string") {
+      return Mustache.render(templateName, datamodel);
+    } else if (typeof templateName === "object") {
+      return Mustache.render(
+        templateName.template,
+        datamodel,
+        templateName.partials
+      );
+    }
+  }
+  return record.title;
 }
 
 function cleanUrl(url) {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 ### next release
 
 * Added `GeoRssCatalogItem` for displaying GeoRSS files comming from rss2 and atom feeds.
+* Added Mustache template Name to the CSWCatalogGroup (itemNameTemplate) to define a child item name
 
 ### v7.11.4
 
--- a/test/Models/CswCatalogGroupSpec.js
+++ b/test/Models/CswCatalogGroupSpec.js
@@ -163,4 +163,34 @@ describe("CswCatalogGroup", function() {
       .then(done)
       .otherwise(done.fail);
   });
+
+  it("honors itemNameTemplate", function(done) {
+    loadText("test/csw/ReferencesWithoutProtocol.xml")
+      .then(function(xml) {
+        jasmine.Ajax.install();
+        jasmine.Ajax.stubRequest(/.*/).andError();
+        jasmine.Ajax.stubRequest("http://gamone.whoi.edu/csw").andReturn({
+          contentType: "text/xml",
+          responseText: xml
+        });
+
+        var group = new CswCatalogGroup(terria);
+
+        group.updateFromJson({
+          name: "USGS Woods Hole pycsw",
+          type: "csw",
+          url: "http://gamone.whoi.edu/csw",
+          getRecordsTemplate:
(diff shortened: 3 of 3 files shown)


ccancellieri:
I've had to cherry-pick changes due to revert operations on the master branch

ccancellieri:
Pipeline fails due to https://github.com/TerriaJS/terriajs/issues/4237 not sure how to proceed, can you help me?

rowanwins:
Hi @ccancellieri 

Thanks for the contribution - I like the idea and did something similar a few months back for CKAN, see 
https://github.com/TerriaJS/terriajs/pull/3878

In terms of your PR
- I'm not sure we should be extending this behaviour all the way up to the `CatalogItem` model given this is a quite specific use case and no other catalog item implements it 
- Putting urls in the catalog item name seems like it would result in really long names... although it doesn't look like the CSW Metadata offers much else though that we could use which is a shame :(
  - I'm wondering if to counteract that we only apply the long name when we have to (eg when there are multiple WMS's available), this would be inline with my CKAN PR
- We also need to be a little careful with exposing mustache templates too liberally 

So I think my suggestion would be, rather than use mustache templates, would be to add an option to the `CswCataloGroup` along the lines of `useCombinedNameWhenMultipleReferencesExist`, which when set to `true` combines the name.

Does that make sense?

Cheers
Rowan

lib/Models/CswCatalogGroup.js
```diff
@@ -765,6 +768,27 @@ function createItemForUri(catalogGroup, record, uri, downloadUrls, legendUrl) {
...
+  if (defined(properties)) {
+    const templateName = properties.itemNameTemplate;
+
+    if (defined(templateName)) {
+      if (typeof templateName === "string") {
+        return Mustache.render(templateName, { record, uri });
+      } else if (typeof templateName === "object") {
+        return Mustache.render(
+          templateName.template,
+          { record, uri },
```
tephenavies: Unfortunately our documentation generation pipeline cannot deal with ES6/6+ syntax. Please replace both occurrences of { record, uri } with the old ES5 syntax: {record: record, uri: uri} (and then prettier will probably reformat that into multiple lines).

ccancellieri: Oh, I apologize, it wasn't clear to me, thanks for clarifying, I'll fix that

ccancellieri:
Thanks for your suport! I really appreciate it.

> Thanks for the contribution - I like the idea and did something similar a few months back for CKAN, see
> #3878

Oh, I didn't know that, this is also why I'm trying to move this functionnality down to the CatalogItem, but let me explain our POV on this task. (inline).

We discussed a lot internally about this problem with the metadata owners (which affected our CKAN and GeoNetwork items), let me explain:

Fetching the following resource from an ISO 19115 metadata using CSW

```

<?xml version="1.0" encoding="UTF-8"?>
<Record xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:dct="http://purl.org/dc/terms/" xmlns:geonet="http://www.fao.org/geonetwork" xmlns:ows="http://www.opengis.net/ows">
   <dc:identifier>9725ddc3-3824-41c1-b436-XXXXXXXX</dc:identifier>
   <dc:date>2020-04-20T16:04:39</dc:date>
   <dc:title>Global Roads Inventory Project (GRIP) dataset</dc:title>
   <dc:subject>roads</dc:subject>
   <dc:subject>GLOBIO</dc:subject>
   <dc:subject>transportation</dc:subject>
   <dc:subject>structure</dc:subject>
   <dct:abstract>The Global Roads Inventory Project (GRIP) .....</dct:abstract>
   <dc:rights>copyright</dc:rights>
   <dc:rights>license</dc:rights>
   <dc:language>eng</dc:language>
   <dc:source />
   <ows:BoundingBox crs="urn:ogc:def:crs:::WGS 1984">
      <ows:LowerCorner>180.00 -90.00</ows:LowerCorner>
      <ows:UpperCorner>-180.00 90.00</ows:UpperCorner>
   </ows:BoundingBox>
   <dc:URI protocol="WWW:LINK-1.0-http--link" name="Global patterns of current and future road infrastructure. Meijer et al (2018) Env. Res. Letters">https://iopscience.iop.org/article/10.1088/1748-9326/aabd42</dc:URI>
   <dc:URI protocol="OGC:WMS-1.3.0-http-get-map" name="globio:grip4_tp1_dens_m_km2" description="highways">https://XXXXXXXXXXXXXXX/map/gsrv/globio/wms</dc:URI>
   <dc:URI protocol="OGC:WMS-1.3.0-http-get-map" name="globio:grip4_tp2_dens_m_km2" description="primary roads">https://XXXXXXXXXXXXXXX/map/gsrv/globio/wms</dc:URI>
   <dc:URI protocol="OGC:WMS-1.3.0-http-get-map" name="globio:grip4_tp3_dens_m_km2" description="secondary roads">https://XXXXXXXXXXXXXXX/map/gsrv/globio/wms</dc:URI>
   <dc:URI protocol="OGC:WMS-1.3.0-http-get-map" name="globio:grip4_tp4_dens_m_km2" description="tertiary roads">https://XXXXXXXXXXXXXXX/map/gsrv/globio/wms</dc:URI>
   <dc:URI protocol="OGC:WMS-1.3.0-http-get-map" name="globio:grip4_tp5_dens_m_km2" description="local roads">https://XXXXXXXXXXXXXXX/map/gsrv/globio/wms</dc:URI>
   <dc:URI protocol="OGC:WMS-1.3.0-http-get-map" name="globio:grip4_total_dens_m_k2" description="total roads density">https://XXXXXXXXXXXXXXX/map/gsrv/globio/wms</dc:URI>
   <dc:URI protocol="WWW:LINK-1.0-http--link" name="Globio info">https://www.globio.info/download-grip-dataset</dc:URI>
   <dc:URI name="GRIP">\/srv/api/records/9725ddc3-3824-41c1-b436-XXXXXXXXXX/attachments/GRIP.PNG</dc:URI>
   <dc:URI name="Glosis">https://XXXXXXXXXXXXXXXx/srv/api/records/9725ddc3-3824-41c1-b436-XXXXXXXXXX/attachments/glosis.PNG</dc:URI>
   <dc:URI>https://XXXXXXXXXXXXXXXx/srv/api/records/9725ddc3-3824-41c1-b436-XXXXXXXXXX/attachments/glosic.PNG</dc:URI>
</Record>

```

Results into this (the resource name is different but the example is very similar, apologize for that):

image

As you see it's impossible to distinguish te layer we are going to add, and this is why we are here :)

My first action was trying to find a common field into the dc:URI which could be declared as layer name to concatenate the resource name to the layer name and this is what I've got:

TerriaWithoutMustache

This is fine but not so beautiful to see (we don't like WMS layername to be shown) nor so generic/standard, let's consider that CSW is a protocol which maps metadata standards (dubilncore, iso19115, etc) into another format.

[Here page 51](https://portal.opengeospatial.org/files/?artifact_id=77949) you see that the onlineresource is simply mapped to a generic anyURI element.

Due to this we could not assume that the resulting <dc:URI/> element have some specific or mandatory element to share (f.e.: name).

This is why I was trying to make it a generic method to 'compose' the title of an item using what the item creator knows to have.
In my example we have a quite short description so I used the following template:

```
...
"itemProperties": {
            "itemNameTemplate":"{{record.title}} - {{uri.description}}"
 }
...
```
Which results like this:
TerriaWithMustache

That said let me try to explain the proposal

> * I'm not sure we should be extending this behaviour all the way up to the `CatalogItem` model given this is a quite specific use case and no other catalog item implements it

Well since that functionnality is quite generic I'd like to port that to all the connectors (at least the CKAN one which have a similar problem, not considering your pullreq which was unknown at that time).

Probably I'm too ambitious :)

If so where do you suggest to move this field?

> * Putting urls in the catalog item name seems like it would result in really long names... although it doesn't look like the CSW Metadata offers much else though that we could use which is a shame :(

As depicted in the images it's up to the metadata owner AND the catalog editor define what should be used for title name AND in which form, in my case I'm showing the description, while in the jasmine tests I'm showing the protocol, etc.

>   
>   * I'm wondering if to counteract that we only apply the long name when we have to (eg when there are multiple WMS's available), this would be inline with my CKAN PR

Following the discussion above I can't see any field (apart of the url itself) which can be used as 'long name' because we could not guarantee that the resource will ship into the <dc:URI/>. In my example you can see:
 - description
 - protocol
 - name

But none of that are mandatory (I know that we are filtering on protocol but it's not so useful in long names, I used that only for tests to avoid adding new resources)

> * We also need to be a little careful with exposing mustache templates too liberally

I agree with you and I see your concerns but I'm using that because the application is already exposing it for the getfeatureinfo. I've also checked in the past it has some problem of csrf vulnerability but most of those are addressed right now. 

> 
> So I think my suggestion would be, rather than use mustache templates, would be to add an option to the `CswCataloGroup` along the lines of `useCombinedNameWhenMultipleReferencesExist`, which when set to `true` combines the name.
> 
> Does that make sense?

Not sure if our point has been shown clearly, in our case showing the layername in addition to the resource is not so acctractive (the same is happening with the CKAN connector, I've planned to discuss that in a second stage but let me extend the problem to that here due to a such detailed description).

Let me say that I appreciated your feedback, I'm just bringhing our previous considerations trying to motivate the implementation.
I tryied to discuss about that on the issue or on gitter but it's really urgent to us so I tryied to speedup the process creating the proposal.

If you really don't like the template functionnality, is there a way to expose the URI description instead of the name? (it could be not so standard but we need that even in the CKAN connector).

If you instead see the point to use the template, I see only a problem, every connector may have it's own (probably different) data model to pass to the template and this is something which should be very well documented on the Item or on the connector (I could do that if needed).

rowanwins:
Thanks for the detailed writeup @ccancellieri and sharing your scenario

The other option I can think of might be to create a group for datasets which have multiple references. The group could have  the title of the dataset and then the sub items could use the uri description as the name. eg
````
  - Global Road Inventory Dataset
    - highways
    - local roads 
    - ...
````
That sort of approach would prevent us from having to duplicate any element of the name which might be nice.
If we did that we could perhaps pass down a field to use (eg "name" or "description"). 

In regards to making this sort of thing more generic than just CSW it seems to me that it's most likely a problem unique to subclasses of  `CatalogGroup` rather than `CatalogItem`.

Sorry for any delays - many of our staff are working on a raft of projects which pay our way so while we do our best to keep up with the open source community sometimes it's a challenge :)

Cheers

lib/Models/CswCatalogGroup.js
ccancellieri: Replaced that and also changed const to var

ccancellieri:
Dear All,
 changing the metadata in our case is not the best option, most of our data have been harvested from other sources and we also have to manage about 9000 metadata.

I'm moving the itemNameTemplate field down to the CSWCatalogGroup to minimize impact of this pull request on other connectors.

Let's see if it is acceptable for you and if we can survive limiting that only to CSW only for now.

ccancellieri:
Thank you for your support, I hope it helps,
Best regards,
Carlo

PR #4235, opened 2020-04-21
Mobx: support for date formatting in UI

rowanwins:
### Adds support for formatting of dates in various spots in UI

This PR adds a `dateFormat` trait to the `TimeVaryingTraits` which formats a js date using the [dateformat](https://www.npmjs.com/package/dateformat#usage) library. This sort of functionality was available in master but I've tweaked it slightly. 
The spots it gets applied are 
- [x] DateTimeSelectorSection.jsx
- [x] BottomDock/Timeline/Timeline.jsx
- [x] BottomDock/Timeline/DateTimePicker.jsx

To test use a catalog item like
````
    {
      "url": "https://ows.services.dea.ga.gov.au",
      "type": "wms",
      "name": "time WMS",
      "layers": "ls8_nbart_geomedian_annual",
      "dateFormat": "yyyy" <---- the important bit
    },
````

One thing I haven't tackled is FeatureInfo - My gut feel is that presentation of dates for feature info should be handled separately but happy to discuss.

### Checklist

-   [x] There are unit tests to verify my changes are correct. 


na9da:
This is a nice change! 

:+1: for the view tests.

As for the feature info date formats, I think we should re-implement the mechanism in master for formatting feature info data [seen here](https://github.com/TerriaJS/terriajs/blob/master/doc/connecting-to-data/customizing-data-appearance/feature-info-template.md#formatting-dates).

na9da (approved):
Looks good to me!

PR #4244, opened 2020-04-22
Stop automatically including all "@ types" packages in compilation

tephenavies:
Fixes `tsc --noEmit` (which was broken due to pulling in `@types/react-native` from styled-components and browser library types)

Also paves the way for https://github.com/TerriaJS/TerriaMap/issues/437 by fixing many/all of the errors that show up there.

See https://www.typescriptlang.org/v2/tsconfig#types


soyarsauce:
Thank you i was wondering why the heck react native types were showing up when you were talking about your type errors

PR #4240, opened 2020-04-22
Bump mobx minor, mobx-react major versions

soyarsauce:
### What this PR does

Addresses requirement for some new syntax/approaches to https://github.com/TerriaJS/terriajs/issues/4183 and further UI development, by updating mobx versions - mainly hooks 

### Checklist

-   [x] unit tests still pass locally
-   [x] I've updated CHANGES.md with what I changed.

PR #4248, opened 2020-04-23
  Fixes FeatureInfoPanel port from #4240 & some wrong proptypes

soyarsauce:
### What this PR does

Fixes FeatureInfoPanel port from https://github.com/TerriaJS/terriajs/pull/4240 & some wrong proptypes

### Checklist

-   [x] no unit tests for port/type update

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -107,6 +107,8 @@ Change Log
 * Tidy Compass UI animations, styles, titles
 * Bumped mobx minor to 4.15.x, mobx-react major to 6.x.x
 * Add `dateFormat` trait to `TimeVaryingTraits` to allowing formatting of datestrings in workbench and bottomdock.
+* Fixed FeatureInfoPanel using old class state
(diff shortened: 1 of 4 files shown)


soyarsauce:
Yeah definitely don't merge this until I fix feature info panel

soyarsauce:
merged mobx now so should be good to go

na9da (approved):
Looks good!

There is some code here that uses the [old prop for FeedbackButton](https://github.com/TerriaJS/terriajs/blob/mobx-update-versions/lib/ReactViews/StandardUserInterface/MapColumn.jsx#L138), but it is commented out, so not a problem.

soyarsauce:
Yeah we'll probably slowly clean up those commented out bits when we get to removing sass (as a requirement for https://github.com/TerriaJS/terriajs/blob/mobx/doc/contributing/styled-components-and-sass.md)

PR #4257, opened 2020-04-27
Mobx difftool (delta v2)

### LINKED ISSUE #4145: MobX - Change detection (Delta) - v2 development
- [x] UX prepares designs of v2 until 20 March 2020

- [x] development of v2 in MobX to be demonstrated in TerriaCube mobx version - to occur between 20 March - 20 April (max)

na9da:
### What this PR does

Fixes #4145 

- [x] Implement the flow as per design
- [x] Remove time controls & location filter for the generated diff item added to workbench
- [x] Lazy load the difftool
- [ ] Try and optimize WMS performance when processing large lists of available dates
- [ ] Styling (rough list)
   - [ ] Title bar for diff tool main panel
   - [ ] Close button on the right
   - [ ] Location picker message style & position
   - [ ] Various selectors in the `MainPanel`
   - [ ] DiffTool icon in the `ViewingControls` menu

Sample item for testing
```
{
      "name": "Delta Blended service",
      "type": "wms",
      "opacity": 1,
      "timeFilterPropertyName": "data_available_for_dates",
      "canDiffImages": true
}
```

### Checklist


lib/ModelMixins/TimeFilterMixin.ts
```diff
@@ -1,37 +1,37 @@
...
 type MixinModel = Model<
   TimeFilterTraits & DiscretelyTimeVaryingTraits & MappableTraits
 > &
-  TimeVarying;
+  DiscretelyTimeVaryingMixin.Instance;
 
 /**
  * A Mixin for filtering the dates for which imagery is available at a location
- * picked by the user.
+ * picked by the user. This Mixin requires `DiscretelyTimeVaryingMixin` to be present.
```
tephenavies: Rather than one Mixin requiring another to be present I think it is better to have TimeFilterMixin mix in DiscretelyTimeVaryingMixin and itself when called. Then TimeFilterTraits would also include DiscretelyTimeVaryingTraits. Unless that isn't technically possible or poses problems somewhere.

lib/Models/WebMapServiceCatalogItem.ts
```diff
@@ -549,9 +605,9 @@ class WebMapServiceCatalogItem
 
   @computed
   private get _currentImageryParts(): ImageryParts | undefined {
-    const imageryProvider = this._createImageryProvider(
-      this.currentDiscreteTimeTag
-    );
+    const imageryProvider = this._createImageryProvider({
+      time: this.currentDiscreteTimeTag
```
tephenavies: I think this change might cause layers to flicker and be re-created when the opacity slider is moved.

soyarsauce:
test for mobx: http://ci.terria.io/mobx-difftool/#https://gist.githubusercontent.com/soyarsauce/427583294a53dfbd13e35d820ac66467/raw/688a57e9d5f184c2b6ddb32b892493a4c567aa6d/delta-mobx.json

soyarsauce:
@na9da is the `leftItem` supposed to be the newly returned layer after requesting the diff? For some reason I'm getting blank `availableDiffStyles` from `{this.diffItem.availableDiffStyles?.map(style => ` on `13d1e23`
image

soyarsauce:
~~Additionally, I think there's an unintended bug when you tried to hide the dates via~~
~~https://github.com/TerriaJS/terriajs/pull/4257/commits/4dedacf9085014e1699fdb419612c66850be0c87~~

~~Now delta doesn't show the dates at the bottom when you load the difftool~~

image
edit: it's just css, the text is dark on dark bg

soyarsauce:
hidden datepicker colours resolved with https://github.com/TerriaJS/terriajs/pull/4300/commits/17d46667c326ca9d087aa0080272fc46ccf38013

But the difference image problem still persists, only on your new changes to `mobx-difftool` cc @na9da

na9da:
> @na9da is the `leftItem` supposed to be the newly returned layer after requesting the diff? For some reason I'm getting blank `availableDiffStyles` from `{this.diffItem.availableDiffStyles?.map(style => ` on `13d1e23`

@soyarsauce I think I know why. I recently made a change to hide the style selector for the generated diff item when shown on the workbench. It works by returning an empty styleSelector  if the item is currently showing difference image. That has an unintended effect of making the availableStyles empty. I'll figure out a fix for that.

na9da:
@soyarsauce the empty diff styles bug should be fixed now.

lib/Models/WebMapServiceCatalogItem.ts
na9da: I have reverted `_createImageryProvider` to the old style accepting only time param. Instead i'm picking extra params inside `_createImageryProvider` if it is showing diff.

lib/ReactViews/SplitPoint.jsx
```diff
@@ -20,4 +20,5 @@ const SplitPoint = createReactClass({
   }
 });
 
+export default SplitPoint;
 module.exports = SplitPoint;
```
tephenavies: It's probably best to remove the old export

soyarsauce:
Some issues for post-delta/testing

- removing the delta source item from workbench keeps the diff tool open, should probably close it if it isn't referenced in workbench anymore
- bottom date pickers cover up the timeline + logo area

tephenavies (approved):

PR #4260, opened 2020-04-29
Mobx tour

soyarsauce:
### What this PR does

Resolves https://github.com/TerriaJS/terriajs/issues/4183

Missing some basic unit tests and docs but you can't readily customise it anyway without rebuilding terria - will address separately.

### Checklist


tephenavies (commented):
Couldn't get this to work in a local build. On http://ci.terria.io/mobx-tour/ it worked really well. There should be a "previous" button on the final page of the tour. Otherwise everything worked well.

`lib/ReactViewModels/GuideSequence.ts` is an empty file.

soyarsauce:
@steve9164 Which part wasn't building when you were trying to run it locally?

soyarsauce:
Good pickup on the "previous" needing to appear on the last slide, it was explicitly removed from last slide as it used to be a "skip" button.

I've also now updated it so that all the dot and overlays follow the correct components upon a window resize (triggered via both resize or zoom)

PR #4270, opened 2020-05-01
Lock moment to 2.24.0 until 2.25 issue resolution

soyarsauce:
### What this PR does

Fixes CI

Given CI will nuke TerriaMap's package lock, terriajs auto upgrades dependencies depending on package json defined versions. (This will never change as package-lock.json from TerriaMap will never sync up with a given branch)


soyarsauce:
Somehow this has made things worse. But can't reproduce locally

PR #4275, opened 2020-05-04
Force moment 2.24.0 for CI builds

soyarsauce:
This is the real real fix for https://github.com/TerriaJS/terriajs/pull/4270. The nuclear option from https://github.com/TerriaJS/terriajs/pull/4270 wasn't nuclear enough


soyarsauce:
Adding that we will need to remove this later once moment issues are resolved, having versions specified like this outside of package.json are always only a bandage fix. 2.25.0 & 2.25.1 both still exhibit issues when building, so we are sticking to 2.24.0

PR #4287, opened 2020-05-07
Rework handling of proxy on url for Cesium3dTilesCatalogItem

rowanwins:
### What this PR does

During deploying `mobx-28` to https://dev.nsw.digitaltwin.terria.io/ we found problem viewing `Cesium3DTilesCatalogItem` where things were getting confused between a url string and a resource,
eg https://dev.nsw.digitaltwin.terria.io/#share=s-zJGCLt3Wi4n8ySadVILll74YvtW

 I think the bug was introduced in #4140 due to a subtle change in where `proxyCatalogItem` was applied.

### Checklist

    - Not super sure how I'd write tests for this... Not sure how specific it was/is to the NSW DT deployment where there is additional complexity around our 3D tiles proxying.

### DIFF
--- a/lib/Models/Cesium3DTilesCatalogItem.ts
+++ b/lib/Models/Cesium3DTilesCatalogItem.ts
@@ -93,15 +93,37 @@ export default class Cesium3DTilesCatalogItem
   }
 
   private loadTileset() {
-    const tileset = this.createNewTileset(
-      this.url,
-      this.ionAssetId,
-      this.ionAccessToken,
-      this.ionServer,
-      this.optionsObj
-    );
+    if (!isDefined(this.url) && !isDefined(this.ionAssetId)) {
+      return;
+    }
 
-    if (tileset && !tileset.destroyed) {
+    let resource = undefined;
+    if (!isDefined(this.url) && isDefined(this.ionAssetId)) {
+      resource = this.createResourceFromIonId(
+        this.ionAssetId,
+        this.ionAccessToken,
+        this.ionServer
+      );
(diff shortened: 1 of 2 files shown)


lib/Models/Cesium3DTilesCatalogItem.ts
```diff
@@ -93,15 +93,37 @@ export default class Cesium3DTilesCatalogItem
...
+        this.ionAssetId,
+        this.ionAccessToken,
+        this.ionServer
+      );
+    } else if (isDefined(this.url)) {
+      resource = this.createResourceFromUrl(
+        proxyCatalogItemUrl(this, this.url),
+        this.ionAssetId,
+        this.ionAccessToken,
+        this.ionServer
```
nf-s: can these be removed?

lib/Models/Cesium3DTilesCatalogItem.ts
```diff
@@ -149,47 +171,44 @@ export default class Cesium3DTilesCatalogItem
...
 
-  private createNewTileset(
-    url: Resource | string | undefined,
+  private createResourceFromUrl(
+    url: Resource | string,
     ionAssetId: number | undefined,
     ionAccessToken: string | undefined,
-    ionServer: string | undefined,
-    options: any
+    ionServer: string | undefined
```
nf-s: can these also be removed?

nf-s (changes requested):

nf-s (approved):
I have merged `mobx` into this to resolve conflicts with `CHANGES.md`

PR #4302, opened 2020-05-13
Enable the item only if it is currently disabled.

na9da:
This fixes an issue where the item was getting added to the workbench twice.

### What this PR does

Fixes a problem where the catalog item gets added twice to the workbench if it is already enabled.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 ### next release
 
 * Added `GeoRssCatalogItem` for displaying GeoRSS files comming from rss2 and atom feeds.
+* Fixed a bug where catalog item returned by TerriaJsonCatalogFunction gets added to the workbench twice.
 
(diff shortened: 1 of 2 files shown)


soyarsauce (changes requested):
This kind of makes sense, but it feels like it doesn't address the root problem of the side effect & goes against the thing where you can set something twice and it should JustDoIt.
https://github.com/TerriaJS/terriajs/blob/mobx/doc/contributing/model-layer.md#avoid-reactions-and-other-side-effects

soyarsauce:
Actually, seeing it in context of `isShownChanged()` it probably makes sense to do it like this. Maybe someone else could also sanity check?

na9da:
Closing. Since the root problem was a mis-configured catalog file that set `isEnabled: "true"` instead of `isEnabled: true`.

PR #4310, opened 2020-05-14
Re-mount WelcomeMessage against disclaimerVisible

soyarsauce:
### What this PR does
Fixes the clash of welcome message against the disclaimer
Now it waits for disclaimer first if it's visible, before
doing its checks against localstorage for preferences/past visits
### Checklist

-   [x] n/a There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)


na9da (approved):
Looks good!

PR #4312, opened 2020-05-14
Mobx diff ui tweaks

soyarsauce:
### What this PR does

addresses some issues in https://github.com/TerriaJS/terriajs/issues/4311

Fix workbench viewing control styling
Updates difference/delta icon, compare/splitter icon and spacing
Add stopgap fix for splitter being offset
Don't show difference control on diffed items

### Checklist

-   [x] n/a There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [x] n/a I've updated CHANGES.md with what I changed.


soyarsauce:
>Nanda Sankaran  2:09 PM
Looks good to me.


Wing Ho  2:16 PM
well, 3 resize events per 'workbench toggle' -i want to leave the third one in, incase a browser doesn't fire onTransitionEnd for whatever reason
image.png 
image.png

2:17
part of me is like, well. mobx is already so slow, better slow and working than buggy-splitter
2:17
alright that's the basis for why i'm just gonna SlapItOn
2:17
slipslapslop

PR #4315, opened 2020-05-15
Re-assign split directions when the tool is reset.

### LINKED ISSUE #4311: Post testing delta fixes
Some issues for post-delta/testing

- [ ]  removing the delta source item from workbench keeps the diff tool open, should probably close it if it isn't referenced in workbench anymore
- [ ]  bottom date pickers cover up the timeline + logo area
- [x] The output diff item in workbench should hide the menu option to invoke Difference tool again
- [x] #4309 
- [x] Left/right items loose the split direction sense when the tool is reset
- [ ] The output item should show legend

na9da:
### What this PR does

Partly fixes #4311

Fixes a difftool bug where left/right items loose their split direction settings when the tool is reset.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Change Log
 * Bumped explicit base typescript to 3.9.2
 * Lock rollbar to 2.15.2
 * Ported disclaimer to mobx with new designs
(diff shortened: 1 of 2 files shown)


soyarsauce:
Delete branches after merge 😀

PR #4318, opened 2020-05-18
Fixes a splitter bug where split direction is not applied to new layers

na9da:
### What this PR does

Fixes a splitter bug where split direction is not applied to new layers

This behavior can be consistently triggered after pressing the Back button to reset the difftool. It is happening because of the order of execution of the splitter reaction & observe model layer reaction in Cesium/Leaflet models. The splitter reaction sets the split direction only for layers currently on the map. So if the splitter reaction runs before the observe model layer reaction, new layers that are not yet added to the map will not get the split direction applied to it. This fix applies the split direction to all imagery layers, not checking whether they are on the map and so removes the dependency on the order in which the reactions are run.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ Change Log
 * Added satellite imagery to help panel
 * Stop disclaimer clashing with welcome message by only loading WelcomeMessage after disclaimer is no longer visible
 * Fixes a difftool bug where left/right items loose their split direction settings when the tool is reset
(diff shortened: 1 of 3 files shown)


tephenavies:
The new implementation does seem more correct. The only problem is that alpha and show won't be set correctly yet for some of the returned imagery. Maybe it would be better to use `_makeImageryLayerFromParts` instead of `_createImageryLayer`.

tephenavies (approved):
Thanks for making the change.

Issue #4280, opened 2020-05-05
[EPIC] MobX UI i18n blockers


soyarsauce:
We have a few things that we are adding to the mobx ui that will affect i18n-ability, namely:
### Map icon button
PR:
https://github.com/TerriaJS/terriajs/pull/4165#discussion_r401348748

Fix:
https://github.com/TerriaJS/terriajs/issues/4281

### Help menu items/panels
https://github.com/TerriaJS/terriajs/issues/4161
https://github.com/TerriaJS/terriajs/issues/4278

Unfortunately we need to ship these, but we also want to make things like this work with the new i18n translation files - and don't want to lose sight of them.

Also, we don't actually have a mechanism right now to swap languages in the UI itself, a TerriaMap consumer would need to do the overrides by forking terriajs which is an extremely unideal situation (might as well call it unsupported right now, even though all the framework is in place to i18n)

zoran995:
There is some initial work on changes of i18next configuration started in #4227. 
To avoid users having to override translations by forking terriajs and updating directly it is necessary to support loading of translation from the backend (second point in draft PR), and for that, it is necessary to pass translation files during build so they are accessible. My experiments with this were to manually copy of `Language` folder to `TerriaMap/wwwroot` and it worked. This way user can edit translation file directly and changes will be visible the same way as for configuration files. It would be much better to keep translation files as default values and providing users with a possibility to override those value from TerriaMap, but have to look if this can be done and how. This way user would end with much smaller `.json` file to edit on his end and also easier upgrade between versions. 

Without this, I don't think it will be possible to i18n-ify help menu items. I have some ideas for it but have to do some testing.

I would like to hear other opinions on this, and if @vicborgy have/make a position for swap languages mechanism in his UI designs I would be happy to finish it.

PR #4324, opened 2020-05-18
load translation override using backend

zoran995:
### What this PR does

I decided to break commit #4227 into smaller parts since something like `eslint` will take a long time to completely configure. The basic idea is described in #4280 ([comment](https://github.com/TerriaJS/terriajs/issues/4280#issuecomment-625207504_)) 
Adds an option for overriding the translation files from TerriaMap. This provides the user with the ability to 
- define language configuration inside TerriaMap `config` file. 
- put a language overrides to file `wwwroot/Language/{{lng}}/languageOverrides.json`, `{{lng}}` is a short name for the language. If the language that user is looking for is not officially supported by Terria, he should specify also create a `translation.json` to avoid `404 not found` error in console. Reason for this is in i18next backend default behaviour to try to load every namespace that is not already loaded. 
- tests are updated and should be correct (at least they are on my computer) to support the removing of `i18.js` from `prerequisites.js`. Tests now use `cimode` language option which returns the translation key, it is defined to run `beforeEach` test so it is not necessary to call it for all tests.

global internationalization track in #3785
### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - I am not sure how to test this


zoran995:
One idea is to use this as an option to enable internationalization of the help and trainer content (if it proves to be possible to handle all use cases hope to have it by end of the week)

soyarsauce:
This should go some of the way to addressing https://github.com/TerriaJS/nsw-digital-twin/issues/425

zoran995:
implemented in #4523

PR #4326, opened 2020-05-18
Mobx: Ensure there aren't more bins than unique values

### LINKED ISSUE #4325: Mobx: Sometimes scalar legends on table data have more bins than values
If the number of unique values in a dataset is very low and there in no `numberOfBins` set in `TableColorStyleTraits` then you can end up with some funky presentation, eg

ScalarBins

rowanwins:
### What this PR does

Fixes #4325

This PR ensures there aren't more bins than there are unique values.

### Checklist


nf-s:
Poor `@computed` getting linked into everything...
I am pretty happy with this, but I think we could make improvements to the colour ramp - for example
image
This has picked two colours which are quite similar, whereas it should have used a light yellow, and deep red.
Maybe we could use https://github.com/d3/d3-scale-chromatic to generate colour scales based on number of bins

nf-s (changes requested):
See https://github.com/TerriaJS/terriajs/pull/4326#issuecomment-673240839

nf-s:
Or, we could start using this approach instead https://github.com/TerriaJS/terriajs/pull/3667

soyarsauce:
Related https://github.com/TerriaJS/terriajs/issues/4263

nf-s:
We have decided to leave continuous colour ramps for another day, I will clean this up and merge it as a quick fix

nf-s (approved):

PR #4329, opened 2020-05-19
Automatically pick searched location to filter imagery

na9da:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4320

Modifies difftool so that it will automatically pick a marked position on the map (like from a search result) and filter imagery for that location.

We get the latlng from the search result, but to make the feature request we need to know the tile x, y coordinates and level. We can get x,y from latlng & level using [positionToTileXY](https://cesium.com/docs/cesiumjs-ref-doc/TilingScheme.html#positionToTileXY) api. For `level`, we simply have to use a high enough value supported by the service (some services don't work on low zoom levels). I have hardcoded it to 30 for now.

### Checklist


lib/ReactViews/Tools/DiffTool/DiffTool.tsx
```diff
@@ -362,6 +355,35 @@ class Main extends React.Component<MainPropsType> {
...
+    // imagery at that location
+    const markerLocation = getMarkerLocation(this.props.terria);
+    const sourceItem = this.props.sourceItem;
+    if (
+      this.location === undefined &&
+      markerLocation &&
+      Mappable.is(sourceItem)
+    ) {
+      const part = sourceItem.mapItems.find(p => ImageryParts.is(p));
+      const im = part && ImageryParts.is(part) && part.imageryProvider;
```
soyarsauce: Just realised this now as it's called `im` in various parts of difftool.tsx but it can be confusing for an `ImageryProvider` to be abbreviated to `im`

soyarsauce (changes requested):
Hey @na9da - a few things
- Can you find a way to add a test for this?
- This doesn't seem to filter the location for the 'right side' difference image
- This still tries to auto select your location even if you clear your search marker, e.g.
1. search for a location (e.g. "queensland")
2. clear that result
3. enable difference mode on a delta item
4. observe it still auto-selects your old searched location

lib/ReactViews/Tools/DiffTool/DiffTool.tsx
na9da: Yes! I did it for a cleaner perttier formatting, but it is not worth it :)

na9da:
> Hey @na9da - a few things
> 
>     * Can you find a way to add a test for this?

I gave it a shot but it wasn't rendering properly because of the `reaction` in DiffTool. I think I'll write some tests for this and some other parts of difftool when I'm come back later to implement the changes from user-testing.

> 
>     * This doesn't seem to filter the location for the 'right side' difference image
I couldn't reproduce that. It worked for me.

>     * This still tries to auto select your location even if you clear your search marker, e.g.
Thanks for spotting that. I have fixed it :+1: 

>     1. search for a location (e.g. "queensland")
> 
>     2. clear that result
> 
>     3. enable difference mode on a delta item
> 
>     4. observe it still auto-selects your old searched location

soyarsauce:
https://github.com/TerriaJS/terriajs/issues/4380

soyarsauce:
@na9da to merge next/mobx into this branch, then tag @soyarsauce  - after this is in next, a small tweak will be required to fulfill #4380

soyarsauce:
~@na9da please link the ticket this PR addresses, thanks~
https://github.com/TerriaJS/terriajs/issues/4320

CHANGES.md
```diff
@@ -3,8 +3,14 @@ Change Log
 
 ### MobX Development
 
+#### mobx-next-release (mobx-31)
```
soyarsauce: Bad merge conflict

soyarsauce (approved):
Works well with the exception that there's no UI indicator that the searched location is being loaded in.

lib/ModelMixins/AsyncChartableMixin.ts
```diff
@@ -52,3 +52,13 @@ export default function AsyncChartableMixin<
 
   return AsyncChartableMixin;
 }
+
+namespace AsyncChartableMixin {
+  export interface AsyncChartableMixin
+    extends InstanceType<ReturnType<typeof AsyncChartableMixin>> {}
+  export function isMixedInto(model: any): model is AsyncChartableMixin {
+    return model && model.isChartable;
```
tephenavies, line 60: I don't think this is a great check for AsyncChartableMixin.
The fact that `Chartable.is(model)` and `model.isChartable()` are not about the same thing is odd. I'm not sure `AsyncChartableMixin.isChartable` should exist. It might have been added to match UI existing logic, but we should prefer `Chartable.is(model)`.

I have no better suggestion though.

And I've just realised this comes straight from AsyncMappableMixin, where it was introduced under a `// TODO`: 
https://github.com/TerriaJS/terriajs/commit/5ea77ed4233a6a770f5e4352ee35edd69e8be1a6#diff-3ad2723f358849722e25021da0c6f044R22-R26

lib/ModelMixins/AsyncMappableMixin.ts
```diff
@@ -49,3 +49,13 @@ export default function AsyncMappableMixin<
 
   return AsyncMappableMixin;
 }
+
+namespace AsyncMappableMixin {
+  export interface AsyncMappableMixin
+    extends InstanceType<ReturnType<typeof AsyncMappableMixin>> {}
+  export function isMixedInto(model: any): model is AsyncMappableMixin {
+    return model && model.isMappable;
```
tephenavies, line 57: Ditto

lib/Models/LocationMarkerUtils.ts
```diff
@@ -91,7 +93,24 @@ export function removeMarker(terria: Terria) {
...
   return catalogItem !== undefined && terria.overlays.contains(catalogItem);
 }
+
+export function getMarkerLocation(terria: Terria): LatLonHeight | undefined {
+  const catalogItem = terria.getModelById(CzmlCatalogItem, MARKER_UNIQUE_ID);
+  if (catalogItem === undefined || !terria.overlays.contains(catalogItem)) {
+    return;
+  }
+  const marker = catalogItem.czmlData[1];
+  // @ts-ignore
```
tephenavies, line 107: I'm not a fan of `// @ts-ignore`. It can be broken by changes in indenting because prettier could reflow lines, either ignoring types for more or less code than intended and it's less clear exactly where it's needed than casting `((o as unknown) as MyType)` or using `<any>o`

tephenavies:
I'll add my comments to a new issue

soyarsauce:
> I'll add my comments to a new issue

@steve9164 do it in https://github.com/TerriaJS/terriajs/pull/4515, this was a functional-merge

soyarsauce:
~functional-merge~

meaning merged so that the separate difference features could be incorporated into flow testing

lib/ModelMixins/AsyncMappableMixin.ts
tephenavies: Recorded in https://github.com/TerriaJS/terriajs/issues/4522

lib/ModelMixins/AsyncChartableMixin.ts
tephenavies: Recorded in https://github.com/TerriaJS/terriajs/issues/4522

lib/Models/LocationMarkerUtils.ts
tephenavies: Moved to #4515.

lib/ModelMixins/AsyncChartableMixin.ts
na9da: The change is from another PR and is being addressed in this issue - https://github.com/TerriaJS/terriajs/issues/4522

lib/ModelMixins/AsyncMappableMixin.ts
na9da: The change is from another PR and is being addressed in this issue - #4522

PR #4341, opened 2020-05-25
Convert Clipboard to use styled-components and TS

soyarsauce:
### What this PR does

Update clipboard to use styled-components and Typescript. Forms part of styles rewrite for client-side theming.

I've also changed the exports of Icon.jsx to be more typescript friendly.

### Checklist

-   [x] Test!
-   [x] Only a refactor, no CHANGES.md entry needed


soyarsauce:
@steve9164 please update desc

lib/ReactViews/Icon.jsx
```diff
@@ -141,13 +141,13 @@ export const StyledIcon = styled(Icon)`
...
 
   ${props => props.opacity && `opacity: ${props.opacity};`}
 `;
 
-export default Icon;
-// (?) are these cjs exports for the doc generator?
-module.exports = Icon;
-module.exports.GLYPHS = GLYPHS;
-module.exports.StyledIcon = StyledIcon;
+export default Object.assign(Icon, { GLYPHS });
```
tephenavies, line 155: Export things in only 1 way (using modules). This should fix the mismatch between Typescript type-checking and webpack imports. Ideally sometime we'd export GLYPHS separately to Icon, but it's used so widely that that is best addressed in a separate PR

soyarsauce:
The padding is ever so slightly different to mobx but at least it is now using SC

tephenavies (approved):

PR #4348, opened 2020-05-26
Mobx - catalog group fix

### LINKED ISSUE #4346: Mobx: handle invalid items when initiating the catalog
When a catalog includes invalid configuration for a single item subsequent items won't be loaded

````
catalog: [
    {
      "type": "geojson",
      "name": "GeoJSON Test",
      "url": null
    },
    {
      "type": "geojson",
      "name": "GeoJSON Another Test",
      "url": "test/bike_racks.geojson"
    }
]
````
The error is handled, but we should keep processing the other items.
Error

### PR DESCRIPTION
**Pull request by rowanwins**
### What this PR does

Fixes #4346

When loading a catalog file if a member of a catalog group is incorrectly defined **all** members of that group will fail to display in the catalog. For example the following catalog
````
catalog: [
    {
      "type": "geojson",
      "name": "GeoJSON Test",
      "url": null
    },
    {
      "type": "geojson",
      "name": "GeoJSON Another Test",
      "url": "test/bike_racks.geojson"
    }
]
````
Order of the catalog items is irrelevant, all items will fail to display in the catalog.

This PR resolves it. ~~Not super sure on the best approach for tests yet but it would be a useful addition. Not sure if I'm testing the changes to the  `fromJson` method in the `modelReferenceArrayTrait`  or the `GroupMixin` where the error manifested itself...~~

### Checklist

### CODE REVIEW

**Review by soyarsauce — COMMENTED** [review]
awesome nice work





5:01
test would just be trying to call fromJSON with some invalid models
5:01
fromJson*
5:01
and with some valid models
5:02
and check that they correctly resolve the ones that still work
5:02
something along those lines

**Comment by soyarsauce** [comment]
You can do it in either or both, but makes a tiny bit more sense from GroupMixin where it gets consumed

**Comment by soyarsauce** [comment]
Just give me a few ticks to test this with catalog editor, swallowing the error whole might have adverse effects with newly created incorrectly-defined-records being lost in the abyss

**Comment by rowanwins** [comment]
Yeah I was trying to think if there was a way to still create it and leave it as a stub member in the group with a note about incorrect configuration or something like that...

**Comment by soyarsauce** [comment]
Time for a StubCatalogItem!

**Comment by soyarsauce** [comment]
This still doesn't work with a reference (meaning it will not fix what you are trying to fix, for the reference catalog items)

**Comment by rowanwins** [comment]
Ok so the approach now makes use of the `isExperiencingIssues` trait on a `CatalogMember` so dodgy items will remain in the catalog but will be flagged.

File: lib/Models/MagdaReference.ts [hunk 1]
```diff
@@ -514,7 +514,11 @@ export default class MagdaReference extends AccessControlMixin(
       if (stratum === "type" || stratum === "id") {
         return;
       }
-      updateModelFromJson(result, stratum, terriaAspect[stratum], true);
+      try {
+        updateModelFromJson(result, stratum, terriaAspect[stratum], true);
+      } catch (err) {
+        result.setTrait('override', 'isExperiencingIssues', true)
```
soyarsauce: >Yeah I think underride is better than definition


so do you want under or over here?

rowanwins: i was a ding dong

soyarsauce:
Tests are still failing @rowanwins ☹️

soyarsauce:
Hi @rowanwins - bad news, I managed to break this branch again, good news - I added a fix for it this time. Please review the changes across https://github.com/TerriaJS/terriajs/pull/4348/commits/5bae1814d05a6473e0f4bd8743ceea4e12a42a4b && https://github.com/TerriaJS/terriajs/pull/4348/commits/c5f5c42900a57bc4b4bd00b84f83a82c38e16360 then :shipit:

soyarsauce (approved):

PR #4352, opened 2020-05-26
Mobx fix timefilter

na9da:
### What this PR does

Fixes a bug which causes time filter to show for all satellite imagery layers

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Change Log
 ### MobX Development
 
 #### mobx-next-release (mobx-32)
-* Fixing a regression bug were the time filter is shown for all satellite imagery items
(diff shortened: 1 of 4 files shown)


soyarsauce (approved):
Awesome work Nanda
tl;dr slack discussion
- why not wms yet? 
>The method in question is defined by the Mixins and not directly by WMS, so I thought I should test the mixins. Could have saved a few lines. I'll refactor it when I  add more tests for the Mixin.
- add extra test assupmtion for wms in a separate pr
- >the reason for all this, is i suspect adding the extra WMS test will also highlight the runtime error we have at mobx where one mixin tries to assign styleSelector, and another sets it as a computed

PR #4355, opened 2020-05-27
Make expanded SOS chart item shareable

na9da:
### What this PR does

A user cannot share the expanded SOS chart item, this PR makes it shareable. 

Currently, when creating the expanded chart item we simply duplicate the source item and then when we create a share link we only copy the items "user" traits. This means that when restoring the chart item from a share link its URL and other definition traits will be missing and the item wont load properly. So instead of duplicating the source item, we create a reference to the source item. This process is asynchronous so we also change `ChartExpandAndDownloadButtons` component to accept a promise of items.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### MobX Development
 
 #### mobx-next-release (mobx-32)
+* Made expanded SOS chart item shareable.
 * Fixed a regression bug where the time filter is shown for all satellite imagery items
 * (💫The next rad feature💫 but please be mostly bug fixes from now until June!)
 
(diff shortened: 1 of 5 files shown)


rowanwins:
Looks ok. From what I can tell sharing of charts is generally pretty broken in mobx brnach at the moment so think some more work is needed elsewhere too. For example the `All generation types` item in the `simple.json` catalog also wont share it's chart prior to this PR, and still won't share it's chart.

rowanwins (approved):
Looks good @na9da

na9da:
@rowanwins Thanks for the merge.

I should get to writing that generic chart component then we can fix the issue across the board.

PR #4356, opened 2020-05-27
Fix diffable & discretelytimevarying mixin methods

soyarsauce:
### What this PR does

* Fixes a bug where WMS items caused type errors in newer babel and typescript builds, due to mixed mixin methods on DiffableMixin & DiscretelyTimeVaryingMixin

### Checklist
-   [x] unit tests n/a - tried to do this but it would involve different tsconfig options which would break a bunch of other bits of terria.


soyarsauce:
Fixes https://github.com/TerriaJS/terriajs/issues/4345

na9da (approved):
Looks good!

PR #4363, opened 2020-05-28
Mobx: set clampToGround trait default on GeoJsonCatalogItem

rowanwins:
### What this PR does

95% of geojson files in the wild aren't 3D and so if they get added with clampToGround: false they float around in space. This PR changes the default.

### Checklist

-   [x] No unit tests required


nf-s:
Thanks rowan, I need this!

PR #4383, opened 2020-05-29
Mobx moment points fixes

na9da:
### What this PR does

* Fixed a chart bug where moment points do not stick to the basis item when they are of different scale.
* Fixed a bug where the moment point selection highlight is lost when changing the satellite imagery date.
* The change to SOS catalog item is just wrapping a block of code in action to fix a whole lot of mobx warnings

Also adds typings for `react-test-renderer`

### Checklist


rowanwins:
~~Not quite right yet @na9da~~

~~If I was going to hazard a guess it would be something about the range of the axis being greater than the range of the values. eg the chart min is 0 but the data min is 0.45~~

Ignore the above - Should've built the changes!

rowanwins (approved):
Looks good

PR #4393, opened 2020-06-01
Mobx Legend improvements for TableMixin

nf-s:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4245

Related to https://github.com/TerriaJS/terriajs/pull/4120

* Add `multipleTitles` and `maxMultipleTitlesShowed` to `LegendItemTraits`
* Aggregate legend items in `createLegendItemsFromEnumColorMap` by colour, that is merge legend items with the same colour (using `multipleTitles`)
* Only generate `tableStyles` for region columns if no other styles exist

merged legend items

### Examples

Not generating region-column styles:  http://ci.terria.io/mobx-table-style-mods/#share=s-njw3Wy76cGdMXlpKDeNgfMMlWCh

Add `NSW LGA NEXIS 2012` layer and `POPULATION ESTIMATE` should be default:   http://ci.terria.io/mobx-table-style-mods/

Generating region-column styles: http://ci.terria.io/mobx-table-style-mods/#share=s-qtpyWzyZjPxyPAFd7jnGBBklMyJ

### Checklist

-   [x] Tests for this are handled by https://github.com/TerriaJS/terriajs/pull/4219 - see https://github.com/TerriaJS/terriajs/pull/4219/files#diff-76cab554c85da4157534e0a1c0e24386


nf-s:
what do you think @rowanwins @steve9164 ?

lib/ModelMixins/TableMixin.ts
```diff
@@ -102,7 +102,20 @@ export default function TableMixin<T extends Constructor<Model<TableTraits>>>(
       const value = super.activeStyle;
       if (value !== undefined) {
         return value;
-      } else if (this.styles && this.styles.length > 0) {
+      } else if (
+        this.styles &&
+        this.styles.length > 0 &&
+        isDefined(this.styles[0].id)
```
tephenavies: Why is `isDefined(this.styles[0].id)` needed?

tephenavies (changes requested):
I think we should discuss the behaviour and come up with better default style selection before we merge this.

I don't think we should default to a style based on a region column when you drag a region-mapped csv onto the map, but maybe there's another way to address this (e.g. maybe we shouldn't generate a single-colour colour scheme unless there are no other styles available).

lib/ModelMixins/TableMixin.ts
nf-s: Just to get rid of TypeError for `this.tableColumns.find(col => col.name === this.styles[i].id)`

lib/ReactViews/Workbench/Controls/Legend.jsx
```diff
@@ -207,7 +207,24 @@ const Legend = createReactClass({
...
+            <div
+              title={
+                isDefined(legendItem.multipleTitles)
+                  ? legendItem.multipleTitles.join(", ")
+                  : legendItem.title
+              }
+            >
+              {isDefined(legendItem.multipleTitles)
+                ? `${[...legendItem.multipleTitles]
+                    .splice(0, legendItem.maxMultipleTitlesShowed)
```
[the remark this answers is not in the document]
nf-s: ah yep! that makes much more sense

lib/ModelMixins/TableMixin.ts
nf-s: I can't remember, It must be residual from an earlier version. All of this is pretty rough.

nf-s:
I totally agree, I think only generating the single-colour scheme if nothing else is available would be definitely better than my approach

nf-s:
I have made the requested changes, let me know if you want me to look into what you suggested, or if you have any other ideas!
I'm not that familiar with `Table` stuff in Terria.

nf-s:
> I think we should discuss the behaviour and come up with better default style selection before we merge this.
> 
> I don't think we should default to a style based on a region column when you drag a region-mapped csv onto the map, but maybe there's another way to address this (e.g. maybe we shouldn't generate a single-colour colour scheme unless there are no other styles available).

I have addressed this by only generate tableStyles for region columns if no other styles exist

lib/Table/TableAutomaticStylesStratum.ts
```diff
@@ -258,11 +260,24 @@ export class ColorStyleLegend extends LoadableStratum(LegendTraits) {
...
+      }
+      return prev;
+    }, {});
+
+    return Object.entries(colorMapValues)
+      .map(value => {
         return createStratumInstance(LegendItemTraits, {
-          title: value,
-          color: colorMap.colors[i].toCssColorString()
+          multipleTitles: value[1],
```
KeyboardSounds: Could you put a comment in here explaining why you're using these particular indices to the value array?

KeyboardSounds (changes requested):

lib/Table/TableAutomaticStylesStratum.ts
nf-s: It is from `Object.entries(colorMapValues)`, so `value` is `[key, value]`

nf-s: It is from `Object.entries(colorMapValues)`, so `value` is `[key, value]`

KeyboardSounds: OH duh sorry of course

KeyboardSounds (approved):

lib/Traits/CatalogMemberTraits.ts
```diff
@@ -39,6 +39,13 @@ export class ShortReportTraits extends ModelTraits {
     description: "The content of the section."
   })
   content?: string;
+
```
nf-s: This is already in `next` not sure why it is showing a diff

KeyboardSounds: merge in `next` and see if it disappears?

lib/Table/TableAutomaticStylesStratum.ts
```diff
@@ -93,8 +92,9 @@ export default class TableAutomaticStylesStratum extends LoadableStratum(
   get defaultChartStyle(): StratumFromTraits<TableStyleTraits> | undefined {
     const scalarColumns = this.catalogItem.tableColumns.filter(
       column =>
-        column.type === TableColumnType.scalar ||
-        column.type === TableColumnType.time
+        !this.catalogItem.excludeStyles?.includes(column.name) &&
```
tephenavies: Why is it necessary to add `excludeStyles`? We already have `"type": "hidden"`. Are there situations where you want a column to be a special type (it looks like `excludeStyles` is only relevant if a column is "scalar" or "time") and not have a style for that column?

nf-s: You're right, absolutely not necessary - i have removed it

lib/Traits/TableTraits.ts
```diff
@@ -9,6 +9,7 @@ import primitiveTrait from "./primitiveTrait";
 import RasterLayerTraits from "./RasterLayerTraits";
 import TimeVaryingTraits from "./TimeVaryingTraits";
 import LayerOrderingTraits from "./LayerOrderingTraits";
+import primitiveArrayTrait from "./primitiveArrayTrait";
```
tephenavies: Unnecessary import

tephenavies (approved):

PR #4395, opened 2020-06-01
Add getAncestors contract

soyarsauce:
### What this PR does

Stops terriajs from breaking https://github.com/TerriaJS/terrace/issues/116 

### Checklist


soyarsauce:
GH just didn't get status updated with api - both build jobs pass

rowanwins (changes requested):
This is a pretty abstract test - can we add some code comments around when this occurs and the expected behaviour?

soyarsauce:
@rowanwins  The tests map almost 1 to 1 with the code comments in https://github.com/TerriaJS/terriajs/blob/083ab51eb0456d8e6512e49202b9234706324f74/lib/Models/getAncestors.ts#L4
But yes can do for extra information - have just added some extra to https://github.com/TerriaJS/terriajs/pull/4395/commits/083ab51eb0456d8e6512e49202b9234706324f74#diff-1adc935206ca05f402b814f5cf908194R29

rowanwins (approved):
Thanks for adding those comments, knowing there are comments in the src helps!

PR #4396, opened 2020-06-02
Mobx: prevent chart legend from displaying for more than 4 items

### LINKED ISSUE #4328: Mobx - charts with lots of variables cause things to be pushed off screen
When there is a only a single param on a chart it works ok
NormalChart

When we enable lots of params some things happen
- the map resizes
- the feature info pushes off the screen
- some chart elements go off the screen
ChartsOffScreen

rowanwins:
### What this PR does

Fixes #4328

This PR prevents a chart legend from displaying if there are more than 4 items on the chart. 

Previously this would result in
- the map being resized
- featureInfo being pushed on the screen

More than 4 is somewhat of an arbitrary number and this is very much a short term fix. 

People can still use the workbench to toggle items on and off and see the colours.

HiddenLegend

### Checklist

-   [ ] Nil tests


na9da (approved):
Looks alright to me :+1:

PR #4397, opened 2020-06-02
Stop stack overflow on addedByUser

soyarsauce:
### What this PR does

Another one to stop a side effect from terriajs breaking TerriaJS/terrace#116

### Checklist


lib/Core/addedByUser.ts
```diff
@@ -1,8 +1,24 @@
...
+  catalogMember: BaseModel,
+  options: {
+    depth: number;
+  } = {
+    depth: 0
+  }
+): boolean {
+  const depth = options.depth;
+  if (depth > 100) {
+    console.error(
```
soyarsauce, line 15: not sure how often this is going to happen so don't want to raise a terriaerror with raisetouser:true here

rowanwins:
So I'm trying to understand the issue we're trying to solve... is it basically when we have a child group which references a parent group somewhere up the hierarchy?

And if I understand we're basically setting a threshold of 100 to prevent infinite looping... TBH this seems like a less than ideal fix - can't we just detect that the modelId has already been added?

soyarsauce:
@rowanwins this just fixes a recursive function in the codebase that doesn't have an exit when it goes far deeper than we would expect it to, while keeping the current behaviour.

The problem is when you somehow load in a catalog that has A > B > C > B.
1. B checks for its known container IDs, which are [A, C]
2. C checks for its known container IDs, which include [B]
3. back to 1 forever

We'd basically have to rewrite the entire `addedByUser` function if you're asking to detect for modelIds that it has checked, as it doesn't do that right now - and I don't know the implications of rewriting as we haven't explicitly defined the exact rules for what is and isn't allowed in a `knownContainerUniqueIds`

I'm happy if you want to raise a separate issue to rewrite `addedByUser`, but I still think having the depth guard on the recursion is better as right now it's still possible to crash the UI without this in place.

rowanwins:
Okay - if it's the best we can do in the short term then let's proceed and not let perfect get in the way of good.

rowanwins (approved):
👍

soyarsauce:
Perfect exists in https://github.com/TerriaJS/terriajs/issues/4403

PR #4398, opened 2020-06-02
Mobx tool tweaks

na9da:
### What this PR does

* Added `ToolButton` for loading/unloading a tool
* Changed the `Tool` interface, now accepting prop `getToolComponent` instead of `toolComponent`
  This was required because when passing `toolComponent={import("tool")}` from top level, the component is loaded eagerly. We fix that by changing it to `getToolComponent={() => import("tool").module}`

### Checklist


lib/ReactViews/Tool.tsx
```diff
@@ -44,4 +47,48 @@ class Tool extends React.Component<PropsType> {
   }
 }
 
+interface ToolButtonProps extends ToolProps {
+  icon: string;
+}
+
+@observer
+export class ToolButton extends React.Component<ToolButtonProps> {
```
soyarsauce, line 55: Where is this used?

soyarsauce (commented):
Things I really like:
actually loading the tool on demand

Things I'm confused by:
I think generalising this (specifically the bits around `showCloseButton`) without also generalising `CloseToolButton` isn't useful enough without having a second example of an implementation of tool, that uses the MapIconButton to toggle it on ad off. Or have you already got another tool in the works that needs this? Any other thoughts @na9da ?

lib/ReactViews/Tool.tsx
na9da: Yep, so I made these changes for the place-editor tool.
You can find the usage here: https://github.com/TerriaJS/qld-digital-twin/blob/use-place-editor/lib/Views/UserInterface.jsx#L43

That tool unlike `DiffTool` is invoked from the Nav, hence the toggle behavior.

soyarsauce: Yeah these changes now makes a lot more sense with the place editor use!

zoran995:
What is the status of this? could really use it for some other tools

lib/ReactViewModels/ViewState.ts
```diff
@@ -592,3 +582,10 @@ export default class ViewState {
     return this.currentTool !== undefined;
   }
 }
+
+interface Tool {
+  toolName: string;
+  getToolComponent: () => React.Component | Promise<React.Component>;
+  showCloseButton: boolean;
+  params: any;
```
zoran995:
```suggestion
  params?: any;
```
Should params be an optional parameter

na9da:
> What is the status of this? could really use it for some other tools

The PR is kinda stable now.

@soyarsauce can you review this sometime, please?

soyarsauce (changes requested):
@na9da could you add a simple render/mount test for Tool.tsx and maybe even the `getToolComponent` prop - just to have something & ensure no regression

lib/ReactViews/Tool.tsx
```diff
@@ -44,4 +47,48 @@ class Tool extends React.Component<PropsType> {
...
+      currentTool.getToolComponent === this.props.getToolComponent
+    );
+  }
+
+  toggleOpen() {
+    const { viewState } = this.props;
+    if (this.isThisToolOpen) {
+      viewState.closeTool();
+    } else {
+      viewState.openTool({ ...this.props, showCloseButton: false });
```
soyarsauce: You spread props here but should it be stored on `params`? (or even a toolProps key) Haven't looked at your place editor implementation but at least we don't pollute the `Tool` interface with props

soyarsauce (commented):
Let's get this in , pending my final question

soyarsauce:
Great work on the tests for direct item vs item-as-promise, btw

lib/ReactViews/Tool.tsx
na9da: The only `ToolButton` specific prop here is icon. The rest are meant for the `Tool` component which will correctly pass down the `params` object as props to the user component. I have improved it a bit by making it more explicit what we pass down.

soyarsauce (approved):
Bloody awesome work - merge it when CI passes

zoran995:
@na9da I am having some troubles when using `ToolButton` component with function `isThisToolOpen` as defined [here](https://github.com/TerriaJS/terriajs/pull/4398/files#diff-8950c72c5625324f47baa2651370eb3eR57-R63) it returns false so the button doesn't work as intentioned (it once returns true and then false). Any advice on this. 
```
<ToolButton
  toolName="Measure"
  icon={Icon.GLYPHS.measure}
  getToolComponent={() =>
    import("../Tools/DrawingTool/MeasureTool").then(
      m => m.default
    )
  }
  key="measure-tool"
/>
```
I tried changing it to `toolName` but then it introduces an issue with possibility to have one tool open and button of second with the closing state (not sure is this handled here).

PR #4399, opened 2020-06-02
Model transformations, 3d tiles feature hiding

na9da:
### What this PR does

* Added `TransformationTraits` that can be used to change position/rotation/scale of a model.
* Place editor traits - currently we have to add it to `TerriaJS`. There is no way to extend traits from outside
* Adds a method to hide 3d tiles feature persistently by generating styling rules for the feature

This makes it possible to change a model and share the changes.

Note that we have also modified `3dTilesCatalogItem` to copy its root tile transformation to `modelMatrix` replacing it with an `IDENITITY` matrix. Through this simplification we can control the models transformation using a single matrix.

### Checklist


lib/Models/Cesium3DTilesCatalogItem.ts
```diff
@@ -43,6 +48,11 @@ class ObservableCesium3DTileset extends Cesium3DTileset {
   }
 }
 
+interface ObservableCesium3DTileset {
+  modelMatrix: Matrix4;
```
KeyboardSounds: Why are we storing a separate copy of the tileset's modelMatrix? is it for reactivity? Or some other reason?

KeyboardSounds: Also I'm not super keen on defining a class and an interface with the same name in the same file

na9da: Noted! This is just to make the type checker happy. `modelMatrix` should ideally be defined in the cesium typings.

soyarsauce:
@na9da to fix broken test

lib/Models/Cesium3DTilesCatalogItem.ts
na9da: Fixed by moving the extra typings to the terriajs-cesium typings file.

wwwroot/test/Cesium3DTiles/tileset.json
```diff
@@ -0,0 +1,95 @@
...
+        0,
+        0,
+        0,
+        5.0375
+      ]
+    },
+    "geometricError": 1,
+    "refine": "REPLACE",
+    "content": {
+      "uri": "dragon_low.b3dm"
```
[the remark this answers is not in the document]
na9da: No, it is not. But since we are just using the file for tests and not actually rendering anything, do we need it?

KeyboardSounds (approved):

PR #4402, opened 2020-06-03
Workaround for feature item outline style.

mwu2018:
### What this PR does

Fixes [this issue](https://github.com/TerriaJS/nsw-digital-twin/issues/413)

If filled colours of polygon is of 100% opacity,  outline of items from feature server can not be rendered correctly in 3D mode when viewing right at the top of the map. This PR add some transparency to solve the problem.

The root cause is still under investigation and this PR provide a work-around solution.

### Checklist


zoran995:
This issue might be connected with the problem that cesium can't render polygons with a stroke-width > 1.0 ([check this comment](https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/GeoJsonCatalogItem.ts#L332-L334)). 
Maybe you can check #4405 I believe this might be fixed there

mwu2018:
> This issue might be connected with the problem that cesium can't render polygons with a stroke-width > 1.0 ([check this comment](https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/GeoJsonCatalogItem.ts#L332-L334)).
> Maybe you can check #4405 I believe this might be fixed there

Thanks @zoran995 . I have a new approach. Please see https://github.com/TerriaJS/terriajs/pull/4404. I will also try yours too.

mwu2018:
Replaced by https://github.com/TerriaJS/terriajs/pull/4404.

PR #4404, opened 2020-06-03
Use either polygon or polyline.

### LINKED ISSUE #4420: Should render polygon outlines from feature server
With the existing implementation, polygon outlines can not be seen from top viewing angle in 3D mode:

image

If viewing away from the top, the outlines can be seen:

image

The outlines should always be seen.

More details can be found [here](https://github.com/TerriaJS/nsw-digital-twin/issues/413).

mwu2018:
### What this PR does

Fixes #4420 

The existing implementation renders both polygon outlines and polylines. If the polylines are opaque, the polygon outlines will be completely blocked if being viewed right from the top angle. When being viewed from non-top angle, because the lighting condition of polylines is reduced, the polygon outlines will become visible.

The above problem is caused by a bug where `symbol.color` is used as polyline material color. Instead `symbol.outline.color` should be used.

This PR also add a new unit test.

Another [PR](https://github.com/TerriaJS/terriajs/pull/4405) might have given a better solution but it seems not working with the latest mobx branch.

### Checklist


rowanwins:
Using https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/FeatureServer/8
at the moment the incorrect outline colour is assigned, it ought to be a light green-ish color.

Color

mwu2018:
> Using https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/FeatureServer/8
> at the moment the incorrect outline colour is assigned, it ought to be a light green-ish color.
> 
> Color

This should have been fixed with the new commit.

mwu2018:
Replaced by https://github.com/TerriaJS/terriajs/pull/4412.

PR #4412, opened 2020-06-04
Feature item

### LINKED ISSUE #4420: Should render polygon outlines from feature server

mwu2018:
### What this PR does

Fixes #4420 and more.

See https://github.com/TerriaJS/terriajs/pull/4405. 

Reproduced here mainly for easy CI tests.

***Limitation***
Add a unit test only for feature polygon/polyline styling.

### Checklist


CHANGES.md
```diff
@@ -3,6 +3,8 @@ Change Log
 
 ### MobX Development
 
+
+#### mobx-next-release (mobx-32)
```
rowanwins: remove

rowanwins: And add the actual changes

package.json
```diff
@@ -179,6 +179,7 @@
     "react-test-renderer": "^16.3.2",
     "regenerator-runtime": "^0.13.2",
     "terriajs-server": "^3.0.1",
+    "webpack-cli": "^3.3.11",
```
rowanwins, line 182: Not sure why this is here?

rowanwins (changes requested):
Almost there!

package.json
mwu2018: you need it to run
```
  yarn dev
```

CHANGES.md
mwu2018: done

package.json
rowanwins: Huh - well there you go, just did a fresh clone and got the same. 
````
> webpack-dev-server --inline --config buildprocess/webpack.config.dev.js --host 0.0.0.0

The CLI moved into a separate package: webpack-cli
Please install 'webpack-cli' in addition to webpack itself to use the CLI
-> When using npm: npm i -D webpack-cli
-> When using yarn: yarn add -D webpack-cli
internal/modules/cjs/loader.js:638
    throw err;
    ^

Error: Cannot find module 'webpack-cli/bin/config-yargs'
````
I wonder if it's related to a new version of webpack? 
Or if the same issue occurs if you run terriajs as a package within a terriamap instance - in my test I just pulled down terriajs. TerriaMap has webpack-cli as a dependency already so that might circumvent the issues when running it nested as a package.

Comment (mwu2018) [reply package.json:182]: I have never run terriajs as a package within a terriamap instance locally.

File: lib/Map/LeafletVisualizer.ts [hunk 3]
```diff
@@ -668,15 +681,22 @@ class LeafletGeomVisualizer {
 
     let layer = details.layer;
     if (!isDefined(layer)) {
-      const polygonOptions = {
+      console.log(outlineColor);
```
rowanwins: delete me

mwu2018: done

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "terriajs",
-  "version": "7.10.0-mobx-33",
+  "version": "7.10.0-mobx-34",
```
rowanwins: Remove this - release manager will bump versions

mwu2018: done

lib/Models/esriLineStyle.ts
```diff
@@ -0,0 +1,43 @@
+const defaultDashArray = [4, 3];
+
+export const esriLineStyleCesium: {
```
zoran995:
```suggestion
const esriLineStyleCesium: {
```

forgot to remove export in the last commit (this is only used in this file)

rowanwins (approved):

PR #4428, opened 2020-06-10
Automate NPM publishing

tephenavies:
### Publish to NPM automatically from GitHub Actions

Currently npm publishing is manual and requires a good internet connection. We should instead publish automatically from GitHub Actions whenever someone pushes a version tag to GitHub.

We should combine this with an automated slack message that notifies everyone (`@here`) in `#terriajs` that terriajs has been published.

E.g. an action like https://github.com/marketplace/actions/post-slack-message

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,32 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+      - run: npm install
+      - run: npx gulp lint release
+
+  publish-npm:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+          registry-url: https://registry.npmjs.org/
+      - run: npm publish --tag test-auto-deploy
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
--- a/package.json
+++ b/package.json
@@ -131,7 +131,6 @@
   },
   "scripts": {
     "prepublish": "gulp post-npm-install",
-    "postpublish": "bash -c \"git tag -a ${npm_package_version} -m \"${npm_package_version}\" && git push origin ${npm_package_version}\"",
     "postinstall": "gulp post-npm-install",
     "gulp": "gulp",
     "make-schema": "gulp make-schema",


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
```
soyarsauce, line 7: Would it not be better to run this on `push` (+ filtering to run only on the `master`+`mobx` branches) instead and check if version has been incremented? The "manual step" can just be the release PR which increments the version + tidies changelog, instead of manually creating a GH release

soyarsauce: This also fixes the problem where you create a GH release but turns out there's a test suite failure, but I guess we have travis to tell us that if we went with the "on GH release" method.

soyarsauce: I missed the context message in slack, the gh release probably works out better if we can bake into the process to check for successful travis checks before tagging a specific commit

soyarsauce: >As long as it tagged & published the first merge commit to master/mobx after a version change (and didn't do the actual version bump commit).

This can really go either way depending on whether you do a merge commit or a rebase merge.

>having a developer push a tag is a better workflow than editing versions anywhere

In any case, both need to happen anyway? `npm publish` must be done on a unique version, and keep in mind creating a GH release is separate to pushing a git tag. Anyway taking all this discussion on board, I think we trial the "create a GH release, to do a terriajs release" way (aka what you've proposed with this PR) and see how it goes.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,32 @@
...
+  publish-npm:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+          registry-url: https://registry.npmjs.org/
+      - run: npm publish --tag test-auto-deploy
```
soyarsauce, line 30: You'll need to update this tag before merging

.github/workflows/npm-publish.yml
tephenavies: > > As long as it tagged & published the first merge commit to master/mobx after a version change (and didn't do the actual version bump commit).
> 
> This can really go either way depending on whether you do a merge commit or a rebase merge.

Actually in [my tests](https://github.com/steve9164/test-npm-publish) it was fine using merges, since it only runs on commits to the main branches, so it's triggered on the merge commit, correctly detects the version change and published that. Rebases should work too, they're the simpler case.

> > having a developer push a tag is a better workflow than editing versions anywhere

That sentence was about TerriaMap clones using `npm run deploy`. A bit of a tangent for this PR.

tephenavies:
See alternative approach here: https://github.com/TerriaJS/terriajs/pull/4436

tephenavies:
Closed in favour of alternative linked above

PR #4436, opened 2020-06-11
Add npm publish GitHub Action

tephenavies:
Add automated NPM publishing. Requires an NPM_TOKEN in secrets.

This approach will publish an NPM package revision any time a version change commit makes it to `mobx`, and since `mobx` is protected for all users that will always be a merge commit from a reviewed PR

~Things I still want to add:~
- [x] A way to send a slack message to `#terriajs` to notify devs of a successful or failed npm publish
    * Hopefully through some github actions / slack official or third-party integration
    * Could instead make a specific slack bot to post specific things
    * Plain webhooks?

Things we might add later:
- [ ] Something that labels PRs that will cause a release with a label called something like "Release PR". And make it red
    * The version change checker supports PRs so this should be fairly easy to do, but we might have to create a custom action script for applying/removing the label

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,40 @@
+name: npm-publish
+on:
+  push:
+    branches:
+      - mobx
+
+jobs:
+  publish:
+    name: Publish to NPM
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Check version changes
+        uses: EndBug/version-check@v1
+        id: check
+        with:
+          diff-search: true
+
+      - name: Version update detected
+        if: steps.check.outputs.changed == 'true'
+        run: 'echo "Version change found! New version: ${{ steps.check.outputs.version }} (${{ steps.check.outputs.type }})"'
+
+      - name: Set up Node.js for NPM
+        if: steps.check.outputs.changed == 'true'
+        uses: actions/setup-node@v1
+        with:
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install dependencies
+        if: steps.check.outputs.changed == 'true'
+        run: npm install
+
+      - name: Publish package to NPM
+        if: steps.check.outputs.changed == 'true'
+        run: npm publish --tag mobx
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -32,6 +32,12 @@ jobs:
       - name: Install dependencies
         if: steps.check.outputs.changed == 'true'
         run: npm install
+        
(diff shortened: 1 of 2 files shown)


tephenavies:
Finally added git tagging (which then causes the standard GitHub releases) and slack messaging. This is ready to go ~after adding an `NPM_TOKEN` and a `SLACK_WEBHOOK_URL`~ (done!). We can add the PR labeller later.

tephenavies:
Thoroughly tested on [a dummy repo](https://github.com/steve9164/test-npm-publish)
* [npm page](https://www.npmjs.com/package/@steve9164/test-npm-publish)
* [GitHub Actions](https://github.com/steve9164/test-npm-publish/actions)

Slack messages:
image

package.json
```diff
@@ -183,7 +183,7 @@
   },
   "scripts": {
     "prepublish": "gulp post-npm-install",
-    "postpublish": "bash -c \"git tag -a ${npm_package_version} -m \"${npm_package_version}\" && git push origin ${npm_package_version}\"",
+    "postpublish": "bash -c \"if [ -z \"$GITHUB_ACTION\" ]; then git tag -a ${npm_package_version} -m \"${npm_package_version}\" && git push origin ${npm_package_version}; else echo 'Running in GitHub Action. Must be tagged separately'; fi \"",
```
KeyboardSounds: What does "tagged separately" mean here?

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,84 @@
...
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          tag-name: ${{ steps.check.outputs.version }}
+          tag-message: ${{ steps.check.outputs.version }}
+          tagger-name: Terria Bot via GitHub Actions
+          tagger-email: TerriaBot@users.noreply.github.com
+
+      # These slack success and failure message actions can be combined but it means there'll be no notification on a failed
+      #  version-check (which is probably unlikely to fail). It also allows for a bit better message customisation
+      - name: action-slack-success
```
KeyboardSounds: Why is it called `action-slack-success` rather than `Send success notification on Slack` or something? Does the name have special meaning for this one?

tephenavies: Very good point. At some point I think I got confused between `id` and `name`.

package.json
KeyboardSounds: Do you think the postpublish script is now long enough that we should put it in a separate file? Then it will be more readable and we can add comments and stuff

KeyboardSounds:
My comments are mostly cosmetic pedantry, I can't see anything functionally wrong here. I am super excited for this, nice work @steve9164 🎉

package.json
tephenavies: I've removed the message, which was confusing worded and printed into logs of the GitHub Action where it could be interpreted wrongly. I think it should remain in package.json. Now that it's shorter it is clearer what it's doing

soyarsauce (approved):

PR #4437, opened 2020-06-11
Mobx: Fix timeline stack handling 

rowanwins:
### What this PR does

Fixes https://github.com/TerriaJS/qld-digital-twin/issues/152

When there were two datasets in the workbench with timeline button turned on the date pickers wouldn't work. Broken example
http://ci.terria.io/mobx/#share=s-wTQol9D8BGhIH8Mb51NVRFreu4m

The issue was that the `top` getter in the TimelineStack finds the first item with dates, however when we clicked the datepicker buttons we want our item to be the top item, however it was added to the end of the array rather than the start hence it would be intercepted by any item before it.

### Checklist


reginapramesti (approved):
Looks good!

PR #4442, opened 2020-06-15
Nominatim search provider

zoran995:
### What this PR does

Port nominatim search provider from master. 

### Checklist


rowanwins:
Strictly speaking using Nominatim in an autocomplete box is against their [terms of service](https://operations.osmfoundation.org/policies/nominatim/). That said we do debounce our requests quite heavily so I think we'd be under the 1 second threshold. 

I think in most of our maps we've tended to steer away from using the Nominatim search provider for some of those reasons, but seeing as it's done it's probably ok to be merged and it's up to app creators to ensure they are within the terms of use. 
@steve9164 what do you think?

tephenavies:
I agree with your suggestion @rowanwins.

lib/Models/NominatimSearchProvider.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+export class ExtendedNominatimSearchProvider extends NominatimSearchProvider {
+  doSearch(
+    searchText: string,
+    searchResults: SearchProviderResults
+  ): Promise<void> {
+    return super.doSearch(searchText, searchResults);
+  }
+}
```
tephenavies: @zoran995 I think it would be better to test this provider through the public `search` function, rather than `doSearch`. Is there a reason this can't be done?

If it can't be done that way then this test helper should be moved into the test file.

lib/Models/SearchResult.ts
```diff
@@ -23,6 +24,7 @@ export default class SearchResult {
   @observable isOpen = false;
   @observable type: string = "search_result";
   @observable location: { longitude: number; latitude: number } | undefined;
+  @observable id: number | undefined;
```
tephenavies: I think to be more general the id should be `string | undefined`. Across the rest of Terria ids are predominately `string`s. And any `number` can be cast to a `string`, but not the other way around.

```suggestion
  @observable id: string | undefined;
```

lib/Models/SearchResult.ts
```diff
@@ -12,6 +12,7 @@ export interface SearchResultOptions {
   clickAction?: () => void;
   catalogItem?: BaseModel;
   location?: { longitude: number; latitude: number };
+  id?: number;
```
tephenavies:
```suggestion
  id?: string;
```
See other comment.

tephenavies (changes requested):

lib/Models/NominatimSearchProvider.ts
zoran995: the only way that I saw that could be possible with `search` was to add some timeout so I moved to doSearch.

zoran995: moved this to test file

pjonsson:
Is this superseded by #7302, or does this contain more things that should go on top of that PR?

PR #4443, opened 2020-06-17
Changes for invoking MyData without using the ExplorerWindow component

na9da:
### What this PR does

As of now it is not possible to open the upload data modal without showing the other tabs like the catalog explorer. In some tools we want the user to upload some data and the upload modal should not show other tabs. This PR makes it possible to open the upload data modal independent of the `ExplorerWindow` component.

List of changes:
     * Makes it possible to invoke MyData tab from a component different from ExplorerWindow
     * The extensions shown in the AddData dropdown are now customizable
     * Receive a callback when files are uploaded

### Checklist

### DIFF
--- a/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
+++ b/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
@@ -33,6 +33,7 @@ const AddData = createReactClass({
     activeTab: PropTypes.string,
     localDataTypes: PropTypes.arrayOf(PropTypes.object),
     remoteDataTypes: PropTypes.arrayOf(PropTypes.object),
+    onFileAddFinished: PropTypes.func.isRequired,
     t: PropTypes.func.isRequired
   },
 
--- a/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
+++ b/lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
@@ -113,13 +113,15 @@ const AddData = createReactClass({
       }
     }
     addUserCatalogMember(this.props.terria, promise).then(addedItem => {
+      if (addedItem && !(addedItem instanceof TerriaError)) {
+        this.onFileAddFinished([addedItem]);
+      }
+      // FIXME: Setting state here might result in a react warning if the
+      // component unmounts before the promise finishes
       this.setState({
         isLoading: false
       });
       this.props.resetTab();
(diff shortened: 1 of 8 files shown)


lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
```diff
@@ -103,19 +113,18 @@ const AddData = createReactClass({
...
     addUserCatalogMember(this.props.terria, promise).then(addedItem => {
-      if (addedItem && !(addedItem instanceof TerriaError)) {
-        this.onFileAddFinished(addedItem);
-      }
       this.setState({
         isLoading: false
       });
       this.props.resetTab();
+      if (addedItem && !(addedItem instanceof TerriaError)) {
+        this.onFileAddFinished([addedItem]);
```
tephenavies: Can you explain this change?

Does moving this processing after setting `isLoading` to false mean that we don't have a loading state? I know the current synchronous loading of catalog items leads to UI freezes, but is is better this way?

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
```diff
@@ -103,19 +113,18 @@ const AddData = createReactClass({
...
+      }
     });
   },
 
-  onFileAddFinished(fileToSelect) {
-    this.props.viewState.myDataIsUploadView = false;
-    this.props.viewState.viewCatalogMember(fileToSelect);
+  onFileAddFinished(files) {
+    this.props.onFileAddFinished(files);
```
tephenavies: I think it would be cleaner to remove this function entirely and just use `this.props.onFileAddFinished` in the places that this was used.

tephenavies (commented):
This is really cool, and I'm pretty impressed that it took so few changes.

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
na9da: So when my custom component receives `onFileAddFinished` its state changes and the explorer component gets unmounted. If we call, `setState` after the unmount has happened, react throws an error (or is it a warning). But you are right, changing the order here is not correct because we don't get the loading indicator for the item load. I'll think of a fix.

na9da: I have reverted the change for now.

tephenavies (approved):

PR #4444, opened 2020-06-17
feedback form is now scrollable

KeyboardSounds:
### What this PR does

Set a max height for the feedback form and makes it scrollable so that it doesn't break on small screens

### Checklist

-   [ ] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ N/A, tiny CSS change
-   [ ] ~I've updated CHANGES.md with what I changed.~ N/A, tiny CSS change


reginapramesti (approved):
This looks good! Thanks 😄

PR #4458, opened 2020-06-19
Add upload as tour point

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4451

* Add "Upload" to tour points

### Checklist

-   [x] unti tets covered in TourPortalSpec.tsx There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)

### DIFF
--- a/lib/ReactViews/SidePanel/SidePanel.jsx
+++ b/lib/ReactViews/SidePanel/SidePanel.jsx
@@ -106,6 +106,7 @@ SidePanelButton.propTypes = {
 };
 
 export const EXPLORE_MAP_DATA_NAME = "ExploreMapDataButton";
+export const SIDE_PANEL_UPLOAD_BUTTON_NAME = "SidePanelUploadButton";
 
 const SidePanel = observer(
   createReactClass({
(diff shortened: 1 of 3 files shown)


soyarsauce:
OK to go into every default tour

lib/ReactViews/SidePanel/SidePanel.jsx
```diff
@@ -209,6 +212,24 @@ const SidePanel = observer(
...
+const RefForUpload = props => {
+  const SidePanelWrappedForExplore = withTerriaRef(
+    SidePanel,
+    EXPLORE_MAP_DATA_NAME
+  );
+  // it's important we pull outt refFromHOC and not override it again on rest
+  // so that the second HOC can correctly provide the EXPLORE_MAP_DATA_NAME prop
+  const { refFromHOC, ...rest } = props;
+  return (
+    <SidePanelWrappedForExplore refFromHOCForUpload={refFromHOC} {...rest} />
```
tephenavies, line 226: Neat, but I agree with the above comment. It would be better to allow for more than 1 ref using the original HOC.

tephenavies (approved):

PR #4475, opened 2020-06-26
Fix console warnings

zoran995:
### What this PR does

There are some warnings in console introduced in recent changes. It is generally recommended to use `useTranslation` hook in functional components instead of `withTranslation` HOC. 
There is also a
```javascript
Warning: Received `false` for a non-boolean attribute `id`.
```
which is coming from [here](https://github.com/TerriaJS/terriajs/blob/mobx/lib/ReactViews/Generic/TooltipWrapper.tsx#L296) but I am not sure about the best way to handle it and where it is used, maybe (`id: applyAriaId ? idForAria : undefined`). 

### Checklist


soyarsauce (approved):

soyarsauce:
Thanks @zoran995 ! You must be really actively using mobx in development to catch these so quickly 🌟 👍

zoran995:
You are welcome @soyarsauce, sorry for introducing the help prompt issue. We moved our project on mobx and need to move everything we developed so catching up as fast as we can

PR #4482, opened 2020-06-29
Mobx wcs fixes

nf-s:
### WCS fixes

http://ci.terria.io/mobx-wcs-fixes/#clean&http://gist.githack.com/nf-s/8a55c04720ec60f5d210f794ad604599/raw/893b14ed48aa9569a09616a0283a5ba4d40a9b9d/wcs-test.json

Fixes:
- ExportData returning undefined instead of null
- WCS error reporting (now parses XML for error message)
- ExportableData now has `exportEnabled` property
- Add leaflet visualiser Rectangle support

### DIFF
--- a/lib/Models/CsvCatalogItem.ts
+++ b/lib/Models/CsvCatalogItem.ts
@@ -78,6 +78,15 @@ export default class CsvCatalogItem
     return isDefined(this._csvFile);
   }
 
+  @computed
+  get exportEnabled() {
+    return (
+      isDefined(this._csvFile) ||
+      isDefined(this.csvString) ||
+      isDefined(this.url)
+    );
+  }
+
   async exportData() {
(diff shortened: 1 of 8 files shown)


nf-s:
Not sure why deployment has failed, the last commit only changes a type import

lib/Models/CsvCatalogItem.ts
```diff
@@ -78,6 +78,15 @@ export default class CsvCatalogItem
     return isDefined(this._csvFile);
   }
 
+  @computed
+  get exportEnabled() {
```
tephenavies: I think this should be `canExportData` to match `canZoomTo`.

tephenavies: I'll fix and push

tephenavies (approved):

PR #4489, opened 2020-06-30
Mobx fix prompt

soyarsauce:
### What this PR does

Fixes #https://github.com/TerriaJS/terriajs/issues/4488
Fix pointer-events on map navigation regression
Only fixes one case found from https://github.com/TerriaJS/terriajs/pull/4477
Possibly revert it and take a better approach on the size of the actual container so we don't have to use pointer events

### Checklist

-   [x] unit test n/a

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 
 #### next release (mobx-36)
 * Added `pointer-events` to `MapNavigation` and `MenuBar` elements, so the bar don't block mouse click outside of the button.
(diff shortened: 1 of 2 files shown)


soyarsauce:
And perhaps a ping for @zoran995 to help trace any other affected bits

reginapramesti (approved):
👍

zoran995:
@soyarsauce will check it, completely missed that this prompt exist at all (it's not on by default)

PR #4507, opened 2020-07-07
Next: Add typeName to CatalogMemberMixin

nf-s:


lib/ModelMixins/CatalogMemberMixin.ts
```diff
@@ -11,6 +11,7 @@ type CatalogMember = Model<CatalogMemberTraits>;
 function CatalogMemberMixin<T extends Constructor<CatalogMember>>(Base: T) {
   abstract class CatalogMemberMixin extends AccessControlMixin(Base) {
     abstract get type(): string;
+    readonly typeName?: string;
```
soyarsauce: Every implementation of this is a getter, would be better to reflect that here?

nf-s: Yeah you're probably right, but wouldn't `readonly` be better than getters for properties which are set in the class definition and never change? 
Like `get type(): string` also

CHANGES.md
```diff
@@ -3,7 +3,7 @@ Change Log
 
 ### MobX Development
 
-#### next release (8.0.0-alpha.37)
```
soyarsauce, line 6: By the way this should be kept in

lib/ModelMixins/CatalogMemberMixin.ts
soyarsauce: Only reason I think typename could change is i18n but that might make more sense to map the type against translations rather than setting them on models

soyarsauce (approved):

PR #4508, opened 2020-07-07
Next: Add header option to loadText

nf-s:
### Add header option to loadText


rowanwins (approved):
Looks good!

rowanwins:
Will need to typescript-ify this at some stage @nf-s but think we can do that down the track

PR #4512, opened 2020-07-07
Add refreshEnabled trait and AsyncMappableMixin to AutoRefreshMixin

nf-s:
### Add `refreshEnabled` trait and `AsyncMappableMixin` to `AutoRefreshMixin`

This adds a 'toggle' trait so the AutoRefreshMixin can be turned off or back on.

Example with 2 CSV catalog items: one with `refreshEnabled=true` and one with `refershEnabled=false`:  
http://ci.terria.io/autorefreshmixin-refreshenabled/#clean&https://gist.githubusercontent.com/nf-s/7add30fc21d000e5e1710faf09e0dc4b/raw/1abf730796ef82d8c332d3aec7d7a2eec3db90d6/polling.json

**Motivation**

This is used in the `CatalogFunctionMixin` and `CatalogFuntionJobMixin` - https://github.com/TerriaJS/terriajs/pull/4229

For example - here https://github.com/TerriaJS/terriajs/blob/79d51860836270eff93eab6008a73298c81b689b/lib/ModelMixins/CatalogFunctionJobMixin.ts


lib/Traits/CsvCatalogItemTraits.ts
```diff
@@ -33,6 +34,7 @@ class PollingTraits extends ModelTraits {
 }
 
 export default class CsvCatalogItemTraits extends mixTraits(
+  AutoRefreshingTraits,
```
nf-s, line 37: Maybe this shouldn't be here, as all `AutoRefreshingTraits` are computed values in `CsvCatalogItem`

lib/ModelMixins/AutoRefreshingMixin.ts
```diff
@@ -3,44 +3,67 @@ import {
...
   T extends Constructor<AutoRefreshing>
 >(Base: T) {
-  abstract class AutoRefreshingMixin extends Base {
-    _autoRefreshDisposer: IReactionDisposer | undefined;
+  abstract class AutoRefreshingMixin extends AsyncMappableMixin(Base) {
+    private autoRefreshDisposer: IReactionDisposer | undefined;
+    private autorunRefreshEnableDisposer: IReactionDisposer | undefined;
 
-    /* Return the interval in seconds to poll for updates. */
+    /** Return the interval in seconds to poll for updates. */
```
rowanwins, line 24: You have very specific code comment formatting :)

nf-s: haha that is just for JSDoc/Vscode - otherwise i don't get the description when I hover over it

rowanwins (approved):
All looks reasonable to me @nf-s

PR #4521, opened 2020-07-08
Mobx: Add infoSectionOrder trait and some extra sections to WebMapServiceCatalogItem info array

### LINKED ISSUE #4506: MobX: display all filled fields of WMS
There are cases when WMS fields are missing from data description/metadata even when they are filled by the data custodian. 
To add the following fields as part of metadata for a WMS: 

- GetCapabilties url 
- Service Contact

Example: 
- go to NSW Digital Twin - National Datasets - Drought -  Vegetation cover - https://nsw.digitaltwin.terria.io/#share=s-3VxqdKGRz51QdG8nUB7U8kgJ7wQ 

image

- same data tested on a non-MobX app map.terria.io, https://map.terria.io/#share=s-kwABt9PyYPCbNtVIsZjgQsVf7J0 produces metadata with Service Contact, Get Capabilities URL; the Service Description has a different title but the information is there:

image

rowanwins:
### What this PR does

Fixes #4506

Adds following sections to WMS info
- GetCapabilties url
- Service Contact

Also slightly improved the repetitive nature of the code.

And use some internationalization stuff which looks like it was lost.

## NOTE
Suggest we wait until #4558 is merged as it also added a trait for the purposes of testing, but we'll need to kill it off in this PR.

### Checklist

-   [ ] no unit tests, fairly minor change


lib/Models/WebMapServiceCatalogItem.ts
```diff
@@ -289,21 +308,25 @@ class GetCapabilitiesStratum extends LoadableStratum(
...
-        result.push(traits);
+        result.push(
+          createInfoSection(
+            i18next.t("models.webMapServiceCatalogItem.accessConstraints"),
+            service.AccessConstraints
+          )
+        );
       }
     }
     return result;
```
nf-s, line 333: For some reason these are displaying in reverse order to being pushed into `result`

**National Map**:  
natmap

**next**:  
next

nf-s (changes requested):

lib/Models/WebMapServiceCatalogItem.ts
rowanwins: Yeah I noticed similiar. 
Ordering is either controlled by some defaults, or by passing in params.

https://github.com/TerriaJS/terriajs/blob/master/lib/ReactViews/Preview/DataPreviewSections.jsx#L32-L46

I think the main change is that what appeared in master as  "Data Description" is now called "Web Map Service Layer Description", hence why it gets shuffled back. But that's not a change that I did.

rowanwins: Maybe I'll define an infoSectionOrder for the WMS item, the user can always override it with the definition stratum

nf-s (dismissed):

zoran995:
Should there be a skip condition for service contact, GeoServer by default provide `The Ancient Geographers` as a contact organisation

nf-s (approved):

lib/Models/WebMapServiceCatalogItem.ts
```diff
@@ -296,26 +308,68 @@ class GetCapabilitiesStratum extends LoadableStratum(
...
+        !layer ||
+        !layer.Abstract ||
+        containsAny(layer.Abstract, WebMapServiceCatalogItem.abstractsToIgnore)
+      ) {
+        continue;
+      }
+      if (this.capabilitiesLayers.size > 1) {
+        suffix = ` - ${layer.Title}`;
+      }
+    }
```
nf-s: I need to extend this to multiple abstracts showing - if using multi layer WMS

nf-s (changes requested):

nf-s (approved):

Issue #4278, opened 2020-05-05
MobX -  Internationalisation of help content


reginapramesti:
We've made the help content to be loaded in from `config.json` but have yet to use i18n for internationalisation. We should ideally support internationalisation for those who would like to have a multilingual map.

tephenavies:
Fixed in #4613.

PR #4523, opened 2020-07-08
Internationalize help content

### LINKED ISSUE #3964: i18n: add ability to load translation in from TerriaMap
It's great that we have i18n now, but no way to actually load in your own file without compiling your own version of terriajs. Ideally some way to merge translation overrides, instead of having to copy any given language wholesale.

zoran995:
### What this PR does

Fixes #3964, #4278

Add an internationalization support for Help content, Trainer Bar and Help terms content. This also enables loading the internationalization using backend service (same as in #4324 - this should probably replace that). Basic idea behind this is to user put help content into `languageOverrides.json` file and then using keys reference it from `config.json`. Exist approach also works without any problems (putting everything to `config.json`. I will start also a PR on TerriaMap for demo/testing purposes.

### Checklist


soyarsauce (changes requested):
This is great work, thanks @zoran995 - are you able to please review the additional changes I've made in https://github.com/TerriaJS/terriajs/pull/4613/files/01a93d1cc23ce9c7a8102f7f5aa5a52a5ecd18d2..345a6d6dc8e4ab50bd9071335a2fd9b5926f8ecc, especially to see if it works with your TerriaMap environment? Then we can get this merged 🙂

soyarsauce:
That's now https://github.com/TerriaJS/terriajs/pull/4613/files/01a93d1cc23ce9c7a8102f7f5aa5a52a5ecd18d2..f5d220dad82e22d4ff7e8ccb0b2af95c3ba3bd0e to pick up the latest commit to fix CI

PR #4525, opened 2020-07-09
Generic chart component

### LINKED ISSUE #4264: Write a generic <chart> component
The current `<chart>` component depends on CsvCatalogItem, so we have to write new components for other items that want to show a chart, eg: `<airquality-chart>`, `<sos-chart>`. 

Write a generic chart component that we can use all catalog items.

KeyboardSounds:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4264

New custom feature info chart components can now be made by inheriting from ChartCustomComponent.

### Checklist


lib/ReactViews/Custom/ChartCustomComponent.ts
```diff
@@ -1,96 +1,138 @@
...
 import ChartExpandAndDownloadButtons from "./Chart/ChartExpandAndDownloadButtons";
 import Chart from "./Chart/FeatureInfoPanelChart";
 import CustomComponent, { ProcessNodeContext } from "./CustomComponent";
+import Model, { BaseModel } from "../../Models/Model";
+import CatalogMemberTraits from "../../Traits/CatalogMemberTraits";
+import CsvCatalogItemTraits from "../../Traits/CsvCatalogItemTraits";
+import hasTraits from "../../Models/hasTraits";
+
+export interface ChartCustomComponentAttributes {
+  /**  The title of the chart.  If not supplied, defaults to the name of the context-supplied feature, if available, or else simply "Chart". */
```
rowanwins, line 17: We used to have an `id` attribute, by removing it we blow up existing charts because of the `checkAllPropertyKeys` method in `ChartCustomComponent`

rowanwins: We should probably have better error handling so the app doesn't whitescreen on unknown attributes

KeyboardSounds: We never actually _used_ the `id` attribute. It was never parsed or accessed in the code, we only used `identifier`. Any chart that's using it is probably not working as intended anyway, so isn't it better to see what's broken so we can fix it?

KeyboardSounds: I don't know. I'm pretty torn on this. I don't want to have to maintain every attribute that anyone has ever used in a chart ever, but breaking things more is bad too.

KeyboardSounds: also it's super confusing having two different identifier properties

lib/Language/en/translation.json
```diff
@@ -251,7 +251,7 @@
   },
   "welcomeMessage": {
     "title": "Let's get you started",
-    "welcomeMessage": "Interested in data discovery and exploration?<br/>Dive right in and get started or check the following help guide options.",
+    "welcomeMessage": "Welcome to the Queensland Digital Twin.<br/> Access a rich catalogue of 2D, 3D and live data from multiple sources in one convenient location. Take a tour guide or start exploring!",
```
rowanwins: Remove this change probably :)

lib/ReactViews/Custom/ChartCustomComponent.ts
rowanwins: I was using the "All generation types" item in the `simple.json` catalog and it broke because it had the id attribute

rowanwins: I can't say I've seen `identifier` used in the wild, but I have `id`, so I'd be inclined towards keeping `id` if we had to chose.

KeyboardSounds: I've seen both :(
How about we just don't throw an error 🤷 there's no good reason to.

KeyboardSounds: The only reason I think we should keep identifier over id is because identifier has always done something but id hasn't. So it's less likely to break something that was already working properly.

rowanwins: > How about we just don't throw an error 🤷 there's no good reason to.

Sounds like a plan.

If I read the code correctly I think `identifier` was only used by the sos-chart as an attribute, i don't think it was on the regular `chart` component. 
 I think back in the master branch `id` was used somewhere, but it didn't make the cut for mobx (probably because it does nothing!) but the doco was never adjusted and in porting items over featureInfoTemplates were just blindly cut and pasted.

rowanwins (approved):
Looks good - merge away!

lib/Models/SensorObservationServiceCatalogItem.ts
```diff
@@ -624,7 +624,7 @@ function createChartColumn(
   name: string | undefined
 ): string {
   const nameAttr = name == undefined ? "" : `name="${name}"`;
-  return `<sos-chart identifier="${identifier}" ${nameAttr}></sos-chart>`;
+  return `<sos-chart identifier="${identifier}" ${nameAttr} can-download="false"></sos-chart>`;
```
tephenavies, line 629: Can you explain this change?

tephenavies (changes requested):

lib/Models/SensorObservationServiceCatalogItem.ts
KeyboardSounds: The API that provides the chart data is a SOAP API, and the download button is essentially just a link, so when you click it you get an error page. I asked Nanda if it ever worked for SOS, and he said no. So essentially, this change hides a button that doesn't work, and hasn't ever worked.
This doesn't stop you from downloading the chart once you've expanded it, though.

tephenavies: Ok. Sounds like this change is an improvement on the current state then. Ideally sos-chart would default to no download button, or provide the ability to download data without expanding, but I don't think we need to do that now.

buildprocess/configureWebpack.js
```diff
@@ -94,7 +94,7 @@ function configureWebpack(terriaJSBasePath, config, devMode, hot, MiniCssExtract
                 loader: require.resolve('string-replace-loader'),
                 options: {
                     search: 'function _get\\(target, property, receiver\\).*',
-                    replace: 'var _get = require(\'../Core/superGet\').default;',
+                    replace: 'var _get = require(\'' + path.resolve(terriaJSBasePath, 'lib') + '/Core/superGet\').default;',
```
tephenavies, line 97: I can't believe we haven't tripped over this til now. Good to have a proper require path.

tephenavies (approved):

PR #4535, opened 2020-07-10
cesium 3d tiles mixin

KeyboardSounds:
### What this PR does

Makes 3D tiles support into a mixin so that it can be reused in multiple catalog items.

### Checklist

-   [ ] ~There are unit tests to verify my changes are correct or unit tests aren't applicable~ — n/a test the catalog item instead i'm in a hurry


lib/Traits/Cesium3dTilesTraits.ts
```diff
@@ -0,0 +1,185 @@
+import { JsonObject } from "../Core/Json";
```
rowanwins, line 1: I take it this file has just been copied and renamed, which is fine, the github diff however isn't showing the deletion of `Cesium3dCatalogItemTraits`?

lib/Models/Cesium3DTilesCatalogItem.ts
```diff
@@ -30,27 +30,11 @@ import createStratumInstance from "./createStratumInstance";
...
-      });
-    });
-  }
-}
+import Cesium3dTilesMixin from "../ModelMixins/Cesium3dTilesMixin";
 
 export default class Cesium3DTilesCatalogItem
   extends FeatureInfoMixin(
-    AsyncMappableMixin(
+    Cesium3dTilesMixin(
```
rowanwins, line 11: The top of the file still references `Cesium3DCatalogItemTraits.ts` which I presume is no longer required?

lib/Models/Cesium3DTilesCatalogItem.ts
```diff
@@ -60,332 +44,4 @@ export default class Cesium3DTilesCatalogItem
   get typeName() {
     return i18next.t("models.cesiumTerrain.name3D");
```
rowanwins, line 19: I know this wasn't your change but perhaps fix while we're here :)

lib/Models/Cesium3DTilesCatalogItem.ts
rowanwins: In fact there are a squillion imports which I guess aren't need anymore

rowanwins (changes requested):
A couple of super-minor things @KeyboardSounds but fine otherwise :)

lib/Models/Cesium3DTilesCatalogItem.ts
rowanwins: Actually turns out thats the correct lang string, it's perhaps just a bit weirdly named

lib/Language/en/translation.json
```diff
@@ -275,7 +275,7 @@
   },
   "welcomeMessage": {
     "title": "Let's get you started",
-    "welcomeMessage": "Interested in data discovery and exploration?<br/>Dive right in and get started or check the following help guide options.",
+    "welcomeMessage": "Welcome to the Queensland Digital Twin.<br/> Access a rich catalogue of 2D, 3D and live data from multiple sources in one convenient location. Take a tour guide or start exploring!",
```
rowanwins: Thought I dropped a comment on this one earlier sorry @KeyboardSounds

lib/Traits/Cesium3dTilesTraits.ts
rowanwins: Ignore this

lib/Models/Cesium3DTilesCatalogItem.ts
rowanwins: Done

lib/Models/Cesium3DTilesCatalogItem.ts
rowanwins: Resolve later I think

rowanwins (approved):
Looks good now!

PR #4553, opened 2020-07-21
Mobx: Add cacheDuration url trait

### LINKED ISSUE #4550: Version8: Add cacheDuration trait
In master we have a `cacheDuration` option on many catalog item types, we don't yet have that in mobx and so in many cases in our models we're manually setting cache length without any option to override. I'm thinking it's possibly a candidate for `UrlTraits`?

rowanwins:
### What this PR does

Fixes #4550

This PR adds a `cacheDuration` to the `UrlTraits`.

Individual models are then responsible for implementing something like which either grabs the configured cacheDuration from the user, or applies an appropriate default duration for that item type.

````
 @computed get cacheDuration (): string {
    if (isDefined(super.cacheDuration)) {
      return super.cacheDuration
    }
    return '0d'
  }
````

### Checklist

### CODE REVIEW

File: lib/Models/MagdaReference.ts [hunk 1]
```diff
@@ -713,6 +714,13 @@ export default class MagdaReference extends AccessControlMixin(
     return undefined;
   }
 
+  @computed get cacheDuration(): string {
+    if (isDefined(super.cacheDuration)) {
+      return super.cacheDuration;
+    }
+    return "1d";
```
nf-s: The previous value of this was `0d`, does that matter?

nf-s (changes requested):
And do we need to pass `cacheDuration` to `proxyCatalogItemUrl` as it is accessed within the function:
```js
return corsProxy.getURL(
    url,
    defaultValue(catalogItem.cacheDuration, cacheDuration)
  );
```

lib/Models/MagdaReference.ts
soyarsauce: Given 0d was the previous default we should also set that to 0d here until we see how that interacts with the proxy, overriding the cache control behaviour of magda-gateway via our proxy wouldn't be ideal

>it's fine when we access it directly, because we've forced our gateway to set private cache headers, but i'm not sure how that interaction changes when we put requests to magda through our terriajs-server proxy

>i would hope it's fine and terriajs-server honours those kind of headers but best to double check

>actually i would say it's essential to double check, auth is new to terriajs and to terriajs-server een more., (in the sense that terriajs-server isn't designed for an auth world.)

soyarsauce: Related: https://github.com/TerriaJS/nsw-digital-twin/issues/345

rowanwins: Yeah that was my bad, good pick up @nf-s

rowanwins: Fixed

KeyboardSounds:
If you're feeling particularly conscientious, there's still a stray cacheDuration in CsvCatalogItemTraits

https://github.com/TerriaJS/terriajs/blob/97b58a4e309d5ca23f8222563c5ee201a3dd0677/lib/Traits/CsvCatalogItemTraits.ts#L58-L65

I don't think this is a deal breaker for this PR because we can do it later and I'm not sure it's actually used anywhere, but I thought I'd mention it.

nf-s (approved):

PR #4555, opened 2020-07-21
added basic nested WMS Group support

### LINKED ISSUE #3674: Model Architecture: WMS CatalogGroup - nested wms group layers don't show up
If you set up a catalog with the two WMS servers:
https://programs.communications.gov.au/geoserver/ows
https://ows.services.dea.ga.gov.au/

The first will load, the second will load but not show the nested layers
image

nf-s:
### Adds basic support for nested WMS groups

Fixes #3674 

To test: 

- http://ci.terria.io/wms-group-fix/
- add [https://ows.services.dea.ga.gov.au/](https://ows.services.dea.ga.gov.au/) as WMS url

### Checklist


nf-s:
Do I need to handle property inheritance?

The WMS spec states:

>Each available map is advertised by a <Layer> element in the service metadata. Conceptually, each Layer is a
distinct entity. However, as a means of classifying and organizing layers, and as a means of reducing the size of
the service metadata, a single parent Layer may enclose any number of additional layers, which may be
hierarchically nested as desired. Some properties defined in a parent layer are inherited by the children it
encloses. These inherited properties may be either redefined or added to by the child. Subclause 7.2.4.8
summarizes whether or how each property is inherited. 

Or is it handled by `WebMapServiceCapabilities`?  
[It looks like it is](https://github.com/TerriaJS/terriajs/blob/b012d3185e1e3201c54fa1c0bc25e6c1fae3e6b5/lib/Models/WebMapServiceCapabilities.ts#L187)

rowanwins (approved):
Looks and works really good!

rowanwins:
CMAR is another good example with nesting multiple levels deep
````
    {
      "type": "wms-group",
      "url": "https://www.cmar.csiro.au/geoserver/wms",
      "name": "CMAR"
    },
````

PR #4556, opened 2020-07-21
Fix postpublish action when running in Github Action

tephenavies:
As in `next` branch. I forgot to change this when adding the action to `master`


rowanwins (approved):
Looks fine to me!

PR #4564, opened 2020-07-23
Add frontend contributing guide

soyarsauce:
### What this PR does

Partially addresses https://github.com/TerriaJS/terriajs/issues/4470

### Checklist

-   [x] unit tests n/a
-   [x] changelog n/a


soyarsauce:
What this doesn't address:
>Imports
    When to use require and import
    Omitting file extensions (for .ts, .tsx, .js or .jsx)

and the `"Raw[Component]" pseudo-naming-convention for "smart / presentational" components` aspect, because I haven't given that too much thought right now

rowanwins (approved):
Looks good to me!

As a general comment I think if we can include examples of things that have been done well, eg gold star components, that will help people get a feel for examples of what we're working towards.

PR #4569, opened 2020-07-24
Provide base url and optional proxy url for senaps items.

### LINKED ISSUE #4567: Load senaps data via proxy with access control
Currently `senaps.io` data are loaded via `terriajs-server` proxy that does not have access control mechanism. Use `datastore-api` where the proxy has accesss-control logic.

mwu2018:
### What this PR does

Fixes #4567 
This PR removes hard coded senaps base url from SenapsLocationsCatalogItem so that senaps data can be protected by a secured proxy service such as `datastore-api` of version `0.0.57-terria-1` and above.

**Technical Note**
The data service proxied by TerriaJS-Server is inherently public to any users. In the case that a data service must be restricted to certain authenticated users, use an alternative proxy server that enforces required access control.

An example using alternative proxy URL `/api/v0/data/proxy/senaps-locations-as-private-record` to proxy senaps service at `https://senaps.io/api/sensor/v2` follows.

```
{
    "id": "senaps-locations-as-private-record",
    "type": "senaps-locations",
    "definition": {
        "url": "/api/v0/data/proxy/senaps-locations-as-private-record/https://senaps.io/api/sensor/v2",
        "name": "Observations",
        "description": "Demonstrate that the catalog is shown as private and the underlying data also requires private api key.",
        "dataCustodian": "Some custodian",
        "locationIdFilter": "some location id"
    }
}
```

### Checklist


lib/Language/en/translation.json
```diff
@@ -275,7 +275,7 @@
   },
   "welcomeMessage": {
     "title": "Let's get you started",
-    "welcomeMessage": "Interested in data discovery and exploration?<br/>Dive right in and get started or check the following help guide options.",
+    "welcomeMessage": "Welcome to the Queensland Digital Twin.<br/> Access a rich catalogue of 2D, 3D and live data from multiple sources in one convenient location. Take a tour guide or start exploring!",
```
soyarsauce: Revert this change, the QLD terriamap translation script will override this so just watch when you are staging files

mwu2018: I was puzzled by this changes. Now clear.

mwu2018: done

rowanwins:
So I'm wondering if we can do something that is a bit more scalable and not specific to senaps? 

Perhaps we could leverage the [AccessControlMixin](https://github.com/TerriaJS/terriajs/blob/mobx/lib/ModelMixins/AccessControlMixin.ts) and the [Models/proxyCatalogItem](https://github.com/TerriaJS/terriajs/blob/mobx/lib/Models/proxyCatalogItemUrl.js) to determine which proxy we used?
````
if (catalogItem.accessType === 'public' ) {
  proxyUrl = the regular proxy url
} else if (catalogItem.accessType !== 'public') {
   proxyUrl = the auth controlled proxy
}
````
Currently we don't have anything in Terria that captures that secondary proxy url (the auth controlled one), but perhaps we should consider it?

Anyway it's just a thought...

**Comment by mwu2018** [comment]
Yes, making it generic is very desirable. I will investigate this.
If a server provides both public and non-public data, it is better to use the same controlled proxy service. Otherwise the data server must be in the regular web server's proxy whitelist, which may increase the attack surface.

**Comment by rowanwins** [comment]
@mwu2018 I'm wondering if there is a way we can achieve this without having to manually define `proxyUrl` property on items? I'm thinking of the scenario where we want to flick a switch between an item being public or not. Having to add or remove the proxyUrl property seems pretty fiddily.

Perhaps we can modify the `Models/MagdaReference.ts` so that proxyUrl is a computed getter based on the information that we have?

**Comment by mwu2018** [comment]
@rowanwins We do not need to remove proxyUrl property in order to switch an item from non-public to public. Access control is decided by OPA only.
BTW, proxyUrl property is optional. It is not required except for senaps. Please see `Technical Note` in the PR description.

**Comment by soyarsauce** [comment]
With @steve9164 & @mwu2018 to discuss

**Comment by tephenavies** [comment]
Maybe we should add these instead in a `urlOptions` obejct. E.g.

```json
{
  "name": "ItemA",
  "type": "csv",
  "url": "https://example.com/a.csv",
  "urlOptions": {
    "proxy": "/my-special-proxy/",
    "parameters": {
      "accesskey": "secret"
    }
  }
}
```

That would allow us to implement extra url parameters for token across all items (we'd have to implement using parameters for each item, but at least we have a central mechanism to express it for all items).

Thoughts @rowanwins @soyarsauce @mwu2018?

mwu2018:
My thought is that the proxy instead of the client should decide how to get the actual data. For example, the proxy may decide to either append an access token or set basic authorization in the request header. Only the proxy knows about access secrets.

tephenavies:
To move:
```json
{
  "proxyUrl:" "/url-to-proxy/"
}
```
to
```json
{
  "urlOptions": {
    "proxyUrl:" "/url-to-proxy/"
  }
}
```

with the view that we will add further `urlOptions` as needed to generally support access tokens in catalogue.

soyarsauce:
Yeah this can get really messy, but agree that having the in one location will be useful in stopping it from spiraling out too much. It's also noteworthy that we would see 2 use cases of both client-side tokens as well as server-side tokens. Given this PR is to address the senaps issue we shouldn't think too hard on the client-side implementation (as we handle this particular case at the proxy level), I think @steve9164 was just trying to demonstrate some other possible future states.

zoran995:
Just my 2 cents, @mwu2018 proxy can't know always about access secrets, at least it is not possible when doing SSO between multiple application, backend application can't get data for user session without client sending JWT (at least I have never seen such implementation - also JWT token changes over time). 

But that use case is out of the scope of this PR, might raise discussion about it somewhere else.

rowanwins:
Reiterating my preference which I mentioned above I'd kinda like to see us make use of a getter within `MagdaReference`, we should be able to construct the proxy url based on the magda record id. 
Perhaps with this approach we would provide a default value that could be overridden?

mwu2018:
@rowanwins Could you please elaborate a bit more on your preferred approach?

mwu2018:
@zoran995 Thanks for your input. Agree that we should aim at a generic approach for access control but that is out of the scope of this PR. We definitely need to have further discussion on this topic.

mwu2018:
@steve9164 The proxyUrl is now inside urlOptions. I've verified that it works in my local deployment environment.
I need to update instructions though.

tephenavies:
@rowanwins Can you expand on your MagdaReference idea? If it's possible to implement it in MagdaReference that would be better.

mwu2018:
@steve9164 and @rowanwins 
I took another look to the changes and found that all of the special changes for senaps data are unnecessary.  @steve9164 once asked out-of-band why proxy url can not be directly added to the the base url of senaps data, just like any other non-public data and my reply was wrong. In fact it is quite ok. So, this PR does not need `urlOptions` and `proxyUrl` anymore ~only needs to keep the revised unit tests~ and I will make the update.

Hope this update will keep everyone happy.

tephenavies (approved):
Thanks for making those changes @mwu2018. This is now ready to merge.

PR #4577, opened 2020-07-28
Mobx time

soyarsauce:
### What this PR does

Fixes some problems with https://github.com/TerriaJS/qld-digital-twin/issues/170
When loading https://services.aremi.data61.io/aemo/v6/csv/all and traversing between the available times, you can only ever go forward and not backwards. The indexing via `DiscretelyTimeVaryingMixin`s `getDiscreteTimeIndex` gets thrown off by duplicates.
### Checklist


lib/Models/CsvCatalogItem.ts
```diff
@@ -141,7 +143,19 @@ export default class CsvCatalogItem
       | undefined = this.strata.get(
       automaticTableStylesStratumName
     ) as TableAutomaticStylesStratum;
-    return automaticTableStylesStratum?.discreteTimes;
+    return automaticTableStylesStratum?.discreteTimes?.reduce?.(
```
rowanwins: I'm wondering if we need to move this down the tree further so it's applicable for other catalog item types too such as WMS?

lib/Models/CsvCatalogItem.ts
```diff
@@ -141,7 +143,19 @@ export default class CsvCatalogItem
...
+      // is it correct for discrete times to remove duplicates?
+      // duplicates will mess up the indexing problem as our `<DateTimePicker />`
+      // will eliminate duplicates on the UI front, so given the datepicker
+      // expects uniques, return uniques here
```
rowanwins: Perhaps we need a second getter on the `DiscretelyTimeVaryingMixin` so that we have the choice to return a deduplicated array or the original array - that might help make it slightly more explicit as to what the expected behaviour is for each of the respective getters is?

soyarsauce:
@rowanwins I don't think we create a second getter if we don't actually need it - given that `discreteTimes()` is an abstract method that is up to each item to implement, how does it sound if I, for now, move this current `reduce` up to `TableAutomaticStylesStratum` (via https://github.com/TerriaJS/terriajs/pull/4577/commits/f6b7292cafc50f2d106950e911585c61658fc757) - and through a separate proposal, more broadly evaluate & talk through the merits of renaming `discreteTimes` to `uniqueTimes` so it's even clearer (or to leave it, based on below)?

`discreteTimes` in the name itself suggests uniqueness for units of time - especially if one looks at `DiscretelyTimeVaryingMixin` and all the indices related functions, those would rely on needing unique values and wouldn't make sense with duplicate times. And thinking more generally about a given time dimension from a WMS, I'll lend on yours & others expertise for if they would ever give duplicate times, but perhaps for the WMS implementation, a "duplicate time" would never occur and is handled for us at the server level. But with a `TableAutomaticStylesStratum`, especially from a CSV, it makes sense that there are duplicate times in the values itself, but calling a `discreteTimes` getter should deduplicate for you. Now writing all this out makes me think we don't even rename it as this is the natural conclusion once you actually think through it. `:just-a-bit-confused:`

lib/Models/CsvCatalogItem.ts
soyarsauce: see https://github.com/TerriaJS/terriajs/pull/4577#issuecomment-669086838

lib/Models/CsvCatalogItem.ts
soyarsauce: see https://github.com/TerriaJS/terriajs/pull/4577#issuecomment-669086838

rowanwins:
Yeah ok I think you've convinced me in writing that out @soyarsauce . `discreteTimes` does imply unique values so adding a getter for non-unique values probably doesn't make much sense. And I don't think WMS will ever send duplicate values. I think by moving the logic up/across to the `TableAutomaticStylesStratum.ts` you've covered off on the places this is mostly likely to occur in.

rowanwins (approved):
Approved based on changes performed and discussion

PR #4589, opened 2020-07-29
Mobx always breadcrumbs

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4169

Currently breadcrumbs only show when you are doing a data search.
This does 2 things:
- makes them always shown
- fixes it so when you click on its parents, it opens up that item in the catalog

### Checklist

### DIFF
--- a/lib/ReactViews/Search/Breadcrumbs.jsx
+++ b/lib/ReactViews/Search/Breadcrumbs.jsx
@@ -53,7 +53,6 @@ class Breadcrumbs extends React.Component {
       // Note: should it reset the text if a person deletes current search and starts a new search?
       <Box
         left
-        // styledHeight={"32px"}
         styledMinHeight={"32px"}
         fullWidth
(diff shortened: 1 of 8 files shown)


lib/ReactViews/Search/Breadcrumbs.jsx
```diff
@@ -53,47 +53,61 @@ class Breadcrumbs extends React.Component {
       // Note: should it reset the text if a person deletes current search and starts a new search?
       <Box
         left
-        styledHeight={"32px"}
+        // styledHeight={"32px"}
+        styledMinHeight={"32px"}
```
reginapramesti: Great idea! I like the use of min height so the height is adjusted when it wraps to the next line - probs just remove the commented out `styledHeight`?

reginapramesti (approved):
Looks good to me! I'd just remove the commented out stuff, but shouldn't need another round of review 👍

Issue #4218, opened 2020-04-09
Mobx: WFS and WMTS support


zoran995:
Are WFS and WMTS supported on the mobx branch, can't find them? If not do you plan to add them?

nf-s:
Hey @zoran995
I'm looking into adding WFS support over the next week or so. Is there anything specific that you are after? 
Eg tiling, mvt support, filter query support
Thanks!

zoran995:
Hi @nf-s thanks for working on this
Haven't thought much about specific stuff, just looking for basic implementation (never got enough time to implement it). All the features you mentioned in PR would be nice but it won't be easy to implement them especially styling (sld to cesium). 
Are you going with WFS 1.1.0 or 1.3.0? 
If you need some help in implementing this I can try to find some spare time to help.

nf-s:
No problem!
Yeah sld to cesium would be a bit of work!
I am using 1.1.0 at the moment. 
There is a checklist of possible features in the [PR](https://github.com/TerriaJS/terriajs/pull/4591) , currently it is looking like I will just add the clip and ship (export layer by selecting an extent) for the first PR, and then we can discuss further development.

soyarsauce:
Blocks https://github.com/TerriaJS/de-africa-map/issues/54 - required for wmts basemaps in de-africa

nf-s:
Closed by #4720

PR #4591, opened 2020-07-29
Basic wfs/wfs group working

### LINKED ISSUE #4549: MobX: support for WFS
This epic is to consolidate work for enabling support for WFS and ~~Esri Feature Server/Service~~.

nf-s:
### WFS support

Fixes #4549 #4218 

Basic WFS item + group support:

- Only for geojson
- WFS catalog item has geojson style traits

Example WFS items: http://ci.terria.io/wfs/#clean&https://gist.githubusercontent.com/nf-s/b71d4aef0e027c8013136a3c6dc895dc/raw/625b1374570e3f2af10bd8a3e3960807ed15db96/wfs.json

### Future development

- [ ] There is a lot of duplicate code across WMSCatalogItem, WMSCatalogGroup, WFSCatalogItem, WFSCatalogGroup, should I refactor some out?
- [ ] Add tiling support
- [x] Add GML `OutputType`
- [ ] Add mvt `OutputType` support
- [ ] Add geojson-vt support for geojson `OutputType`
- [ ] Add client-side vector styling?
- [ ] Add SLD style support?
- [ ] Support join operations/multiple layers (featureTypes)?
- [ ] Support filters?

### Checklist

- [x] Go through all WFS layers in old maps
- [x] ~Add clip and ship support~ There isn't clip and ship support, but you can download the geojson


KeyboardSounds:
>There is a lot of duplicate code across WMSCatalogItem, WMSCatalogGroup, WFSCatalogItem, WFSCatalogGroup, should I refactor some out?

please

nf-s:
NEII and Kaleidoscope for WFS layers

soyarsauce:
Look in NEII and Kaleidoscope for wfs items

lib/Models/WebFeatureServiceCatalogItem.ts
```diff
@@ -0,0 +1,371 @@
...
+import isDefined from "../Core/isDefined";
+import AsyncMappableMixin from "../ModelMixins/AsyncMappableMixin";
+
+class GetCapabilitiesStratum extends LoadableStratum(
+  WebFeatureServiceCatalogItemTraits
+) {
+  static load(
+    catalogItem: WebFeatureServiceCatalogItem
+  ): Promise<GetCapabilitiesStratum> {
+    console.log("Loading GetCapabilities");
```
rowanwins: remove me

lib/Models/WebFeatureServiceCatalogItem.ts
```diff
@@ -0,0 +1,371 @@
...
+        service &&
+        service.Abstract &&
+        !containsAny(
+          service.Abstract,
+          WebFeatureServiceCatalogItem.abstractsToIgnore
+        ) &&
+        service.Abstract !== firstDataDescription
+      ) {
+        const traits = createStratumInstance(InfoSectionTraits);
+        traits.name = "Web Map Service Description";
```
rowanwins: Don't forget to internationalize some of these names in the info section

lib/Traits/WebFeatureServiceCatalogItemTraits.ts
```diff
@@ -0,0 +1,35 @@
...
+    type: "string",
+    name: "Layer(s)",
+    description: "The layer or layers to display."
+  })
+  layers?: string;
+
+  @anyTrait({
+    name: "Parameters",
+    description:
+      "Additional parameters to pass to the MapServer when requesting images."
```
rowanwins: `MapServer` cut-and-pasta

rowanwins:
I tested with this monster WFS server
http://www.cmar.csiro.au/geoserver/ows?service=wfs&version=1.3.0&request=GetCapabilities

Mostly working great, a couple of smaller things I noticed so far
- Feature picking isn't highlighting the selected feature

- The export tool is exporting `[object object]`

- Splitter doesn't seem to be working

**Note:** that some of those things might not actually be your fault, they might be underlying bugs

splitter

lib/Models/WebFeatureServiceCatalogItem.ts
```diff
@@ -0,0 +1,486 @@
...
+  @computed
+  get canExportData() {
+    return isDefined(this.geojsonCatalogItem?.geoJsonData);
+  }
+
+  async exportData() {
+    if (isDefined(this.geojsonCatalogItem?.geoJsonData)) {
+      return {
+        name: `${this.name} export.json`,
+        file: new Blob([this.geojsonCatalogItem!.geoJsonData.toString()])
```
rowanwins: Perhaps `json.stringify` rather than `.toString()` to fix the export

lib/Traits/WebFeatureServiceCatalogItemTraits.ts
```diff
@@ -0,0 +1,51 @@
...
+  UrlTraits,
+  MappableTraits,
+  CatalogMemberTraits
+) {
+  @primitiveTrait({
+    type: "string",
+    name: "Type Name(s)",
+    description: "The type name or names to display."
+  })
+  typeNames?: string;
```
rowanwins, line 31: Should we allow multiple layers in a single `WebFeatureServiceCatalogItem` do you think? I know this is currently ported from how we do this in Terria master... but should we re-evaluate?
I'm trying to think when it would be helpful to have multiple layers represented as a single layer in Terria...?

lib/Traits/WebFeatureServiceCatalogItemTraits.ts
```diff
@@ -0,0 +1,51 @@
...
+    name: "Parameters",
+    description:
+      "Additional parameters to pass to the WFS Server when requesting features."
+  })
+  parameters?: JsonObject;
+
+  @objectTrait({
+    type: StyleTraits,
+    name: "Style",
+    description: "Styling rules that follow simplestyle-spec"
```
rowanwins: I think we can use markdown to include a link to the simple style spec which might be helpful.

lib/Models/WebFeatureServiceCatalogItem.ts
nf-s: Yep good point! thanks

nf-s:
> I tested with this monster WFS server
> http://www.cmar.csiro.au/geoserver/ows?service=wfs&version=1.3.0&request=GetCapabilities
> 
> Mostly working great, a couple of smaller things I noticed so far
> 
> * Feature picking isn't highlighting the selected feature
> * The export tool is exporting `[object object]`
> * Splitter doesn't seem to be working
> 
> **Note:** that some of those things might not actually be your fault, they might be underlying bugs

Splitter ins't implemented to geojson right?
I have removed `  readonly supportsSplitting = true;` from wfs catalog item.

Highlighting features is working for me - albeit super slowly - sometimes it takes a while for the features to highlight

image

lib/Traits/WebFeatureServiceCatalogItemTraits.ts
nf-s: I would say may as well leave it in, so people could add operations using the `parameters` trait if they really want to.
But yeah, probably not very helpful - I haven't ever seen it used.

lib/Traits/WebFeatureServiceCatalogItemTraits.ts
nf-s: Yep good idea!

rowanwins:
So the highlight things seems to be problematic for lines and points, this is probably due to some handling in the GeoJsonCatalogItem.

rowanwins:
`typeNames` trait doesn't seem to work. It should show a grid off WA, and a bunch of points around PNG
````
        {
            "name": "test",
            "type": "wfs",
            "url": "http://www.cmar.csiro.au/geoserver/ows",
            "typeNames": "nerp:1996-2000-Data_All_SeismicSurveys__2D_2013_09,mnf:SS2012_V05_ADCP"
        }
````

**Comment by nf-s** [comment]
> `typeNames` trait doesn't seem to work. It should show a grid off WA, and a bunch of points around PNG
> 
> ```
>         {
>             "name": "test",
>             "type": "wfs",
>             "url": "http://www.cmar.csiro.au/geoserver/ows",
>             "typeNames": "nerp:1996-2000-Data_All_SeismicSurveys__2D_2013_09,mnf:SS2012_V05_ADCP"
>         }
> ```

This seems to be happen due to the 1000 feature limit. Maybe I should add a warning if we hit that limit?

**Comment by rowanwins** [comment]
Ahhh that would make sense. TBH I wouldn't over engineer it, its a pretty bad user experience for WFS anyway that needs way more thought, so if there is anything simple do it, but if not, dont worry about it!

**Comment by nf-s** [comment]
> 
> 
> Ahhh that would make sense. TBH I wouldn't over engineer it, its a pretty bad user experience for WFS anyway that needs way more thought, so if there is anything simple do it, but if not, dont worry about it!

Ok I'll just add a short report to say something

File: lib/Models/WebFeatureServiceCapabilities.ts [hunk 7]
```diff
@@ -0,0 +1,171 @@
...
+import {
+  CapabilitiesGeographicBoundingBox,
+  CapabilitiesService
+} from "./WebMapServiceCapabilities";
+
+export interface FeatureType {
+  readonly Name?: string;
+  readonly Title: string;
+  readonly Abstract?: string;
+  readonly WGS84BoundingBox?: CapabilitiesGeographicBoundingBox;
```
zoran995, line 19: Coundn't get ideal zoom to work. WFS has different bounding box response
```
interface Wgs84BoundingBox {
  LowerCorner: string;
  UpperCorner: string;
}
```

lib/Models/WebFeatureServiceCapabilities.ts
```diff
@@ -0,0 +1,171 @@
...
+    return service;
+  }
+
+  get featureTypes(): FeatureType[] {
+    const featureTypesJson = this.json.FeatureTypeList?.FeatureType as Array<
+      any
+    >;
+    if (!isDefined(featureTypesJson) || !Array.isArray(featureTypesJson)) {
+      return [];
+    }
```
zoran995: This will cause the server with only one layer to fail (say the group is empty). It is pretty common to have such services when using isolated workspace in geoserver.

zoran995:
Some things I run into during quick tests
- Data can't be hidden from the map
- Is there an option for user to define if data is exportable? I don't see it either for other layers that has export option. If there is no such option maybe add an exportable trait class, to be reused later.

lib/Models/WebFeatureServiceCapabilities.ts
rowanwins: Good pickup @zoran995 

For example this WFS URL
 https://data.gov.au/geoserver/australia-s-indigenous-forest-estate-2018/wfs

nf-s: Thanks @zoran995 , all fixed!

nf-s: @rowanwins that URL doesn't have any <FeatureType>, but I tested with my own Geoserver and it seems to work fine

rowanwins:
@nf-s 
Got an error adding trying to add one of the layers underneath this
https://mapping.launceston.tas.gov.au/arcgis/services/Public/Transport/MapServer/WFSServer

````
TypeError: getCapabilitiesStratum.outputFormats.find is not a function
````
Note this is WFS from an ArcGIS server

lib/Models/WebFeatureServiceCapabilities.ts
rowanwins: Hmm weird, looks like the data.gov.au WFS thing might be broken...

nf-s:
I have fixed `show` for WFS Catalog item. 
Change `ExportableData` interface to `Mixin` + add `disableExport` trait.
@rowanwins I have fixed that, but that layer only has gml and currently it only supports geojson.
I'll add gml using that `gmlToGeojson` function

rowanwins (approved):
Looks awesome @nf-s !

PR #4592, opened 2020-07-30
Alt proxy url

mwu2018:
### What this PR does

### Checklist


rowanwins:
@mwu2018 I'm wondering if there is a way we can achieve this without having to manually define `proxyUrl` property on items? I'm thinking of the scenario where we want to flick a switch between an item being public or not. Having to add or remove the `proxyUrl` property seems pretty fiddily. 

Perhaps we can modify the `Models/MagdaReference.ts` so that `proxyUrl` is a computed getter based on the information that we have?

mwu2018:
@rowanwins Sorry, this is a draft PR only for myself. The correct one is https://github.com/TerriaJS/terriajs/pull/4569.

Issue #4563, opened 2020-07-23
Version8: Problem with Gulp and Windows paths


hardreddata:
I thought I would take a look at a clean clone of the _next_ branch which I cloned to `D:\dev\TerriaNext` on my Windows 10 PC.

The `npm install` worked. Shown below is an extract of the Gulp output. Notice the `Module not found: Error: Can't resolve 'D:devTerriaNext`, and the following `\newline` and `\tab` from **N**ode and **T**erria. 

I am using Git Bash.

```
...
 f5ad078d6d4ff7fc08ef6a3105dc2ef9.jpg   24.9 KiB                 [emitted]
f65ca6d17be7dd1611465960983a262b.json     64 KiB                 [emitted]
facde137ccef1e22b31b75bb0db0c985.json   64.3 KiB                 [emitted]
                  index.TerriaMap.css   48.2 KiB          index  [emitted]  index
                   index.TerriaMap.js    110 KiB          index  [emitted]  index
          vendors~index.TerriaMap.css    327 KiB  vendors~index  [emitted]  vendors~index
           vendors~index.TerriaMap.js   43.7 MiB  vendors~index  [emitted]  vendors~index
Entrypoint main = TerriaMap.js

ERROR in ./node_modules/terriajs/lib/Map/CesiumTileLayer.ts
Module not found: Error: Can't resolve 'D:devTerriaNext
ode_modules     erriajslib/Core/superGet' in 'D:\dev\TerriaNext\node_modules\terriajs\lib\Map'
 @ ./node_modules/terriajs/lib/Map/CesiumTileLayer.ts 104:11-79
 @ ./node_modules/terriajs/lib/Models/Leaflet.ts
 @ ./node_modules/terriajs/lib/ViewModels/TerriaViewer.ts
 @ ./node_modules/terriajs/lib/Models/Terria.ts
 @ ./index.js
 @ ./entry.js

ERROR in ./node_modules/terriajs/lib/ModelMixins/Cesium3dTilesMixin.ts
Module not found: Error: Can't resolve 'D:devTerriaNext
ode_modules     erriajslib/Core/superGet' in 'D:\dev\TerriaNext\node_modules\terriajs\lib\ModelMixins'
 @ ./node_modules/terriajs/lib/ModelMixins/Cesium3dTilesMixin.ts 154:11-79
 @ ./node_modules/terriajs/lib/Models/Cesium3DTilesCatalogItem.ts
[...]

PR #4605, opened 2020-07-30
Version 8: Replace gov.au/page-alerts with a handmade WarningBox

### LINKED ISSUE #4214: Remove pancake and gov-au as dependencies of TerriaJS mobx
`@gov.au/page-alerts` is used in NSWDT for one alert. We should decide either to use `@gov.au` design system more widely or remake this alert with simple divs.

Another solution would be to remove the NSWDT-custom things in TerriaJS and replace them with places where apps can inject their own views.

tephenavies:
### What this PR does

Fixes #4214

Removes blocker to solving #4563.

Add a WarningBox to replace `@gov.au/page-alerts`. This will remove dependency on the `pancake` pre-build system which sometimes caused problems on `yarn`/`npm install` and probably slowed down dependency installation.

### Checklist


nf-s (approved):

nf-s:
This PR makes me happy

PR #4613, opened 2020-08-04
Internationalize help content - extended

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/3964, https://github.com/TerriaJS/terriajs/issues/4278
Follows on from https://github.com/TerriaJS/terriajs/pull/4523

This further extends the work done in https://github.com/TerriaJS/terriajs/pull/4523 such that flexibility required to configure TerriaMap to load translations from, magda for instance, will exist.

### Checklist


lib/Models/Internationalization.ts
```diff
@@ -48,11 +82,15 @@ class Internationalization {
...
         debug: languageConfig.debug,
+        react: languageConfig.react,
         fallbackLng: languageConfig.fallbackLanguage,
         whitelist: Object.keys(languageConfig.languages),
 
         // to allow en-US when only en is on the whitelist - nonExplicitWhitelist must be set to true
-        nonExplicitWhitelist: true,
+        // nonExplicitWhitelist: true,
+        // deprecated, it's now the inverse
+        nonExplicitSupportedLngs: false,
```
soyarsauce: @zoran995 nonExplicitWhitelist is now deprecated, is it right to do `nonExplicitSupportedLngs` as `false` (which is the default anyway) to get the same behaviour?

zoran995: also `whitelist` [L87](https://github.com/TerriaJS/terriajs/pull/4613/files#diff-89ed42d66d62ba7736eb980b2e4778a7R87) was changed to `supportedLanguages`, there is no much documentation about migration to the new definition, going to test this today

zoran995: from my understanding it is not inverse, `nonExplicitWhitelist` was also default to `false`. On my end everything works fine with both settings.

lib/Core/injectTerms.ts
```diff
@@ -71,14 +77,28 @@ const injectTerms = (string: string, termDictionary: Term[]): string => {
...
+
     termDictionary.forEach(term => {
-      term.aliases?.forEach(alias => {
-        tooltipTerms.set(alias.toLowerCase(), term);
-      });
+      const aliasesTranslated = term.aliases
+        ? i18next.exists(term?.aliases)
+          ? i18next.t(term?.aliases, { returnObjects: true })
+          : term?.aliases
+        : [];
```
zoran995: When the `helpContentTerms` reference value in override file it will throw error if they doesn't exist in that file. In that case the value is not an array but a string
override error

```suggestion
      const aliasesTranslated = term.aliases
        ? i18next.exists(term.aliases)
          ? <Array<string>>i18next.t(term?.aliases, { returnObjects: true })
          : Array.isArray(term.aliases.slice())
          ? term.aliases
          : [term.aliases]
        : [];
```

zoran995 (changes requested):
Added few suggestion, everything else work fine on my end.

lib/Models/Internationalization.ts
soyarsauce: Yes you're right it should be `true` here - also in my mind I conflated the config surrounding this a little bit with `load: "languageOnly"` - but upon looking at the source in i18next this confirms that they both intend to do similar things:
```
    if (this.options.load === 'languageOnly' || this.options.nonExplicitSupportedLngs) {
      code = this.getLanguagePartFromCode(code);
    }
```
https://github.com/i18next/i18next/blob/80a38100d21a7e7c1f9cb2acff5f709063027b9f/src/LanguageUtils.js#L78-L80

lib/Core/injectTerms.ts
soyarsauce: Thanks good pickup, I ran into this as well when testing out overrides but hadn't gone to fix it

nf-s (approved):

CHANGES.md
```diff
@@ -10,6 +10,10 @@ Change Log
 
 #### next release (8.0.0-alpha.44)
 * Pass `format` trait on `TableColumnTraits` down to `TableAutomaticStylesStratum` for generating legends
+* Added an option for translation override from TerriaMap
+* Help content, trainer bar & help terms can use translation overrides
+* Accepts `backend` options under a new `terria.start()` property, `i18nOptions`
+* [The next improvement]
```
reginapramesti: Need to move this to the next release? I think we've already release 8.0.0-alpha.44 right?

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
```diff
@@ -102,8 +102,16 @@ class HelpPanelItem extends React.Component {
           itemString={this.props.content.itemName}
           paneMode={this.props.content.paneMode}
           markdownContent={this.props.content.markdownText}
-          videoUrl={this.props.content.videoUrl}
-          placeholderImage={this.props.content.placeholderImage}
+          videoUrl={
+            this.props.content.videoUrl
+              ? t(this.props.content.videoUrl)
+              : this.props.content.videoUrl
```
reginapramesti, line 108: I think it would make sense to use `i18n.exists(this.props.content.videoUrl)` here like in`HelpVideoPanel.jsx` for example. Or is there a reason it's not used in this case? I'm just thinking for cases where `this.props.content.videoUrl` exists but the translation of it doesn't

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
```diff
@@ -102,8 +102,16 @@ class HelpPanelItem extends React.Component {
...
+          }
+          placeholderImage={
+            this.props.content.placeholderImage
+              ? t(this.props.content.placeholderImage)
+              : this.props.content.placeholderImage
+          }
```
reginapramesti, line 114: Same as above

reginapramesti (changes requested):
Looks really really good thanks for your work @soyarsauce @zoran995 :) just need to clarify a few things before we merge

lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
```diff
@@ -105,7 +105,7 @@ class HelpPanel extends React.Component {
                 ${p => p.theme.addTerriaPrimaryBtnStyles(p)}
               `}
             >
-              {"Take the tour"}
+              {t("helpPanel.takeTour")}
```
reginapramesti, line 108: Also here, is there a reason why we chose to not do a `i18n.exists` check? I might be missing something but just wanted to clarify this to make sure I understand everything. Is the translation always guaranteed to be there?

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
reginapramesti: Oh wait I'm an idiot I just realised this is a link/url. In which case do we need translations for them?

zoran995: This enables users to provide a different help video made for different languages (although there could be cc provide for video). 

`i18next.exists` is needed in two cases (if I recall correctly):
- when loading an objects from translation (arrays)
- when dealing with larger text that can get into translation function. Default behavior of `t` function should be to put a translation text if there is no translation, but when there are larger or quiet complex text it happens to be trimmed, so we use `i18next.exists` to check if that string exists in any of translation files.
 
just noticed a few places that we can add `i18next.exists` to avoid second issue , `markdown` in `StyledHtml`.

lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
zoran995: this was just a string that wasn't i18ified, `i18next.exists` is needed only for configurable translation (loading from config files) terms so we avoid two issues mentioned in previous comment.

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
reginapramesti: Ah that's fair thanks for clarifying!

soyarsauce: Yes that's right, I've addressed & simplified this in https://github.com/TerriaJS/terriajs/pull/4613/commits/434f49af1c5b3f7f1833b572a4789da8ee597ae2 & https://github.com/TerriaJS/terriajs/pull/4613/commits/880900543647021301246e164540db8ed06e3c77 now

soyarsauce: & actually more importantly doc update so this info isn't lost to just this PR https://github.com/TerriaJS/terriajs/pull/4613/commits/782bde028ee82ac33b69db615d61ff4ed26ae2a5

reginapramesti (approved):
All the changes look good to me!

PR #4657, opened 2020-08-20
Mobx i18n magda

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/pull/4613 breaking magda based configs

### Checklist

-   [x] ~There are unit tests to verify my changes are correct or~ unit tests aren't applicable - no start suite


soyarsauce:
I'm just doing a few more tests for this one

lib/Models/Terria.ts
```diff
@@ -404,15 +404,19 @@ export default class Terria {
...
+              options.i18nOptions
+            );
           }
 
-          Internationalization.initLanguage(
-            config.parameters.languageConfiguration,
-            options.i18nOptions
-          );
-
           if (config.aspects) {
```
mwu2018, line 407: Should `config.aspects` be tested before `config.parameters`? There is no logical issue though.

soyarsauce: You're right in that there isn't a logical issue - if it's a magda config, `config.parameters` will never exist - however I've clarified that in https://github.com/TerriaJS/terriajs/pull/4657/commits/f5399c8f3c69f8d2903f944eca300dd0cd7d58b3 now

soyarsauce:
@mwu2018 could you do an explicit approval so we can merge it and do a new release as soon as CI passes? thanks 🙏

mwu2018 (approved):

PR #4677, opened 2020-08-27
Add sortMembers method to GroupMixin

rowanwins:
### What this PR does

I needed a way to sort members of a catalog group **after** they'd be added to a group, this is because you don't always have all the members of a group prior to adding them. 

So this PR add a `sortMembers` method to the `GroupMixin` which accepts a sorting function.

I also implemented the sorting of `ArcGisPortalGroup`, it can be tested with a config such as 
````
    {
      "type": "arcgis-portal-group",
      "name": "All open Content",
      "url": "https://portal.spatial.nsw.gov.au/portal",
      "groupBy": "portalCategories"
    },
````
Where you will notice that the contents of all the sub-groups are sorted by their title.

### Checklist

### DIFF
--- a/lib/ModelMixins/GroupMixin.ts
+++ b/lib/ModelMixins/GroupMixin.ts
@@ -50,6 +50,24 @@ function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
     }
 
     @action
+    sortMembers(
+      stratumId: string,
+      sortFunction: (a: BaseModel, b: BaseModel) => number
+    ) {
+      const members: any = this.getTrait(stratumId, "members");
+      const models = this.memberModels as Array<BaseModel>;
+      if (members === undefined || models.length === 0) return;
+      const unsortedModels = models.splice(0);
+      members.clear();
+      const sortedModels = filterOutUndefined(
+        unsortedModels.sort(sortFunction)
+      );
+      sortedModels.forEach((m: BaseModel) => {
+        this.add(stratumId, m);
+      }, this);
+    }
+
+    @action
     toggleOpen(stratumId: string) {
       this.setTrait(stratumId, "isOpen", !this.isOpen);
     }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@ Change Log
 ### MobX Development
 
 #### next release (8.0.0-alpha.47)
-* [The next improvement]
+* Ad `sortMembers` method to `GroupMixin`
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by nf-s** [comment]
This seems related to our ordering of objectArrayTraits across strata - https://github.com/TerriaJS/terriajs/issues/4559
I think we should try to kill two birds with one stone if possible.

I think you would want to sort members across all strata right? Not just definition stratum.

File: lib/ModelMixins/GroupMixin.ts [hunk 1]
```diff
@@ -49,6 +49,24 @@ function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
...
+    @action
+    sortMembers(
+      stratumId: string,
+      sortFunction: (a: BaseModel, b: BaseModel) => number
+    ) {
+      const members: any = this.getTrait(stratumId, "members");
+      const models = this.memberModels as Array<BaseModel>;
+      if (members === undefined || models.length === 0) return;
+      const unsortedModels = models.splice(0);
+      members.clear();
```
nf-s, line 61: Not sure why `this.getTrait(stratumId, "members")` isn't giving correct type with the `clear` method on it - I'll have a look into it

rowanwins: Yeah that annoyed me too, hence going the `any` approach.

rowanwins:
You might be right @nf-s , good pickup linking the issues.

TBH I'm not sure if I've got a good enough grasp on how that might work across strata. Do @soyarsauce or @steve9164 want to chime in on this PR?

soyarsauce:
Essentially you'd have to determine how to resolve order preference after the  https://github.com/TerriaJS/terriajs/issues/4559 is figured out.

For this PR specifically, it seems you'd only want to sort members once all members strata are resolved (e.g. additions / removals, then apply this sort preference), not at any individual stratum level.

tephenavies:
Maybe the members should be exposed through a computed in an automatic stratum? Or define a computed `get members()` function on the Group that either merges super.members into it's own member before returning a sorted array of members, or similar.

nf-s:
@steve9164 what did you think of my idea of adding a `sortByProperty` to the actual `ObjectArrayTraits` class?
-> https://github.com/TerriaJS/terriajs/issues/4559#issuecomment-681614101

rowanwins:
I'm going to close this as I've resolved the pressing case elsewhere and it seems like this isn't likely to end approach of sorting across strata.

PR #4685, opened 2020-08-29
Unify info

### LINKED ISSUE #4684: Replace many createStratumInstance calls that don't use the second initialisation paramter
The following pattern is often used in MobX computeds:

https://github.com/TerriaJS/terriajs/blob/7d0f41c1aaf19b679361ed18971e5bd2cc05d2ba/lib/Models/ArcGisFeatureServerCatalogItem.ts#L312-L319

We can avoid calling `runInAction` inside computeds (or using helper functions that do the same like `newInfo` and `createInfoSection`) by using the second parameter of `createStratumInstance`, which allows an object of initial values to be passed:

https://github.com/TerriaJS/terriajs/blob/7d0f41c1aaf19b679361ed18971e5bd2cc05d2ba/lib/Models/createStratumInstance.ts#L12-L3

Calling `createStratumInstance` should be preferred over small utility functions

tephenavies:
### What this PR does

Fixes #4684

Unify all info section creation using `createStratumInstance` rather than `createInfoSection` or `newInfo`.

Also I have run `Organize imports` on many of the files touched as many had unused imports. I should have put them into a separate commit from my changes, but it's only import lines affected so not too bad. There's a couple of prettier updates that have also made their way in, but nothing substantial.

### Checklist

-   [x] No unit tests as this is a refactor without change in functionality
-   [x] I don't think this needs a CHANGES.md entry


lib/Models/ArcGisPortalItemReference.ts
```diff
@@ -136,37 +136,28 @@ export class ArcGisPortalItemStratum extends LoadableStratum(
...
-      newInfo(
-        i18next.t("models.arcgisPortal.openInPortal"),
-        `<a href="${this.portalItemUrl}"><button>${i18next.t(
+      createStratumInstance(InfoSectionTraits, {
+        name: i18next.t("models.arcgisPortal.openInPortal"),
+        content: `<a href="${this.portalItemUrl}"><button>${i18next.t(
           "models.arcgisPortal.openInPortal"
         )}</button></a>`
-      )
+      })
```
tephenavies: Does this also need to be sanitised? It doesn't look like we could ever get "style" or "font" attributes/tags. And everything in traits should be sanitised before being used anyway. (See below/above comment on how it used to be sanitized)

lib/Models/ArcGisPortalItemReference.ts
```diff
@@ -136,37 +136,28 @@ export class ArcGisPortalItemStratum extends LoadableStratum(
...
-      runInAction(() => {
-        traits.name = name;
-        // traits.content = content;
-        traits.content = DOMPurify.sanitize(content, {
-          FORBID_ATTR: ["style"],
-          FORBID_TAGS: ["font"]
-        });
-      });
-      return traits;
-    }
```
tephenavies, line 150: `newInfo` used to call `DOMPurify.sanitize`

lib/Models/ArcGisPortalItemReference.ts
tephenavies: @rowanwins do you know the answer to this? Once I clear this up the PR is good to merge.

rowanwins: No sanitzation require here, we only need to do it on the portal item `descriptions` which comes as horrendous html.

rowanwins (approved):
Looks good!

PR #4691, opened 2020-09-01
Ports chartColor trait for DiscretelyTimeVarying items from master.

na9da:
### What this PR does

Ports the `chartColor` trait for DiscretelyTimeVarying items from master.

### Testing

You can use the following config for testing. Charts have become really slow for some reason, you will be able to see the effect of setting `chartColor` but interacting with the chart itself will be painful.

```
{
          "type": "wms",
          "name": "Daily Landsat and Sentinel-2 satellite images",
          "opacity": 1,
          "chartColor": "yellow",
          "timeFilterPropertyName": "data_available_for_dates",
          "chartType": "momentPoints"
}
```

### Checklist

### DIFF
--- a/lib/Traits/DiscretelyTimeVaryingTraits.ts
+++ b/lib/Traits/DiscretelyTimeVaryingTraits.ts
@@ -47,6 +47,13 @@ export default class DiscretelyTimeVaryingTraits extends mixTraits(
   chartDisclaimer?: string;
 
   @primitiveTrait({
+    type: "string",
+    name: "Chart color",
+    description: "The color to use when the data set is displayed on the chart"
+  })
+  chartColor?: string;
+
+  @primitiveTrait({
     type: "boolean",
     name: "Disable date time selector",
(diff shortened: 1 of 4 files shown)


lib/Traits/DiscretelyTimeVaryingTraits.ts
```diff
@@ -46,6 +46,13 @@ export default class DiscretelyTimeVaryingTraits extends mixTraits(
   })
   chartDisclaimer?: string;
 
+  @primitiveTrait({
+    type: "string",
+    name: "Chart color",
+    description: "The color to use when the data set is displayed on the chart"
```
rowanwins: Are we expecting any type of value here? eg hex string or rgb string? Perhaps we can just make it clearer?

na9da: Any html color value should work. I'll add that to the description.

rowanwins (approved):
👍

PR #4696, opened 2020-09-02
Chart cursor perf

na9da:
### What this PR does

Slightly improves chart cursor performance by preventing re-rendering of the x-axes on mouse moves.
Issue: https://github.com/TerriaJS/terriajs/issues/4663

### Checklist


lib/ReactViews/Custom/Chart/BottomDockChart.jsx
```diff
@@ -402,6 +396,12 @@ class Plot extends React.Component {
 }
 
 class XAxis extends React.PureComponent {
```
na9da, line 398: Accepting only simple types as props so that `shouldComponentUpdate` of the PureComponent works as expected.

nf-s:
This seems to have fixed it for me, thanks @na9da !

nf-s (approved):

PR #4700, opened 2020-09-02
add warning for IE users

### LINKED ISSUE #4699: Warn users that we're ending IE support
As per https://github.com/TerriaJS/terriajs/blob/next/architecture/0003-drop-ie11-support.md we're dropping support for IE.

- [ ] Show a message warning IE users that IE support is ending on 1 November 2020

KeyboardSounds:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4699

If `showIEMessage` is `true` in config.json, warns IE11 users that support is ending.

### Checklist

-   [X] n/a There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 
 ### Next Release
 
+* If `showIEMessage` is `true` in config.json, warn IE11 users that support is ending.
 
 ### v7.11.10
 
--- a/lib/ReactViews/InternetExplorerOverlay/InternetExplorerOverlay.jsx
+++ b/lib/ReactViews/InternetExplorerOverlay/InternetExplorerOverlay.jsx
@@ -1,0 +1,175 @@
+import React, { useState } from "react";
+import createReactClass from "create-react-class";
+import PropTypes from "prop-types";
+import classNames from "classnames";
+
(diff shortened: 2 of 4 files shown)


KeyboardSounds:
Note that this won't show up in the CI deployment because it requires setting something in config.json

soyarsauce:
Really great work @KeyboardSounds - just tweaked copy to not force 2 options & the tertiary style so it would actually load https://github.com/TerriaJS/terriajs/pull/4700/commits/a889618d5e7c0481c3c30e68acc57dc0b60ba7d8#diff-70a21a5121a3ab2d8c6ac7a1cef0f332L156 (see casing) and I munged it even further because we rarely use the tertiary styles in v7 - but it will tide us over.

soyarsauce:
merging when ci is good then let's get it on some maps

KeyboardSounds:
> just tweaked copy to not force 2 options & the tertiary style so it would actually load a889618#diff-70a21a5121a3ab2d8c6ac7a1cef0f332L156

Oh, nice catch! thank you :)

soyarsauce (approved):

PR #4701, opened 2020-09-02
Fix styling of WFS and GeoRSS

### LINKED ISSUE #4682: WFS always same color
All WFS layer features will get the same color. For example, all polygons have #93aa00 as `fillColor`. 
This is happening because `GeoJsonCatalogItem`, when generated from WFS, will have an undefined name so `defaultColor` won't do proper randomization of colors. Much better UX would be that features from different WFS layers get a different color.

zoran995:
### What this PR does

Fixes #4682

Sets the name of GeojsonCatalogItem to the `sourceReference` name if the name is not defined. The name is used to choose default styling if a style is not defined some other way.

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable - Small change


nf-s:
Looks good, thanks @zoran995 !

nf-s (approved):

PR #4702, opened 2020-09-02
Fix issue with CSV retireved from THREDDS

### LINKED ISSUE #4665: Cannot display timeseries on click with nc file
Migrated from TerriaMap
https://github.com/TerriaJS/TerriaMap/issues/474
> We have some climate data served using ncWMS protocol (using THREDDS) and also a version using Geoserver (ncWMS plugin), we would like to display the time series graph for a specific point on click on the map. Is this feature available and is there some conf we can take as an example?
Many thanks

cc @davidedelerma

davidedelerma:
### What this PR does

Fixes #4665 

header of CSV returned from THREDDS contain `"` caracter that are
seen as "delimiter" by default and cause the csv parsing to crash.
Extensive descritpion and related conversation can be seen looking at the issue #4665 

### Checklist

I didn't do any change in functionalities or modified any function. I just passed a different delimiter as input to the parser function.

### DIFF
--- a/lib/Map/TableStructure.js
+++ b/lib/Map/TableStructure.js
@@ -394,6 +394,7 @@ TableStructure.fromCsv = function(csvString, result) {
     csvString += "\r\n";
   }
   var json = csv.toArrays(csvString, {
+    delimiter: "|",
     onParseValue: castToScalar
   });
   // Remove any blank lines. Completely blank lines come back as [null]; lines with no entries as [null, null, ..., null].


lib/Map/TableStructure.js
```diff
@@ -394,6 +394,7 @@ TableStructure.fromCsv = function(csvString, result) {
     csvString += "\r\n";
   }
   var json = csv.toArrays(csvString, {
+    delimiter: "|",
```
rowanwins, line 397: Hi @davidedelerma 

We need to be careful about hardcoding values that might potentially break parsing of other csv's. 

My preference would be to see this added as a configurable option on the `CsvCatalogItem` which get's passed down to here. And the default should be the current behaviour, rather than the `|`.

Hope that makes sense

rowanwins (changes requested):

PR #4712, opened 2020-09-03
Fix remove all bug

### LINKED ISSUE #4342: Remove all doesn't remove all items from map
Repro steps:

* Add two vector items (or open this link http://ci.terria.io/mobx/#share=s-wRuUNPqGylt0zdxlPf2OulrqIGg)
* Click remove all
* All the items get removed from the workbench, but on of them still shows on the map

reginapramesti:
### What this PR does

Fixes #4342

In the `observeModelLayer` (in Cesium.ts and Leaflet.ts), the for loop is removing catalogue items from the globe without updating the indices in the process so some of the items are skipped in the process, and this PR fixes that

Tried on the share links posted on the issues page:
* http://ci.terria.io/fix-remove-all-bug/#share=s-wRuUNPqGylt0zdxlPf2OulrqIGg
* http://ci.terria.io/fix-remove-all-bug/#share=s-8JkiUGTmZw64CwrkOCrgPp1EnEb

### Checklist

-   [x] It's a small fix and I imagine there would already be unit tests that already cover that 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Change Log
 * Replaced all instances of `createInfoSection` and `newInfo` with calls to `createStratumInstance` using an initialisation object.
 * Fix styling of WFS and GeoRSS.
 * Fixed a bug that caused re-rendering of xAxis of charts on mouse move. Chart cursor should be somewhat faster as a result of this fix.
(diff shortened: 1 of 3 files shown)


rowanwins (approved):
I like!

PR #4713, opened 2020-09-04
Region mapping time-support

### LINKED ISSUE #4674: Region mapping not working with time-dimension
Test CSV: https://gist.githubusercontent.com/nf-s/401a821abfbc106f93e191058cea9d86/raw/fec78ec7aae650982104bcdb29812c226723f378/region-time.csv

http://ci.terria.io/next/#share=s-kIR3aIAXmLW8FuUWA9UrC2V8knH

nf-s:
### Region mapping time-support

Fixes #4674 

### Checklist


rowanwins:
Tested with this CSV
https://gist.githubusercontent.com/rowanwins/1884c2df4d2985868408ef5c929de802/raw/c908b3fcad80acdfcc024cfc84b4bcdc82d784fb/sample.csv

But not quite there. It seemed to work moving from the first time to the second time, but then it wouldn't go from the second to the third...

Screen_2020_09_04

nf-s:
Yep I noticed that - i think I know how to fix it!

nf-s:
Fixed the last timeInterval issue

lib/ModelMixins/TableMixin.ts
```diff
@@ -496,6 +538,69 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
           .colorMap;
         const valuesAsRegions = regionColumn.valuesAsRegions;
 
+        let currentTimeRows: number[];
+
+        // TODO: this is already implemented in RegionProvider.prototype.mapRegionsToIndicesInto, but regionTypes require "loading" for this to work. I think the whole RegionProvider thing needs to be re-done in TypeScript at some point and then we can move stuff into that.
```
nf-s, line 543: This sums up how I feel about this PR. 
At some point we need to sort out all of the `RegionProvider...` JavaScript stuff - which is mostly unused at the moment.
It would be nice if we could separate region-mapping complexity from `TableMixin`

nf-s: See #4715 for more info about this

rowanwins (approved):
- Works with a time-enabled CSV. 
- Works with a time-enabled csv that only has a single entry.
- Works with time-enabled csv that has multiple region with different dates available
- Works with a csv that doesn't have time

PR #4714, opened 2020-09-04
add short report for esri feature server when exceeding  limit

rowanwins:
### What this PR does

This PR adds a short report to esri-featureServer items if the response only contains a portion of the available features. By default most esri feature services only return 1000 features.

See this example which has one feature with the warning, and one without
http://ci.terria.io/feature-server-limit/#share=s-tMnlk8spb2igWcPeaSfjwzapL2f

Screen_2020_09_04

### Checklist


nf-s (approved):
looks good!

PR #4716, opened 2020-09-04
change region-map current time to discrete date

nf-s:
### What this PR does

One-line change for previous PR #4713 
This changes currentTime to a discrete time - which reduces the number of times the transformer is used


rowanwins (approved):
Approved

PR #4720, opened 2020-09-07
Implement WMTS service

### LINKED ISSUE #1573: Problem accessing ESRI WMTS
When I try to add this dataset
https://www.arcgis.com/home/item.html?id=af0614c88e0d4cad9a72115a944b37ff
using the WMTS endpoint at
http://gis.ngdc.noaa.gov/arcgis/rest/services/bag_hillshades/ImageServer/WMTS/1.0.0/WMTSCapabilities.xml

I get no errors, but nothing appears.   When I check for issues in developer tools, I find: 

```
Unusable ArcGIS Mapserver layers metadata: 
[object Object](anonymous function) @ ArcGisMapServerCatalogItem.js:226
createCatalogItemFromUrl.js:27RuntimeError
```

### LINKED ISSUE #4218: Mobx: WFS and WMTS support

zoran995:
### What this PR does

Ports WMTS implementation to `next` branch with few additions that seemed necessary (minLevel, use of ResourceUrl so Restful encoding can be used). Global tracking issue #4218. This should also fix #1573 and #2927 if I haven't missed something.

### Checklist


lib/Models/WebMapTileServiceCatalogItem.ts
```diff
@@ -0,0 +1,688 @@
...
+        continue;
+      }
+
+      const levelZeroScaleDenominator = 559082264.0287178; // from WMTS 1.0.0 spec section E.4.
+      if (
+        Math.abs(levelZeroMatrix.ScaleDenominator - levelZeroScaleDenominator) >
+        1
+      ) {
+        continue;
+      } */
```
zoran995: @nf-s what do you think about this, is this part necessary, I changed `imageryProvider` to use provided `tileWidth` and `tileHeight`. Also, it can happen that the list of matrices doesn't contain levelZeroMatrix (it start from 10 for example)

nf-s: I don't much experience working with tile-sets like this (or the WMTS spec) - but this seems reasonable to me!
Let me have a look into it and I'll get back to you

nf-s:
Note to self - testing URLs:
- https://a.s2maps-tiles.eu/wmts/?service=WMTS&version=1.0.0&request=GetCapabilities
- https://github.com/TerriaJS/drought-map/blob/7328e4849cb7afe12854674e86cdb3ac52f19705/wwwroot/init/terria.json
- http://geoserver.nationalmap.nicta.com.au/region_map/gwc/service/wmts?REQUEST=GetCapabilities
- https://gibs.earthdata.nasa.gov/wmts/epsg3857/best/wmts.cgi
- https://services.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/WMTS/1.0.0/WMTSCapabilities.xml
- https://services.arcgisonline.com/arcgis/rest/services/Reference/World_Boundaries_and_Places/MapServer/WMTS/1.0.0/WMTSCapabilities.xml

zoran995:
Updated to better handle SRS definitions, there are many of them possible in web mercator projection. Should we add some more, check alternative codes [epsg:3857](https://epsg.io/3857)?

lib/Models/WebMapTileServiceCatalogItem.ts
nf-s: Yep I think you're right!

nf-s:
> 
> 
> Updated to better handle SRS definitions, there are many of them possible in web mercator projection. Should we add some more, check alternative codes [epsg:3857](https://epsg.io/3857)?

The new ones you added seem to cover every layer I threw at it!

nf-s:
Hey @zoran995 - I made a few changes (I hope you don't mind!)
Please let me know if you aren't happy with any of the changes - or if I have broken anything!

My changes:
- There was an issue with selecting the `baseUrl` from `ResourceURLs` in `get imageryProvider`
- Fixed legend and style - for some reason the `LegendURL` is different between WMS/WMTS ?!
- Removed a few console.logs and names copied from WMS/WFS

lib/Models/WebMapTileServiceCatalogItem.ts
```diff
@@ -0,0 +1,657 @@
...
+        tileMatrixSetLinks = [layer.TileMatrixSetLink];
+      }
+    }
+
+    let tileMatrixSetId: string =
+      "urn:ogc:def:wkss:OGC:1.0:GoogleMapsCompatible";
+    let maxLevel: number = 0;
+    let minLevel: number = 0;
+    let tileWidth: number = 256;
+    let tileHeight: number = 256;
```
nf-s, line 612: If a layer is added which doesn't support WebMercator, Terria will crash when it is added to the map. My guess is it has something to do with these default values.
Would it be worth throwing an error if no usable tile matrix can be found - or do you need these defaults for certain WMTS servers?

zoran995: it was done this way in master, I tested with few layers that support only GeographicTilingScheme and Terria didn't crash there was just no layer on the map (although there was no error that it failed to load). For me, it would be much better to remove those defaults and throw an error (just maybe drop a comment in code that there were defaults if we run into a server that by default works with the above matrix set and doesn't return it - never saw one indeed).

zoran995:
all changes look fine to me, just need to double-check the GetFeatureInfo part for some reason it calls 
```
http://127.0.0.1:3001/?service=WMTS&version=1.0.0&request=GetFeatureInfo&tilematrix=EPSG%3A900913%3A6&layer=region_map%3AFID_SUA_2011_AUST&style=&tilerow=37&tilecol=57&tilematrixset=EPSG%3A900913&format=image%2Fpng&i=86&j=24&infoFormat=text%2Fhtml
```
instead of using the WMTSServer url

zoran995:
GetFeatureInfo part is in `terriajs-cesium` creating a pull request for it.

lib/Models/WebMapTileServiceCatalogItem.ts
```diff
@@ -0,0 +1,710 @@
...
+        found = true;
+        tileMatrixSetId = tileMatrixSet;
+        tileMatrixSetLabels = usableTileMatrixSets[tileMatrixSet].identifiers;
+        tileWidth = Number(usableTileMatrixSets[tileMatrixSet].tileWidth);
+        tileHeight = Number(usableTileMatrixSets[tileMatrixSet].tileHeight);
+        break;
+      }
+    }
+
+    if (!found) return;
```
nf-s: I have added this check - and also a warning which will display if no tileMatrixSet is found.
Does this affect any of your layers?

test/Models/WebMapTileServiceCatalogItemSpec.ts
```diff
@@ -0,0 +1,152 @@
...
+      ) {
+        const url = new URI(mapItems[0].imageryProvider.url)
+          .search("")
+          .toString();
+        expect(url).toEqual("test/WMTS/with_tilematrix.xml");
+      }
+    } finally {
+      cleanup();
+    }
+  });
```
nf-s: These tests are failing due to an error occurring in Cesium here 

https://github.com/TerriaJS/cesium/blob/ecb3ede1797b0a2b79e3ee1dbd8bbff57da3edb4/Source/Scene/WebMapTileServiceImageryProvider.js#L320

```
mobx.module.js?7f83:1847 [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Autorun@1307]' TypeError: Cannot read property 'x' of undefined
    at new WebMapTileServiceImageryProvider (WebMapTileServiceImageryProvider.js?8e6a:302)
    at WebMapTileServiceCatalogItem.get (WebMapTileServiceCatalogItem.ts?1b33:1)
    at trackDerivedFunction (mobx.module.js?7f83:786)
    at ComputedValue.computeValue (mobx.module.js?7f83:1271)
    at ComputedValue.trackAndCompute (mobx.module.js?7f83:1256)
    at ComputedValue.get (mobx.module.js?7f83:1216)
    at shouldCompute (mobx.module.js?7f83:706)
    at ComputedValue.get (mobx.module.js?7f83:1215)
    at shouldCompute (mobx.module.js?7f83:706)
    at Reaction.runReaction (mobx.module.js?7f83:1787)
```

Do you have any idea why this is happening? I'm not sure

lib/Models/WebMapTileServiceCatalogItem.ts
nf-s: Warning looks like this - 
image

test/Models/WebMapTileServiceCatalogItemSpec.ts
zoran995: Will run it later tonight and provide a feedback, but this is something with tiles

zoran995: In [L320](https://github.com/TerriaJS/cesium/blob/ecb3ede1797b0a2b79e3ee1dbd8bbff57da3edb4/Source/Scene/WebMapTileServiceImageryProvider.js#L320) Cesium calls function `positionToTileXY` which then checks if the supplied position is falling in the maximum rectangle defined for that scheme ([this func](https://github.com/TerriaJS/cesium/blob/ecb3ede1797b0a2b79e3ee1dbd8bbff57da3edb4/Source/Scene/WebMapTileServiceImageryProvider.js#L320)). 
The rectangle defined for the scheme has the following bounds
`{west: -3.141592653589793, south: -1.4844222297453324, east: 3.141592653589793, north: 1.4844222297453322}`
while we send 
`{west: -3.141592653589793, south: -1.4844222335884323, east: 3.141592653589793, north: 1.4844222335884323}`
here is north coordinate bigger than the coordinate defined for tiling scheme (last decimal).

There are two solutions that come to my mind to create TillingScheme with the specific rectangle size, or to reduce the size of the rectangle if it is falling outside tillingScheme rectangle.

lib/Models/WebMapTileServiceCatalogItem.ts
nf-s: Ok so my changes are breaking a few layers - for example  https://gibs.earthdata.nasa.gov/wmts/epsg3857/best/wmts.cgi

nf-s:
Note to self: check pixel interpolation for WMS and WMTS
image

lib/Models/WebMapTileServiceCatalogItem.ts
nf-s: I have reverted this

test/Models/WebMapTileServiceCatalogItemSpec.ts
nf-s: I have added a simple check here https://github.com/TerriaJS/cesium/pull/78/files#diff-95a7930932cbafb5bb06b2cae7637aa4R319-R326
I will revisit this later

nf-s:
getFeature requests are not perfect, but the problem lies in https://github.com/TerriaJS/cesium/blob/ecb3ede1797b0a2b79e3ee1dbd8bbff57da3edb4/Source/Scene/WebMapTileServiceImageryProvider.js#L818

nf-s (approved):

nf-s:
Test URL for de-africa http://ci.terria.io/wmts/#clean&https://gist.githubusercontent.com/nf-s/075689074a843bcf99182241406d7858/raw/fdcd0e88617effb3fc46ae64af31b79d1fa88735/wmts.json

zoran995:
thanks for merging this
just one side note those were indeed OWS interfaces but they are used only in WMTS, they are taken from owsGetCapabilities.xsd, and will be referenced if any other specification is implemented that is using that specification (for example WFS but it is handled differently in terria).

nf-s:
Sorry @zoran995, I was a bit under pressure and didn't realise - I'll revert that change next time I'm working on OWS stuff.
Thanks for all of your work!

PR #4728, opened 2020-09-09
Register all catalog members in tests

tephenavies:
### What this PR does

Removes the copied and pasted of `CatalogMemberFactory.register` code from `registerCatalogMembers.ts` to each test that uses a specific CatalogMember through the factory by calling `registerCatalogMembers` before running tests. We also have all of our `UrlMapping`s in `registerCatalogMember`, and I find it an even worse idea to copy and paste those as I don't think we've finished developing them, and any changes to that should be done using TDD to ensure we don't break the recognition of some item types when improving others.

I also improved some existing `createUrlReferenceFromUrl` tests using `async` `await` to pass what were unhandled promise rejections back to Jasmine.

### Checklist


nf-s (approved):
Love it!

PR #4730, opened 2020-09-10
Handle tile errors

na9da:
### What this PR does

Introduces `TileErrorHandlerMixin` for handling tile errors and if required gracefully retry the tiles. This is a port of the same functionality from master as [seen here](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/ImageryLayerCatalogItem.js#L960).

A bunch of options that control the behavior of the mixin is moved under a common trait object called `tileErrorHandlingOptions` available as part of `RasterLayerTraits`.

Testing:
One way to test the behavior is to load any WMS layer and from the network tab of your browser block the URL pattern matching tile requests for the layer. Although, this will only test a narrow case because when the browser blocks the URL it wont set any status code for the blocked request.

### Checklist


lib/Traits/RasterLayerTraits.ts
```diff
@@ -16,4 +50,16 @@ export default class RasterLayerTraits extends ModelTraits {
...
   })
   leafletUpdateInterval?: number;
+
+  @objectTrait({
+    type: TileErrorHandlingTraits,
+    name: "tileErrorHandling",
+    description: "Options for handling tile errors"
+  })
+  tileErrorHandlingOptions: TileErrorHandlingTraits = {
+    thresholdBeforeDisablingItem: 0,
```
rowanwins: I think  in master the `thresholdBeforeDisablingItem` default is 5, see https://github.com/TerriaJS/terriajs/blob/master/lib/Models/ImageryLayerCatalogItem.js#L203

lib/ModelMixins/TileErrorHandlerMixin.ts
```diff
@@ -0,0 +1,366 @@
...
+      );
+
+      // Give up loading this (definitively, unexpectedly bad) tile and
+      // possibly give up on this layer entirely.
+      const failTile = action((e: Error) => {
+        this.tileFailures += 1;
+        const opts = this.tileErrorHandlingOptions;
+        const thresholdBeforeDisablingItem =
+          opts.thresholdBeforeDisablingItem === undefined
+            ? 5
```
rowanwins: See comment below about trait default value

lib/ModelMixins/TileErrorHandlerMixin.ts
```diff
@@ -0,0 +1,366 @@
...
+        // There was an intervening success, so restart our count of the tile failures.
+        this.tileFailures = 0;
+        this.tileRetriesByMap.clear();
+      }
+
+      operation.attempt(
+        action(async attemptNumber => {
+          if (this.show === false) {
+            // If the layer is no longer shown, ignore errors and don't retry.
+            tellMapToSilentlyGiveUp();
```
rowanwins, line 207: I get an uncaught exception saying 
````
TypeError: tellMapToSilentlyGiveUp is not a function
````

**Review by rowanwins — CHANGES_REQUESTED** [review]
A few minor tweaks required @na9da but looking pretty close

File: lib/Traits/RasterLayerTraits.ts (hunk 1 above)
Comment (na9da) [reply lib/Traits/RasterLayerTraits.ts]: Good catch! I forgot to change it back after testing.

File: lib/ModelMixins/TileErrorHandlerMixin.ts (hunk 3 above)
Comment (na9da) [reply lib/ModelMixins/TileErrorHandlerMixin.ts:207]: I have relocated the helper method definitions before its usage.

**Review by tephenavies — DISMISSED** [review]

File: lib/ModelMixins/TileErrorHandlerMixin.ts [hunk 4]
```diff
@@ -0,0 +1,369 @@
...
+     * There are several mechanisms in play here:
+     *   - Cesium's Resource automatically keeps trying to load any resource
+     *     that fails until told to stop, but tells us each time.
+     *   - The "retry" library knows how to pace the retries, and when to actually stop.
+     */
+    onTileLoadError(tileProviderError: TileProviderError): void {
+      const operation = retry.operation(this.tileRetryOptions);
+      // result.reject = stop trying to load this tile
+      // result.resolve = retry loading this tile
+      const result = when.defer();
```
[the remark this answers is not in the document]
na9da: Yep, agree on not using `when`. Here we are forced to use it because our Cesium fork expects a `when` object here - https://github.com/TerriaJS/cesium/blob/terriajs/Source/Core/TileProviderError.js#L161.

I'll add a comment explaining that bit.

lib/ModelMixins/TileErrorHandlerMixin.ts
```diff
@@ -0,0 +1,369 @@
...
+      tileProviderError.retry = result;
+
+      if (tileProviderError.timesRetried === 0) {
+        // There was an intervening success, so restart our count of the tile failures.
+        this.tileFailures = 0;
+        this.tileRetriesByMap.clear();
+      }
+
+      operation.attempt(
+        action(async attemptNumber => {
```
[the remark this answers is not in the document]
na9da: Yep this is kind of a lazy attempt that will protect us till the first `await`. After that I usually wrap things explicitly inside a `runInAction`. In this case I think we need to guard things only till the first `await` - I'll recheck that.

lib/ModelMixins/TileErrorHandlerMixin.ts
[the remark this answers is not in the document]
na9da: This is done.

lib/ModelMixins/TileErrorHandlerMixin.ts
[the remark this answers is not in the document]
na9da: Have made a few changes making sure we are not accessing any mobx variables outside an action.

na9da:
@rowanwins @steve9164 - I have made the requested changes. Let me know if it is good to merge.

rowanwins (approved):

PR #4750, opened 2020-09-16
Fixes a bug that causes the feature info chart for SOS items to not load.

na9da:
### What this PR does

Fixes a bug that causes the feature info chart for SOS items to not load.

This was caused by a mixin ordering bug  that overrides the correct version of `forceLoadChartItems` provided by AsyncChartableMixin.  Removing the method from  `DiscretelyTimeVaryingMixin` simplifies the mixin and saves us from committing the same mistake in the future.

Testing:
Sample SOS catalog for testing - https://gist.github.com/na9da/06aa9694a511e404be5287594f879637

### Checklist


lib/ReactViews/BottomDock/ChartDisclaimer.tsx
```diff
@@ -18,14 +19,18 @@ interface ChartDisclaimerProps {
 const ChartDisclaimer: React.FC<ChartDisclaimerProps> = ({ terria }) => {
   const chartView = new ChartView(terria);
 
-  const chartableItemsWithDisclaimers: (Chartable &
-    DiscretelyTimeVaryingMixin.DiscretelyTimeVaryingMixin)[] = chartView.chartableItems
-    .filter(DiscretelyTimeVaryingMixin.isMixedInto)
-    .filter(item => item.chartDisclaimer);
-  if (chartableItemsWithDisclaimers.length === 0) return null;
-  const uniqueChartDisclaimers = [
-    ...new Set(chartableItemsWithDisclaimers.map(i => i.chartDisclaimer))
```
na9da, line 27: I rewrote this bit to fix a type error I was getting after removing `forceLoadChartItems` from the Mixin.

rowanwins (approved):
Looks good to me

PR #4776, opened 2020-09-24
Upgrade to cesium v1.73

### LINKED ISSUE #4722: update terriajs-cesium to 1.73
(putting the issue here because since the terriajs-cesium repo is forked, we can't make issues in it)

This will give us:
- an updated default ion key— the old one is broken, so any maps using it won't work
- anti-aliasing (once we've merged https://github.com/TerriaJS/terriajs/pull/4651)

**This merge needs to be done very carefully, because changes we made in our fork were accidentally merged upstream and reverted. If we merge the reversion commits in, they'll revert the changes in our fork. Each change will need to be merged manually.**

reginapramesti:
### What this PR does

Partially fixes #4722

* Removed references to Cesium's BingMapsApi (deprecated as of v1.72)
* Upgraded terriajs-cesium 
* Added a new` domhandler` dependency in `package.json` (for some reasons this wasn't there before even though we've been using it!). Using v2.4.2 because that's what we've been using implicitly and things break when we use `domhandler` v3.0.0

### Checklist

-   [x] There are unit tests to verify my changes are correct ~~or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~~


lib/Models/BingMapsSearchProvider.ts
```diff
@@ -39,7 +38,7 @@ export default class BingMapsSearchProvider extends SearchProvider {
     if (this.url.length > 0 && this.url[this.url.length - 1] !== "/") {
       this.url += "/";
     }
-    this.key = BingMapsApi.getKey(options.key);
+    this.key = options.key;
```
KeyboardSounds, line 41: Just double checking, do we need to do this for the other search providers as well? I noticed that this change was made to them as well.

reginapramesti: So the ones that were specifically deprecated and removed as of v1.73 were just the `BingMapsApi` and `MapboxApi`. I only fixed these because `BingMapsApi` (or I suppose the lack thereof 😆) was causing errors. I'm not sure what happened to the rest 🤔 but they seem to "work" just fine

KeyboardSounds (approved):

PR #4781, opened 2020-09-25
Update GtfsCatalogItem to use AutoRefreshingMixin

### LINKED ISSUE #4766: Stop automatically updating autoupdating layers when the layer is disabled in the workbench
For compute intensive layers like Live Buses that reload themselves every 5 seconds, for performance and other reasons we sometimes want to include them in Share/Stories as added to the workbench but not actually enabled. However, in this state the timer is still running and the reloading of the data still occurs, even though we are not displaying the data.

Screen Shot 2020-09-22 at 17 48 31

When autoupdating layers like Live Buses, Current Power Generation etc have been added to the workbench but disabled, we should no longer display/run the timer, and we should no longer perform the data update in the background.

rowanwins:
### What this PR does

Fixes #4766

This PR updates the `GtfsCatalogItem` to use the `AutoRefreshingMixin`.
It also adds a condition to the `AutoRefreshingMixin` to stop unnecessary polling when an item is disabled in the workbench.

### Checklist

-   [ ] No tests exist for either of these models or mixins currently :(

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@ Change Log
 #### next release (8.0.0-alpha.51)
 * Fix story prompt being permanent/un-dismissable
 * Fixed a bug that caused the feature info chart for SOS items to not load.
+* Update the `GtfsCatalogItem` to use the `AutoRefreshingMixin`.
+* Add a condition to the `AutoRefreshingMixin` to prevent unnecessary polling when an item is disabled in the workbench.
 * [The next improvement]
 
 #### 8.0.0-alpha.50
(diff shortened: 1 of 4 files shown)


rowanwins:
Test with http://ci.terria.io/gtfs-autorefreshing/#share=s-qSWEG4nLKHgF4jyQBPUPXA9ADAh
when you disable the item in the workbench you should see requests stop.

KeyboardSounds (commented):
I don't think this was caused by your changes, but it looks like the gtfsFeedEntities are getting recreated every time new data is pulled from the API. This is a visual and performance issue— notice that in the CI deployment of this branch (and also next, weirdly enough), the buses all move at slightly different times and reappear and disappear, compared to prod. Do you want to give fixing it a crack, or should we raise a separate issue for it?

I suspect this code here at the beginning of forceLoadMapItems is the issue:
```
    GtfsStratum.load(this).then(stratum => {
      runInAction(() => {
        this.strata.set(GtfsStratum.stratumName, stratum);
      });
    });
```

KeyboardSounds (approved):

PR #4786, opened 2020-09-25
Fix up duplicating loading of GetCapabilities

### LINKED ISSUE #4740: WMS getCapabilities loaded twice for WMS group and WMS item
Same for WFS - and probably other things.
We should find a way to copy `WebMapServiceCapabilities`

Related:
- #4726

rowanwins:
### What this PR does

Fixes #4740

This PR prevents the duplicate loading of GetCapabilities statements,
- The capabilities are now passed from a WebMapServiceCatalogGroup to a child WebMapServiceCatalogItem
- Capabilities are only requested when the strata doesn't already exist

### Checklist

-   [ ] Don't think tests are required but happy to be told otherwise

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Change Log
 #### next release (8.0.0-alpha.51)
 * Fix story prompt being permanent/un-dismissable
 * Fixed a bug that caused the feature info chart for SOS items to not load.
+* Prevent duplicate loading of GetCapabilities
 * [The next improvement]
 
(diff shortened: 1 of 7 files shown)


nf-s:
I added WMTS group/item + refactored a few things to help me with data catalog performance optimisation - https://github.com/TerriaJS/terriajs/issues/4726
and https://github.com/terriajs/terriajs/issues/4741

nf-s (approved):

PR #4806, opened 2020-09-29
Port gpx support

zoran995:
### What this PR does

The GPX catalog item is supported on the master, this ports it supports to next branch. Also, there is some movement in cesium regarding GPX support.

### Checklist


lib/Models/GpxCatalogItem.ts
```diff
@@ -0,0 +1,128 @@
...
+        );
+        return this._geoJsonItem.loadMapItems();
+      });
+  }
+
+  protected forceLoadMetadata(): Promise<void> {
+    return Promise.resolve();
+  }
+
+  get mapItems(): import("./Mappable").MapItem[] {
```
rowanwins: This looks slightly quirky

lib/Models/GpxCatalogItem.ts
```diff
@@ -0,0 +1,128 @@
...
+  duplicateLoadableStratum(newModel: BaseModel): this {
+    return new GpxStratum(newModel as GpxCatalogItem) as this;
+  }
+
+  static async load(item: GpxCatalogItem) {
+    return new GpxStratum(item);
+  }
+}
+
+StratumOrder.addDefinitionStratum(GpxStratum.stratumName);
```
rowanwins: This stratum doesn't look like it does anything so it can probably be culled. And then below remove the call to 
`GpxStratum.load(this).then(stratum => {`

lib/Models/GpxCatalogItem.ts
```diff
@@ -0,0 +1,128 @@
...
+import TerriaError from "../Core/TerriaError";
+import AsyncMappableMixin from "../ModelMixins/AsyncMappableMixin";
+import CatalogMemberMixin from "../ModelMixins/CatalogMemberMixin";
+import UrlMixin from "../ModelMixins/UrlMixin";
+import GpxCatalogItemTraits from "../Traits/GpxCatalogItemTraits";
+import CommonStrata from "./CommonStrata";
+import CreateModel from "./CreateModel";
+import GeoJsonCatalogItem from "./GeoJsonCatalogItem";
+import proxyCatalogItemUrl from "./proxyCatalogItemUrl";
+import getFilenameFromUri from "terriajs-cesium/Source/Core/getFilenameFromUri";
```
rowanwins: unused import

rowanwins (changes requested):
A couple of minor things @zoran995 but pretty close, thanks!!

lib/Models/GpxCatalogItem.ts
zoran995: most of this unused code is from trying to use getFilenameFromUri for the name of the item (when data is loaded from URL) but was always getting an error

zoran995:
will go once more through everything to make sure it all works fine

rowanwins (approved):
Looks good thanks @zoran995 !

PR #4809, opened 2020-09-29
Add preserveOrder trait on CatalogGroup

rowanwins:
### What this PR does

This PR ports the `preserveOrder` trait for the `CatalogGroup` model. I noticed it was responsible for many errors in the catalog converter. 

It's perhaps not completely ideal, in some regards it would be nice if this was somehow implemented on the `GroupMixin` but that seemed a somewhat more complicated task...

If folks are happy with the approach I'll write a few tests, but I thought I'd whack this up for initial thoughts.

### Checklist


nf-s:
I don't think we should alphabetically order members by default - I think we should preserve order instead

nf-s:
## TODO:

- [ ] Try to move sorting from `CatalogGroup` to `GroupMixin`
- [ ] Port tests from `CatalogGroupSpec.js`

nf-s:
Replaced by https://github.com/TerriaJS/terriajs/pull/6098

Issue #4723, opened 2020-09-08
Stratum choice


zoran995:
Are there more extensive any information around the decisions on CommonStrata choices and when to use which? 

I am working on the implementation of more generalised data search (search in the data catalog) and the usage of `CommonStrata.underride` for all layer data obtained from capabilities seems little wrong to me. Currently, in wms and wfs there are `isOpenInWorkbench` and layer `name` on the same stratum level

nf-s:
Yep totally agree, and we should add some JSDoc strings to the CommonStrata enum

zoran995:
I solved the search part by adding `CommonStrata.search` locally and updating some catalog items to use it, but something like `CommonStrata.catalogItem` is maybe better naming for it to store information about catalog item. 
If you want I can create a pull request with the change mentioned above?

PR #4812, opened 2020-09-29
Catalog search

### LINKED ISSUE #5041: V7 - V8: Catalog search loading spinner never stops
Try searching for data in a v8 map catalog.
image
The "loading" spinner never stops.

zoran995:
### What this PR does

Fixes #5041
@soyarsauce this is how I would update catalog search, my initial implementation was pretty much the same to DT one

Update of the catalog search to use more of traits in search. I went to using `InfoSectionTraits` instead of using the `CommonStratum`s because it produces much cleaner results. 
Although at first, I went for adding search stratum (#4723, https://github.com/TerriaJS/terriajs/issues/3657#issuecomment-700619828) there were few issues with that approach. Usage of `CommonStratum` and model conversion to json will add a lot of additional elements to return search result which is not desired. For example, if an item has legend defined in init file it will be picked as `CommonStratum.definition` and return search results for any legend element. 
When using `InfoSectionTraits` in combination with name, description and id of catalog item returns only results that contain data that user can find in the data preview. 

There was an issue with a wrong prop sent to `SearchHeader` so no message was showing when there are no results or search error occurred. 

The only way that I found to not load all members is to add another trait `loadForSearch` (default `true`) so the administrator can decide which catalog groups should not load for search.

When the #4681 it might be needed to update some of the stuff here

### Checklist


soyarsauce:
@steve9164 says : unsure if another stratum is best approach, we'll think about it a bit more

zoran995:
This uses `InfoSectionTraits` to avoid adding another stratum. I also updated our implementation to remove the additional stratum and use this approach

zoran995:
this is obsolete as terriajs v8 supports catalog search for quite some time

PR #4814, opened 2020-09-30
Add ExtendModel mixin for extending an existing model with new traits.

na9da:
### What this PR does

Adds `ExtendModel` mixin which can be used for extending an existing model with new traits. This can be used for instance in an external tool, to extend an existing model without the pain of duplicating and keeping the Mixin stack in sync.

eg: `const MyCustomWMS = ExtendModel(WebMapServiceCatalogItem, mixTraits(WebMapServiceCatalogItemTraits, MyCustomWMSTraits))`

Ideally, we should be able to pass just the new traits for the extended model (instead of mixing it with the original traits), but I found the type trickery to implement that too hard. Open to suggestions on how that can be done.

For a concrete example check #4815 

### Checklist


lib/Models/ExtendModel.ts
```diff
@@ -0,0 +1,179 @@
...
+  BaseClass extends ModelConstructor<BaseModel>
+>(
+  Base: BaseClass,
+  Traits: T
+): ModelConstructor<InstanceType<BaseClass> & ModelType<InstanceType<T>>> {
+  type Traits = InstanceType<T>;
+  type StratumTraits = StratumFromTraits<Traits>;
+
+  const MixinBase = Base as new (...a: any[]) => BaseModel;
+
```
na9da, line 40: We need this cast because TS requires mixin base classes to have a constructor with a specific shape.

The rest of the code is mostly the same as the previous `CreateModel` impelmentation.

na9da:
Current decision is not to go with this implementation. The same thing can be achieved by moving all `WebMapServiceCatalogItem` bits into a new `WebMapServiceMixin` and then custom catalog items can simply include the mixin.

PR #4815, opened 2020-09-30
Diffable WMS

na9da:
Merge after #4814 

### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4667
Moves the custom code for diffing WMS catalog item to a new catalog item `DiffableWebMapServiceCatalogItem`.

Testing:
Use a catalog item like [this one](https://gist.github.com/na9da/12e8c0dd6de431748228b459c1a5f0ea) with `type` set to `diffable-wms`.

### Checklist


lib/Models/DiffableWebMapServiceCatalogItem.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+}
+
+/**
+ * A custom WMS catalog item that supports diffing imagery from two dates.
+ */
+export default class DiffableWebMapServiceCatalogItem extends DiffableMixin(
+  ExtendModel(WebMapServiceCatalogItem, DiffableWebMapServiceCatalogItemTraits)
+) {
+  static readonly type = "diffable-wms";
```
rowanwins, line 86: So does this mean if adding from a catalog json that you need to do something like 

````
{
   "type": "diffable-wms",
   "url": "http://something",
   "name": "my diffable wms"
}
````

File: lib/Models/DiffableWebMapServiceCatalogItem.ts [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+
+  @computed
+  get disableDateTimeSelector() {
+    return this.catalogItem.isShowingDiff;
+  }
+
+  @computed
+  get shortReport() {
+    const catalogItem = this.catalogItem;
+    console.log("**here**", catalogItem.isShowingDiff);
```
rowanwins: stray console

lib/Models/DiffableWebMapServiceCatalogItem.ts
na9da: Yeah that's right. Sorry I didn't make that clear :grimacing:

soyarsauce:
based on extendmodel, but redo with mixins

PR #4818, opened 2020-09-30
Add support for initializationUrls when loading configuration from Magda

rowanwins:
### What this PR does

This PR allows `initializationUrls` config option to be used when retrieving configuration from Magda.

### Checklist


soyarsauce (approved):
Just waiting for CI

PR #4821, opened 2020-09-30
Fix CI PR build for external contributors

tephenavies:
This PR brings us most of the way to solving https://github.com/TerriaJS/terriajs/issues/4703 I think.
Still needs to be tested by an external contributor. I think we need to merge it to master first then get an external contributor to test pushing a PR to master.

Changes are:
* Upgrade Travis dist from Trusty to Xenial
* Remove `pyenv` which was failing on Xenial. The python version provided in Xenial works and is guaranteed to be consistent.
* Switch Travis CI external PR builds from trying to run on Electron to using Firefox, [as recommended by Karma](http://karma-runner.github.io/5.2/plus/travis.html)
* Remove electron from dependencies. It was only used on Travis CI external PR builds, which always failed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,8 @@ matrix:
       - xvfb
     script:
     - npm run prettier-check
-    - pyenv shell 2.7
     - pip install -r doc/requirements.txt
     - gulp lint docs release
-    - gulp test-firefox
     - gulp test-travis
     env:
     - NODE_OPTIONS=--max_old_space_size=3072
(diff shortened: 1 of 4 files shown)


zoran995:
If I create a PR against this branch will it trigger build using the changes from this branch? Just a dummy PR to test if everything works correctly.

tephenavies:
Hi @zoran995. I think Travis will use the YAML config from the branch being PR'ed to for the PR build. Once this is merged to `master` a dummy PR to test would be great!

KeyboardSounds (approved):
💯 looks good to me!

PR #4823, opened 2020-10-01
Fix traits to model class memoization

na9da:
### What this PR does

Fix memoization of `traitsClassToModelClass`.

Since this method is called during model class definition stage (check `objectTrait` & `objectArrayTrait`), it will always be run outside an `autorun` or an `observer`, hence `createTransformer` will not memoize and throws a warning in the console. In this PR we switch to `memoize` function from lodash.

From the mobx guide:
```
Always use transformations inside a reaction like @observer or autorun. Transformations will, like any other computed value, fall back to lazy evaluation if not observed by something, which sort of defeats their purpose.
```

### Checklist


lib/Traits/traitsClassToModelClass.ts
```diff
@@ -1,11 +1,11 @@
-import { createTransformer } from "mobx-utils";
+import memoize from "lodash-es/memoize";
 import CreateModel from "../Models/CreateModel";
 import ModelTraits from "./ModelTraits";
 import TraitsConstructor from "./TraitsConstructor";
 
-const traitsClassToModelClass = createTransformer(function<
-  T extends ModelTraits
->(traitsClass: TraitsConstructor<T>) {
+const traitsClassToModelClass = memoize(function<T extends ModelTraits>(
```
KeyboardSounds, line 11: Could you please add a version of your comment from the PR as a comment in the code here? It's useful context for understanding why we're using memoize here.

KeyboardSounds (changes requested):

lib/Traits/traitsClassToModelClass.ts
na9da: Good point! I have added a note.

KeyboardSounds (approved):

PR #4824, opened 2020-10-01
Update Basemaps url and attributions

skgsergio:
### What this PR does

This PR updates CARTO Basemaps URL to the new CDN URL and fixes the attributions to match current requirements.

I did not add the proper attribution HTML since there is a comment regarding that was not working properly, which I don't know if is still true:
https://github.com/TerriaJS/terriajs/blob/8ec9a207164d719ea898e5ac19876c5c42181b6a/lib/ViewModels/createGlobalBaseMapOptions.js#L53-L57

If the issue is now fixed the preferred attribution (both by OSM and CARTO) is:

```html
&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>, &copy; <a href="https://carto.com/about-carto/">CARTO</a>
```

#### Context
We've recently changed our CDN (last week) and we have a plan mid-long term (depends on how quick we manage to stop all the traffic there) to deprecate our `.global.ssl.fastly.net` domains.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 Change Log
 ==========
 
+### v7.11.14
+
+* Update CARTO Basemaps CDN URL and attribution.
+
 ### v7.11.13
 
 * Upgraded to Cesium v1.73.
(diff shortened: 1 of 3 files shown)


rowanwins (approved):
Thanks @skgsergio , looks great!

PR #4825, opened 2020-10-01
Update Carto basemaps

zoran995:
### What this PR does

Fixes 

As stated by @skgsergio in #4824. This updates Carto base maps to updated URLs, and attribution to the correct one. This was changed in #971, but it seems that new URLs support https. If the footer is the only place the attribution is used I can confirm that it works.
2D
3D

### Checklist


rowanwins (approved):
Looks good thanks @zoran995

skgsergio:
Thanks! Didn't notice the next branch.

PR #4826, opened 2020-10-01
Correctly handle rounded corners on isolated buttons in menu-bar

rowanwins:
### What this PR does

This PR adds an `:only-child` selector to `menu-bar.scss` to ensure that the menu bar is correctly styled in terms of border radius when there is only one item. 

This is the sort of thing it fixes.

Screen_2020_10_02

### Checklist

-   [ ] no tests added


rowanwins:
THis is what things now look like
Fixed

Easiest way to check to just to open up the ci deploy and delete one of the top left menu items using the "elements" tab in devtools.

soyarsauce (approved):
Hi @rowanwins - thanks for this! Totally understanding this is a complex, convoluted component - just keeping in back of mind https://github.com/TerriaJS/terriajs/blob/next/doc/contributing/frontend-style-guide.md#mobx-ui-refactor

>Any work to existing components should include an effort to remove sass from it.

PR #4827, opened 2020-10-01
Add clipToRectangle trait

rowanwins:
### What this PR does

This PR ports the `clipToRectangle` property that used to exist in master on [`ImageryLayerCatalogItem`](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/ImageryLayerCatalogItem.js#L108-L119).

I've added the implementation to a few of the key catalog item types. Mostly it was implemented however you couldn't disable it.

### Checklist

### DIFF
--- a/lib/Models/WebMapServiceCatalogItem.ts
+++ b/lib/Models/WebMapServiceCatalogItem.ts
@@ -953,7 +953,14 @@ class WebMapServiceCatalogItem
 
       let rectangle;
 
-      if (this.clipToRectangle && this.rectangle !== undefined) {
+      if (
+        this.clipToRectangle &&
+        this.rectangle !== undefined &&
+        this.rectangle.east !== undefined &&
+        this.rectangle.west !== undefined &&
+        this.rectangle.north !== undefined &&
+        this.rectangle.south !== undefined
+      ) {
         rectangle = Rectangle.fromDegrees(
           this.rectangle.west,
           this.rectangle.south,
--- a/lib/Models/WebMapServiceCatalogItem.ts
+++ b/lib/Models/WebMapServiceCatalogItem.ts
@@ -953,14 +953,7 @@ class WebMapServiceCatalogItem
 
       let rectangle;
 
-      if (
-        this.clipToRectangle &&
-        this.rectangle !== undefined &&
-        this.rectangle.east !== undefined &&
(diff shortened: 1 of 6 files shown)


lib/Models/WebMapServiceCatalogItem.ts
```diff
@@ -954,6 +954,7 @@ class WebMapServiceCatalogItem
       let rectangle;
 
       if (
+        this.clipToRectangle &&
         this.rectangle !== undefined &&
         this.rectangle.east !== undefined &&
```
na9da, line 959: Should we perhaps remove these individual checks for each east, west, north, south so that we have a uniform behavior for both WMS & ArcGIS? Their values [default to 0](https://cesium.com/docs/cesiumjs-ref-doc/Rectangle.html?classFilter=rectangle).

na9da:
Looks good @rowanwins. Just one discussion point.

No ci because of the camel case branch name.

lib/Models/WebMapServiceCatalogItem.ts
rowanwins: Huh we'll there you go, I should've checked the defaults, definitely need to work out how to consider them because that could have unintended consequences!

rowanwins: Actually @na9da I think they default to undefined, this is our rectangle, not the Cesium Rectangle
https://github.com/TerriaJS/terriajs/blob/next/lib/Traits/MappableTraits.ts#L7-L35

na9da: I meant in [this call](https://github.com/TerriaJS/terriajs/blob/next/lib/Models/WebMapServiceCatalogItem.ts#L963) if one of east, west, south, north is undefined it will default to 0. So right now if the catalog definition is like `{west: 30, east: 50, south: -10}` then for WMS no imagery will be shown (because north is undefined), but for ArcGIS we show imagery for the region  `{west: 30, east: 50, north: 0, south: -10}`.

It is a trivial thing but good to be consistent, right?

rowanwins: Ahh ok, it's hard to imagine a situation where you might end up with a partial rectangle but anything could happen I guess!

na9da (approved):

PR #4830, opened 2020-10-02
Various fixes to menu bar in mobile

### LINKED ISSUE #4085: Mobx/Digital twin UI issues on small screen - Tools menu
When a user resizes the browser window there are a few edge cases that he can see

- related maps get moved to the second row when the user is logged, and there are three rows (first image), in my case between 768px and 797px (on longer names it will happen even earlier)
- before switching to mobile view there is a possibility to see something like on the second image (blue line around logo)
- on a screen with a width less than 1000px related buttons get out the window (user can't see the close button)
first image
blue line
related maps

reginapramesti:
### What this PR does

Partially fixes #4085. Specifically the one Ana raised on missing hamburger icon and the branding not being on the menu bar

Things I did in this PR:
* Refactored some of `MobileHeader.jsx`, themeifying the left hand side of the mobile menu bar (the hamburger menu + branding section)
* Part of the reason why the branding was located off the menu bar was because it was too long if we were to display all 3 of them. The simple fix would've been to just fit it somehow, with the buttons on the right hand side of the menu bar overlapping the branding, but I figured we should do what master does. I made it so that branding is kind of like in master where only one of the brands appear on the mobile menu, rather than all 3 (example I referred to was nationalmap), so I added a new `configParameter` that allows us to choose which of the brands will appear on mobile. Though in master it's done through the TerriaMap `global.scss` file hiding the 0th and 2nd branding image. Not entirely sure which way is more preferable though. Regardless, it shouldn't affect the way it displays on desktop I believe

### Checklist


lib/ReactViews/SidePanel/Branding.jsx
```diff
@@ -18,18 +19,26 @@ const Branding = createReactClass({
       .brandBarElements;
     if (!defined(brandingHtmlElements)) {
       brandingHtmlElements = [
-        '<a target="_blank" href="http://terria.io"><img src="images/terria_logo.png" height="52" title="Version: {{ version }}" /></a>'
+        "",
+        '<a target="_blank" href="http://terria.io"><img src="images/terria_logo.png" height="52" title="Version: {{ version }}" /></a>',
+        ""
```
reginapramesti: Note I had to change this so that the "default" would be an array of 3 like in TerriaMap's `config.json` just to make it work with the way I implemented the branding stuff

soyarsauce:
image

soyarsauce (approved):

lib/ReactViews/SidePanel/Branding.jsx
soyarsauce: As discussed, resolved via https://github.com/TerriaJS/terriajs/pull/4830/commits/462171b89b2b6e6b53143a70974618da4100ddce

PR #4831, opened 2020-10-02
Don't merge - testing #4821

zoran995:
Testing the build for external PR, after the changes from #4821.


zoran995:
@steve9164

tephenavies:
@zoran995 Looks like the test runner worked, but it failed `prettier-check`. And that's because an external PR got merged to `master` without proper formatting due to the check being part of a permanently failing PR build (the very issue we've just now solved). I'll push a prettier fix to master and I think we can say that this works.

PR #4835, opened 2020-10-05
Mobx fix magda init

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4833

Allow Magda backed maps to use an inline `terria-init` catalog without it getting overwritten by map-config before it can be parsed

Fixes https://github.com/TerriaJS/terriajs/issues/4834

Deprecates `proxyableDomainsUrl` config param

### Checklist


soyarsauce:
Restarted job, seems like it got stuck on saucelabs

soyarsauce:
Tests were actually broken - I even mentally noted to myself to uninstall the inlined stub but ended up not. Fixed them now!

lib/Models/Terria.ts
```diff
@@ -876,16 +876,20 @@ export default class Terria {
     const configParams =
       aspects["terria-config"] && aspects["terria-config"].parameters;
 
-    configParams.initializationUrls =
-      aspects["terria-config"] && aspects["terria-config"].initializationUrls;
+    // TODO: what purpose does this serve? it's not a config parameter
+    // configParams.initializationUrls =
+    //   aspects["terria-config"] && aspects["terria-config"].initializationUrls;
```
rowanwins: Agreed, I think this can be culled, from my reading of the code it serves no purpose.

rowanwins (approved):
Looks good

lib/Models/Terria.ts
soyarsauce: image

Thanks for confirming 😀

rowanwins: I gitblame you @soyarsauce 😂 
Screen_2020_10_07

soyarsauce: Hahaha sorry for misleading with half-pseudo-code!! You'll note my suggestion doesn't change the `initializationUrls` key on `configParams`. It looks like a naming problem, should have been called `initUrls` or something - `const initUrls = aspects["terria-config"] && aspects["terria-config"].initializationUrls;` and done what you will with the list of init urls there.

Anyway the extra mocked config map tests will help somewhat ensure things are loaded the right way 👍 it was quite difficult to verify without those unit tests.

PR #4838, opened 2020-10-06
properly render polygons with holes

### LINKED ISSUE #4255: Support for 2D polygons with holes
When importing a vector polygon that has holes in it, all holes are rendered correctly in 3D modes, but are disregarded in 2D mode because the hierarchyToLatLngs function (https://github.com/TerriaJS/terriajs/blob/master/lib/Map/LeafletVisualizer.js#L530) does not handle polygons with holes. 
[example-holes-geojson.zip](https://github.com/TerriaJS/terriajs/files/4530278/example-holes-geojson.zip)

Screenshot from 2020-04-24 10-51-59
Screenshot from 2020-04-24 10-52-17

Desired 2D rendering: 
Screenshot from 2020-04-24 10-56-28

PR on its way:
```
function hierarchyToLatLngs(hierarchy) {
  let holes = [];
  var positions = Array.isArray(hierarchy) ? hierarchy : hierarchy.positions;

  // Manage polygons with holes
  if (hierarchy.holes.length > 0) {
    hierarchy.holes.forEach((hole) => {
      holes.push(convertEntityPositionsToLatLons(hole.positions));
    });
    return [convertEntityPositionsToLatLons(positions), holes];
  } else {
    return convertEntityPositionsToLatLons(positions);
  }
}

### PR DESCRIPTION
**Pull request by rowanwins**
### What this PR does

Fixes #4255

This PR fixes rendering of polygons with holes in the leaflet viewer.

Use this geojson as an example
https://gist.githubusercontent.com/rowanwins/d8226663338d2514438511fa765ca063/raw/720571451c30823e53b7fe295cf86189fea11b02/polygon-holes.geojson

### Checklist

### CODE REVIEW

File: lib/Map/LeafletVisualizer.ts [hunk 1]
```diff
@@ -1226,10 +1226,16 @@ function positionToLatLng(
 }
 
 function hierarchyToLatLngs(hierarchy: PolygonHierarchy) {
-  // This function currently does not handle polygons with holes.
-
+  let holes: any = [];
```
rowanwins: Couldn't get typings to work properly :(

lib/Map/LeafletVisualizer.ts
```diff
@@ -1226,10 +1226,16 @@ function positionToLatLng(
...
   const positions = Array.isArray(hierarchy) ? hierarchy : hierarchy.positions;
-  return convertEntityPositionsToLatLons(positions);
+  if (hierarchy.holes.length > 0) {
+    hierarchy.holes.forEach(hole => {
+      holes.push(convertEntityPositionsToLatLons(hole.positions));
+    });
+    return [convertEntityPositionsToLatLons(positions), holes];
```
na9da: Should this be `return [convertEntityPositionsToLatLons(positions), ...holes];` instead?
That seems to satisfy the type and also matches the [documentation](https://leafletjs.com/reference-1.7.1.html#polygon) (2nd example)

rowanwins: Good pickup @na9da - that also fixed my typing problem!!

na9da (approved):

PR #4840, opened 2020-10-07
 Hide the chart download button when there is no downloadable source.

na9da:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/4836

The fix reflects the behavior in master.
eg from master: https://maps.dea.ga.gov.au/#share=s-fhMSo2vlaqT8iFFH3r4sZ0WU2zh

### Checklist


rowanwins (approved):
Looks good!

PR #4848, opened 2020-10-08
ArcGisReferenceItem and ArcGisPortalGroup fixes

rowanwins:
### What this PR does

- [x] Only gets rectangle from an arcgis portal item where it is correctly defined
- [x] Improve support for when a portal item is in multiple groups

### Checklist

-   [x] Unit tests added


lib/Models/ArcGisPortalCatalogGroup.ts
```diff
@@ -317,7 +317,7 @@ export class ArcGisPortalStratum extends LoadableStratum(
       return;
     }
     const id = this._catalogGroup.uniqueId;
-    const itemId = id + "/" + arcgisDataset.id;
+    const itemId = `${id}/${arcgisDataset.id}`;
```
mwu2018, line 320: It seems there are no fundamental changes here. How does it solve the problem raised in https://github.com/TerriaJS/nsw-digital-twin/issues/459?

rowanwins: This change was just for niceness, this is called a [template string](https://developers.google.com/web/updates/2015/01/ES6-Template-Strings), they tend to be a bit nicer that using `"string1" + "string2" + "string3"` pattern

mwu2018: I am trying your fixes in my local deployment.

mwu2018: Yes, I can confirm that both issues https://github.com/TerriaJS/nsw-digital-twin/issues/458 and https://github.com/TerriaJS/nsw-digital-twin/issues/459 are resolved by this PR. 

However, I don't understand why by checking `arcgisPortalItem.extent` can solve items missing in catalog problem. Is there anything I missed?

rowanwins: So I think the extent can also be retrieved via the `Cesium3DTilesCatalogItem` which reads it from the `SceneServer` endpoint rather than the portal item. So I suspect the dodgy value was overriding the correct value.

lib/Models/ArcGisPortalItemReference.ts
```diff
@@ -114,7 +114,10 @@ export class ArcGisPortalItemStratum extends LoadableStratum(
...
     if (this.arcgisPortalItem === undefined) return undefined;
 
-    if (this.arcgisPortalItem.extent !== null) {
+    if (
+      this.arcgisPortalItem.extent !== null &&
+      this.arcgisPortalItem.extent.length === 2
+    ) {
       return createStratumInstance(RectangleTraits, {
         west: this.arcgisPortalItem.extent[0][0],
         south: this.arcgisPortalItem.extent[0][1],
```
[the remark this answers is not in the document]
rowanwins: Yeah so sometimes a portal item will report an extent like
````
{
   ...
   extent: [],
   ...
}
````
rather than a valid value like
````
{
   ...
   extent: [[0,0], [1,1]],
   ...
}
````

**Comment by rowanwins** [comment]
If you're happy with this @mwu2018 do you want to approve?

**Comment by mwu2018** [comment]
> If you're happy with this @mwu2018 do you want to approve?

There is one more thing I need to check: Although the item that was missing previously now appears in the catalog and can be added to the map, it can not be visualised in 3D map for unknown reasons. However, the item can be visualised in 2D map. It seems that it has something to do with the values of extent.

The preview map shows that the bbox and items do not line up:
image

After clicking on the preview map, it becomes
image

This might be the problem -- The `west` and the `east` have the same value.
image


There is no problem if the item is harvested by the old portal crawling.
With old portal crawling, the url is 

  `https://portal.spatial.nsw.gov.au/server/rest/services/Wollondilly_Proposed_Plan/MapServer`.

With the real time portal search, the url is 

  `https://portal.spatial.nsw.gov.au/server/rest/services/Wollondilly_Proposed_Plan/MapServer/0`

File: lib/Models/ArcGisPortalItemReference.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (mwu2018) [reply lib/Models/ArcGisPortalItemReference.ts]: Maybe we also need to validate the values, at least to make sure `east` > `west` and `north` > `south`. Otherwise, wrong values might screw up other things.

**Comment by rowanwins** [comment]
Good pickup @mwu2018 

So the portal item itself has an extent which is reporting as 
````
"extent": [
   [150,-34],
   [150,-33]
]
````
The problematic bit is that east and west are both returned as 150. Terria tries to be smart and use the rectangle to limit requests but when a rectangle becomes impossibly small it doesn't request anything.

The MapServer that the portal item points to gives a more accurate extent with the following values 
````
"xmin": 150.52507310014357,
"ymin": -34.29876999963574,
"xmax": 150.83069979965614,
"ymax": -33.899034400217374
`````
So the xmin and xmax are higher precision

End result is I that I've culled the rectangle calculation based on the portal item json and we'll just rely on the services that they point to.

**Comment by mwu2018** [comment]
I will run another test.

File: lib/Models/ArcGisPortalItemReference.ts [hunk 3]
```diff
@@ -111,20 +111,6 @@ export class ArcGisPortalItemStratum extends LoadableStratum(
...
-      return createStratumInstance(RectangleTraits, {
-        west: this.arcgisPortalItem.extent[0][0],
-        south: this.arcgisPortalItem.extent[0][1],
-        east: this.arcgisPortalItem.extent[1][0],
-        north: this.arcgisPortalItem.extent[1][1]
-      });
-    }
-    return undefined;
-  }
-
```
mwu2018, line 127: Yep, by removing this function the items can be properly visualised in the map.

mwu2018 (approved):
Local test has verified the issues had been resolved.

Issue #4409, opened 2020-06-03
Wrong position of help panel opened from compass


zoran995:
On CI version of mobx help panel opened from compass shows in the middle of the map, I believe that this is happening because it doesn't find help item to open.

![compass help](https://user-images.githubusercontent.com/19208948/83675611-9d135e80-a5d9-11ea-9b7b-c3b131bc50bc.png)

zoran995:
this is resolved

Issue #4582, opened 2020-07-29
DE UX Testing: Improve compass interactivity


vicborgy:
Covering issue no. 31 of [spreadsheet](https://docs.google.com/spreadsheets/d/10eWTU44vUEk-NKm2UcUPEK7qeHSDnRzrXUbMV6nDuBE/edit#gid=0) (medium priority):

**Issue**
> The new compass was annoying some participants, it opens too quickly and they kept accidentally opening it. There's no delay on hover over to prevent accidental opening

**Fixes**
Vic to improve:
- Buttons & styling
- Interaction with compass

soyarsauce:
@vicborgy to attach new solution design

vicborgy:
Here are the designs for the revised compass interaction - https://invis.io/GDYDNTW2UTQ

- Hover state of compass shows a chevron that pops out from behind the compass - rolling away from the compass will deactivate the hover state of the compass
- Clicking the chevron will activate the compass hover info modal
- Closing the modal will close the modal and the open state of the compass

soyarsauce:
Thanks for this @vicborgy - this should really improve things for the compass. Bar the compass not actually working for tablets, what do you think the interaction there should be (for touch devices that are large enough to load the "non mobile UI"?)

Semi related, are you able to progress https://github.com/TerriaJS/terriajs/issues/4189 (& subsequently https://github.com/TerriaJS/terriajs/issues/4673)?

vicborgy:
For touch devices the action to make it expand and show the chevron (and expand compass) would be clicking on the compass or a click + hold.

I will check out those issues you sent.

rowanwins:
Resolved by #4849

PR #4849, opened 2020-10-08
Fix compass help bug

### LINKED ISSUE #4673: Mobx: compass points to non-existent default navigation help
The compass will tempt you to `Find out more about the controls and how to use them.` , only to open out to a dead help panel (as well as the `help` button that comes out of the compass)

This was supposed to be filled out by https://github.com/TerriaJS/terriajs/issues/4189 but that's been backlogged now

We should improve the defaults

image

reginapramesti:
### What this PR does

Fixes #4673 and implements #4582

Implemented new compass design, which in turn fixes the compass help button bug. Also enabled users to permanently disable compass controls from the dropdown. 

### Checklist


zoran995:
also fixes #4409

reginapramesti:
There's still the bug described https://github.com/TerriaJS/terriajs/issues/4177. I tried to fix it but after some investigation, I think it's a bigger problem than it looks. Tried adding the `onDismissed` props in `CleanDropdownPanel` but it behaves strangely, which, after some more investigation, I found is because `CleanDropdownPanel` never calls `this.open()`, unlike the generic `DropdownPanel` class. That function does some trickery to work with the `onDismissed` props. 

Anyway I think I'll leave this bug for now and will focus on other more important bugs

KeyboardSounds:
This looks great! The only thing I'm wondering about is whether the little arrow dropdown icon should be replaced by a ❓ icon instead. Otherwise it looks like a settings menu.

reginapramesti:
Yeah that's a good idea, I was initially just following the designs but then Wing suggested for the dropdown to be below the button rather than to the left (for various reasons) and so we just changed it from a left arrow to a down arrow 😂

KeyboardSounds (approved):

PR #4850, opened 2020-10-08
Fix mobile layout 

zoran995:
### What this PR does

The current implementation of the mobile layout had an issue with map navigation bar being hidden in the bottom of the screen. The only way to reveal the navigation bar is to hide the address bar by scrolling on the map header. This now fix this by fixing the address bar to be visible and removes `position:relative` from `.ui-root` (not sure why it is needed but it worked, after few hours of debugging how to fix it). Haven't seen anything broken, everything also worked on the desktop.

old behaviour
image  image

new behaviour 
image

### Checklist


wibily (approved):

wibily:
thanks @zoran995!

PR #4857, opened 2020-10-09
Fix presentation of the WMS Dimension metadata

rowanwins:
### What this PR does

The Dimension Metadata on a WebMapServiceCatalogItem was displaying all weird. The json object is a it funky so I've had to strip out the non-functioning bits in a weird way. It's probably something to do with the xml parsing but given that they might just changing this is a short term fix.
Screen_2020_10_09

### Checklist


lib/Models/WebMapServiceCatalogItem.ts
```diff
@@ -322,7 +322,19 @@ class GetCapabilitiesStratum extends LoadableStratum(
         {},
         this.capabilitiesLayers.get(this.catalogItem.layersArray[0])
       ) as any;
+
       if (out !== undefined) {
+        // The Dimension object is really weird and has a bunch of stray text in there
+        if ("Dimension" in out) {
+          const goodDimension: any = {};
+          Object.keys(out.Dimension).forEach((k: any) => {
```
rowanwins, line 330: Tried a bunch of prettier ways to do this but it was being super fiddily so sorry it's ugly

tephenavies:
This one is weird. The XML parser (our local `ThirdParty/xml2json.js`) turns:
```xml
<Dimension name="time" default="2014-01-01T00:00:00Z" units="ISO8601">2014-01-01T00:00:00.000Z/2014-12-31T23:00:00.000Z/PT1H</Dimension>
```

into a `String` (yes uppercase String object) with the value of the inner text and then adds the attributes as keys/values to the String object. When MetadataTable tries to display the object iterating over `Object.keys(a)` it gets all the numeric keys.

I think your fix is the right sort of fix. We shouldn't give such a weird object to MetadataTable and expect it to be able to display it properly.

tephenavies (approved):

PR #4859, opened 2020-10-09
Add itemProperties trait

rowanwins:
### What this PR does

This ports the `itemProperties` property on `WebMapServiceCatalogGroup` from master to next

### Checklist

-   [x] Added a basic test


na9da (approved):
Looks good to me!

I think, ideally we should have something like `@partialObjectTrait` for traits of these kinds, but that will need some knife fight with the compiler :grinning:

PR #4866, opened 2020-10-13
Chart fixes for #4852

### LINKED ISSUE #4852: Time series plotting in V8 behaves differently (unwanted) to V7 for DEA Waterbodies
##DEAMaps V8 testing issue
### Expected behavior
In V7, when you click on a waterbody and hit expand, it will automatically only plot the 'percentage of surface area (not volume) observed as wet'. This is the preferred behavior.
image

### Actual behaviour
In V8, if you click on a waterbody and hit expand it will automatically plot both timeseries
image

In addition, the labels we applied to the timeseries graphs do not seem to be carried over to V8. I.e the legend refers to them as 'wet pixel percentage' and 'wet pixel count (n=x)', not 'percentage of surface area (not volume) observed as wet' and 'wet pixel count'.

### Steps to reproduce the behaviour
V7 sharelink http://maps.dea.ga.gov.au/#share=s-cI4nR0Elhu3IdyDg8gl8jeH45iM
V8 sharelink https://de-australia.prod.saas.terria.io/#share=s-sX8oOxariSJpj7tO2kKACYJ0BbO

na9da:
### What this PR does

Fixes #4852 

* Chart expanded from feature info panel will now by default show only the first chart line.

* Chart component attributes `column-titles` and `column-units` will now accept a simpler syntax like: `"Time,Speed"` or `"ms,kmph"` just like in `master`. 

  Previously we required the user to provide a string like `"Column-name-1:Time,Column-name-2:Speed"` keyed by the column name. This was because `@objectArrayTrait columns` is indexed by `name` and not the column number. To set column title we need to know the column name first, and since we haven't fetched the data yet, the user has to provide it. 

  The resulting syntax is difficult to write and sometimes impossible too, because the column names might be generated dynamically by the server. This fix stores the `titles` & `units` array as a parent item trait (`columnTitles` & `columnUnits`) and derives the `title` and `unit` for a column from it (if set).

Test link: http://ci.terria.io/chart-4852/#share=s-dBV6vxoZg9BG4PmkNSHg8JqrmbG

### Checklist


rowanwins (approved):
Looks good 👍

AnaBelgun:
Is this ready to merge?

PR #4872, opened 2020-10-16
Fix handling of `ArcGisPortalItemReference` for when a feature layer …

rowanwins:
### What this PR does

This PR fixes a condition where a portal item supports multiple sub-layers. 

A portal item contains a url, often this points to something like https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/FeatureServer
which as an entire feature service contains 17 sublayers.

However sometimes the portal item also lists a subset of layers at the `/data` endpoint for the portal item, for example it might specify that this portal item only includes layers 1,2 & 3.

The condition I originally wrote was broken. Basically the condition should be that if there is only 1 layer associated with the portal item we update the url to point at the sublayer, eg 
so
https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/FeatureServer
becomes
https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/FeatureServer/2 
(or whatever the relevant layer id is) 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Change Log
 #### next release (8.0.0-alpha.56)
 * Add `itemProperties` trait to `WebMapMapCatalogGroup`.
 * Add support for `formats` traits within `featureInfoTemplate` traits.
+* Fix handling of `ArcGisPortalItemReference` for when a feature layer contains multiple sublayers.
 * [The next improvement]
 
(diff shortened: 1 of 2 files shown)


mwu2018:
Verified in local deployment.

mwu2018 (approved):

PR #4884, opened 2020-10-21
Fix zoom for satellite imagery in DEA maps

### LINKED ISSUE #4853: 'Ideal zoom' button zooms out waaaay too much for satellite data
### Expected behavior
N/A - this is a new V8 feature

### Actual behaviour
If you load in a satellite scene, then select 'ideal zoom', it zooms to an unhelpful map view
2D basemap:
image

3D basemap
image

### Steps to reproduce the behaviour
Load some satellite data, e.g. https://de-australia.prod.saas.terria.io/#share=s-t9qIomI0RVVey7AjNiIH4LMaXr7
Then select 'ideal zoom'

reginapramesti:
### What this PR does

Fixes #4853 

In master we perform some kind of "check" before zooming in (see [here](https://github.com/TerriaJS/terriajs/blob/master/lib/Models/CatalogItem.js#L1002)), to make sure that zooming doesn't break for datasets with bounding boxes that are overly large (or invalid), which is the underlying problem in the linked issue. The default zoom extent in such cases should be the `homeCamera` view (this is the behaviour in master). I am basically replicating that in mobx.

I added the code to `zoomTo` in `ViewingControls`, which seems like a weird place to add it. Though it allows me to _not_ have to add the code to both Leaflet and Cesium's `zoomTo`. The other alternative I thought of was to replace the `rectangle` property when it's first initialised in the catalogue item, but doing it this way allows me to _not_ have to do that for all catalogue items with `RectangleTraits` (ArcGis/WebMap/everything else that has `.rectangle`) so 🤷 

### Checklist

### DIFF
--- a/lib/ReactViews/Workbench/Controls/ViewingControls.jsx
+++ b/lib/ReactViews/Workbench/Controls/ViewingControls.jsx
@@ -111,7 +111,16 @@ const ViewingControls = observer(
 
     zoomTo() {
       const viewer = this.props.viewState.terria.currentViewer;
-      viewer.zoomTo(this.props.item);
+      const item = this.props.item;
+      let zoomToView = item;
+      if (
+        item.rectangle !== undefined &&
+        item.rectangle.east - item.rectangle.west >= 360
+      ) {
+        zoomToView = this.props.viewState.terria.mainViewer.homeCamera;
+        console.log("Extent is wider than world so using homeCamera.");
+      }
+      viewer.zoomTo(zoomToView);
     },
 
     openFeature() {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 
 #### next release (8.0.0-alpha.57)
 * Fix memoization of `traitsClassToModelClass`.
+* Fix zooming bug for datasets with invalid bounding boxes.
 * [The next improvement]
   
 #### 8.0.0-alpha.56


lib/ReactViews/Workbench/Controls/ViewingControls.jsx
```diff
@@ -111,7 +111,16 @@ const ViewingControls = observer(
 
     zoomTo() {
       const viewer = this.props.viewState.terria.currentViewer;
-      viewer.zoomTo(this.props.item);
+      const item = this.props.item;
+      let zoomToView = item;
+      if (
+        item.rectangle !== undefined &&
+        item.rectangle.east - item.rectangle.west >= 360
```
rowanwins, line 118: The only question mark I have about this is whether it will work for things in the Pacific that cross the anti-meridian...

reginapramesti: I think if it does get to beyond 360 degrees from east to west there would still be something wrong with the bounding box, given that the only scenario this would break is if the east >= 180 and west <= -180, in which case the east should be the west value and vice versa and it seems like a weird bounding box anyway? 🤔 Though correct me if I'm wrong but I don't think it would break pacific map (as I imagine the behaviour in master doesn't break it either)

reginapramesti: I'm thinking even in the worst case scenario where a bounding box has `east: 179.9` and `west`-179.9`, which is reasonably likely given the location of pacific islands 😂 , it still would zoom to the desired location

rowanwins: Given that I can't find a good example and it's an edge case let's just role with what you've done :)

rowanwins (approved):
👍

PR #4887, opened 2020-10-21
Fix magda based maps when mix matching init styles

soyarsauce:
### What this PR does

Fix magda based maps when mix matching init styles

When loading in magda based maps with `initializationUrls`, the id of the map config record would be used as the child records now referenced a root group with ID `map-config` instead of `/` in a static-json based catalog.

I also realise this is going to be hard to reproduce/test, basically, with this branch, and terrace (https://github.com/TerriaJS/terrace / https://github.com/TerriaJS/handbook/blob/master/DeveloperGuides/SaasMaps.md), point it at any magda based map, and observe 

`//Corona Virus - Cases`
vs
`map-config-pacificmap/Corona Virus - Cases`

image

image

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ Change Log
 * Chart expanded from feature info panel will now by default show only the first chart line.
 * Chart component attribtues `column-titles` and `column-units` will now accept a simpler syntax like: "Time,Speed" or "ms,kmph"
 * Fix presentation of the WMS Dimension metadata.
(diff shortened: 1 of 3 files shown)


soyarsauce:
This was the initial approach I took, and deciding it was not the best idea given the unpredictability it would have introduced: https://github.com/TerriaJS/terriajs/pull/4888/files

nf-s (approved):

nf-s:
Btw I didn't test this locally because I haven't used Magda before - but looks good!

soyarsauce:
image

PR #4907, opened 2020-10-27
Fix CKAN naming

### LINKED ISSUE #4903: Investigate naming of CKAN dynamic items in version 8
CKAN group seems to name it's items by CKAN Dataset instead of Resource.

- [x] Investigate why this is happening on http://pacificmap.prod.saas.terria.io/#share=s-600IuxFH0DlYfU0Gjspb88pe7Dm
- [x] Fix ASAP

image

tephenavies:
### What this PR does

Fixes #4903.

Add a stratum of computeds so that CkanItemReference inherits shared traits from a containing CkanCatalogGroup

### Checklist

### DIFF
--- a/lib/Models/CkanItemReference.ts
+++ b/lib/Models/CkanItemReference.ts
@@ -1,5 +1,5 @@
 import i18next from "i18next";
-import { computed, runInAction } from "mobx";
+import { action, computed, runInAction } from "mobx";
 import { createTransformer } from "mobx-utils";
 import URI from "urijs";
 import isDefined from "../Core/isDefined";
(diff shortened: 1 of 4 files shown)


lib/Models/CkanItemReference.ts
```diff
@@ -124,16 +127,16 @@ export class CkanDatasetStratum extends LoadableStratum(
...
-        return this.ckanDataset.title + " - " + this.ckanResource.format;
+        return this.ckanDataset.title + " - " + this.ckanResource.name;
       }
       if (
-        this.ckanItemReference.useCombinationNameWhereMultipleResources &&
+        this.ckanItemReference.useDatasetNameAndFormatWhereMultipleResources &&
         this.ckanDataset.resources.length > 1
       ) {
-        return this.ckanDataset.title + " - " + this.ckanResource.name;
+        return this.ckanDataset.title + " - " + this.ckanResource.format;
```
tephenavies, line 145: Switched the order that these are checked so the default is checked last. Otherwise setting `useCombinationNameWhereMultipleResources` without setting `useDatasetNameAndFormatWhereMultipleResources` to false doesn't do anything

soyarsauce: This would better be inline comments in ckanitemreference so it's not lost to the PR

soyarsauce (approved):
image

Merging as code commit changed passed

Issue #4910, opened 2020-10-27
load*! loadMapItems | loadMetadata | loadReference | loadChartItems


soyarsauce:
No action/"issue" on this at the moment, just to capture some things surrounding this.

Notes for load*!

load*! loadMapItems | loadMetadata | loadReference | loadChartItems

When writing the mobx branch ,
- every mixin basically ended up implementing an interface
- ended up with stacks of interfaces
- it was clear you wouldn't use the interface outside of the mixin.

`mappable` predates the above, which is why mappable the interface, and asyncmappablemixin 

Do we need Mappable (when considering both `AsyncMappableMixin` & `Mappable`)?
In some cases we have things that are using asyncmappablemixin, but aren't async
Should we just always use `asyncmappablemixin`

In UI land, might be useful to abstract all these load paths to one single load method

We always call loadchartitems. if chart item getter is not empty then we show charts. There's no special mode where you only show a chart. e.g. csv items that aren't mappable (but we want to chart)

In v8, we show charts by using the chart item getter.
CSV, SOS - mix in the tablemixin. TableMixin defines chart item, so for csv/sos they don't need their own.

tangent on chart item - we used to have a button in v7 which is displayed , if something is in workbench, which basically boiled to: "if workbenchitem is not mappable, then it must be chartable, show an icon".

WPS has helper function that loads metadata and map items

PR #4911, opened 2020-10-27
Remove isMixedInto from AsyncMappable and AsyncChartable

nf-s:
### Remove isMixedInto from AsyncMappable and AsyncChartable

Related to #4910

Remove confusion between checking `Mappable.is` or `AsyncMappable.isMixedInto` (and for `Chartable`)

### Checklist


nf-s:
I think we still need `isMixedInto` for these.

For example, how do I check if a catalog member has `isLoadingMapItems`?

PR #4919, opened 2020-10-27
Workbench loading issues 

### LINKED ISSUE #4719: Adding items to workbench isn't consistent across code-base
Maybe we should replace all instances of `workbench.add` with `addToWorkbench` as this function includes all loading of mappable/chartable...
Or we could combine `addToWorkbench` with `workbench.add`.

**Why**:
- It is confusing
- We seem to run into loading issues every now and then, for example drag-dropping files wasn't triggering `loadMetadata` or `loadMapItems`. Similar issue with splitting items.
- Some catalog items recreate functionality of `addToWorkbench` -  eg WPS
- If we want to change how loading catalog items works in the future, it is difficult because there are calls scattered around, whereas most could be removed if we moved everything into `workbench.add`

Also - we need feedback if catalog member is loading. For example, when you Add a catalog item to the workbench, the data catalog panel will stay visible until it has finished loading.

### LINKED ISSUE #4900: Race condition for displaying data when using &start= ?
There seems to be a race condition for displaying data when using the `start` param in `next`

So for PacificMap we often want to do map embeds using a url like
````
<iframe src="https://map.pacificdata.org/#hideExplorerPanel=1&start={%20%22version%22:%228.0.0%22,%20%22initSources%22:[%20{%20%22stratum%22:%22user%22,%20%22models%22:{%20%22//UserData%22:{%20%22type%22:%22ckan-item%22,%20%22url%22:%20%22https://pacificdata.org/data%22,%20%22datasetId%22:%20%22e5027bdf-ab88-4473-8148-f04e5a67b083%22,%20%22resourceId%22:%20%22e5fbdd9a-01a9-4fd9-bd82-9bd3bfd90d6a%22%20}%20},%20%22workbench%22:[%22//UserData%22]%20}%20]%20}"></iframe>
````

or with the start component slightly more legibly
````
{
    "version":"8.0.0",
    "initSources":[
        {
            "stratum":"user",
            "models":{
                "//UserData":{
                    "type":"ckan-item",
                    "url": "https://pacificdata.org/data",
                    "datasetId": "e5027bdf-ab88-4473-8148-f04e5a67b083",
                    "resourceId": "e5fbdd9a-01a9-4fd9-bd82-9bd3bfd90d6a"
                }
            },
            "workbench":["//UserData"]
        }
    ]
}
````
The data is being added to the workbench and I see the geojson being requested, however it's not added to the map. If I then open the 'About Data' for the catalog item it will appear on the preview map and regular map.

So I'm guessing perhaps the catalog item is ready before the map is or something like that???

### PR DESCRIPTION
**Pull request by nf-s**
### Fix async bug with loading `ReferenceMixin` in `initSources` and removed `addToWorkbench`

Fixes #4900

Pacific map were having an issue with embedding maps which use CKAN references - the problem was in how `loadReference()` and `loadMapItems()` were called in `applyInitData`.

For example http://ci.terria.io/workbench-loading-issues/#hideExplorerPanel=1&start={%20%22version%22:%228.0.0%22,%20%22initSources%22:[%20{%20%22stratum%22:%22user%22,%20%22models%22:{%20%22//UserData%22:{%20%22type%22:%22ckan-item%22,%20%22url%22:%20%22https://pacificdata.org/data%22,%20%22datasetId%22:%20%22e5027bdf-ab88-4473-8148-f04e5a67b083%22,%20%22resourceId%22:%20%22e5fbdd9a-01a9-4fd9-bd82-9bd3bfd90d6a%22%20}%20},%20%22workbench%22:[%22//UserData%22]%20}%20]%20}

This is now fixed.

I also reworked `addToWorkbench` with `workbench.add` - This fixes issue "Adding items to workbench isn't consistent across code-base" #4719 

### Checklist

### CODE REVIEW

File: lib/ReactViews/Custom/Chart/ChartExpandAndDownloadButtons.tsx [hunk 1]
```diff
@@ -70,8 +70,9 @@ class ChartExpandAndDownloadButtons extends React.Component<PropsType> {
             }
           });
         });
-        workbench.add(itemToExpand);
+
         try {
+          workbench.add(itemToExpand);
```
tephenavies: `expandItem` isn't an `async` function so you can't catch promise rejections from `workbench.add`. I think `workbench.add` should be outside the the `try` so that it's clear that no exceptions are expected to be thrown from it.

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
```diff
@@ -54,21 +53,19 @@ export const DataCatalogItem = observer(
     },
 
     toggleEnable(event) {
-      runInAction(() => {
+      runInAction(async () => {
```
tephenavies: `runInAction(async` is never a good mix. There are some good ideas to fix this here: https://mobx.js.org/actions.html#asynchronous-actions

lib/ReactViews/DataCatalog/DataCatalogReference.jsx
```diff
@@ -69,30 +69,22 @@ const DataCatalogReference = observer(
...
-          this.props.terria.timelineStack.addToTop(this.props.reference);
-        } else {
-          this.props.terria.timelineStack.remove(this.props.reference);
-        }
+        try {
+          if (!this.props.terria.workbench.contains(this.props.reference)) {
+            await this.props.terria.workbench.add(this.props.reference);
+          } else {
+            await this.props.terria.workbench.remove(this.props.reference);
+          }
```
tephenavies: What happened to the timeline stuff?

lib/ReactViews/Preview/MappablePreview.jsx
```diff
@@ -43,25 +42,22 @@ class MappablePreview extends React.Component {
   };
 
   @action.bound
-  toggleOnMap(event) {
+  async toggleOnMap(event) {
```
tephenavies: Again action + async.

lib/ReactViews/Preview/MappablePreview.jsx
```diff
@@ -43,25 +42,22 @@ class MappablePreview extends React.Component {
...
-      toAdd
-    ).then(() => {
+    try {
+      if (toAdd) {
+        this.props.terria.timelineStack.addToTop(this.props.previewed);
+        await this.props.terria.workbench.add(this.props.previewed);
+      } else {
+        this.props.terria.timelineStack.remove(this.props.previewed);
+        this.props.terria.workbench.remove(this.props.previewed);
+      }
```
tephenavies, line 54: This is common to a lot of files. Maybe it should be a function in a separate file `addToOrRemoveFromWorkbenchAndTimelineStack`.

tephenavies (changes requested):
Could you add some tests to WorkbenchSpec.ts? Something for resolving references, something for adding the same item twice (it shouldn't do anything the second time). I think there might be bug if you add an item twice that has a `sourceReference`, so a test for that would be good.

lib/Models/Workbench.ts
```diff
@@ -150,6 +155,59 @@ export default class Workbench {
...
+   * be removed from the workbench. If it is mappable, `loadMapItems` will be called.
+   * If it is chartable, `loadChartItems` will be called.
+   *
+   * @param item The item to add to or remove from the workbench.
+   */
+  public async add(item: BaseModel | BaseModel[]): Promise<void> {
+    if (Array.isArray(item)) {
+      return await Promise.all(item.map(i => this.add(i))).then(
+        () => undefined
+      );
```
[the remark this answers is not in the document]
nf-s: yep definitely!

lib/ReactViews/DataCatalog/DataCatalogReference.jsx
nf-s: my bad, just re-added it

lib/ReactViews/Custom/Chart/ChartExpandAndDownloadButtons.tsx
nf-s: 👍

lib/ReactViews/DataCatalog/DataCatalogItem.jsx
nf-s: Yep I forgot about that, thanks for picking it up - `runInAction` wasn't needed here so I've removed it

lib/ReactViews/Preview/MappablePreview.jsx
nf-s: fixed

lib/ReactViews/Preview/MappablePreview.jsx
nf-s: 😑

nf-s:
Let me know if you like my tests @steve9164

AnaBelgun:
High priority because it blocks race condition / PacificMap

lib/Models/Terria.ts
```diff
@@ -857,18 +855,18 @@ export default class Terria {
...
+        return Promise.all(
+          newItems.map(async model => {
+            if (ReferenceMixin.is(model)) {
+              await model.loadReference();
+              model = model.target || model;
+            }
 
-        return Promise.all(promises).then(() => undefined);
+            if (Mappable.is(model)) {
+              await model.loadMapItems();
```
tephenavies, line 905: This could be quite slow. What's the problem with loading multiple workbench items at once?

tephenavies: Ohhhh. Whoops. Just realised what this is doing. It synchronises loadReference -> loadMap for each, but runs every item in parallel to every other item. That's exactly what it should do.

tephenavies (approved):
Awesome. Thanks for those tests.

PR #4923, opened 2020-10-28
WMS legend improvements for default styles

### LINKED ISSUE #1513: Picking the first legend for WMS isn't always right
Sometimes this happens:
1. Layer has a default style, which is not the first in the list
2. When we load GetCaps, we get the legend from the first style (which includes a PALETTE parameter)
3. When we request tiles, we don't pass that parameter, and get the first style.

Hence: a legend that doesn't match the tile style.

http://nationalmap.research.nicta.com.au/#share=CAot5S

nf-s:
Fixes https://github.com/TerriaJS/terriajs/issues/1513

### WMS legend improvements for default styles

For some reason WMS doesn't include information about the default style.

Before we assumed that the first style was default, but only used it to request legend and didn't use it to request tiles &mdash; sometimes this resulted in the legend not matching the actual style of the tiles.

Now, if no style is specified (ie we are using the default style) we will try to request a legend with no `style` parameter (**not the first style in the list**).

### Added transparency and Geoserver legend options (thanks @zoran995)

If the WMS `isGeoserver` then we can add a few options to make legend more pretty

### Checklist


nf-s:
From [OGC WMS-SLD spec](https://www.ogc.org/standards/sld) &mdash; about `style` property for `GetLegendGraphic` request:

> Style of layer for which to produce legend graphic. **If not present, the default style is selected.** The style may be any valid style available for a layer, including non-SLD internally-defined styles.

zoran995:
this triggered me to notice missing transparent legend option for WMS from Geoserver. It was implemented in #1373.

lib/ReactViews/Workbench/Controls/DimensionSelectorSection.tsx
```diff
@@ -71,7 +71,7 @@ export default class DimensionSelectorSection extends React.Component<{
               {(typeof dim.selectedId === "undefined" ||
                 dim.allowUndefined) && (
                 <option key="__undefined__" value="">
-                  Not specified
+                  {dim.undefinedLabel ?? "Not specified"}
```
KeyboardSounds: You should probably i18n-ify the "Not specified" here

nf-s: ah yep, good point

nf-s: done

nf-s:
~Actually before we merge this I might merge in https://github.com/TerriaJS/terriajs/pull/4939~

Done

KeyboardSounds (approved):

PR #4947, opened 2020-10-30
Correctly load basemap and viewermode from init files

### LINKED ISSUE #4265: Mobx baseMap and viewer from init file
Loading of basemap from init file is not ported from master, so currently it is defined only from user local preferences and if its first time user then the first one from the list is used. 
The viewer mode option is always set to the value from the init file, although there is a value in local preferences.
This is mainly so it doesn't end up forgotten :)

zoran995:
### What this PR does

Fixes #4265

This fixes an issue of not loading base maps from init files. It also fixes the loading of viewerMode not override the persisted viewerMode. There is a change of using baseMapId instead of baseMapName, this is done this way to avoid using `@ts-ignore`, or update core classes not related to this on a deep level.

### Checklist


lib/Models/Terria.ts
```diff
@@ -516,8 +520,14 @@ export default class Terria {
       this.mainViewer.baseMap = baseMapSearch.mappable;
     } else {
       console.error(
-        `Couldn't find a basemap for unique id ${persistedBaseMapId}`
+        `Couldn't find a basemap for unique id ${persistedBaseMapId}. Trying to load init base map.`
```
na9da, line 560: For clarity, it might be good to rename this method as `loadPersistedOrInitBaseMap()`

na9da:
Looks great @zoran995!
Just the minor suggestion I made and we are good to merge.

lib/Models/Terria.ts
zoran995: changed

CHANGES.md
```diff
@@ -5,6 +5,7 @@ Change Log
 
 
 #### next release (8.0.0-alpha.60)
+* Fixed an issue with not loading the base map from init file and an issue with viewerMode from init files overriding the persisted viewerMode
 * [The next improvement]
```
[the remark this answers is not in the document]
zoran995: done

na9da (approved):
Approving!

PR #4953, opened 2020-11-02
Revert routing for longer review

tephenavies:
### What this PR does

Revert routing PR to fix CI & relative URL issues. We should put the routing changes through a longer review process to consider all of the impacts of those changes.

### Checklist

-   [ ] Fix up changelog


nf-s (approved):

nf-s:
https://github.com/TerriaJS/terriajs/pull/4679 has broken quite a few things:

- CI doesn't work - http://ci.terria.io/next/ is requesting http://ci.terria.io/build/TerriaMap.js instead of http://ci.terria.io/next/build/TerriaMap.js

- the default proxy URL is `proxy/` not `/proxy/` - so if I add a WMS item from the catalog - my legend has URL: http://localhost:3001/catalog/proxy/_0d/...

- If you go to http://localhost:3001/catalog and then click a WMS group the catalog disappears

- How do we host terria statically? Can we disable this?

- Does this issue - https://github.com/TerriaJS/catalog-converter/issues/19 - mean all catalog URLs have to be absolute now? We shouldn't have to write absolute URLs

I think https://github.com/TerriaJS/terriajs/pull/4679 and the TerriaMap PR (https://github.com/TerriaJS/TerriaMap/pull/500) should be reverted, it needs more to be more thought out and more thoroughly tested.

PR #4957, opened 2020-11-03
Fix missing padlock

reginapramesti:
### What this PR does

Fixes https://github.com/TerriaJS/nsw-digital-twin/issues/464

NSW Digital Twin now fetches straight from portal for datasets with access control restrictions. So far access control has only been implemented on `MagdaReference` items, so this PR implements it on `ArcGisPortalCatalogGroup` and `ArcGisPortalItemReference`.

### Checklist

-   [x] ~~There are unit tests to verify my changes are correct or~~ unit tests aren't applicable (if so, write quick reason why unit tests don't exist): it's a small bug fix


rowanwins:
As I read it this PR will apply the padlock to all items which come from portal which probably isn't ideal as some content is public. For  the current real-world use in the NSW Digital Twin this is probably ok because we're only retrieving secured items via the `ArcGisPortalCatalogGroup`, however it is possible to use it for public items and so the public isn't really relevant in those cases.

Perhaps it ought to be a trait that can be set?

reginapramesti:
I'm a bit confused, probably because I don't fully know how everything is fully integrated. What do you mean by applying the padlock to all items? 

At the moment it gives the portal items an `access` property from the `AccessControlMixin` (just like how it's done on `MagdaReference`), to determine whether it's private/public. From this, `"public"` datasets wouldn't have the padlock whereas private (anything else that isn't `"public"`) datasets would. I figured the implementation of whether or not it's visible by the public is handled elsewhere but I could be wrong? 🤔 how would it work using a trait?

rowanwins:
Another thought... The `ArcGisPortalItemReference` and `ArcGisPortalGroup` should be able to determine if an item has a padlock or not as they are created.

The [REST API documents](https://developers.arcgis.com/rest/users-groups-and-items/item.htm) an `access` property on a portal item. The same can also be done [groups API](https://developers.arcgis.com/rest/users-groups-and-items/group.htm) which also has an  `access` property

So I think perhaps you've done the first step of adding the mixin, and now you just need to populate the individual items with whether they are padlocked or not...

reginapramesti:
I think that's already handled by the mixin itself through the `isPrivate` get function, and the UI already handles the adding of padlocks to private datasets

lib/Models/ArcGisPortalItemReference.ts
```diff
@@ -6,6 +6,7 @@ import URI from "urijs";
 import isDefined from "../Core/isDefined";
 import { JsonObject } from "../Core/Json";
 import loadJson from "../Core/loadJson";
+import AccessControlMixin from "../ModelMixins/AccessControlMixin";
```
KeyboardSounds, line 9: What does the AccessControlMixin do here in the Reference version of the CatalogGroup?

reginapramesti: It's so that we can call the `setAccessType` function in `ArcGisPortalCatalogGroup`, which allows us to give the catalog item an `access` property (`"public"/"private"/etc`). Basically with this mixin for each catalog item we can assign it an access value based on its access control settings in portal. By default `AccessControlMixin` is already incorporated in a `CatalogMemberMixin`, so I didn't have to add it to `ArcGisPortalCatalogGroup`, but not in references like `MagdaReference` and `ArcGisPortalItemReference`, so I had to add that mixin (just like how it was done in `MagdaReference`).

KeyboardSounds (approved):

AnaBelgun:
this is medium only because of Regina's availability constraint

PR #4959, opened 2020-11-03
Replace Google Groups and Gitter with Github Discussions

tephenavies:


soyarsauce (approved):
Thanks for tidying this up! (& again for setting it up to begin with)

PR #4961, opened 2020-11-04
Mobx tjs new routing v2

soyarsauce:
### TODO
- [ ] Revert the branch reference in https://github.com/TerriaJS/terriajs/pull/4961/commits/a7c620708c555b2ec9fda6418ba718b92bfeaaf5
- [x] add base href information to BrowserRouter
- [x] (new) - update all the default url references that don't have a leading slash, like `proxydomains/` to include a leading slash

- [ ] revert the hardcoded values in https://github.com/TerriaJS/terriajs/pull/4961/commits/899e663ae9ac37d7c2fdd30b19b7deac4004aaa1#diff-0f560813c9eb05adde52919f9ab57c87f8af770d2b6e5ea83a7569c4625d23bcR495-R499 - comments are there for a suggested way (means we need a terriajs-server update, to expose the `appBaseUrl` on `/serverconfig/`

~~- [ ] figure out the approach for disclaimer-enabled maps~~

Related:
- Disclaimer https://github.com/TerriaJS/terriajs/issues/4928

### What this PR does

Addressing the concerns coming out of https://github.com/TerriaJS/terriajs/pull/4953

>#4679 has broken quite a few things:
    CI doesn't work - http://ci.terria.io/next/ is requesting http://ci.terria.io/build/TerriaMap.js instead of http://ci.terria.io/next/build/TerriaMap.js

This is simply a matter of updating CI to utilise the base href mechanism detailed out at https://github.com/TerriaJS/terriajs/blob/8.0.0-alpha.59/doc/deploying/deploying-terriamap.md

>Ensure that you specify the baseHref path in your devserverconfig.json if you are serving your TerriaMap from a directory, and an appBaseUrl so a sitemap & other links can be generated.

>    the default proxy URL is proxy/ not /proxy/ - so if I add a WMS item from the catalog - my legend has URL: http://localhost:3001/catalog/proxy/_0d/...
    If you go to http://localhost:3001/catalog and then click a WMS group the catalog disappears

Fixing the above basehref and a tweak on proxy to use `/proxy/` will address this

>    How do we host terria statically? Can we disable this?

There is nothing stopping you from using terria statically, the app will still load at the base url. The only thing you could get tripped on, is copying a URL that isn't `/` (e.g. `/catalog/[foo]`)and trying to share that , if it is not configured to do SPA routing (e.g. serve up index.html for known app routes). In fact, the entire feature right now is opt-in anyway, you will not prerender unless you update terriamap to allow for it.

>    Does this issue - TerriaJS/catalog-converter#19 - mean all catalog URLs have to be absolute now? We shouldn't have to write absolute URLs

No. Relative URLs should be used wherever possible and whenever it makes sense, see my comment at https://github.com/TerriaJS/catalog-converter/issues/19#issuecomment-721451168

> I think #4679 and the TerriaMap PR (TerriaJS/TerriaMap#500) should be reverted, it needs more to be more thought out and more thoroughly tested.

I believe all the above should make sense, but let me address the CI point and demonstrate everything working here!

### Checklist

-   [ ] n/a There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [ ] n/a I've updated CHANGES.md with what I changed.


nf-s:
> Fixing the above basehref and a tweak on proxy to use /proxy/ will address this

I don't think we should change the default proxy URL - won't it will break stuff if people having terria sitting on a route other than `/`? - for example `www.something.com/terria`

> There is nothing stopping you from using terria statically, the app will still load at the base url. The only thing you could get tripped on, is copying a URL that isn't / (e.g. /catalog/[foo])and trying to share that , if it is not configured to do SPA routing (e.g. serve up index.html for known app routes). In fact, the entire feature right now is opt-in anyway, you will not prerender unless you update terriamap to allow for it.

By "opt-in" does that mean it is disabled by default on TerriaMap? I think we need to be able to disable it, and have all links work for static sites

soyarsauce:
Thanks @nf-s for extra Qs - I attempt to address below

### Proxy

>I don't think we should change the default proxy URL - won't it will break stuff if people having terria sitting on a route other than /? - for example www.something.com/terria

That will depend on how they have things routed. For instance,
1. if they are serving off `www.something.com/terria`, and terriajs-server is also served/configured off `/terria/` - (e.g. you only reach proxy through `www.somethign.com/terria/proxy/`)

then it makes sense (in the frontend) for references to the proxy to be using `/proxy/`.
base href of `/terria/`
frontend request to `/proxy/` will go to `/terria/proxy/`

2. if they are serving off `www.something.com/terria`, but the "terria application"/terriajs-server is proxied through `/` - (e.g. you only reach proxy through `www.something.com/proxy/`)

then this is a very strange, non-standard configuration that they are savvy enough to navigate given they were able to break them apart like that, and we can deal with it if that comes - one case that comes to mind is nationalmap and /investormap/ case, but that is basically case #1 in 2 parallel terria applications rather than #2 here.

3. if they are statically hosting off `www.something.com/terria`

then they aren't running terriajs-server, there is no proxy anyway, simple!


### Opt-in

>By "opt-in" does that mean it is disabled by default on TerriaMap? I think we need to be able to disable it, and have all links work for static sites

I see where you're getting at here now! The _prerendering_ portion is opt-in, but the _routing_ is not. I've tried my darn hardest to make this as progressively enhanced as possible, coming from a place of me hating it when there are major breaking changes in other libs without good reason!

It will be a maintenance nightmare to support turning routing off as we'd get two sources of truth that we would develop against for a very small use case (although important case). I think that's the right balance to strike here, it's not unreasonable to expect users using a single page app to configure their static hosting so that they can get the most out of the app, and is very straightforward. A statically hosted terria application is already severely limited.

Lots of resources if you just search "static hosting single page routing [insert static hosting option here]"

Just a few:
apache: https://gist.github.com/davemackintosh/8ac35fff747dbf2c95e1
nginx: https://gist.github.com/dimitardanailov/7a7c4e3be9e03d1b578a
s3: https://gist.github.com/bradwestfall/b5b0e450015dbc9b4e56e5f398df48ff

nf-s:
Ah ok, yeah using `${baseHref}${proxyUrl}` makes sense to me - but currently the baseHref isn't used for the proxy url right?

I don't agree with a few points about the routing:

- I don't think static terria sites are that limited - they really only miss the proxy and have limited share

- I would be nice if people didn't have to setup single page routing for static sites

But if it is too hard to support both then I guess that is fine

soyarsauce:
OK - let me dig a bit, I think I might be able to get something halfway for alt routing avenues

soyarsauce:
Hey @nf-s - [I've addressed your concerns about affecting static hosting here (more details in the ADR)](https://github.com/TerriaJS/terriajs/pull/4961/commits/fd85003f3e0db790f1066f65ecb2faf49f62d6d6), and varied on `experimentalFeatures` - however I'm thinking we could go one step further and give this feature its own feature flag (`browserRouting` or something). What do you think, of both the flag and this general approach for more full compatibility of static-hosted maps?

soyarsauce:
~Still todo:~

moved to pr description 

~- [ ] add base href information to BrowserRouter~
~- [ ] (new) - update all the default url references that don't have a leading slash, like `proxydomains/` to include a leading slash~

zoran995:
Hi @soyarsauce 
I just tried to test these changes to see how they would work in our environment, but having troubles with building and starting this. It seems a little unstable, but probably there is some stuff that I didn't understand correctly.
I have an app running behind Nginx routing to the localhost address.
1. Setting the `baseHref: "/portal/"` throws the `Uncaught SyntaxError: Unexpected token '<'` in the console for localhost access and it is returning the content of `index.ejs` as `TerriaMap.js`.
2. Setting the `appBaseUrl` without prerendering completed throws the following (didn't test with prerendering completed)
```
Uncaught TypeError: Cannot read property 'react' of undefined
    at useTranslation (useTranslation.js?f4aa:36)
```
3. What is the standard build time with prerendering (npm run gulp release), started it locally on windows and it didn't finish after more than 2 hours.
4. Also looking at the ci it looks a bit odd to have a server application on the URL (http://ci.terria.io/mobx-tjs-new-routing-v2/) and then have catalog URLs as the http://ci.terria.io/catalog/. It will for sure cause issues, wouldn't it be much better to have it as http://ci.terria.io/mobx-tjs-new-routing-v2/catalog/.

soyarsauce:
Hi @zoran995 

Yes this is still very WIP, but getting close, thanks so much for playing around with it in the meantime.

>1 Setting the baseHref: "/portal/" throws the Uncaught SyntaxError: Unexpected token '<' in the console for localhost access and it is returning the content of index.ejs as TerriaMap.js.

`index.ejs` shouldn't get served up, it should be the 'compiled' index.html (which just adds the base name at build time)

>2 Setting the appBaseUrl without prerendering completed throws the following (didn't test with prerendering completed)

Could you elaborate a little more on what pre and post "prerendering complete" means here?

> 3 What is the standard build time with prerendering (npm run gulp release), started it locally on windows and it didn't finish after more than 2 hours.

Definitely strange one, how many routes/pages/how big is your catalog? We have a map that has about 100-odd pages/catalog items that prerender, and doesn't take nearly as long - it was maybe ~ 2-3 minutes to build before prerender, and about 5 ~ 10 minutes after adding the prerender step depending on the specs of machine used. edit: I think adding some more feedback during the steps somehow will alleviate the "sit there and wonder if it's actually progressing" scenario.

Are you using direct windows or WSL windows? Kind of sounds like the prerender has stalled, maybe something to look into. Have a look at https://github.com/TerriaJS/terriajs/blob/66f4080147e7fb109f0be54ddf53e14761778b22/doc/deploying/prerender-troubleshooting.md if under WSL

>4 Also looking at the ci it looks a bit odd to have a server application on the URL (http://ci.terria.io/mobx-tjs-new-routing-v2/) and then have catalog URLs as the http://ci.terria.io/catalog/. It will for sure cause issues, wouldn't it be much better to have it as http://ci.terria.io/mobx-tjs-new-routing-v2/catalog/.

Yes - this is what my todo "`add base href information to BrowserRouter`" will fix, it should most definitely be your latter example.

soyarsauce:
Finally I should add - our approach for this will be to add it as an opt-in, especially as we don't have any sort of "migration guide"-esque, general guide, to how all the bits fit together just yet. The knowledge atm is kind of internal to the team and unfortunately, even more specifically, me at the moment which I seek to address after the actual implementation is all sound.

zoran995:
1. you are right my bad. 
I went that it will serve the application using the base href for localhost access also. It looks for the `TerriaMap.js` on the http://127.0.0.1:3001/portal/build/TerriaMap.js, and receives the index.html, while it is being served on http://127.0.0.1:3001/build/TerriaMap.js. Are the localhost URLs expected to work with this setup? 
2. My bad for this, wrong config.json
3. Using the Windows 10 directly. Will do little more debugging for this, and return with findings.

Also, is it default behaviour that the user be just redirected to the base application if the catalog URL couldn't be resolved or there should be some sort of warning, "this catalog item is not available you can keep using the application normally"?

zoran995:
Update on 3. This is connected to the 1. Running the `npm run gulp release` with `baseHref` other than `/` will cause the prerendering to keep running forever since it can't resolve the localhost.
Prerendering completes in under than 2 minutes on my computer but can't use the baseHref.

tephenavies:
We should try to progress and merge this in April. It looks close to mergeable. More work is needed on synchronising explorer panel state and URL state (one should derive from the other if possible)

E.g. `viewState.closeCatalog()` and `viewState.history?.push(ROOT_ROUTE)`

tephenavies:
- [ ] Check whether https://github.com/TerriaJS/terriajs/issues/5486 is related.

tephenavies:
In some cases opening the catalog using `/catalog/X` routing circumvents certain loading processes. This can lead to `loadReferences` not being called on references, and an "Unable to resolve reference" message, or in other cases tabbed catalogs don't open to the correct tab because ancestor information is built up by `loadMembers`, which is done when opening groups in the catalog.

Another thing to consider is magda catalogs. Loading terria at `/catalog/X` only works if we can load all ancestors of `X`. This isn't possible with magda catalogs (this is the same problem as catalog re-organisation breaking old share links). Possibly if we always use MemoryRouter with magda catalogs we can call the correct load functions in the correct place such that we still use the fake `/catalog/X` route to store state for navigating through the catalog (in fact this should be similar/identical? to our current approach).

We need to add tests for each of these situations, and many others, before this is merged.

wibily:
requires unit tests and de-duping state between routing and view

wibily:
Ana: don't break share links
Nanda: catalog routes should be indexable but when you launch it, it should set state. Catalogs to be entry points to the app. Terria shouldn't try and change the url when you navigate to different areas. to be discussed during tech discussion meeting

AnaBelgun:
Connected to: https://github.com/TerriaJS/product/issues/117#issuecomment-977577639

zoran995:
hi @steve9164, what happened to this PR, is it removed from terria scope for future?

tephenavies:
Hi @zoran995. We decided that we didn't want URL routes inside terria for the catalog. Instead we will likely support `/catalog/:id` on Terria application load which will show the specified member inside the catalog (part of #6151, more details and the ability to reproduce on CI coming shortly). Building pre-rendering on top of these routes should be possible.

zoran995:
thanks, just wondered as it got closed without any comment, and pretty sure that a lot of work went in into it

PR #4968, opened 2020-11-05
Various fixes for droughtmap

na9da:
### What this PR does

Various fixes for drought map release.

* Added `DataCustodianTraits` to `WebMapServiceCatalogItem`
* Added `disableDimensionSelectors` trait to `WebMapServiceCatalogItem`. Acheives the same effect of `disableUserChanges` in v7.
* Temporarily stopped using `papaparse` for fetching Csv urls till an upstream bug is fixed.

### Checklist


lib/Table/Csv.ts
```diff
@@ -52,12 +53,21 @@ export default class Csv {
...
+      }
     });
+    // There is currently a bug when using papaparse to fetch CSV URLs
+    // See: https://github.com/mholt/PapaParse/pull/832
+    // return new Promise<string[][]>((resolve, reject) => {
+    //   papaparse.parse(url, {
+    //     ...getParseOptions(columnMajor, resolve, reject),
+    //     download: true,
+    //   });
+    // });
```
soyarsauce, line 70: Perhaps a changelog entry to reflect this change

lib/Traits/WebMapServiceCatalogItemTraits.ts
```diff
@@ -232,10 +234,10 @@ export default class WebMapServiceCatalogItemTraits extends mixTraits(
 
   @primitiveTrait({
     type: "boolean",
-    name: "Disable style selector",
-    description: "When true, disables the style selector in the workbench"
+    name: "Disable dimension selectors",
+    description: "Disables the dimension selectors when true"
```
soyarsauce: Still keep the "in the workbench" wording?
The way I see it, if I was using the catalog editor and reading this description, and didn't know what a dimension selector was, it would be confusing. So at least some hint that it's workbench related would be helpful

soyarsauce (changes requested):
Awesome work, just some suggestions surrounding documentation, then let's 🚀

lib/Table/Csv.ts
soyarsauce: Scratch this, I'm so sorry - I missed the last entry in the changelog that covered this!

soyarsauce:
The last code-change passing at https://github.com/TerriaJS/terriajs/runs/1355790491

soyarsauce (approved):

PR #4979, opened 2020-11-10
Improve magda root group behaviour for sharing

soyarsauce:
### What this PR does

Fixes https://github.com/TerriaJS/qld-digital-twin/issues/237 & https://github.com/TerriaJS/terriajs/issues/4978

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@ Change Log
 * Temporarily stopped using `papaparse` for fetching Csv urls till an upstream bug is fixed.
 * Improve handling of `ArcGisMapServerCatalogItem` when dealing with tiled layers.
 * Ensure there aren't more bins than unique values for a `TableStyle`
+* Improves magda based root group mimic behaviour introdcued in 8.0.0-alpha.57 by adding `/` to `knownContainerUniqueIds` when `map-config*` is encountered
 
 #### 8.0.0-alpha.60
 * Fix WMS legend for default styles.
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -75,6 +75,43 @@ import ViewerMode from "./ViewerMode";
 import Workbench from "./Workbench";
 // import overrides from "../Overrides/defaults.jsx";
 
+interface InitModels {
+  [key: string]: {
+    [key: string]: JsonValue;
+    knownContainerUniqueIds: string[];
+  };
+}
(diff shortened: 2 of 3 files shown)


nf-s:
I'm not that happy with this solution, it is applying a quick fix to a quick fix which led to a bug. 

I'd much rather just reverting the change for maps this issue affects - eg just commenting/uncommenting lines for qld/nsw digital twin terriaJS:

https://github.com/TerriaJS/terriajs/blob/69f045dc5f82efb3c91c2a8193a240f968177834/lib/Models/Terria.ts#L920-L923

I believe those maps are already using custom terriaJS versions

soyarsauce:
Hey @nf-s - thanks for taking a look! And I agree it's definitely not ideal at all. However it _is inline_ and a halfway measure to achieving the 'share key'/'alias id' concept in https://github.com/terriajs/terriajs/issues/4774.

Re: QLD/NSW - they've historically used official (alpha) terriajs releases, only recently has QLD diverted slightly for the WIP outlines work. NSW is still using official. SaaS as well.

The problem with reverting it for SaaS however, means we then break sharing in a different way by unfixing what the original quick fix addressed. Do you have any other ideas on how we can move forward without share keys being finalised?

nf-s:
Would we need to revert this change for SaaS? I thought it was working well in SaaS.
In my mind, this revert would only need to happen for QLD (and NSW when the next release occurs - if they upgrade terriajs to the problem version).

I would much prefer a rough `shareKeys` implementation than a fix like this if possible.   
If we had that we could just add `shareKeys = ['/']` for the root magda group.

What do you think?  
I don't have a good idea of how urgent this is right now, so I you think we need this fix now I'm happy to approve it.

soyarsauce:
SaaS has the same problem as QLD(and NSW should it update), in that some of those saasmaps that are magda based (QLD+NSW are "fully magda based maps"). So any share links created before the change, are no longer working at the moment - I've slated today for architecture documentation and was hoping for someone to be able to look after this in the meantime.

None of the current prod saas maps that are magda based are actually in "production" yet, the only production saas maps are utilising other means of catalogue-ing - DE africa is the one magda-based catalog that comes to mind, which is the only reason why I am ok with taking a slower approach for it, but IMO QLD definitely needs action (in the form of a roll back, or a fix like this).

nf-s (approved):

PR #4980, opened 2020-11-10
Merge catalog groups with the same name/id into one group

### LINKED ISSUE #4938: Merge catalog groups with the same name
As a catalog organiser,
I want all references in a catalog to a catalog groups with the same name to be merged
So that I can define a single catalog group's members across multiple catalog.json files.

This works in v7, as seen in the current version of NEII Viewer, but doesn't work in v8.

reginapramesti:
### What this PR does

Fixes #4938 

Implement v7 feature that allows having multiple catalog groups with the same name but different members, to allow having different catalog json files with the same top level group name

### Checklist


tephenavies (approved):
Good to merge @KeyboardSounds or @nf-s when tests pass (tests pass locally).

PR #4982, opened 2020-11-11
4851 chart fixes

### LINKED ISSUE #4851: Missing chart disclaimer for DEA Waterbodies in V8
Same issue as #4841 , but for DEA Waterbodies

V7:
image

V8:
image

na9da:
### What this PR does

Fixes #4851 and more

* Chart disclaimers were not being shown in shared urls, this fixes it 
  Testing: Opening, the test link below should show chart disclaimers

* Fixes a bug where chart disclaimers were shown even for chart items disabled in the workbench - 
  Testing: enabling/disabling a chart item in the workbench should also remove its disclaimer

* Fixes a bug where charts with titles containing the text "lat" or "lon" were hidden in feature info panel.
  Testing: Feature info panel opened when clicking a coastal dot should show a chart (previously it was empty because the name contains "Relative" which includes the fragment "lat")

Test link: http://ci.terria.io/4851-chart-fixes/#share=s-k901bmQ3p0S3IBQ1GDEEzFdANqD

### Checklist


lib/ReactViews/BottomDock/ChartDisclaimer.tsx
```diff
@@ -22,17 +22,21 @@ const ChartDisclaimer: React.FC<ChartDisclaimerProps> = ({ terria }) => {
   const uniqueChartDisclaimers: string[] = [
     ...new Set(
       filterOutUndefined(
-        chartView.chartableItems.map(item =>
-          hasTraits(item, DiscretelyTimeVaryingTraits, "chartDisclaimer")
-            ? item.chartDisclaimer
+        chartView.chartItems.map(chartItem =>
+          chartItem.showInChartPanel &&
+          hasTraits(
```
na9da, line 27: Collect disclaimers only for items that are shown in chart panel.

lib/ReactViews/BottomDock/ChartDisclaimer.tsx
```diff
@@ -22,17 +22,21 @@ const ChartDisclaimer: React.FC<ChartDisclaimerProps> = ({ terria }) => {
...
+          )
+            ? chartItem.item.chartDisclaimer
             : undefined
         )
       )
     )
   ];
-  const chartItems = chartView.chartItems.filter(c => c.showInChartPanel);
-  if (uniqueChartDisclaimers.length === 0 || chartItems.length === 0)
-    return null;
```
na9da, line 35: We don't need the `chartItems.length` test as the above change covers it.

lib/ReactViews/Custom/Chart/FeatureInfoPanelChart.jsx
```diff
@@ -34,10 +34,7 @@ class FeatureInfoPanelChart extends React.Component {
   @computed
   get chartItem() {
     return this.props.item.chartItems.find(
-      chartItem =>
-        chartItem.type === "line" &&
-        chartItem.name.indexOf("lat") === -1 &&
-        chartItem.name.indexOf("lon") === -1
+      chartItem => chartItem.type === "line"
```
na9da, line 37: Not sure why we were doing this, but this hides charts with names "Relative something" from the chart panel.
Seems to be related to - https://github.com/TerriaJS/terriajs/commit/79efd766296b89af6f2b4b779fe708b31ff77474

lib/ReactViews/Custom/ChartCustomComponent.ts
```diff
@@ -231,22 +231,9 @@ export default abstract class ChartCustomComponent<
...
-                hasTraits(
-                  item,
-                  DiscretelyTimeVaryingTraits,
-                  "chartDisclaimer"
-                ) &&
-                chartDisclaimer !== undefined
-              ) {
-                item.setTrait(
-                  CommonStrata.definition,
-                  "chartDisclaimer",
```
na9da, line 246: Apart from some refactoring to move setting of `chartDisclaimer` (and other traits from parent) to a common method `setTraitsFromParent`, we also make sure the trait is set in `CommonStrata.user` so that it gets added to the shared url.

tephenavies (approved):
Thanks for all the PR comments.

PR #4984, opened 2020-11-11
Detect Terria embed (in iframe), if so - hide modals/popups/workbench

### LINKED ISSUE #3832: Embed state
Following on from https://github.com/TerriaJS/terriajs/issues/3648 & https://github.com/TerriaJS/terriajs/pull/3826 (& tangentially, https://github.com/TerriaJS/terriajs/issues/3806) We should have some explicit way to check, via viewState, if the map is embedded (simply check if we're running in an iframe) - then we can do things like making sure all the popups don't show even if config files say yes (or at least things like welcome messages - most of the time embedding things is to show some datasets immediately)

nf-s:
### What this PR does

* Added embed detection - it is used to automatically reduce modals/pop-ups if Terria is embedded into a webpage (for example - in an iframe) - which is equivalent to setting `showWelcomeMessage=false` `showInAppGuides=false`, `showFeaturePrompts=false` and `#hideWorkbench=1`.
* Added `ConfigParameters.disableEmbedDetection` and URL parameter `disableEmbedDetection`, which can disable feature previously outlined
* Added the following missing params to `ConfigParameters`: `googleAnalyticsOptions`, `globalDisclaimer`, `printDisclaimer`, `disableSplitter`, `showFeaturePrompts`

I also cleaned up doc\deploying\controlling-with-url-parameters.md and doc\customizing\client-side-config.md. They now have "to be ported form v7" sections

Fixes #3832

#### Without `disableEmbedDetection`

```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="utf-8">
    <title>NationalMap Embed Test</title>
</head>

<body>
    <iframe id="embeddedNationalMap"
        src="http://ci.terria.io/embedded-mode/"
        width="1024"
        height="768"></iframe>
</body>
</html>
```

#### With `disableEmbedDetection`

```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="utf-8">
    <title>NationalMap Embed Test</title>
</head>

<body>
    <iframe id="embeddedNationalMap"
        src="http://ci.terria.io/embedded-mode/#disableEmbedDetection=1"
        width="1024"
        height="768"></iframe>
</body>
</html>
```

### Checklist

-   [x] Added some configParameter, userProperties tests


nf-s:
These failed tests have stumped me. If someone could take a look that would be amazing! 😊

tephenavies:
@steve9164 to talk to @philipgrimmett before reviewing.

nf-s:
Stale PR

PR #4985, opened 2020-11-11
Add `multiplierDefaultDeltaStep` Trait, which tries to calculate sensible multiplier for `DistrectelyTimeVarying` datasets

nf-s:
### Add `multiplierDefaultDeltaStep` Trait, which tries to calculate sensible multiplier for `DistrectelyTimeVarying` datasets

Fixes  https://github.com/PacificCommunity/Terria-PacificMap/issues/53

Add `multiplierDefaultDeltaStep` Trait, which tries to calculate sensible multiplier for `DistrectelyTimeVarying` datasets. By default it is set to 2, which results in a new timestep being displayed every 2 seconds (on average) if timeline is playing.

### Checklist


nf-s:
Close because CI doesn't like uppercase branch NAME 😑 

Moved to here - https://github.com/TerriaJS/terriajs/pull/4986

Issue #4987, opened 2020-11-13
When loading config from magda, initializationUrls should be applied even if `group` aspect is not set


na9da:
Currently when defining magda config, one has to remember to set `group` aspect for `initializationUrls` to be loaded and applied. Ideally, `initializationUrls` should be loaded even when the group is not set.

[Relevant code](https://github.com/TerriaJS/terriajs/blob/next/lib/Models/Terria.ts#L957-L980).

soyarsauce:
Thanks @na9da - as mentioned, may just be an artifact of magda-based-loading. (i think the logic that was initially written probably used the existence of that aspect as one check/hint that it's a magda map, ie https://github.com/TerriaJS/terriajs/blob/8c5397b39cad00be3610e4dc7d21963a9fa077e9/lib/Models/Terria.ts#L957)

soyarsauce:
(& then subsequently, this block gets skipped:)
https://github.com/TerriaJS/terriajs/blob/8c5397b39cad00be3610e4dc7d21963a9fa077e9/lib/Models/Terria.ts#L980-L991

PR #4990, opened 2020-11-16
Magda init urls

KeyboardSounds:
### What this PR does

Fixes #https://github.com/TerriaJS/terriajs/issues/4987

### Checklist

### DIFF
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -971,26 +971,22 @@ export default class Terria {
       reference.setTrait(CommonStrata.definition, "url", magdaRoot);
       reference.setTrait(CommonStrata.definition, "recordId", id);
       reference.setTrait(CommonStrata.definition, "magdaRecord", config);
-      await reference.loadReference().then(() => {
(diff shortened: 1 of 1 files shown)


lib/Models/Terria.ts
```diff
@@ -971,26 +971,22 @@ export default class Terria {
       reference.setTrait(CommonStrata.definition, "url", magdaRoot);
       reference.setTrait(CommonStrata.definition, "recordId", id);
       reference.setTrait(CommonStrata.definition, "magdaRecord", config);
-      await reference.loadReference().then(() => {
```
na9da, line 974: @KeyboardSounds to fix #4987, I think we need to remove the check for `groups` and `groups.members` at https://github.com/TerriaJS/terriajs/blob/891aafb9bfb1ceb153b14931073ab7a599a98bde/lib/Models/Terria.ts#L957

KeyboardSounds: whoops! it would help if I fixed the actual bug 😂

na9da (approved):
Looks good to me @KeyboardSounds

PR #4995, opened 2020-11-18
Adds new config parameter `defaultBaseMapId`

na9da:
### What this PR does

Adds a new config parameter `defaultBaseMapId` which is used as a fallback if no persisted map is found.

Required for https://github.com/TerriaJS/RaPPMap/issues/87

### Checklist


zoran995:
This can't be done through init files like in v7? Or this should be third way do define the basemap? There is #4947 to port it.

na9da:
Ooh! That's right, I forgot that `baseMapName` existed. I'll close this PR in favor of #4947.

PR #4997, opened 2020-11-19
Add trait `showInWorkbenchOnLoad` for all catalog members.

na9da:
### What this PR does

Adds trait `showInWorkbenchOnload` for all catalog members which adds the item to the workbench when the init file is loaded. This is equivalent to setting `isEnabled: true` in the v7.

Required for https://github.com/TerriaJS/RaPPMap/issues/87

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### MobX Development
 
 #### next release (8.0.0-alpha.62)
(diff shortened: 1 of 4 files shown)


na9da:
`CatalogMemberTraits` is probably not the place to add this trait as it also gets mixed into Groups.

na9da:
Closing as it was decided to re-use the existing `workbench` array instead of adding a new trait to achieve this. New issue tracked here - https://github.com/TerriaJS/catalog-converter/issues/26

PR #4999, opened 2020-11-19
Hide empty info sections

na9da:
### What this PR does

Hide info sections with empty content in explorer preview.

eg:

in v7 

image

vs in v8

image

### Checklist


nf-s (approved):
👍

PR #5000, opened 2020-11-19
Fix tabs bug

reginapramesti:
### What this PR does

Enable tabbed catalog to work properly by fixing several bugs, for https://github.com/PacificCommunity/Terria-PacificMap/issues/56 and https://github.com/TerriaJS/drought-map/issues/203

From the discussion that we had over sprint planning, we decided that the same catalog items in different tabs should have different ids 

### Checklist


lib/Models/upsertModelFromJson.ts
```diff
@@ -29,16 +29,12 @@ export default function upsertModelFromJson(
...
-      while (terria.getModelById(BaseModel, uniqueId) !== undefined) {
-        uniqueId = id + "(" + idIncrement + ")";
-        idIncrement++;
-      }
+      uniqueId = (parentId || "") + "/" + localId;
     }
 
+    // To handle duplicate items, e.g. same item in different tabs
+    uniqueId = incrementIfAlreadyExists(uniqueId, terria);
+
```
na9da: The reason this breaks the  [TerriaSpec](https://github.com/TerriaJS/terriajs/blob/next/test/Models/TerriaSpec.ts#L248) is that terria automatically creates a group with id [USER_ADDED_CATEGORY_ID.](https://github.com/TerriaJS/terriajs/blob/next/lib/Models/CatalogNew.ts#L38). When [updateApplicationUrl](https://github.com/TerriaJS/terriajs/blob/next/lib/Models/Terria.ts#L591) is called with a share link containing `USER_ADDED_CATEGORY_ID` group, the above change will auto increment and create `USER_ADDED_CATEGORY_ID (1)` etc. So the group members are restored under a new group instead of the original one.

nf-s (approved):
Thanks for all of your work @reginapramesti !

Sorry @na9da , hope you don't mind me stealing this!
I had some spare time and wanted to get this ready to SaaS update today

PR #5001, opened 2020-11-19
Fix error when previewing an item in a open by default group.

### LINKED ISSUE #4994: `knownContainerUniqueIds` not set correctly for group members when the parent group has `isOpen=true`
Currently `knownContainerUniqueIds` for members is set as side-effect of loading the parent group.

https://github.com/TerriaJS/terriajs/blob/next/lib/ModelMixins/GroupMixin.ts#L69

If the source catalog has groups with `isOpen=true` then the group appears open in the explorer but `loadMembers` for the group is never called and so `knownContainerUniqueIds` for the group members will be empty.

This breaks a few things, for eg, previewing a group item will throw an error here:

https://github.com/TerriaJS/terriajs/blob/next/lib/ReactViewModels/ViewState.ts#L570

na9da:
### What this PR does

Fixes #4994 

Fixes console errors when previewing an item in a open by default group (i.e `isOpen: true` in init file).

The errors are thrown because `loadMembers()` are not called for groups that are already open, so the `knownContainerUniqueIds` are not properly set for the members. 

This change ensures `loadMembers()` is called for all open groups by invoking them in `componentDidMount` and `componentWillMount` hooks of `DataCatalogGroup` component.

### Checklist

### DIFF
--- a/lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
+++ b/lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
@@ -88,6 +88,18 @@ const DataCatalogGroup = observer(
       return url.replace(/^https?:\/\//, "");
     },
 
+    componentDidMount() {
+      if (this.props.group.isOpen) {
+        this.props.group.loadMembers();
+      }
+    },
+
+    componentDidUpdate() {
+      if (this.props.group.isOpen) {
+        this.props.group.loadMembers();
+      }
+    },
+
     render() {
       const group = this.props.group;
       const { t } = this.props;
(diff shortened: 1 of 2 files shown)


lib/ReactViews/DataCatalog/DataCatalogGroup.jsx
```diff
@@ -88,6 +88,18 @@ const DataCatalogGroup = observer(
       return url.replace(/^https?:\/\//, "");
     },
 
+    componentDidMount() {
+      if (this.props.group.isOpen) {
+        this.props.group.loadMembers();
+      }
+    },
```
na9da, line 95: On mount/update ensure that the group is loaded if it is open, this is so that `GroupMixin.refreshKnownContainerIds()` method is run so that the items ancestors are set correctly.

nf-s (approved):

PR #5002, opened 2020-11-19
Fix errors when switching explorer tabs.

na9da:
### What this PR does

Fixes console errors when switching explorer tabs. On tab switch `isOpen` was being set, but `loadMembers` were not getting called. This change use `openGroup` method to correctly perform both. Also switched to `.uniqueId` for tracking tabs instead of `.name`.

### Checklist


na9da:
Closing in favor of https://github.com/TerriaJS/terriajs/pull/5000

PR #5003, opened 2020-11-19
Wps fixes

na9da:
### What this PR does

Fixes WPS issues that came up during testing for https://github.com/TerriaJS/RaPPMap/issues/87

* Now uses `catalog-converter` to convert terriajs json in WPS response from v7 to v8.
* Fixes a bug in `UserDrawing` which caused points to not be plotted on the map.
* Fixes app crash when switching between different types of parameter in `GeoJsonParameterEditor`.
* Fixes several mobx warnings in the WPS codepath.

### Checklist

### DIFF
--- a/lib/ModelMixins/CatalogFunctionJobMixin.ts
+++ b/lib/ModelMixins/CatalogFunctionJobMixin.ts
@@ -130,6 +130,7 @@ function CatalogFunctionJobMixin<
      */
     protected abstract async _invoke(): Promise<boolean>;
 
+    @action
     public async invoke() {
       this.setTrait(CommonStrata.user, "jobStatus", "running");
       try {
@@ -165,6 +166,7 @@ function CatalogFunctionJobMixin<
     /**
      * This function adapts AutoRefreshMixin's refreshData with this Mixin's pollForResults - adding the boolean return value which triggers refresh disable
      */
+    @action
     refreshData() {
       if (this.pollingForResults) {
         return;
@@ -201,6 +203,7 @@ function CatalogFunctionJobMixin<
      * - `pollForResults` returns true {@link CatalogFunctionJobMixin#refreshData}
      * - on `loadMetadata` if `jobStatus` is "finished", and `!downloadedResults`  {@link CatalogFunctionJobMixin#forceLoadMetadata}
      */
+    @action
     private async onJobFinish(addResultsToWorkbench = this.inWorkbench) {
       // Download results when finished
       if (
@@ -289,6 +292,7 @@ function CatalogFunctionJobMixin<
     }
     protected async forceLoadMapItems() {}
 
+    @action
     protected async forceLoadMetadata() {
       if (this.jobStatus === "finished" && !this.downloadedResults) {
         await this.onJobFinish();
--- a/lib/ModelMixins/CatalogFunctionJobMixin.ts
+++ b/lib/ModelMixins/CatalogFunctionJobMixin.ts
@@ -130,11 +130,10 @@ function CatalogFunctionJobMixin<
      */
     protected abstract async _invoke(): Promise<boolean>;
 
-    @action
     public async invoke() {
       this.setTrait(CommonStrata.user, "jobStatus", "running");
       try {
-        const finished = await this._invoke();
+        const finished = await runInAction(() => this._invoke());
         if (finished) {
           this.setTrait(CommonStrata.user, "jobStatus", "finished");
           this.onJobFinish(true);
--- a/lib/Models/FunctionParameters/FunctionParameter.ts
+++ b/lib/Models/FunctionParameters/FunctionParameter.ts
@@ -3,6 +3,7 @@ import combine from "terriajs-cesium/Source/Core/combine";
 import isDefined from "../../Core/isDefined";
 import JsonValue, { JsonObject } from "../../Core/Json";
 import CatalogFunctionMixin from "../../ModelMixins/CatalogFunctionMixin";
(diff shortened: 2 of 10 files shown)


lib/Models/FunctionParameters/FunctionParameter.ts
```diff
@@ -64,11 +65,10 @@ export default abstract class FunctionParameter<
   clearValue(strataId: string) {
     if (isDefined(this.catalogFunction.parameters?.[this.id])) {
       runInAction(() => {
-        delete this.catalogFunction.parameters![this.id];
         this.catalogFunction.setTrait(
           strataId,
           "parameters",
-          this.catalogFunction.parameters
+          omit(this.catalogFunction.parameters, this.id)
         );
```
na9da, line 72: `delete` doesn't work correctly for mobx objects, so we use omit.

lib/Models/UserDrawing.ts
```diff
@@ -126,6 +126,8 @@ export default class UserDrawing extends CreateModel(EmptyTraits) {
     });
   }
 
+  async loadMapItems() {}
+
```
na9da, line 130: workbench/overlay `add` method requires this.

lib/Models/WebProcessingServiceCatalogFunctionJob.ts
```diff
@@ -491,17 +496,21 @@ function formatOutputValue(title: string, value: string | undefined) {
...
+    const { member, messages } = convertMember(input);
+    if (member === null || messages.length > 0)
+      throw new Error("Error converting v7 item to v8");
+    return member;
+  } else {
+    const { result, messages } = convertCatalog(input);
+    if (result === null || messages.length > 0)
+      throw new Error("Error converting v7 catalog to v8");
+    return result;
   }
```
na9da, line 513: Use catalog-converter to properly convert terria json in WPS response. The response object could be a full catalog or a single catalog item.

lib/ReactViews/Analytics/GenericParameterEditor.tsx
```diff
@@ -16,12 +16,13 @@ export default class GenericParameterEditor extends React.Component<{
   }
 
   render() {
+    const value = (this.props.parameter.value || "") as string;
     return (
```
na9da, line 20: Ensure `value` is never `undefined` to prevent React warning about [uncontrolled input](https://reactjs.org/docs/uncontrolled-components.html).

nf-s (changes requested):

lib/ModelMixins/CatalogFunctionJobMixin.ts
```diff
@@ -130,6 +130,7 @@ function CatalogFunctionJobMixin<
      */
     protected abstract async _invoke(): Promise<boolean>;
 
+    @action
```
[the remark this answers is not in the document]
na9da: Yeah possibly! I haven't used it before, but happy to look at it.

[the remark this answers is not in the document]
nf-s: There is an example of it here
https://github.com/TerriaJS/terriajs/blob/6702000db4b9908d1db2e23ef4b7c28a550ed984/lib/ReactViews/DragDropFile.tsx#L29-L66

But it may be easier to get rid of `@action` just wrap all synchronous calls inside the function with `runInAction`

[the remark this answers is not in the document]
na9da: @nf-s I have wrapped the `_invoke()` call in an action and now the warnings are gone. Let me know if that works. I think I had it in a working branch before but somehow missed it when splitting into separate PRs.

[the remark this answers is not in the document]
nf-s: Awesome thanks @na9da !

lib/ModelMixins/CatalogFunctionJobMixin.ts
```diff
@@ -201,6 +202,7 @@ function CatalogFunctionJobMixin<
      * - `pollForResults` returns true {@link CatalogFunctionJobMixin#refreshData}
      * - on `loadMetadata` if `jobStatus` is "finished", and `!downloadedResults`  {@link CatalogFunctionJobMixin#forceLoadMetadata}
      */
+    @action
```
nf-s, line 205: ~Sorry to be a pain @na9da can you please also change this async function to call runInActions where required~

lib/ModelMixins/CatalogFunctionJobMixin.ts
```diff
@@ -289,6 +291,7 @@ function CatalogFunctionJobMixin<
     }
     protected async forceLoadMapItems() {}
 
+    @action
```
nf-s, line 294: ~Does this need an `@action` here?
Sorry I'm not a fan of wrapping async functions in actions - I find it a bit unintuitive where the action starts and then the promise starts with async/await syntax~

nf-s (changes requested):

lib/ModelMixins/CatalogFunctionJobMixin.ts
nf-s: scratch this too

lib/ModelMixins/CatalogFunctionJobMixin.ts
nf-s: actually scratch that - sorry I'm being too pedantic.

nf-s (approved):

PR #5005, opened 2020-11-20
Fix mobx warnings for geojson catalog item.

na9da:
### What this PR does

Wraps bunch of stuff inside `action` to fix mobx warnings when loading a geojson item.

### Checklist

### DIFF
--- a/lib/Models/GeoJsonCatalogItem.ts
+++ b/lib/Models/GeoJsonCatalogItem.ts
@@ -1,5 +1,5 @@
 import i18next from "i18next";
-import { computed, observable, runInAction, toJS } from "mobx";
+import { action, computed, observable, runInAction, toJS } from "mobx";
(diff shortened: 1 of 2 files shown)


lib/Models/GeoJsonCatalogItem.ts
```diff
@@ -119,48 +119,50 @@ class GeoJsonCatalogItem extends AsyncMappableMixin(
...
-                  "</a>",
-                firefox:
-                  '<a href="http://www.mozilla.org/firefox" target="_blank">' +
-                  i18next.t("models.userData.firefox") +
-                  "</a>"
-              })
-            });
+    return new Promise<JsonValue | undefined>(
+      action((resolve, reject) => {
+        if (isDefined(this.geoJsonData)) {
```
na9da, line 124: Indentation change after wrapping in action.

nf-s (approved):

PR #5010, opened 2020-11-24
Fix broken highlighting for raster layers.

### LINKED ISSUE #4983: MobX - Inconsistent Feature Highlighting
Not yet ported to MobX (TBC):
In some of the features when you select the layer the selected polygon is not being highlighted e.g., if you select Murray Darling Region in the Australian Drainage Division Layer then you can see the feature information pop-ups but the boundary is not been selected; https://uat.drought.gov.au/#share=s-cwf9HLHblkoA7BQdQ4zupzaa70C 

**Note this particular layer has been fixed**

image

na9da:
### What this PR does

Fixes #4983 

Fixes broken feature highlighting for raster layers.

### Testing
* Visit http://ci.terria.io/next/#share=s-tNvxWDqG6g0fUBvssB1I7LGDwoC
* Try selecting any of the polygons on the map - they fail to highlight
* Now test again with - http://ci.terria.io/fix-raster-highlight/#share=s-6yBT9A2ISpT3TMu535y1VDIlX8q
* Clicking on any polygon should highlight it
* Clicking another polygon should un-hilight the previous selection and highlight the new one

### Checklist


lib/Models/GlobeOrMap.ts
```diff
@@ -158,9 +158,9 @@ export default abstract class GlobeOrMap {
     rectangle: Rectangle
   ): () => void;
 
-  _highlightFeature(feature: Feature | undefined) {
+  async _highlightFeature(feature: Feature | undefined) {
     if (isDefined(this._removeHighlightCallback)) {
-      this._removeHighlightCallback();
+      await this._removeHighlightCallback();
       this._removeHighlightCallback = undefined;
```
na9da, line 164: `await` is required because if `_removeHighlightCallback` is unset before the callback is complete, then we might break the functioning of the callback and features will fail to un-highlight.

nf-s (changes requested):
Looks good Nanda!

One minor issue, if you select a feature and then remove the layer - the feature doesn't get removed:

image

na9da:
@nf-s Good catch! You snagged an ancient bug from master :grimacing: 

Have fixed it now.

nf-s (approved):

PR #5018, opened 2020-11-26
Nullable time column

na9da:
### What this PR does

Makes `timeColumn` in table styles nullable. Useful for turning off auto-detection of time columns in tables.

### Checklist


tephenavies (dismissed):
This looks good. Can you also add a test that all the correct entities are created in the dataSource when using `null` timeColumn and a timeseries CSV? Or at least the right number of entities. Maybe with this test file: https://github.com/TerriaJS/terriajs/blob/next/wwwroot/test/csv/lat_lon_timeseries.csv

There should be 11 entities created, one for each row (though they'll all be in 2 locations)

nf-s:
AREMI Power generation test URL http://ci.terria.io/nullable-time-column/#clean&https://gist.githubusercontent.com/nf-s/9ff1422d9528c16a9e71403ec720558a/raw/fe86f1c9bf2dd4053148475dd1361fc7e205ec22/aremi-elec-infra.json

nf-s (approved):

PR #5061, opened 2020-12-07
implement here maps geocoding service

zoran995:
### What this PR does

Implements the support for the here maps geocoding service, something we completed a few months ago but never committed to upstream

### Checklist


zoran995:
This has to be implemented from scratch as a new model for location search providers is implemented

PR #4869, opened 2020-10-15
Shapefile drag n drop

KeyboardSounds:
### What this PR does

Fixes #4721

Adds support for dragging and dropping shapefiles into Terria by converting them to GeoJson in the browser.

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [ ] I've updated CHANGES.md with what I changed.

Issue #2907, opened 2018-03-13
Support Shapefile in browser natively


stevage:
It should be possible, using Mike Bostock's [Shapefile](https://github.com/mbostock/shapefile) library and an unzip library like [unzip](https://www.npmjs.com/package/unzip), to open a zip file containing a Shapefile directly in the browser - without using Terria-Server. That would make it much faster, remove the need for the server, and remove the crippling 1MB limit.

stevage:
Calvin Metcalf's [Shapefile library](https://github.com/calvinmetcalf/shapefile-js) supports projections and zip files.

nf-s:
Closed by https://github.com/TerriaJS/terriajs/pull/5080

PR #5080, opened 2020-12-10
Shapefile support in browser

### LINKED ISSUE #4721: Mobx: drag and drop of shapefiles
2 problems:
1. Currently drag & dropping shapefiles into a v8 map won't even attempt to convert it - the UI suggests uploading, but clicking "upload" does nothing.
2. drag and dropping a zip file does not get rid of the overlay as if it's stuck in a drag over event and the "convert/upload" path hasn't been considered/realised for v8

Related:
https://github.com/TerriaJS/terriajs/issues/4633

https://github.com/TerriaJS/terriajs/issues/3082

https://github.com/TerriaJS/terriajs/issues/2907

zoran995:
### What this PR does

Fixes #4721, #2907

Continuation of the work done in #4869.
- added a `GeoJsonMixin` and moved most of `GeoJsonCatalogItem` code to it
- added a `ShapefileCatalogItem` which handles the loading of shapefiles
- extended `GeojsonCatalogItem` to support loading of the dropped zip file, loading is currently done using the `ShapefileCatalogItem`, I didn't find a way to register multiple loaders for the file with the same extension and how to determine which one to use.
- added tests for `GeojsonCatalogItem` and `ShapefileCatalogItem`

### Checklist


zoran995:
any idea why tests are failing, all tests work perfectly fine locally 
image
image

lib/Models/UrlReference.ts
```diff
@@ -63,11 +63,14 @@ export default class UrlReference extends UrlMixin(
       return Promise.resolve(undefined);
     }
 
+    // Does the mapping at this index match this url?
+    // Can we load it if we need to?
```
tephenavies, line 67: Thanks for these comments

lib/Core/Json.ts
```diff
@@ -16,6 +16,15 @@ export function isJsonObject(
...
 
+export function isJsonArray(value: JsonValue | undefined): value is JsonArray {
+  return (
+    value !== undefined &&
+    value !== null &&
+    Array.isArray(value) &&
+    value.every(child => isJsonObject(child) || isJsonArray(child))
+  );
+}
+
```
tephenavies: ~This will return false for `isJsonArray(['a'])` and `['a']` is otherwise a valid `JsonArray`. This function needs to be renamed to avoid this confusion.~
Moved to ShapefileCatalogItem and renamed to `isJsonArrayOrDeepArrayOfObjects`. I don't object to this function being in `Json.ts` if it's more widely applicable, but using this function instead of `Array.isArray` could cause errors/confusion and could be much slower given it's recursive nature.

tephenavies (approved):

tephenavies:
CI should be up at http://ci.terria.io/zoran-shapefile/ soon.

ayeb-nermin:
Hello, how can i add the shp format in the import formats of a layer?

PR #5085, opened 2020-12-16
Auto-convert MapServer layer names to IDs before requesting

### LINKED ISSUE #5084: ArcGIS layer names are not converted to layer id's when requesting tiles in v8
ArcGIS does not do the right thing when requesting layers by name and need to be requested using their id's.

In v7 when using an ArcGIS layer like the Contours layer in NatMap, the layer names used are given as text, but when requesting the actual tiles they have been converted into the layers numerical id's:

v7

v7

v7

In v8:

v8

v8

v8

As can be observed the output is very different. If you manually change the layer name to be the numerical id's in v8, then the layer works successfully.

na9da:
### What this PR does

Fixes #5084 

Testing: As described in the issue, the following link should show a contour map instead of dots.
Test link: http://ci.terria.io/mapserver-layer-names-to-ids/#share=s-ujk3rpqAJMsy27hwYYLRFhT2jws

### Checklist

### DIFF
--- a/lib/Models/ArcGisMapServerCatalogItem.ts
+++ b/lib/Models/ArcGisMapServerCatalogItem.ts
@@ -359,6 +359,13 @@ export default class ArcGisMapServerCatalogItem
     });
   }
 
+  @computed
+  get mapServerStratum(): MapServerStratum | undefined {
+    return this.strata.get(MapServerStratum.stratumName) as
+      | MapServerStratum
+      | undefined;
+  }
+
   loadMapItems() {
     return this.loadMetadata();
   }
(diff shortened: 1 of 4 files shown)


lib/Models/ArcGisMapServerCatalogItem.ts
```diff
@@ -371,10 +378,7 @@ export default class ArcGisMapServerCatalogItem
   }
 
   @computed get imageryProvider() {
-    const stratum = <MapServerStratum>(
-      this.strata.get(MapServerStratum.stratumName)
-    );
-
+    const stratum = this.mapServerStratum;
```
na9da, line 381: Just a minor refactor.

PR #5101, opened 2020-12-18
Don't set recordId since magdaRecord is available

tephenavies:
The existing code misrepresents the recordId as "/". While the Terria id used for the root group is "/" the recordId used to talk to magda about the root group is still (usually) map-config[-something].


tephenavies:
Test failures show that this is probably the wrong fix. Perhaps instead `recordId` should be set, except using `config.id`, as from before this commit: https://github.com/TerriaJS/terriajs/blob/12eb658b3a64ecd17dc4527d26183c82873e71d9/lib/Models/Terria.ts. I don't know why that commit changed the `recordId`.

PR #5113, opened 2020-12-25
Fix pluralisation of mapDataState to support other languages

zoran995:
### What this PR does

The pluralisation of `mapDataState` string is not done correctly which causes issues with languages more complex than English. I18next already contains a built-in algorithm that handles this really nice and is described [here](https://www.i18next.com/translation-function/plurals). The documentation also lists a nice [jsfiddle](https://jsfiddle.net/sm9wgLze) where it is possible to check the pluralisation option for language and provide the translation. 

### Checklist


wibily (approved):

wibily:
thanks!

PR #5123, opened 2021-01-08
Fixed pickedFeatures for dynamic layers

nf-s:
### Fixed pickedFeatures for dynamic layers

Fixes https://github.com/TerriaJS/terriajs/issues/4768

Layers which are dynamic (i.e. `mapItems` change due to things like time or other dimensions) - would lose `pickedFeatures` when `mapItems` change. 

#### Replicate problem

For example, auto updating datasets will hide `FeatureInfoPanel` when the dataset is updated:

- http://ci.terria.io/next/#clean&https://gist.githubusercontent.com/nf-s/875ba30b1fb6cad9a3450e08ef8eef33/raw/27812f29500669911c03f93eb60d59c8565cef46/power-gen.json
- Add `Current Power Generation` dataset
- Select point
- Wait for update
- `FeatureInfoPanel` will disappear after update

This was due to `feature._catalogItem` not being set correctly.

#### Fixed

- http://ci.terria.io/picked-feature-catalogitem/#clean&https://gist.githubusercontent.com/nf-s/875ba30b1fb6cad9a3450e08ef8eef33/raw/27812f29500669911c03f93eb60d59c8565cef46/power-gen.json
- Add `Current Power Generation` dataset
- Select point
- Wait for update

### Future work

While the `FeatureInfoPanel` will stay open when the dataset changes, the actual feature info does **not** update &mdash; it will show the feature info from the feature that was selected before the change.

~This is the same behavior in master. At some point it may be worth updating `pickedFeatures` as the datasets change.~

This is a bug, it has an issue here https://github.com/TerriaJS/terriajs/issues/5127

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### MobX Development
 
 #### next release (8.0.0-alpha.66)
+* Fixed `pickedFeatures` not setting `_catalogItem` property correctly
 * [The next improvement]
 
 #### 8.0.0-alpha.65
--- a/lib/Map/PickedFeatures.ts
+++ b/lib/Map/PickedFeatures.ts
@@ -46,7 +46,7 @@ export default class PickedFeatures {
    * Gets or sets the array of picked features.  The array is observable and may be updated up until the point that
    * {@see PickedFeatures#allFeaturesAvailablePromise} resolves.
    */
-  @observable features: Entity[] = [];
+  @observable features: Feature[] = [];
 
   /**
    * Gets or sets a message describing an error that occurred while picking features.
--- a/lib/Models/Cesium.ts
+++ b/lib/Models/Cesium.ts
@@ -1,3 +1,4 @@
+import i18next from "i18next";
 import { autorun, computed, runInAction } from "mobx";
 import { createTransformer } from "mobx-utils";
 import BoundingSphere from "terriajs-cesium/Source/Core/BoundingSphere";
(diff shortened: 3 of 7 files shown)


zoran995:
> While the `FeatureInfoPanel` will stay open when the dataset changes, the actual feature info does **not** update — it will show the feature info from the feature that was selected before the change.
> 
> This is the same behavior in master. At some point it may be worth updating `pickedFeatures` as the datasets change.

In master, feature info will update after dataset update, just checked it with `Current Power Generation` dataset. 
Screenshot 2021-01-08 144043
Screenshot 2021-01-08 144113

nf-s:
Good pickup, thanks @zoran995, I thought I tested polling csvs, but apparently not! 
The `FeatureInfoPanel` doesn't seem to update for changes in time and other dimensions (eg display variables) for other datasets.

nf-s:
As far as I can tell, it has to do with how point features are created/modified in master for `TableDataSource` - https://github.com/TerriaJS/terriajs/blob/9125dbf2d3481bd9d02edc273dbdca713d4759c2/lib/Models/TableDataSource.js#L369

In `next`, we just create new features every update - we never just update feature properties https://github.com/TerriaJS/terriajs/blob/48ba953cf90abdce99f89741e65e5504f0a753be/lib/ModelMixins/TableMixin.ts#L534

nf-s:
I'm going to leave that issue for another day - `TableMixin optimisation: update existing features instead of re-creating them on dataset change`  https://github.com/TerriaJS/terriajs/issues/5127

This PR is ready to go!

nf-s:
I have decided this is a bad approach. 

Ideally we should "re-select" features if they update

PR #5135, opened 2021-01-13
TableMixin styles (display variables) use column title

nf-s:
### TableMixin styles (display variables) use column title

If `style.title` is undefined, use `column.title` for display variable dropdown.

### Checklist


lib/ModelMixins/TableMixin.ts
```diff
@@ -321,7 +321,10 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
         options: this.tableStyles.map(style => {
           return {
             id: style.id,
-            name: style.styleTraits.title || style.id
+            name:
```
KeyboardSounds, line 324: Should this use `??` instead?

KeyboardSounds (approved):
Looks good, aside from one tiny nitpick which you can ignore if you disagree.

lib/ModelMixins/TableMixin.ts
nf-s: 👍

PR #5144, opened 2021-01-19
Create merge TraitsOption for objectArrayTrait

nf-s:
### Create `merge` `TraitsOption` for `objectArrayTrait`

Fixes https://github.com/TerriaJS/terriajs/issues/5048

#### DEA coastlines WMS uses custom legend url in catalog:

```json
{
  "legends": [
    {
      "url": "https://data.dea.ga.gov.au/projects/coastlines/DEACoastLines_legend 1.png"
    }
  ]
}
```

#### WMS GetCapabilitiesStratum will then add default legends on top:

For GeoServer WMS we request high DPI legends and use imageScaling=0.5 to scale it back to original size

```json
{
  "legends": [
    {
      "url": "https://geoserver.dea.ga.gov.au/geoserver/wms?service=WMS&version=1.3.0&request=GetLegendGraphic&format=image%2Fpng&layer=dea%3ADEACoastLines&LEGEND_OPTIONS=fontName%3ACourier%3BfontStyle%3Abold%3BfontSize%3A12%3BforceLabels%3Aon%3BfontAntiAliasing%3Atrue%3BlabelMargin%3A5%3BfontColor%3A0xffffff%3Bdpi%3A182&transparent=true",
      "imageScaling": 0.5,
      "urlMimeType": "image/png"
    }
  ]
}
```

#### This results with:
```json
{
  "legends": [
    {
      "url": "https://data.dea.ga.gov.au/projects/coastlines/DEACoastLines_legend 1.png",
      "imageScaling": 0.5,
      "urlMimeType": "image/png"
    }
  ]
}
```

Which is wrong, as this legend should have no `imageScaling` or `urlMimeType` traits.

### Before vs After

image

### After 

We have set `merge: false` in `LegendTraits`

```ts
@objectArrayTrait({
    name: "Legend URLs",
    description: "The legends to display on the workbench.",
    type: LegendTraits,
    idProperty: "index",
    merge: false
  })
  legends?: LegendTraits[];
```

Now we don't merge trait values across strata for a given `idProperty` (`index` in this case) - so the manually defined legend will **completely** overwrite the WMS `GetCapabilitiesStratum` legend.

#### So we get the correct result

```json
{
  "legends": [
    {
      "url": "https://data.dea.ga.gov.au/projects/coastlines/DEACoastLines_legend 1.png"
    }
  ]
}
```

### Checklist


lib/Traits/CatalogMemberTraits.ts
```diff
@@ -155,7 +155,8 @@ export default class CatalogMemberTraits extends ModelTraits {
     name: "Legend URLs",
     description: "The legends to display on the workbench.",
     type: LegendTraits,
-    idProperty: "index"
+    idProperty: "index",
+    merge: false
```
nf-s, line 175: `LegendTraits` is the only `ObjectArrayTrait` where I have set `merge: false` - I'm sure there will be others

tephenavies (changes requested):
Please add a note a the top of alpha.66 in CHANGES.md that there is a breaking change to `legends` in `CatalogMemberTraits`. Something like:

Potentially breaking changes:
* Changed merging behaviour of Trait `legends` (of type `LegendTraits`) in `CatalogMemberTraits`. This affects legends on all CatalogMember models. Legend objects in higher strata now replace values in lower strata that match by index, rather than merging properties with them.

nf-s:
> 
> 
> Please add a note a the top of alpha.66 in CHANGES.md that there is a breaking change to `legends` in `CatalogMemberTraits`. Something like:
> 
> Potentially breaking changes:
> 
>     * Changed merging behaviour of Trait `legends` (of type `LegendTraits`) in `CatalogMemberTraits`. This affects legends on all CatalogMember models. Legend objects in higher strata now replace values in lower strata that match by index, rather than merging properties with them.

@steve9164 done!

tephenavies (approved):

PR #5145, opened 2021-01-19
Configurable search providers from config

### LINKED ISSUE #5141: Search bar configurable from config
I am considering an option to make the search bar configurable from the config files. Currently, the users need to edit the TerriaMap to configure another SearchProviders. This is similar to BaseMap configuration options. 
I am thinking of dropping this into `configParameters`, and maybe an init option to disable some of them.

zoran995:
### What this PR does

Fixes #5141
Requires https://github.com/TerriaJS/TerriaMap/pull/515

After investigation as the best option appears to utilize the existing model system for SearchProviders.
- wrote proposal as architecture decision record
- added SearchProviderMixin to connect searchProviders with a model system
- ported existing search providers to the proposed model
- added defaults to BingMapsSearchProviders and AustralianGazetteer for testing purposes.

### Testing
Default SearchProviders are BingMaps and AustralianGazetteer, need to apply registerSearchProviders and comment out reference to BingMapsSearchProviderViewModel.

### Pending
-   [x] Make everything support i18next code and load correct names in GUI
-   [x] Tests
-   [x] Documentation
-   [x] Create TerriaMap PR

### Future work
- Introduce server-side WFS SearchProvider which will perform a search using terriajs-server and return the results.

### Checklist


zoran995:
Docs are still missing the `timeline` and `models` property of `initData`, I wasn't sure if `timeline` is used.
For the `models` I didn't know where it is used, can anyone else document it?

lib/Models/Terria.ts
```diff
@@ -725,7 +822,16 @@ export default class Terria {
...
       if (this.configParameters.hasOwnProperty(key)) {
-        this.configParameters[key] = parameters[key];
+        if (key === "searchBar") {
+          // merge default and new
+          //@ts-ignore
+          this.configParameters[key] = {
+            ...this.configParameters[key],
+            ...parameters[key]
+          };
+        } else {
```
zoran995: merging searchBar params so specified defaults don't get overridden with undefined. We might even run deep merge on config parameters something as `_.merge(this.configParameters, parameters)`

zoran995:
This is ready for review, if anyone has time to check.
p.s. Sorry, it turned out to be a monster PR.

zoran995:
This is ready for review, if anyone has time to check.
p.s. Sorry, it turned out to be a monster PR.

zoran995:
Hi @steve9164, just to check is this still interesting for Terria team to rebase the changes when I am free?

zoran995:
Hi @nf-s I have synced the code with the latest main and it is ready for review. 
p.s. Would you like me to completely rebase and squash the code in a single commit to avoid polluting history with 2+ years old commits?

nf-s:
Hi @zoran995 
Thanks so much!
A rebase would be great. We don't think squash is required - but whatever you prefer!

We will try to review it in the next few weeks.

Thanks again 🙂

lib/ReactViewModels/SearchState.ts
```diff
@@ -46,28 +42,36 @@ export default class SearchState {
...
+      this.terria.searchBarModel.catalogSearchProvider =
+        options.catalogSearchProvider ||
+        new CatalogSearchProvider("catalog-search-provider", options.terria);
+    });
+
+    const self = this;
 
     this._catalogSearchDisposer = reaction(
-      () => this.catalogSearchText,
+      () => self.catalogSearchText,
```
ljowen, line 61: I don't believe this is needed since arrow functions use the enclosing `this` value

lib/Traits/SearchProviders/CesiumIonSearchProviderTraits.ts
```diff
@@ -0,0 +1,20 @@
+import primitiveTrait from "../Decorators/primitiveTrait";
+import mixTraits from "../mixTraits";
+import LocationSearchProviderTraits, {
+  SearchProviderMapCenterTraits
+} from "./LocationSearchProviderTraits";
+
+export default class BingMapsSearchProviderTraits extends mixTraits(
```
ljowen:
```suggestion
export default class CesiumIonSearchProviderTraits extends mixTraits(
```

doc/customizing/search-providers.md
```diff
@@ -0,0 +1,132 @@
...
+
+-   if items are renamed, or moved
+-   dynamic groups are updated (for example, WMS server publishes new layers)
+
+For more details see [/buildprocess/generateCatalogIndex.ts](/buildprocess/generateCatalogIndex.ts)
+
+## Location search providers
+
+Location search providers are used to search for locations on the map. TerriaJS currently supports two implementations of search providers:
+
```
ljowen, line 54: We should add CesiumIonSearchProvider since that's now the default in TerriaMap

lib/Models/SearchProviders/CesiumIonSearchProvider.ts
```diff
@@ -28,74 +32,77 @@ interface CesiumIonGeocodeResult {
...
-    );
+    runInAction(() => {
+      if (!this.key && this.terria.configParameters.cesiumIonAccessToken) {
+        this.setTrait(
+          CommonStrata.defaults,
+          "key",
+          this.terria.configParameters.cesiumIonAccessToken
+        );
+      }
+      this.showWarning();
```
ljowen: I'm seeing this warning logged if I provide the cesium access key in the `searchProviders: [{ ... }]` object but not as `cesiumIonAccessToken`. However the location search still works correctly, same thing for `BingMapsSearchProvider`

test/Models/SearchProviders/BingMapsSearchProviderSpec.ts
ljowen, line 1: Thanks for adding this!

ljowen (commented):
Looks great! Tested locally with Bing and Cesium search providers. Couple of small comments

lib/Traits/SearchProviders/CesiumIonSearchProviderTraits.ts
zoran995: fixed

doc/customizing/search-providers.md
zoran995: added a section about CesiumIonSearchProvider

lib/Models/SearchProviders/CesiumIonSearchProvider.ts
zoran995: Good catch, stratum values won't be defined till `updateModel` is called and it was happening later in initialization. Fixed by moving the warning log to `SearchBarModel` and having default noop `showWarning` method on `LocationSearchProviderMixin`

ljowen (approved):

PR #5156, opened 2021-01-20
Line and point custom chart component

### LINKED ISSUE #72: DE Australia - Waterbodies - dots on chart line
Ability to put dots on the line chart to show where the valid observations are. 

This is a symbology change, so quite different to the Water discharge level example. 

Instead of a line chart the irrigators would like a line with markers chart. Here is a very basic example using excel. Data is on the left. 

Please note that there are fewer points in the DEA waterbodies time series than there are in the imagery stack (surface area isn't calculated for cloud affected / partial coverage of waterbody), so the plot symbology needs to be drawn from the data in the CSV's not from the imagery service stack at that location. 

image

KeyboardSounds:
### What this PR does

Fixes https://github.com/TerriaJS/de-australia-map/issues/72
Fixes https://github.com/TerriaJS/terriajs/issues/3981

- Adds a new Line and Point chart type
- CustomChartComponent now has a chart-type attribute

### Checklist

-   [x] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ Changes covered by existing tests


lib/ModelMixins/TableMixin.ts
```diff
@@ -270,7 +270,7 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
             name: yColumn.title,
             categoryName: this.name,
             key: `key${this.uniqueId}-${this.name}-${yColumn.name}`,
-            type: "line",
+            type: this.chartType,
             xAxis,
```
na9da, line 274: Currently [chartType](https://github.com/TerriaJS/terriajs/blob/line-and-point-custom-chart-component/lib/Traits/DiscretelyTimeVaryingTraits.ts#L37) defaults to `momentLines`. So this change forces the chart type of all existing table items that don't specify one to `momentLines`.

To avoid that, I think we should:
1) Remove the default value for [chartType](https://github.com/TerriaJS/terriajs/blob/line-and-point-custom-chart-component/lib/Traits/DiscretelyTimeVaryingTraits.ts#L37)
2) Change this line to `type: this.chartType || "lineAndPoint"`

I think 1) is alright because the [existing code that uses chartType](https://github.com/TerriaJS/terriajs/blob/line-and-point-custom-chart-component/lib/ModelMixins/DiscretelyTimeVaryingMixin.ts#L335) redundantly sets a default.

na9da:
Looks good to me @KeyboardSounds. Just one suggestion from me.

lib/ModelMixins/TableMixin.ts
KeyboardSounds: ah, well caught! that would have been bad.

`chartType` in itself probably needs reconsidering, because a catalog item can theoretically have multiple chart types and why is it in DiscretelyTimeVaryingTraits, etc. But I'll make a ticket for it :)

na9da (approved):

PR #5146, opened 2021-01-19
Translation and configuration documentation

zoran995:
### What this PR does

Fixes #<insert issue number here if relevant>

This PR documents the ways translation language files should be used. It also provides an updated list of configuration parameters in the config documentation. I didn't plan to update the entire config documentation but it happened while adding the `languageConfiguration` parameter.

There are some parameters I didn't know how to describe so they are missing a description. 
Parameters missing in configParameters list when compared with v7 are, I am not sure which ones are deprecated and which one will be implemented again:
***
`autoPlay`
`googleAnalyticsOptions`
`printDisclaimer`
`mobileDefaultViewerMode`
`openAddData`
`showFeaturePrompts`
`cesiumTerrainUrl`
***

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [ ] I've updated CHANGES.md with what I changed.

PR #5158, opened 2021-01-20
Translation documentation

zoran995:
### What this PR does

This is an extract of only documentation related to translation from #5146, updating and validation of config documentation could take some time, and this might be accepted much faster.

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - **documentation**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### MobX Development
 
 #### next release (8.0.0-alpha.66)
+* Wrote translation documentation [#5158](https://github.com/TerriaJS/terriajs/pull/5158)
 * [The next improvement]
 
 #### 8.0.0-alpha.65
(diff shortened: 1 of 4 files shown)


zoran995:
I am thinking that it might be more intuitive for users if we changed the name of `languageOverrides.json` to `additional.json` representing additional translations, and much easier to understand the structure, which probably will be used in much more cases to provide new translation key (for use in configuration files) then to override the translation. 
Also, maybe change the name of `translation.json` to `common.json`, representing the common language file for that language, although it doesn't make much difference. 
@steve9164, what do you think? If doing the change, it will probably be easier to do it before weblate is fully configured?

zoran995:
already merged in #5146

PR #5140, opened 2021-01-14
Configurable basemaps from init json

zoran995:
### What this PR does

Fixes #907

- Enables defining the list of baseMaps in init files. This was requested a long time ago (#907) but never implemented. When using bingMaps, we check if there is a specified bingMapsKey and, if not, set the default one from `configParameters`.
- Creates a default baseMap list to reflect the existing state
- Implements `previewBaseMapId` init file parameter so dataset preview baseMap can be changed (there is no more a fixed list of available baseMaps so we can't be sure that positron baseMap will exist)

### Testing

Add this to the init file and check that there are only 2 base maps.

```
"baseMaps": [
    {
      "item": {
        "id": "basemap-positron",
        "name": "Positron (Light)",
        "type": "open-street-map",
        "url": "https://basemaps.cartocdn.com/light_all/",
        "attribution": "© <a href='https://www.openstreetmap.org/copyright'>OpenStreetMap</a>, © <a href='https://carto.com/about-carto/'>CARTO</a>",
        "subdomains": ["a", "b", "c", "d"],
        "opacity": 1.0
      },
      "image": "/images/positron.png"        
    },
    {
      "item": {
        "id": "basemap-darkmatter",
        "name": "Dark Matter",
        "type": "open-street-map",
        "url": "https://basemaps.cartocdn.com/dark_all/",
        "attribution": "© <a href='https://www.openstreetmap.org/copyright'>OpenStreetMap</a>, © <a href='https://carto.com/about-carto/'>CARTO</a>",
        "subdomains": ["a", "b", "c", "d"],
        "opacity": 1.0
      },
      "image": "/images/dark-matter.png" 
    }
  ]
}
```

### Pending 

- [x] Check if the base map with a specific id is already added to avoid duplicates
- [x] Tests
- [x] Documentation - *edit all init properties should be documented now*
- [x] Create TerriaMap PR to remove references to deleted functions

### Future work

- Fix share URL so that base map set in it works correctly. @nf-s can we update the parameter for share to `baseMapId` instead of `baseMapName` - _edit I couldn't get share basemap to work on ci next_
- URL parameter `baseMapId` - so user can explicitly set the basemap
- Add attribution traits for all `catalogItems` - _done in #5167_

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [x] I've updated CHANGES.md with what I changed.

PR #5167, opened 2021-01-21
Mappable attribution

### LINKED ISSUE #5162: Port `attribution` property to Mappable in v8
> ... `attribution` property to catalog items.  The attribution is displayed on the map when the catalog item is enabled.

zoran995:
### What this PR does

Fixes #5162

This might also be considered as needed by #5140, to credit basemap items properly.

- Added a `AttributionTraits` and included into `MappableTraits`.
- Implemented it for most of the catalog items. I think that only `GtfsCatalogItem` is missing implementation for attribution since I couldn't figure out where to hook the attribution. 

### Checklist


tephenavies (approved):
Looks great. Thanks!

PR #5174, opened 2021-01-25
Add fallback for Color.fromCssColorString

nf-s:
### Add fallback for Color.fromCssColorString

The type signature for `Color.fromCssColorString(col:string) => Color` is incorrect as it can return undefined.

For example, this catalog has a CSV with an enum color map with an invalid CSS color:
- http://ci.terria.io/next/#clean&https://gist.githubusercontent.com/nf-s/e89bc5da84fb5f41d205d02287841abe/raw/2dfad06e6421b7942fd3f3da4b1d62a1cbf50f2b/colorstringtest.json

When the catalog item is added to the workbench the map will crash.

This PR adds default colors for wherever Color.fromCssColorString is used.

### I also deleted some old JS files:

- lib/Models/LegendHelper.js
- lib/Models/TableColumnStyle.js
- lib/Models/TableDataSource.js
- test/Models/LegendHelperSpec.js
- test/Models/TableDataSourceSpec.js 

### Checklist


lib/Models/GlobeOrMap.ts
```diff
@@ -170,6 +170,7 @@ export default abstract class GlobeOrMap {
 
       if (isDefined(feature._cesium3DTileFeature)) {
         const originalColor = feature._cesium3DTileFeature.color;
+        const defaultColor = Color.fromCssColorString("#fffffe");
```
KeyboardSounds, line 173: Why is this nearly white instead of white?

KeyboardSounds (approved):
Can you fix the type declaration in terriajs-cesium as well? We're using custom type decs for Cesium still I think so it's a problem on our end. Unless we're not, in which case we should create an issue in Cesium, because sharing is caring.

PR #5177, opened 2021-01-27
Improve item search zoom.

na9da:
### What this PR does

This improves zoom behavior for item search (for 3d tiles).

Previously when zooming to a small building/feature, it sometimes gets hidden behind a bigger building. With this change, when zooming to a small building we show a top angle view.

### Checklist


lib/Models/ItemSearchProvider.ts
```diff
@@ -1,6 +1,7 @@
 import BoundingSphere from "terriajs-cesium/Source/Core/BoundingSphere";
 import Rectangle from "terriajs-cesium/Source/Core/Rectangle";
 import DataSource from "terriajs-cesium/Source/DataSources/DataSource";
+import LatLonHeight from "../Core/LatLonHeight";
```
KeyboardSounds: Is this import being used?

KeyboardSounds (changes requested):
Other than the possible stray import, everything looks good to me :)

lib/Models/ItemSearchProvider.ts
na9da: Good catch.

KeyboardSounds (approved):

PR #5180, opened 2021-01-27
Configure item search parameters.

na9da:
### What this PR does

This change allows configuring the item search parameters through the catalog file.

Currently we support setting a human readable `name` for the parameter and also passing index specific `queryOptions`.

eg:
```
{
      "name": "Buildings 3D",
      "type": "3d-tiles",
      "url": "https://nsw.digitaltwin.terria.io/api/v0/data/485be627-677f-47bb-be13-decb3d58168b/Buildings3D/tileset.json",
      "search": {
        "providerType": "indexed",
        "options": {
          "indexRootUrl": "/Buildings3D/indexRoot.json"
        },
        "resultTemplate": "Building #{{OBJECTID}}",
        "parameters": [
          {
            "id": "GEOSCAPE_ROOF_SLOPE",
            "name": "Roof Slope",
            "queryOptions": {
              "prefix": true,
              "fuzzy": 2
            }
          }
        ]
      }
    }
```

### Checklist

### DIFF
--- a/lib/Models/ItemSearchProviders/IndexedItemSearchProvider.ts
+++ b/lib/Models/ItemSearchProviders/IndexedItemSearchProvider.ts
@@ -249,20 +249,29 @@ export default class IndexedItemSearchProvider extends ItemSearchProvider {
         `No valid zoom point defined for data record at index ${dataIdx}`
       );
     }
-    const center = Cartesian3.fromDegrees(_longitude, _latitude);
-    const boundingSphere = new BoundingSphere(
-      center,
-      // Prefer the user specified feature height over the tile radius which is
-      // automatically calculated by the indexer.
-      // We also don't want to zoom close to small objects, so threshold the zoom radius to 50.
-      isNaN(_featureHeight) ? _tileRadius : Math.max(_featureHeight * 2, 50)
-    );
+
+    let zoomToTarget;
+    if (_featureHeight) {
+      zoomToTarget = {
+        latitude: _latitude,
+        longitude: _longitude,
+        featureHeight: _featureHeight
+      };
+    } else {
+      const center = Cartesian3.fromDegrees(_longitude, _latitude);
+      const boundingSphere = new BoundingSphere(
+        center,
+        // Prefer the user specified feature height over the tile radius which is
+        // automatically calculated by the indexer.
+        // We also don't want to zoom too close to small objects, so threshold the zoom radius.
+        _tileRadius
+      );
+      zoomToTarget = boundingSphere;
+    }
     return {
       id,
       idPropertyName: this.indexRoot.idProperty,
-      zoomToTarget: {
-        boundingSphere
-      },
+      zoomToTarget,
       properties
     };
   }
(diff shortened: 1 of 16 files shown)


lib/Models/ItemSearchProviders/IndexedItemSearchProvider.ts
```diff
@@ -59,23 +61,27 @@ export default class IndexedItemSearchProvider extends ItemSearchProvider {
...
-          name: propertyId,
+          name: parameterOptions?.name || propertyId,
           range: index.range
         };
       case IndexType.enum:
         return {
           type: "enum",
           id: propertyId,
-          name: propertyId,
+          name: parameterOptions?.name || propertyId,
```
KeyboardSounds: Here too

lib/Models/ItemSearchProviders/TextIndex.ts
```diff
@@ -49,14 +49,18 @@ export default class TextIndex implements IndexBase<TextSearchQuery> {
...
    *
-   * @param value The value to be searched.
+   * @param value         The value to be searched.
+   * @param queryOptions  MiniSearch.SearchOptions
    * @param return The IDs of objects matching the search text.
    */
-  async search(value: TextSearchQuery): Promise<Set<number>> {
+  async search(
+    value: TextSearchQuery,
+    queryOptions: any
```
KeyboardSounds: For better typing, use the more specific MiniSearch.SearchOptions type here and cast the caller's input param to `MiniSearch.SearchOptions`

lib/Models/ItemSearchProviders/Types.ts
```diff
@@ -1,7 +1,7 @@
 export type IndexBase<QueryType> = {
   type: IndexType;
   load(indexRootUrl: string, valueHint: QueryType): Promise<void>;
-  search(query: QueryType): Promise<Set<number>>;
+  search(query: QueryType, queryOptions: any): Promise<Set<number>>;
```
KeyboardSounds: Rather than passing a queryOptions to every index type, including the ones that don't use it, I would make it optional and remove it as a `search` parameter in the things that aren't using it. It's not a huge deal, but the code is slightly clearer that way. What do you think?

KeyboardSounds (changes requested):

lib/Models/ItemSearchProviders/IndexedItemSearchProvider.ts
```diff
@@ -59,23 +61,27 @@ export default class IndexedItemSearchProvider extends ItemSearchProvider {
...
+    const parameterOptions = this.parameterOptions.find(
+      ({ id }) => id === propertyId
+    );
     switch (index.type) {
       case IndexType.numeric:
         return {
           type: "numeric",
           id: propertyId,
-          name: propertyId,
+          name: parameterOptions?.name || propertyId,
```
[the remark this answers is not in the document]
na9da: Good catch and `??` is neat!

lib/Models/ItemSearchProviders/Types.ts
na9da: Good point!

KeyboardSounds (approved):

PR #5186, opened 2021-01-28
Pedestrian mode

### LINKED ISSUE #5019: [EPIC] Implement pedestrian view - tool
### Problem Statement
_What problem or need are we trying to solve or fulfil?_

**Speculative / to be validated:**

How might we help **urban planners**(?) remotely assess and communicate the spatial impact of buildings/infrastructure on pedestrians?

### Impact
_What's the impact of this problem on our customers and to our business?_

**Speculative / to be validated:**

- Negative spatial properties of buildings/infrastructure (past/present/future) are less likely to be picked up early in the planning process, resulting in lost time and money
- Stakeholders have less visibility over the planning process, resulting a less diverse feedback
- Planners are required to use expensive and/or complex 3D/GIS software to simulate buildings/infrastructure from the perspective of a pedestrian

### How do we judge success?
_What are the goals of the project and the success criteria by which they will be measured? These need to be specific and measurable._

**Refine with client**
1. Number of visitors independently activate and engage with this tool (Google Analytics)
2. Number of planners who report using this tool to assess and/or communicate buildings/infrastructure proposals
3. Positive feedback from users (via interviews, email, map feedback tool etc.)

### What do we already know?
_What data or insights do we have to validate the problem?_

**Speculative examples to prompt data collection:**

**Anecdotes from client or target user group**
- Expensive/time consuming pedestrian POV analysis?
- Difficulty communicating pedestrian POV to others e.g. stakeholders unable to review proposals remotely?
- Negative spatial impacts of new buildings/infrastructure missed in planning stage? Cost of changes?

**Competitor analysis**
- Who else is already supporting this use case or offering similar functionality?
- What are the gaps in their offering?
- … ?

**Visitor analytics**
- number of visitors who have added reality mesh data or 3D model to the map in the past month?
- ... ?

### What do we need to answer?
_What assumptions are we making that need to be validated/refuted?_

- Users can access either Chrome, Firefox or Edge browser…
- Users have access to a keyboard and mouse and are able to use them simultaneously
- … ?

---

User comment: 
"put the two points (location and direction) on the map to create a view from our viewpoint. To expand this capability to create a walking view mode. Basically a user can draw a line and simulate a walking view along the path. 
Inspiration: https://kartta.hel.fi/3d/#/
image

na9da:
### What this PR does

Fixes #5019 

This PR implements the pedestrian mode for easy navigation at street level.

### Checklist


na9da:
@philipgrimmett  @KeyboardSounds  I have attempted to implement most things in the design spec.

I had to leave out a few things as it might require more experimentation:
1) Hiding the zoom buttons - not working when I attempted to hide it using a `Terria.configParameter` flag.
2) MiniMap direction
3) Smoother MiniMap. Right now the minimap flickers when moving fast due to slow updates. Might have to try out some plugin for leaflet to buffer tiles like this one - https://github.com/TolonUK/Leaflet.EdgeBuffer

KeyboardSounds:
It's up to you which of these you want to fix right now, butt there are a few weird things I found:
- If you change the basemap for the map, the basemap on  the minimap doesn't change
- the bevel on the minimap looks odd. it doesn't match any of the other UI components in Terria.
- this is purely a matter of opinion but strafing from side to side with the A and D keys should move the camera more slowly than moving forward. My guess is that users are more likely to want to make small adjustments in those directions rather than "walking"— people naturally want to walk in the direction of the camera
- There are a few cases where the position on the minimap doesn't seem to match the world position. this is most obvious when you're looking at a basemap with roads on it. [Here's an example](http://localhost:3001/#start=%7B%22version%22%3A%228.0.0%22%2C%22initSources%22%3A%5B%7B%22stratum%22%3A%22user%22%2C%22models%22%3A%7B%7D%2C%22workbench%22%3A%5B%5D%2C%22timeline%22%3A%5B%5D%2C%22initialCamera%22%3A%7B%22west%22%3A115.9070262661226%2C%22south%22%3A-31.970212523524022%2C%22east%22%3A115.90714040624808%2C%22north%22%3A-31.970143318024046%2C%22position%22%3A%7B%22x%22%3A-2366235.0734716365%2C%22y%22%3A4871538.567856443%2C%22z%22%3A-3357635.102075125%7D%2C%22direction%22%3A%7B%22x%22%3A-0.28197193036837687%2C%22y%22%3A0.302694671450872%2C%22z%22%3A0.9104217519147815%7D%2C%22up%22%3A%7B%22x%22%3A-0.41438362104194343%2C%22y%22%3A0.8174299403135719%2C%22z%22%3A-0.40011811667446084%7D%7D%2C%22homeCamera%22%3A%7B%22west%22%3A109%2C%22south%22%3A-45%2C%22east%22%3A158%2C%22north%22%3A-8%7D%2C%22baseMapName%22%3A%22Positron+%28Light%29%22%2C%22viewerMode%22%3A%223dSmooth%22%2C%22currentTime%22%3A%7B%22dayNumber%22%3A2459243%2C%22secondsOfDay%22%3A47203.361%7D%2C%22showSplitter%22%3Afalse%2C%22splitPosition%22%3A0.5%2C%22stories%22%3A%5B%5D%7D%5D%7D) My guess is that this is because of perspective/projection weirdnesses and might be hard to fix, but users could find it confusing.
- While using the mouse to rotate the camera, I accidentally rotated so far that the horizon ended up upside down. We should prevent users rotating past their "feet" to avoid disorientation.
- If you're already in pedestrian mode in 3D, and then you switch to 2D, the Minimap and control panel are still visible and there's no way to get rid of them.

lib/ReactViews/Tools/PedestrianMode/Marker.tsx
```diff
@@ -0,0 +1,40 @@
+import Cartesian3 from "terriajs-cesium/Source/Core/Cartesian3";
```
KeyboardSounds: Should this be a .tsx file? Since it's a Model, it probably doesn't belong in ReactViews, either.

KeyboardSounds (changes requested):
Only one tiny thing, but otherwise, this looks solid. Nice work putting all this together so quickly!

lib/ReactViews/Tools/PedestrianMode/Marker.tsx
na9da: Good catch. I have renamed it to `.ts`, however the class is not generally useful so I don't want to take it out of context and place it in `Models/`.

na9da:
Hey @KeyboardSounds all are valid points, I have fixed or improved some of them in the latest changes:

> If you change the basemap for the map, the basemap on the minimap doesn't change

Good catch! Fixed. 

> the bevel on the minimap looks odd. it doesn't match any of the other UI components in Terria.

I removed the minmap shadow for now. However this requires more tweaking to match @philipgrimmett designs

>  this is purely a matter of opinion but strafing from side to side with the A and D keys should move the camera more slowly than moving forward. My guess is that users are more likely to want to make small adjustments in those directions rather than "walking"— people naturally want to walk in the direction of the camera

Good point.  I have halved the strafe rate wrt forward/backward rate. We can tweak this further after feedback from others.

> There are a few cases where the position on the minimap doesn't seem to match the world position. this is most obvious when you're looking at a basemap with roads on it. Here's an example My guess is that this is because of perspective/projection weirdnesses and might be hard to fix, but users could find it confusing.

The point on the minmap is the exact position of the camera. This is actually hidden in the main map because it only shows things in front of the camera. This could be something we can improve in the next iteration.

>    While using the mouse to rotate the camera, I accidentally rotated so far that the horizon ended up upside down. We should prevent users rotating past their "feet" to avoid disorientation.

I'll defer this one to the next iteration as well as it requires more experimentation.

>    If you're already in pedestrian mode in 3D, and then you switch to 2D, the Minimap and control panel are still visible and there's no way to get rid of them.

Good catch! I have fixed it now.

KeyboardSounds (approved):

lib/ReactViews/Tools/PedestrianMode/MovementsController.tsx
```diff
@@ -0,0 +1,342 @@
+import { runInAction } from "mobx";
```
zoran995, line 1: I wonder why this is `tsx` and in `ReactViews` it is more for a `Map`? Personally think that this could be used a general feature in TerriaJS.

PR #5187, opened 2021-01-28
Misc AREMI fixes

nf-s:
### Misc AREMI fixes

* Fix `CkanCatalogGroup.groupBy = "none"` members
* Fix `TableMixin` region mapping feature props and make Long/Lat features use column titles (if it exists) to match v7 behaviour.
  * This includes an awful thing in `FeatureInfoSection.jsx`. In v7 - we used to add this to each `Feature` - `properties._terria_columnAliases = tableStructure.getColumnAliases()` - which I think is worse...
* Add support for `CkanItemReference` `wms_layer` property
* Add support for `ArcGisMapServerCatalogGroup` to use `sublayerIds`.
* Remove `wordBreak="break-all"` from `Box` surrounding `DataPreview`
  * This is because were getting weird breaks in `Description` and other elements that didn't reverse this

### Checklist

-   [x] Small bug fixes - so no tests.


lib/ReactViews/FeatureInfo/FeatureInfoSection.jsx
```diff
@@ -86,6 +87,18 @@ export const FeatureInfoSection = observer(
...
+        let aliases = [];
+        if (TableMixin.isMixedInto(this.props.catalogItem)) {
+          aliases = this.props.catalogItem.columns
+            .filter(col => col.name && col.title && col.name !== col.title)
+            .map(col => [col.name, col.title]);
+        }
+        aliases.forEach(aliasMap => {
+          propertyData[aliasMap[0]] = propertyData[aliasMap[1]];
+        });
+
```
tephenavies: Should `aliasMap[1]` keys be removed from `propertyData`? And is it possible or likely that an `aliasMap[0]` could also be an `aliasMap[1]` later in `aliases` and hence the key will be overwritten in `propertyData`?

nf-s: I think we need both in `propertyData` - as some feature templates use column name and others use column title (even if both are defined).
Overwriting aliases is definitely possible, but it seems unlikely someone would name a `column.title` after a different `column.name`. I have added this so at least we aren't overriding properties which are defined:

```js
propertyData[aliasMap[0]] = propertyData[aliasMap[0]] ?? propertyData[aliasMap[1]];
```

I have also shallow cloned `propertyData` so we aren't messing with `feature.propertes|currentProperties|...`

tephenavies (approved):

PR #5205, opened 2021-02-02
Clean up Terria.start and remove Magda specifics

nf-s:
### Clean up Terria.start and remove Magda specifics

For https://github.com/TerriaJS/saas/issues/118

- Rename `registryConfigurationId` to `clientConfigUrl` in server config 
- Load server config before client config + override clientConfigUrl

### Checklist


nf-s:
No longer required

PR #5206, opened 2021-02-02
Merge properties from vector tiles into feature info

tephenavies:
### What this PR does

Re-implement vector tile feature & csv row property merging in v8.

### Checklist

-   [x] Not easily testable (best I can think of is: set a custom regionMapping, have a small sample vector tileset with just a 0/0/0 tile, have a small sample csv that uses the tileset, emulate feature picking)
-  [x] I've updated CHANGES.md with what I changed.


tephenavies:
Whoops. This PR comes from my personal repo. I've now updated my `gh` cli so hopefully I don't do that again.

PR #5208, opened 2021-02-02
Tsify loader function

zoran995:
### What this PR does

Created a typescript version of `Loader`. Was receiving the following error on ci  `Cannot create styled-component for component: [object Object].`, and this resolves it.

### Checklist


tephenavies (changes requested):
Hey @zoran995. I fully support this change, but I get a white screen when I add ABC Photo Stories:
You can reproduce by going to http://ci.terria.io/zoran-loader-tsx/ and adding "ABC Photo Stories" to the map.

I think it breaks for a lot of TableMixin items.

zoran995:
Nice catch, forgot to update imports. It's resolved now

tephenavies (approved):
Looks good. I'll merge this after we publish 8.0.0-alpha.66 so we have some time to find additional problems.

PR #5211, opened 2021-02-03
Merge properties from vector tiles into feature info

tephenavies:
### What this PR does

Re-implement vector tile feature & csv row property merging in v8.

### Checklist

-   [x] Not easily testable (best I can think of is: set a custom regionMapping, have a small sample vector tileset with just a 0/0/0 tile, have a small sample csv that uses the tileset, emulate feature picking)
-  [x] I've updated CHANGES.md with what I changed.

### DIFF
--- a/lib/ModelMixins/TableMixin.ts
+++ b/lib/ModelMixins/TableMixin.ts
@@ -717,7 +717,7 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
 
                 return this.featureInfoFromFeature(
                   regionType,
-                  d,
+                  Object.assign({}, feature.properties, d),
                   feature.properties[regionType.uniqueIdProp]
                 );
               }
--- a/lib/ModelMixins/TableMixin.ts
+++ b/lib/ModelMixins/TableMixin.ts
@@ -717,7 +717,8 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
 
                 return this.featureInfoFromFeature(
                   regionType,
-                  Object.assign({}, feature.properties, d),
+                  // Preserve values from d and insert feature properties after entries from d
+                  Object.assign({}, d, feature.properties, d),
                   feature.properties[regionType.uniqueIdProp]
                 );
               }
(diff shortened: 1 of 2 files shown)


lib/ModelMixins/TableMixin.ts
```diff
@@ -717,7 +717,7 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
 
                 return this.featureInfoFromFeature(
                   regionType,
-                  d,
+                  Object.assign({}, feature.properties, d),
```
nf-s:
```suggestion
                  Object.assign(
                    {},
                    d,
                    // Only add feature.properties which don't exist in $d
                    Object.entries(feature.properties).reduce<any>(
                      (props, [key, value]) => {
                        if (d![key] === undefined) {
                          props[key] = value;
                        }
                        return props;
                      },
                      {}
                    )
                  ),
```

Is it worth reversing the order of properties so column values are listed above region feature properties in FeatureInfoSection?

nf-s (changes requested):

lib/ModelMixins/TableMixin.ts
tephenavies: Good catch. I raise your suggestion with:
```suggestion
                  Object.assign(d, Object.assign({}, feature.properties, d)),
```

I know this modifies d, but we were modifying it previously in `featureInfoFunc`. Or:
```suggestion
                  Object.assign({}, d, Object.assign({}, feature.properties, d)),
```
Or we could use what you've written if you think that conveys intent better.

------

(the same as above but with a quick comment, but maybe not a good comment)
```suggestion
                  // Preserve values from d and insert feature properties after entries from d
                  Object.assign(d, Object.assign({}, feature.properties, d)),
```

Not modifying d:
```suggestion
                  // Preserve values from d and insert feature properties after entries from d
                  Object.assign({}, d, Object.assign({}, feature.properties, d)),
```

nf-s: Yep the last one looks good to me!
But is it equivalent to `Object.assign({}, d, feature.properties, d)`?

image

nf-s: > Properties in the target object are overwritten by properties in the sources if they have the same key. Later sources' properties overwrite earlier ones.

tephenavies: Yep. That's equivalent

tephenavies:
```suggestion
                  // Preserve values from d and insert feature properties after entries from d
                  Object.assign({}, d, feature.properties, d),
```

nf-s (approved):
soz forgot to merge

PR #5235, opened 2021-02-11
Fix tool switching

na9da:
### What this PR does

Merge after #5232

Fixes app crash when switching between multiple tools.

eg: Open item search tool, then open pedestrian mode tool, witness the app crash :boom: 

Tool mode uses [SplitPoint](https://github.com/TerriaJS/terriajs/blob/cc0db318b94b75f6b12abf518db129deb0a4f268/lib/ReactViews/SplitPoint.jsx#L21) for lazy loading tool components. 

SplitPoint uses react state variable to track the loaded component and during render it gets forwarded the latest props. 

This works fine when opening/re-opening the same tool component. But when opening a new tool, since react state updates lags behind `render()` the old component will receive the new props meant for the new component. This causes the app to crash.

I have re-written `Tool` component using react-lazy and ErrorBoundary also fixing the original problem.

### Checklist

### DIFF
--- a/lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx
+++ b/lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx
@@ -216,7 +216,7 @@ class DateTimePicker extends React.Component<PropsType> {
       granularity: defaultGranularity
     };
 
-    window.addEventListener("click", this.closePickerEventHandler.bind(this));
+    window.addEventListener("click", this.closePickerEventHandler);
 
     // Update currentDateIndice when currentDate changes
     this.currentDateAutorunDisposer = reaction(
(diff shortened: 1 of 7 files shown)


lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx
```diff
@@ -216,7 +216,7 @@ class DateTimePicker extends React.Component<PropsType> {
       granularity: defaultGranularity
     };
 
-    window.addEventListener("click", this.closePickerEventHandler.bind(this));
+    window.addEventListener("click", this.closePickerEventHandler);
```
na9da, line 220: This is not related to the main issue, but some error I found during testing. We are not removing the same event listener we bound to!

KeyboardSounds (approved):

KeyboardSounds (approved):

na9da:
Fixed all i18next translation errors. Merging with the localeString error which has already in the #next branch.

PR #5247, opened 2021-02-16
Fix WMS GetLegendGraphic style param

nf-s:
#### How we determine WMS legends (in order)

1.   Defined manually in catalog JSON
2.   If style is undefined, and server doesn't support GetLegendGraphic, we must select first style as default - as there is no way to know what the default style is, and to request a legend for it
3.   If style is is set and it has a legendUrl -> use it!
4.   If server supports GetLegendGraphic, we can request a legend (with or without style parameter)

### Fix WMS GetLegendGraphic `style` param

Fixes https://github.com/TerriaJS/terriajs/issues/5021

I forgot to add `style` parameter to `GetLegendRequest`, so:
- if a `style` was set and didn't provide a `legendUrl`
- and the server `supportsGetLegendGraphic`
- then we would request `GetLegendGraphic` for the default style (that is `style` is undefined)

So I added `style` to the request if it is defined.

Sorry no example catalog items - I haven't seen anyone ever use this, but it follows the spec.

image

### Other changes

- Loosen Legend `mimeType` check - so now it will treat the Legend URL as an image if the `mimeType` matches **OR** the file extension matches (previously, if `mimeType` was defined, then it wouldn't look at filetype extension)

#### How to test

- https://services.ga.gov.au/gis/great-artesian-basin/wms?REQUEST=GetCapabilities&SERVICE=WMS
- `Potentiometric Surface` -> `Potentiometric Surface 1900-1920`
- Change style to `Potentiometric Surface Grayscale`

### Checklist


lib/ReactViews/Workbench/Controls/Legend.tsx
```diff
@@ -33,11 +30,10 @@ const DISPLAYABLE_MIME_TYPES = [
...
-  if (legend.urlMimeType) {
-    return !!DISPLAYABLE_MIME_TYPES[legend.urlMimeType];
-  }
-
-  return !!legend.url?.match(IMAGE_URL_REGEX);
+  return (
+    (legend.urlMimeType && !!DISPLAYABLE_MIME_TYPES[legend.urlMimeType]) ||
+    !!legend.url?.match(IMAGE_URL_REGEX)
+  );
 }
```
nf-s, line 37: Loosen Legend `mimeType` check - so now it will treat the Legend URL as an image if the `mimeType` matches **OR** the file extension matches (previously, if `mimeType` was defined, then it wouldn't look at filetype extension)

nf-s: This is because some people use weird `mimeTypes` for legends (eg `image/png;charset=UTF-8`) - but have correct file extensions (eg `png`)

nf-s: If a Legend fails `checkMimeType` - this will be displayed

image

tephenavies (approved):

tephenavies:
Whoops

PR #5249, opened 2021-02-17
Upgrade @babel/parser and eslint

zoran995:
### What this PR does

Fixes 
Upgrade `babel-eslint` to `@babel/eslint-parser` and new version of `eslint`. There was a bunch of unexpected token errors/warnings in vscode due to some changes in used babel modules.
Most of the changes are regarding
```
warning  'require' is already defined as a built-in global variable  no-redeclare
```
otherwise, this would be really in-place upgrade.

### Checklist

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,22 @@
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "corejs": 3,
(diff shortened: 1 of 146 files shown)


tephenavies:
Changes merged into #5253.

PR #5253, opened 2021-02-19
Fix CI

tephenavies:
### What this PR does

Fix CI

Closes #5249.

### DIFF
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -582,28 +582,24 @@ export default class Terria {
     const launchUrlForAnalytics =
       options.applicationUrl?.href || getUriWithoutPath(baseUri);
     return loadJson5(options.configUrl, options.configUrlHeaders)
-      .then((config: any) => {
-        return runInAction(() => {
-          // If it's a magda config, we only load magda config and parameters should never be a property on the direct
-          // config aspect (it would be under the `terria-config` aspect)
-          if (config.aspects) {
-            return this.loadMagdaConfig(
-              options.configUrl,
-              config,
-              baseUri
-            ).then(() => {
-              Internationalization.initLanguage(
-                this.configParameters.languageConfiguration,
-                options.i18nOptions
-              );
-            });
-          }
-
+      .then(async (config: any) => {
+        // If it's a magda config, we only load magda config and parameters should never be a property on the direct
+        // config aspect (it would be under the `terria-config` aspect)
+        let languageConfiguration: LanguageConfiguration | undefined;
+        if (config.aspects) {
+          await this.loadMagdaConfig(options.configUrl, config, baseUri);
+          languageConfiguration = this.configParameters.languageConfiguration;
+        }
+        runInAction(() => {
           // If it's a regular config.json, continue on with parsing remaining init sources
           if (config.parameters) {
             this.updateParameters(config.parameters);
+            languageConfiguration = config.parameters.languageConfiguration;
+          }
+
+          if (!options.i18nOptions?.skipInit) {
             Internationalization.initLanguage(
-              config.parameters.languageConfiguration,
+              languageConfiguration,
               options.i18nOptions
             );
           }
--- a/lib/Table/TableAutomaticStylesStratum.ts
+++ b/lib/Table/TableAutomaticStylesStratum.ts
@@ -208,11 +208,12 @@ export class ColorStyleLegend extends LoadableStratum(LegendTraits) {
             })
           ]
         : [];
-    let numberFormatOptions: JsonObject | undefined = undefined;
-    if (colorColumn !== undefined) {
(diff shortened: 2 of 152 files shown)


zoran995:
The solution for i18next errors is to use an older version of the i18next, just tested with fixed version i18next@19.8.7 and everything works correctly. It will require more investigation to update to the latest version (19.8.9) of `i18next`.

The issue with the locales is coming from the implementation locale, which sometimes might be defined to some other value than "en" or "en-AU", for example specifying locale "bs-BS" will return `A$ 1,500`. To fix it we should introduce locale trait (default undefined) and use it in `toLocaleString`, or use `toLocaleString` with undefined locale in tests. This is at least from my understanding of Intl API

tephenavies:
@zoran995 I think the tests are badly written. At most we should test `toLocaleString` specifically with "en-au" (and maybe "en") locale(s) to ensure that our use of Intl is correct and adds thousands separators to numbers.

For i18next I've done some digging. I've changed initialising to call `i18next.init` only once before any specs. Now I'm tracking down why i18next starts to return English results (instead of "cimode") for any specs run after this spec: 
```
Terria terria start via loadMagdaConfig should dereference uniqueId to `/`
```

lib/Models/Terria.ts
```diff
@@ -582,28 +582,24 @@ export default class Terria {
...
           // If it's a regular config.json, continue on with parsing remaining init sources
           if (config.parameters) {
             this.updateParameters(config.parameters);
+            languageConfiguration = config.parameters.languageConfiguration;
+          }
+
+          if (!options.i18nOptions?.skipInit) {
             Internationalization.initLanguage(
-              config.parameters.languageConfiguration,
+              languageConfiguration,
```
tephenavies, line 602: Refactored slightly so that `Internationalization.initLanguage` is only called from one place with different arguments if from magda config or config.json. Also moved runInAction to wrap the synchronous parts.

The only change in logic is that `initLanguage` can now be skipped using the option `i18nOptions.skipInit`.

lib/Table/TableAutomaticStylesStratum.ts
```diff
@@ -283,7 +281,10 @@ export class ColorStyleLegend extends LoadableStratum(LegendTraits) {
     ];
   }
 
-  private _formatValue(value: number, format: JsonObject | undefined): string {
+  private _formatValue(
+    value: number,
+    format: Intl.NumberFormatOptions | JsonObject | undefined
+  ): string {
```
tephenavies, line 287: Just a small refactor. Earlier in the PR I made a change to this, then rolled it back and left the refactor.

nf-s (changes requested):

package.json
```diff
@@ -154,10 +154,11 @@
     "worker-loader": "^2.0.0"
   },
   "devDependencies": {
+    "@babel/eslint-parser": "^7.12.16",
+    "@babel/plugin-proposal-decorators": "^7.12.13",
```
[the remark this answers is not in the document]
tephenavies: Babel stuff normally goes into devDependencies, but I think in this case we need to put any babel stuff related to transpiling into dependencies.
```suggestion
```

nf-s (approved):

zoran995:
@steve9164  just figured out that we could just add `fallbackLng: "cimode"` for specs instead of all work for `skipInit`. Do we need `skipInit` for anything else (I wasn't a big fan of adding it from start)? 
We also need to lock the version of `i18next` on 19.8.9 till `i18next-http-backend` pick with the latest changes, the install is failing (creating a PR for it).

tephenavies:
@zoran995 According to https://www.i18next.com/overview/api we should never call `i18next.init` more than once. Setting fallbackLng might work for this version but we don't have any assurances this won't change in a minor or patch version in the future.

I added `skipInit` so that you could customise your i18next setup without modifications to terriajs further than just adding languageOverrides. Maybe there's a better way. The other way I was considering was to stop using the global default i18next instance and instead create an i18next instance, maybe storing it in `Terria`. Then the way to initialise yourself (in CI or custom TerriaMap) would be to pass in your own i18next instance. But there are a lot of places where a Terria object isn't available. (Those places are probably a subset of the places where results from i18next.t won't update when `changeLanguage` is called, so we may end up modifying them anyway if we want to support "hot" language changing).

zoran995:
Call another to i18next.init from tests is 99.9% my mistake, if in the first place (when implementing it for tests) I mocked the `Internationalization.initLanguage` and placed `i18next.init` in it (callFake) we wouldn't face this issue. The best call is to leave this as is and move to other issues, I wasn't thinking that this enables setting different i18next configuration from TerriaMap nice point 😃

Regarding hot language reloading, there might be some edge cases but generally, it works out of the box (at least in our implementation). Promised a few months ago that I will create a PR to implement it without UI so users can just connect to it but didn't got enough time for it

Issue #4191, opened 2020-04-01
Support for date and number formating


zoran995:
Since internationalization support is introduced some time ago, it would be also nice to add support for localized date and number formats. 
Dates example, Germany (as well as many European countries) is using format DD.MM.YYYY (01.04.2020), while in USA date format is MM/DD/YYYY (04/01/2020).
Numbers example, in Germany thousand delimiter is space and a decimal delimiter is a comma (10 000,05), while in a lot of other countries thousand delimiter is a comma and a decimal delimiter is a dot (10,000.05).
I want to hear another opinion about this since it might require to introduce at least one new library (moment-js and numeral-js), need to check more about libraries. Currently, the date format is fixed on dd/mm/yyyy.

rowanwins:
Hi @zoran995 

Date formatting in mobx is partially addressed in https://github.com/TerriaJS/terriajs/pull/4235

I haven't looked at number formatting but we could possibly look at using the browser native [`.tolocalestring` method](https://www.w3schools.com/jsref/jsref_tolocalestring_number.asp). I'm trying to work out where in the UI we'd need to apply this... feature info, legends, anywhere else...?

zoran995:
Didn't know about date formatting. Is it possible to make default date formatting to depend on the format defined for current language, perhaps format might be defined in language translation file?

For numbering format maybe also measure tool result, location bar. I wasn't sure if we are using numbers as parameters user can insert, so I looked into library that can make it easy to go between formated value and original value. Recently I was doing some internal changes for locationBar and since latitude and longitude are stored as string it caused some problems.

PR #5261, opened 2021-02-21
Fix ci and expose locale trait for table columns

zoran995:
### What this PR does

Fixes CI

This PR is targeting `next-fix-ci` branch feel free to change to `next` if think that it is better (this includes all changes from #5253). Everything is done according to the recommendation in https://github.com/TerriaJS/terriajs/pull/5253#issuecomment-782620714

### Checklist

### DIFF
--- a/lib/Table/TableAutomaticStylesStratum.ts
+++ b/lib/Table/TableAutomaticStylesStratum.ts
@@ -208,20 +208,29 @@ export class ColorStyleLegend extends LoadableStratum(LegendTraits) {
             })
           ]
         : [];
-    const numberFormatOptions:
-      | Intl.NumberFormatOptions
-      | JsonObject
-      | undefined = { useGrouping: true };
-    if (colorColumn?.traits?.format !== undefined) {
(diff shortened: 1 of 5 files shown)


tephenavies:
@zoran995 This fixes our current CI problem (if you switch $1,500 & A$1,500 in specs), but I don't really want to add the ability to set locale in traits for number formatting. We should probably follow the browser's locale settings for all formatted numbers.

zoran995:
We might leave in the description that it is recommended to not define the locale and leave it to browser default (in eastern Europe countries there is an issue that most users have default en settings but they should see the results in the specific locale).
We might look into setting a locale for git actions

tephenavies:
I've decided to disable the 2 tests that fail due to being run in an unexpected locale, so I will close this PR for now. If this feature is important outside of CI tests can you open an issue describing the problem it solves so we can discuss further? Thanks

zoran995:
@steve9164 there is already some discussion in #4191 regarding returning the date in a specific locale based on the language. This PR went in little different direction, in my opinion better option would be to have possibility setting output locale based on language.

PR #5282, opened 2021-02-25
Add dd/mm/yyyy date format support

tephenavies:
### What this PR does

Add back support for dd/mm/yyyy common Australian date format to version 8.

### Checklist

### DIFF
--- a/lib/Table/TableColumn.ts
+++ b/lib/Table/TableColumn.ts
@@ -246,7 +246,6 @@ export default class TableColumn {
           return null;
         }
         return new Date(centuryFix(y), m - 1, d);
-        return new Date(`${centuryFix(y)}/${m}/${d}`);
       } else {
         parsingFailed = true;
         return null;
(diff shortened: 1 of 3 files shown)


lib/Table/TableColumn.ts
```diff
@@ -158,47 +158,163 @@ export default class TableColumn {
...
+      const d = +dayString;
+      const m = +monthString;
+      const y = +yearString;
+      if (Number.isInteger(d) && Number.isInteger(m) && Number.isInteger(y)) {
+        if (d > 31 || m > 12 || y > 9999) {
+          parsingFailed = true;
+          return null;
         }
+        return new Date(centuryFix(y), m - 1, d);
+        return new Date(`${centuryFix(y)}/${m}/${d}`);
```
nf-s: There are two returns here

nf-s: and omg months start at 0 - that is very upsetting

nf-s (changes requested):
Looks good, just needs that extra return statement removed.

At some point we should bring in missing date parsing from master - https://github.com/TerriaJS/terriajs/blob/6153d332aa3bd160ddbbd968be37a339a4f6ad96/lib/Map/TableColumn.js#L485

For example `2020-Q1` type format

lib/Table/TableColumn.ts
tephenavies: ikr

tephenavies:
```suggestion
```

nf-s (approved):

PR #5297, opened 2021-03-03
Update catalog-converter

tephenavies:
### What this PR does

Update catalog-converter to latest published version. The default options have changed with this version so that the CLI and exported functions use the same conversion options by default.

### DIFF
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -1,4 +1,3 @@
-import { convertCatalog, convertShare } from "catalog-converter";
 import i18next from "i18next";
 import { action, computed, observable, runInAction, toJS, when } from "mobx";
 import { createTransformer } from "mobx-utils";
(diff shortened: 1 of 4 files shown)


lib/Models/Terria.ts
```diff
@@ -564,7 +563,8 @@ export default class Terria {
           .filter((v7initUrl: any) => isJsonString(v7initUrl))
           .map(async (v7initUrl: string) => {
             const catalog = await loadJson5(v7initUrl);
-            const convert = convertCatalog(catalog);
+            const { convertCatalog } = await import("catalog-converter");
+            const convert = convertCatalog(catalog, { generateIds: false });
```
na9da: Just a suggestion. We could maybe use `Promise.all` here to do the load & import requests in parallel? Maybe even extract it out as an async function called `convertV7Catalog` if the resulting code is a bit dense.

na9da (approved):
Looks great @steve9164. 

I just made a small suggestion which you could consider implementing if it makes sense.

lib/Models/Terria.ts
tephenavies: Done

PR #5314, opened 2021-03-09
Quaterly dates support (eg 2020-Q1)

nf-s:
Add support for TableColumn quarterly dates in the format yyyy-Qx (eg 2020-Q1)

Test CSV https://gist.githubusercontent.com/nf-s/c7968085cf15db1a32d130d77fbb9110/raw/618037a0a3ccd39c9e3a422b6f01476e75ee387c/yyyy-Qx.csv

### Checklist

### DIFF
--- a/lib/Table/TableColumn.ts
+++ b/lib/Table/TableColumn.ts
@@ -252,6 +252,30 @@ export default class TableColumn {
       }
     };
 
+    let yyyyQQ: StringToDateFunction = value => {
+      // Is it quarterly data in the format yyyy-Qx ? (Ignoring null values, and failing on any purely numeric values)
+      if (value.indexOf("-Q") === 4) {
+        var year = value.slice(0, 4);
+        var quarter = value.slice(6);
+        var monthString;
+        if (quarter === "1") {
+          monthString = "01/01";
+        } else if (quarter === "2") {
+          monthString = "04/01";
+        } else if (quarter === "3") {
+          monthString = "07/01";
+        } else if (quarter === "4") {
+          monthString = "10/01";
+        } else {
+          parsingFailed = true;
+          return null;
+        }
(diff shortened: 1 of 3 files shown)


lib/Table/TableColumn.ts
```diff
@@ -252,6 +252,30 @@ export default class TableColumn {
       }
     };
 
+    let yyyyQQ: StringToDateFunction = value => {
+      // Is it quarterly data in the format yyyy-Qx ? (Ignoring null values, and failing on any purely numeric values)
+      if (value.indexOf("-Q") === 4) {
```
na9da: I think this will be a tad more faster when parsing many rows if we do something like `value.slice(4,6) === "-Q"` as we wont be searching the whole string. Or even better `value[4] === '-' && value[5] === 'Q'`.

lib/Table/TableColumn.ts
```diff
@@ -252,6 +252,30 @@ export default class TableColumn {
...
+          monthString = "04/01";
+        } else if (quarter === "3") {
+          monthString = "07/01";
+        } else if (quarter === "4") {
+          monthString = "10/01";
+        } else {
+          parsingFailed = true;
+          return null;
+        }
+        return new Date(year + "/" + monthString);
```
na9da: Maybe we should also handle the situation where `year` is invalid and the whole date object becomes invalid.

na9da:
Looks great @nf-s. I have just made a couple of suggestions.

lib/Table/TableColumn.ts
nf-s: done!

lib/Table/TableColumn.ts
nf-s: yep good idea!

na9da (approved):

PR #5315, opened 2021-03-09
Region map timeseries improvements

nf-s:
### Region map timeseries improvements

* Added `momentChart` to region-mapped timeseries
* Add time-series chart (in FeatureInfo) for region-mapped timeseries
* Only show `TableMixin` chart if it has more than one
* Add `TableChartStyle` name trait.

Test CSV https://gist.githubusercontent.com/nf-s/91e162f2fef03ac6df3d0f9a3eaae4dd/raw/755dab77221da47f020d409bb4bb793ccb73b976/regiontime.csv

Click on region to see time-series chart

### Checklist

-   [x] Not sure how to write tests for this stuff


lib/ModelMixins/TableMixin.ts
```diff
@@ -692,24 +702,65 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
...
+                    yName: this.activeTableStyle.colorColumn?.title,
+                    units: chartColumns.map(column => column.units || ""),
+                    csvData: [
+                      chartColumns.map(col => col!.title).join(","),
+                      ...regionIds.map(i =>
+                        chartColumns
+                          .map(col => col.valueFunctionForType(i))
+                          .join(",")
+                      )
+                    ].join("\n")
```
na9da: Maybe we could split this off into a new method called `setChartDetailsFunctionForFeatureInfo` to simplify the code a bit?

nf-s: I took some stuff out of `private readonly createRegionMappedImageryLayer` and put it into two new functions `getImageryLayerFilteredRows` and `getImageryLayerFeatureInfo`

na9da: That's better than my suggestion and looks less dense now!

na9da (approved):

Issue #5233, opened 2021-02-11
Fix AsyncLoader.isLoading + other loading issues


nf-s:
At the moment, we can't observe `isLoading` without cause these errors:

```
Error: [mobx] Computed values are not allowed to cause side effects by changing observables that are already being observed. Tried to modify...
```

nf-s:
Might be able to do something with `fromPromise(this.loadKeepAlive).state`...

tephenavies:
That actually might not work because I think observing `fromPromise(this.loadKeepAlive)` will cause the AsyncLoader to reload without `load` being called again.

nf-s:
Yep! that is what happened when I tried it.
I think that approach may work as long as we wait until the first `load` is called before observing `this.loadKeepAlive`.

tephenavies:
I think it'll still continue to reload in some circumstances without subsequent calls to `load`. It's important to the mechanics of AsyncLoader that `this.loadKeepAlive` is never observed. I think `.load` should even be an action so that calling `.load` from inside a computed doesn't end up with `this.loadKeepAlive` being observed.

KeyboardSounds:
Also see https://github.com/TerriaJS/terriajs/issues/5324

tephenavies:
Just talked to @nf-s about potential AsyncLoader improvements.

One thing we discussed is that AsyncLoder doesn't work when promises need to be chained. It's not possible to chain AsyncLoaders to do this. However you could apply the inner idea of AsyncLoader (which is to run parts where you decide/calculate how to load data/metadata etc. inside a keepAlive computed that's unobserved) and allow creating these keepAlives from promises. If you extend this to making your own `mobx.flow` function you could achieve something like this:
```js
const load = loaderFlow(function*() {
  const url = this.url;
  const res = yield loadFromJson(url);
  yield function() {
    this.setTrait(x, y, res.foo);
  };
  const url2 = `${this.url}/${res.path}`;
  const res2 = yield loadFromJson(url2);
  yield function() {
    this.setTrait(x2, y2, res2.bar);
  }
});
```
Note due to the computeds that will be applied to each synchronous part any actions that need to be performed need to be yielded as functions. These would be wrapped in `action`.

Potential difficulties:
* Avoiding reloading from secondary URLs when data from the first load is the same and it could be detected that loading from the secondary URL is not necessary
* Working out how to clean up the keepAlives. Hopefully `loaderFlow` could keep track of these

nf-s:
[...]

PR #5322, opened 2021-03-12
Initial message

### LINKED ISSUE #5160: Port `initialMessage` property for CatalogMember to v8

KeyboardSounds:
### What this PR does

Fixes #5160

Plus, the Notification refactor that nobody asked for!

### Checklist


lib/ModelMixins/AsyncMappableMixin.ts
```diff
@@ -29,7 +30,8 @@ function AsyncMappableMixin<T extends Constructor<Model<MappableTraits>>>(
      * If the map items are already loaded or already loading, it will
      * return the existing promise.
      */
-    loadMapItems(): Promise<void> {
+    async loadMapItems(): Promise<void> {
+      await this.showInitialMessageIfRequired();
```
nf-s: I'm not sure this is the best place for this. What about putting something like:
```ts
if (ShowableMixin.isMixedInto(item) {
   await item.showInitialMessageIfRequired();
}
```
in https://github.com/TerriaJS/terriajs/blob/6c205d9fcb48857b098993e5723dedd5a5e24683/lib/Models/Workbench.ts#L195-L201

CHANGES.md
```diff
@@ -12,6 +12,8 @@ Change Log
 * Add support for TableColumn quarterly dates in the format yyyy-Qx (eg 2020-Q1).
 * Fix region mapping feature highlighting.
 * Update clipboard to fix clipboard typing error.
+* Move notification state change logic from ViewState into new class NotificationState
+* Catalog items can now show a disclaimer or message before loading through specifying InitialMessageTraits
 * [The next improvement]
```
nf-s: Can you please add an entry about `ShowableMixin`

lib/ReactViewModels/ViewState.ts
```diff
@@ -24,6 +22,7 @@ import {
   TourPoint
 } from "./defaultTourPoints";
 import DisclaimerHandler from "./DisclaimerHandler";
+import NotificationState from "./NotificationState";
```
nf-s: There is still a `Notification` interface in this file

nf-s: (at line 43 - it won't let me comment on it)

lib/ReactViewModels/ViewState.ts
```diff
@@ -319,37 +318,15 @@ export default class ViewState {
     this.terria = terria;
 
     this._unsubscribeNotificationListener = terria.notification.addEventListener(
```
nf-s: I think that if we're refactoring `Notification` stuff, we should probably get rid of the un-typed Cesium event handler in 

https://github.com/TerriaJS/terriajs/blob/20304b657d5e96766009f770a0c985f2345f8264/lib/Models/Terria.ts#L258

I know that `notificationState` should probably be in `ViewState`, but it needs to be visible to the entire codebase (especially the model layer). Could we maybe move it into `Terria.ts` and remove it from `ViewState`? As everywhere has access to the `terria` object, and then we can get rid of the `CesiumEvent` passing `Notifications` between `Terria` and `ViewState`.

lib/ReactViewModels/ViewState.ts
nf-s: https://github.com/TerriaJS/terriajs/blob/5b7d0155473d3537b7a2402a6fe40566080113cc/lib/ReactViewModels/ViewState.ts#L42-L53

lib/ReactViewModels/ViewState.ts
```diff
@@ -639,14 +616,10 @@ export default class ViewState {
...
-  }
-
   hideMapUi() {
     return (
-      isDefined(this.getNextNotification()) &&
-      this.getNextNotification()!.hideUi
+      this.notificationState.currentNotification !== undefined &&
+      this.notificationState.currentNotification!.hideUi
     );
   }
```
nf-s: Can we make this a `@computed` property?

lib/ReactViews/Notification/Notification.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const viewState = props.viewState;
+  const notificationState = viewState?.notificationState;
+  const notification = notificationState?.currentNotification;
+
+  if (
+    viewState === undefined ||
+    notificationState === undefined ||
+    notification === undefined
+  ) {
+    return <></>;
```
nf-s, line 19: I like returning `null` in these situations

nf-s: probs doesn't matter though

lib/Traits/CatalogMemberTraits.ts
```diff
@@ -4,6 +4,7 @@ import anyTrait from "./anyTrait";
 import LegendTraits from "./LegendTraits";
 import ModelTraits from "./ModelTraits";
 import objectArrayTrait from "./objectArrayTrait";
+import objectTrait from "./objectTrait";
```
nf-s:
```suggestion
```

nf-s (changes requested):
When I click on an item in the Data Catalog (using the default `simple` catalog)  - I get this

image

I also get this, when I click on `Marine water quality grids for the Great Barrier Reef region - monthly data` and then `number_of_valid_observations`

image

### EDIT

The change I suggest here - https://github.com/TerriaJS/terriajs/pull/5322#discussion_r592938193 - fixes the `@computed` error

Ahhh #5233 is going to be the end of me

lib/ReactViewModels/ViewState.ts
nf-s: This approach could also be used for the `terria.error` Cesium event handler

KeyboardSounds:
TODO: Rewrite to observe changes to viewstate in an autorun, and run then, rather than in load

nf-s:
Also, when you merge in `next` please take into account changes made to `Notification.jsx` here - https://github.com/TerriaJS/terriajs/pull/5413 

https://github.com/TerriaJS/terriajs/commit/08ea7a227a36902a00357710306225aa6f40e6cb#diff-a07b9cfd11367978cf4ae8e66b86910d059019e214c2a9bb5bf7403b2f95c617

nf-s (changes requested):

lib/ModelMixins/AsyncMappableMixin.ts
nf-s: We discussed this, you were right

lib/ReactViewModels/ViewState.ts
nf-s: Never mind, not needed right now

lib/Models/Terria.ts
```diff
@@ -1632,10 +1635,14 @@ async function interpretHash(
 
         // Show warning messages if converted
         if (result.converted) {
-          terria.notification.raiseEvent({
+          // terria.notification.raiseEvent({
+          //   title: i18next.t("share.convertNotificationTitle"),
+          //   message: shareConvertNotification(result.messages)
+          // } as Notification);
```
nf-s: stray comments

lib/ModelMixins/MappableMixin.ts
```diff
@@ -7,8 +7,10 @@ import ImageryProvider from "terriajs-cesium/Source/Scene/ImageryProvider";
 import AsyncLoader from "../Core/AsyncLoader";
 import Constructor from "../Core/Constructor";
 import Model from "../Models/Model";
+import saveModelToJson from "../Models/saveModelToJson";
```
nf-s: this is unused

lib/ModelMixins/ShowableMixin.ts
```diff
@@ -0,0 +1,66 @@
...
+import i18next from "i18next";
+import Property from "terriajs-cesium/Source/DataSources/Property";
+import { action, computed, observable } from "mobx";
+
+const t = i18next.t.bind(i18next);
+
+type ShowableModel = Model<ShowableTraits>;
+type InitialMessageJSON = {
+  [Property in keyof InitialMessageTraits]: any;
+};
```
nf-s: This isn't used

lib/ModelMixins/ShowableMixin.ts
```diff
@@ -0,0 +1,66 @@
+import Model from "../Models/Model";
+import Constructor from "../Core/Constructor";
+import ShowableTraits, { InitialMessageTraits } from "../Traits/ShowableTraits";
+import i18next from "i18next";
+import Property from "terriajs-cesium/Source/DataSources/Property";
```
nf-s: This isn't used

nf-s (changes requested):

nf-s (changes requested):
Did we decide to get rid of `ShowableTraits`? I can't remember

nf-s (approved):

PR #5323, opened 2021-03-12
Always initialise i18n

tephenavies:
### Checklist


nf-s (approved):
Hey @steve9164 I made a few changes;

- tsified `TerriaError` and `raiseErrorToUser`
- created `wrapErrorMessage` function to wrap error message in something more user friendly (using `models.raiseError.errorMessage` translation string)
- created a `TerriaError` if loading config fails

Are you happy with these changes?

lib/Core/TerriaError.ts
```diff
@@ -0,0 +1,32 @@
...
+  sender?: unknown;
+  /** True if the user has seen this error; otherwise, false. */
+  raisedToUser?: boolean;
+}
+
+/**
+ * Represents an error that occurred in a TerriaJS module, especially an asynchronous one that cannot be raised
+ * by throwing an exception because no one would be able to catch it.
+ */
+export default class TerriaError implements TerriaErrorOptions {
```
tephenavies: It seems weird to me that a class implements its own initialisation options. Is there anywhere that this property is used?

nf-s: Yeah I just did that to drag across JSDoc comments. I can get rid of it

nf-s: i got rid of it

PR #5328, opened 2021-03-16
 fix CSW xml types

nf-s:
### Fix CSW xml types

See comment below

### Checklist

-   [x] I don't think tests are required


lib/Models/CswCatalogGroup.ts
```diff
@@ -325,7 +325,10 @@ class CswStratum extends LoadableStratum(CswCatalogGroupTraits) {
       records.push(...(json?.SearchResults?.Record ?? []));
 
       // Get next start position - or stop pageing
-      const nextRecord = json?.SearchResults?.nextRecord;
+      const nextRecord =
+        typeof json?.SearchResults?.nextRecord === "string"
+          ? parseInt(json?.SearchResults?.nextRecord ?? "0")
+          : json?.SearchResults?.nextRecord;
```
nf-s: When paging for CSW records, we use the `nextRecord` ID in the response to set `startPosition`. 

This was a `string`, but should have been a `number` - so we were getting this error:

```
startPosition must be an integer greater than zero.
```

KeyboardSounds:
Github won't let me add comments because it's outside the part of the code you changed, but I can see places in the code that assume `numberOfRecordsMatched` and `numberOfRecordsReturned` are numbers but they could be strings. Is that okay?

nf-s:
> Github won't let me add comments because it's outside the part of the code you changed, but I can see places in the code that assume `numberOfRecordsMatched` and `numberOfRecordsReturned` are numbers but they could be strings. Is that okay?

Fixed!

KeyboardSounds (approved):
Once you make sure this PR isn't the thing breaking CI, merge away.

PR #5332, opened 2021-03-17
Fix issue #5331.

### LINKED ISSUE #5331: CkanCatalogGroup should handle filterQuery correctly
The current implementation assumes that each element in filterQuery array of CkanCatalogGroup is an Object that has key-value pair such as
```
{
  "fq": "(organization:alpine-shire-council AND res_format:(geojson OR GeoJSON OR kml OR KML OR kmz OR KMZ OR wms OR WMS OR CSV-GEO-AU OR csv-geo-au OR aus-geo-csv OR \"Esri REST\"))"
}
```

However, an array element can also be a pure query string such as 
```
  "fq=+(res_format%3Awms%20OR%20res_format%3AWMS)"
```
The code should also handle that case.

mwu2018:
### What this PR does

Fixes #5331 

### Checklist


lib/Models/CkanCatalogGroup.ts
```diff
@@ -84,9 +84,13 @@ export class CkanServerStratum extends LoadableStratum(CkanCatalogGroupTraits) {
         .segment("api/3/action/package_search")
         .addQuery({ start: 0, rows: 1000, sort: "metadata_created asc" });
 
-      Object.keys(filterQuery).forEach((key: string) =>
-        uri.addQuery(key, filterQuery[key])
-      );
+      if (filterQuery.toString().includes("fq=")) {
+        uri.query(uri.query() + "&" + filterQuery.toString());
+      } else {
```
nf-s: If the `filterQuery` trait can have string elements, it needs to be reflected in the `Trait`

https://github.com/TerriaJS/terriajs/blob/6e4d4df9dc368b155d283abd48bbe0c788960616/lib/Traits/CkanCatalogGroupTraits.ts#L25-L43

nf-s: At the moment it only allows `JsonObject` elements

nf-s: A possible change could be:
```ts
  filterQuery?: {
    [key: string]: JsonValue | string;
  }[]
```

nf-s (changes requested):

lib/Models/CkanCatalogGroup.ts
mwu2018: If `filterQuery` can only have `JsonObject` elements. An element like "fq=+blahblah" is considered valid JsonObject.

mwu2018: E.g.  filterQuery?: Object[] = ["fq=+blahblah"];

Therefore I don't think the Trait needs to be changed. @nf-s

nf-s: String isn't a valid JsonObject;

```
export interface JsonObject {
  [key: string]: JsonValue;
}
```

But I just realised that my suggestion is totally wrong! I meant something like this:

```ts
filterQuery?: JsonObject[] | string[] = [
    "Some string",
    {
      fq:
        '(res_format:(czml OR CZML OR geojson OR GeoJSON OR WMS OR wms OR kml OR KML OR kmz OR KMZ OR WFS OR wfs OR CSV-GEO-AU OR csv-geo-au OR "Esri REST"))'
    }
  ];
```

nf-s: ~But also, why do we need a `string`?~
 
~Isn't `["fq=+blahblah"]` equivalent to `{fq: "blahblah"}`~

Never mind, just realised master supports this

mwu2018: @nf-s You are right. I made a mistake by using Object instead of JsonObject in the example
 ```filterQuery?: Object[] = ["fq=+blahblah"];```
which is valid but useless. Has adopted your recommendation, which will cause other improvement. See the new commit.

lib/Models/CkanCatalogGroup.ts
```diff
@@ -84,9 +85,13 @@ export class CkanServerStratum extends LoadableStratum(CkanCatalogGroupTraits) {
         .segment("api/3/action/package_search")
         .addQuery({ start: 0, rows: 1000, sort: "metadata_created asc" });
 
-      Object.keys(filterQuery).forEach((key: string) =>
-        uri.addQuery(key, filterQuery[key])
-      );
+      if (typeof filterQuery === "string") {
+        uri.query(uri.query() + "&" + filterQuery.toString());
```
nf-s: Looks good @mwu2018 
Sorry to nitpick, but is it possible to use `uri.addQuery()` instead of `uri.query(uri.query() + ...`?

nf-s (changes requested):

lib/Models/CkanCatalogGroup.ts
mwu2018: @nf-s You did not nitpick at all. Instead, you have opened a can of worms 😃 
1. The `filterQuery` trait is changed similar to what you suggested initially. Your later suggestion does not allow mixed elements in the array. 
2. We can not use `uri.addQuery(filterQuery)` because the URI encoder will escape `=`, which is not what we want. I have added some comments in the code.
3. The code is slightly refactored in order to do simple unit tests.

Can you please take another look?

nf-s (approved):

Issue #5219, opened 2021-02-07
Upgrade terriajs to typescript 4.x


na9da:
Currently when using typescript 4.x with terriajs, we get a bunch of error [TS2611](https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json#L2593). This ticket is to find a solution for the error and upgrade TS to the latest 4.x version.

nf-s:
I want [logical nullish assignment](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#short-circuiting-assignment-operators)! 💥

zoran995:
Thought that https://github.com/microsoft/TypeScript/pull/41994 might solve this issue, but just got the same error using it (maybe someone else wants to test it)

na9da:
Thanks @zoran995 for following it up with the TS team.

I was thinking if we can find a workaround for this particular error.

When building terriajs with TS4.x we have around 256 errors out of which 223 are TS2611 and out of them 132 happens in `LoadableStratum` definitions and 91 in other models and mixins.

I was looking at the particular case of `LoadableStratum`, here is a simplified example of what is happening in [LoadableStratum](https://github.com/TerriaJS/terriajs/blob/a2e6b7cbc23d88b0b83beb6c8dd645e686724059/lib/Models/Definition/LoadableStratum.ts).

```
export type Constructor<T> = new (...args: any[]) => T;
export type PropertiesOf<T> = { [K in keyof T]: T[K] };

// A class factory for generating LoadableStratum
export function LoadableStratum<T extends Constructor<ModelTraits>>(
  Traits: T
): Constructor<PropertiesOf<InstanceType<T>>> {
  type Traits = InstanceType<T>;

  abstract class Loadable {
     // ... code that dynamically defines getters/setters for traits
  }

  return Loadable as any;
}

class ModelTraits {}

class FooTraits extends ModelTraits {
   get foo() { return 0; }
}

class FooLoadableStratum extends LoadableStratum(FooTraits) {
    // TS2611 error here even though foo is a getter in FooTraits
    get foo() {
         return 1;
    }
}
```
[...]

PR #5342, opened 2021-03-19
Tsify styled components

zoran995:
### What this PR does

Most of the styled-components are written in javascript and not in TS so we have to import them using CJS format. This is a beginning to tsify them and move import to the ES module format. This need to go through an iterative process to determine the best rules enforcement since some options should not be allowed together as they will only confuse users.

I might also propose some of the changes to the styled-components (non-UI related) but to enforce more strict rules (i.e. button themes/types `primary`, `secondary`).

p.s. Upgrade to TypeScript v4 would enable us to write more strict types checks (i.e. for styledWidth) using [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html).

### Checklist


tephenavies:
Thanks @zoran995. I've wanted to get this done for a while now.

Those template literal types look odd. It's super cool what they'll allow us to do for Box & similar though. We're tracking our upgrade to Typescript 4 in #5219, which at the moment is waiting on support for accessors overriding properties (fixing TS2611 errors).

zoran995:
After doing a little investigation probably the best is to go with this as is and incrementally improve components. There is a bunch of places that components are overridden with custom style.

We might get inspiration for components from [Atlassian design system](https://atlassian.design/components/select/examples) (they are using styled-components)  ([source](https://bitbucket.org/atlassian/atlaskit-mk-2)). Also https://bitbucket.org/atlassian/design-system-mirror/src/master/design-system/

Also, the styled-system looks nice https://styled-system.com/. 
@steve9164 Is there a plan for terria to support multi-theme UI?

tephenavies:
Finally merging! Sorry for the delay.

I like the look of styled-system.

I'm not sure what you mean by multi-theme UI. Could you expand on that?

zoran995:
i.e. light/dark UI option for the user to choose between, or some other color themes

AnaBelgun:
That's interesting @zoran995 and it might be connected by our Theme-ing ticket / thinking https://github.com/TerriaJS/terriajs/issues/4661; feel free to comment / suggest options.

tephenavies:
@zoran995 It is possible to dynamically switch themes by setting `props.themeOverrides` on `StandardUserInterface`.

zoran995:
My idea was mainly should we go in direction to provide a possibility for end users of the application an option to choose from predefined color schemes (button in UI), in that case we would need to support multiple themes at the same time.

PR #5352, opened 2021-03-22
Fix new starter issues

wibily:
### What this PR does

Fixes missing linting dependencies when following the getting-started.md instructions for the first time.
Running the default `gulp` task for the first time after a fresh `npm install` fails as it is missing the lint dependencies.
This change makes it so that gulp build will be run as a npm post-install task.

Also fixes a typo in the docs

### Checklist

-   No unit tests as there are no source code changes

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Change Log
   
 * Updated terriajs-cesium to version 1.79.1
 * Make base maps configurable from init files and update documentation for init files [#5140](https://github.com/TerriaJS/terriajs/pull/5140).
+* Fixed missing dependency issue for a fresh build
 
 #### 8.0.0-alpha.68
 * Remove points from rectangle `UserDrawing`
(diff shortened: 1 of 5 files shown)


tephenavies:
Hey @wibily.
Does this fix the original issue? Both TerriaMap and terriajs have their own `gulpfile.js` and gulp tasks. Both have a `lint` and a `build` task. `gulp build` in terriajs (this repo) only builds the test suite. If `gulp lint` is broken in TerriaMap it likely needs to be fixed there.

When fixing it there can you let me know why you need to build first? We've been able to run `gulp lint` on a clean checkout forever.

Also it's a preference not to mirror `gulp` tasks in npm scripts. The reason we have `gulp` as a script is so users can run `npm run gulp build` or similar. We could instruct users to use `npx gulp x` instead, however I think `npx` has been removed in `npm` 7 anyway.

PR #5353, opened 2021-03-22
Fix load persisted basemap and share of basemaps

### LINKED ISSUE #5348: Share links don't set basemap
Related https://github.com/TerriaJS/terriajs/issues/4620

- Go to ci.terria.io/next
- Select basemap
- Create story
- Open story in another browser (or clear browser storage)

zoran995:
### What this PR does

Fixes #5348 (partially see comment below)

I wrongly placed some of the functions in #5140 so loading of persisted basemaps stopped working. This fixes that along with basemap share using baseMapId. 
This fixes sharing only for v8. v7 is using parameter baseMapName which probably needs to be mapped on catalog-converter level.

I am not completely happy with how sharing of basemap work, if the user already visited the map in the past shared version will be overridden with persisted basemap. But that is how it worked in the v7. To truly fix that sepparation for init data is needed so we can know exactly from where configuration comes.

### Checklist


nf-s:
Thanks @zoran995 ! 
I have merged this into this PR https://github.com/TerriaJS/terriajs/pull/5310 as there is some overlap

PR #5366, opened 2021-03-25
Hide basemap entry in basemap menu 

na9da:
### What this PR does

Added option `hideInBaseMapMenu` to hide basemap entry in basemap listing.

### Why we need this? 
Some of our basemaps are composite catalog items combining multiple base maps. For this use-case we need to add base map entries as member of the composite map item, but the member itself should not be listed in the base map menu. This option can be used to hide the composite member from the basemap menu.

```
"baseMaps": [
...
{
      "item": {
        "id": "basemap-topographic-member",
        "type": "wms",
        "url": "https://base.maps.someservice.com/service?service=wms",
        "layers":"SOME_LAYER",
        "tileWidth": 512,
        "tileHeight": 512
      },
      "image": "images/basemaps/topographic.png",
      "hideInBaseMapMenu": true
    },
  {
      "item": {
        "id": "basemap-topographic",
        "name": "VIC Topographic",
        "type": "composite",
        "members": ["basemap-topographic-member", "basemap-bing-roads"]
      },
      "image": "images/basemaps/topographic.png"
    }
]
```

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 ### MobX Development
 
 #### next release (8.0.0-alpha.70)
+* Add `hideInBaseMapMenu` option to `BaseMapModel`.
 * Change default basemap images to relative paths.
 * Add `tileWidth` and `tileHeight` traits to `WebMapServiceCatalogItem`.
 * [The next improvement]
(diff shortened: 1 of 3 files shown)


nf-s (changes requested):
If possible, can we leave the `terria.baseMaps` array untouched, and just hide them in https://github.com/TerriaJS/terriajs/blob/a430b8505370e13183db66e7122e1eb646a59622/lib/ReactViews/Map/Panels/SettingPanel.jsx#L320

nf-s:
> 
> 
> If possible, can we leave the `terria.baseMaps` array untouched, and just hide them in
> 
> https://github.com/TerriaJS/terriajs/blob/a430b8505370e13183db66e7122e1eb646a59622/lib/ReactViews/Map/Panels/SettingPanel.jsx#L320

Actually nah never mind, we can move it later if we every need to

nf-s (approved):

PR #5373, opened 2021-03-28
Api table catalog item

KeyboardSounds:
### What this PR does

Adds an ApiTableCatalogItem.

### Checklist


KeyboardSounds:
Here's a test catalog item:
```
{
      "name": "Pedestrian Traffic",
      "type": "api-table",
      "refreshInterval": 60,
      "removeDuplicateRows": true,
      "idKey": "sensor_id",
      "dateTimeKey": "date_time",
      "defaultStyle": {
        "time": {
          "idColumns": ["sensor_id"]
        }
      },
      "apis": [
        {
          "apiUrl": "https://data.melbourne.vic.gov.au/resource/d6mv-s43h.json",
          "keyToColumnMapping": [
            { 
              "columnName": "Total in both directions",
              "keyInApiResponse": "total_of_directions"
            },
            { 
              "columnName": "Time",
              "keyInApiResponse": "date_time"
            },
            { 
              "columnName": "id",
              "keyInApiResponse": "sensor_id"
            }
          ]
        },
        {
          "apiUrl": "https://data.melbourne.vic.gov.au/resource/h57g-5234.json",
          "kind": "PER_ID",
          "keyToColumnMapping": [
            { 
              "columnName": "latitude",
              "keyInApiResponse": "latitude"
            },
            { 
              "columnName": "longitude",
              "keyInApiResponse": "longitude"
            }
          ]
        }
      ]
    }
  ```

File: lib/Models/ApiTableCatalogItem.ts [hunk 1]
```diff
@@ -0,0 +1,197 @@
...
+
+          // Add per id data for this response
+          Object.assign(
+            mergedResponse,
+            perIdData.get(mergedResponse[this.idKey!])
+          );
+          return mergedResponse;
+        });
+      });
+    });
```
nf-s, line 99: Ideally there would be no observables in this before setting `apiResponses` - but we can't do this until `TableMixin` has been sorted (see https://github.com/TerriaJS/terriajs/issues/5392)

lib/Models/ApiTableCatalogItem.ts
```diff
@@ -0,0 +1,197 @@
...
+      .then(() => undefined);
+  }
+
+  refreshData(): void {
+    this.loadDataFromApis().then(() => {
+      runInAction(() => {
+        this.append(this.apiDataColumnMajor);
+      });
+    });
+  }
```
nf-s: As we talked about, this isn't proper usage of `@computed` to set `dataColumnMajor`, but until `TableMixin` allows catalog items to override `@computed dataColumnMajor` there isn't a good way to do this. (see https://github.com/TerriaJS/terriajs/issues/5392)

So - until `TableMixin` has been sorted - I think this should just call `this.apiResponseToTable()` instead of using a `@computed` here as `apiDataColumnMajor` will is going to be re-computed after every time `loadDataFromApis` is called

nf-s (changes requested):

lib/ModelMixins/TableMixin.ts
```diff
@@ -53,19 +53,59 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
...
+            (col, idx) =>
+              (newDataColumnMajor[idx] = col.filter(
+                cell => cell !== duplicateString
+              ))
+          );
+        }
+      }
+      this._dataColumnMajor = newDataColumnMajor;
+    }
+
```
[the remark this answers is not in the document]
nf-s: https://github.com/TerriaJS/terriajs/issues/5392

lib/ModelMixins/TableMixin.ts
```diff
@@ -53,19 +53,63 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
...
+          dataColumnMajor.forEach(
+            // Remove all the cells marked as duplicates
+            (col, idx) =>
+              (dataColumnMajor[idx] = col.filter(
+                cell => cell !== duplicateString
+              ))
+          );
+        }
+      }
+      return dataColumnMajor;
```
nf-s: I don't think you should have to clone `this._dataColumnMajor`, or change values in `dataColumnMajor`.

Maybe a better approach would be to have an array of row indices which need to be removed, and then return:

```ts
return this._dataColumnMajor.filter((row, idx) => rowIndicesToRemove.includes(idx))
```

nf-s: Also, I think it is important that if `removeDuplicateRows` is `false`, then return `this._dataColumnMajor` immediately.

Your current approach clones `this._dataColumnMajor` regardless of  what `removeDuplicateRows` is set to

nf-s (changes requested):

nf-s (changes requested):
Please add something about these to CHANGES:

- Fixes a bug where FeatureInfoDownload tries to serialize a circular object
- Added `removeDuplicateRows` `TableTrait`
- `forceLoadTableData` can now return undefined - which will leave `dataColumnMajor` unchanged

nf-s (approved):

PR #5399, opened 2021-04-08
Fix Cesium rendering crash when clipping rectangle is not within tiling scheme rectangle

### LINKED ISSUE #5396: Cesium rendering crashes when capabiltiies document returns a larger rectangle than what's supported by its tiling scheme
See e.g. ArcGIS server https://lsa2.geohub.sa.gov.au/server/rest/services/LSA/LocationSAViewerv7/MapServer/124.

Probably also applies to WMS.

This is because we don't ensure the following assumption is always true (written in comments in "terriajs-cesium/Source/Scene/ImageryLayer.js"):
> The imagery TilingScheme's rectangle always fully contains the ImageryProvider's rectangle.

In version 7 we created ImageryProviders, but didn't provide a `rectangle` option. We only added the clipping rectangle when constructing the ImageryLayer from the ImageryProviders. See:

#### v7 ArcGIS & WMS ImageryProvider creation:

https://github.com/TerriaJS/terriajs/blob/6dc8cc397d62de54b52c09bad42965f62db18d06/lib/Models/ArcGisMapServerCatalogItem.js#L369-L391

https://github.com/TerriaJS/terriajs/blob/6dc8cc397d62de54b52c09bad42965f62db18d06/lib/Models/WebMapServiceCatalogItem.js#L1487-L1503

#### v7 ImageryLayerCatalogItem passing rectangle from catalog items (incl. WMS & ArcGIS) onto Cesium/Leaflet

https://github.com/TerriaJS/terriajs/blob/6dc8cc397d62de54b52c09bad42965f62db18d06/lib/Models/ImageryLayerCatalogItem.js#L940-L949

#### v7 Cesium.js creating ImageryLayers

https://github.com/TerriaJS/terriajs/blob/6dc8cc397d62de54b52c09bad42965f62db18d06/lib/Models/Cesium.js#L1294-L1299

### Action

We should copy the version 7 method and only add `rectangle` on the ImageryLayer. This means adding `rectangle` and `clipToRectangle` to `ImageryParts` so it can be passed through to Cesium.ts/Leaflet.ts.

tephenavies:
### What this PR does

Fixes #5396.

* Add computed cesiumRectangle and change all other places that use item.rectangle to create a Cesium Rectangle to use new computed.
* Move clipping rectangle from ImageryProviders to ImageryLayer creation points to match terriajs v7 and fix Cesium render crash bug
* Update tests affected by these changes
* Fix some type errors in tests. I've started experimenting with throwing exceptions in tests for assertions that help with ts types in the remainder of the test.

(example for last point - changing:
```js
expect(long.expression.to.derive(condition)).toBeTruthy();
if (long.expression.to.derive(condition)) {
  // Remainder of test can now execute and types will recognise that condition is true
}
```

to:

```js
if (!long.expression.to.derive(condition)) throw new Error("Expected <explain condition meaning here>");
// remainder of test will also recognise condition is true
```
which has less repetition and as long as the exception message starts with "Expected" or similar it'll still be clear that it's an expectation failure of sorts)

### Checklist

-   [ ] No tests exist yet for ImageryLayer creating in Cesium.ts

### DIFF
--- a/lib/Models/Cesium.ts
+++ b/lib/Models/Cesium.ts
@@ -1,6 +1,6 @@
 import i18next from "i18next";
 import { autorun, computed, runInAction } from "mobx";
-import { createTransformer } from "mobx-utils";
+import { computedFn } from "mobx-utils";
 import BoundingSphere from "terriajs-cesium/Source/Core/BoundingSphere";
 import Cartesian2 from "terriajs-cesium/Source/Core/Cartesian2";
 import Cartesian3 from "terriajs-cesium/Source/Core/Cartesian3";
@@ -51,9 +51,6 @@ import CesiumRenderLoopPauser from "../Map/CesiumRenderLoopPauser";
 import CesiumSelectionIndicator from "../Map/CesiumSelectionIndicator";
 import MapboxVectorTileImageryProvider from "../Map/MapboxVectorTileImageryProvider";
 import PickedFeatures, { ProviderCoordsMap } from "../Map/PickedFeatures";
-import TileErrorHandlerMixin from "../ModelMixins/TileErrorHandlerMixin";
-import SplitterTraits from "../Traits/SplitterTraits";
-import TerriaViewer from "../ViewModels/TerriaViewer";
 import MappableMixin, {
   ImageryParts,
   isCesium3DTileset,
@@ -61,6 +58,9 @@ import MappableMixin, {
   isTerrainProvider,
   MapItem
 } from "../ModelMixins/MappableMixin";
+import TileErrorHandlerMixin from "../ModelMixins/TileErrorHandlerMixin";
+import SplitterTraits from "../Traits/SplitterTraits";
+import TerriaViewer from "../ViewModels/TerriaViewer";
 import CameraView from "./CameraView";
 import Feature from "./Feature";
 import GlobeOrMap from "./GlobeOrMap";
@@ -114,9 +114,12 @@ export default class Cesium extends GlobeOrMap {
   private readonly _disposeSplitterReaction: () => void;
 
   private _createImageryLayer: (
-    ip: ImageryProvider
-  ) => ImageryLayer = createTransformer((ip: ImageryProvider) => {
-    return new ImageryLayer(ip);
+    ip: ImageryProvider,
+    clippingRectangle: Rectangle | undefined
+  ) => ImageryLayer = computedFn((ip, clippingRectangle) => {
+    return new ImageryLayer(ip, {
+      rectangle: clippingRectangle
+    });
   });
 
   constructor(terriaViewer: TerriaViewer, container: string | HTMLElement) {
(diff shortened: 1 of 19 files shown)


tephenavies:
To test, see: http://ci.terria.io/fix-v8-rectangle/#clean&https://gist.githubusercontent.com/steve9164/809d0e6ce49b1e2b13cdb5250f15782e/raw/8ccd1510ed87316ce7d6ed6385169373666065ca/clipRect-v8.json

And next: http://ci.terria.io/next/#clean&https://gist.githubusercontent.com/steve9164/809d0e6ce49b1e2b13cdb5250f15782e/raw/8ccd1510ed87316ce7d6ed6385169373666065ca/clipRect-v8.json

The first of the 2 datasets should no longer cause a Cesium hang/crash when zooming in and out. It can be a bit hard to get it to crash in the existing `next` deployment. Add it before any other dataset and play around with zoom levels that can see all of Australia.

The second dataset is testing that clipping still works. It should work the same in next. It's clipped to: 
image

lib/Models/Cesium.ts
```diff
@@ -114,9 +114,12 @@ export default class Cesium extends GlobeOrMap {
   private readonly _disposeSplitterReaction: () => void;
 
   private _createImageryLayer: (
-    ip: ImageryProvider
-  ) => ImageryLayer = createTransformer((ip: ImageryProvider) => {
-    return new ImageryLayer(ip);
+    ip: ImageryProvider,
+    clippingRectangle: Rectangle | undefined
+  ) => ImageryLayer = computedFn((ip, clippingRectangle) => {
```
KeyboardSounds, line 119: Why the move from createTransformer to computedFunction? I'm not super confident that I understand the difference between the two, but I thought createTransformer memoized functions that were a pure function of their arguments, whereas a computedFn memoized functions that are a function of props and observable state.

lib/Models/Leaflet.ts
```diff
@@ -105,12 +105,16 @@ export default class Leaflet extends GlobeOrMap {
   private _disposeSplitterReaction: () => void;
 
   private _createImageryLayer: (
-    ip: ImageryProvider
-  ) => GridLayer = createTransformer((ip: ImageryProvider) => {
+    ip: ImageryProvider,
+    clippingRectangle: Rectangle | undefined
+  ) => GridLayer = computedFn((ip, clippingRectangle) => {
```
KeyboardSounds, line 110: See prev question about computedFn vs createTransformer

lib/Models/Cesium.ts
tephenavies: I am also not entirely sure what the difference is between them. One difference is that `computedFn` can take multiple arguments (as long as the number of arguments is always the same) and those arguments are allowed to be undefined.

`computedFn` is newer than `createTransformer` and the creator of mobx seems to recommend `computedFn` where it's possible to use it. The only thing I can think of that would make it not possible to use it is if you're using `createTransformer`'s destructor option, but I'm sure there are some other differences too.

KeyboardSounds (approved):

PR #5407, opened 2021-04-09
Handle errors while loading init sources

nf-s:
### Handle errors while loading init sources

- Add `title` parameter in `raiseErrorToUser` to overwrite error title.
- Added some error handling in `Terria.ts` to deal with loading init sources.
- Added error handling for selecting a basemap

### Checklist


lib/Models/Terria.ts
```diff
@@ -759,23 +760,73 @@ export default class Terria {
...
+          jsonValue = (await initSource).data;
+        }
+
+        if (jsonValue && isJsonObject(jsonValue)) {
+          return jsonValue;
+        }
+        return undefined;
+      }
+    );
+
```
nf-s, line 804: This has been moved here from outside `Terria` class - because I wanted access to instance variable for `raiseErrorToUser`

nf-s:
Also, I believe this will cause `loadInitSources` to be called twice:

https://github.com/TerriaJS/terriajs/blob/55d6681aa60b46496dc9d77950905dcb29dc3e80/lib/Models/Terria.ts#L730-L745

As `updateApplicationUrl()` is called in `start()`:

https://github.com/TerriaJS/terriajs/blob/55d6681aa60b46496dc9d77950905dcb29dc3e80/lib/Models/Terria.ts#L656

and the common pattern is to call this in `TerriaMap/index.js`

https://github.com/TerriaJS/TerriaMap/blob/48348b2ee4d5b68f3f01cd2d2d2af6da9fa26040/index.js#L66-L79

lib/Models/Terria.ts
```diff
@@ -740,7 +741,6 @@ export default class Terria {
         .query("")
         .hash("")
     );
-    await this.loadInitSources();
```
nf-s, line 743: This was causing `loadInitSources` to be called twice:

As `updateApplicationUrl()` is called in `start()`:

https://github.com/TerriaJS/terriajs/blob/55d6681aa60b46496dc9d77950905dcb29dc3e80/lib/Models/Terria.ts#L656

and the common pattern is to call this in `TerriaMap/index.js`

https://github.com/TerriaJS/TerriaMap/blob/48348b2ee4d5b68f3f01cd2d2d2af6da9fa26040/index.js#L66-L79

nf-s: To fix the case where `updateApplicationUrl` is called when the URL has changes, I have added `loadInitSources` to https://github.com/TerriaJS/terriajs/blob/28692dca384d39c8452966a349814080d5a3ed64/lib/ViewModels/updateApplicationOnHashChange.ts#L13-L26

tephenavies (approved):

PR #5426, opened 2021-04-14
Continuous color map

nf-s:
### Continuous color map

**Example http://ci.terria.io/continuous-color-map/#share=s-14cIDS8sGwd5gtkumJoU5C7sjt4**

https://github.com/TerriaJS/terriajs/issues/3441

* Moved `TableStyle.colorMap` into `TableColorMap`
* Replaced `colorbrewer.json` with `d3-scale-chromatic` - we now support d3 color scales (in addition to color brewer) -  see https://github.com/d3/d3-scale-chromatic
* Added `ContinuousColorMap` - it will now be used by default for `scalar` columns
  * To use `DiscreteColorMap` - you will need to set `numberOfBins` to something other than `0`.
* `TableColorMap` default color palette for `scalar` columns is not `Reds` instead of `RdYlOr`
* Legends for `scalar` columns will now calculate optimal `numberFormatOptions.maximumFractionDigits` and  `numberFormatOptions.minimumFractionDigits`

## Todo

- [x] Fix `Minimum and maximum values must be different.`

### Checklist


na9da:
@nf-s looks awesome! This is definitely a great enhancement.

Setting an invalid value for `colorPalette` results in app crash.

I'll update if I find any more problems.

na9da (approved):

PR #5428, opened 2021-04-15
Zoom status indicators

na9da:
### What this PR does

- [x] Adds a toast and zoom button spinner when map is zooming similar to the attached design.
- [x] Modifies the `zoomTo()` call to return a promise that resolves when the zoom animation is complete.
- [x] Implement slide in/out animation for the toast
- [x] Fixed toast positioning on mobile - shown 40px from bottom left.

image

### Testing

* Visit http://ci.terria.io/zoom-status-indicators/#share=s-hQ3zR006Fur8ObS3fZh1fiveM4V
* Click ideal zoom for the first catalog item. 
* The zoom button should show a spinner and a toast should appear at bottom center of the screen
* The zoom button will be disabled while the zoom is in progress
* Click ideal zoom for the second catalog item
* You should see the same results

### Checklist


lib/ReactViews/AnimatedSpinnerIcon.tsx
```diff
@@ -0,0 +1,34 @@
+import styled from "styled-components";
```
na9da, line 1: Extracted from `Loader.tsx`

lib/ReactViews/Workbench/Controls/ViewingControls.jsx
```diff
@@ -113,17 +120,22 @@ const ViewingControls = observer(
...
-      let zoomToView = item;
-      if (
-        item.rectangle !== undefined &&
-        item.rectangle.east - item.rectangle.west >= 360
-      ) {
-        zoomToView = this.props.viewState.terria.mainViewer.homeCamera;
-        console.log("Extent is wider than world so using homeCamera.");
-      }
-      viewer.zoomTo(zoomToView);
+      runInAction(() => {
```
na9da: Just wraps in `runInAction`

lib/ReactViews/AnimatedSpinnerIcon.tsx
zoran995: It might be better to move this to `lib/Styled`

lib/ReactViews/Loader.tsx
```diff
@@ -1,9 +1,8 @@
 import { TFunction } from "i18next";
```
zoran995, line 1: Also might be better to move this to `lib/Styled`

na9da: I'm a bit confused about that. How do we now decide what goes in `lib/Styled`?

zoran995: Not sure, but I see `Loader` as a component that is reused in combination with other components and doesn't stand alone. It just might be easier to find it later when needed.

tephenavies (approved):

PR #5431, opened 2021-04-18
Don't render help panel on init

### LINKED ISSUE #4670: Mobx: iOS safari is white screen
Repro:
1. visit http://ci.terria.io/next/
2. observe loading gif
3. observe loading gif fading out into a white screeen

zoran995:
### What this PR does

Currently we are rendering entire help panel at the start although it is not visible, and it might also fix #4670. Profiler is reporting significant increase of the initial rendering time. 
before this change
image

after this change
image

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - small UI change

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Change Log
 ### MobX Development
 
 #### next release (8.0.0-alpha.77)
-
+* Stopped rendering help panel when it is closed.
 * [The next improvement]
 
 
--- a/lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
+++ b/lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
@@ -454,7 +454,9 @@ const StandardUserInterface = observer(
                   animationDuration={animationDuration}
                 />
               )}
-            <HelpPanel terria={terria} viewState={this.props.viewState} />
+            {this.props.viewState.showHelpMenu && (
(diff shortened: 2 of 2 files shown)


tephenavies (dismissed):

tephenavies (changes requested):
I like the idea but this disables slide-in, slide-out animations. We should improve this component, but I'm not sure this is the way to do that.

Another idea would be to render only a white box while animating (that should be much cheaper than rendering the whole help panel) or maybe only render the first 320px column, not the HelpVideoPanel too while animating. What do you think @philipgrimmett @wibily ?

tephenavies:
BTW you can test this PR as-is at http://ci.terria.io/zoran-help-panel/

wibily:
I don't think its that big of an issue that we break animations to save on initial rendering times as we don't have a consistent animation pattern within the app. For example, the story panel isn't animated either and the workbench sliding animation could also do with some work. My opinion is that we conditionally render the HelpPanel and remove the css transition properly so that we can deal with sliding panels holistically at a later date

na9da:
Sorry I didn't see this PR when I made a fix for the problem here #5807. It should be fixed while retaining the animation. Good to know that it also increases render performance. Thanks @zoran995.

zoran995:
Closing this as it is resolved in #5807

PR #5442, opened 2021-04-22
Add translations

wibily:
### What this PR does

Adds language translation toggle to enable switching between languages

### Checklist


lib/ReactViews/Mobile/MobileMenuItem.jsx
```diff
@@ -34,7 +37,9 @@ export default function MobileMenuItem(props) {
 MobileMenuItem.propTypes = {
   href: PropTypes.string,
   onClick: PropTypes.func,
-  caption: PropTypes.string
+  caption: PropTypes.string,
+  showMobileIcon: PropTypes.bool,
```
nf-s: Is `showMobileIcon` used anywhere?

nf-s (changes requested):
This isn't super high priority - but if it isn't too much work - it would be nice if the wrapping was a bit better on small screens (before transition to mobile screen):

## With language switcher 

image

## Without language switcher

image

lib/ReactViews/Mobile/MobileMenuItem.jsx
```diff
@@ -34,7 +37,9 @@ export default function MobileMenuItem(props) {
...
+  icon: PropTypes.string
```
nf-s: I think this should be `PropTypes.object` as you are passing in an object here

https://github.com/TerriaJS/terriajs/blob/d1515b43b0c24350717699c0ac056a3d0d6f9484/lib/ReactViews/Map/Panels/LangPanel/LangPanel.tsx#L31-L43

nf-s: Which gets passed in here
https://github.com/TerriaJS/terriajs/blob/d1515b43b0c24350717699c0ac056a3d0d6f9484/lib/ReactViews/Map/Panels/MobilePanel.jsx#L27-L31

lib/ReactViews/Map/Panels/LangPanel/LangPanel.tsx
```diff
@@ -0,0 +1,59 @@
...
+              <RawButton onClick={() => i18n.changeLanguage(key)}>
+                {value}
+              </RawButton>
+            </Li>
+          ))}
+        </Ul>
+      </Box>
+    </MenuPanel>
+  );
+};
```
nf-s, line 59: I'm getting this error, but not sure what it's about 
image

lib/ReactViews/Mobile/MobileMenuItem.jsx
wibily: ah, good catch

lib/ReactViews/Mobile/MobileMenuItem.jsx
wibily: sorry i don't understand, are you suggesting that MobileMenuItem should have a single prop with the caption and the icon as keys?

lib/ReactViews/Map/Panels/LangPanel/LangPanel.tsx
wibily: will investigate

lib/ReactViews/Mobile/MobileMenuItem.jsx
nf-s: Sorry I meant `icon` should be `PropTypes.object` - as you are passing in an `object` through:
- `MenuPanel` `mobileIcon` prop -> `MobileMenuItem` `icon` prop

I was getting an error in console about the prop being an `object`, even though it's prop type is `string`

lib/ReactViews/Map/Panels/LangPanel/LangPanel.tsx
wibily: hmm, i couldn't reproduce this one

nf-s: i can't either!
it has disappeared

nf-s (approved):

PR #5448, opened 2021-04-25
TSify feedbackForm and introduce Checkbox component

zoran995:
### What this PR does

Converted the FeedbackForm to typescript with better accessibility. 
- each input field has its associated label so a screen reader can properly read text.
- tested checkbox component
- added option for comment minLength so the administrators can easily activate it if needed.

Added checkbox component
- built with accessibility in mind (this is the reason why hidden `<input type="checkbox"/>` is rendered).
- with multiline text, a checkbox is aligned with the first line of text, can easily add centered property if needed (maybe @philipgrimmett would like to share his opinion on this from UI/UX side).
- possibility to have a fully controlled or uncontrolled component.

Few places should utilize the checkbox component, but it will be pretty straightforward to update them when this is accepted.

### Checklist


lib/Models/Terria.ts
```diff
@@ -375,7 +385,9 @@ export default class Terria {
     languageConfiguration: undefined,
     customRequestSchedulerLimits: undefined,
     persistViewerMode: true,
-    openAddData: false
+    openAddData: false,
+    feedbackPreamble: "feedback.feedbackPreamble",
+    feedbackMinLength: 0
```
wibily, line 398: i'm not sure if this needs to be a configuration option as this would allow users to submit a feedback form with no feedback?

package.json
```diff
@@ -160,8 +160,13 @@
     "@babel/eslint-parser": "^7.12.16",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@types/dateformat": "^3.0.1",
+    "@types/enzyme": "^3.10.8",
+    "@types/enzyme-adapter-react-16": "^1.0.6",
+    "@types/jasmine-enzyme": "^7.0.1",
     "babel-plugin-styled-components": "^1.10.7",
     "documentation": "12.3.0",
+    "enzyme": "^3.11.0",
```
wibily, line 169: nice! thanks for setting this up for us

lib/ReactViews/Feedback/FeedbackForm.tsx
```diff
@@ -0,0 +1,454 @@
...
+      styledHeight={styledMinHeight}
+      styledMinHeight={styledMinHeight}
+      styledMaxHeight={styledMaxHeight}
+      onChange={onChangeHandler}
+      invalidValue={!valueIsValid}
+    ></StyledTextArea>
+  );
+};
+
+const StyledTextArea = styled(StyledInput).attrs({
```
wibily, line 365: this should probably be in its own file

lib/ReactViews/Feedback/FeedbackForm.tsx
```diff
@@ -0,0 +1,454 @@
...
+  changeSendShareUrl(e: React.ChangeEvent<HTMLInputElement>) {
+    this.setState((prevState: IState) => ({
+      sendShareURL: !prevState.sendShareURL
+    }));
+  }
+
+  onSubmit(e: React.FormEvent<HTMLFormElement | HTMLDivElement>) {
+    e.preventDefault();
+
+    if (
```
wibily, line 138: instead of detecting this onSubmit, we could instead set the minLength attribute on the textarea. Since the textarea has a name attribute, this will trigger the browser's form validation and we also get a nice warning message on why the submit failed.

wibily (approved):

lib/Models/Terria.ts
zoran995: Till now, it worked without any limitation (user can send completely empty feedback with no info), so I left it like that by default

PR #5455, opened 2021-04-27
Add `ignoreErrors` url parameter

nf-s:
### Add `ignoreErrors` url parameter

Add `ignoreErrors` url parameter

### Checklist


tephenavies:
Here's a test: http://ci.terria.io/ignore-error-flag/#share=s-u4au2PVppKSEYgMx2HdQuojrEpb&ignoreErrors=1
Remove ignoreErrors=1 and it throws an error because the shared catalog item is in this catalog: https://gist.githubusercontent.com/steve9164/807aaebe2eacf8c5e0eb76e2aee917db/raw/e8d99d55a3cdd5d134dbd0d846fde961bd51a03d/test-ignoreErrors.json

tephenavies (approved):

PR #5473, opened 2021-05-06
Change behaviour of Disclaimer to only render deny button if there is a denyAction

tephenavies:
### What this PR does

For https://github.com/TerriaJS/drought-map/issues/235.

Rather than show a useless cancel button this PR now hides that button unless it does something.

In most cases with sites derived from v8 TerriaMap the denyAction is a redirect to a site, by default https://terria.io. Maybe there should be a follow up PR to remove that default and default to no cancel button when a custom site is not provided. @nf-s what do you think? 

### Checklist

-   [x] Small UI change, no tests


nf-s (approved):

nf-s:
yep looks good
and I agree, default to no `denyAction`
Having a cancel button which takes you to terria.io doesn't seem useful

PR #5478, opened 2021-05-07
Wc weblate

wibily:
### What this PR does

records down the decision to use weblate for our translation needs

### Checklist

- no code changes
- purely documentation changes ADR so no updates to changes.md


architecture/0007-use-weblate-for-i18n.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Context
+
+We have a request to provide a french translationed version of de-africa. We are using the i18next and i18next-react libraries with all our UI strings within en/translation.json. There are also two long life branches master & next representing our current production release (v7) and the next production release (v8). Due to the time lag between releases, there are hundreds more strings in the v7 translation.json versus the master version.
+
+It would not be workable or scalable for translators to provide these translations into the github repository directly as this requires a level of technical expertise and we won't have an easy way of keeping track of which strings haven't been translated yet. We also require some quality check mechanism to ensure the quality of the contributed translations.
+
+## Decision
+
+We will use hosted weblate at https://hosted.weblate.org/projects/terriajs/ to allow external translators to contribute to the project. There will be two projects, one connected to the master branch and another connected to the next branch.
```
tephenavies, line 17: I'd like to see some extra details here about what we intend to do when the next branch becomes master. Should we:
* Try to port translations from v7 to v8?
* Provide instructions for contributors to do this?
* Deprecate/terminate the v7 translation project on some schedule without a way to port the translations?

Should we continue making the v7 weblate commit back to a `7.x.x` branch of terriajs and release it every `n` months? I don't have a good answer to any of these questions.

architecture/0007-use-weblate-for-i18n.md
```diff
@@ -0,0 +1,27 @@
...
+
+When the en/translation.json is updated, our weblate plugin and associated webhook will let the hosted weblate know and weblate will pull the updated changes. When a contributing translator provides translations, weblate will create a PR into the corresponding branch allowing a terria mainter to review the changes before merging the changes.
```
tephenavies: Would pull requests with whole translation files for other localisations fit into this approach e.g. https://github.com/TerriaJS/terriajs/pull/5241 & https://github.com/TerriaJS/terriajs/pull/5302?

I'm not saying it should. Maybe Weblate has something that could help us get these efforts into the project through the Weblate workflow below.

tephenavies (dismissed):
This is great @wibily. Thanks so much for researching all of this and setting it up. Here's a few things that I picked up on.

architecture/0007-use-weblate-for-i18n.md
zoran995: I see that weblate has a mechanism to review translations. I am not sure who is in charge of reviewing, but I wonder when we get entire terriajs translated to French and approve those translations. When some source string changes, will it set its translation to unreviewed or need editing status?

zoran995:
In the context of using weblate and translating to french, we should address strings from #4519 (with or without lint support), since some strings are still not extracted for translation.
Also, weblate supports pluralization of string, and #5113 should be addressed to support pluralization of all strings that need it (at least it was like that when I created that PR). This will show a nice translation form 
image

architecture/0007-use-weblate-for-i18n.md
wibily: @steve9164 New whole translation files won't be an issue as it'll get picked up by the github webhook and then it'll follow the existing weblate process. If it was an existing translation file however, this might cause merge conflicts between our respository and the weblate repository.
@zoran995 - I believe that if the source string changes, it will change the status of the translated string to 'needs editing'.

zoran995: Just checked above with weblate support to be sure. When source strings update, their state will change automatically to "Needs editing", and if we are using reviewing new review will be required. Can we add that somewhere to adr, and also, should reviewing be activated?

wibily: Keep in mind that we will need additional work to support RTL languages currently

wibily: I don't think we should enable reviewing at this stage, maybe once we have identified someone who could do it? Thoughts? @steve9164

architecture/0007-use-weblate-for-i18n.md
wibily: That's a tricky one because currently there is also a possibility of duplication if two different translators contribute translations for the same string but in different branches. I think for the v8 release, we should try to port translations from v7 to v8 and then going forward only have one weblate project to avoid confusion. I'll update the ADR to reflect this

tephenavies (approved):

PR #5481, opened 2021-05-10
fix CSW stratum name bug

nf-s:
### Checklist

### DIFF
--- a/lib/Models/CswCatalogGroup.ts
+++ b/lib/Models/CswCatalogGroup.ts
@@ -711,15 +711,17 @@ export default class CswCatalogGroup extends UrlMixin(
   }
 
   protected async forceLoadMetadata(): Promise<void> {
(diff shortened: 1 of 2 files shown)


lib/Models/CswCatalogGroup.ts
```diff
@@ -711,15 +711,17 @@ export default class CswCatalogGroup extends UrlMixin(
...
   }
 
   protected async forceLoadMembers(): Promise<void> {
-    const cswStratum = <CswStratum | undefined>this.strata.get(CswStratum.name);
+    const cswStratum = <CswStratum | undefined>(
+      this.strata.get(CswStratum.stratumName)
+    );
     if (cswStratum) {
       await runLater(() => cswStratum.createMembersFromLayers());
     }
```
nf-s, line 727: my bad....

KeyboardSounds (approved):

PR #5487, opened 2021-05-12
Leaflet hack to fix white gap between tiles

nf-s:
### Hack to fix "Space between tiles on fractional zoom levels in Webkit browsers" 

https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-688644225

### After vs Before

image

### Checklist


lib/Sass/global/_leaflet.scss
```diff
@@ -65,6 +65,12 @@
     max-width: none !important;
   }
 
+  // Hack to fix "Space between tiles on fractional zoom levels in Webkit browsers" (https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-688644225)
+  .leaflet-tile-container img {
+    width: 256.5px !important;
+    height: 256.5px !important;
+  }
+
```
nf-s: Currently we only support 256x256 tiles in leaflet, so this should be fine 🤷

zoran995: Love this, those white lines always annoyed me; there is also one more nice workaround https://github.com/ngageoint/mapcache-electron/commit/53dcfa608f477d1ccd088320cf76de7b8ce8bc5f

nf-s: Thanks, that is nicer than the CSS approach. 
I ended up doing something similar but with the `tileloadstart` event - so I don't upset our Leaflet types

https://github.com/TerriaJS/terriajs/blob/16f82fce81c84dc8b39a863aa21167d46b7ecb11/lib/Map/CesiumTileLayer.ts#L84-L88

na9da (approved):
Looks great!

zoran995:
@nf-s It seems that it actually should be `+ 1px`, when it is `0.5px` I can still see the gaps

image

PR #5501, opened 2021-05-21
weblate merge master

tephenavies:
Merge from weblate.

Sorry if this interferes with something you've been doing @wibily. I was wondering what the conflict between our repo and weblate was and then once I'd resolved that I thought I might as well PR the resolution (conflict was weblate using 4 spaces for translation files. I'm not sure if it just detected the spaces we use incorrectly or if it'll continue to do that).


wibily (approved):
I've set up the two space add-on inside weblate and can see that new weblate PRs have the correct spacing. Either way, I'm happy to approve since the worst case scenario would be that we have the run the formatter again.

PR #5502, opened 2021-05-21
Fix calculation of minimum level for WMTS

tephenavies:
### What this PR does

Part of https://github.com/TerriaJS/terriajs/issues/5503.

It can be merged independently of the change proposed in #5503, but alone it doesn't result in a corrected visualisation.

### Checklist


zoran995:
fixed in #7371

PR #5512, opened 2021-05-25
Disable pedestrian mode in mobile

### LINKED ISSUE #5511: Disable pedestrian mode in mobile

na9da:
### What this PR does

Fixes #5511 

Disables pedestrian mode in mobile.

### Testing
* Go to http://ci.terria.io/next/
* Turn on mobile view from the dev panel in your browser
* You will see the pedestrian mode button 
image
* Now visit http://ci.terria.io/disable-pedestrian-mode-in-mobile/
* Switch to mobile view
* You shouldn't see the pedestrian mode button

### Checklist

### DIFF
--- a/lib/ReactViews/Map/MapNavigation.jsx
+++ b/lib/ReactViews/Map/MapNavigation.jsx
@@ -141,6 +141,7 @@ class MapNavigation extends React.Component {
               <If
                 condition={
                   !this.props.terria.configParameters.disablePedestrianMode &&
+                  !this.props.viewState.useSmallScreenInterface &&
                   this.props.terria.currentViewer instanceof Cesium
                 }
               >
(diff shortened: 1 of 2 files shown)


lib/ReactViews/Map/MapNavigation.jsx
```diff
@@ -141,6 +141,7 @@ class MapNavigation extends React.Component {
               <If
                 condition={
                   !this.props.terria.configParameters.disablePedestrianMode &&
+                  !this.props.viewState.useSmallScreenInterface &&
```
wibily, line 144: this may cause a problem if a user starts pedestrian mode and then resizes their screen. Perhaps another check to see whether they are already in pedestrian mode?

wibily (approved):

lib/ReactViews/Map/MapNavigation.jsx
na9da: Good point! If a `ToolButton` unmounts it [closes the tool automatically](https://github.com/TerriaJS/terriajs/blob/6296fccd92983b5f0154dd47061dd438b7941ed1/lib/ReactViews/Tool.tsx#L85) so the pedestrian mode will exit correctly when switching to mobile view.

PR #5514, opened 2021-05-26
ModelDimensions

nf-s:
### `ModelDimensions`

This provides ability to set model JSON through SelectableDimensions (a dropdown).

### Example

This adds a dropdown to set `colorPalette` to `Reds` or `Blues`.

It does this by applying the options `value` object as model stratum:

eg 
```json
{
  "defaultStyle": {
    "color": {
       "colorPalette": "Reds"
    }
  }
}
```

http://ci.terria.io/poc-model-dimensions/#clean&https://gist.githubusercontent.com/nf-s/83699e3e90bca7a75fb550580e16ad90/raw/7760fe47447b6e68b69b833f1ec64a4d406d50a7/model-dim.json

```json
{
      "type": "csv",
      "url": "test/NSW_LGA_NEXIS_201212.csv",
      "name": "NSW LGA NEXIS 2012",
      "modelDimensions": [
        {
          "id": "Cols",
          "selectedId": "Red",
          "options": [
            {
              "id": "Red",
              "value": {
                "defaultStyle": {
                  "color": {
                    "colorPalette": "Reds"
                  }
                }
              }
            },
            {
              "id": "Blue",
              "value": {
                "defaultStyle": {
                  "color": {
                    "colorPalette": "Blues"
                  }
                }
              }
            }
          ]
        }
      ]
    }
```

### Checklist


lib/Traits/CatalogMemberTraits.ts
```diff
@@ -205,4 +206,13 @@ export default class CatalogMemberTraits extends ModelTraits {
...
   dataCustodian?: string;
+
+  @objectArrayTrait({
+    type: DimensionTraits,
+    idProperty: "id",
+    name: "Model dimensions",
+    description:
+      "This provides ability to set model JSON through SelectableDimensions (a dropdown)."
+  })
+  modelDimensions?: DimensionTraits[];
```
nf-s, line 217: .

lib/Traits/objectTrait.ts
```diff
@@ -75,6 +75,7 @@ export class ObjectTrait<T extends ModelTraits> extends Trait {
             message: `${propertyName} is not a valid sub-property of ${this.id}.`
           })
         );
+        return;
```
nf-s, line 78: This was supposed to be here, as we need to stop if `trait` is `undefined`

lib/Traits/ShadowTraits.ts
```diff
@@ -1,9 +1,9 @@
+import CatalogMemberTraits from "./CatalogMemberTraits";
 import primitiveTrait from "./primitiveTrait";
-import ModelTraits from "./ModelTraits";
 
 export type Shadows = "CAST" | "RECEIVE" | "BOTH" | "NONE";
 
-export default class ShadowTraits extends ModelTraits {
+export default class ShadowTraits extends CatalogMemberTraits {
```
KeyboardSounds: This seems kind of lazy to me. Something with ShadowTraits is not inherently a catalog member. It might be one, in fact it almost certainly is, but it's a coincidence rather than something that should be codified into the inheritance tree, IMO.

KeyboardSounds: Can you split out the traits from CatalogMemberTraits that you want to use here and share it across CatalogMember and ShadowTraits?

nf-s: I agree, I don't really like it. But I have done it this way as I need to access `super.selectableDimensions` in `ShadowMixin` to merge them:

https://github.com/TerriaJS/terriajs/blob/e5546b6cca4a3bb8200c1095c3487d1e5cff13ca/lib/ModelMixins/ShadowMixin.ts#L33-L53

We don't currently have a mechanism to merge arrays across Mixins (similar to how we deal with arrays across strata). So there isn't really a nice way to merge all of the `selectableDimensions` across all of the different Mixins.

nf-s: Actually I'll just rename `ShadowMixin.selectableDimensions` to `ShadowMixin.shadowDimensions` or something and then merge them upstream

nf-s: Done

KeyboardSounds (approved):

PR #5517, opened 2021-05-26
API support in GeoJSON catalog item

KeyboardSounds:
### What this PR does

GeoJSON Mixin based catalog items can now call an API to retrieve their data as well as fetching it from a url.

This requires changes to loadJson and loadJsonBlob to POST a request body rather than always make a GET request.

### Checklist


lib/Traits/ApiRequestTraits.ts
```diff
@@ -0,0 +1,59 @@
...
+  queryParameters: QueryParamTraits[] = [];
+
+  @objectArrayTrait({
+    name: "Query parameters for updates",
+    type: QueryParamTraits,
+    description:
+      "Query parameters to supply to the API on subsequent calls after the first call.",
+    idProperty: "name"
+  })
+  updateQueryParameters: QueryParamTraits[] = [];
```
nf-s, line 44: These have only been implemented in `ApiTableCatalogItem` - not `GeoJsonCatalogItem` - so I think they should be moved into `ApiTableRequestTraits`

nf-s: `QueryParamTraits` should probs also be moved into `lib/Traits/ApiTableCatalogItemTraits.ts`

nf-s (changes requested):

lib/Traits/ApiRequestTraits.ts
```diff
@@ -0,0 +1,59 @@
...
+      "The value of the query parameter. Parameter values starting with" +
+      " `DATE!`, eg. `DATE!HH:MM`, will be replaced wih the current date and" +
+      " time, formatted according to the format string following the `!`." +
+      " For more information on the format string format, see the " +
+      " [dateformat](https://github.com/felixge/node-dateformat) library."
+  })
+  value?: string;
+}
+
+export default class ApiRequestTraits extends UrlTraits {
```
nf-s, line 28: With this change, you've added `forceProxy` and `cacheDuration` - but they aren't implemented in `ApiTableCatalogItem`

nf-s (changes requested):

lib/Models/saveModelToJson.ts
```diff
@@ -11,9 +11,13 @@ export default function saveModelToJson(
...
-  const includeStrata = options.includeStrata
-    ? options.includeStrata
-    : Array.from(model.strata.keys());
+  if (options.includeStrata === undefined && model.strata === undefined) {
+    // There aren't any strata. Because of some dodgy typing, this can happen for anyTraits.
+    // We stringify and then parse to remove stray mobx properties, etc.
+    return JSON.parse(JSON.stringify(model));
+  }
+  const includeStrata =
+    options.includeStrata ?? Array.from(model.strata.keys());
```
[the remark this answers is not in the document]
KeyboardSounds: Yeah I really wasn't sure about this one. AnyTraits end up getting the type Model<T> but yeah, they're not stratified or reactive.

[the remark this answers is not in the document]
nf-s: Where does `AnyTrait` get the type `Model`?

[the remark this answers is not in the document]
KeyboardSounds: I don't know where it's getting the type _from_, but on hover I get:
image

lib/Traits/ApiRequestTraits.ts
KeyboardSounds: Lots of things that mixin UrlTraits don't implement either of those, eg. MapboxVectorTileCatalogItem, Cesium3DTilesCatalogItem. I thought the principle here was that just because a value is set, doesn't necessarily mean that something is done with that value.

lib/Models/saveModelToJson.ts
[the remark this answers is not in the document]
nf-s: I think the problem is in here

https://github.com/TerriaJS/terriajs/blob/2b482eba0802e7f5f2100aeaa0d485187478aeed/lib/Models/ModelPropertiesFromTraits.ts#L6-L34

It doesn't like when traits have type `any` - to fix this you can change `requestData?: any` to `requestData?: JsonObject`. I can't think of a reason why any trait should have type `any` over `JsonObject`

lib/Traits/ApiRequestTraits.ts
nf-s: I don't think a Model should have a trait which isn't implemented. Maybe we should split up `UrlTraits` into `UrlTraits` and `UrlProxyTraits` or something

nf-s (approved):

lib/Traits/ApiRequestTraits.ts
nf-s: This is a later problem

lib/Traits/ApiRequestTraits.ts
nf-s: This is also a later problem

PR #5518, opened 2021-05-27
Wc about french

wibily:
### What this PR does
This PR is to support detection of language changes in client maps by using the useTranslation() hook.

### Checklist


tephenavies (approved):
Looks good. I can't see why there was a problem before, but it's much better to use `useTranslation` & `withTranslation` than import `i18next` anywhere (and especially in React component files).

PR #5529, opened 2021-06-08
Compare

na9da:
### What this PR does

Implements new compare workflow.

### Other changes
- Adds `ignoreSplitterWhenPicking` option to MapInteractionMode. This is needed so that when user picks a filter location on the left or right and only one side has visible imagery, the interaction should still resolve the location correctly. https://github.com/TerriaJS/terriajs/pull/5529/commits/cbf8317a7ab6446bb674726cfd6741ec35a09849
- Adds WorkflowPanel used for implementing workflows outside the workbench. https://github.com/TerriaJS/terriajs/blob/compare/lib/Styled/WorkflowPanel.tsx
- Activates/deactivates terria timeline stack when the `Timeline.jsx` component mounts or unmounts.
- Adds an `index` parameter to `Workbench.add()` method specifying the index where the item should be added - https://github.com/TerriaJS/terriajs/blob/4f925bcea1acfad4a12af63a1f1b2600550ffed5/lib/Models/Workbench.ts#L171

### Description
- [x] Implement the basic workflow based on the new designs.
- [x] Feature gate the workflow so that the old workflow is used by default and the new workflow is shown only when a feature flag is turned on
- [x] Disable location filter when interacting with the splitter or other Compare UI elements.
- [x] Just found a bug where the location filter does not get set the first time.
- [x] Use `#useExperimentalCompareWorkflow` hash parameter to enable this feature.
- [ ] Write tests
- [ ] Date picker popup should drop shadow to create some contrast when using lighter base maps.

### Testing
1. Open this [CI link](http://ci.terria.io/compare/#share=s-dbkoEJG7aNdPrbTrXcBPEJreGc6&configUrl=https://gist.githubusercontent.com/na9da/fc6d95a548fb352338becffb0dd9b0c9/raw/cbb17ea72c01f25c324ab0752b970b3d5ec09045/useExperimentalCompareWorkflow.json).
2. Play with the workflow and try to break things

### Checklist


lib/Models/TimelineStack.ts
```diff
@@ -74,7 +76,7 @@ export default class TimelineStack {
     });
   }
 
-  destroy() {
+  deactivate() {
     if (this._disposeClockAutorun) {
```
na9da, line 80: `TimelineStack` has a reaction that keeps the time of workbench items in sync with the timeline clock. Previously this reaction was created inside the constructor, so it kept running even when the `Timeline.jsx` component is unmounted/hidden. As a result when the user sets a date for an item, the reaction will override it with the current clock time. I have moved the reaction to a an `activate` method which gets called in the [componentDidMount](https://github.com/TerriaJS/terriajs/pull/5529/files#diff-327ef8da4defdcbc3243f9ab7a41901cc5664e80d0c94d40c4a6256dc3dff379R41-R47) method of `Timeline.jsx` and created a `deactivate` method that gets called in unmount hook. So now, if the `Timeline.jsx` is not shown, the user can set dates on item without it being reset by this reaction.

lib/Models/Workbench.ts
```diff
@@ -167,14 +166,15 @@ export default class Workbench {
    *
    * @param item The item to add to or remove from the workbench.
    */
-  public async add(item: BaseModel | BaseModel[]): Promise<void> {
+  public async add(
+    item: BaseModel | BaseModel[],
+    index: number = 0
+  ): Promise<void> {
```
na9da: Added index parameter to specify the position in the workbench where we want to insert the item.

lib/Models/Workbench.ts
```diff
@@ -167,14 +166,15 @@ export default class Workbench {
...
     if (Array.isArray(item)) {
-      await Promise.all(item.reverse().map(i => this.add(i)));
+      await Promise.all(item.reverse().map(i => this.add(i, index)));
       return;
     }
 
-    this.insertItem(item);
-
```
na9da: Moved this call to later, because if the item is a reference, we need to load it first.

lib/ReactViews/BottomDock/BottomDock.jsx
```diff
@@ -77,4 +78,4 @@ const BottomDock = observer(
   })
 );
 
-module.exports = measureElement(BottomDock, false);
+module.exports = withControlledVisibility(measureElement(BottomDock, false));
```
na9da: So that we can show/hide the `BottomDock`

lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx
```diff
@@ -16,9 +16,9 @@ import {
   ObjectifiedYears
```
na9da, line 16: Changes to make the DateTimePicker more customizable... Ideally, when we have more time, we should rewrite this component.

lib/Models/Cesium.ts
```diff
@@ -983,7 +983,7 @@ export default class Cesium extends GlobeOrMap {
    * specified and set terria.pickedFeatures based on this.
    *
    */
-  pickFromScreenPosition(screenPosition: Cartesian2, ignoreSplitter: boolean) {
+  pickFromScreenPosition(screenPosition: Cartesian2, ignoreSplitter?: boolean) {
```
na9da, line 986: Changes to read `ignoreSplitter` setting from MapInteractionMode.

lib/ReactViews/DragDropFile.tsx
```diff
@@ -105,7 +105,7 @@ class DragDropFile extends React.Component<PropsType> {
           [Styles.isActive]: this.props.viewState.isDraggingDroppingFile
         })}
       >
-        {isDefined(this.props.viewState.isDraggingDroppingFile) ? (
+        {this.props.viewState.isDraggingDroppingFile ? (
```
na9da: This resulted in an invisible div below the main app all the time and the browser sometimes scrolls to reveal a white blank space. When `this.props.viewState.isDraggingDroppingFile` is true, the div is correctly positioned at the center of the screen.

lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
```diff
@@ -247,7 +248,7 @@ const StandardUserInterface = observer(
         !this.shouldUseMobileInterface();
       const showStoryPanel =
         this.props.terria.configParameters.storyEnabled &&
-        this.props.terria.stories.length &&
+        this.props.terria.stories.length > 0 &&
```
na9da: This check caused `showStoryPanel` to be 0 if the length is 0 and when used in JSX for conditional rendering it printed a stray 0 in the dom.

lib/ReactViews/Tools/DiffTool/DiffTool.tsx
```diff
@@ -1015,20 +1016,6 @@ function removeSplitItem(item: DiffableItem) {
...
-  const imageryProvider = feature.imageryLayer?.imageryProvider;
-  if (imageryProvider === undefined) return false;
-  return (
-    item.mapItems.find(
-      m => ImageryParts.is(m) && m.imageryProvider === imageryProvider
-    ) !== undefined
-  );
-}
-
 function setTimeFilterFromLocation(
```
na9da, line 1030: Moved to a separate file in `lib/Models`

lib/Models/Terria.ts
```diff
@@ -1638,6 +1659,8 @@ async function interpretHash(
             message: { key: "parsingStartDataErrorMessage" }
           });
         }
+      } else if (property === "useExperimentalCompareWorkflow") {
+        terria.configParameters.useExperimentalCompareWorkflow = true;
```
wibily: should this be inside a else if block? Ie could it be possible that we'd want to set useExperimentalCompareWorkflow in combination with any of the other properties in the hash?

wibily: ah sharing seems to use #start, so when sharing from using #useExperimentalCompareWorkflow, it reverts back to to the original compare

lib/Models/Workbench.ts
```diff
@@ -167,14 +166,15 @@ export default class Workbench {
...
    * @param item The item to add to or remove from the workbench.
    */
-  public async add(item: BaseModel | BaseModel[]): Promise<void> {
+  public async add(
+    item: BaseModel | BaseModel[],
+    index: number = 0
+  ): Promise<void> {
     if (Array.isArray(item)) {
-      await Promise.all(item.reverse().map(i => this.add(i)));
+      await Promise.all(item.reverse().map(i => this.add(i, index)));
```
wibily: wait why do we need to reverse the item array?

lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx
```diff
@@ -250,6 +275,15 @@ class DateTimePicker extends React.Component<PropsType> {
     window.removeEventListener("click", this.closePickerEventHandler);
   }
 
+  componentDidUpdate() {
+    // Scroll into view to reveal the selected time item
```
wibily, line 279: this will fire whenever any of the props for the component change. It might be better to abstract out the <Grid className="time-grid"> JSX into its own component and then use an effect to conditionally scroll the time into view. This was a very nice touch though :)

lib/Models/Workbench.ts
zoran995: If I recall correctly, it is done so the item gets placed in the correct position in the workbench. 
When an array is supplied to the `add` function items should be added in the same order so the first array item gets at the top of the workbench (or desired index), without reverse that order would be messed up

lib/Models/doesImageryFeatureBelongToCatalogItem.ts
```diff
@@ -0,0 +1,20 @@
...
+  feature: Feature,
+  item: BaseModel
+): Boolean {
+  if (!MappableMixin.isMixedInto(item)) return false;
+  const imageryProvider = feature.imageryLayer?.imageryProvider;
+  if (imageryProvider === undefined) return false;
+  return (
+    item.mapItems.find(
+      m => ImageryParts.is(m) && m.imageryProvider === imageryProvider
+    ) !== undefined
```
wibily: could use Array.includes here

wibily (approved):

lib/Models/Terria.ts
na9da: The change looks fine to me. The if/else statements are inside a `forEach` loop, so it will correctly process all hash paremeters. The generated share URL currently doesn't capture the flag, which is alright I think.

lib/Models/doesImageryFeatureBelongToCatalogItem.ts
na9da: `Array.includes` only takes a concrete value, no matcher function, so I don't think we can use it here.

wibily: ahh you are right, I was thinking of `Array.some` which takes in a matcher function

lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx
na9da: Good point! Extracting the JSX would be a bigger change I would hold off until a grand rewrite of the component :) . For now I have added a check with prevProps for `currentDate`.

lib/Models/doesImageryFeatureBelongToCatalogItem.ts
na9da: Ah yes, that works. I have made the change :+1:

tephenavies (approved):

lib/Styled/Checkbox/Elements/CheckboxIcon.tsx
```diff
@@ -6,9 +6,6 @@ import { CheckboxIconProps } from "../types";
 const StyledCheckboxIcon = styled(StyledIcon).attrs({
   styledWidth: "1em"
 })<{ disabled?: boolean }>`
-  top: 0.125em;
-  align-self: flex-start;
-  position: relative;
```
zoran995: this is added with an intention to force a checkbox aligned with a top row with multi-row label.

lib/ReactViews/Compare/ItemList.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <div
+            css={`
+              flex-grow: 1;
+            `}
+          >
+            <Checkbox
+              isChecked={item.show}
+              onChange={ev => onChangeSelection(item, ev.target.checked)}
+              label={<SelectorText medium>{item.name}</SelectorText>}
+            />
```
zoran995: Font size should be set on the top checkbox element and text should inherit it that way we get text and checkbox icon to have the same size and they are aligned vertically.

lib/Styled/Checkbox/Elements/CheckboxIcon.tsx
zoran995: I tried to avoid this in feedback component 
image

lib/ReactViews/Compare/ItemList.tsx
zoran995: There are some changes to how the checkbox work in #5826 which enables us to easier work with text sizing.

zoran995 (changes requested):
Hi @na9da I have just a small suggestion regarding the checkbox component

lib/ReactViews/Compare/ItemList.tsx
zoran995: just created a #5919 with only checkbox changes from #5826

zoran995:
@na9da We completely remove the existing compare tool from the navigation menu? I am not a fan of doing that, it will be really hard to communicate that to our users. Is there a chance to leave it but make it hidden by default, at first I understood that both options will be available and that this will be marked as experimental?

AnaBelgun:
hi @zoran995 the changes are following testing we've done with users which were very confused by the current workflow. Understandably, this is a major and we'd like to test it again before it's merged into main. We'll definitely keep you in the loop. and we'll talk about options.

zoran995:
I am afraid that it will confuse user where is option. Some of things that might be useful in new workflow :
- Leave compare button to at least open new workflow (it will be much easier for user to find it and transition) 
- collapse option for left right dataset (big legends issue) 
- reorder of datasets

Also we lose option to put more dataset to one side and use opacity (ie street and addresses for two time periods) 

p.s. I know that this is still in development but better to mention it in earlier

sidenote from above, closing and opening compare again cause something like this (compare is at 50% but the dataset is visible as it was in last compare)
image

na9da:
@zoran995 thanks for pointing out other use cases. This PR is still a WIP and I hope we will be able to address all the concerns. I'll add back the experimental flag if we merge it anytime soon.

PR #5468, opened 2021-05-04
Revert "add tests"

nf-s:
### What this PR does

This reverts commit 668ee565004766b64184cd2941bbd53e05068ebb.

This added tests for https://github.com/TerriaJS/terriajs/pull/5448 - which included adding devDependency for `enzyme`:

https://github.com/TerriaJS/terriajs/blob/7efefad863f1e94c25288c481b874b368b52171d/package.json#L169-L170

This resulted in clashes in `domhandler` versions, which upsets yarn:

image

image

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [x] I've updated CHANGES.md with what I changed.

PR #5532, opened 2021-06-09
Upgrade html-to-react dependency and fix type errors.

na9da:
### What this PR does

Upgrades `html-to-react` to latest version and fix the type errors. This is to fix the CI build errors.

### Checklist


lib/ReactViews/Custom/ChartCustomComponent.ts
```diff
@@ -338,7 +338,8 @@ export default abstract class ChartCustomComponent<
    * column contains a `<chart>`?
    * @param node The node to test
    */
-  private isFirstColumnOfChartRow(node: DomElement): boolean {
+  private isFirstColumnOfChartRow(_node: DomElement): boolean {
+    const node = _node as any;
```
na9da: Casting to `any` here and below as I think it is not worth the effort typing this properly. The nested typing is gnarly!

package.json
```diff
@@ -85,14 +84,13 @@
     "d3-transition": "^1.0.0",
     "d3-zoom": "^1.8.3",
     "dateformat": "^3.0.3",
-    "domhandler": "^2.4.2",
```
tephenavies, line 88: I think we shouldn't import "domhandler" without having a direct dependency on "domhandler".

tephenavies (changes requested):

zoran995:
@na9da I am afraid that this could cause the same issue as #5468 although I might be missing something, check the https://github.com/TerriaJS/terriajs/pull/5468#issuecomment-832607495

na9da:
Thanks @zoran995 for pointing that out. Considering this and also @steve9164's comment, I think we should define the type locally and use that instead of depending on domhandler. We are currently importing it only for the type definition. Upgrading `html-to-react` itself is not a problem though.

na9da:
Sorry that, I overwrote the previous changes.

tephenavies (approved):

PR #5534, opened 2021-06-10
More geojson

KeyboardSounds:
### What this PR does

Adds basic per feature styling and time support to GeoJsonMixin-based catalog items.

### Checklist


lib/Traits/GeoJsonTraits.ts
```diff
@@ -1,17 +1,46 @@
...
 import StyleTraits from "./StyleTraits";
 import UrlTraits from "./UrlTraits";
 
+export class PerPropertyGeoJsonStyleTraits extends ModelTraits {
+  @anyTrait({
+    name: "Properties",
+    description:
+      "If the properties of a feature match these properties, then apply the style to that feature"
+  })
+  properties?: any;
```
nf-s: This should be `JsonObject` - see https://github.com/TerriaJS/terriajs/pull/5517#discussion_r643606957

nf-s (changes requested):
Also, it is possible that multiple `perPropertyStyles` match to each feature. At the moment only the first `perPropertyStyle` is actually applied. Maybe not worth dealing with at this point in time though.

http://localhost:3001/#clean&https://gist.githubusercontent.com/nf-s/7e29fb1a7b12e02ce629ecbfa348ef91/raw/3fcbd2584f96f183d4353305c30121c29a4d0686/geojson-style-test-catalog.json

KeyboardSounds:
> At the moment only the first perPropertyStyle is actually applied. Maybe not worth dealing with at this point in time though.
I've fixed this! Thanks for spotting it

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -150,6 +159,51 @@ export default function GeoJsonMixin<
...
+                return featKey === styleKey && featValue === styleValue;
+              }) !== undefined
+          );
+        });
+
+        if (matchedStyles !== undefined) {
+          for (let matched of matchedStyles) {
+            for (let trait of Object.keys(matched.style.traits)) {
+              // @ts-ignore - TS can't tell that `trait` is of the correct index type for style
+              featureProperties[trait] = matched.style[trait];
```
nf-s: This will overwrite values with `undefined`

Whereas this would keep existing values:

```suggestion
              const styleValue = matched.style[trait];
              if (isDefined(styleValue)) featureProperties[trait] = styleValue;
```

nf-s (changes requested):

lib/ModelMixins/GeojsonMixin.ts
KeyboardSounds: Fixed! But the `@ts-ignore` is still necessary, typescript still complains

nf-s (approved):

PR #5536, opened 2021-06-10
Make GyroscopeGuidance ready for i18n

ykiu:
### What this PR does

I noticed some strings in the GyroscopeGuidance component are hardcoded, blocking translation. This patch moves them out of the source code to the i18n message registry.

### Checklist

    - Unit tests aren't applicable because this patch just adds some strings to translation.json
    - I haven't updated CHANGES.md because this patch seems too small to be listed there. I would update it if that's required, however.


tephenavies (approved):
Thanks for the contribution!

tephenavies (changes requested):
Actually, could you add a CHANGES.md entry and add yourself to CONTRIBUTORS.md?

ykiu:
OK, edited both files.

tephenavies (approved):

ykiu:
@steve9164 Thanks!

PR #5538, opened 2021-06-11
Fix bug `selectableDimensions` bug in `Cesium3dTilesMixin`.

nf-s:
### Checklist

### DIFF
--- a/lib/ModelMixins/Cesium3dTilesMixin.ts
+++ b/lib/ModelMixins/Cesium3dTilesMixin.ts
@@ -249,7 +249,7 @@ export default function Cesium3dTilesMixin<
     }
 
     @computed get selectableDimensions(): SelectableDimension[] {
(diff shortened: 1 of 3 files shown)


lib/ModelMixins/Cesium3dTilesMixin.ts
```diff
@@ -249,7 +249,7 @@ export default function Cesium3dTilesMixin<
     }
 
     @computed get selectableDimensions(): SelectableDimension[] {
-      return [...this.selectableDimensions, this.shadowDimension];
+      return [...super.selectableDimensions, this.shadowDimension];
     }
```
nf-s, line 256: Fix circular reference

tephenavies (approved):

KeyboardSounds (changes requested):
No no wait don't merge this yet you need to fix gltf catalog item too

KeyboardSounds (approved):

PR #5549, opened 2021-06-15
Upgrade to react to v17 and more

zoran995:
### What this PR does

Upgrade to react17, there should be no breaking changes, changes are mostly related to optimizations. 
Also, removed unused dependencies
- `class-list`
- `d3-axis`
- `d3-dispatch`
- `simple-statistics`
- `ts-loader`

Still testing if everything works correctly. Need to check if an upgrade of `react-virtual` breaks something in `search inside items`, @na9da do you maybe recall of specific reason to lock the version. _edit_: `useRef` should be used instead of `createRef` in functional components. Everything works correctly after changing that.

- [x] Upade react imports
- [x] Fix issue with react-virtual upgrade

### Checklist


zoran995:
regarding
```
error TS2686: 'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.
```
using react 17 named imports can't work properly with typescript until it's upgraded to 4.1+. Currently, typescript upgrade is blocked (see #5219). 
Using `import React from "react"` won't be supported in future versions of `React` (https://twitter.com/dan_abramov/status/1308736035321511939). 
What we can do, is use `import * as React from "react"`, or as a temp solution, we can revert commits around upgrade of imports and go just with a version upgrade, and when we upgrade typescript then we can update imports. @steve9164 any thought on this?

tephenavies:
Sorry for the slow reply. I like the `import * as React from "react"` solution if that will work in future React versions.

PR #5550, opened 2021-06-15
Checkbox fix click event

zoran995:
### What this PR does

Checkbox component has wrong placement of label element which results in non-clickable checkbox icon. 

### Checklist


steve9164:
I think this leads to invalid HTML. <label> cannot have <div> nested inside. At least `<Spacing>` would have to be replaced with its `<span>` equivalent (which I think is `<SpacingSpan>`). I haven't looked at the other components to tell if they need to be changed too. Also there should be only 1 control inside the `<label>` (though that should already be correct)

Other than that I think this should work well.

tephenavies (changes requested):
I think this leads to invalid HTML. <label> cannot have <div> nested inside. At least `<Spacing>` would have to be replaced with its `<span>` equivalent (which I think is `<SpacingSpan>`). I haven't looked at the other components to tell if they need to be changed too. Also there should be only 1 control inside the `<label>` (though that should already be correct)

Other than that I think this should work well.

zoran995:
nice catch, fixed. Also updated id so it's not only number

tephenavies:
Waiting on test in #5588 and then I'll merge this and close that PR.

PR #5552, opened 2021-06-18
Wc de aus ga

wibily:
### What this PR does

Adds tracking for more user interactions

### Checklist


na9da:
Looks great @wibily!

I would recommend two good to have changes:

1. I think we should extract all the `category` tags into a typescript Enum called `AnalyticsCatageory` or something and use that as the first parameter of the `logEvent` call. That will help reduce spelling errors and make it easier to figure out what all categories are in use.
2. Maybe we should define the [Analytics](https://github.com/TerriaJS/terriajs/blob/wc-de-aus-ga/lib/Models/Terria.ts#L273) type as `{ logEvent: (category: AnalyticsCategory, action: string, ...): void }` and remove the `?` from `terria.analytics?.logEvent` in all the call sites. I don't think there is a chance for it to be `undefined` as it gets defined in the `Terria` [constructor](https://github.com/TerriaJS/terriajs/blob/5cb18157128572d3a3d10f50d010566806402721/lib/Models/Terria.ts#L502-L504).

wibily:
@na9da for 1) yup, i'll clean up the TerriaJS events and refactor all the logEvent calls to use an enum. I can't change the type as downstream maps also log events
2) If `option.analytics` is falsy in config, then that entire block won't run and `this.analytics` will be undefined. I really want to use something like https://gcanti.github.io/fp-ts/modules/Option.ts.html but that's a discussion for another time  :)

wibily:
changes made, @na9da could you please approve if you are happy with the changes?

na9da (approved):

PR #5553, opened 2021-06-18
Make clickable moment points optional.

na9da:
### What this PR does

Partial fix for https://github.com/TerriaJS/de-australia-map/issues/94

Right now the moment points expects an `onClick` handler and throws a console error if not set. This change makes it optional. 

### Testing
1. Visit this [next branch link](http://ci.terria.io/next/#share=s-100uNrxDtubTYVMIWovjoukHywV)
2. Observe, when hovering the dot points, they appear clickable and throws a console error when clicking.
3. Now, visit [this link](http://ci.terria.io/non-clickable-moment-points/#share=s-oab4cshjiUvcvgfzIUEvsipmJXP).
4. Observe, when hovering the dot points, they are not clickable and also the fill color changes on hover.

### Checklist


lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
```diff
@@ -62,26 +63,42 @@ class MomentPointsChart extends React.Component {
...
     const baseKey = `moment-point-${chartItem.categoryName}-${chartItem.name}`;
     const fillColor = chartItem.getColor();
+    const isClickable = chartItem.onClick !== undefined;
+    const clickProps = point => {
+      if (isClickable) {
+        return {
+          pointerEvents: "all",
+          cursor: "pointer",
+          onClick: () => chartItem.onClick(point)
+        };
```
na9da, line 73: Populate these props only if the point is clickable (i.e `onClick` handler is set)

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
```diff
@@ -62,26 +63,42 @@ class MomentPointsChart extends React.Component {
...
   }
 }
 
+const StyledCircle = styled(Circle)`
+  &:hover {
+    opacity: 1;
+    fill: white;
+  }
+`;
+
```
na9da, line 101: Adds a hover effect on the point (Circle).

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
wibily: why not just go with chartItem?.onClick(point)? Up to you if you want to go this way or your current way

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
wibily: up to you but do we need this hover effect if the item is not clickable?

wibily (approved):

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
na9da: Yes, I think it makes it easier for the user to identify the point on which the mouse is hovering.

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
na9da: I re-wrote it this way because there are a bunch of props I want to set depending on whether the item is clickable or not.

PR #5556, opened 2021-06-18
Switch from Travis CI to GitHub Actions

peterhassall:
### What this PR does

Fixes devops#47

This PR switches deployment of TerriaMap to ci.terria.io from Travis CI to GitHub Actions, and updates the GitHub authentication method to fix devops#45

### Checklist


wibily:
hmm is the url in the ci-cleanup.js script still valid? (https://api.github.com/repos/' + repo + '/branches';)
As far as i can see, shouldn't it be (/repos/{owner}/branches}?
https://docs.github.com/en/github-ae@latest/rest/reference/repos#list-branches

One dodgy way to check might be view that secret and then curl that url to see if it works :/

wibily (approved):

tephenavies:
~`$GITHUB_REPOSITORY` is `owner/repo` so it should work.~ Whoops misread, but in the case of ci-cleanup.js the `repo` variable is also in the form `owner/repo` so it should continue to work under GHA.

tephenavies:
@peterhassall Are there any interoperability problems if this gets merged and some branches continue to be deployed and call the cleanup script on Travis? I can't see any (other than maybe helm version problems, though we should have sorted that when we upgraded the Travis scripts to helm v3).

tephenavies (approved):

peterhassall:
> @peterhassall Are there any interoperability problems if this gets merged and some branches continue to be deployed and call the cleanup script on Travis?

It _should_ be fine. Both methods create the ingress resources after checking the active branches, so they shouldn't interfere with each other.

PR #5564, opened 2021-06-22
Fixed issue when there were multiple features for the same date in GeoJsonMixin

KeyboardSounds:
### What this PR does

Tiny bugfix to make time-based geojson work when there are multiple features with the same time property value.

### Checklist


nf-s (changes requested):
I'm pretty sure that `discreteTimesAsSortedJulianDates` should be unique

nf-s:
https://github.com/TerriaJS/terriajs/blob/ae210977724d5e14bc371d5f73ba28cb52e22f46/lib/ModelMixins/TableMixin.ts#L555-L559

KeyboardSounds:
> I'm pretty sure that discreteTimesAsSortedJulianDates should be unique

@nf-s le done

lib/ModelMixins/DiscretelyTimeVaryingMixin.ts
```diff
@@ -96,21 +96,24 @@ function DiscretelyTimeVaryingMixin<
...
+            asJulianMap.set(asJulian.tag, asJulian);
           }
         } catch {}
       }
-      asJulian.sort((a, b) => JulianDate.compare(a.time, b.time));
-      return asJulian;
+
+      const asJulianArray = Array.from(asJulianMap.values());
+      asJulianArray.sort((a, b) => JulianDate.compare(a.time, b.time));
+      return asJulianArray;
```
nf-s: I'd rather not have this change in `DiscretelyTimeVaryingMixin` as it could have performance implications. 
I think `discreteTimes` should be implemented to be unique in https://github.com/TerriaJS/terriajs/blob/4f9a5f84aac1b73d0a29f125fcea10ee1f39a319/lib/ModelMixins/GeojsonMixin.ts#L518

nf-s (changes requested):

lib/ModelMixins/DiscretelyTimeVaryingMixin.ts
KeyboardSounds: What do you mean by performance implications? For things that use `discreteTimesAsSortedJulianDates`? This is why I originally had it as a separate getter. I don't want to put it in GeoJsonMixin because otherwise everything that wants no duplicates (which is most things!) will have to reimplement it, so we might as well put it in DiscretelyTimeVaryingMixin.

nf-s: I'm saying that `discreteTimes` should only ever return unique dates - which is the case for every other implementation of `discreteTimes`. 
So with this change, it is de-duplicating twice - which I think may cause some performance issues - as we can have datasest with millions of dates

KeyboardSounds: OH
OH sorry I have no idea what I was thinking thank you for being patient with my nonsense
I think I've fixed it now

nf-s (approved):
I just added a commit to fix `CHANGES.md` - as I think I added some duplicate lines in a previous PR

PR #5568, opened 2021-06-23
GeoJSON CZML template

nf-s:
### What this PR does

CZML template to be used to replace each GeoJSON Point feature. Feature coordinates and properties will automatically be applied to CZML packet, so they can be used as references. If this is defined, `clampToGround`, `style`, `perPropertyStyles`, `timeProperty` and `heightProperty` will be ignored.

For example - this will render a green cylinder for every point (and use the length and radius feature properties)

http://ci.terria.io/geojson-czml-template/#clean&https://gist.githubusercontent.com/nf-s/327a06724f1f00aaafa2052036ed9243/raw/8779ecbcc18b29c827736084be7e54692d5a6cf0/czml-geojson.json

```json
{
      "name": "geojson-czml-test",
      "url": "build/TerriaJS/test/GeoJSON/points.geojson",
      "type": "geojson",
      "czmlTemplate": {
        "cylinder": {
          "length": {
            "reference": "#properties.height"
          },
          "topRadius": {
            "reference": "#properties.radius"
          },
          "bottomRadius": {
            "reference": "#properties.radius"
          },
          "material": {
            "solidColor": {
              "color": {
                "rgba": [
                  0,
                  200,
                  0,
                  255
                ]
              }
            }
          }
        }
      }
    },
```

image

### Checklist

-   [X] There are unit tests to verify my changes are correct

### DIFF
--- a/lib/ModelMixins/GeojsonMixin.ts
+++ b/lib/ModelMixins/GeojsonMixin.ts
@@ -1,7 +1,8 @@
-import { Feature, FeatureCollection, GeoJsonObject } from "geojson";
+import { Feature, FeatureCollection, GeoJsonObject, Point } from "geojson";
 import i18next from "i18next";
-import { action, computed, observable, runInAction } from "mobx";
+import { action, computed, observable, runInAction, toJS } from "mobx";
 import Cartesian3 from "terriajs-cesium/Source/Core/Cartesian3";
+import clone from "terriajs-cesium/Source/Core/clone";
 import Color from "terriajs-cesium/Source/Core/Color";
 import defaultValue from "terriajs-cesium/Source/Core/defaultValue";
 import DeveloperError from "terriajs-cesium/Source/Core/DeveloperError";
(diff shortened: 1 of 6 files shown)


lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -112,51 +114,51 @@ export default function GeoJsonMixin<
...
+          this._dataSource = await this.loadCzmlDataSource(geoJsonWgs84);
+        } else {
+          this._dataSource = await this.loadGeoJsonDataSource(geoJsonWgs84);
+        }
+      } catch (e) {
+        throw TerriaError.from(e, {
+          title: i18next.t("models.geoJson.errorLoadingTitle"),
+          message: i18next.t("models.geoJson.errorParsingMessage")
         });
+      }
```
nf-s: Mostly turning into async/await.

This is the only change - on line 151-155

```ts
        if (isDefined(this.czmlTemplate)) {
          this._dataSource = await this.loadCzmlDataSource(geoJsonWgs84);
        } else {
          this._dataSource = await this.loadGeoJsonDataSource(geoJsonWgs84);
        }
```

lib/Models/Cesium.ts
```diff
@@ -1179,7 +1179,7 @@ export default class Cesium extends GlobeOrMap {
       // Try to find catalogItem for picked feature, and use catalogItem.getFeaturesFromPickResult() if it exists - this is used by FeatureInfoMixin
       const catalogItem = picked?.primitive?._catalogItem ?? id?._catalogItem;
 
-      if (typeof catalogItem.getFeaturesFromPickResult === "function") {
+      if (typeof catalogItem?.getFeaturesFromPickResult === "function") {
```
nf-s, line 1182: `catalogItem` can be undefined here

na9da (approved):
Looks great @nf-s!

Something to think about in the future: If the geojson has geometries other than points, maybe we could create both czml & geojson datasources with the geojson datasource containing the other (non-point) geometries (?)

PR #5575, opened 2021-06-25
Disable `TableMixin` time column if only one unique time interval

nf-s:
### Disable `TableMixin` time column if only one unique time interval

Fixes https://github.com/TerriaJS/terriajs/issues/5571

#### Example

Now timeline shouldn't show - and "Expand" + download button shouldn't be in feature info section

http://ci.terria.io/table-time-single-interval/#share=s-9krZqFCp53Vkjo2U40Xp8LQVgwF

vs

http://ci.terria.io/next/#share=s-9krZqFCp53Vkjo2U40Xp8LQVgwF

### Checklist


lib/Models/OpenDataSoftCatalogItem.ts
```diff
@@ -500,13 +500,6 @@ export class OpenDataSoftDatasetStratum extends LoadableStratum(
...
   }
 
-  /** Disable date time selector if there is only 1 sample per point */
-  @computed get disableDateTimeSelector() {
-    return isDefined(this.maxPointSamples) && this.maxPointSamples === 1
-      ? true
-      : undefined;
-  }
-
   /** Get fields with useful infomation (for visualisation). Eg numbers, text, not IDs, not region... */
```
nf-s, line 503: This is been replaced by disableDateTimeSelector in lib/Table/TableAutomaticStylesStratum.ts

KeyboardSounds (approved):

PR #5578, opened 2021-06-29
Make attribution & disclaimer link text customizable.

na9da:
### What this PR does

Makes the attribution & disclaimer links in the cesium map credit section customizable.

### Checklist

### DIFF
--- a/lib/Language/en/translation.json
+++ b/lib/Language/en/translation.json
@@ -789,6 +789,16 @@
       "csvRegionMappingMessageLoadError": "Couldn't load region boundaries for region {{regionName}} {{exception}}"
     }
   },
+  "cesiumExtraCredits": {
+    "attribution": {
+      "link": "/about.html#data-attribution",
+      "text": "Data attribution"
+    },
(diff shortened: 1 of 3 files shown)


lib/Language/en/translation.json
```diff
@@ -789,6 +789,16 @@
       "csvRegionMappingMessageLoadError": "Couldn't load region boundaries for region {{regionName}} {{exception}}"
     }
   },
+  "cesiumExtraCredits": {
```
wibily, line 792: hmm, i don't think the actual link should be in the translation.json as this won't change if its a different language. Keep in mind that external translators contribute to this file as well through weblate

wibily: i reckon keep the link hardcoded if it doesn't have to change now but if it does, perhaps a better place would be inside config.json

na9da: Yeah, it was sort of a quick fix. I was holding back from adding more new config parameters. I guess I'll redo the PR adding config parameters and making it less hackier.

PR #5581, opened 2021-06-29
Fix `TableMixin` region only (no useful `colorColumn`)

nf-s:
### Fix `TableMixin` region only (no useful `colorColumn`)

There was an issue where CSVs which only had region mapping would show no data - even if `colorColumn` is set to a region column.

```json
{
      "type": "csv",
      "name": "region color test",
      "defaultStyle": {
        "color": {
          "colorColumn": "LGA_code",
          "regionColor": "orange"
        }
      },
      "url": "dcp-lgas.csv"
    },
```

### Checklist


tephenavies (approved):

tephenavies (changes requested):
This still doesn't act as before. Try https://data.gov.au/data/dataset/76327768-7e34-4464-b4a2-6bba32876d32/resource/de4745ba-44b5-496a-a1ca-ded413deb508/download/dcp-lgas.csv on http://ci.terria.io/fix-region-csv/. You should only get the regions that are in the dataset, not all regions coloured.

lib/Table/TableStyle.ts
```diff
@@ -471,29 +468,52 @@ export default class TableStyle {
...
+        title: this.tableModel.name,
+        nullColor
+      });
     }
   }
 
+  @computed get defaultColor() {
+    return Color.fromCssColorString(defaultColor);
+  }
+
```
tephenavies: This shouldn't be a computed because it doesn't use anything observable. Does it even need to be in this class?

nf-s: Sorry I forgot to remove that - done!

lib/Table/TableStyle.ts
```diff
@@ -471,19 +468,35 @@ export default class TableStyle {
...
+        color = this.regionColor;
+      } else if (colorTraits.nullColor) {
         color = Color.fromCssColorString(colorTraits.nullColor);
-      } else if (this.binColors.length > 0) {
-        color = this.binColors[0];
+      } else if (colorTraits.binColors && colorTraits.binColors.length > 0) {
+        color = Color.fromCssColorString(colorTraits.binColors[0]);
       } else if (colorId) {
         color = Color.fromCssColorString(getColorForId(colorId));
       }
```
nf-s, line 484: I changed this.binColors to `colorTraits.binColors` as I think we should only use binColors here if they are defined manually (not the default binColors).

This means that `Color.fromCssColorString(getColorForId(colorId))` is used more - which makes it easier to see different datasets

tephenavies:
It still doesn't behave well when adding https://data.gov.au/data/dataset/76327768-7e34-4464-b4a2-6bba32876d32/resource/de4745ba-44b5-496a-a1ca-ded413deb508/download/dcp-lgas.csv via web data. I'd expect only the regions in the file to be coloured (either in 1 colour, or in many colours I don't really mind)

nf-s:
> 
> 
> It still doesn't behave well when adding https://data.gov.au/data/dataset/76327768-7e34-4464-b4a2-6bba32876d32/resource/de4745ba-44b5-496a-a1ca-ded413deb508/download/dcp-lgas.csv via web data. I'd expect only the regions in the file to be coloured (either in 1 colour, or in many colours I don't really mind)

Fixed

tephenavies:
Thanks for fixing this up. It works with my examples now.

tephenavies (approved):

PR #5582, opened 2021-06-30
Configurable credit links

na9da:
### What this PR does

Replaces the hardcoded credit links with links configurable via config parameters.

This is required especially for maps that have about pages that are not located at the hardcoded locations.

### Testing

* Visit the [ci link](http://ci.terria.io/custom-credit-links/#configUrl=https://gist.githubusercontent.com/na9da/6c5c76313faae147c2589de3069fe046/raw/0be24f0807aad0852094835147aab321373b9e16/custom-credit-links.json).
* Check out the 3 links at the bottom credit line that configured via the config parameters.

### Checklist

### DIFF
--- a/lib/Models/Cesium.ts
+++ b/lib/Models/Cesium.ts
@@ -243,14 +243,16 @@ export default class Cesium extends GlobeOrMap {
           ?.reverse()
           .forEach(({ url, text }) => {
             // Create a link and insert it after the logo node
+            // Defaults to the given text if no translation is provided
+            const translatedText = i18next.t(text);
             const el = document.createElement("div");
-            el.innerHTML = `<a href="${url}" target="_blank" rel="noopener noreferrer">${text}</a>`;
+            el.innerHTML = `<a href="${url}" target="_blank" rel="noopener noreferrer">${translatedText}</a>`;
             const linkNode = el.firstChild as Element | null;
             if (linkNode) {
               logoContainer?.insertAdjacentElement("afterend", linkNode);
             }
           });
-        expandLink.innerText = "Basemap";
+        expandLink.innerText = i18next.t("map.extraCreditLinks.basemap");
       }
     }
 
--- a/lib/Models/Cesium.ts
+++ b/lib/Models/Cesium.ts
@@ -245,12 +245,12 @@ export default class Cesium extends GlobeOrMap {
             // Create a link and insert it after the logo node
             // Defaults to the given text if no translation is provided
             const translatedText = i18next.t(text);
-            const el = document.createElement("div");
-            el.innerHTML = `<a href="${url}" target="_blank" rel="noopener noreferrer">${translatedText}</a>`;
-            const linkNode = el.firstChild as Element | null;
-            if (linkNode) {
-              logoContainer?.insertAdjacentElement("afterend", linkNode);
-            }
+            const a = document.createElement("a");
+            a.href = url;
+            a.target = "_blank";
+            a.rel = "noopener noreferrer";
+            a.innerText = translatedText;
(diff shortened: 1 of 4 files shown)


lib/Models/Cesium.ts
```diff
@@ -239,27 +239,17 @@ export default class Cesium extends GlobeOrMap {
...
-          if (attributionToAboutPage && attributionToAboutPage.firstChild) {
-            logoContainer.parentNode.insertBefore(
-              attributionToAboutPage.firstChild,
-              logoContainer.nextSibling
-            );
-          }
-        }
-
+        this.terria.configParameters.extraCreditLinks
+          ?.reverse()
```
wibily: this will mutate the extraCreditLinks array, is this the behaviour you want or do you simply want to make a reversed copy?

wibily (commented):
Just for my understanding, how come we are reaching into the dom and adding elements outside of React?

lib/Models/Cesium.ts
```diff
@@ -239,27 +239,17 @@ export default class Cesium extends GlobeOrMap {
...
+          .forEach(({ url, text }) => {
+            // Create a link and insert it after the logo node
+            const el = document.createElement("div");
+            el.innerHTML = `<a href="${url}" target="_blank" rel="noopener noreferrer">${text}</a>`;
+            const linkNode = el.firstChild as Element | null;
```
wibily: instead of doing this, it might be easier to create an 'a' element and appendChild to your el element, that way you'll know for sure it exists and can get rid of the null checks

lib/Models/Terria.ts
```diff
@@ -416,7 +421,12 @@ export default class Terria {
     persistViewerMode: true,
     openAddData: false,
     feedbackPreamble: "feedback.feedbackPreamble",
-    feedbackMinLength: 0
+    feedbackMinLength: 0,
+    extraCreditLinks: [
```
wibily, line 425: would this leave them out of the translation.json flow? It might be better to have the extraCreditLinks be an object so that use the object key to refer to the url and have that same key be referenced in translation.json

wibily:
i can't believe you rick rolled me...well played

wibily (approved):

lib/Models/Cesium.ts
na9da: > Just for my understanding, how come we are reaching into the dom and adding elements outside of React?

You are right. I think the reason why it was done that way originally is that we are dealing with DOM elements created by Cesium. We could still separate it out as a renderless component. I have created an issue for it here - https://github.com/TerriaJS/terriajs/issues/5584

na9da:
> Just for my understanding, how come we are reaching into the dom and adding elements outside of React?

You are right. I think the reason why it was done that way originally is that we are dealing with DOM elements created by Cesium. We could still separate it out as a renderless component. I have created an issue for it here - https://github.com/TerriaJS/terriajs/issues/5584

lib/Models/Terria.ts
na9da: Good point about translating the text. I have changed the default link text to i18n keys. However I am keeping it as an array as it makes the ordering more explicit (although JS object key ordering is in the lexical order).

lib/Models/Cesium.ts
na9da: Makes sense. I think we don't need the `el` if we create `a` directly (but you loose the convenience of innerHTML). I have updated the code to do so.

lib/Models/Cesium.ts
na9da: Good catch! That would have been a bug.

PR #5587, opened 2021-06-30
Fixes a bug with numeric item search where it sometimes fails to return all matching values.

### LINKED ISSUE #5562: Search: Indexed search does not return all matching results for some numeric queries
When searching an indexed dataset, if we provide the same min & max value for a numeric attribute, then only one result is returned even though there are multiple valid matches.

This is probably happening because, `startIndex` & `endIndex` in the below code is the same index position. To fix the problem we need to iterate `startIndex` backwards and iterate `endIndex` forwards until we reach a non matching value.

https://github.com/TerriaJS/terriajs/blob/80e6c3e26ac17e9ae83d53576e3d3b188f124805/lib/Models/ItemSearchProviders/NumericIndex.ts#L64-L81

Link to reproduce issue: TBD.

na9da:
### What this PR does

Fixes #5562 

Basically if a numeric index has multiple repeating values, currently the search might not return all the matching results.

### Checklist

### DIFF
--- a/lib/Models/ItemSearchProviders/NumericIndex.ts
+++ b/lib/Models/ItemSearchProviders/NumericIndex.ts
@@ -21,7 +21,7 @@ type NumericSearchQuery = {
 export default class NumericIndex implements IndexBase<NumericSearchQuery> {
   readonly type = IndexType.numeric;
 
-  private idValuePairs?: Promise<{ dataRowId: number; value: number }[]>;
+  idValuePairs?: Promise<{ dataRowId: number; value: number }[]>;
 
   /**
    * Constructs a NumericIndex.
@@ -73,8 +73,12 @@ export default class NumericIndex implements IndexBase<NumericSearchQuery> {
       { value: endValue },
       (a, b) => a.value - b.value
     );
-    const startIndex = i < 0 ? ~i : i;
-    const endIndex = j < 0 ? ~j : j;
+    let startIndex = i < 0 ? ~i : i;
+    let endIndex = j < 0 ? ~j : j;
+    // iterate startIndex backward till we find the first matching value
+    while (idValuePairs[startIndex - 1]?.value === value.start) startIndex -= 1;
+    // iterate endIndex forward till we find the last matching value
+    while (idValuePairs[endIndex + 1]?.value === value.end) endIndex += 1;
     const matchingIds = idValuePairs
       .slice(startIndex, endIndex + 1)
       .map(({ dataRowId }) => dataRowId);
--- a/lib/Models/ItemSearchProviders/NumericIndex.ts
+++ b/lib/Models/ItemSearchProviders/NumericIndex.ts
@@ -69,14 +69,13 @@ export default class NumericIndex implements IndexBase<NumericSearchQuery> {
       { dataRowId: 0, value: startValue },
       entry => entry.value
     );
-    const endIndex =
-      sortedLastIndexBy(
-        idValuePairs,
-        { dataRowId: 0, value: endValue },
-        entry => entry.value
-      ) - 1;
+    const endIndex = sortedLastIndexBy(
+      idValuePairs,
+      { dataRowId: 0, value: endValue },
+      entry => entry.value
+    );
     const matchingIds = idValuePairs
-      .slice(startIndex, endIndex + 1)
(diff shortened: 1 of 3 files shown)


lib/Models/ItemSearchProviders/NumericIndex.ts
```diff
@@ -73,8 +73,12 @@ export default class NumericIndex implements IndexBase<NumericSearchQuery> {
...
       (a, b) => a.value - b.value
     );
-    const startIndex = i < 0 ? ~i : i;
-    const endIndex = j < 0 ? ~j : j;
+    let startIndex = i < 0 ? ~i : i;
+    let endIndex = j < 0 ? ~j : j;
+    // iterate startIndex backward till we find the first matching value
+    while (idValuePairs[startIndex - 1]?.value === value.start) startIndex -= 1;
+    // iterate endIndex forward till we find the last matching value
+    while (idValuePairs[endIndex + 1]?.value === value.end) endIndex += 1;
```
na9da: Basically  the binary search could stop somewhere in the middle of a sequence of repeated values. 

Eg, Let's say we are searching for the value `10` in the sequence `7, 8, 9, 10, 10, 10, 10, 10, 11, 12, 13..`. `binarySearch()` might return index value `6` for start & end index. In that case we have to walk backwards & forwards in the array to discover the first & last matching value, i.e 3 & 7.

tephenavies: It seems a shame to do a binary search and then iterate linearly to find the final solution. Maybe you could switch out the `binarySearch` for [_.sortedIndexOf](https://lodash.com/docs/4.17.15#sortedIndexOf) and [_.sortedLastIndexOf](https://lodash.com/docs/4.17.15#sortedLastIndexOf).

tephenavies: Is it possible that most of this list is the same value? Like 10,000 entries with the same value (and some that are different, or the whole list is the same, or whatever would make this solution act worst)?

tephenavies (changes requested):
Needs clarification whether adding O(n) operation to O(logn) operation will have a significant impact in common cases.

na9da:
@steve9164 good point! In the worst case scenario where you have long runs of the same value that could be a problem. I'll see if I can use repeated `binarySearch()` instead to narrow it down.

na9da:
Lodash has `sortedIndex` and `sortedLastIndex` which can also do the job.

lib/Models/ItemSearchProviders/NumericIndex.ts
```diff
@@ -63,18 +64,17 @@ export default class NumericIndex implements IndexBase<NumericSearchQuery> {
...
-      (a, b) => a.value - b.value
-    );
-    const startIndex = i < 0 ? ~i : i;
-    const endIndex = j < 0 ? ~j : j;
+    const endIndex =
+      sortedLastIndexBy(
+        idValuePairs,
+        { dataRowId: 0, value: endValue },
+        entry => entry.value
+      ) - 1;
```
tephenavies: This seems to work well if `endValue` is present or if it isn't present, but I'd like to have more tests on this.

test/Models/ItemSearchProviders/NumericIndexSpec.ts
```diff
@@ -0,0 +1,23 @@
...
+        { dataRowId: 4, value: 3 },
+        { dataRowId: 5, value: 3 },
+        { dataRowId: 6, value: 3 },
+        { dataRowId: 7, value: 3 },
+        { dataRowId: 8, value: 9 },
+        { dataRowId: 9, value: 10 }
+      ]);
+      const ids = await index.search({ start: 3, end: 3 });
+      expect([...ids]).toEqual([3, 4, 5, 6, 7]);
+    });
```
tephenavies: Could you add similar tests for:
- [ ] Start and end are different, and no values in the range [start, end] are present in the index
- [ ] Same as above but the range is lower than any values

and maybe a couple of others.

na9da: I have added 3 more test cases to validate behavior for out of range search queries and in range queries.

tephenavies (approved):

PR #5589, opened 2021-07-01
Show searchtool load progress

### LINKED ISSUE #5563: Search: Show a progress indicator when opening the search tool.
Currently when searching datasets, there is a delay in opening the search tool if the network is already saturated by tile load requests. Add some kind of progress indicator to show that the search tool is loading.

na9da:
### What this PR does

Fixes #5563 

Adds a `LazyItemSearchTool` to show a loading shell while the actual `ItemSearchTool` loads.

This change is required because currently there is no progress indicator between clicking the `Search` button and showing the search form. This could take several seconds depending on the available network speed and the user can get the feeling that the app is broken.

### Checklist

### DIFF
--- a/lib/ReactViews/Errors/RaiseToUserErrorBoundary.tsx
+++ b/lib/ReactViews/Errors/RaiseToUserErrorBoundary.tsx
@@ -1,0 +1,29 @@
+import i18next from "i18next";
+import React, { ErrorInfo } from "react";
+import ViewState from "../../ReactViewModels/ViewState";
+
+/**
+ * An error boundary that raise the error to the user.
+ */
+export default class RaiseToUserErrorBoundary extends React.Component<{
+  viewState: ViewState;
+}> {
+  state = { hasError: false };
+
+  static getDerivedStateFromError(error: any) {
+    return {
+      hasError: true
+    };
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    const newError = new Error(i18next.t("itemSearchTool.toolLoadError"));
+    newError.name = error.name;
+    newError.stack = error.stack;
+    this.props.viewState.terria.raiseErrorToUser(newError);
+  }
+
+  render() {
+    return this.state.hasError ? null : this.props.children;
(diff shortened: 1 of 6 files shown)


lib/ReactViews/Tools/ItemSearchTool/LazyItemSearchTool.tsx
```diff
@@ -0,0 +1,67 @@
...
+    </Suspense>
+  );
+};
+
+const Wrapper = styled(Main)`
+  align-items: center;
+  justify-content: center;
+`;
+
+class LazyLoadErrorBoundary extends React.Component<{ viewState: ViewState }> {
```
KeyboardSounds: To avoid duplicate implementations of near-identical components in the future, do you think it would be a good idea to put this in its own file and genericise the error message so that it's more visible and reusable?

na9da: Good point @KeyboardSounds. I have extracted it out as a component called `RaiseToUserErrorBoundary`.

lib/ReactViews/Errors/RaiseToUserErrorBoundary.tsx
```diff
@@ -0,0 +1,29 @@
...
+  state = { hasError: false };
+
+  static getDerivedStateFromError(error: any) {
+    return {
+      hasError: true
+    };
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    const newError = new Error(i18next.t("itemSearchTool.toolLoadError"));
```
KeyboardSounds: I think you should make the error message here configurable, otherwise it will always be ""Error loading search tool." even if this component isn't wrapping the search tool.

na9da: Haha yes.. I have changed it to accept terria error options to customize the error message shown to user.

KeyboardSounds (approved):
Thanks, Nanda!

PR #5604, opened 2021-07-02
Move trait decorators to new folder

zoran995:
### What this PR does

Fixes 

Move the decorators to their own folders so it is easier to find them when needed. I always struggle when need to find them
~~Replace isGeoserver, isEsri, isThredds with single trait. There is no point to assign them all at the same time. In the future we can create an enumTrait that will accept enum value.~~ I am not sure if I correctly understood how those traits are used so reverted this part.

### Checklist


nf-s:
Hi @zoran995 

I really like this idea - but was thinking if we should also move the other "core" Trait files into a different folder - like:
- `ArrayNestedStrataMap.ts`
- `mixTraits.ts`
- `ModelReference.ts`
- `ModelTraits.ts`
- `NestedStrataMap.ts`
- `Stratified.ts`
- `Trait.ts`
- `traitsClassToModelClass.ts`
- `TraitsConstructor.ts`

What do you think @steve9164 ?

tephenavies:
@nf-s, @zoran995 I also like those ideas. Maybe we could have inside `lib/Traits` a `TraitsClasses` folder with all the various Traits classes and a `decorators` folder with the decorators and the other files @nf-s listed would be the only files in `lib/Traits`.

zoran995:
@nf-s, @steve9164 moved all traits classes to `lib/Traits/TraitsClasses`, decorators are in `lib/Traits/Decorators` and core traits classes (listed above) are left in `lib/Traits`

nf-s (approved):

PR #5605, opened 2021-07-02
Replace isGeoserver, isEsri, isThredds with one trait

zoran995:
### What this PR does

Continues on #5604
Replace isGeoserver, isEsri, isThredds with single trait.  I don't see when user will need to assign multiple of them at the same time.

### Checklist


nf-s:
At the moment we are only really using `isGeoServer` and `isNcWMS`.
`isEsri` isn't used for anything, and `isThredds` is only used to set `isNcWMS`.

Our `Traits` system doesn't deal with those string enum union types (`"esri" | "geoserver" | ...`) very well. Until we have proper support from them I would prefer to keep using boolean traits.

nf-s:
I'd be happy to get rid of `isEsri` and replace `isThredds` with `isNcWMS` to make it more concise

zoran995:
@nf-s I played a little with adding `enumTrait`, is it valid solution? It can be useful in other places. `Thredds` is replaced with `NcWMS`
It can be tested with
- valid server definition
```
{
  "name": "Solar_Satellite_DNI_2014",
  "type": "wms",
  "layers": "Solar_Satellite_DNI_2014",
  "url": "https://gis.aremi.data61.io/bom/wms",
  "server": "geoserver"
}
```

- invalid server definition
```
{
      "name": "Solar_Satellite_DNI_2014",
      "type": "wms",
      "layers": "Solar_Satellite_DNI_2014",
      "url": "https://gis.aremi.data61.io/bom/wms",
      "server": "thredds"
}
```

PR #5607, opened 2021-07-05
Limit `SelectableDimension` options to 1000 values

nf-s:
### Limit `SelectableDimension` options to 1000 values

Fixes https://github.com/TerriaJS/terriajs/issues/5618

This is to limit time taken to render SelectableDimensions

I also added the limit to `SdmxDataflowStratrum` so we don't compute `options` if there are more than 1000 values

### To test

- **BEFORE** https://beta.nationalmap.terria.io/
- **AFTER** http://ci.terria.io/sdmx-options-fix/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/nationalmap/map-config/prod.json

- National Datasets -> Social and Economic -> ABS.Stat (Beta) -> SNAPSHOTS -> Socio-Economic Indexes for Areas (SEIFA) -> SEIFA 2016 by Statistical Area 1 (SA1)

### Before

image

### After

image

### Checklist


lib/ReactViews/Workbench/Controls/DimensionSelectorSection.tsx
```diff
@@ -38,6 +38,7 @@ class DimensionSelectorSection extends React.Component<PropsType> {
       dim =>
         !dim.disable &&
         isDefined(dim.options) &&
+        dim.options.length < 1000 &&
```
wibily: could you put the 1000 into an exported const and use it in both files so that we don't accidently miss one of these if we ever need to change the limit?

lib/Models/SdmxJson/SdmxJsonDataflowStratum.ts
```diff
@@ -269,55 +270,62 @@ export class SdmxJsonDataflowStratum extends LoadableStratum(
...
+          let options: StratumFromTraits<DimensionOptionTraits>[] = [];
+
+          // Only create options if less then 1000 values
+          if (allowedOptionIds.length < 1000) {
+            // Get codes by merging allowedOptionIds with codelist
+            let filteredCodesList =
+              (allowedOptionIds.length > 0
+                ? codelist?.codes?.filter(
+                    code =>
+                      allowedOptionIds && allowedOptionIds.includes(code.id!)
```
wibily: i know this is outside the scope of this review, but we don't need to convert allowedOptionsIds back into an array, we could instead use `allowedOptionsIdsSet.has` which is faster than the `includes` method on arrays. Also if we make allowedOptionsIdsSet an empty set if constraints is not an array, then we save ourselves a few null checks.

wibily (approved):

lib/Models/SdmxJson/SdmxJsonDataflowStratum.ts
nf-s: Yep that is much nicer!

lib/ReactViews/Workbench/Controls/DimensionSelectorSection.tsx
nf-s: Yep good idea - done

PR #5611, opened 2021-07-05
Use translation prefix with translate#

### LINKED ISSUE #5603: Update useTranslation to require prefix
Currently `useTranslation` will check any value for translation and if it exists apply that translation. This can lead to getting translation when we actually want to receive the exact string in only string in specified language. 
We should change this to require prefix `translate#string` like bellow
https://github.com/TerriaJS/terriajs/blob/c49f3f948801b1b56167c3bc98e31c5ad6ed7f75/lib/Language/languageHelpers.ts#L6-L23
this is a breaking change (used for help in `config.json`) so it will be good to handle it before v8 release (in that case we don't need deprecation warning)
file above is from https://github.com/TerriaJS/terriajs/pull/5145/

zoran995:
### What this PR does

Fixes #5603 

Update `useTranslationIfExists` function to require `translate#` prefix in front of translation key. See #5603 and https://github.com/TerriaJS/terriajs/issues/4819#issuecomment-763167436 for more details.
In future for all strings that could come from config or init we will have to use `useTranslationIfExists`.
This requires https://github.com/TerriaJS/TerriaMap/pull/534 so we don't break help panel

### Checklist


tephenavies:
~I'll merge this after alpha 88 is released.~

tephenavies (approved):

zoran995:
@steve9164 can we just get TerriaMap part merged 
> This requires [TerriaJS/TerriaMap#534](https://github.com/TerriaJS/TerriaMap/pull/534) so we don't break help panel

PR #5612, opened 2021-07-05
fix wms get feature info feature count

### LINKED ISSUE #4925: WMS select only one feature & missing highlight selected objects
- When a WMS layer is added on v7 and user click on the object it will select all available objects, on v8 currently, it will select only one object. 
- The v7 had an option to highlight the selected object, there is no such feature in v8. AFAIK it was done getting the geometry from feature response, but the WMS is not required to return Geometry in it (on MapServer it can be disabled, on GeoServer it is possible to use new geojson template and hide geometry). My thought on this is that we could utilise additional style request and load an overlay layer with a highlighted object (it should be set through catalog config).

[next test](http://ci.terria.io/next/#share=s-dKwaKSi84gryYm82uHd0zvgHy1I)
[nationalmap test](https://nationalmap.gov.au/#share=s-ebMPvcqmr6GOBP92iIhFFfxslae)

zoran995:
### What this PR does

Fixes #4925 

Added `feature_count` to `getFeatureInfo` parameters as it was done in v7 (#1050).

### Checklist

### DIFF
--- a/lib/Models/WebMapServiceCatalogItem.ts
+++ b/lib/Models/WebMapServiceCatalogItem.ts
@@ -1186,7 +1186,7 @@ class WebMapServiceCatalogItem
           ...dimensionParameters,
           feature_count:
             1 +
-            (this.maximumShownFeatureInfos ||
+            (this.maximumShownFeatureInfos ??
               this.terria.configParameters.defaultMaximumShownFeatureInfos)!,
           styles: this.styles === undefined ? "" : this.styles
         },
--- a/lib/Models/WebMapServiceCatalogItem.ts
+++ b/lib/Models/WebMapServiceCatalogItem.ts
@@ -1187,7 +1187,7 @@ class WebMapServiceCatalogItem
           feature_count:
             1 +
             (this.maximumShownFeatureInfos ??
-              this.terria.configParameters.defaultMaximumShownFeatureInfos)!,
+              this.terria.configParameters.defaultMaximumShownFeatureInfos),
           styles: this.styles === undefined ? "" : this.styles
         },
         tileWidth: this.tileWidth,
--- a/lib/Models/WebMapServiceCatalogItem.ts
+++ b/lib/Models/WebMapServiceCatalogItem.ts
@@ -1158,10 +1158,7 @@ class WebMapServiceCatalogItem
         "width",
         "height",
         "bbox",
-        "layers",
-        // This is here as a temporary fix until Cesium implements this fix
-        // https://github.com/CesiumGS/cesium/issues/9021
-        "version"
+        "layers"
       ];
 
       const baseUrl = queryParametersToRemove.reduce(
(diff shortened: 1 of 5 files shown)


lib/Models/WebMapServiceCatalogItem.ts
```diff
@@ -1187,6 +1184,10 @@ class WebMapServiceCatalogItem
         parameters: parameters,
         getFeatureInfoParameters: {
           ...dimensionParameters,
+          feature_count:
+            1 +
+            (this.maximumShownFeatureInfos ||
+              this.terria.configParameters.defaultMaximumShownFeatureInfos)!,
```
nf-s: Thanks so much for fixing this @zoran995 !

Can you please use nullish coalescing instead of logical OR here?

```suggestion
            1 +
            (this.maximumShownFeatureInfos ??
              this.terria.configParameters.defaultMaximumShownFeatureInfos)
```

And can you please change the type of `this.terria.configParameters.defaultMaximumShownFeatureInfos` to be `number` instead of `number | undefined`

https://github.com/zoran995/terriajs-i18n/blob/a878b51da78cb8756cf2f78ed20552a3d4197c48/lib/Models/Terria.ts#L103-L106

nf-s (changes requested):

lib/Models/WebMapServiceCatalogItem.ts
zoran995: changing the type of `defaultMaximumShownFeatureInfos` to be number will require an update of tests depending on Terria, if I recall correctly? Maybe handle this in separate PR to handle all `configParameters` that have a default value (half of them)
updated to nullish coalescing

zoran995: actually, this would be better? maximumShownFeatureInfos is coming from init file and it can be anything
```
(parseInt(this.maximumShownFeatureInfos) ||
              this.terria.configParameters.defaultMaximumShownFeatureInfos)
```
we don't want to specify "11" in init by mistake and end having 111 instead of 12

nf-s: The type of trait value is validated when it is updated/upserted from JSON

https://github.com/TerriaJS/terriajs/blob/33f03c0477e56ba7d7a4cc40771406cf5fc801e2/lib/Traits/primitiveTrait.ts#L44-L62

So it should only be `number | undefined`

nf-s: Also just checked, you should be able to change `configParameters.defaultMaximumShownFeatureInfos` to type `number` in `Terria.ts` without breaking tests.

zoran995: done

nf-s (approved):

nf-s:
Thanks a lot @zoran995 !

PR #5613, opened 2021-07-05
`WebMapServiceCatalogGroup` will now create layer auto-IDs using `Name` field to avoid ID clashes.

### LINKED ISSUE #102: Harvesting from DEA ows server doesn't return all catalogue
ws endpoint: https://ows.dea.ga.gov.au

from older version of terria

image

v8
image

nf-s:
### `WebMapServiceCatalogGroup` will now create layer auto-IDs using `Name` field to avoid ID clashes.

Fixes https://github.com/terriajs/de-australia-map/issues/102

Unfortunately, because of this change I had to add more shareKeys :(

See https://github.com/TerriaJS/terriajs/pull/5613#discussion_r663903084

### Test links

Using https://ows.dea.ga.gov.au/?service=WMS&version=1.3.0&request=GetCapabilities - which has duplicate layer `Title`s (which is valid WMS - but we don't handle properly)

- After http://ci.terria.io/wms-group-id-fix/#share=s-bgXPnmHxWdu59rfRXjtC3pYwBRG

- Before http://ci.terria.io/next/#share=s-bgXPnmHxWdu59rfRXjtC3pYwBRG

image

### Checklist

- [x] Test with live share link buckets

### DIFF
--- a/lib/ModelMixins/GroupMixin.ts
+++ b/lib/ModelMixins/GroupMixin.ts
@@ -108,17 +108,29 @@ function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
        * - member.uniqueId = 'some-group-id/some-member-id'
        * - group.shareKeys = 'old-group-id'
        * - So we want to create member.shareKeys = ["old-group-id/some-member-id"]
+       *
+       * We also repeat this process for each shareKey for each member
        */
 
       this.memberModels.forEach((model: BaseModel) => {
         // Only add shareKey if model.uniqueId is an autoID (i.e. contains groupId)
         if (isDefined(model.uniqueId) && model.uniqueId.includes(groupId)) {
-          shareKeys.forEach(groupShareKey =>
+          shareKeys.forEach(groupShareKey => {
+            // Get shareKeys for current model
+            const modelShareKeys = this.terria.modelIdShareKeysMap.get(
+              model.uniqueId!
+            );
+            modelShareKeys?.forEach(modelShareKey => {
+              this.terria.addShareKey(
+                model.uniqueId!,
+                modelShareKey.replace(groupId, groupShareKey)
+              );
+            });
             this.terria.addShareKey(
               model.uniqueId!,
               model.uniqueId!.replace(groupId, groupShareKey)
-            )
-          );
+            );
+          });
         }
       });
     }
--- a/lib/ModelMixins/GroupMixin.ts
+++ b/lib/ModelMixins/GroupMixin.ts
@@ -14,7 +14,7 @@ import ModelReference from "../Traits/ModelReference";
 import CatalogMemberMixin from "./CatalogMemberMixin";
 
 function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
-  abstract class GroupMixin extends Base implements Group {
+  abstract class Klass extends Base implements Group {
     private _memberLoader = new AsyncLoader(this.forceLoadMembers.bind(this));
 
     /**
@@ -91,7 +91,7 @@ function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
     }
 
     @action
-    addShareKeysToMembers(): void {
+    addShareKeysToMembers(members = this.memberModels): void {
       const groupId = this.uniqueId;
       if (!groupId) return;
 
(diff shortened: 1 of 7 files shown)


lib/ModelMixins/GroupMixin.ts
```diff
@@ -108,17 +108,29 @@ function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
...
+                modelShareKey.replace(groupId, groupShareKey)
+              );
+            });
             this.terria.addShareKey(
               model.uniqueId!,
               model.uniqueId!.replace(groupId, groupShareKey)
-            )
-          );
+            );
+          });
```
nf-s, line 133: This adds another level of `shareKey` generation. 
This means that if the group has `shareKeys` and a member has `shareKeys` - new `shareKeys` will be added by replacing the `groupId` element of each model `shareKey` with each group `shareKey`

For example

- We have a WMS group with id `"/groupID"` and sharekeys `["groupShareID"]`
- This group has a WMS item with id `"/groupID/itemID"` and sharekeys `["groupID/itemShareID"]`

With this change, the WMS item will now get an extra `shareKey` 
- `["groupID/itemShareID", "groupShareID/itemShareID"]`

This is generated by replacing the group's ID (`groupID`) with the groups shareKey (`groupShareID`)

lib/Models/WebMapServiceCatalogGroup.ts
```diff
@@ -288,7 +288,18 @@ class GetCapabilitiesStratum extends LoadableStratum(
...
+
+  /** For backward-compatibility.
+   * If layer.Name is defined, we will use it to create layer autoID (see `this.getLayerId`).
+   * Previously we used layer.Title, so we now add it as a shareKey
+   */
+  getLayerShareKeys(layer: CapabilitiesLayer) {
+    if (isDefined(layer.Name) && layer.Title !== layer.Name)
+      return [`${this.catalogGroup.uniqueId}/${layer.Title}`];
+
+    return [];
```
nf-s, line 302: These changes should keep most v7 IDS working

https://github.com/TerriaJS/terriajs/blob/6153d332aa3bd160ddbbd968be37a339a4f6ad96/lib/Models/WebMapServiceCatalogGroup.js#L263-L271

nf-s: I can always add a shareKey for `layer.Abstract` to support all values of `titleField`

nf-s: Actually I just found this in the `/0006-dynamic-groups-and-shareKeys.md` ADR  - which I never actually implemented

https://github.com/TerriaJS/terriajs/blob/5d13c2ba093fa0e35aeaaf27178622c4a1034625/architecture/0006-dynamic-groups-and-shareKeys.md#L106-L108

This won't fix the problem, but I should I implement this anyway?

tephenavies (approved):
This seems like a sane way of allowing change in auto id generation within dynamic groups. Although it adds some complexity to the shareKeys system the changes are well contained within the function in GroupMixin that already deals with shareKeys, and the dynamic groups

lib/ModelMixins/GroupMixin.ts
```diff
@@ -108,17 +108,33 @@ function GroupMixin<T extends Constructor<Model<GroupTraits>>>(Base: T) {
...
+          // If member is a Group -> apply shareKeys to the next level of members
+          if (GroupMixin.isMixedInto(model)) {
+            this.addShareKeysToMembers(model.memberModels);
+          }
```
nf-s, line 137: This is needed here, because some nested groups won't propagate `shareKeys` properly.

For example `WebMapServiceCatalogGroup` - which doesn't set autoIDs in nested groups the same as `CatalogGroup`



Here you would expect the `"Surface Reflectance 25m Annual Geomedian (Landsat 8)"` catalog item to have an auto ID of: 

- `"\\Digital Earth Australia - OGC Web Services\`**`Surface Reflectance`**`\Surface Reflectance 25m Annual Geomedian (Landsat 8)"`

But it doesn't. WMS group generates IDs based on `${rootGroupID}\{layer.Name}` - to its actual ID is:

- `"\\Digital Earth Australia - OGC Web Services\Surface Reflectance 25m Annual Geomedian (Landsat 8)"`

-----------

Another thing, some dynamic catalog groups (eg WMS, WFS, SDMX) will create nested `CatalogGroup`s, but won't load them - which will therefore not call `group.addShareKeysToMembers()`.

PR #5616, opened 2021-07-07
Fix `StyledIcon` css `display` clash

nf-s:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/5614

Before http://ci.terria.io/next/#share=s-dr1vf3l917Tu61liGsZGabDL13s

After http://ci.terria.io/fix-icon-css/#share=s-dr1vf3l917Tu61liGsZGabDL13s

### Checklist


lib/Styled/Icon.tsx
```diff
@@ -165,10 +162,10 @@ export const StyledIcon = styled(Icon)<IStyledIconProps>`
   ${props =>
     props.displayInline &&
     `
-  display: inline; 
+  display: inline;
   vertical-align: middle;
   `}
-  
+
```
zoran995: Should we add a `display: block` option, it was a default option when `displayInline` wasn't specified in props. 

```css
  ${props =>
    !props.displayInline &&
    `
  display: block
  `}
```

zoran995: Along with above, when vertically centering elements it might be good to put following to outer element (in this catalog case it's button).
```css
display: flex;
align-items: center;
```
With that, we should get pixel-perfect center without having to think if icon is `block` or `inline` (currently it is slight moved down)

wibily (approved):

lib/Styled/Icon.tsx
nf-s: Done!

zoran995 (approved):

PR #5620, opened 2021-07-08
Wc typity type

wibily:
### What this PR does

Refactor story component into tsx

### DIFF
--- a/lib/ReactViews/Story/Story.tsx
+++ b/lib/ReactViews/Story/Story.tsx
@@ -9,21 +9,24 @@ import Text from "../../Styled/Text";
 import Spacing from "../../Styled/Spacing";
 import parseCustomHtmlToReact from "../Custom/parseCustomHtmlToReact";
 import Icon, { StyledIcon } from "../../Styled/Icon";
-import classNames from "classnames";
+
+export type Story = {
+  title: string;
+  text: string;
+  id: string;
+  shareData?: any;
+};
 
 interface Props {
-  story: any;
-  editStory: any;
-  viewStory: any;
-  deleteStory: any;
-  recaptureStory: any;
-  recaptureStorySuccessful: any;
-  onMouseDown: any;
-  onTouchStart: any;
-  style: any;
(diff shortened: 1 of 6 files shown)


lib/ReactViews/Story/Story.tsx
```diff
@@ -0,0 +1,321 @@
...
+
+import Box from "../../Styled/Box";
+import { RawButton } from "../../Styled/Button";
+import Ul from "../../Styled/List";
+import Text from "../../Styled/Text";
+import Spacing from "../../Styled/Spacing";
+import parseCustomHtmlToReact from "../Custom/parseCustomHtmlToReact";
+import Icon, { StyledIcon } from "../../Styled/Icon";
+
+export type Story = {
```
KeyboardSounds: Why use `type` here and not `interface`?

wibily: yup will change to an interface

lib/ReactViews/Story/StoryBuilder.jsx
```diff
@@ -417,15 +417,16 @@ const StoryBuilder = observer(
                       key={`${story.id}`}
                       story={story}
                       sortData={story}
-                      deleteStory={this.removeStory.bind(this, index)}
-                      recaptureStory={this.recaptureScene}
+                      deleteStory={() => this.removeStory(index, story)}
```
KeyboardSounds, line 420: I feel like I'm missing something obvious but why are all these event handlers wrapped in closures now?

wibily: ah this is so that the function that's passed in as props has the story bound already. Previously, we had code like `props.deleteStory(props.story)` which doesn't quite make sense to me.

The reason why i didn't do something like `this.removeStory.bind(this, index, story)` is that react throws up a warning about how it auto-binds this in a performant way which is why i have the anonymous functions.

KeyboardSounds (approved):

PR #5642, opened 2021-07-16
Feature/story default

bpmweel:
Fixing postconfirmation function

- Now actually copies story 0
- Also copies the image on the s3 bucket


bpmweel:
Ignore this, picked the wrong upstream. sorry!

PR #5644, opened 2021-07-17
Basemap config

### LINKED ISSUE #5382: Enable choosing of basemaps that are available
After playing around with basemaps and checking the #5366. We probably should do some changes to the basemap model:
- change the name of the array `baseMaps` to `baseMapCatalog` or `baseMapItems`
- add an array basemaps which will say which basemaps should be available to the user (similarly to how workbench property works)
- in this case default base maps list should be merged in the baseMapCatalog so the user doesn't have to redefine it in catalog file, but also can provide another one with different id and say use this one.

zoran995:
### What this PR does

Fixes #5382

- baseMaps is now a core object for entire base map configuration
- list of available basemaps is defined in `baseMaps.init`. This list is combined with default base maps so it is easy to override some parameter
 - definition of `initBaseMapId` and `initBaseMapName` are moved to `baseMaps.defaultBaseMapId` and `baseMaps.defaultBaseMapName`
- `previewBaseMapId` is moved to `baseMaps.previewBaseMapId`
- implemented `baseMaps.useBaseMaps` array of base map ids so it's easy to define the list of basemaps that will be available to user. The order of baseMaps in settings panel will be the same as in this list. When undefined show all baseMaps (default basemaps first)

Init config for baseMaps now uses the Traits system. It might be good to do entire init file that way.

### Checklist


zoran995:
Really happy about one feature - it's possible to reference an item from catalog members using it's id and specifying image :astonished:. Will provide config samples after tests

zoran995:
a sample file with referencing basemap from catalog and override the positron basemap image and name https://gist.githubusercontent.com/zoran995/d29eb334ead7c49e1cc00ffc52ab8e25/raw/985349b90c1d9e1b82d683eae0276f5df043cf4f/basemaps.json

zoran995:
This is ready for review

lib/Models/Terria.ts
```diff
@@ -1239,27 +1224,16 @@ export default class Terria {
...
-    if (isJsonString(initData.baseMapName)) {
-      this.initBaseMapName = initData.baseMapName;
-    }
-
-    if (isJsonString(initData.previewBaseMapId)) {
-      this.previewBaseMapId = initData.previewBaseMapId;
-    }
+    this.baseMapsModel.initializeDefaultBaseMaps().catchError(error => {
+      errors.push(error);
+    });
```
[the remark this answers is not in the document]
nf-s: https://github.com/TerriaJS/terriajs/blob/c32ea2521deb847d97525317c0379a8ab188ed0a/lib/Models/Terria.ts#L770-L786

lib/Traits/TraitsClasses/BaseMapTraits.ts
```diff
@@ -0,0 +1,65 @@
...
+    type: "string",
+    name: "Image",
+    description: "Path to the basemap image"
+  })
+  image?: string;
+
+  @modelReferenceTrait({
+    factory: CatalogMemberFactory,
+    name: "Base map item",
+    description: "Base map item definition"
```
nf-s: Could this please include a sentence about referencing existing catalog items (eg `"//Surface Geology"`)

lib/Traits/TraitsClasses/BaseMapTraits.ts
```diff
@@ -0,0 +1,65 @@
...
+  items?: BaseMapTraits[];
+
+  @primitiveArrayTrait({
+    type: "string",
+    name: "useBaseMaps",
+    description:
+      "Array of base maps ids that is available to user. Use this do define order of the base maps in settings panel. Leave undefined to show all basemaps."
+  })
+  useBaseMaps?: string[];
+}
```
nf-s: Do you mind renaming this to something like `enabledBaseMaps` ?

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
```diff
@@ -95,6 +95,8 @@ function addStratum(terria, stratumId, initSource) {
   });
 }
 
+function addBaseMaps(terria, initSource) {}
+
```
nf-s, line 99: empty function

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
```diff
@@ -65,7 +65,7 @@ export function getShareData(
   addTimelineItems(terria, initSource);
   addViewSettings(terria, viewState, initSource);
   addFeaturePicking(terria, initSource);
-
+  addBaseMaps(terria, CommonStrata.definition, initSource);
```
nf-s, line 68: Maybe remove this until we sort out `basemapId`...

nf-s (changes requested):
Thanks @zoran995. This is a great feature!

A few small issues:
- See comment about moving `this.baseMapsModel.initializeDefaultBaseMaps()` to `terria.start()`
- See comment about removing `baseMapsModel._defaultBaseMaps` property, and added default basemaps to `CommonStrata.defaults` instead of `CommonStrata.definition`
- A few comments about names of things

Another potential issue - I think we should remove all of the basemap logic from `TerriaViewer.ts`

https://github.com/TerriaJS/terriajs/blob/0cdd4743f5548124f295e83ec4956fb2aeb93f92/lib/ViewModels/TerriaViewer.ts#L38-L69

I think this should be handled by `BaseMapsModel`.

I also think we should rename `defaultBaseMapId` to `baseMapId` - then all user interaction can be saved to the `user` stratum of `BaseMapsModel` - and then for share link, we can serialize the `user` stratum in the same way we do for all other models.

For example - to set the basemap we currently use this:
https://github.com/TerriaJS/terriajs/blob/96f1a661687a120bcc934605870d56487cff6d20/lib/ReactViews/Map/Panels/SettingPanel.jsx#L66-L68

But it would become something like this

```ts
this.props.terria.baseMapsModel.setTrait(CommonStrata.user, "baseMapId", baseMap)
```

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
nf-s, line 99: If you end up moving `TerriaViewer.basemaps` into `BaseMapsModel` - you will need to use something like `addModelStratum()` to add `basemaps` model to the `initSource`

zoran995:
@nf-s thanks for reviewing this, I just have one concern for moving the base map logic from `TerriaViewer` to `BaseMapsModel`. 
We have multiple instances of `TerriaViewer` (data preview and pedestrian tool mini-map), and this change would require `BaseMapsModel` to know a base map for each instance or to use the same base map in each. 
IMHO this would just complicate the implementation of the `BaseMapsModel`, also it would be pretty bad to enforce users using the same base map.
I might be missing something but I don't see a clean way to move that part to `BaseMapsModel`.

lib/Models/BaseMaps/BaseMapsModel.ts
```diff
@@ -0,0 +1,138 @@
...
+  // Can't do this in constructor since {@link CatalogMemberFactory} doesn't
+  // have any values at the moment of initializing Terria class.
+  initializeDefaultBaseMaps(): Result {
+    if (this._defaultBaseMaps.length === 0 && this.items.length === 0) {
+      this._defaultBaseMaps.push(...defaultBaseMaps(this.terria));
+      return this.loadFromJson(CommonStrata.definition, <any>{
+        items: this._defaultBaseMaps
+      });
+    }
+    return Result.none();
```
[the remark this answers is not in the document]
zoran995: done

lib/Models/Terria.ts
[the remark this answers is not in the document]
zoran995: done

lib/Traits/TraitsClasses/BaseMapTraits.ts
zoran995: done

lib/Traits/TraitsClasses/BaseMapTraits.ts
zoran995: done

lib/Traits/Decorators/modelReferenceTrait.ts
```diff
@@ -0,0 +1,112 @@
...
+      constructor.traits = {};
+    }
+    constructor.traits[propertyKey] = new ModelReferenceTrait(
+      propertyKey,
+      options
+    );
+  };
+}
+
+export class ModelReferenceTrait extends Trait {
```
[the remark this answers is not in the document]
zoran995: will do after this is merged

nf-s:
Fair point. Yeah we would also need to add `previewBaseMap` or something to `BaseMapsModel`.

I'm happy to leave it for now, but I'd like to investigate it in the future - as I think it may be a bit cleaner to keep all basemap logic and state in `BaseMapsModel`.

nf-s (approved):
Thanks again @zoran995 for this PR - it is a great feature!

PR #5344, opened 2021-03-21
Tsify workbench

zoran995:
### What this PR does

Fixes #<insert issue number here if relevant>

Moving workbench to styled-components and typescript. This depends on #5342.

For testing this, basically, try adding every possible dataset and verify that workbench is correct.

Some of the changes, although most of them are just move to typescript and styled components
- No longer use calculated height in workbench, let flex do this
- Changed checkbox in item visibility to use checkbox component instead of a button
- chart selector uses a checkbox
- Time section now fills the entire width
- Fixed some small issues identified while working on this that I no longer recall

### Checklist

-   [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - existing tests pass succesfully
-   [ ] I've updated CHANGES.md with what I changed.

PR #5648, opened 2021-07-19
refactored BadgeBar to handle longer strings

wibily:
### What this PR does

Fixes de-africa-map#98

Refactored BadgeBar to use flexbox instead of floats. To test suffix ?lng=fr or ?lng=en to trigger the translation switch
ie
http://ci.terria.io/wc-fr/?lng=fr
vs
http://ci.terria.io/next/?lng=fr

Screen Shot 2021-07-19 at 5 28 38 pm
Screen Shot 2021-07-19 at 5 28 31 pm
Screen Shot 2021-07-19 at 5 28 11 pm
Screen Shot 2021-07-19 at 5 28 00 pm

### Checklist


zoran995:
this moves the remove all button to the center of badge bar? 
There is a ts version of badge bar in #5344 which should work correctly https://github.com/TerriaJS/terriajs/blob/63b8cc0e3870cac1710af28d3b7e7060fa5a0c0c/lib/ReactViews/BadgeBar.tsx feel free to take it

lib/ReactViews/BadgeBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <TextSpan
+            textLight
+            uppercase
+            css={`
+              padding-left: 5px;
+            `}
+          >
+            ({props.badge})
+          </TextSpan>
+        )}
```
zoran995:
```suggestion
        <TextSpan textLight uppercase>
          {props.label} {props.badge ? `(${props.badge})` : null}
        </TextSpan>
```

Last night I pushed a small fix for badge bar when having longer texts, just saw that I posted a commit link not branch link https://github.com/zoran995/terriajs-i18n/blob/workbench-ts/lib/ReactViews/BadgeBar.tsx

wibily: many thanks @zoran995

wibily: thanks @zoran995, I have also added the nowrap to ensure that it remains on one-line. One thing I might investigate tommorrow is to add ellipses and a title so that we might be able to fail a bit more gracefully when dealing with extra-long translation strings

na9da (approved):
Looks good @wibily !

Issue #5275, opened 2021-02-25
GeoJsonMixin forceLoadMapItems needs rewrite


tephenavies:
https://github.com/TerriaJS/terriajs/blob/cd5bae29dfbbbed7721da9566e41ecf4c3ade8b3/lib/ModelMixins/GeojsonMixin.ts#L122-L138

Currently the whole function runs even if `customDataLoader` loads something. `customDataLoader` should return a Promise when it loads something and undefined when it doesn't load something. I think instead of using the Promise constructor it would be neater to use `Promise.resolve().then(` or maybe async/await.

nf-s:
`GeoJsonMixin` now follows an approach similar to `TableMixin` https://github.com/TerriaJS/terriajs/pull/5900

https://github.com/TerriaJS/terriajs/blob/92f47a6872b4d521093d01ddf776178d3fb71d67/lib/ModelMixins/GeojsonMixin.ts#L305-L317

Issue #5136, opened 2021-01-13
Implement featureInfoUrlTemplate for other catalog items


zoran995:
It would be nice to implement `featureInfoUrlTemplate` for other catalog items as it is already implemented for Cesium3DTiles. There is already an indication in v8 documentation about its support coming from `FeatureInfoTraits`, used in most of the `CatalogItems`.
When the user clicks on the feature, he might get only the feature id and get the rest of the info from the API accepting that ID as parameter. This allows organizations to maintain only alphanumeric data and publish static geojson, while the map always returns the accurate feature info. This is especially useful when one organization holds the alphanumeric data while the other maintains the geometry.

PR #5032, opened 2020-11-28
Revert some mobx.action changes

tephenavies:
### What this PR does

Removes some `mobx.action` usages that could cause memoisation problems.

Fixes failing test added #5026.

### Checklist

-   [x] There are unit tests to verify my changes are correct (see #5026)
-   [x] Changes are fixing a previous change within this version, so no CHANGES.md entry needed

Issue #4542, opened 2020-07-16
GeoJSON feature info only top feature, 2D


zoran995:
In the 2D viewer, when having multiple overlapping elements in GeoJSON the GetFeatureInfo tool will select only the top element and the underlying elements won't be selectable. 

[example layer](https://gist.githubusercontent.com/zoran995/42ba2355a38099a93de58be4352fe1f4/raw/9b8d0aca5046f1146bb298abc0be8398c6361a56/tiger_landmarks.json)

[Map share](http://ci.terria.io/next/#share=s-hNbifgzA51tKIdbXZZeDFg3Gg6j)

nf-s:
Closed by https://github.com/TerriaJS/terriajs/pull/5660

PR #5660, opened 2021-07-21
Implement `featureInfoUrlTemplate` for Geojson data

### LINKED ISSUE #5659: 2d vector feature info bug
When working in 2D map viewer feature info will always return only one feature. In v7 it returned all features for position
v8
image

v7
image

zoran995:
### What this PR does

Fixes #5659, #4542, #5275

Leaflet monkey patch for onClick event is still valid, there was no major changes regarding this part. Also updated it to match the changes up to 1.7.1. This resolves selection of multiple vector features in 2D viewer.
Also implemented support for FeatureInfoMixin in 2D viewer, and connected it to GeojsonMixin.

This is also a start towards the implementation of #5136, other catalog items can be done later.

For fixing #5275 had to move the loading logic into a single `dataLoader`, when providing additional data loading method we will have to call `super.dataLoader()`. The reason for this is when using an old combination with `customDataLoader`, adding
```
const geojsonData = await this.customDataLoader();
if (isDefined(geojsonData)) return geojsonData;
```
triggered memoization somewhere which results in failed test (see #5026 and #5032 for more details). I might miss something but this is the only working solution I found.

### Checklist


lib/Map/geoRssConvertor.js
```diff
@@ -115,10 +115,7 @@ function getGeoRssPropertiesRecursively(gmlNode, properties) {
       continue;
     }
 
-    if (
-      child.hasChildNodes() &&
-      getGeoRssPropertiesRecursively(child, properties)
-    ) {
```
zoran995: this led to features not having any properties

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -110,31 +118,47 @@ export default function GeoJsonMixin<
...
+     */
+    buildFeatureFromPickResult(
+      _screenPosition: Cartesian2 | undefined,
+      pickResult: any
+    ): Feature | undefined {
+      if (pickResult instanceof Entity) {
+        return Feature.fromEntityCollectionOrEntity(pickResult);
+      } else if (isDefined(pickResult?.id)) {
+        return Feature.fromEntityCollectionOrEntity(pickResult.id);
+      }
```
zoran995: 2D pick result is entity so we need to resolve it like that

lib/Models/OpenDataSoftCatalogItem.ts
```diff
@@ -592,13 +593,18 @@ export default class OpenDataSoftCatalogItem
     _screenPosition: Cartesian2 | undefined,
     pickResult: any
   ) {
-    const feature = new Feature(pickResult?.id);
+    let feature: Feature | undefined;
+    if (pickResult instanceof Entity) {
+      feature = Feature.fromEntityCollectionOrEntity(pickResult);
+    } else {
+      feature = new Feature(pickResult?.id);
+    }
```
zoran995: same as for GeojsonMixin

zoran995:
This turned into converting some catalog items to use `GeojsonMixin`, @nf-s I tried to apply changes from error handling PR

zoran995:
This is ready for review

zoran995:
Converted this to draft as it doesn't work on MVT layers (they are using raster feature picking), it will take some time to provide a fix

zoran995:
This now works for MVT layers, it can also be easily  implemented for other raster layers by applying Mixin and wrapping the imagery provider (tested locally with ArcGIS map server), I just decided to skip it in this iteration and implement in follow up if this is accepted.

If a user is using cesium primitives the feature info will be shown while fetch is done in the background and after each fetch is done feature info for that feature is updated, on the other hand with raster layers fetch will be awaited and everything will be shown at the same time.

tephenavies:
I like the idea of being able to use featureInfoUrlTemplate more widely (there's a few datasets I'd really want to use it for that are TableMixin region mapping). What's the status of this?

zoran995:
@steve9164  This is ready for review, just needed to fix GeoRSS tests after changing it to use GeojsonMixin. Thought that it will take much more time but it was really straightforward. 
Here is a basic catalog that can be used for testing https://gist.githubusercontent.com/zoran995/8200b51b38ff6c76a954b52b223e4319/raw/6ef1e640a1732e45ca59a75baf157be6b95db0e8/feature-info-url-template-test-data.json

lib/Core/isDefined.ts
```diff
@@ -1,3 +1,3 @@
-export default function isDefined<T>(value?: T): value is T {
-  return value !== undefined;
+export default function isDefined<T>(value: T | undefined | null): value is T {
+  return <T>value !== undefined && <T>value !== null;
```
nf-s: Are the type assertions here needed?

lib/Map/LeafletPatched.js
```diff
@@ -0,0 +1,27 @@
...
+        layers.push(layer);
+      }
+    }
+  }
+  if (layers.length) {
+    this._fireEvent(layers, e);
+  }
+};
+
+export default L;
```
nf-s: Thanks for porting this!

lib/ModelMixins/FeatureInfoMixin.ts
```diff
@@ -57,6 +62,17 @@ export default function FeatureInfoMixin<T extends Constructor<Target>>(
...
               });
+              // Update description of the feature after it is resolved from
+              // feature info template url
+              feature.description = new ConstantProperty(
+                describeFromProperties(
+                  feature.properties,
+                  (feature._catalogItem as any)?.currentTimeAsJulianDate ||
+                    JulianDate.now(),
+                  (feature._catalogItem as any)
+                    ?.showStringIfPropertyValueIsNull ?? false
```
nf-s:
```suggestion
                  (TimeVarying.is(this)
                    ? this.currentTimeAsJulianDate
                    : undefined) ?? JulianDate.now(),
                  TimeVarying.is(this)
                    ? this.showStringIfPropertyValueIsNull
                    : false
```

lib/ModelMixins/FeatureInfoMixin.ts
```diff
@@ -71,7 +87,87 @@ export default function FeatureInfoMixin<T extends Constructor<Target>>(
...
+                } else {
+                  feature.properties[property] = featureInfo[property];
+                }
+              });
+              // Update description of the feature after it is resolved from
+              // feature info template url
+              feature.description = describeFromProperties(
+                feature.properties,
+                (this as any)?.currentTimeAsJulianDate || JulianDate.now(),
+                (this as any)?.showStringIfPropertyValueIsNull ?? false
```
nf-s:
```suggestion
                (TimeVarying.is(catalogItem)
                  ? catalogItem.currentTimeAsJulianDate
                  : undefined) ?? JulianDate.now(),
                TimeVarying.is(catalogItem)
                  ? catalogItem.showStringIfPropertyValueIsNull
                  : false
```

lib/ModelMixins/FeatureInfoMixin.ts
```diff
@@ -71,7 +87,87 @@ export default function FeatureInfoMixin<T extends Constructor<Target>>(
...
+        if (
+          isDefined(catalogItem.featureInfoUrlTemplate) &&
+          isDefined(features) &&
+          features.length < catalogItem.maxRequests
+        ) {
+          for (let i = 0; i < features.length; i++) {
+            const feature = features[i];
+            const resource = new Resource({
+              url: proxyCatalogItemUrl(
+                this,
```
nf-s:
```suggestion
                catalogItem,
```

lib/Map/LeafletPatched.js
nf-s: I'm currently getting this when testing locally with the "GeoJSON Test" catalog item

image

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -794,6 +813,9 @@ function GeoJsonMixin<T extends Constructor<Model<GeoJsonTraits>>>(Base: T) {
         }
       ];
 
+      // _catalogItem property is needed for some feature picking functions (eg FeatureInfoMixin)
+      (feature as any)._catalogItem = this;
+
```
nf-s: This is overwriting the `feature` import from `"@turf/helpers"`

Can this be done generically at the bottom of `forceLoadMapItems`? - for example

```ts
        this._dataSource?.entities.values.map(
          entity => ((entity as any)._catalogItem = this)
        );
```

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -997,6 +1019,9 @@ function GeoJsonMixin<T extends Constructor<Model<GeoJsonTraits>>>(Base: T) {
 
         const properties = entity.properties;
 
+        // _catalogItem property is needed for some feature picking functions (eg FeatureInfoMixin)
+        (entity as any)._catalogItem = this;
+
```
nf-s: See above - regarding doing the generically in `forceLoadMapItems`

lib/Models/Catalog/CatalogItems/GeoRssCatalogItem.ts
```diff
@@ -178,7 +127,7 @@ class GeoRssStratum extends LoadableStratum(GeoRssCatalogItemTraits) {
 
 StratumOrder.addLoadStratum(GeoRssStratum.stratumName);
 
-export default class GeoRssCatalogItem extends MappableMixin(
+export default class GeoRssCatalogItem extends GeoJsonMixin(
   UrlMixin(CatalogMemberMixin(CreateModel(GeoRssCatalogItemTraits)))
 ) {
```
nf-s: This should have  `implements HasLocalData`

nf-s (changes requested):
Thanks for the PR @zoran995 !
Sorry it took so long to review

lib/Map/LeafletPatched.js
zoran995: Initially, when I implemented this we were on 1.7 of Leaflet, in 1.8 fakeStop is removed and apparently not needed anymore as it is removed in the original 1.8 implementation
https://github.com/Leaflet/Leaflet/pull/7439/files#diff-f4127fe2c74bbf5c584e01246849fac1954ead8521449585b61d268bfb7b059eL358

lib/ModelMixins/FeatureInfoMixin.ts
zoran995: thanks

zoran995:
This is a new init file that can be used for testing (geojson, georss and czml)
https://gist.githubusercontent.com/zoran995/8200b51b38ff6c76a954b52b223e4319/raw/aacaec1bf485a281c42aa5aac8a2e372de4a8df3/feature-info-url-template-test-data-2.json

nf-s (approved):
Works well!

Thanks a lot @zoran995

PR #5663, opened 2021-07-22
Clean up ViewControl traits + colors

nf-s:
### Clean up ViewControl traits + colors 

* Clean up `SplitterTraits` and `ViewControl` colors
  * `$color-splitter` and `theme.colorSplitter` has been replaced with `$color-secondary` and `theme.colorSecondary`
* `SplitterTraits` is now included in `RasterLayerTraits`
  * Removed `supportsSplitter` variable
  * Added `disableSplitter` trait
  * This fixes splitter support for a few catalog items
* Clean up `canZoomTo`
  * Replaced with `disableZoomTo` in `MappableTraits`
* Clean up `showsInfo`
  * Replaced with `disableAboutData` in `CatalogMemberTraits`

### Checklist


CHANGES.md
```diff
@@ -9,6 +9,9 @@ Change Log
   - `colorPalette` no longer supports a list of CSS colors (eg `rgb(0,0,255)-rgb(0,255,0)-rgb(255,0,0)`). Instead please use `binColors`.
   - Organise `Traits` folder into `Traits/Decorators` and `Traits/TraitsClasses`
   - Renamed all mixin instance type definitions to `XMixin.Instance`.
+  - `$color-splitter` and `theme.colorSplitter` has been replaced with `$color-secondary` and `theme.colorSecondary`
```
na9da, line 12: This change means we loose the current orangeish color for splitter controls right? Is that intended?

lib/ReactViews/Workbench/Controls/ViewingControls.jsx
```diff
@@ -403,7 +407,7 @@ const ViewingControls = observer(
       const classList = {
         [Styles.noZoom]: !canZoom,
         [Styles.noSplit]: !canSplit,
-        [Styles.noInfo]: !item.showsInfo
```
na9da, line 406:
```js
const canSplit =
        !item.terria.configParameters.disableSplitter &&
        item.supportsSplitting &&
        defined(item.splitDirection) &&
        item.terria.currentViewer.canShowSplitter;
```

We should change the `supportsSplitting` to `disableSplitter` or remove the entire `classList`. I don't think it has any effect any more since the classes are not defined in the `.scss` style file.

na9da (approved):
Hi @nf-s,

I fully endorse the clean up :-)

CHANGES.md
nf-s: Yep - i've made `colorSecondary` orange

lib/ReactViews/Workbench/Controls/ViewingControls.jsx
nf-s: Nice find, I've removed all SCSS

PR #5664, opened 2021-07-22
Fix docs generation

zoran995:
### What this PR does

When moving Traits classes to a new folder I skipped to update `generateDocs.js` which result in the failure of docs script. This just fixes the import location

### Checklist


na9da (approved):
Thanks @zoran995 !

PR #5684, opened 2021-07-28
Wc disable popup for story shares

wibily:
### What this PR does

- hides welcome popup for story shares
- fixes view stories tab in mobile menu after closing popup

### Checklist


lib/Models/Terria.ts
```diff
@@ -1663,6 +1663,7 @@ async function interpretHash(
       } else if (property === "hideWelcomeMessage") {
         terria.configParameters.showWelcomeMessage = false;
       } else if (property === "start") {
+        terria.configParameters.showWelcomeMessage = false;
         try {
           // a share link that hasn't been shortened: JSON embedded in URL (only works for small quantities of JSON)
           const startData = JSON.parse(propertyValue);
```
nf-s: This will only work with share links which haven't been shortened - eg JSON embedded in the URL.

One method to solve this would be to check if there are any stories after `loadInitSources()` has been called. 

For example - this could go at the bottom of `terria.forceLoadInitSources()`

https://github.com/TerriaJS/terriajs/blob/051f94ea59c51feb74e76fc23995576b9480d895/lib/Models/Terria.ts#L999-L1008

Another approach could be to look through the share JSON data here in `interpretHash()` - to see if stories exist

https://github.com/TerriaJS/terriajs/blob/051f94ea59c51feb74e76fc23995576b9480d895/lib/Models/Terria.ts#L1713

nf-s (changes requested):

nf-s (approved):

PR #5454, opened 2021-04-27
Error handling ROUND 2 - ding ding ding

nf-s:
Fixes https://github.com/TerriaJS/terriajs/issues/5031

### What this PR does

* Add `TerriaErrorSeverity` enum, values can be `Error` or `Warning`.
  * Errors with severity `Error` are presented to the user. `Warning` will just be printed to console.
  * By default, errors will use `Error`
  * `TerriaErrorSeverity` will be copied through nested `TerriaErrors` on creation (eg if you call `TerriaError.from()` on a `Warning` then the parent error will also be `Warning`)
  * Loading models from share links or stories will use `Warning` if the model is **not in the workbench**, otherwise it will use `Error`.
* In `terriaErrorNotification` - show `error.message` (as well as `error.stack`) if `error.stack` is defined
* `AsyncLoader` loadXXX methods now return `Result` with `errors`.
* `AsyncLoader` now has an observable `result` property.
* `viewState.viewCatalogMember()` now handles loading catalog members, opening groups and showing "Add Data" window.
* Removed `openGroup()` - it is replaced by `viewState.viewCatalogMember`
* Fix `MagdaReference` `forceLoadReference` bug.
* Clean up `CkanCatalogGroup` loading - errors are no-longer swallowed.
* Clean up `3dTilesMixin` loading - errors are no-longer swallowed.
* Fix `DataPreviewSections` info section bug.
* Move `FeedbackForm` `z-index` to same as `Notification` - this is so it will appear above Data catalog.
* Added `result.raiseError()`, `result.pushErrorTo()` and `result.clone()` helper methods - and `Result.combine()` convenience function
* Renamed `ReferenceMixin.is` to `ReferenceMixin.isMixedInto`

### Test URLs

http://ci.terria.io/terria-error-workbench/#clean&https://gist.githubusercontent.com/nf-s/89c9da8fd43e7f61c71255cc881d17e4/raw/3130470b9c9cf82fb5f3d83c38985f2164975db1/error-test.json

This has a few broken models to play with

**Note** you should get an error on startup - this is due to an incorrect trait in the catalog

### Which maps have I tested (locally):

- [x] Natmap beta
- [x] De-aus
- [x] De-africa
- [ ] TerriaCube
- [ ] Kaleidoscope
- [x] Pacificmap
- [x] GEO-RAPP
- [ ] NEII
- [x] QLD DT
- [x] NSW DT
- [ ] VIC DT

#### Share tests

[...]

PR #5691, opened 2021-07-29
Log to external error service

### LINKED ISSUE #5627: Re-enable Rollbar - error tracking service
(or look into Sentry as alternative)

na9da:
### What this PR does

Fixes #5627 

Merge only after #5454 

* Adds a generic error service interface that can be configured through `configParameters`.
* Adds `RollbarErrorServiceProvider`. (by default this logs all uncaught errors and unhandled promise errors).
* Logs to error service in `terria.raiseErrorToUser`.

### Checklist


lib/Models/Terria.ts
```diff
@@ -561,6 +574,7 @@ export default class Terria {
         !terriaError.raisedToUser)
     ) {
       terriaError.raisedToUser = true;
+      this.errorService.error(terriaError);
       this.error.raiseEvent(terriaError);
     } else {
       console.log(terriaError);
```
nf-s: I think we should call `this.errorService.error(terriaError)` regardless of whether the error is shown to the user

na9da: Have changed this to log to service for every call.

na9da:
Thanks for the review @nf-s, I'll merge into `next` once the upstream branch is merged.

nf-s (changes requested):
Hey @na9da 

I was having a play around to see if we could get more useful stack trace in there.

What do you think of this change?

https://github.com/TerriaJS/terriajs/pull/5701

na9da:
Hi @nf-s 

That looks like a much better stack trace. I'll merge it in!

nf-s (approved):

PR #5692, opened 2021-07-29
Set default orientation for GLTFs to zero degrees HPR.

### LINKED ISSUE #5690: Use 0 default rotation for GltfCatalogItem
Currently when no rotation is specified for a GltfCatalogItem, the orientation of the model is wrong as shown in the screenshot below. 

This is because we default to a `Quaternion.IDENTITY` rotation instead of `HPR=(0,0,0)`:

https://github.com/TerriaJS/terriajs/blob/cb56deb620b8268a6c42e23d4ef273e8a65e77f6/lib/Models/GltfCatalogItem.ts#L86-L89

image

na9da:
### What this PR does

Fixes #5690 

Sets the default orientation for `GltfCatalogItem` to 0 wrt ENU so that it is aligned correctly to the ground at its origin location.

### Checklist

### DIFF
--- a/test/Models/GltfCatalogItemSpec.ts
+++ b/test/Models/GltfCatalogItemSpec.ts
@@ -1,10 +1,18 @@
+import HeadingPitchRoll from "terriajs-cesium/Source/Core/HeadingPitchRoll";
 import JulianDate from "terriajs-cesium/Source/Core/JulianDate";
+import CesiumMath from "terriajs-cesium/Source/Core/Math";
+import Matrix3 from "terriajs-cesium/Source/Core/Matrix3";
+import Matrix4 from "terriajs-cesium/Source/Core/Matrix4";
+import Quaternion from "terriajs-cesium/Source/Core/Quaternion";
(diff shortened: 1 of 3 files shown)


test/Models/GltfCatalogItemSpec.ts
```diff
@@ -28,18 +36,27 @@ describe("GltfCatalogItem", function() {
...
-          expect(orientation.x).toEqual(0);
-          expect(orientation.y).toEqual(0);
-          expect(orientation.z).toEqual(0);
-          expect(orientation.w).toEqual(1);
+        if (orientation !== undefined) {
+          const { heading, pitch, roll } = getEntityHprInDegrees(entity);
+          expect(heading).toEqual(0);
+          expect(pitch).toEqual(0);
+          expect(roll).toEqual(0);
         }
```
na9da, line 60: Refactored the test so that we directly check the HPR values in degrees.

KeyboardSounds (approved):
Thanks Nanda!

PR #5702, opened 2021-08-02
Allow translation of some components

ykiu:
### What this PR does

This patch ports some hardcoded strings into the i18n message registry.

### Checklist

    - Unit tests aren't applicable because the changes are only about translation. I manually tested the affected components renders as before (in English :-) ).


wibily (approved):
thank you for contributing this

ykiu:
Thank you for your review @wibily !

PR #5705, opened 2021-08-03
Make CompositeCatalogItem sync visibility to its members

ykiu:
### What this PR does

The show/hide checkbox on composite catalog items was not working. This PR makes it work.

Below is a comparison of the behavior before and after the fix.

| Before | After |
| ------ | ----- |
| composite-before | composite-after |

The catalog definition used for the comparison:

```json
    {
      "name": "Composite",
      "type": "composite",
      "members": [
        {
          "type": "csv",
          "url": "test/incidents_notime.csv",
          "name": "CSV Test"
        },
        {
          "type": "csv",
          "url": "test/NSW_LGA_NEXIS_201212.csv",
          "name": "NSW LGA NEXIS 2012"
        }
      ]
    }
```

### Checklist


nf-s (approved):
Hi @ykiu 

Thanks a lot for the PR! 😄 
Works perfectly!

PR #5706, opened 2021-08-03
Make useTranslationIfExists take an i18n object as a parameter

### LINKED ISSUE #6352: Fix naming of map tools bug appearing as translation strings
image

Related to #6340  to fix map attribution labels.

tephenavies:
### What this PR does

Change useTranslationIfExists to ensure that React components that use it to render translation strings use a i18next-react method of re-rendering when the language is changed

And add unit tests for useTranslationIfExists

Fixes #6352 #6336

### Checklist


lib/Core/injectTerms.ts
```diff
@@ -15,7 +15,7 @@ const findFirstTerm = (
   terms.forEach((_, term) => {
     const foundIndex = text
       .toLowerCase()
-      .indexOf(useTranslationIfExists(term).toLowerCase(), fromIndex);
+      .indexOf(useTranslationIfExists(term, i18next).toLowerCase(), fromIndex);
```
tephenavies: Probably should document somewhere that this file should also be improved to not import global i18next

lib/Language/languageHelpers.ts
```diff
@@ -1,15 +1,25 @@
...
+ * @param keyOrString Either a prefixed translation key or a string
+ * @param i18n An i18next instance
  */
-export function useTranslationIfExists(keyOrString: string) {
-  if (keyOrString && keyOrString.indexOf("translate#") === 0) {
-    const translationKey = keyOrString.substr("translate#".length);
-    return i18next.exists(translationKey)
-      ? i18next.t(translationKey)
+export function useTranslationIfExists(keyOrString: string, i18n: i18n) {
+  if (keyOrString.indexOf(TRANSLATE_KEY_PREFIX) === 0) {
```
zoran995: can it happen that `keyOrString` be undefined (specified by user)?

tephenavies: It shouldn't, since the type signature doesn't include `undefined`, but the function was being called wrongly by a .jsx file (`lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx `) with a parameter that was sometimes undefined. That's also been fixed in this PR

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
```diff
@@ -113,10 +117,14 @@ class HelpPanelItem extends React.Component {
...
-          placeholderImage={useTranslationIfExists(
-            this.props.content.placeholderImage
-          )}
+          videoUrl={
+            this.props.content.videoUrl &&
+            useTranslationIfExists(this.props.content.videoUrl, i18n)
+          }
+          placeholderImage={
+            this.props.content.videoUrl &&
+            useTranslationIfExists(this.props.content.placeholderImage, i18n)
```
zoran995: this should actually check if `placeholderImage` is defined

lib/Language/languageHelpers.ts
zoran995: I just run a check on `trainerItems` if the title is undefined it will throw an error although it is typescript file
```json5
"trainerItems": [
  {
    //"title": "stepbystepguide.gettingstarted.title", // this will throw rror
    "footnote": "stepbystepguide.gettingstarted.footnote",
    "steps": [
      {
        //"title": "stepbystepguide.gettingstarted.step1.title", // this will throw error when opening steps guide
        "markdownDescription": "stepbystepguide.gettingstarted.step1.markdownDescription"
      }
    ]
  }
]
```

lib/Core/injectTerms.ts
staffordsmith83: Added a TODO comment, but could be resolved before merging if easy?

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
staffordsmith83: Done

lib/Language/languageHelpers.ts
staffordsmith83: @steve9164 which way do you want to go on this?

nf-s (changes requested):
Got an error when using `Tour` functionality

image

Steps:
- http://ci.terria.io/utif-i18n-parm/
- Change language to French
- Go to step by step guide
  - image
- Press "commence"
  - image

lib/Language/languageHelpers.ts
tephenavies: Probably there's an implicit `any` in the code that deals with `trainerItems` - anything that deals with JSON. You could fix it by explicitly checking existence of each property of the JSON object. Using `unknown` helps out here - e.g. `fetchJson(...) as unknown` and then `if (o.title) {...}`

tephenavies:
The error occurred because a couple of calls to `applyTranslationIfExists` didn't have the i18n parameter. This should now be fixed for all uses of the function.

lib/Language/languageHelpers.ts
staffordsmith83: Resolved:
```ts
export function applyTranslationIfExists(
  keyOrString: unknown,
  i18n: i18n,
  options?: { [key: string]: string }
): string {
  // keyOrString could be undefined in some cases even if we type it as string
  if (isJsonString(keyOrString)) {
    if (keyOrString.indexOf(TRANSLATE_KEY_PREFIX) === 0) {
      const translationKey = keyOrString.substr(TRANSLATE_KEY_PREFIX.length);
      return i18n.exists(translationKey)
        ? i18n.t(translationKey, options)
        : translationKey;
    } else {
      return keyOrString;
    }
  }
  // Return an empty string if keyOrString was undefined. All functions that call this require a string to be returned
  return "";
}
```

Or should I still require keyOrString to be a string so TS can check what we pass it in most cases?

staffordsmith83:
@nf-s think this is OK? https://github.com/TerriaJS/terriajs/pull/5706#discussion_r909294321

lib/Language/languageHelpers.ts
staffordsmith83: Perhaps this is better:

```ts
export function applyTranslationIfExists(
  keyOrString: string,
  i18n: i18n,
  options?: { [key: string]: string }
): string {
  // keyOrString could be undefined in some cases even if we type it as string
  if (isJsonString(keyOrString as unknown)) {
    if (keyOrString.indexOf(TRANSLATE_KEY_PREFIX) === 0) {
      const translationKey = keyOrString.substr(TRANSLATE_KEY_PREFIX.length);
      return i18n.exists(translationKey)
        ? i18n.t(translationKey, options)
        : translationKey;
    } else {
      return keyOrString;
    }
  }
  // Return an empty string if keyOrString was undefined. All functions that call this require a string to be returned
  return "";
}
```

staffordsmith83: Resolved

staffordsmith83 (approved):

staffordsmith83 (approved):

staffordsmith83 (approved):

staffordsmith83 (approved):

staffordsmith83:
Oops on those repeated approvals, process got a bit mixed up with crispy initiating the PR and me taking over! Waiting for @nf-s to review final changes

lib/Language/languageHelpers.ts
```diff
@@ -1,18 +1,35 @@
...
-  if (keyOrString.indexOf("translate#") === 0) {
-    const translationKey = keyOrString.substr("translate#".length);
-    return i18next.exists(translationKey)
-      ? i18next.t(translationKey, options)
-      : translationKey;
-  } else {
-    return keyOrString;
+): string {
+  // keyOrString could be undefined in some cases even if we type it as string
+  if (isJsonString(keyOrString as unknown)) {
```
nf-s, line 23: I would type `keyOrString` as `unknown` if you're using `isJsonString`

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
```diff
@@ -21,20 +21,35 @@ class HelpPanelItem extends React.Component {
...
+    // const MenuIconWrapper = styled(Box).attrs({
+    //   centered: true
+    // })`
+    //   flex-shrink: 0;
+    //   width: 64px;
+    //   height: 64px;
+    //   display: table-cell;
+    //   vertical-align: middle;
+    //   padding-left: 25px;
+    // `;
```
nf-s: Can these comments be removed?

lib/ReactViews/Map/TrainerBar/TrainerBar.tsx
```diff
@@ -405,7 +408,7 @@ export const TrainerBar = observer((props: TrainerBarProps) => {
           }
           selectedTrainer={selectedTrainerItem}
           theme={theme}
-          t={t}
+          // t={t}  // TODO: Can this be safely removed?
```
nf-s: This can be removed

lib/ReactViews/Tools/Tool.tsx
```diff
@@ -69,16 +69,16 @@ export class ToolButtonController extends MapNavigationItemController {
...
+
+  // TODO: The intention was for this to be translated in the getter. This is now removed as it was an incorrect use of useTranslationIfExists()
   @computed
   get title() {
     const buttonState = this.active ? "open" : "closed";
-    return useTranslationIfExists(`tool.button.${buttonState}`, {
-      toolName: this.name,
-      toolNameLowerCase: this.name.toLowerCase()
-    });
+    return `tool.button.${buttonState}`;
```
nf-s: Instead of removing the `applyTranslationifExists` - you can import global `i18next` and use `applyTranslationIfExists(this.props.toolName, i18next)`

nf-s (changes requested):

lib/Language/languageHelpers.ts
zoran995: @nf-s I am more into this change as it will force us to use string type while and non-string value can come only from config and in the future values in config will be part of the traits system and validated as a string, just my thoughts on this

lib/ReactViews/Map/TrainerBar/TrainerBar.tsx
staffordsmith83: Done

lib/ReactViews/Tools/Tool.tsx
staffordsmith83: @nf-s the issue here is that applyTranslationifExists will only translate anything that has "translate#" at the start of the string otherwise it will just return the passed in value. So its not using applyTranslationIfExists in the right way. Should I still leave it in if its doing nothing?

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
staffordsmith83: Yes, not needed any more, resolved.

lib/Language/languageHelpers.ts
staffordsmith83: Yes that what I was thinking @zoran995, if you see my previous commit I typed it as unknown, but then thought better to type as string so we are forced to pass a string in ost occurrences, and if a non-string value sneaks through we handle it inside `applyTranslationIfExists()`. Happy to defer to you @nf-s but do you think that intention is valid? Perhaps my type assertion in `if (isJsonString(keyOrString as unknown))` is not the correct way to do things.

nf-s (approved):

lib/ReactViews/Tools/Tool.tsx
nf-s: Ok, I wasn't aware that tool name/title couldn't get string which started with `translate#`.

lib/Language/languageHelpers.ts
nf-s: Ok happy to leave it then

lib/ReactViews/Tools/Tool.tsx
staffordsmith83: Nah you might be right @nf-s  Ill just check with @na9da about his intention for this part he had some ideas I think he wrote initially

lib/Language/languageHelpers.ts
na9da: In my latest debugging, I think `name` and `title` doesn't get used anywhere. Will clean it up sometime - https://github.com/TerriaJS/terriajs/issues/6368

lib/ReactViews/Tools/Tool.tsx
na9da: In my latest debugging, I think `name` and `title` doesn't get used anywhere. Will clean it up sometime - https://github.com/TerriaJS/terriajs/issues/6368

na9da:
I have done a few tests to see if it impacts other tools like `DiffTool`, but they seem to be working fine.

PR #5711, opened 2021-08-08
Docs generation for v8 model layer

nf-s:
### Checklist


tephenavies:
Auto generated pages shouldn't be put into `doc/` and committed to the repo. Our doc publishing site (netlify) doesn't commit the results back to the repo so the generated pages will quickly get out of date and sometimes pollute other diffs. I think the only exception to this that we have at the moment is the attribution page, which I'm not sure what to do about. I think we need it to be committed, so at the moment it gets committed periodically.

tephenavies:
Maybe you didn't run generateDocs with cwd = `build/`?

nf-s:
Replaced by https://github.com/TerriaJS/terriajs/pull/5713

PR #5713, opened 2021-08-09
Docs generation for model layer v8 (version 2)

nf-s:
### Docs generation for model layer v8 (version 2)

Fixes https://github.com/terriajs/terriajs/issues/5637

* Add `description` and `example` static properties to `Trait`, and added `@traitClass` decorator.
* Add `parent` property to `Trait`, which contains parent `TraitClass`.
* New model-generated documentation in `generateDocs.ts`
* Refactored some `Traits` classes so they use `mixTraits` instead of extending other `Traits` classes.

### Checklist


lib/Traits/TraitsClasses/WebMapServiceCatalogItemTraits.ts
```diff
@@ -136,17 +137,17 @@ export class WebMapServiceAvailableLayerDimensionsTraits extends ModelTraits {
...
+
+<strong>Note:</strong> <i>To present all layers in an available WMS as individual items in the catalog use the \`WebMapServiceCatalogGroup\`.</i>`,
+  example: {
+    type: "wms",
+    name: "Mangrove Cover",
+    url: "https://ows.services.dea.ga.gov.au",
+    layers: "mangrove_cover_v2_0_2"
+  }
+})
 export default class WebMapServiceCatalogItemTraits extends mixTraits(
```
nf-s, line 151: An example of setting `description` and `example` for a Traits class

package.json
```diff
@@ -161,7 +161,6 @@
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@types/dateformat": "^3.0.1",
     "babel-plugin-styled-components": "^1.10.7",
-    "documentation": "12.3.0",
```
tephenavies, line 164: What did you do about the examples and comments that documentation used to pull out? (If it was even working).

I still aim to have examples when we get time to make a better system for examples, comments, etc. but I think instead of parsing code it'll

nf-s: I moved them to the new `@traitsClass()` decorator
See https://github.com/TerriaJS/terriajs/pull/5713/files#discussion_r684918251

The JSDoc stuff wasn't working for me

tephenavies: Yeah JSDoc doesn't like some ES2017 syntax, or something I can't remember what.

tephenavies (approved):

zoran995:
can we just link  translation guides to doc as part of this PR or to create a separate PR

PR #5724, opened 2021-08-11
Feature info template for composite catalog items

ykiu:
### What this PR does

This PR enables to use featureInfoTemplate in conjunction with CompositeCatalogItem.

Previously, there was no way to set featureInfoTemplate on members of a composite catalog item. This PR fixes the problem.

Say we have the following catalog definition:

```json
{
    "name": "Composite",
    "type": "composite",
    "members": [
        {
            "type": "csv",
            "url": "test/incidents_notime.csv",
            "name": "CSV Test",
            "featureInfoTemplate": {
                "template": "here comes the template"
            }
        }
    ]
}
```

In this example, when one clicks on a feature from the CSV dataset, the default attribute table is rendered even though the member item ("CSV Test") has a template.

This was because when FeatureInfoPanel determines the template to use, it does not look into the individual members of a composite model and thinks that the model does not have a template.

In this PR, I tweaked FeatureInfoPanel so that when it sees a composite model, it looks into the members of the model and uses an appropriate template if found one.

To test the changes, I turned the test code for FeatureInfoPanel into TypeScript and added two tests for the determineCatalogItem() function. The first one is a basic test for ordinary catalog items. The other one is for composite catalog items.

### Checklist


tephenavies (approved):

tephenavies:
Thanks for updating those tests for v8!

ykiu:
@steve9164 Thank you!

Issue #5281, opened 2021-02-25
FeatureInfo still displayed for hidded layers (WMS minScaleDenominator + hideLayerAfterMinScaleDenominator)


geoffrowland:
For a WMS Catalog Item, if minScaleDenominator is set and hideLayerAfterMinScaleDenominator is true, tiles are hidden after zooming in past minScaleDenominator. 

However, FeatureInfo from the hidden layer is still displayed from clicking on the map. The expected behaviour is that a hidden map layer would not generate a FeatureInfo response.
In WebMapServiceCatalogItem.js, perhaps enablePickFeatures should be set as false in these circumstances

PR #5735, opened 2021-08-14
Add maxScaleDenominator to WMS and MapServer

### LINKED ISSUE #3972: Web Map Service (WMS) (item): MaxScaleDenominator
Web Map Service (WMS) (item) should support WMS ver 1.3 MaxScaleDenominator, in much the same way as MInScaleDenominator is already supported.

So, WebMapServiceCatalogItem.js should include **maxScaleDenominator** and the corresponding **hideLayerBeforeMaxScaleDenominator**, similar to the existing **minScaleDenominator** and **hideLayerAfterMinScaleDenominator**.

Also, I have noticed that when a layer is hidden by hideLayerAfterMinScaleDenominator/minScaleDenominator, featureInfo picking for the hidden layer still works. I think expected behaviour should be that you don't pick information from a layer you can't see.

zoran995:
### What this PR does

Fixes #3972, #5281

Added support for maxScaleDenominator
- moved entire logic from catalog items to Mixin (`MinMaxLevelMixin`) and Traits (`MinMaxLevelTraits`) level
- extracted function scaleDenominatorToLevel for conversion between scale and zoom level
- **Don't set `minimumLevel` on imageryProvider as it will cause performance issues at globe view (requesting images for at least specified zoom level)**
- Changed from raising error to user to a message in workbench, since error doesn't say what user can do.

### Checklist


zoran995:
we are already using this in our version in production (dataset is private). For testing you can use https://gist.githubusercontent.com/zoran995/2e56bc27d6363cda83e0cd3bd8f042b0/raw/630c782147ce842c9dbe9fcd070dbfee5b5e1bbc/min-max.json

tephenavies (approved):

PR #5757, opened 2021-08-23
Fix model update

### LINKED ISSUE #4119: Story v2.0 - Allow user to "reset" the scene after moving/adding/removing things
When we are in a Story and the user has changed the view, added, or removed layers etc, allow the user to click on the currently selected Scene "dot" to reset the scene to what it is in the story. I.e. clicking on the dot will reset the view, the loaded layers, with the very same effect as if they had clicked Next and then Previous to return to the scene they are on.

na9da:
### What this PR does

Fixes #4119 and also will improve performance when switching between story scenes.

Changes:
- Wraps `AsyncLoader.load()` in action so that mobx correctly caches load item promise and `mapItems`.
-  Fully resets a model state when switching story scenes

### Testing - model state reset
- Open this `main` branch [story](http://ci.terria.io/main/#share=s-uHFdujNkQUq9wukMQVnZdQd40cf)
- Switch story scenes from `A -> B -> A`
- Observe that the style is not reset back to True color when switching back to scene `A` from `B`.
- Now repeat the same for this [branch](http://ci.terria.io/fix-model-update/#share=s-uHFdujNkQUq9wukMQVnZdQd40cf).
- Observe that the style is reset back to True color.

### Checklist

### DIFF
--- a/lib/ReactViews/Story/StoryPanel.jsx
+++ b/lib/ReactViews/Story/StoryPanel.jsx
@@ -37,7 +37,7 @@ export async function activateStory(scene, terria) {
         try {
           await terria.applyInitData({
             initData: initSource,
-            replaceStratum: false,
+            replaceStratum: true,
             canUnsetFeaturePickingState: true
           });
         } catch (e) {
--- a/lib/ReactViews/Story/StoryPanel.jsx
+++ b/lib/ReactViews/Story/StoryPanel.jsx
@@ -159,14 +159,7 @@ const StoryPanel = observer(
     },
 
     onCenterScene(story) {
-      if (story.shareData) {
-        this.props.terria
-          .updateFromStartData(
-            story.shareData,
-            `Story data: \`${story.title ?? story.id}\``
-          )
-          .raiseError(this.props.terria);
-      }
+      activateStory(story, this.props.terria);
     },
 
     goToPrevStory() {
--- a/lib/Traits/Decorators/anyTrait.ts
+++ b/lib/Traits/Decorators/anyTrait.ts
@@ -1,3 +1,4 @@
+import { computed } from "mobx";
 import Result from "../../Core/Result";
 import { BaseModel } from "../../Models/Definition/Model";
 import Trait, { TraitOptions } from "../Trait";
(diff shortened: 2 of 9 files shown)


lib/ReactViews/Story/StoryPanel.jsx
```diff
@@ -159,14 +159,7 @@ const StoryPanel = observer(
...
-      if (story.shareData) {
-        this.props.terria
-          .updateFromStartData(
-            story.shareData,
-            `Story data: \`${story.title ?? story.id}\``
-          )
-          .raiseError(this.props.terria);
-      }
+      activateStory(story, this.props.terria);
     },
```
na9da, line 163: Can't think of a reason why we shouldn't be using `activateStory` instead of `updateFromStartData`. Calling `updateFromStartData` has a few problems:
1. Every time you "center the story" it adds a duplicate story share data in `Terria.initSources`
2. There is no way to pass `replaceStratum` so that we can get a clean state when resetting the scene.

When looking at the history of `onCenterScene` in `v7` branch, it looks like to me that both `activateStory` and `onCenterScene` used to call `updateFromStartData` then changes were made to activateStory but not to onCenterScene, which could be an error.

lib/Traits/Decorators/anyTrait.ts
```diff
@@ -19,6 +20,8 @@ export default function anyTrait(options: TraitOptions) {
 }
 
 export class AnyTrait extends Trait {
+  readonly decoratorForFlattened = computed.struct;
+
```
na9da, line 24: Cesium3DTilesCatalogItem uses anyTrait for specifying styles. Using structural comparison here will prevent unnecessary recomputes of `mapItems` when style object changes without actual values changing (eg when stratum is recreated during story scene change)

nf-s (changes requested):
Works perfectly for me!
Can you please just add a CHANGES entry?

na9da:
> Can you please just add a CHANGES entry?

Forgot to push. Done now.

nf-s (approved):

Issue #5759, opened 2021-08-24
Add support for custom glyphs in point chart


na9da:
Currently point charts can only show points as filled circles. Extend this to support glyphs provided by the [@visx/glyphs](https://airbnb.io/visx/glyphs) package.

PR #5760, opened 2021-08-24
Add chart glyph styles.

### LINKED ISSUE #5758: Use @visx instead of @vx and update chart dependencies
@vx has been moved to [@visx](https://www.npmjs.com/org/visx?tab=packages).

Upgrade the dependencies and test.

na9da:
### What this PR does

Fixes #5758 #5759 
Also required for https://github.com/terriajs/de-australia-map/issues/72

Changes:
- Changed @vx/* dependencies to @visx/* which is the new home of the chart library
- Added support to change glyph styles for charts with points
- Fixes required for https://github.com/terriajs/de-australia-map/issues/72

### Testing
- Open this share [link](http://ci.terria.io/chart-glyphs/#share=s-i7hq5sJJaQbo51BUs3TCli4zlQt&clean&https://gist.githubusercontent.com/na9da/657916be03e1541629f2018501e80a4d/raw/95cfe3ce7ee4deb8c8fb730fecb28f25aaf04be5/waterbodies.json).
- You should see a chart with different glyphs for different types of points

### Checklist


lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
```diff
@@ -31,36 +33,37 @@ class MomentPointsChart extends React.Component {
...
+      const interpolatedPoints = chartItem.points.map(p => ({
+        ...p,
+        ...interpolate(p, basisItem.points, basisToSourceScale)
+      }));
       return interpolatedPoints;
     }
     return chartItem.points;
   }
 
   doZoom(scales) {
```
wibily, line 45: not part of this PR but we may want to consider better typing of this function either through enforcing it to exist through an interface or through typescript. Right now we kinda expect it to just be there in <LineAndPointChart />

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
```diff
@@ -31,36 +33,37 @@ class MomentPointsChart extends React.Component {
...
-    const points = this.props.chartItem.points;
-    if (points.length === 0) return;
-    const n = points.length;
-    const spacing = (scales.x(points[n - 1].x) - scales.x(points[0].x)) / n;
-    const dots = document.querySelectorAll(`g#${this.props.id} circle`);
-    dots.forEach((dot, i) => {
+    const points = this.points;
+    if (points.length === 0) {
```
wibily, line 47: could points be undefined? ie if the chartItem from props didn't have a points key?

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
na9da: Valid point. I would like to simplify some of the chart components at some point and also convert them to `tsx`.

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
na9da: `chartItem.points` is [defined here](https://github.com/TerriaJS/terriajs/blob/9e6a18f8c656701f2cad8d550d32da468f8eccc0/lib/ModelMixins/ChartableMixin.ts#L60) as array. And because all catalog items are `.ts` we can be confident that this will always be defined.

test/ReactViews/Custom/Chart/BottomDockChartSpec.tsx
```diff
@@ -51,7 +51,9 @@ describe("BottomDockChart", function() {
     ];
   });
 
-  it("renders all points on map for active chart items", function() {
+  // xit because the new version of `withParentSize` from `@vx/responsive` uses ResizeObserver to
+  // trigger render which doesn't seem to work correctly in tests
+  xit("renders all points on map for active chart items", function() {
```
wibily: instead of excluding it, it may be better just to delete it

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
```diff
@@ -31,36 +33,37 @@ class MomentPointsChart extends React.Component {
...
+      return;
+    }
+    const glyphs = document.querySelectorAll(
```
wibily, line 50: this bit seems weird to me in that we are reaching into the DOM to modify things instead of using React but i recognise that this pattern is used by all the different charting components and will be a big task to migrate over

wibily (approved):

lib/ReactViews/Custom/Chart/MomentPointsChart.jsx
na9da: The reason for modifying DOM directly is performance. Because we chart many thousands of points, any updates to those points (like when you zoom the chart) results in react destroying and re-creating DOM elements for those points. This is painfully slow. Instead what we do here is after react has initially rendered those (svg) DOM elements and when we have to update the position of those points, we just change the attributes of the existing DOM elements. There is a drastic improvement in performance when we do that. This is one of those special cases were we have to use the escape hatch.

test/ReactViews/Custom/Chart/BottomDockChartSpec.tsx
na9da: It is a useful test to have, so I have just commented it out now.

PR #5764, opened 2021-08-25
Difftool fixes

na9da:
### What this PR does

- Uses `null` instead of `undefined` to unset `currentTime` from DiffTool.
- Fixes `CzmlCatalogItem` to react to input data changes. Previously this resulted in marker location not updating on clicking a different location when using difftool.

### Testing

- Open this share [link](http://ci.terria.io/difftool-fixes/#clean&share=s-z1VBtkLJqLxdRg0Mml6jYI73Jtu&https://gist.githubusercontent.com/na9da/131cb6308113ec3fc3e4a1211e2ecca3/raw/1c6e8f5c29db3afb970978a3cde3cedfc0577c31/sentinel.json).
- Open the "Difference" tool from the context menu for the first item.
- Try to generate a difference image
- Everything should work.

### Checklist


lib/Models/Catalog/CatalogItems/CzmlCatalogItem.ts
```diff
@@ -89,34 +89,35 @@ export default class CzmlCatalogItem
...
+        title: i18next.t("models.czml.unableToLoadItemTitle"),
+        message: i18next.t("models.czml.unableToLoadItemMessage")
+      });
+    }
+
+    return makeRealPromise<CzmlDataSource>(
+      CzmlDataSource.load(loadableData, {
+        credit: attribution
+      })
+    )
```
na9da, line 117: Moved the checks to top level so that `mobx` correctly registers the dependency on the values.

lib/ModelMixins/DiscretelyTimeVaryingMixin.ts
```diff
@@ -38,7 +38,7 @@ function DiscretelyTimeVaryingMixin<
     @computed
     get currentTime(): string | undefined {
       const time = super.currentTime;
-      if (time === undefined) {
+      if (time === undefined || time === null) {
```
wibily, line 41: this is just a suggestion, but would it be possible to do something like
```
export const MISSING = undefined;
export const UNSET = null;
```
in Model.ts or some other more 'global' file?
that way this line can read as
`if(time === Model.MISSING || time === Model.UNSET)`

lib/Traits/TraitsClasses/TimeVaryingTraits.ts
```diff
@@ -8,7 +8,7 @@ export default class TimeVaryingTraits extends mixTraits(CatalogMemberTraits) {
     description: "The current time at which to show this dataset.",
     type: "string"
   })
-  currentTime?: string;
+  currentTime?: string | null;
```
wibily, line 12: does this trait need to set the isNullable field to true? Also, why do we have a separate isNullable field in the trait system?

wibily (approved):

lib/ModelMixins/DiscretelyTimeVaryingMixin.ts
na9da: I could do that but I think because `undefined` and `null` or so ubiquitous no one might use it? ~But it gives me another idea, I can maybe define a method called `unsetTrait` which does the right thing by setting the value to `null`.~ It is hard to add a generic method that does that because not all trait values accept `null`. I don't have the type-fu to express that atleast.

The following doesn't type check.

```ts
    unsetTrait<Key extends keyof StratumTraits>(
      stratumId: string,
      trait: StratumTraits[Key] extends null ? Key : never
    ): void {
      this.setTrait(stratumId, trait, null);
    }
```

lib/Traits/TraitsClasses/TimeVaryingTraits.ts
na9da: You are right! 

I think `isNullable` is used only during [de-serialization](https://github.com/TerriaJS/terriajs/blob/0e5ff1bd512597808274ab58805d5d1ce6b75acf/lib/Traits/Decorators/primitiveTrait.ts#L55-L65) of a model from json. And that is because we don't have the type information during runtime.

There is low chance that a json serialized model will have `null` as `currentTime`, but I'll enable that anyways.

PR #5774, opened 2021-08-30
Catalog Index

### LINKED ISSUE #5592: Improve item search

nf-s:
### Catalog indexer experiment

Fixes #5592 https://github.com/terriajs/nationalmap/issues/1048

### This depends on this small terriajs-cesium PR 

This change is only required for catalog index generation - not for using catalog index client-side

- [ ] https://github.com/TerriaJS/cesium/pull/82

### Why

Searching through many dynamic groups is slow - because we have to load each group to find group models.

This method uses a new node.js script to load all groups and references, and then save a fully resolved catalog tree (with model `id`, `name` and `description`) into JSON file which can then be indexed.

### How

Read docs in `client-side-config.md` https://github.com/TerriaJS/terriajs/blob/fc8377f69327f49980048f8d302d78381d43ff5e/doc/customizing/client-side-config.md#catalogindex

### Changes

* Add `CatalogIndex`, `CatalogIndexReference` and `generateCatalogIndex()` script. These can be used to generate a static JSON in dex of a terria catalog - which can then be searched through using `flexsearch`
* Added `weakReference` flag `ReferenceMixin`, this can be used to treat References more like a shortcut (this means that `sourceReference` isn't used when models are shared/added to the workbench - the `target` is used instead)
* GroupMixin.isMixedInto and MappableMixin.isMixedInto are now more strict - and won't pass for for References with `isMappable` or `isGroup`.
* `Workbench.add` can now handle nested `References` (eg `CatalogIndexReference -> CkanReference -> WMSCatalogItem`).
* Add `description` trait to `CatalogMemberReferenceTraits`

### Test links

National map with new index - http://ci.terria.io/catalog-index-experiment/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/nationalmap/map-config/test.json

National map without index - http://ci.terria.io/catalog-index-experiment/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/nationalmap/map-config/prod.json

Pacific map with new index - http://ci.terria.io/catalog-index-experiment/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/dev.json

### Checklist


nf-s:
## FIXME

image

lib/ModelMixins/GroupMixin.ts
```diff
@@ -270,7 +270,13 @@ namespace GroupMixin {
...
     extends InstanceType<ReturnType<typeof GroupMixin>> {}
   export function isMixedInto(model: any): model is Instance {
-    return model && "isGroup" in model && model.isGroup;
+    return (
+      model &&
+      "isGroup" in model &&
+      model.isGroup &&
+      "forceLoadMembers" in model &&
+      typeof model.forceLoadMembers === "function"
+    );
```
nf-s, line 316: This change is required, as `ReferenceTraits` includes `isGroup` property

lib/ModelMixins/MappableMixin.ts
```diff
@@ -191,7 +191,12 @@ namespace MappableMixin {
   export interface Instance
     extends InstanceType<ReturnType<typeof MappableMixin>> {}
   export function isMixedInto(model: any): model is Instance {
-    return model && model.isMappable;
+    return (
+      model &&
+      model.isMappable &&
+      "forceLoadMapItems" in model &&
+      typeof model.forceLoadMapItems === "function"
+    );
```
nf-s, line 196: This change is required, as `ReferenceTraits` includes `isMappable ` property

lib/Traits/TraitsClasses/ReferenceTraits.ts
```diff
@@ -0,0 +1,11 @@
...
+import primitiveTrait from "../Decorators/primitiveTrait";
+
+export default class ReferenceTraits extends ModelTraits {
+  @primitiveTrait({
+    type: "boolean",
+    name: "Is catalog item open in workbench",
+    description: "Whether the item in the workbench open or collapsed."
+  })
+  isOpenInWorkbench?: boolean = true;
+}
```
nf-s, line 11: This is needed so we can see loading spinner in the workbench, while a reference is loading

doc/customizing/client-side-config.md
```diff
@@ -174,3 +175,28 @@ Configuration of items to appear in the search bar
...
+### CatalogIndex
+
+If your TerriaMap has many (>50) dynamic groups (groups which need to be loaded - for example CKAN, WMS-group...) it may be worth generating a static catalog index JSON file. This file will contain ID, name and description fields of all catalog items, which can be used to search through the catalog very quickly without needing to load dynamic groups.
+
+The https://github.com/nextapps-de/flexsearch library is used to index and search the catalog index file.
+
+To generate the catalog index:
+
+- `npm run build-tools`
+- `node .\build\generateCatalogIndex.js config-url base-url` where
```
na9da, line 193: It failed to run when using node 10.x (which is outdated now I guess) due to `TextEncoder` not being defined by `jsdom`. But would be good to leave a note to use at least 12.x.

na9da:
So one metric of `flexsearch` i'm interested in is the serialized size of its index. I think `minisearch` uses a clever suffix tree like data structure that greatly reduces the size of the index thus reducing the overall size of the file that has to be downloaded over the network. This is important in the case of item search. I'll do some tests around that in the future.

doc/customizing/client-side-config.md
nf-s: done!

lib/Models/Catalog/CatalogReferences/CatalogIndexReference.ts
```diff
@@ -0,0 +1,98 @@
+import filterOutUndefined from "../../../Core/filterOutUndefined";
+import flatten from "../../../Core/flatten";
+import TerriaError from "../../../Core/TerriaError";
+import GroupMixin from "../../../ModelMixins/GroupMixin";
+import ReferenceMixin from "../../../ModelMixins/ReferenceMixin";
+import CatalogIndexReferenceTraits from "../../../Traits/TraitsClasses/CatalogIndexReferenceTraits";
+import CreateModel from "../../Definition/CreateModel";
+import { BaseModel } from "../../Definition/Model";
+
```
na9da, line 9: Some top level comment on the purpose of this class might be a good idea.

lib/Models/SearchProviders/CatalogIndex.ts
```diff
@@ -0,0 +1,131 @@
+import { ObservableMap, action, observable } from "mobx";
+import CatalogIndexReference from "../Catalog/CatalogReferences/CatalogIndexReference";
+import loadJson from "../../Core/loadJson";
+import updateModelFromJson from "../Definition/updateModelFromJson";
+import CommonStrata from "../Definition/CommonStrata";
+import Terria from "../Terria";
+import CatalogIndexReferenceTraits from "../../Traits/TraitsClasses/CatalogIndexReferenceTraits";
+import SearchResult from "./SearchResult";
+
+const { Document } = require("flexsearch");
```
na9da: No type definitions available right now?

lib/Models/SearchProviders/CatalogIndex.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  constructor(private readonly terria: Terria, private readonly url: string) {
+    this.loadPromise = this.loadCatalogIndex();
+  }
+
+  @action
+  private async loadCatalogIndex() {
+    // Load catalog index
+    try {
+      const index = (await loadJson(this.url)) as CatalogIndexFile;
```
na9da, line 44: I don't have a preference here but one thing to consider is whether to load this eagerly or only when the user opens search for the first time and maybe document the choice here?

lib/Models/SearchProviders/CatalogIndex.ts
```diff
@@ -0,0 +1,131 @@
...
+  [id: string]: CatalogIndexReferenceTraits;
+}
+
+export interface ModelIndex {
+  name: string;
+  knownContainerUniqueIds: string[];
+}
+
+export default class CatalogIndex {
+  private _models: ObservableMap<string, CatalogIndexReference> | undefined;
```
na9da: Does this have to be an `ObservableMap`?

lib/ReactViewModels/SearchState.ts
```diff
@@ -14,13 +14,13 @@ import CatalogSearchProvider from "../Models/SearchProviders/CatalogSearchProvid
...
-  catalogSearchProvider?: any;
-  locationSearchProviders?: any[];
+  catalogSearchProvider?: CatalogSearchProvider;
+  locationSearchProviders?: SearchProvider[];
 }
 
 export default class SearchState {
   @observable
-  catalogSearchProvider: SearchProvider | undefined;
+  catalogSearchProvider: CatalogSearchProvider | undefined;
```
na9da: Can the type here be the abstract `SearchProvider` so that it leaves the window open for others to write a custom search provider if they want one?

lib/ReactViews/DataCatalog/DataCatalog.jsx
```diff
@@ -59,7 +58,11 @@ export const DataCatalog = observer(
               <DataCatalogMember
                 viewState={this.props.viewState}
                 member={item}
-                manageIsOpenLocally={isSearching}
+                // manage group `isOpen` flag locally if searching through models dynamically (i.e. not using catalog index)
+                // This must be false if usingCatalogIndex - so group references open correctly in the search
+                manageIsOpenLocally={
+                  isSearching && !catalogSearchProvider.usingCatalogIndex
+                }
```
na9da: If `catalogSearchProvider` is a generic `SearchProvider` like in the previous suggestion then this flag name has to be more generic I guess. What's the difference btw?

na9da (approved):
Hey @nf-s, this is a major improvement and it looks great :tada:

I have tested and it is working without errors.
Approving with a few recommendations.

lib/Models/SearchProviders/CatalogIndex.ts
nf-s: They just got published 8 days ago!
https://www.npmjs.com/package/@types/flexsearch

So i added them

lib/Models/SearchProviders/CatalogIndex.ts
nf-s: Added a comment

lib/Models/SearchProviders/CatalogIndex.ts
nf-s: Oh yeah good pickup, no it doesn't - i've turned it into a normal map

lib/ReactViews/DataCatalog/DataCatalog.jsx
nf-s: Done, I added a property to `SearchProvider` called `resultsAreReferences`

lib/ReactViewModels/SearchState.ts
nf-s: Yep done

PR #5775, opened 2021-08-31
Wc help

### LINKED ISSUE #5636: Create UX wireframes for Help menu redesign
Discuss with UX team about reconfiguring Help menu as follows: 

- [ ] remove Related Maps button - to be included as configuration change when releasing NatMap as standalone map https://github.com/TerriaJS/nationalmap/issues/1054 - NOT TO BE DONE PART OF THIS TICKET
- [x] moving Help content from Help button at the bottom right corner under new tab
- [x] include link to End User manual into Help content (https://terriajs.gitbook.io/terriajs/)
- [ ] have default generic Get Started video and workflows under Help page 

note to Phil: set up metrics for GA when we do it

wibily:
### What this PR does

Fixes #5636

### Checklist


wibily:
check with Ana before merge

wibily:
http://ci.terria.io/wc-help/

lib/ReactViews/Map/HelpButton/HelpButton.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+
+import ViewState from "../../../ReactViewModels/ViewState";
+import Icon from "../../../Styled/Icon";
+import Text from "../../../Styled/Text";
+import Prompt from "../../Generic/Prompt";
+
+import Styles from "./help-button.scss";
```
na9da, line 9: I think we should use styled components here instead of SCSS (unless it is not possible due to dependency on some code that is still not ported).

lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
```diff
@@ -118,18 +117,43 @@ class HelpPanel extends React.Component {
...
-              </For>
-            )}
+              </Box>
+            </For>
+          )}
+          <Box styledHeight={"64px"}>
+            <Box styledMargin={"auto 25px"}>
+              <Text semiBold extraLarge textDark uppercase>
+                <a
+                  href="//terriajs.gitbook.io/terriajs/"
```
na9da: I think this should always be `https` and open in a new window perhaps, to avoid user loosing any datasets he has loaded?

na9da (changes requested):
Looks great @wibily. 

Just one suggestion.

lib/ReactViews/Map/HelpButton/HelpButton.tsx
wibily: i'm using help-button.scss in order to extend _buttons.scss and retain the existing styling of the existing buttons. I agree we should use styled components but i felt that de-scssing all the buttons in the app was a bit more work than i was willing to take on :/

na9da (approved):
Looks good @wibily !

PR #5795, opened 2021-09-03
Chart fixes

na9da:
Depends on https://github.com/TerriaJS/terriajs/pull/5760

### What this PR does

- Makes feature info panel chart labels more readable - https://github.com/TerriaJS/de-australia-map/issues/104
- Do not replace expanded charts if they have different titles - https://github.com/TerriaJS/de-australia-map/issues/105
- Fix broken chart data download button (commit https://github.com/TerriaJS/terriajs/pull/5795/commits/d3b48cb83ec9516510effe6d6c990fc74ed5ba64)
- Changed chart tooltip and legend bar implementation so that it can fit more legends gracefully. (https://github.com/TerriaJS/terriajs/pull/5795/commits/120e5269150510982456e241135772a7fb314389 https://github.com/TerriaJS/terriajs/pull/5795/commits/8a2f7a18c9ed543e8a48b1eb9390962a686f3752)
- Improve status message when feature info panel chart is loading
  Previously:
  
image

  Now:

image

### Testing

- Open this [link](http://ci.terria.io/chart-fixes/#share=s-dShjXJ4zG9AUnTm2KAOHFFQVZ09&https://gist.githubusercontent.com/na9da/3c46309baf6cca7be9d301e6348eb75e/raw/e35b8ef8d21d9016a3ec8f23d0df5ab56d940532/coastal.json).
- Click a coastal point to see feature info panel with a chart
- Check out the x-axis labels (there should be enough labels to make the chart readable)
- Expand the chart
- Now click another point and expand chart again
- You should be able to see both charts together (previously only the new chart replaced the existing one).
- Clicking the download button should trigger download of the chart data without erroring.

### Checklist


robbibt:
This now works exactly how I would expect, and the updated x-axis is much more useful and user-friendly. Thanks @na9da!

lib/ReactViews/Custom/Chart/FeatureInfoPanelChart.jsx
```diff
@@ -54,15 +56,30 @@ class FeatureInfoPanelChart extends React.Component {
   }
 
   render() {
-    if (!ChartableMixin.isMixedInto(this.props.item)) return null;
-    if (!this.chartItem) return null;
+    const catalogItem = this.props.item;
+    const height = this.props.height || this.props.parentHeight;
+    const width = this.props.width || this.props.parentWidth;
+    if (!ChartableMixin.isMixedInto(catalogItem)) {
+      return null;
```
wibily: should this silently fail or should this throw an error?

wibily (approved):

lib/ReactViews/Custom/Chart/FeatureInfoPanelChart.jsx
na9da: Good point. I have fixed it now.

lib/ReactViews/Custom/Chart/Legends.jsx
```diff
@@ -21,10 +22,7 @@ export default class Legends extends React.PureComponent {
     });
 
     return (
-      <div
-        className={Styles.legends}
-        style={{ maxHeight: `${Legends.maxHeightPx}px` }}
-      >
+      <Container style={{ maxWidth: this.props.width }}>
```
wibily: is there a reason we aren't using our <Box> component? I'm a bit wary of introducing another general purpose container component

wibily: Furthermore, is there a reason why we are giving this div a max width?

lib/ReactViews/Custom/Chart/Legends.jsx
```diff
@@ -52,14 +50,40 @@ class Legend extends React.PureComponent {
...
+const Container = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+  margin: auto;
+  padding: 7px;
+  font-size: 0.8em;
+`;
+
+const Label = styled.div`
```
wibily: could use a <Text /> component here. An optional thing might do will be to add an aria-label for this element which is               id={sanitizeIdString(chartItem.key + 'label')}, then we can add in an aria-labelledby on the chart itself.

lib/ReactViews/Custom/Chart/Legends.jsx
```diff
@@ -52,14 +50,40 @@ class Legend extends React.PureComponent {
...
+  align-items: center;
+  justify-content: center;
+  min-width: 0px;
+`;
+
+const LabelText = styled.span`
+  margin-left: 4px;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+  overflow: hidden;
```
wibily: oh i added this to <Text/> as overflowEllipsis

lib/ReactViews/Custom/ChartCustomComponent.ts
```diff
@@ -224,7 +224,15 @@ export default abstract class ChartCustomComponent<
...
-          const id = `${context.catalogItem.uniqueId}:${source}`;
+          // When expanding a chart for this item and there is already an
+          // expanded chart for the item, there are 2 possibilities.
+          // 1. Remove it an show the new chart
+          // 2. Show the new chart alongside the existing chart
+          //
+          // If title & source names for the two expanded charts are the same then
+          // we only show the latest one, otherwise we show both.
+          // To do this we make the id dependant on the parentId, title & source.
+          const id = `${context.catalogItem.uniqueId}:${attrs.title}:${source}`;
```
wibily, line 235: is there an example dataset where I can see this?

wibily (approved):

lib/ReactViews/Custom/ChartCustomComponent.ts
na9da: - Example link for behavior 2 i.e when expanding chart from a catalog item and chart title is different from the currently expanded one, create a new chart item - [link](http://ci.terria.io/chart-fixes/#share=s-dShjXJ4zG9AUnTm2KAOHFFQVZ09&https://gist.githubusercontent.com/na9da/3c46309baf6cca7be9d301e6348eb75e/raw/e35b8ef8d21d9016a3ec8f23d0df5ab56d940532/coastal.json)
- Example link for behavior 1 i.e when expanding chart from a catalog item and chart title is the same, replace the currently expanded chart for the item with the new chart. I couldn't locate another example dataset for this, but I'm sure it exists :-)

lib/ReactViews/Custom/Chart/Legends.jsx
na9da: Ah, that's a local container defined here, not a generic or exported one. I have now renamed it as `LegendsContainer` for clarity.

https://github.com/TerriaJS/terriajs/blob/947e040504bfe05e7bc48d12d27eb070c4060811/lib/ReactViews/Custom/Chart/Legends.jsx#L68

We need `max-width` to constraint the legends div to the size of the chart plot. Without max-width (and I think because we use white-space: nowrap) the div expands to fit the content which can exceed the whole page width.

lib/ReactViews/Custom/Chart/Legends.jsx
na9da: I tend to use locally defined components instead of the `Text` `Box` generic ones. I wish we had tailwind. I have now replaced it with `TextSpan` and `overflowEllipsis`.

lib/ReactViews/Custom/Chart/Legends.jsx
na9da: Can you clarify this? The `Label` here is not form label although it is called `Label`. Maybe I should rename it to something else less confusing. I like the suggestion to use `aria-label` in addition to `title`.

na9da: I have renamed it as `LegendWrapper` and `LegendText`

Issue #5782, opened 2021-09-02
Mobile: Hamburger menu changes


wibily:
- make About link the same style
- Add 'Help" option to link to external help page (https://terriajs.gitbook.io/terriajs/)
 ![IMG_2224 PNG](https://user-images.githubusercontent.com/4196561/131801653-16d5c6ad-d720-4dfe-90c4-13e83bfc0752.png)

na9da:
- Fix for about menu item style - #5796 
- Help link, to discuss with @AnaBelgun

na9da:
This needs to go in `config.json` as the last array entry under `parameters > helpContent`

```json
{
        "itemName": "mapuserguide",
        "title": "translate#helpPanel.mapUserGuide",
        "paneMode": "externalLink",
        "url": "https://terriajs.gitbook.io/terriajs/"
   }
```

PR #5796, opened 2021-09-07
HOC `withControlledVisibility` should extend propTypes of the wrapped component

na9da:
### What this PR does

https://github.com/TerriaJS/terriajs/blob/9e6a18f8c656701f2cad8d550d32da468f8eccc0/lib/ReactViews/StandardUserInterface/processCustomElements.jsx#L71-L79

Here we check  `propTypes` to decide whether to forward `smallScreen` prop to the child components. But if the child component (like [MenuItem](https://github.com/TerriaJS/terriajs/blob/main/lib/ReactViews/StandardUserInterface/customizable/MenuItem.jsx)) is wrapped by `withControlledVisibility` then the check fails and the prop is discarded. This resulted in some buttons to not be styled correctly in mobile.

eg from #5782 

before:
image

after:
image

### Testing

- Open [link](http://ci.terria.io/extend-proptypes/)
- Switch to mobile view
- Open the hamburger menu
- The About link should have the same style as the rest

### Checklist


tephenavies (approved):

wibily:
Nanda to hit merge button

PR #5807, opened 2021-09-13
Restyle help panel

### LINKED ISSUE #5636: Create UX wireframes for Help menu redesign

na9da:
### What this PR does

Fixes #5636 #4670 

Changes:
- Re-styles help panel menu items based on the new design from @philipgrimmett 
- Adds map user guide link to mobile menu if it is supplied in `Terria.configParameters.helpItems`.
- Fixes app crash when loading in mobile Safari browser. This is happening because of a render loop in `TooltipWrapperRaw` component as the following check fails. Even though help panel is not shown anywhere in mobile, the help panel is rendered eagerly. I assume it was done to animate opening of help panel. I have changed it so that the `HelpPanel` component is rendered only when activated (https://github.com/TerriaJS/terriajs/pull/5807/commits/daea39115c015108ce9deaeafacd0c758601347e). I have also made changes to retain the animation when help panel is opened (https://github.com/TerriaJS/terriajs/pull/5807/commits/2195fc73411b27e933a8c908e2558a4899ab3e35).

https://github.com/TerriaJS/terriajs/blob/3c090fa5bccbe88c030f17b317b839a05080271d/lib/ReactViews/Generic/TooltipWrapper.tsx#L106-L111

On mobile Safari the above check passes each time (difference is above 5 maybe because of high pixel density?) and an infinite update loop is set up crashing the app. Fixing the root cause will require a deeper surgery. My fix above is still a good fix because we shouldn't be rendering help panel in mobile. I have made a ticket capturing the problem https://github.com/TerriaJS/terriajs/issues/5827

### Testing

* Open [this link](http://ci.terria.io/restyle-help-panel/#configUrl=https://gist.githubusercontent.com/na9da/b326147a3cc7a066d6501ecef2b93960/raw/834af0d311ccf1b4d6b73872b0b3cfdb6a914412/config-with-map-user-guide.json)
* Check out the help panel. You should see a link to the map user guide.
* Now switch to mobile view from the developer console
* Open the hamburger menu
* You should see a link to the map user guide.

### Checklist


lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
```diff
@@ -120,46 +120,17 @@ class HelpPanel extends React.Component {
           </Box>
         </Box>
         <Spacing bottom={10} />
-        <Box centered displayInlineBlock>
+        <Box centered displayInlineBlock fullWidth styledPadding="0 26px">
           {helpItems && (
             <For each="item" index="i" of={helpItems}>
-              <Box displayInlineBlock fullWidth styledHeight={"70px"}>
```
na9da, line 126: Removed the `Box` as the new styling doesn't need it.

lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
```diff
@@ -120,46 +120,17 @@ class HelpPanel extends React.Component {
...
-                  <StyledIcon
-                    glyph={Icon.GLYPHS.externalLink}
-                    displayInline
-                    styledWidth={"15px"}
-                    fillColor={this.props.theme.textDark}
-                  />
-                </a>
-              </Text>
-            </Box>
-          </Box>
```
na9da, line 162: Map user guide link is now generated from `Terria.configParameters.helpItems` just like the other help menu items.

lib/ReactViews/Map/Panels/HelpPanel/HelpPanelItem.jsx
```diff
@@ -32,91 +29,94 @@ class HelpPanelItem extends React.Component {
...
-            <Text
-              semiBold
-              extraLarge
-              uppercase
-              textDark
-              css={`
-                padding-right: 25px;
-                padding-left: 5px;
-                display: table-cell;
-                vertical-align: middle;
```
na9da, line 97: Extracted into separate components.

lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
wibily: Has the user guide link been added to all the other maps?

lib/ReactViews/Mobile/MobileMenu.jsx
```diff
@@ -77,12 +79,41 @@ const MobileMenu = observer(
       this.props.viewState.toggleFeaturePrompt("mapGuidesLocation", true, true);
     },
 
+    /**
+     * If the help configuration defines an item named `mapuserguide`, this
+     * method returns props for showing it in the mobile menu.
+     */
+    mapUserGuide() {
```
wibily, line 86: :/ this seems really hacky, what if we instead moved this into the map specific code ie similar to how the About link is generated ie inside UserInterface.jsx:35 inside TerriaMap? That way, the responsibility of the translation is also deferred to the map specific translation code and we don't have any special treatment for the link inside terrirajs

wibily: Also the external icon link should be right aligned in the mobile menu as per Phil's designs

wibily (approved):
some minor comments to consider

wibily:
also tested on browserstack ios on real iphone, seems to load fine

lib/ReactViews/Map/Panels/HelpPanel/HelpPanel.jsx
na9da: Nope. Have to do that post deploy to avoid an extra link in the help menu. I have noted that in the parent issue.

lib/ReactViews/Mobile/MobileMenu.jsx
na9da: Yes it is a bit hacky, but I think we should treat this as a stop gap before doing https://github.com/TerriaJS/terriajs/issues/5845 as part of the larger mobile UI rework.

Good point about the icon alignment. I'll fix it.

PR #5820, opened 2021-09-16
Fix diverging col scale

nf-s:
### Fix diverging col scale 

Diverging color scales need to be centered around 0.

### Before

http://ci.terria.io/main/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/prod.json&share=s-kPQChAwbjgNvV568gtNBa6laOwh

image

### After

http://ci.terria.io/fix-diverging-col-scale/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/prod.json&share=s-kPQChAwbjgNvV568gtNBa6laOwh

image

### Checklist


lib/Table/TableColorMap.ts
```diff
@@ -300,6 +313,10 @@ export default class TableColorMap {
     return Color.fromCssColorString(this.colorTraits.regionColor);
   }
 
+  @computed get isDiverging() {
```
na9da, line 343: Might be good to add a comment on what we mean by diverging - i.e values cross zero.

na9da (approved):

lib/Table/TableColorMap.ts
nf-s: Done! 
I also realised that I needed to add a test to make sure the colorPalette hasn't been set to something that is not diverging - so I have done that

PR #5822, opened 2021-09-16
Table outlier filter

nf-s:
#### Issues

- [ ] https://github.com/PacificCommunity/Terria-PacificMap/issues/70
- [ ] https://github.com/TerriaJS/nationalmap/issues/1063
- [ ] https://github.com/TerriaJS/terriajs/issues/5805
- [ ] https://github.com/TerriaJS/terriajs/issues/5750
- [ ] https://github.com/TerriaJS/terriajs/issues/5470

#### Changes

* Filter table column values by Z Score. This is controlled by the following `TableColorStyleTraits`:
  * `zScoreFilter` - Treat values outside of specifed z-score as outliers, and therefore do not include in color scale. This value is magnitude of z-score - it will apply to positive and negative z-scores. For example a value of `2` will treat all values that are 2 or more standard deviations from the mean as outliers. This must be defined to be enabled - currently it is only enabled for SDMX (with `zScoreFilter=4`). (See https://towardsdatascience.com/a-brief-overview-of-outlier-detection-techniques-1e0b2c19e561 for background on z-score)
  * `zScoreFilterEnabled` - True, if z-score filter is enabled
  * `rangeFilter` - This is applied after the `zScoreFilter`. It is used to effectively 'disable' the zScoreFilter if it doesn't cut at least the specified percange of the range of values (for both minimum and maximum value). For exmaple if `rangeFilter = 0.2`, then the zScoreFilter will only be effective if it cuts at least 20% of the range of values from the minimum and maximum value
* Add `outlierColor` to `ContinuousColorMap`
* Add `placement` to `SelectableDimension`. This can be used to put `SelectableDimension` below legend using `placement = "belowLegend`
* Add `SelectableDimensionCheckbox` (and rename `SelectableDimension` to `SelectableDimensionSelect`)
* Add `outlierFilterDimension` checkbox `SelectableDimension` to workbench to enable/disable dimension
* Extend `tableStyle.rowGroups` to regions
* Fix `spreadFinishTime` bug

### Before

http://ci.terria.io/main/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/prod.json&share=s-xSGyOIcXSVOhcwaFgWDJ5Ug0oR6

### After

http://ci.terria.io/table-outlier-filter/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/prod.json&share=s-xSGyOIcXSVOhcwaFgWDJ5Ug0oR6

image

#### Further work

- Texture for outlier (diagonal lines)
- maybe grey for outlier color
- Add percentile binning + kmeans binning - use for datasets which aren't linear/normal
- goodness-of-fit for some distributions
  - exponential - apply transformation to make more linear

#### Tests

- [x] Test selectable dimension checkbox
- [x] Test zscore min/max values (with regions)
- [x] Test legend with outlier color

### Checklist


lib/Table/TableColorMap.ts
```diff
@@ -366,6 +332,126 @@ export default class TableColorMap {
...
+    // Array of row group values
+    const rowGroupValues = rowGroups.map(
+      group =>
+        group[1].map(row => values[row]).filter(val => val !== null) as number[]
+    );
+
+    // Get average value for each row group
+    const rowGroupAverages = rowGroupValues.map(val => getMean(val));
+    const std = getStandardDeviation(filterOutUndefined(rowGroupAverages));
+    const mean = getMean(filterOutUndefined(rowGroupAverages));
```
tephenavies: Could this be 50k values for an SA1 csv, or are row groups grouped by something else? I think we should store `filterOutUndefined(rowGroupAverages)` rather than constructing it twice.

tephenavies (changes requested):

lib/Table/TableColorMap.ts
nf-s: yep good point, done

tephenavies (approved):

PR #5823, opened 2021-09-16
Sdmx improvements

nf-s:
### SDMX improvements

- [ ] https://github.com/PacificCommunity/Terria-PacificMap/issues/70
- [x] https://github.com/PacificCommunity/Terria-PacificMap/issues/68
- [x] https://github.com/TerriaJS/nationalmap/issues/1063
- [x] https://github.com/TerriaJS/terriajs/issues/5816

* SDMX will now disable the region column if less than 2 valid regions have been found
* Set `spreadStartTime` and `spreadFinishTime` to `true` for SDMX
* Add SDMX `metadataURLs` from dataflow annotations
* Improve SDMX chart titles
* `TableMixin` will now remove data if an error occurs while calling `forceLoadTableData`
* Make `regionColumn` `isNullable` - this means region column can be disabled by setting to `null`.
* Fix scalar column color map with a single value
* TableMixin will now clear data if an error occurs while calling `forceLoadTableData`
* `TableMixin` will now not return `mapItems` or `chartItems` if `isLoading` 
* SDMX will now use `initialTimeSource = stop`
* Fix `duplicateModels` duplicating observables across multiple models

### New chart example

As this example only has a single region it will display as time-series

image

#### Before

http://ci.terria.io/main/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/prod.json&share=s-sczmoCwlrf0NLyjMkMGtZDyvjzL

#### After

http://ci.terria.io/sdmx-improvements/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/map-config/prod.json&share=s-sczmoCwlrf0NLyjMkMGtZDyvjzL

#### MetadataUrls

image

#### Spread start/finish time

image

### Checklist


lib/ModelMixins/TableMixin.ts
```diff
@@ -605,22 +601,30 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
...
+        // Load region IDS if region mapping
+        const activeRegionType = this.activeTableStyle.regionColumn?.regionType;
+        if (activeRegionType) {
+          await activeRegionType.loadRegionIDs();
+        }
+      } catch (e) {
+        // Clear data if error occurs
         runInAction(() => {
-          this.dataColumnMajor = dataColumnMajor;
+          this.dataColumnMajor = [[]];
```
na9da: Should this be set to `undefined` instead because it could affect things like `_canExportData` returning true when it should be false.

lib/Table/TableColorMap.ts
```diff
@@ -120,6 +120,18 @@ export default class TableColorMap {
           maxValue: this.maximumValue,
           nullColor
         });
+
+        // If we only have one value, create color map with single value
+      } else if (this.colorColumn?.uniqueValues.values.length === 1) {
```
na9da, line 132: Is it possible to rewrite this as a simple `else` instead of `else if` - because it makes it slightly easier to follow? (Assuming the first condition covers all other cases. If not, it makes sense to document the left out cases.)

lib/Models/Catalog/SdmxJson/SdmxJsonDataflowStratum.ts
```diff
@@ -601,6 +626,63 @@ export class SdmxJsonDataflowStratum extends LoadableStratum(
...
 
+  /** Get region TableColumn by searching catalogItem.tableColumns for region dimension
+   * NOTE: this is searching through catalogItem.tableColumns to find the completely resolved regionColumn
+   * This can only be used in computeds/fns outside of ColumnTraits - or you will get infinite recursion
+   */
+  @computed get resolvedRegionColumn() {
+    return this.catalogItem.tableColumns.find(
+      col =>
+        col.name ===
+        this.dimensionColumns.find(col => col.type === "region")?.name
```
na9da: Is it  better to use `TableColumnType.region` here?

na9da (approved):
Looks great @nf-s. Just got a few minor recommendations.

lib/ModelMixins/TableMixin.ts
nf-s: yep done!

lib/Table/TableColorMap.ts
nf-s: I have added some docs to try to make it more clear. This has to be `else if`, as it is possible to have a scalar column with no unique values (i.e. no valid numerical values) - for this case we will need to use a `ConstantColorMap` - which is done at the end of the function

lib/Models/Catalog/SdmxJson/SdmxJsonDataflowStratum.ts
nf-s: Sorry I should use different local variables for the nested `.find()` calls. 

The second one is searching through `dimensionColumns` - which is `TableColumnTraits` - so it needs to search for the `"region"` string instead of `TableColumnType.region`

PR #5826, opened 2021-09-20
Tsify map settings panel

zoran995:
### What this PR does

Convert map settings panel to typescript and use styled components. 
Fixed a check for Leaflet viewer so image optimisation options don't show when the viewer is 2D

### Checklist


lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -0,0 +1,458 @@
...
+    viewer.viewerMode = ViewerMode.Cesium;
+    viewer.viewerOptions.useTerrain = viewerMode === "3d";
+  } else if (viewerMode === "2d") {
+    viewer.viewerMode = ViewerMode.Leaflet;
+  } else {
+    console.error(
+      `Trying to select ViewerMode ${viewerMode} that doesn't exist`
+    );
+  }
+}
```
na9da: I think it is a good decision to extract this out but maybe we should move it to `ViewerMode.ts` and also make `Terria.ts` use this function from `Terria.setViewerMode`. Let met know what you think.

na9da (changes requested):
Thanks @zoran995 for the PR. It looks great. I have just made a small recommendation.

lib/ReactViews/Map/Panels/SettingPanel.tsx
zoran995: good point
it was a plan to extract it as a separate function when `Terria.setViewerMode` get merged but completely forgot about it

zoran995:
@na9da I have extracted `setViewerMode` as you suggested. I have also taken a new version of Checkbox from our implementation and added it here as it is more flexible to use it with child elements instead of label prop

na9da (approved):
Thanks @zoran995 !

PR #5837, opened 2021-09-21
Support group models in workbench

### LINKED ISSUE #5825: Allow group id in workbench
See https://github.com/TerriaJS/nsw-digital-twin/issues/538

mwu2018:
### What this PR does

Fixes #5825 

This is mainly to address this issue: https://github.com/TerriaJS/nsw-digital-twin/issues/538, where magda will post init data where workbench might contain group like models such as esri map server group that can not be added to the map. This PR will add group members to the map.

### Checklist


lib/Models/Terria.ts
```diff
@@ -1241,6 +1241,47 @@ export default class Terria {
...
+    } else {
+      errors.push(
+        TerriaError.from(
+          "Can not load an un-mappable item to the map. Item Id: " +
+            model.uniqueId
+        )
+      );
+    }
+  }
+
```
nf-s: I think it would be better to add the loading group logic to `workbench.add()` - as it already has the loading reference and mapitems logic

https://github.com/TerriaJS/terriajs/blob/a406adddced1cb994b03bfcdb44748b618cb86ae/lib/Models/Workbench.ts#L161-L212

This would require some changes to `applyInitData`, see next comment https://github.com/TerriaJS/terriajs/pull/5837#discussion_r713890143

lib/Models/Terria.ts
```diff
@@ -1375,6 +1416,14 @@ export default class Terria {
...
 
+    const newItems: BaseModel[] = [];
+
+    await Promise.all(
+      newItemsRaw.map(async model => {
+        await this.pushAndLoadMapItems(model, newItems, errors);
+      })
+    );
+
     runInAction(() => (this.workbench.items = newItems));
```
nf-s, line 1452: ... continuing from previous comment

Here we could just have something like

```ts
    this.workbench.add(newItems);
```

This may need `workbench.add` to be refactored to return `Result` objects instead of throwing `TerriaError`s. Because at the moment, if one model fails to load in the workbench, then it all will fail.

nf-s (changes requested):

lib/Models/Terria.ts
mwu2018: After reading the `workbench.add()` code, I feel that it requires too many code changes if moving the loading group logic there. The primary use case of this logic is to process init data. It seems that my changes are simple and easy to follow. Besides, with my current changes, if one model fails to load in the workbench, it will have no impact on the rest of model loading.

lib/Models/Terria.ts
mwu2018: The original code also uses `loadMapItems` logic in `applyInitData` method.

mwu2018:
@nf-s To move the loading group logic to `Workbench`, the potential changes are given in this PR https://github.com/TerriaJS/terriajs/pull/5853. No unit tests yet. I am not sure if it is better than the current PR.

lib/Models/Terria.ts
nf-s: No problem, I'm happy to keep it in `Terria.ts` for the moment, but I think that in the long-term this should all be in `workbench.add` as it is directly related to adding models to the workbench, and doesn't really have anything to do with core Terria functionality - `Terria.ts` is too large at the moment.

nf-s:
No problem, I'm happy to keep it in `Terria.ts` for the moment, but I think that in the long-term this should all be in `workbench.add` as it is directly related to adding models to the workbench, and doesn't really have anything to do with core Terria functionality - `Terria.ts` is too large at the moment.

test/Models/TerriaSpec.ts
```diff
@@ -865,6 +867,225 @@ describe("Terria", function() {
...
+            if (url.indexOf("request=GetCapabilities") !== -1) {
+              args[0] = "test/Terria/applyInitData/WmsServer/capabilities.xml";
+            } else {
+              args[0] = "test/Terria/applyInitData/empty.json";
+            }
+          }
+
+          const result = realLoadWithXhr(...args);
+          return result;
+        });
```
nf-s, line 973: Is there a reason why this doesn't follow the normal pattern we use for mocking xhr calls?

For example

https://github.com/TerriaJS/terriajs/blob/6f04346ca07899a69c05f3ffb07d8f8caf97482c/test/Table/TableStyleSpec.ts#L25-L41

lib/Models/Terria.ts
```diff
@@ -1241,6 +1241,47 @@ export default class Terria {
     );
   }
 
+  // This method is created to support unit tests.
+  async loadMapItems(model: BaseModel, errors: TerriaError[]) {
+    (await (<any>model).loadMapItems()).pushErrorTo(errors);
+  }
```
nf-s: I don't think this is needed - instead of using `spyOn(terria, 'loadMapItems')` - you can check if a model's map items as been loaded by checking `loadMapItemsResult` is defined

Also you shouldn't type cast to `any` here - you can use `MappableMixin.isMixedInto`.

nf-s (changes requested):

test/Models/TerriaSpec.ts
```diff
@@ -865,6 +867,225 @@ describe("Terria", function() {
...
+        loadMapItems = spyOn(terria, "loadMapItems").and.callThrough();
```
nf-s: See https://github.com/TerriaJS/terriajs/pull/5837#discussion_r715505258

nf-s (changes requested):
Another issue, the items in the workbench aren't in order.

For example - http://ci.terria.io/issue-5825b/#clean&https://gist.githubusercontent.com/nf-s/32ae934cbb72126a175096f8921c51d6/raw/774b8185b2f402a78bdb5771bc8c17e385e0177b/group-workbench-test.json

Specifies:

```json
"workbench": [
    "test-item-1",
    "test-group",
    "test-item-2"
  ],
```

But the order after loading is:

```json
"workbench": [
    "test-item-1",
    "test-item-2",
    "test-group"
  ],
```

lib/Models/Terria.ts
mwu2018: This function is needed because I've revised the unit test so that it spies on `loadMapItems` but does not actually call through -- Calling through will make unit test unnecessarily complicated.

mwu2018:
> Another issue, the items in the workbench aren't in order.
> 
> For example - http://ci.terria.io/issue-5825b/#clean&https://gist.githubusercontent.com/nf-s/32ae934cbb72126a175096f8921c51d6/raw/774b8185b2f402a78bdb5771bc8c17e385e0177b/group-workbench-test.json
> 
> Specifies:
> 
> ```json
> "workbench": [
>     "test-item-1",
>     "test-group",
>     "test-item-2"
>   ],
> ```
> 
> But the order after loading is:
> 
> ```json
> "workbench": [
>     "test-item-1",
>     "test-item-2",
>     "test-group"
>   ],
> ```

The order is now maintained.

lib/Models/Terria.ts
```diff
@@ -1241,6 +1241,49 @@ export default class Terria {
...
+      await this.loadMapItems(model, errors);
+    } else {
+      errors.push(
+        TerriaError.from(
+          "Can not load an un-mappable item to the map. Item Id: " +
+            model.uniqueId
+        )
+      );
+    }
+  }
```
tephenavies: * `loadMapItems`  shouldn't be in Terria.ts. It should be exported from MappableMixin as a separate named export.
* `pushAndLoadMapItems` can stay in Terria.ts, since it's a replacement for the old workbench loading code.

mwu2018: `loadmapItems` is only a wrapper -- convenient for unit test to spy on.

nf-s (approved):
Hi Mike, I'm happy with these changes, but can you please move the `loadMapItems` outside of `Terria.ts`

mwu2018:
> Hi Mike, I'm happy with these changes, but can you please move the `loadMapItems` outside of `Terria.ts`

Do you have any suggestion as where to move this function to?

nf-s:
I'm happy with @steve9164 suggestion:
>It should be exported from MappableMixin as a separate named export

mwu2018:
This method is purely for the purpose of unit test. By moving `loadMapItems` outside of `Terria.ts`, it is difficult to test if it is called as expected. 

It is not ideal to verify if `loadMapItems` is called within `applyInitData` that is being tested because it is an implementation detail but I can not think of a better way than this without making some major changes. Any suggestion is welcome.

PR #5838, opened 2021-09-21
Add excludeMembers (blacklist) support to all groups

### LINKED ISSUE #5153: Port "blacklist" for ArcGis(Feature | Map)ServerCatalogGroup to v8
`blacklist` exists in `lib\Traits\ArcGisPortalCatalogGroupTraits.ts`

We need it for:

- ArcGisCatalogGroup
- ArcGisMapServerCatalogGroup
- ArcGisFeatureServerCatalogGroup
- WMSGroup

nf-s:
### Add excludeMembers (blacklist) support to all groups

Fixes https://github.com/TerriaJS/terriajs/issues/5153

Added `excludeMembers` property to `GroupTraits`. It is an array of strings of excluded group and item names. A group or item name that appears in this list will not be shown to the user. This is case-insensitive and will also apply to all child/nested groups

Accompanying `catalog-converter` PR - https://github.com/TerriaJS/catalog-converter/pull/52

### Example

```json
{
  "catalog": [
    {
      "type": "sdmx-group",
      "name": "Pacific Data Hub .Stat",
      "url": "https://stats-nsi-stable.pacificdata.org/rest",
      "excludeMembers": [
        "population",
        "Balance of payments"
      ]
    }
  ]
}
```

image

#### Before

http://ci.terria.io/main/#clean&https://gist.githubusercontent.com/nf-s/71cb8315cd5695d1698fc7465e919def/raw/45a6c17b5c61ba0ba745bde7aae046990a70799a/excludeMembers.json

#### After

http://ci.terria.io/blacklist/#clean&https://gist.githubusercontent.com/nf-s/71cb8315cd5695d1698fc7465e919def/raw/45a6c17b5c61ba0ba745bde7aae046990a70799a/excludeMembers.json

### Checklist


na9da (approved):
Looks great @nfs!

I know it was already in v8 but if it is not too difficult now then maybe we should replace blacklist with excludeList?

nf-s:
Thanks @na9da - yep I'll rename it to something like `excludeMembers`

PR #5843, opened 2021-09-22
Update protomaps and refactor socrata to use geojsonmixin

nf-s:
### What this PR does

- `SocrataCatalogGroup` will not not return groups for Facets if there is only one - so it skips an unnecessary group level.
- Refactor `SocrataMapViewCatalogItem` to use `GeoJsonMixin`
  - So we can ideal zoom and other geojson mixin stuff
- Update protomaps.js to `1.5.0`
  - Fix to circle symoblizer

image

### Checklist

-   [x] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ tests already exist
-   [x] ~I've updated relevant documentation in `doc/`.~ none needed


lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -107,6 +107,18 @@ function GeoJsonMixin<T extends Constructor<Model<GeoJsonTraits>>>(Base: T) {
...
+    constructor(...args: any[]) {
+      super(...args);
+      if (this.strata.get(GeoJsonStratum.stratumName) === undefined) {
+        runInAction(() => {
+          this.strata.set(
+            GeoJsonStratum.stratumName,
+            GeoJsonStratum.load(this)
+          );
+        });
+      }
```
na9da, line 119: Any reason to move this to constructor from `forceLoadMetadata`?

na9da (approved):

lib/ModelMixins/GeojsonMixin.ts
nf-s: I moved it because it doesn't load anything (there is no async call) - it is only used to set `rectangle`

nf-s: https://github.com/TerriaJS/terriajs/blob/90931527d585fda06e87152f94235045802259ff/lib/Core/AsyncLoader.ts#L9

nf-s: We should only use AsyncLoaders if there is an async call

PR #5860, opened 2021-09-27
Changed mobile header icons.

na9da:
### What this PR does

* Change mobile header icons (hamburger, add data & search) and improved the styling.
* Locks the svg icon sizes to 20x20

### Checklist


lib/Sass/exports/_variables-export.scss
```diff
@@ -43,7 +43,7 @@
   colorPrimary: $color-primary;
   colorSecondary: $color-secondary;
   textLight: $text-light; // essentially #fff until we sort out the colour palette from vic
-  textLightTranslucent: rgba($text-light, 0.5);
+  textLightTranslucent: rgba($text-light, 0.3);
```
na9da, line 46: This is only used for the mobile hamburger menu border. So I think it is safe to change it.

lib/ReactViews/Mobile/mobile-header.scss
```diff
@@ -60,6 +61,8 @@
   padding-right: $padding;
   position: relative;
   width: 40px;
+  border-radius: 4px;
+
```
na9da, line 62: This is a repetition, I think we should fix it when we remove `scss` and move it to styled-components.

KeyboardSounds (approved):

PR #5873, opened 2021-09-30
Wc traitdoc

### LINKED ISSUE #5829: update trait system documentation in model-layer.md

wibily:
### Additional trait documentation

Adds additional context and information about how the Trait system works

### Checklist


doc/contributing/traitsInDepth.md
```diff
@@ -0,0 +1,45 @@
...
+5. User (highest priority)
+
+### Default Stratum
+This is the lowest priority stratum best thought of as a sensible fallback value.
+
+### Underride Stratum
+When loading data from external sources, we might want to set values at a lower priority. For example, if a field exists both at a group level and at an individual member level. We might want to set the value to underride at the group level so that the value can be overriden at the individual member level.
+
+### Definition Stratum
+This is the default priority level we should use for values loaded from configuration whether it be from a config file or from an asynchronous call to an external API.
```
nf-s: This is not entire true, traits from asynchronous calls are usually put into their own LoadableStratum.
`LoadableStratum` sit between ~Definition and Override~ Underide and Definition

nf-s (changes requested):

doc/contributing/traitsInDepth.md
nf-s: It might be worth adding a Loadable Stratum section with a description like this

>Load Strata: (singular: Load Stratum) Strata that provide values for a subset of the properties defined on model's traits class, usually by loading them from an external source such as WMS GetCapabilities service. Example: GetCapabilitiesStratum.

From https://github.com/TerriaJS/terriajs/blob/main/doc/contributing/model-layer.md#types-of-classes-in-the-model-layer

nf-s: There can be multiple Load Strata - for example CKAN reference that resolved to a WMS item will have the following strata:

- Default
- Underride
- CkanDatasetStratum (LoadableStratum)
- (WMS)GetCapabilitiesStratum (LoadableStratum)
- Definition
- Override
- User

wibily: oh what's the difference between Definition and Load? I unfortunately missed Loadable Stratum as I was only looking at the constructor but i see it in the priority numbers.

wibily: yup, I'll add in the Loadable Stratum in the docs/images. I'll probably abstract out the different types of LoadableStratum though

nf-s: Definition is mostly just configuration from JSON initialisation (catalog) files

nf-s: The usage is quite inconsistent thought 😢 

I would say that Definition should **only** be set through initialisation files.

And Underride should be used when models are created dynamically (eg WebMapServiceCatalogGroup creating WebMapServiceCatalogItems)

nf-s: Another point about Defaults, these are set in trait definitions

doc/contributing/traitsInDepth.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Override Stratum
+This is the priority level we use when we want to override a configuration value that isn't a result of a user interaction.
+
+### User Stratum
+This is the highest priority stratum. We use this when a user explicitly sets some value and we want that value to override the values loaded through configuration.
+
+## What happens at runtime
+![Runtime](./img/runtime.png)
```
tephenavies, line 43: 1. nit: in StratumOrder the ordering topToBottom means highest priority (user) to lowest (defaults). It'd be nice, but not essential, for the diagram to match and order these top to bottom from highest to lowest priority.
2. And if you make the image again it'd be nice for the accessing code to look more like normal Trait code (just `item.opacity`, or `model.opacity`)

tephenavies: ~Actually have I got that wrong? Is topToBottom actually deafults to user?~ I'm pretty sure I'm right, but I'm not great at reading `sort` functions:
https://github.com/TerriaJS/terriajs/blob/65c3956256db46b5a2e2d7a0d7f9284515b84915/lib/Models/Definition/StratumOrder.ts#L89-L121

doc/contributing/traitsInDepth.md
```diff
@@ -0,0 +1,45 @@
+# Traits in Depth
+
+The Trait system is Terria's way of managing a priority order for state. For example, we might have a default opacity property from a config file that should be overriden by a user provided value. This article will go through how Traits are made, how the priority system works and how updates happen in the UI.
+
+## Making a Trait
+![Making a trait](./img/making_a_trait.png)
+
+For base traits such as `BaseMapsTraits`, we extend ModelTraits. However when composing multiple traits to form a new trait, we use the `mixTraits` function. For example, `ArcGisTerrainCatalogItemTraits` is composed out of `UrlTraits, MappableTraits, CatalogMemberTraits`.
+
+When creating a new property within a Trait, we use [decorators](https://babeljs.io/docs/en/babel-plugin-proposal-decorators) to ensure that the field takes part in the Trait system.
```
tephenavies: I'm worried about pluralisation here ("Trait" and "Traits"). Trait only refers to a single value <sub>which could be a single object (and if the type of that trait is `objectTrait` its shape is specified by a class of Traits)... but I digress</sub>

In this diagram and doc I think most instances of "Trait" should be "Traits". And you wouldn't `mixTraits` on TraitA & TraitB, but you could mix ATraits & BTraits.

nf-s (changes requested):
Hey @wibily do you mind if we put this on pause until we sort out the whole `underride` vs `loadable` strata thing.

I think this is a good opportunity for us to finally fix some of these issues.

I have started a `strata-examples.md` doc to try to articulate and help clarify the problem

https://github.com/TerriaJS/terriajs/blob/wc-traitdoc-nfs-example/doc/contributing/strata-examples.md

doc/contributing/traitsInDepth.md
wibily: will change

PR #5882, opened 2021-09-30
Zoom on enable

zoran995:
### What this PR does

Fixes #5836

Add `zoomOnEnable` to mappable traits. When it's true, zoom to the item when it's added to workbench

### Checklist

-   [x] ~~I've updated relevant documentation in `doc/`.~~


ykiu:
This is nice.

malnajdi:
will this be merged in next release ? also what about isEnabled ?

na9da:
Hey @zoran995 - Thanks for fixing this :hearts: 

It looks good to me, however I suggest changing the trait name to `zoomOnAddToWorkbench` as it makes it clear that the zoom happens only when the item is added to workbench and not on show/hide.

@malnajdi - did you mean what is the equivalent for `isEnabled` trait in v8? That would be the `show` trait.

malnajdi:
@na9da - What i want to do is on loading the terriaMap i want to show 2 default layers that they should be showing all time.
It seems the v7 converter tool creates a list called workbench and it contains all ids of the layers need to be presented. 

But isn't there a way to do it like the version7 ?

zoran995:
Hi @na9da good point for name, updated

na9da (approved):

na9da:
Hi @malnajdi - the `isEnabled` behavior of auto adding to workbench is deprecated in v8. Adding to the workbench array is the way to do it. I'm assuming it was removed because of conflicting behavior with stories and share URLs.

PR #5885, opened 2021-10-01
Fix a couple of bugs that caused the app to show scrollbar in certain instances.

na9da:
### What this PR does

Fix a couple of bugs that caused the app to show scrollbar in certain instances.

### Testing

Since the scrollbar is currently triggered only in the natmap repo, the changes need to be tested against that repo.

### Checklist

### DIFF
--- a/lib/ReactViews/DragDropFile.tsx
+++ b/lib/ReactViews/DragDropFile.tsx
@@ -119,7 +119,7 @@ class DragDropFile extends React.Component<PropsType> {
           [Styles.isActive]: this.props.viewState.isDraggingDroppingFile
         })}
       >
-        {isDefined(this.props.viewState.isDraggingDroppingFile) ? (
+        {this.props.viewState.isDraggingDroppingFile ? (
           <div className={Styles.inner}>
             <Trans i18nKey="dragDrop.text">
               <h3 className={Styles.heading}>Drag & Drop</h3>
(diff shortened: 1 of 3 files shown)


lib/ReactViews/StandardUserInterface/StandardUserInterface.jsx
```diff
@@ -248,7 +248,7 @@ const StandardUserInterface = observer(
         !this.shouldUseMobileInterface();
       const showStoryPanel =
         this.props.terria.configParameters.storyEnabled &&
-        this.props.terria.stories.length &&
+        this.props.terria.stories.length > 0 &&
```
na9da, line 251: All conditionals used in react jsx should be true/false. In this case `showStoryPanel` is `0` when there are no story scenes. This resulted in a stray `0` to be printed in the dom.

lib/ReactViews/DragDropFile.tsx
```diff
@@ -119,7 +119,7 @@ class DragDropFile extends React.Component<PropsType> {
           [Styles.isActive]: this.props.viewState.isDraggingDroppingFile
         })}
       >
-        {isDefined(this.props.viewState.isDraggingDroppingFile) ? (
+        {this.props.viewState.isDraggingDroppingFile ? (
```
na9da, line 122: This resulted in the drag-drop container being shown even when the user is not dragging anything.

tephenavies (approved):

PR #5904, opened 2021-10-11
Demo PR for upgrading to mobx6.

na9da:
### What this PR does

Demonstrates what is required to upgrade terria to mobx6. 

#### Changes made
- This is the result of running `npx mobx-undecorate --keepDecorators` and then fixing what was broken.
- By doing `--keepDecorators` we can keep using the decorator style, just need to call `makeObservable(this)` in the constructor.
- As explained in the [mobx docs](https://mobx.js.org/subclassing.html) we can only use the `override` decorator for decorating a method that is already decorated in the base class (as an `action`, `observable` or `computed`)
- Since we auto decorate all traits in `CreateModel` we have to always use `override` when overriding a trait.
- Because mobx will throw an error if you try to decorate an already decorated method we have to ensure that our Mixins are included only once. I had to add a guard for some of our Mixins to do this. Eg [CatalogMemberMixin](https://github.com/TerriaJS/terriajs/blob/b2945ec562176958b1d1a3ccf956069e24534325/lib/ModelMixins/CatalogMemberMixin.ts#L22-L24)

#### What to expect
- There can be random crashes (~for eg when zooming chart~ fixed)
- If I have forgotten to call `makeObservable(this)` from a class constructor it can result in loss of reactivity.
- It can also cause performance issues if some computed value not getting picked up and recomputing all the time.
- ~Tests are breaking which I'll fix soon.~ Tests pass in local, CI is broken for some reason.
- If you have custom catalog items and code in terriamap, you will have to run `npx mobx-undecorate --keepDecorators` and fix the errors. The errors will be mostly due to the script not setting up the constructor arguments correctly for the models.

#### TODOs
- [ ] Explore using [jscodeshift](https://github.com/facebook/jscodeshift) to automate the migration
- [ ] Attempt to port NSW, VIC & QLD instances.

### Checklist


lib/ModelMixins/MappableMixin.ts
```diff
@@ -51,8 +51,19 @@ export function isDataSource(object: MapItem): object is DataSource {
 }
 
 function MappableMixin<T extends Constructor<Model<MappableTraits>>>(Base: T) {
+  if (Base.prototype.isMappable) {
+    return Base as Constructor<MappableMixin> & T;
+  }
+
```
na9da, line 57: We have some Mixin inheritance chains that result in some Mixins getting included twice. Mobx6 will raise an error when this happens because of an attempt to annotate a member that was already annotated. To prevent this we have to write a guard that will apply the mixin only if it is not already applied and otherwise returns the parameter unmodified. I had to do the same for `CatalogMemberMixin` and `AccessControlMixin`.

wibily:
need to set up meeting with dev team to discuss this and ts changes

wibily:
requires catalog items conversion (e.g nsw catalog items) - possibly migration script?

wibily:
0.5 estimate to get this to a stage where we can go to Ana with an estimate for timing when this should proceed with this

pjonsson:
Does #6687 supersede this PR so it can be closed?

PR #5911, opened 2021-10-14
Add all knownContainerIds to sharelink models 

nf-s:
### Add all knownContainerIds to sharelink models 

* Make flex-search usage (for `CatalogIndex`) web-worker based
* Add `completeKnownContainerUniqueIds` to `Model` class - This will recursively travese tree of knownContainerUniqueIds models to return full list of dependencies
* Add all models from `completeKnownContainerUniqueIds` to shareData.models (even if they are empty)

### Changes to share JSON

image

### Before

http://ci.terria.io/main/#configUrl=https://nationalmap.gov.au/config.json&share=s-ycMoOeupODIhbZ5IH0pWuidNw9N

### After

http://ci.terria.io/sharelink-modelids/#configUrl=https://nationalmap.gov.au/config.json&share=s-89q4Fvaz6O3Y1iz5l3dLhxgi1m3

### Checklist

### DIFF
--- a/lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
+++ b/lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
@@ -11,6 +11,7 @@ import hashEntity from "../../../../Core/hashEntity";
 import CommonStrata from "../../../../Models/Definition/CommonStrata";
 import saveStratumToJson from "../../../../Models/Definition/saveStratumToJson";
 import { BaseModel } from "../../../../Models/Definition/Model";
+import { runInAction } from "mobx";
 
 const CatalogMember = {}; // TODO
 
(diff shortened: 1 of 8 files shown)


lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.js
```diff
@@ -142,7 +145,9 @@ function addModelStratum(terria, model, stratumId, force, initSource) {
     model.knownContainerUniqueIds &&
     model.knownContainerUniqueIds.length > 0
   ) {
-    models[id].knownContainerUniqueIds = model.knownContainerUniqueIds.slice();
+    models[
+      id
+    ].knownContainerUniqueIds = model.completeKnownContainerUniqueIds.slice();
```
tephenavies: I think we should add each container to the models array, similar to if the model had user stratum values. This gives us more information than putting all of the container ids into an array and allows Terria to know the dependency tree to be able to load them in order.

nf-s: done

tephenavies (approved):

PR #5918, opened 2021-10-17
Config traits

zoran995:
### What this PR does

TerriaMap PR: https://github.com/TerriaJS/TerriaMap/pull/557

Update TerriaJS config parameters to use traits system. Some of the parameters are renamed
- `showWelcomeMessage` and `welcomeMessageVideo` are now parameters on object `welcomeMessage` as `show` and `video`
- `helpContent` is now `helpItems`
- `helpContentTerms` are now `helpTerms` (they are not strictly related to `helpContent`)
- `feedbackPreamble` and `feedbackMinLength` are now parameters on object `feedback` as `preamble` and `minLength`
- `brandBarElements`, `brandBarSmallElements`, `displayOneBrand` are now parameters on object `brandBar` as `elements`, `smallElements`, `displayOneBrand`
- `googleUrlShortenerKey` is deprecated long time ago but still left as a config parameter
- `disclaimer` is not used anywhere so I added it to deprecated list
- `globalDisclaimer` is also not used anywhere - not sure if IT should be marked as deprecated as it is set from TerriaMap?

TODO
- [x] Check if some conditions depend only on the existence of specific `configParameter` in case when it is not `primitiveTrait`, `objectTrait` will always be defined and that condition should be updated
- [x] Convert deprecated options to the new ones - _edit: added basic conversion, as I don't see that we really need anything on model level for now_
- [ ] Documentation should generate automatically (as for catalog items) - Best to do this as a separate PR
- [ ] Introduce required option for trait system (at least for configuration if not doable for generic) - Probably best to do this as a separate PR (maybe target this one at first if this is not merged before)

p.s. this was much easier than initially expected

### Checklist

-   ~[ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ - All existing test pass


lib/Core/TerriaError.ts
```diff
@@ -218,7 +218,7 @@ export default class TerriaError {
 
     // shouldRaiseToUser will be true if at least one error includes shouldRaiseToUser = true
     const shouldRaiseToUser = filteredErrors
-      .map(error => error._shouldRaiseToUser ?? false)
+      .map(error => error.shouldRaiseToUser ?? false)
```
zoran995: @nf-s when directly accessing property no error was raised to user when there is more than one error. `updateFromJson` will generate errors of `error` severity but it won't be checked when directly accessing property instead of using getter.

lib/ReactViews/Map/TrainerBar/TrainerBar.tsx
```diff
@@ -303,17 +304,18 @@ interface TrainerBarProps extends WithTranslation {
...
+  // map-owner defined (helpItems)content which could be malformed
+  if (!viewState.trainerBarVisible || !helpItems) {
     return null;
   }
 
   const selectedTrainer = getSelectedTrainerFromHelpContent(
     viewState,
-    helpContent
+    //@ts-ignore
+    helpItems
```
zoran995, line 318: was getting type error

lib/Core/TerriaError.ts
nf-s: Hey @zoran995 , `_shouldRaiseToUser` is more like `overrideRaiseToUser` - i have created a PR to rename it https://github.com/TerriaJS/terriajs/pull/5923/files

So it needs to stay as is, but you did make me realise there is a bug here!

### Before

https://github.com/TerriaJS/terriajs/blob/62c69365b376eb67f7f38bd10bfc0b4a318ea003/lib/Core/TerriaError.ts#L219-L222

`overrideRaiseToUser` was being incorrectly assigned `false`, which was causing `get shouldRaiseToUser` to return false.

### After

https://github.com/TerriaJS/terriajs/blob/948bc3d11ac33c3b9936d3940db58ac924cfe1c4/lib/Core/TerriaError.ts#L219-L232

Can you please see if this fixes your issue?

zoran995: thanks @nf-s everything works correctly after merging your changes

zoran995:
This is ready for review, added a basic conversion so no breaking changes are introduced in patch or minor version.

zoran995:
did another pass over this, also tried loading prod map-configs from https://github.com/TerriaJS/saas-catalogs-public, and no issue is raised to the user so I am pretty much confident that this can go without having to update the configs right away

lib/Traits/Configuration/ConfigParametersTraits.ts
```diff
@@ -0,0 +1,403 @@
...
+    type: "boolean",
+    description: "True to display in-app guides."
+  })
+  showInAppGuides?: boolean = false;
+
+  @objectArrayTrait({
+    type: HelpItemTraits,
+    name: "Help content items",
+    description: "The content to be displayed in the help panel.",
+    idProperty: "index"
```
tephenavies, line 327: Just dropping a comment here for a future full review: I think this should be:
```suggestion
    idProperty: "itemName",
```
though I haven't thought through the consequences of either `idProperty` choices.

lib/Traits/Configuration/HelpItemTraits.ts
```diff
@@ -0,0 +1,116 @@
...
+  @objectArrayTrait({
+    type: StepItemTraits,
+    name: "Footnote",
+    description: "Footnote for the trainer item.",
+    idProperty: "index"
+  })
+  steps?: StepItemTraits[];
+}
+
+export class HelpItemTraits extends ModelTraits {
```
tephenavies, line 52: Relating to https://github.com/TerriaJS/terriajs/pull/5918/files#r894124712. If using `itemName` as an id you could add:
```ts
static isRemoval(helpItem: HelpItemTraits) {
  // This condition looks sensible
  return helpItem.markdownText === null && helpItem.trainerItems === null;
}
```
And then you could have default helpItems in terriajs that get overriden/removed in TerriaMap config using the strata system, I think. I have to play around with this branch sometime. And default help items in terriajs means translation support!

lib/Traits/Configuration/ConfigParametersTraits.ts
nf-s: I think `itemName` is good - then makes it easy to remove specific default help items

PR #5919, opened 2021-10-18
Checkbox label as child element

zoran995:
### What this PR does

Taken from #5826. Update checkbox component to support label added as child elements instead of using label prop, this enables easier styling of the label. All checkbox child elements receive `isDisabled`, `isChecked` and style `font-size: inherit` as a props. Inheritance of font size is needed so checkbox icon and label have the same size.
Checkbox element now also accepts all `Text` props so 

### Checklist

-   [x] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ - units tests for checkbox are still not merged


AnaBelgun:
@na9da is it possible to review this? thank you

zoran995:
merged as part of #5826

PR #5922, opened 2021-10-19
Fix a bug in `Cesium.prototype.observeModelLayer()`

ykiu:
### What this PR does

This PR fixes a potential bug in `Cesium.prototype.observeModelLayer()`.

The function was not handling the removal of 3D Tilesets correctly, and some tilesets were left out even if they were no longer in the workbench.

Previously, this bug was hidden behind a [`runLater()` hack](https://github.com/TerriaJS/terriajs/blob/8.1.4/lib/ModelMixins/Cesium3dTilesMixin.ts#L58-L65) in ObservableCesium3DTileset. I spotted the bug while I was experimenting with a custom implementation of a 3D-tileset-like catalog item. Now the bug is fixed, `Cesium.prototype.observeModelLayer()` can remove 3D tilesets correctly without any hack.

### Checklist

    - I did not make any changes to the tests because I do not intend to change any existing behavior.
    -  No, because this is only an internal change.


na9da:
Thanks @ykiu and good catch! 

Looks like the same thing happens for `dataSources` and `imageryLayers`. Would be good to test them out as well. I'll try to create examples of failing cases.

ykiu:
Thanks for your feedback @na9da!

I don't think this problem happens with `dataSources` and `imageryLayers` as the loop counter is decremented when an item is removed.

https://github.com/TerriaJS/terriajs/blob/8.1.4/lib/Models/Cesium.ts#L541
https://github.com/TerriaJS/terriajs/blob/8.1.4/lib/Models/Cesium.ts#L568

Maybe I should have done the same way for consistency. I'll fix my patch if you prefer that style.

na9da:
Ah ok. I think I like the way you have written the loop starting from the end because in the other case there is a risk of forgetting to decrement.

ykiu:
Okay, so I'll let it as it is.

ykiu:
@na9da Is there anything I should do for moving forward? Let me know if I'm missing something.

na9da:
Hi @ykiu, sorry for the delay. 

I have made similar change to the loops for datasource and imagerylayers and also added a test case to catch it if there is any regression in the future.

na9da (approved):

ykiu:
Thanks @na9da !

PR #5923, opened 2021-10-19
Fix `TerriaError.shouldRaiseToUser` bug

nf-s:
### Fix `TerriaError.shouldRaiseToUser` bug

* Rename `TerriaError._shouldRaiseToUser` to `overrideRaiseToUser`
* Fix `overrideRaiseToUser` bug causing `overrideRaiseToUser` to be set to `true` in `TerriaError.combine`

#### Before

https://github.com/TerriaJS/terriajs/blob/62c69365b376eb67f7f38bd10bfc0b4a318ea003/lib/Core/TerriaError.ts#L219-L222

`overrideRaiseToUser` was being incorrectly assigned `false`, which was causing `get shouldRaiseToUser` to return false.

#### After

https://github.com/TerriaJS/terriajs/blob/948bc3d11ac33c3b9936d3940db58ac924cfe1c4/lib/Core/TerriaError.ts#L219-L232

### Checklist

-   [x] ~I've updated relevant documentation in `doc/`.~

### DIFF
--- a/lib/Core/TerriaError.ts
+++ b/lib/Core/TerriaError.ts
@@ -223,7 +223,7 @@ export default class TerriaError {
 
     // overrideRaiseToUser will be false if:
     // - NO errors includes overrideRaiseToUser = true
-    // - and at least one error includes overrideRaiseToUser = true
+    // - and at least one error includes overrideRaiseToUser = false
     if (
       !isDefined(overrideRaiseToUser) &&
       filteredErrors.map(error => error.overrideRaiseToUser).includes(false)
--- a/lib/Core/TerriaError.ts
+++ b/lib/Core/TerriaError.ts
@@ -53,7 +53,7 @@ export interface TerriaErrorOptions {
   sender?: unknown;
 
   /** True if error message should be shown to user *regardless* of error severity. If this is undefined, then error severity will be used to determine if forceRaiseToUser (severity `Error` are presented to the user. `Warning` will just be printed to console) */
-  forceRaiseToUser?: boolean;
+  forceRaiseToUser?: true;
 
   /** True if the user has seen this error; otherwise, false. */
   raisedToUser?: boolean;
@@ -119,7 +119,7 @@ export default class TerriaError {
   readonly stack: string;
 
   /** Override shouldRaiseToUser (see `get shouldRaiseToUser()`) */
-  forceRaiseToUser: boolean | undefined;
+  forceRaiseToUser: true | undefined;
   @observable showDetails: boolean;
 
   /**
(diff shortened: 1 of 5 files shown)


lib/Models/Terria.ts
```diff
@@ -580,9 +580,9 @@ export default class Terria {
     // Set shouldRaiseToUser:
     // - `true` if forceRaiseToUser agrument is true
     // - `false` if ignoreErrors userProperties is set
-    if (forceRaiseToUser) terriaError.shouldRaiseToUser = true;
+    if (forceRaiseToUser) terriaError.overrideRaiseToUser = true;
     else if (this.userProperties.get("ignoreErrors") === "1")
-      terriaError.shouldRaiseToUser = false;
+      terriaError.overrideRaiseToUser = false;
```
wibily: if we have an underlying TerriaError with overrideRaiseToUser set to true, this would mean that would take precendence over the userProperty 'ignoreError'

lib/Core/TerriaError.ts
```diff
@@ -216,10 +216,20 @@ export default class TerriaError {
         ? TerriaErrorSeverity.Error
         : TerriaErrorSeverity.Warning;
 
-    // shouldRaiseToUser will be true if at least one error includes shouldRaiseToUser = true
-    const shouldRaiseToUser = filteredErrors
-      .map(error => error._shouldRaiseToUser ?? false)
-      .includes(true);
+    // overrideRaiseToUser will be true if at least one error includes overrideRaiseToUser = true
+    let overrideRaiseToUser: boolean | undefined =
```
wibily, line 221: minor nitpick but you could use `filteredErrors.some(error => !!error.overrideRaiseToUser) || undefined`. I think Array.map will create another mapped array although i'm not sure if that will be optimised out by the compiler

wibily (approved):

lib/Core/TerriaError.ts
nf-s: 👍

lib/Models/Terria.ts
nf-s: Yeah i was thinking about this - I'm not sure what the best behavior is.
What do you think?

wibily: i think there's a difference between "has an error occured?" and "does the user want to see it?". I'm not completely across the whole error system but what if the "ignoreError" user property only affected `    this.notificationState.addNotificationToQueue(terriaError.toNotification());` So we still log the error in the console and send it off to rollbar but the only difference is that the user doens't get notified

nf-s: Yeah that is a much better idea. Are you happy with my new changes?
https://github.com/TerriaJS/terriajs/pull/5923/commits/674cac5c2a80efccfc37af48d9ee0fe77886e211

wibily: yeah looks good to me and hey we can always change it again later

wibily (approved):

lib/Models/Terria.ts
nf-s: sounds good, thanks!

PR #5891, opened 2021-10-03
Translations update from Weblate

weblate:
Translations update from [Weblate](https://hosted.weblate.org/projects/terriajs/terriajsnext/) for TerriaJS/TerriaJSNext.

Current translation status:

Weblate translation status

PR #5925, opened 2021-10-20
Update Internationalization.ts with Italian language

LorenzoStucchi:
### What this PR does

Fixes #5891 

Add the option to have the language in Italian

### Checklist


tephenavies:
This PR is the easiest way to add the Italian translation, but it's not scalable. Every translation we add in this way adds to JS bundle size of every TerriaMap, even TerriaMaps that don't make all of the translations available. We'll need to rethink how the set of "default" translations (probably will be en, fr, it, jp) can be included. Maybe the suggestion in https://github.com/TerriaJS/terriajs/pull/5891#issuecomment-948598146 will work

LorenzoStucchi:
Hi @steve9164, I agree with you, this solution is not scalable. The one proposed, seems more intelligent and reusable for all the languages. I closed this PR

PR #5927, opened 2021-10-21
Fix `MapNavigation` getter/setter `visible` bug.

nf-s:
### Fix `MapNavigation` getter/setter `visible` bug.

Some classes only implemented `get visible` and not `set visible` - which upset mobx
- I have replaced `CompositeBarItemController` `visible` setter with `setVisible` function

### For example

This init config doesn't work

```json
{
  "elements": {
    "show-workbench": {
      "visible": false
    },
    "map-data-count": {
      "visible": false
    },
    "help": {
      "visible": false
    },
    "menu-bar": {
      "visible": false
    },
    "map-navigation": {
      "visible": true
    },
    "compass": {
      "visible": false
    },
    "zoom": {
      "visible": true
    },
    "my-location": {
      "visible": false
    },
    "split-tool": {
      "visible": false
    },
    "about-link": {
      "visible": false
    },
    "measure-tool": {
      "visible": false
    },
    "feedback": {
      "visible": false
    },
    "related-maps": {
      "visible": false
    }
  },
  "homeCamera": {
    "west": 109,
    "south": -45,
    "east": 158,
    "north": -8
  },
  "services": [],
  "catalog": []
}
```

image

### Before

http://ci.terria.io/main/#https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/mbd.json

### After

http://ci.terria.io/map-nav-get-set-bug/#https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/pacificmap/mbd.json

### Checklist

-   [] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)

### DIFF
--- a/lib/ReactViews/Map/Navigation/Items/AugmentedVirtualityTool.tsx
+++ b/lib/ReactViews/Map/Navigation/Items/AugmentedVirtualityTool.tsx
@@ -119,11 +119,7 @@ export class AugmentedVirtualityRealignController extends MapNavigationItemContr
 
   @computed
   get visible(): boolean {
-    return this.props.augmentedVirtuality.active && this._visible;
-  }
-
-  set visible(value: boolean) {
-    this._visible = value;
+    return this.props.augmentedVirtuality.active && this.visible;
   }
 
   handleClick(): void {
@@ -219,11 +215,7 @@ export class AugmentedVirtualityHoverController extends MapNavigationItemControl
 
   @computed
   get visible(): boolean {
-    return this.props.augmentedVirtuality.active && this._visible;
-  }
-
-  set visible(v: boolean) {
-    this._visible = v;
+    return this.props.augmentedVirtuality.active && this.visible;
   }
 
   handleClick(): void {
--- a/lib/ReactViews/Map/Navigation/Items/AugmentedVirtualityTool.tsx
+++ b/lib/ReactViews/Map/Navigation/Items/AugmentedVirtualityTool.tsx
@@ -119,7 +119,11 @@ export class AugmentedVirtualityRealignController extends MapNavigationItemContr
 
   @computed
   get visible(): boolean {
-    return this.props.augmentedVirtuality.active;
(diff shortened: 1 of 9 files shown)


lib/ReactViews/Map/Navigation/Items/AugmentedVirtualityTool.tsx
```diff
@@ -119,7 +119,11 @@ export class AugmentedVirtualityRealignController extends MapNavigationItemContr
 
   @computed
   get visible(): boolean {
-    return this.props.augmentedVirtuality.active;
+    return this.props.augmentedVirtuality.active && this._visible;
```
zoran995: is it better to call `super.visible` here

lib/ReactViews/Map/Navigation/Items/AugmentedVirtualityTool.tsx
```diff
@@ -119,7 +119,11 @@ export class AugmentedVirtualityRealignController extends MapNavigationItemContr
...
+  }
+
+  set visible(value: boolean) {
+    this._visible = value;
```
zoran995: at first, this seemed a little contra intuitive to me since setter is already declared at the upper level, but now I see it is hidden by overridden getter

zoran995 (approved):
LGTM
In the long run, we can look at other options so we don't expose base controller property directly (I am not a fan of accessing/exposing values directly if there is a getter somewhere).

just as an idea
What if we update `CompositeBarItemController` to have `setVisible(value: boolean){...}` it will always be available and it contains base logic for setting visible which we can override and call super so everything can chain correctly

nf-s:
@zoran995 good idea! - I have replaced `set visible` with `setVisible`

na9da (approved):
The same risk exists for the other settables like `pinned`. We should probably create a new issue to convert them to the same style.

PR #5933, opened 2021-10-25
Wc print

wibily:
### What this PR does

Refactor to remove the last usage of create-react-class by converting PrintView to use tsx classes. I have also changed the renderLegend function to use `currentTime` instead of `discreteTime` as I'm pretty sure that discreteTime doens't exist.

To compare, I am using the dataset: Marine water quality grids for the Great Barrier Reef region - monthly data > number of observations

Main: http://ci.terria.io/main/#share=s-pCl9Rg6AWyfP0oF3GcBRkI5jnKu
wc-print: http://ci.terria.io/wc-print/#share=s-xVrTjAV430M2Mn5dUbpx3RCoOHl

### Checklist


lib/ReactViews/Map/Panels/SharePanel/PrintView.tsx
```diff
@@ -44,30 +64,24 @@ const PrintView = createReactClass({
         const printWindow = this.props.window;
         const mainWindow = window;
 
-        const printWindowIntervalId = printWindow.setInterval(
+        this.printWindowIntervalId = printWindow.setInterval(
           this.checkForImagesReady,
           200
         );
```
tephenavies: I don't think this will work because methods are not bound when deriving from React.Component. I think you need the following
```suggestion
        this.printWindowIntervalId = printWindow.setInterval(
          this.checkForImagesReady.bind(this),
          200
        );
```

lib/ReactViews/Map/Panels/SharePanel/PrintView.tsx
```diff
@@ -44,30 +64,24 @@ const PrintView = createReactClass({
...
-        const mainWindowIntervalId = mainWindow.setInterval(
+        this.mainWindowIntervalId = mainWindow.setInterval(
           this.checkForImagesReady,
           200
         );
```
tephenavies: Also needs method binding

lib/ReactViews/Map/Panels/SharePanel/PrintView.tsx
```diff
@@ -199,9 +212,71 @@ const PrintView = createReactClass({
...
+   * @param {Terria} options.terria The Terria instance.
+   * @param {ViewState} options.viewState The terria ViewState instance.
+   * @param {Window} [options.printWindow] The window in which to create the print view. This is usually a new window created with
+   *                 `window.open()` or an iframe's `contentWindow`. If undefined, a new window (tab) will be created.
+   * @param {Function} [options.readyCallback] A function that is called when the print view is ready to be used. The function is
+   *                   given the print view window as its only parameter.
+   * @param {Function} [options.closeCallback] A function that is called when the print view is closed. The function is given
+   *                   the print view window as its only parameter.
+   */
+  create(options: CreateOptions) {
```
tephenavies: This should be exported separately (or, if necessary made a `static` function, but a separate export sounds better to me). I get this error in CI:
```
Uncaught TypeError: _PrintView.default.create is not a function
    createPrintView SharePanel.jsx:230
    showPrintView SharePanel.jsx:216
```

lib/ReactViews/Map/Panels/SharePanel/PrintView.tsx
```diff
@@ -130,23 +149,17 @@ const PrintView = createReactClass({
...
     );
-  },
-
-  renderAttribution(attribution) {
-    // For reasons I don't entirely understanding, using parseCustomHtmlToReact instead
-    // of dangerouslySetInnerHTML here doesn't work in IE11 or Edge. All elements after
-    // the first attribution end up just completely missing from the DOM.
-    const html = { __html: DOMPurify.sanitize(attribution) };
-    return <li key={attribution} dangerouslySetInnerHTML={html} />;
-  },
```
tephenavies: I can't find where attribution is dealt with now. Chances are this workaround is not needed any more since we don't support IE11 and I'm 99.9% sure the Edge that doesn't work is the pre-chromium one that we also don't support.

lib/ReactViews/Map/Panels/SharePanel/PrintView.tsx
```diff
@@ -130,23 +149,17 @@ const PrintView = createReactClass({
...
+  }
 
-  renderLegend(catalogItem) {
+  renderLegend(catalogItem: any) {
```
tephenavies: Even if we do need to use `any`, I prefer to add one or more strongly typed type first. Like:
```suggestion
  renderLegend(catalogItem: MappableMixin.Instance | BaseModel | any) {
```
Although this won't throw type errors for bad usage, it enables IDE features like auto complete, and also "Get References" on any MappableMixin.Instance properties/functions used.

The type signature I suggested is only an example. Let me know if you want help working out an appropriate one.

lib/ReactViews/Map/Panels/SharePanel/PrintView.tsx
```diff
@@ -199,9 +212,71 @@ const PrintView = createReactClass({
...
+   * @param {Function} [options.readyCallback] A function that is called when the print view is ready to be used. The function is
+   *                   given the print view window as its only parameter.
+   * @param {Function} [options.closeCallback] A function that is called when the print view is closed. The function is given
+   *                   the print view window as its only parameter.
+   */
+  create(options: CreateOptions) {
+    const {
+      terria,
+      viewState,
+      printWindow = window.open(),
```
tephenavies: I did not know that you can put side effects in defaults when using object destructuring. I know this is in the previous version too, but it's a TIL for me. I guess the behaviour has to match default parameters because object destructuring should be semantically the same whether it's used in function parameters or with `const`.

tephenavies (changes requested):
At the moment print view is not working at http://ci.terria.io/wc-print/. I think the issue is the `PrintView.create` function

wibily:
latest branch is wc-preportal

tephenavies:
I think this is still in draft stage. Please let me know if that's wrong.

PR #5935, opened 2021-10-25
Add support for arbitrary Cesium primitives

ykiu:
### What this PR does

This PR adds support for arbitrary cesium primitives.

Primitive is a low-level API of Cesium and is suitable for building advanced visualizations. A notable example that demonstrates the power of primitives is this [wind visualization](https://cesium.com/blog/2019/04/29/gpu-powered-wind/) by Rayman Ng ([source on GitHub](https://github.com/RaymanNg/3D-Wind-Field)). With the changes in this PR, a future terriajs user will be able to present advanced visualizations like this one using primitives.

We are almost there to support primitives. Cesium3DTileset is one implementation of primitive, and terriajs already supports 3D tiles. Much of the code for working with 3D tiles works just fine with primitives in general. This PR updates some TypeScript type definitions and a type guard function at MappableMixin.ts.

As a side note, below is a code example demonstrating how one can extend terriajs with Cesium primitives. Here, I implement a custom catalog type that returns a Cesium primitive from the `mapItems` getter. This example is taken from Cesium's [tutorial on primitives](https://cesium.com/learn/cesiumjs-learn/cesiumjs-geometry-appearances/#combining-geometries) and adapted for terriajs:

```ts
import Primitive from "terriajs-cesium/Source/Scene/Primitive";
import GeometryInstance from "terriajs-cesium/Source/Core/GeometryInstance";
import RectangleGeometry from "terriajs-cesium/Source/Core/RectangleGeometry";
import Rectangle from "terriajs-cesium/Source/Core/Rectangle";
import Color from "terriajs-cesium/Source/Core/Color";
import PerInstanceColorAppearance from "terriajs-cesium/Source/Scene/PerInstanceColorAppearance";
import ColorGeometryInstanceAttribute from "terriajs-cesium/Source/Core/ColorGeometryInstanceAttribute";

import MappableMixin from "terriajs/lib/ModelMixins/MappableMixin";
import CatalogMemberMixin from "terriajs/lib/ModelMixins/CatalogMemberMixin";
import CreateModel from "../Definition/CreateModel";
import mixTraits from "terriajs/lib/Traits/mixTraits";
import MappableTraits from "terriajs/lib/Traits/TraitsClasses/MappableTraits";
import CatalogMemberTraits from "terriajs/lib/Traits/TraitsClasses/CatalogMemberTraits";
import { computed } from "mobx";

class PrimitiveTraits extends mixTraits(MappableTraits, CatalogMemberTraits) {}

export default class PrimitiveCatalogItem extends MappableMixin(
  CatalogMemberMixin(CreateModel(PrimitiveTraits))
) {
  static readonly type = "primitive";

  get type() {
    return PrimitiveCatalogItem.type;
  }

  forceLoadMapItems() {
    return Promise.resolve();
  }

  @computed
  get mapItems() {
    const instances = [];

    for (let lon = -180.0; lon < 180.0; lon += 5.0) {
      for (let lat = -85.0; lat < 85.0; lat += 5.0) {
        instances.push(
          new GeometryInstance({
            geometry: new RectangleGeometry({
              rectangle: Rectangle.fromDegrees(lon, lat, lon + 5.0, lat + 5.0),
              vertexFormat: PerInstanceColorAppearance.VERTEX_FORMAT
            }),
            attributes: {
              color: ColorGeometryInstanceAttribute.fromColor(
                Color.fromRandom({ alpha: 0.5 })
              )
            }
          })
        );
      }
    }

    return [
      new Primitive({
        geometryInstances: instances,
        appearance: new PerInstanceColorAppearance()
      })
    ];
  }
}
```

The custom catalog item type should be registered to CatalogMemberFactory. It can be done in `TerriaMap/index.js`:

```diff
// ...
+ import CatalogMemberFactory from "terriajs/lib/Models/Catalog/CatalogMemberFactory";

registerCatalogMembers();
+ // Register the custom catalog item type.
+ CatalogMemberFactory.register(
+   PrimitiveCatalogItem.type,
+   PrimitiveCatalogItem
+ );
// ...
```

Finally, in your catalog file:

```js
{
	"catalog": [
		{
			"type": "primitive",
			"name": "Primitive Test"
		}
	]
}
```

This will give us an output like the following:

TerriaMap rendering a Cesium primitive

### Checklist

    - This PR does not modify any existing behavior.


lib/ModelMixins/MappableMixin.ts
```diff
@@ -34,10 +41,8 @@ export namespace ImageryParts {
   }
 }
 
-export function isCesium3DTileset(
-  mapItem: MapItem
-): mapItem is Cesium3DTileset {
-  return "allTilesLoaded" in mapItem;
+export function isPrimitive(mapItem: MapItem): mapItem is AbstractPrimitive {
+  return "destroy" in mapItem;
```
tephenavies: I think this could misclassify MapItems. E.g. I checked code for KmlDataSource and [it has a `destroy` function](https://github.com/CesiumGS/cesium/blob/6e37166f002620f2e159eacfa12524d84e53b9df/Source/DataSources/KmlDataSource.js#L3880). This type guard function needs to discriminate between Primitives and other types of MapItems in all cases.

If it's not possible to make type guards that can always classify arbitrary DataSources and Primitives we may need to change the type of MapItem to add an enum property to tag the MapItem type, or something similar.

tephenavies (changes requested):
Thank you for the PR! I think there's just 1 issue with merging this.

wibily:
cripsy to chase up

ykiu:
Sorry for being silent for such a long time. I'm having a heavy workload these days and haven't been able to give enough thoughts on this :(
But I'm definitely interested in moving this forward so I'll make appropriate changes as soon as I find time to do so!

ykiu:
Hi @steve9164. Sorry again for not being able to get back to you sooner.

So, I looked for properties that can distinguish primitives from other types, and found the only properties that all primitives have in common are `show`, `destroy()`, and `isDestroyed()`.

As you pointed out, DataSources have `destroy`. They also have `show`, so the only thing that's unique to primitives is `isDestroyed()`. A quick search in the source code of terriajs and Cesium reveals no other types that count as a MapItem have `isDestroyed`. So I think this can be used as a distinguisher.

Also, while testing with KML datasets, I noticed that Cesium implicitly adds a primitive to `this.scene.primitives` when we add a data source like KML. This means we cannot remove a primitive just because it's not in `this._allMapItems`. Rather, we need to take a diff of `this._allMapItems` before and after a change and detect what's removed. I added a commit that does that, too.

Regarding your suggestion:
> If it's not possible to make type guards that can always classify arbitrary DataSources and Primitives we may need to change the type of MapItem to add an enum property to tag the MapItem type, or something similar.

I see the point of it is to change the definition of type `MapItem` to make it look like this:

```
type MapItem = 
    | { type: "ImageryParts" item: ImageryParts }
    | { type: "DataSource" item: DataSource }
    | { type: "Primitive" item: Primitive }
    | { type: "TerrainProvider" item: TerrainProvider }
```

and look at the `type` property to distinguish types (correct me if I'm getting it wrong). I like this approach but at the same time it requires large refactoring so perhaps it deserves a separate PR?

tephenavies (approved):

PR #5936, opened 2021-10-25
Fix a bug where share URLs become long in some cases

ykiu:
### What this PR does

This PR fixes a bug introduced by #5705 where share URLs become long when there are many CompositeCatalogItems in the catalog.

#5705 made CompositeCatalogItem set the `show` flag to the user stratum of child members. This resulted in all the members of CompositeCatalogItems having a user stratum. Because BuildShareLink.js embeds all the models with a user stratum defined into the share data, this led to longer share URLs as there are more composite items in the catalog.

This PR fixes this problem by syncing the `show` flag on a per-stratum basis. This way, the user stratum of members remains empty until the user toggles the visibility checkbox in the UI, a behavior BuildShareLink.js seems to expect.

p.s. this should fix #5851.

### Checklist


KeyboardSounds (changes requested):
Hi there @ykiu! Thanks for your PR. Sorry it took us so long to get to back to you. 

I'm not sure this is documented anywhere (sorry again 😞), but the underlying principle behind the stratum system in TerriaJS is to keep track of which values come from where. The approach you've taken here, while effective, doesn't fit with this. It ignores what the different strata layers mean and obscures the original values by setting "show" in every stratum.

The easy solution in this case is instead of setting "show" in each of the composite's children in every stratum that exists already, just set it in the `underride` stratum. Only the `user` stratum gets serialized in share urls, so this will avoid them getting long and slow, but showing and hiding composite catalog items should still work.

ykiu:
No problem at all, I my self have been experiencing hectic weeks lately and haven't been able to afford my time on this.

And thanks for explaining how stratum system works. I'll fix the code as you suggested and push changes, hopefully early next week.

Thanks!

ykiu:
Hi @KeyboardSounds! I just fixed it. I hope it is better now.

KeyboardSounds (approved):

KeyboardSounds:
Thanks for being so patient with us @ykiu! And for the PR :)

PR #5940, opened 2021-10-26
Table improvments for geojson styling 

nf-s:
### Why

To make default styling more intuitive (mainly for new geojson stuff)

#### Better default styles

**Note "Disable style" option in styles dimension selector**

##### Example 1

http://ci.terria.io/table-improve-geojson/#configUrl=https://gist.githubusercontent.com/nf-s/b8b839e77a1bc34b731de15a36eb5429/raw/d3a311ed3282a90adbd63f2d697cf3a92d127f4a/ckan-vic-config.json&share=s-wM82r6qWfXOopInppLxmCf4vjs3

vs 

http://ci.terria.io/main/#configUrl=https://gist.githubusercontent.com/nf-s/b8b839e77a1bc34b731de15a36eb5429/raw/d3a311ed3282a90adbd63f2d697cf3a92d127f4a/ckan-vic-config.json&share=s-wM82r6qWfXOopInppLxmCf4vjs3

##### Example 2

http://ci.terria.io/table-improve-geojson/#configUrl=https://gist.githubusercontent.com/nf-s/b8b839e77a1bc34b731de15a36eb5429/raw/d3a311ed3282a90adbd63f2d697cf3a92d127f4a/ckan-vic-config.json&share=s-nir7lytLVrknIL0RSTFhLZrdCqM 

vs 

http://ci.terria.io/main/#configUrl=https://gist.githubusercontent.com/nf-s/b8b839e77a1bc34b731de15a36eb5429/raw/d3a311ed3282a90adbd63f2d697cf3a92d127f4a/ckan-vic-config.json&share=s-nir7lytLVrknIL0RSTFhLZrdCqM

#### Disable time selector and Time label

image

**Note will not work until checkboxes are fixed**

http://ci.terria.io/table-improve-geojson/#configUrl=https://nationalmap.gov.au/config.json&share=s-kTGK5cUA66zr0IflQFDexCKMCyn

vs 

http://ci.terria.io/main/#configUrl=https://nationalmap.gov.au/config.json&share=s-kTGK5cUA66zr0IflQFDexCKMCyn

#### Basemap contrast color and Remove border from regions with no data

image

http://ci.terria.io/table-improve-geojson/#share=s-8oYjqY5HrdjcgWnC6c3uYX3trvF

vs

http://ci.terria.io/main/#share=s-8oYjqY5HrdjcgWnC6c3uYX3trvF

### Changes

For this PR:
- [x] https://github.com/TerriaJS/terriajs/pull/5900

* Refactored `TableColumn get type()` to move logic into `guessColumnTypeFromValues()`
* `TableMixin.activeStyle` will set `TableColumnType = hidden` for `scalar` columns with name `"id"`, `"_id_"` or `"fid"`
* Fix bug `TableColumn.type = scalar` even if there were no values.
* Treat `TableColumnType.height` as `scalar` until we actually do something with the height data.
* Table columns named `"easting"` and `"northing"` are now hidden by default from styles
* `TableColumn.type = enum` requires at least 2 unique values (including null) to be selected by default
* Tweak automatic `TableColumn.type = Enum` for wider range of values
* Exporting `TableMixin` will now add proper file extensions
* Added `TimeVaryingTraits.timeLabel` trait to change label on `DateTimeSelectorSection` (defaults to "Time:")
  * This is set to `timeColumn.title`
* `TableColumn` will try to generate prettier `title` by un-camel casing, removing underscores and capitalising words
* `TableStyle` `startDates`, `finishDates` and `timeIntervals` will only set values for valid `rowGroups` (invalid rows will be set to `null`). For example, this means that rows with invalid regions will be ignored.
* Add "Disable style" option to `TableMixin.styleDimensions` - it can be enabled with `TableTraits.showDisableStyleOption`
* Added `timeDisableDimension` to `TableMixin` - this will render a checkbox to disable time dimension if `rowGroups` only have a single time interval per group (i.e. features aren't "moving" across time) - it can be enabled with `TableTraits.showDisableTimeOption` - `TableAutomaticStylesStratum` will automatically enable this if at least 50% of rowGroups only have one unique time interval (i.e. they don't change over time)\
* Remove border from region mapping if no data
* Add `baseMapContrastColor` and `constrastColor` to `BaseMapModel`
* Fixed `TableMixin.defaultTableStyle.legends` - `defaultTableStyle` is now not observable - it is created once in the `contructor`
* Removed `Terria.configParameters.enableGeojsonMvt` - geojson-vt/Protomaps is now used by default
* `GpxCatalogItem` now uses `GeojsonMixin`

### Checklist


lib/ModelMixins/TableMixin.ts
```diff
@@ -48,6 +49,24 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
...
+
+      const tableStyle = new TableStyle(this, -1);
+      tableStyle.colorTraits.setTrait(
+        CommonStrata.defaults,
+        "legend",
+        new ColorStyleLegend(this, -1)
+      );
+      this.defaultTableStyle = tableStyle;
+    }
+
```
nf-s: I have moved this here as it doesn't need to be observable. This also simplifies `tableAutomaticStyleStratum` a bit

lib/ModelMixins/TableMixin.ts
```diff
@@ -195,12 +205,6 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
       );
     }
 
-    @computed
-    get disableOpacityControl() {
-      // disable opacity control for point tables - or if no mapItems
-      return this.activeTableStyle.isPoints() || this.mapItems.length === 0;
-    }
-
```
nf-s, line 203: Just moved this down the file

lib/Table/TableAutomaticStylesStratum.ts
```diff
@@ -136,13 +137,6 @@ export default class TableAutomaticStylesStratum extends LoadableStratum(
...
     }
 
-    // If still no styles - try to create a style using text columns
-    if (columns.length === 0) {
-      columns = this.catalogItem.tableColumns.filter(
-        column => column.type === TableColumnType.text
-      );
-    }
-
     return columns.map((column, i) =>
```
nf-s, line 140: We don't need this now, as `TableMixin.defaultTableStyle` will handle it https://github.com/TerriaJS/terriajs/pull/5940/files#r756501263

lib/Language/en/translation.json
```diff
@@ -1297,8 +1298,11 @@
       "bulkGeocoderErrorMessage": "Unable to map addresses to lat-long coordinates, as an error occurred while retrieving address coordinates. Please check your internet connection or try again later.",
       "legendZFilterLabel": "Extreme values",
       "legendNullLabel": "(No value)",
+      "styleDisabledLabel": "Disable style",
```
wibily, line 1301: Hmm, to me Display Variable: Disable style doesn't read very well. Furthermore, in the bike path example, selecting disable style does more than just disabling the style, it shows all the data. I would suggest that it be reworded to "Show all" with the disabling of the style being a consequence of being unable to cram multiple legends when showing all the data

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -614,48 +617,47 @@ function GeoJsonMixin<T extends Constructor<Model<GeoJsonTraits>>>(Base: T) {
...
       const options = {
         describe: describeWithoutUnderscores,
-        markerSize: parseMarkerSize(style["marker-size"]) ?? 20,
-        markerSymbol: style["marker-symbol"], // and undefined if none
-        markerColor: defaultColor(style["marker-color"], this.name ?? ""),
-        stroke: getColor(style.stroke ?? "#000000"),
-        polygonStroke: getColor(style.stroke ?? "#000000"),
-        polylineStroke: defaultColor(style.stroke, this.name || ""),
-        strokeWidth: style["stroke-width"] ?? 2,
+        markerSize: parseMarkerSize(this.style["marker-size"]) ?? 20,
```
wibily: an optional suggestion would be to have a `const DEFAULTS = {...}` object with the same type as this options object. That way, we can ensure that all the fields in option have a fallback value and it separates out the default value from the fallback logic a bit.

lib/ModelMixins/TableMixin.ts
```diff
@@ -48,6 +49,24 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
...
+    /**
+     * The default {@link TableStyle}, which is used for styling
+     * only when there are no styles defined.
+     */
+    private readonly defaultTableStyle: TableStyle;
+
+    constructor(...args: any[]) {
+      super(...args);
+
+      const tableStyle = new TableStyle(this, -1);
```
wibily: I know that this isn't part of the changed code, but can we make the stylenumber an optional parameter in TableStyle so that if a styleNumber isn't provided, it falls back to a default style? I had to spend a bit of time figuring out what a style -1 was

lib/ModelMixins/TableMixin.ts
```diff
@@ -460,6 +477,10 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
             };
           }),
         selectedId: this.activeStyle,
+        allowUndefined: this.showDisableStyleOption ? true : undefined,
```
wibily: do we need the conditional or can we just set this to `allowUndefined: !!this.showDisableStyleOption`

lib/Table/TableAutomaticStylesStratum.ts
```diff
@@ -166,11 +160,54 @@ export default class TableAutomaticStylesStratum extends LoadableStratum(
...
+    // Return true if at least 50% of rowGroups only have one unique time interval (i.e. they don't change over time)
+    let flat = 0;
+
+    for (
+      let i = 0;
+      i < this.catalogItem.activeTableStyle.rowGroups.length;
+      i++
+    ) {
+      const group = this.catalogItem.activeTableStyle.rowGroups[i];
+      const dates = group[1]
```
wibily: is there something special about group[1] and can we be sure that rowGroups[i][1] will always exist?

lib/Table/TableAutomaticStylesStratum.ts
```diff
@@ -166,11 +160,54 @@ export default class TableAutomaticStylesStratum extends LoadableStratum(
...
+        .map(
+          row =>
+            this.catalogItem.activeTableStyle.timeColumn?.valuesAsDates.values[
+              row
+            ]
+        )
+        .filter(date => date) as Date[];
+      if (uniq(dates).length <= 1) flat++;
```
wibily, line 190: i don't think uniq works with native array of Dates. https://jsfiddle.net/zm5yd2j7/

lib/Table/TableColumn.ts
```diff
@@ -592,7 +603,16 @@ export default class TableColumn {
...
+      // If no title set, use `name` and:
+      // - un-camel case
+      // - remove underscores
+      // - capitalise
       this.name
+        .replace(/[A-Z][a-z]/g, letter => ` ${letter.toLowerCase()}`)
+        .replace(/_/g, " ")
+        .trim()
+        .toLowerCase()
+        .replace(/(^\w|\s\w)/g, m => m.toUpperCase())
```
wibily, line 615: since we have lodash in our package.json already, we could do `_.flow([_.lowerCase, _.capitalize])(this.name)`

lib/Table/TableColumn.ts
```diff
@@ -822,20 +815,85 @@ export default class TableColumn {
     return nullFunction;
   }
 
+  private guessColumnTypeFromValues(): TableColumnType {
+    let type: TableColumnType | undefined;
```
wibily, line 819: this is some dark magic here

lib/Table/TableStyle.ts
```diff
@@ -513,9 +537,14 @@ export default class TableStyle {
         ? this.timeColumn.valuesAsJulianDates.maximum
         : sortedStartDates[sortedStartDates.length - 1];
 
-    return startDates.map(date => {
+    const finishDates: (JulianDate | null)[] = new Array(
```
wibily, line 542: what's the benefit of doing it this way over the old map?

lib/Table/TableStyle.ts
```diff
@@ -413,21 +433,25 @@ export default class TableStyle {
     }
 
     const startDates = timeColumn.valuesAsJulianDates.values;
+    const finishDates: (JulianDate | null)[] = new Array(
```
wibily, line 438: hmm, what's the difference between this and the old code?

wibily (approved):

lib/Language/en/translation.json
nf-s: I agree, I don't like the wording very much.

But in most cases, if style is disabled, you won't seen any more features, everything will just be the same color.
So I'm not sure about "Show all" as an alternative.

What about "No style" or "None"?

lib/ModelMixins/GeojsonMixin.ts
nf-s: Yeah good idea!

lib/ModelMixins/TableMixin.ts
nf-s: Done!

lib/ModelMixins/TableMixin.ts
nf-s: 👍

lib/Table/TableAutomaticStylesStratum.ts
nf-s: I have renamed same vars to make it a bit more clear - `rowGroups[i]: [string, string]`

So i've now made it

```ts
 const [rowGroupId, rowIds] = this.catalogItem.activeTableStyle.rowGroups[i]
```

lib/Table/TableAutomaticStylesStratum.ts
nf-s: Yeah nice pickup!
I have changed it to date -> date.getTime() and it now works

lib/Table/TableColumn.ts
nf-s: I prefer to minimise lodash usage, as I hope to remove it one day 😄

lib/Table/TableColumn.ts
nf-s: Yeah it is a bit like that

lib/Table/TableStyle.ts
nf-s: Performance, we sometimes have a million dates or so

lib/Table/TableStyle.ts
nf-s: Same as above, these two bits of code can be really slow with lots of dates.
Seems to be a bit faster now using `for` loops instead of `forEach` and `map`

lib/Language/en/translation.json
wibily: I'd prefer "Disable style" over "None" since Display Variable: None reads like nothing should be shown on the screen. Perhaps this should be a checkbox option instead of adding it to the drop down but we can always do that later

wibily (approved):

PR #5948, opened 2021-10-28
Set WMS layers property from CKAN resource url query string

### LINKED ISSUE #1090: Most VIC data portal WMS datasets (Either via "Open in NationalMap" button or Built-in national map catalog)
### Most VIC data portal WMS datasets (Either via "Open in NationalMap" button or Built-in national map catalog)

E.g. this one: 

https://data.gov.au/dataset/ds-vic-4ce61bcf-f5f2-4f8a-bbe7-4d5a3c07b918

It renders well in Magda map preview but when you open it in nationalMap, it will complaint about:

http status 494 (Request too big)

image

### My theory / Investigation
 
It looks like the original dataset url from vic is like the following (Note it contains the `LAYERS` parameters):

http://services.land.vic.gov.au/catalogue/publicproxy/guest/dv_geoserver/wms?SERVICE=WMS&VERSION=1.1.1&REQUEST=GetMap&WIDTH=512&HEIGHT=512&LAYERS=ENERGY_ELECTRICITY_DISTRIBUTOR&STYLES=&FORMAT=image%2Fpng&SRS=EPSG%3A4283&BBOX=141%2C-39%2C150%2C-34

And seems when the url is add to nationalMap (either via open in national map button or manually add in national map), the `layers` parameter is not utilised.

Because of it, the nationalMap will try to fetch all layers --- which lead to the Request too big error.

### Impact

It seems all VIC WMS dataset are structure in that way. So it will impact all vic WMS datasets on data.gov.au.

Furthermore, seems any of VIC nationalMap built-in catalog also not work (for the same error).

image

tephenavies:
### What this PR does

Fixes TerriaJS/nationalmap#1090.

Test here: http://ci.terria.io/ckan-wms-url-layers/#clean&https://gist.githubusercontent.com/steve9164/bdd8cd04ea0ca32ceb851c3b462089c9/raw/852e90fbf7ed8d9a0bc53817c877f5b51c83313f/test-ckan-resource.json

Note the item doesn't work due to dodgy CORS on the VIC server (for some reason this request https://discover.data.vic.gov.au/api/3/action/resource_show?id=7fc14241-b1ff-4fef-a336-c885f9060b57 doesn't have CORS, but a lot of other APIs at the same URL have CORS).

It'd be a good idea to do similar for MagdaReference.

### Checklist

### DIFF
--- a/test/Models/Catalog/Ckan/CkanItemReferenceSpec.ts
+++ b/test/Models/Catalog/Ckan/CkanItemReferenceSpec.ts
@@ -1,28 +1,16 @@
-import { configure, runInAction } from "mobx";
-import _loadWithXhr from "../../../../lib/Core/loadWithXhr";
-import Terria from "../../../../lib/Models/Terria";
-import CommonStrata from "../../../../lib/Models/Definition/CommonStrata";
 import i18next from "i18next";
+import { runInAction } from "mobx";
 import CkanItemReference, {
   CkanDatasetStratum
 } from "../../../../lib/Models/Catalog/Ckan/CkanItemReference";
 import WebMapServiceCatalogItem from "../../../../lib/Models/Catalog/Ows/WebMapServiceCatalogItem";
-import InfoSectionTraits from "../../../../lib/Traits/TraitsClasses/CatalogMemberTraits";
+import Terria from "../../../../lib/Models/Terria";
 
-configure({
-  enforceActions: "observed",
-  computedRequiresReaction: true
-});
-
-interface ExtendedLoadWithXhr {
-  (): any;
-  load: { (...args: any[]): any; calls: any };
-}
-
-const loadWithXhr: ExtendedLoadWithXhr = <any>_loadWithXhr;
+const taxationStatisticsPackage = require("../../../../wwwroot/test/CKAN/taxation-statistics-package.json");
+const taxationStatisticsWmsResource = require("../../../../wwwroot/test/CKAN/taxation-statistics-wms-resource.json");
+const vicWmsLayerResource = require("../../../../wwwroot/test/CKAN/vic-wms-layer-resource.json");
 
 describe("CkanItemReference", function() {
-  const ckanServerUrl = "http://data.gov.au";
   let terria: Terria;
   let ckanItemReference: CkanItemReference;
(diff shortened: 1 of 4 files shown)


test/Models/Catalog/Ckan/CkanItemReferenceSpec.ts
```diff
@@ -34,17 +22,33 @@ describe("CkanItemReference", function() {
...
+    jasmine.Ajax.stubRequest(
+      "https://example.com/api/3/action/resource_show?id=1234"
+    ).andReturn({
+      responseText: JSON.stringify(taxationStatisticsWmsResource)
+    });
+
+    jasmine.Ajax.stubRequest(
+      "https://example.com/api/3/action/resource_show?id=5678"
+    ).andReturn({
+      responseText: JSON.stringify(vicWmsLayerResource)
```
tephenavies, line 44: Sometimes the rabbit hole chooses you, but hey I think this is nicer now that I've changed it. Now we know CKAN is making exactly the right requests. Also the fail and log requests by default bit is pretty nice for debugging tests while you've got requests stubbed.

test/Models/Catalog/Ckan/CkanItemReferenceSpec.ts
```diff
@@ -53,11 +57,11 @@ describe("CkanItemReference", function() {
   describe("Can load an item by datasetId - ", function() {
     beforeEach(async function() {
       runInAction(() => {
-        ckanItemReference.setTrait("definition", "url", "somedataset");
+        ckanItemReference.setTrait("definition", "url", "https://example.com");
         ckanItemReference.setTrait("definition", "name", "Taxation Statistics");
         ckanItemReference.setTrait("definition", "datasetId", "1234");
       });
-      await ckanItemReference.loadReference();
+      (await ckanItemReference.loadReference()).throwIfError();
```
tephenavies, line 64: I want to reject the promise in `beforeEach` if the item can't load. It's probably worth checking (in Result.ts) if this is the right thing to do.

t83714 (approved):
Looks good to me~
And the test cases also looks neater now 💯  👍 
We also tested in CI with vic ckan data

PR #5959, opened 2021-11-01
TSify split screen control and remove dependency on some deprecated CSS styles.

na9da:
### What this PR does

#5826 removed some css styling that the splitter control depended on which was confusing in the first place. I have tsified the splitter control and removed the dependency. 

### Testing
[Before](http://ci.terria.io/main/#share=s-eJJ6JyFcWszPl549QTIBz3I2TA8) vs [After](http://ci.terria.io/tsify-splitscreen-control/#share=s-eJJ6JyFcWszPl549QTIBz3I2TA8)

### Checklist


lib/ReactViews/Workbench/WorkbenchSplitScreen.tsx
```diff
@@ -0,0 +1,224 @@
...
+    `
+    background-color: ${p.theme.colorSecondary};
+  `}
+  &:hover,
+  &:focus {
+    background-color: ${p => p.theme.colorSecondary};
+  }
+`;
+
+export default WorkbenchSplitScreen;
```
na9da, line 224: I have TSified this file and removed dependency on `scss`

nf-s (approved):
Looks good!

PR #5972, opened 2021-11-11
Support catalog refresh after login

### LINKED ISSUE #5973: Support catalog refresh after login
Related to https://github.com/TerriaJS/vic-digital-twin/issues/201

mwu2018:
### What this PR does

Fixes #5973 

Related to https://github.com/TerriaJS/vic-digital-twin/issues/201

Add a method `refreshCatalogMembersFromMagda` to `Terria` class. It optionally reloads `map-config` record with `group` aspect. Reloading `group` aspect of `map-config` record is required because some top level catalogue items (including groups) originated from magda registry might be subject to access control, different users might have different catalogue items.  This method should be called after user signing in so that those authorised only top level catalogue items can be shown automatically without refreshing the page.

A demo is available at https://dev.vic.digitaltwin.terria.io/.

### Checklist


nf-s (approved):
Looks good!

PR #5975, opened 2021-11-11
Add maincode aliases for all SA regions

### LINKED ISSUE #5974: when region mapping SA codes we should recognise MAINCODE
At the moment when we region map SA codes we recognise columns named e.g.:

```
SA2
SA2_CODE
SA2_CODE_2016
```
See e.g.:

https://github.com/TerriaJS/terriajs/blob/main/wwwroot/data/regionMapping.json#L385-L389

However, the official ABS way to refer to this region type is `SA2_MAINCODE_2016`, which is the only one that will not work. We should add that alias too.

See ABS pages here: https://www.abs.gov.au/Ausstats/ABS@.nsf/7d12b0f6763c78caca257061001cc588/2d50fc5d7082cf99ca257801000dcd3d!OpenDocument

meh9:
### What this PR does

Fixes terriajs#5974

I've added `maincode` aliases for all SA regions that were missing them to ensure we support the default ABS column naming.

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - no idea
-   [ ] I've updated relevant documentation in `doc/`. - none required

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@ Change Log
 ==========
 
 #### next release (8.1.10)
+* Added `MAINCODE` aliases for all ABS Statistical Area regions that were missing them.
 * [The next improvement]
 
 #### 8.1.9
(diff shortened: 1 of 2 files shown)


na9da (approved):
Looks good @meh9 :rocket:

PR #5981, opened 2021-11-15
Make zoom controls bigger and change the icons

na9da:
### What this PR does

Made the zoom controls bigger and changed the icons to suit the larger size. Also changed the zoom reset icon to home icon. This is to make the reset behavior more discoverable.

### Checklist


lib/ReactViews/Map/Navigation/Items/ZoomControl.tsx
```diff
@@ -222,15 +222,18 @@ const StyledZoomControl = styled(Box).attrs(props => ({
...
-  styledWidth: "24px",
-  styledHeight: "60px"
+  styledWidth: "32px"
 }))`
   border-radius: 100px;
   svg {
-    height: 10px;
-    width: 10px;
+    height: 20px;
+    width: 20px;
```
na9da, line 231: Same dimensions as other map nav icons.

nf-s (approved):
Looks good!

Can you please update `CHANGES` before merging

nf-s (changes requested):
Actually, I think the colors might be slightly different. All other icons are `#4d5766`

image

Is that intentional?

na9da:
Fixed the icon colors.

nf-s (approved):

PR #5986, opened 2021-11-16
Closer zoom

na9da:
### What this PR does

Modified "ideal zoom" to zoom closer to 3d tileset and data sources.

### Checklist


nf-s (approved):
Looks good!

PR #5987, opened 2021-11-16
Ckan improvements

nf-s:
### Ckan improvements

* Moved CKAN default `supportedFormats` to `CkanDefaultFormatsStratum`
* Add properties to `CkanResourceFormatTraits`
  * `maxFileSize` to filter out resources with large files (default values: GeoJSON = 150MB, KML = 30MB, CZML = 50MB)
  * `removeDuplicates` (which defaults to true) so we don't get duplicate formats for a dataset (it will check `resource.name`)
    * If there are multiple matches, then the newest (from resource.created property) will be used
  * `onlyUseIfSoleResource` to give a given resource format unless that is all that exists for a dataset
* Add CKAN `useSingleResource`, if `true`, then the highest match from `supportedResourceFormats` will be used for each dataset
* ArcGis Map/Feature Service will now set CRS from `latestWkid` if it exists (over `wkid`)
* Fix CKAN ArcGisFeatureService resources

### Checklist


lib/Traits/TraitsClasses/CkanResourceFormatTraits.ts
```diff
@@ -14,11 +14,41 @@ export default class CkanResourceFormatTraits extends ModelTraits {
...
+  })
+  urlRegex?: string;
+
+  @primitiveTrait({
+    name: "Maximum file size (MB)",
+    description:
+      "The maximum allowed file size for this format (in megabytes).",
+    type: "number"
+  })
+  maxFileSize?: number;
```
na9da: Should this be nullable for it to be able to override the defaults?

nf-s: Yeah good idea

nf-s: done!

lib/Models/Catalog/Ckan/CkanCatalogGroup.ts
```diff
@@ -246,35 +258,113 @@ export class CkanServerStratum extends LoadableStratum(CkanCatalogGroupTraits) {
...
+              dupe !== -1 &&
+              uniqueResources[dupe].resource.created <
+                currentResource.resource.created
+            ) {
+              uniqueResources[dupe] = currentResource;
+            } else if (dupe === -1) {
+              uniqueResources.push(currentResource);
+            }
+            return uniqueResources;
+          }, []);
```
na9da: Not critical but just a nitpick - if the accumulator here is a `{}` instead of a `[]` we can get rid of the `findIndex` and trivially check the existence of the item. Then use `Object.values()` on the result of the reduce.

nf-s: Yeah that is nicer, I have changed to `{}`

na9da (approved):
Looks great @nf-s!

Issue #5962, opened 2021-11-02
remember the state of the "use native device resolution" checkbox 


meh9:
Because I have a high resolution screen I always want to have "use native device resolution" enabled. At the moment I have to manually enable that every session. Let's make it persistent on a per-browser basis using a cookie or local-storage.

Bonus points if you do the same for the Quality slider!

<img width="320" alt="Screen Shot 2021-11-02 at 11 12 40" src="https://user-images.githubusercontent.com/5283003/139758979-6619ecf7-dc79-4027-b86b-eff9fa7734e2.png">

meh9:
Zoran fixed it!

PR #5990, opened 2021-11-18
Reenable use native device resolution on mobile 

### LINKED ISSUE #5989: new terria version doesn't show "Use Native Device Resolution" on mobile
The latest version of Terria is for some reason not showing the "Use Native Device Resolution" setting on mobile, but it's still there on a computer. Please enable the setting on mobile again as it does improve the rendering quality (I just used it a couple of days ago).

On computer:
Screen Shot 2021-11-18 at 09 18 26

On mobile:
Screenshot_2021-11-18-09-18-03-86_3aea4af51f236e4932235fdada7d1643

zoran995:
### What this PR does

Fixes #5989, #5962

Reenable "Use native device resolution" on mobile and provides an option to save them in local storage.

### Checklist


lib/Models/Terria.ts
```diff
@@ -882,6 +882,33 @@ export default class Terria {
...
+      this.setUseNativeResolution(useNativeResolution);
+    }
+
+    const baseMaximumScreenSpaceError = parseFloat(
+      this.getLocalProperty("baseMaximumScreenSpaceError")?.toString() || ""
+    );
+    if (
+      isDefined(baseMaximumScreenSpaceError) &&
+      baseMaximumScreenSpaceError !== null &&
+      baseMaximumScreenSpaceError !== NaN
```
tephenavies: This comparison will not work (`NaN` cannot be used in comparisons, e.g. `NaN !== NaN` is `true`). I think `!isNaN(baseMaximumScreenSpaceError)` is sufficient (since `parseFloat` must return `NaN` or a valid number).

lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -128,6 +128,26 @@ class SettingPanel extends React.Component<PropTypes> {
     this.props.terria.currentViewer.notifyRepaintRequired();
   }
 
+  onBaseMaximumScreenSpaceErrorChange(bmsse: number) {
+    this.props.terria.setBaseMaximumScreenSpaceError(bmsse);
+    if (bmsse) {
```
tephenavies: When would this condition be false? Can bmsse be 0? If it's a 0 comparison I think it's better to be explicit `bmsse !== 0`.

tephenavies (changes requested):

tephenavies:
We want to merge this for a release today, so I'll fix up those couple of things and get it in.

tephenavies (approved):

meh9:
@zoran995 this was awesome, thanks mate! :D

Amazing quality of life improvement for me!

PR #5993, opened 2021-11-22
Handle external link

mwu2018:
### What this PR does

Fixes https://github.com/TerriaJS/vic-digital-twin/issues/147
Add an external link icon to any hyperlink that will open a new tag or window.

The feature can be verified at https://dev.vic.digitaltwin.terria.io/#share=s-lPTZsucouqFYPVfx7hNWW0fqUQL

### Checklist

### DIFF
--- a/lib/ReactViews/Custom/parseCustomHtmlToReact.js
+++ b/lib/ReactViews/Custom/parseCustomHtmlToReact.js
@@ -36,10 +36,8 @@ const ExternalLinkIcon = styled(StyledIcon).attrs({
   displayInline: true
 })`
   margin-left: 5px;
-  fill: ${props => props.theme.colorPrimary};
+  fill: currentColor;
 `;
-
-const externalIcon = React.createElement(ExternalLinkIcon, {});
 
 const isExternalLink = url => {
   const tmp = document.createElement("a");
@@ -81,6 +79,7 @@ function getProcessingInstructions(context) {
         context.disableExternalLinkIcon !== true &&
         isExternalLink(node.attribs.href)
       ) {
+        const externalIcon = React.createElement(ExternalLinkIcon, {});
         children.push(externalIcon);
       }
 
--- a/lib/ReactViews/Custom/parseCustomHtmlToReact.js
+++ b/lib/ReactViews/Custom/parseCustomHtmlToReact.js
@@ -66,7 +66,7 @@ function getProcessingInstructions(context) {
       };
       node.attribs = combine(node.attribs, elementProps);
 
-      if (isExternalLink(node.attribs.href)) {
+      if (context.disableIcon !== true && isExternalLink(node.attribs.href)) {
         children.push(externalLinkIconReact);
       }
 
--- a/lib/ReactViews/Custom/parseCustomHtmlToReact.js
+++ b/lib/ReactViews/Custom/parseCustomHtmlToReact.js
@@ -39,12 +39,6 @@ const ExternalLinkIcon = styled(StyledIcon).attrs({
   fill: currentColor;
 `;
 
-const isExternalLink = url => {
-  const tmp = document.createElement("a");
-  tmp.href = url;
-  return tmp.host !== window.location.host;
-};
-
 /**
  * @private
  */
@@ -62,6 +56,13 @@ function getProcessingInstructions(context) {
       processNode: customComponent.processNode.bind(customComponent, context)
     });
   }
+
+  // Has to put this inside the function. Otherwise may cause unexpected result.
+  const isExternalLink = url => {
+    const tmp = document.createElement("a");
+    tmp.href = url;
+    return tmp.host !== window.location.host;
+  };
 
   // Make sure any <a href> tags open in a new window
   processingInstructions.push({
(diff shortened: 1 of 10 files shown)


na9da:
Hey @mike.wu - good start, just a few things I noticed:

* We should test for external link before enabling this. So you may want to write a small function to do the test - here is one way of doing it (https://www.designcise.com/web/tutorial/how-to-check-if-a-string-url-refers-to-an-external-link-using-javascript)
* We already have an external link icon available here - https://github.com/TerriaJS/terriajs/blob/main/wwwroot/images/icons/external-link.svg - so might want to check if you can reuse it.
* Since we might want to style it a bit, it might be better to implement this entirely using CSS ([here is an example](https://electrictoolbox.com/add-offsite-link-icon-after-external-links-css/)) instead of pushing a node to children. This can be done by tagging the element as an external link by passing `className: "external-link"` in the `elementProps` object and then  writing a new CSS rule for `a.external-link` in the [global section](https://github.com/TerriaJS/terriajs/blob/0aea0e1884bb8d059df96048ca3b7e2af1333b5c/lib/ReactViews/StandardUserInterface/standard-user-interface.scss#L15) of standard-user-interface.scss. 
* There has to be a way to disable this feature for things like branding logos for example. So maybe an option to turn it off should be accepted via the `context` parameter and passed when called from [Branding.tsx](https://github.com/TerriaJS/terriajs/blob/0aea0e1884bb8d059df96048ca3b7e2af1333b5c/lib/ReactViews/SidePanel/Branding.tsx#L84-L86).

mwu2018:
@philipgrimmett  @na9da The external link icon indicates if a new tab or window will be opened, regardless if it is the same domain or not, similar to the about page button. With this approach, the email links will also have the icons. With @na9da suggestion, we only add icons to true external links.

Which way do you prefer?

na9da:
Ah, now I feel that external link icon can be confused with open-in-new-tab. So i'm not sure if we should test for external link or whether it opens in new tab. I leave that decision to @philipgrimmett . But in that case we should tag as external only when `target !== "_self"` after the call to `combine()`. Because when `target` equals `_self`, the browser opens the link in the same window. This is sometimes done to reload the page when clicking the brand icon.

lib/ReactViews/Custom/parseCustomHtmlToReact.js
```diff
@@ -26,6 +29,24 @@ const shouldProcessEveryNodeExceptWhiteSpace = function(node) {
...
+  glyph: Icon.GLYPHS.externalLink,
+  styledWidth: "10px",
+  styledHeight: "10px",
+  displayInline: true
+})`
+  margin-left: 5px;
+  fill: ${props => props.theme.colorPrimary};
+`;
+
+const externalIcon = React.createElement(ExternalLinkIcon, {});
```
na9da: I think you should create this element inside the `processNode` call each time we need it - re-using the same react element might result in unexpected behavior.

na9da:
Looks good @mwu2018 - just that one comment. 

On the whole, I think the problem with this approach is that the icon style may not exactly match the link color everywhere. For example, look at the [workbench and feature info links](http://ci.terria.io/handle-external-link/#share=s-m5hvWgOwQtN8AbsowBKxl5wtXXQ&https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/dev/terria/dea-maps-v8.json).

I'm not sure if there is a nice way to fix it. The other approach is to open in new tab show an redirect message in the new tab. That'll need extending the server to add a `/redirect` endpoint.

zoran995:
@na9da Icon color can be fixed by setting `fill: currentColor` (the same approach is used for checkbox). To make it work in feature info, this style will have to be removed (don't see any issues/breaking colors when changing it in dev console)
https://github.com/TerriaJS/terriajs/blob/0aea0e1884bb8d059df96048ca3b7e2af1333b5c/lib/ReactViews/FeatureInfo/feature-info-panel.scss#L10-L12
 or use `!important`

na9da:
Ah nice tip on  `currentColor`. Thanks @zoran995 :heart:

mwu2018:
Yeah, I used currentColor in the original inline svg string.

KeyboardSounds:
I don't think external link component should be in `lib/ReactViews/Custom/parseCustomHtmlToReact.js`, it should be in its own file. It uses `parseCustomHtmlToReact`, but it's a separate thing unto itself, like our feature info tables.

wibily:
blocked by terriajs#6003

mwu2018:
> I don't think external link component should be in `lib/ReactViews/Custom/parseCustomHtmlToReact.js`, it should be in its own file. It uses `parseCustomHtmlToReact`, but it's a separate thing unto itself, like our feature info tables.

External links may appear anywhere in the page. They are all parsed by `parseCustomHtmlToReact.js`. Therefore adding the logic there should be ok, I think. Maybe we can add a global flag to enable or disable the feature of appending the icon (default to `disabled`, for example)?

KeyboardSounds:
> > I don't think external link component should be in `lib/ReactViews/Custom/parseCustomHtmlToReact.js`, it should be in its own file. It uses `parseCustomHtmlToReact`, but it's a separate thing unto itself, like our feature info tables.
> 
> External links may appear anywhere in the page. They are all parsed by `parseCustomHtmlToReact.js`. Therefore adding the logic there should be ok, I think. Maybe we can add a global flag to enable or disable the feature of appending the icon (default to `disabled`, for example)?

Ah yeah that's true, never mind then 😅

na9da (approved):
Looks good @mwu2018!

PR #6008, opened 2021-11-26
Tsify sendFeedback and cleanup errors/notifications

nf-s:
### Tsify sendFeedback and cleanup errors/notifications

After vs before

image

### Checklist


lib/Models/sendFeedback.ts
```diff
@@ -0,0 +1,105 @@
...
+        method: "POST",
+        data: JSON.stringify(feedbackData),
+        headers: {
+          "Content-Type": "application/json"
+        }
+      });
+    })
+    .then(function(json) {
+      if (json instanceof String) {
+        json = JSON.parse(json.toString());
```
wibily, line 70: this is subtly different to the original jsx version, although from a cursory test in the console `JSON.parse(new String('{"a":1}')` seems to work?

lib/Models/sendFeedback.ts
```diff
@@ -0,0 +1,105 @@
...
+  email: string;
+  sendShareURL: boolean;
+  comment: string;
+  additionalParameters?: Record<string, string | undefined>;
+}) {
+  if (!isDefined(options) || !isDefined(options.terria)) {
+    throw TerriaError.from("options.terria is required.");
+  }
+
+  var terria = options.terria;
```
wibily: could be a const, but its such a nitpick, i'm not fussed

wibily (approved):

lib/Models/sendFeedback.ts
nf-s: Yeah I was just trying to please the type checker here. Probably not necessary though.
I'm not sure how we'd even get a `String` here

nf-s: I might leave it, just to be careful

lib/Models/sendFeedback.ts
nf-s: 👍

PR #6009, opened 2021-11-29
Long overdue README update

tephenavies:
Update README. There's lots of outdated information, especially now after version 8 release. This should fix that.


meh9 (approved):
Looks good.

PR #6013, opened 2021-11-30
ApiGeoJsonCatalogItem

KeyboardSounds:
### What this PR does

### Checklist


lib/Models/Catalog/CatalogItems/GeoJsonCatalogItem.ts
```diff
@@ -74,19 +75,42 @@ class GeoJsonCatalogItem extends GeoJsonMixin(
...
+      fc = toFeatureCollection(
+        jsonData.map(item => {
+          let geojson = this.responseGeoJsonPath
+            ? _get(item, this.responseGeoJsonPath)
+            : item;
+          if (typeof geojson === "string") {
+            geojson = JSON.parse(geojson);
+          }
+          // add extra properties back to geojson so they appear in feature info
+          geojson.properties = item;
```
na9da, line 107: Doubt: Should we exclude the object at responseGeoJsonPath and include just the outer item if it is easy enough to do?

na9da (approved):
Looks great @KeyboardSounds

lib/Models/Catalog/CatalogItems/GeoJsonCatalogItem.ts
KeyboardSounds: ah yeah you're totally right let's do that

PR #6018, opened 2021-12-01
Catalog index improvements

nf-s:
### What this PR does

* Moved `CatalogIndex` loading from constructor (called in `Terria.start`) to `CatalogSearchProvider.doSearch` - this means the index will only be loaded when the user does their first search
* Add basic auth support to `generateCatalogIndex`, fix some bugs and improve performance
* Update terria-js cesium to `1.81.2`
* Add `uniqueId` as fallback to `nameInCatalog`
* Remove duplicated items from OpenDataSoftGroup and SocrataGroup

### After merge

- [ ] This will require a PR to TerriaMap to update terria-js cesium to `1.81.2`

### Checklist


lib/Models/SearchProviders/CatalogIndex.ts
```diff
@@ -31,10 +31,20 @@ export default class CatalogIndex {
     return this._searchIndex;
   }
 
-  readonly loadPromise: Promise<void>;
+  private _loadPromise: Promise<void> | undefined;
```
na9da, line 35: Should this be an `observable`? Considering we use it in a computed here:

https://github.com/TerriaJS/terriajs/blob/7510a87c0041f0ac4d2a9fe98dc4fc7b395c5ae5/lib/Models/SearchProviders/CatalogSearchProvider.ts#L118-L123

lib/Models/Catalog/CatalogGroups/OpenDataSoftCatalogGroup.ts
```diff
@@ -227,7 +233,10 @@ export class OpenDataSoftCatalogStratum extends LoadableStratum(
   }
 
   getDatasetId(dataset: ValidDataset) {
-    return `${this.catalogGroup.uniqueId}/${dataset.dataset_id}`;
+    // Use OpenDataSoft server hostname for datasets, so we don't create multiple across facets
+    return `${URI(this.catalogGroup.url ?? "").hostname()}/${
+      dataset.dataset_id
+    }`;
```
na9da: Doubt: Should we fallback to parent group id when hostname is empty?

buildprocess/generateDocs.ts
```diff
@@ -291,12 +299,12 @@ export default async function generateDocs() {
...
+    .map((section: any) => section["Connecting to Data"])
+    .filter((x: any) => x !== undefined)[0];
   const typeDetailsSubSection =
     connectingToDataSection &&
     connectingToDataSection.find(
-      (subSection:any) => "Catalog Type Details" in subSection
+      (subSection: any) => "Catalog Type Details" in subSection
     );
   if (typeDetailsSubSection === undefined) {
     throw new Error(
```
na9da, line 310: I guess all of these are just prettier changes.

nf-s: yep!

lib/Models/SearchProviders/CatalogIndex.ts
nf-s: Oh yeah good pickup!

lib/Models/Catalog/CatalogGroups/OpenDataSoftCatalogGroup.ts
nf-s: Yeah I have added the fallback. Practically it will never be empty, but good to be safe

buildprocess/generateCatalogIndex.ts
```diff
@@ -1,99 +1,245 @@
...
+
+    //   load.apply(this, args);
+    // };
+
+    // A fun method to add Auth headers to all requests with baseUrl
+    const newFetch = (
+      input: RequestInfo,
+      init?: RequestInit
+    ): Promise<Response> => {
+      const url = typeof input === "string" ? input : input.url;
```
na9da: I know this is experimental at this stage, but would be nice if we can at some point extract code for patching requests to a separate file for readability.

na9da (approved):
Looks great @nf-s!

buildprocess/generateCatalogIndex.ts
nf-s: Good idea! I've moved it to `buildprocess/patchNetworkRequests.ts`

mwu2018:
Should we merge this PR now?

lib/ModelMixins/CatalogMemberMixin.ts
```diff
@@ -95,7 +95,7 @@ function CatalogMemberMixin<T extends Constructor<CatalogMember>>(Base: T) {
 
     @computed
     get nameInCatalog(): string | undefined {
-      return super.nameInCatalog || this.name;
+      return super.nameInCatalog || this.name || this.uniqueId;
```
tephenavies, line 98: `nameInCatalog` should only be used in the catalog. It's not the official name of an item. Maybe  `name` should use `uniqueId` if there isn't a name present, and then `nameInCatalog` can stay as it's original definition.

tephenavies: Maybe that's not the best way to phrase it. Having a backup for `nameInCatalog` is great, but surely you'll still have blank labels in the terria UI once it's added. Or is this a "before reference is resolved" problem that's specific to the exploring the catalogue?

PR #6021, opened 2021-12-02
Generate catalog index bug

nf-s:
### Generate catalog index bug

I had removed `loadMembers` for `CatalogGroup` groups - so IDs weren't being refreshed correctly

### Checklist


na9da (approved):
Looks good!

PR #6029, opened 2021-12-06
Fix geojson legends

nf-s:
### Fix geojson legends

Fix `GeoJsonMixin.legend` for Cesium primitives (`TableMixin` will now check for `super.legends`, and it will not return legends unless it has data)

### Before vs After

image

### Checklist


nf-s:
No longer relevant - #6046

PR #6038, opened 2021-12-09
Warn user if downloading large files

### LINKED ISSUE #2638: Warn the user before loading large vector data files
As reported in TerriaJS/nationalmap#564, large vector data files can cause the browser to hang or crash.  The web platform doesn't give us great tools for defending against this, but we can at least set a conservative size limit, above which we warn the user before trying to load and display the file.

nf-s:
### Warn user if downloading large files

Fixes https://github.com/terriajs/terriajs/issues/2638

Adds `fileSizeToPrompt` and `maxFileSize` options:
- If request is larger than `fileSizeToPrompt`, then a prompt will appear asking the user if the want to "continue" the download (note it does not interrupt the download until the user presses "cancel)
- If request is larger than `maxFileSize`, then download will be canceled

The network request size is calculated in two ways:
- `content-length` from header (if it exists)
- counting the bytes while downloading

The messaging shown to user will change depending on how request size is calculated.
See test links for messaging if `content-length` header is missing

#### 250MB geojson

- http://ci.terria.io/max-file-size-download/#configUrl=https://gist.githubusercontent.com/nf-s/b8b839e77a1bc34b731de15a36eb5429/raw/d3a311ed3282a90adbd63f2d697cf3a92d127f4a/ckan-vic-config.json&share=s-6A2oemcb2pTtDHNWhJtczOUIMnj
- Contours 50cm - City of Greater Geelong

You will get two messages

image

image

#### 70MB geojson

- http://ci.terria.io/max-file-size-download/#configUrl=https://gist.githubusercontent.com/nf-s/b8b839e77a1bc34b731de15a36eb5429/raw/d3a311ed3282a90adbd63f2d697cf3a92d127f4a/ckan-vic-config.json&share=s-6A2oemcb2pTtDHNWhJtczOUIMnj
- Trees - City of Greater Geelong

You will get one message

image

#### Canceled download message

image

### Todo

- [ ] Replace load*** with fetch*** ?
- [ ] CSV
- [ ] KML
- [ ] Other files...
- [ ] Get Phil to word it nicely

### Checklist


sheelanvaez:
Hi @AnaBelgun and @nf-s this looks really great. This is what we exactly looking for. Thank you so much.

sheelanvaez:
Capture10

philipgrimmett:
@nf-s If possible we should give users the option to dismiss all future download warnings. This should reduce the number of popups for users a strong internet connection and unrestrictive data plans.

Figma prototype: https://www.figma.com/proto/jtdgEiY0gxtZlGtxSJTzcZ/Terria---2021?page-id=8998%3A28482&node-id=9016%3A28437&viewport=241%2C48%2C0.39&scaling=min-zoom&starting-point-node-id=8999%3A27508

Also: I'm not sure there's much value for the user in displaying a 'cancelled download' message in the catalogue. If the concern that users will forget which datasets they cancelled earlier in the session?

PR #6045, opened 2021-12-14
Fix legends in TableMixin

KeyboardSounds:
### What this PR does

Fixes an issue where custom legends defined in the catalog were always ignored by TableMixin.

This bug resulted from the `legends` trait being overridden in TableMixin in such a way that it ignored any other legend definitions that might exist in higher strata.

### Checklist

-   [X] I've updated relevant documentation in `doc/`.


nf-s (changes requested):
This means that we now have duplicated `LegendTraits` for `TableMixin`. 
If you want to create a legend for all `TableStyles` you can set it in `defaultStyle.color.legend`

PR #6046, opened 2021-12-14
Remove legend-related traits from CatalogmMemberTraits

KeyboardSounds:
### What this PR does

Removes legend-related traits from CatalogMemberTraits.

This fixes an issue where TableMixin ends up with the legends trait, which would imply that its legends can be set in a catalog by specifying `legends`, which doesn't work, because TableMixin's legends are controlled by the TableStyle model rather than the legends trait.

### Checklist

    - [x] Add test to check that `LegendOwnerTraits` is present if `!TableMixin.isMixedInto`
    - [x] Add test to check that Models don't specify a `legends` property except through LegendOwnerTraits, unless they're TableMixin-based.


nf-s (approved):
I may have snuck something in 

https://github.com/TerriaJS/terriajs/pull/6046/commits/05eceffbd52d9dc1952e801e4840552b69f86ea1

PR #6054, opened 2021-12-21
Fix `GeoJsonCatalogItem.legends`

nf-s:
### Fix `GeoJsonCatalogItem.legends`

https://github.com/TerriaJS/terriajs/pull/6046 removed `LegendTraits` from `GeoJsonMixin`, but we need them

### Test catalog

https://gist.githubusercontent.com/nf-s/7babad578ac9b9d7412aaa6dc2f9602b/raw/56f95f04be02e84f59efefc7c7aa0a7207c116f6/geojson-legend-fix.json

```json
{
  "catalog": [
    {
      "type": "geojson",
      "name": "Distribution powerlines",
      "id": "649c5937",
      "url": "https://uts-nom.herokuapp.com/capacity/published/combined-distribution-powerlines.geojson",
      "legends": [
        {
          "url": "https://network-opportunity-maps.s3-ap-southeast-2.amazonaws.com/capacity/legends/nom_legend_lines.png"
        }
      ],
      "forceCesiumPrimitives": true
    }
  ],
  "workbench": ["649c5937"]
}
```

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,7 @@ Change Log
 * Fix picked feature highlighting for ArcGis REST API features (and TSify `featureDataToGeoJson`)
 * Re-enable GeoJSON simple styling - now if more than 50% of features have [simple-style-spec properties](https://github.com/mapbox/simplestyle-spec) - automatic styling will be disabled (this behaviour can be disabled by setting `forceCesiumPrimitives = false`)
 * Don't show `TableMixin` `legends` or `mapItems` if no data
+* Fix `GeoJsonCatalogItem.legends`
 * [The next improvement]
 
 #### 8.1.14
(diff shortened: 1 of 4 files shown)


KeyboardSounds (approved):
For the record, I hate that we have to do this. Not that there's anything wrong with this PR, what you've done here is the best solution right now, but maybe someday we can kill TableMixin's special sauce legends with fire and this will no longer be necessary (probably wishful thinking, but hey).
Part of the reason we removed LegendOwnerTraits from CatalogMemberMixin and therefore TableMixin was to avoid catalog items having two different legend specification systems, which makes everything a bit more complicated and harder to use and maintain. But I think it probably makes sense for GeoJsonMixin, and this is the only way to do it.

PR #6059, opened 2021-12-23
Various fixes for SOS.

### LINKED ISSUE #191: NEII - Data missing frequency level dropdown
To test: 
- In v7, the Storage Level data had 6 frequency levels to select: http://neii.gov.au/viewer/#share=s-vsJ2KYFqOrvvI5CiFUoE6OPk3ZR
 
image

- in v8, same data had lost the drop down completely https://beta.neiiviewer.nicta.com.au/#share=s-ht27SbZh4gwCDgTbICpUN4oMwy7

image

na9da:
### What this PR does

- Re-enables procedure and observable selectors - Fixes https://github.com/TerriaJS/neii-viewer/issues/191
- Fixes feature info template rendering that was broken because of using `title` instead of `name` as property names.
  Previously:
    
image

- Temporary workaround for an app crash when rendering feature info chart. The actual cause is explained here https://github.com/TerriaJS/terriajs/issues/6058
- Fixes "ideal zoom" for table mixin items.
- Removes clamping of min y-axis value to 0 in bottom panel chart. This resulted in the feature info panel chart looking very different from the bottom panel chart for scales >> 0. Not sure of the original reasoning but I think it is a harmless change.

Test link:
http://ci.terria.io/fix-sos/#clean&https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/neii/water/prod.json

### Checklist

### DIFF
--- a/lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem.ts
+++ b/lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem.ts
@@ -105,6 +105,9 @@ class SosAutomaticStylesStratum extends TableAutomaticStylesStratum {
         pointSize: createStratumInstance(TablePointSizeStyleTraits, {
           pointSizeColumn: p.identifier
         }),
+        // table style is hidden by default when the table uses only 1 color (https://github.com/TerriaJS/terriajs/blob/bbe8a11ae9bf6c0eb78c52d7b5c9b260d5ddc8cf/lib/Table/TableStyle.ts#L82)
+        // force hidden to false so that the frequency and procedure selector will always be shown
+        // Ideally we should rewrite frequency & procedure selector using selectable dimensions and stop using styles to display them.
         hidden: false
       });
     });
--- a/lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem.ts
+++ b/lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem.ts
@@ -104,7 +104,8 @@ class SosAutomaticStylesStratum extends TableAutomaticStylesStratum {
         }),
         pointSize: createStratumInstance(TablePointSizeStyleTraits, {
           pointSizeColumn: p.identifier
-        })
+        }),
+        hidden: false
       });
     });
   }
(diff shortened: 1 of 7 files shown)


lib/ReactViews/Custom/Chart/BottomDockChart.jsx
```diff
@@ -515,7 +515,7 @@ function calculateDomain(chartItems) {
   const ymax = Math.max(...chartItems.map(c => c.domain.y[1]));
   return {
     x: [xmin, xmax],
-    y: [Math.min(0, ymin), ymax]
+    y: [ymin, ymax]
```
KeyboardSounds, line 518: I have a suspicion that this is going to break some dataset, somewhere, that was depending on us clamping the min of the y domain to zero. But this is probably one of those situations where we tell people to fix their own data.

lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem.ts
```diff
@@ -104,7 +104,8 @@ class SosAutomaticStylesStratum extends TableAutomaticStylesStratum {
         }),
         pointSize: createStratumInstance(TablePointSizeStyleTraits, {
           pointSizeColumn: p.identifier
-        })
+        }),
+        hidden: false
```
KeyboardSounds, line 111: Why do we need this here? Could you add a comment?

KeyboardSounds (changes requested):

lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem.ts
na9da: SOS was implemented long before we got selectable dimensions, so the frequency selector is implemented a "style" when it should actually just be a plain selectable dimension. Style selector is [hidden by default](https://github.com/TerriaJS/terriajs/blob/main/lib/Table/TableStyle.ts#L82) when the table uses only 1 color. We set `hidden: false` to override that behavior and show the selector.

Ideal fix is to rewrite the selectors using simple selectable dimensions. But I think in the interest of time that should be a new PR.

I'll comment the code.

lib/ReactViews/Custom/Chart/BottomDockChart.jsx
na9da: Yeah, i think so too, but this shouldn't be the default behavior. If we find that we need it somewhere we can add a flag to force it per item. The problem with clamping automatically is that the when displaying a single chart that has y value >>> 0 the chart line looks flat loosing all the bumps and details which is bad.

KeyboardSounds (approved):

na9da:
Thanks for the merge @KeyboardSounds!

PR #6060, opened 2021-12-24
Cast property to number in show expression generated from filters for 3d tiles.

na9da:
### What this PR does

Cesium throws an error when the property used with a `filter` is a string instead of a number which is the case for some datasets. This PR adds an explicit cast from string to number to fix the problems.

### Checklist


KeyboardSounds (approved):

na9da:
Thanks @KeyboardSounds!

PR #6093, opened 2022-01-17
Issue 6088

### LINKED ISSUE #6088: MagdaReference should accept request of overriding magda record aspects
In some magda applications such as https://data.nsw.gov.au and https://data.gov.au, the "Open in DigitalTwin" or "Open in NationalMap" action will request the targeted terria (DigitalTwin or NationalMap) to render a dataset distribution referenced by a magda record. If the record's "dataset-format" aspect specifies the "WMS" format, the terria will try to render all layers at the WMS endpoint, which may cause some problems:
1. The WMS service endpoint may represent a complicated hierarchical structure, some layers may contain errors, which will cause the terria to completely abandon the rendering entirely.
2. The number of layers may be too large to be rendered at the same time.

In any cases, users will see an error message that does not tell the root causes. See https://github.com/TerriaJS/nsw-digital-twin/issues/519 for example.

A possible solution to the above problems follows.
When a record has "wms" format, treat it as "wms-group". This approach has been successfully used in DigitalTwin managed catalogue items because its magda records will have customised aspect called "terria" that tells MagdaReference the record is of "wms-group" type (format). However, magda records living outside of DigitalTwin do not have such aspect. If a record has no "terria" aspect, MagdaReference will get type info from the record's "dataset-format" aspect. We need a mechanism to override that "dataset-format" aspect. 

Luckily the external magda application can include "override" attribute in the message to DigitalTwin or NationalMap. This approach does not change external magda application's internal format logic. It only needs to make minimal changes in the post message in `magda-web-client/src/Components/Common/DataPreviewMapOpenInNationalMapButton.tsx` in magda repo. The code looks like this
```
            const id = "data.gov.au-postMessage-" + distribution?.identifier;
            const overriddenFormat = (distribution?.format.toLowerCase() === "wms") ? "wms-group" : undefined;
            const override = overriddenFormat
                ? {
                    "aspects": {
                        "dataset-format": {
                            "format": "wms-group"
                        }
                    }
                }
                : undefined;
            
            catConfig = {
                version: "8.0.0",
                initSources: [
                    {
                        stratum: "user",
                        catalog: [
                            {
                                name: distribution?.title,
                                type: "magda",
                                recordId: distribution?.identifier,
                                url: config.baseExternalUrl,
                                override: override,
                                id
                            }
                        ],
                        workbench: [id],
                        previewedItemId: override? id : undefined
                    }
                ]
            };
```

where `previewedItemId` is to tell the target to open the catalogue panel so the user can see the catalogue group.

Of cause MagdaReference needs to be modified to meet the override request.

mwu2018:
### What this PR does

Fixes #6088 

Catalogue items In DigitalTwin are stored as magda records that typically have "terria" aspects. However, magda records outside of the DigitalTwin catalogue system usually do not have "terria" aspects. This PR allows a "magda" type of json object to provide "terria" aspect via its "addOrOverrideAspects" trait.
  
### Test me
  
See demo in https://github.com/magda-io/magda/pull/3299

### Checklist


lib/ReactViewModels/ViewState.ts
```diff
@@ -567,32 +574,21 @@ export default class ViewState {
...
+          (await theItem.loadMembers()).throwIfError();
         }
-      } else if (MappableMixin.isMixedInto(item))
-        (await item.loadMapItems()).throwIfError();
-      else if (CatalogMemberMixin.isMixedInto(item))
-        (await item.loadMetadata()).throwIfError();
+      } else if (MappableMixin.isMixedInto(theItem))
+        (await theItem.loadMapItems()).throwIfError();
+      else if (CatalogMemberMixin.isMixedInto(theItem))
+        (await theItem.loadMetadata()).throwIfError();
```
nf-s, line 599: I'm confused about the change to `viewCatalogMember()`

I also think we nee to keep the recursion, as it is needed to handle nested references

I have added a test for nested references https://github.com/TerriaJS/terriajs/pull/6093/commits/d41943cd9387f8f95b478db7d2f25261be8c9dec (the test fails with this change, but works in `main`)

lib/Models/Catalog/CatalogReferences/MagdaReference.ts
```diff
@@ -213,7 +221,21 @@ export default class MagdaReference extends AccessControlMixin(
...
+  ) {
+    if (record && override && isJsonObject(override.aspects)) {
+      if (isJsonObject(record.aspects)) {
+        for (let key in override.aspects)
+          record.aspects[key] = override.aspects[key];
+      } else {
+        record.aspects = override.aspects;
+      }
+    }
+  }
```
nf-s, line 239: I think this approach is going against our Traits/Strata system. Also the `override` trait (https://github.com/TerriaJS/terriajs//blob/4e29fbb3519058ed4be8319359b463a6ec810caa/lib/Traits/TraitsClasses/MagdaReferenceTraits.ts#L30-L36) is being used for two different things:

- replacing record.aspects in `overrideRecordAspects` - here it is being treated as Magda JSON (with `aspects` property)
- added to `override` stratum (in [`createGroupFromRecord`](https://github.com/TerriaJS/terriajs//blob/347e4e192abc5700bdc89c037e5d0f3b705a854f/lib/Models/Catalog/CatalogReferences/MagdaReference.ts#L507), [`createMemberFromTerriaAspect`](https://github.com/TerriaJS/terriajs//blob/347e4e192abc5700bdc89c037e5d0f3b705a854f/lib/Models/Catalog/CatalogReferences/MagdaReference.ts#L568) and [`createMemberFromDistributionFormat`](https://github.com/TerriaJS/terriajs//blob/347e4e192abc5700bdc89c037e5d0f3b705a854f/lib/Models/Catalog/CatalogReferences/MagdaReference.ts#L676)) - here it is being treated as Terria JSON (with traits - similar to `itemProperties`)

I think that the `override` trait should just be used for the latter (treated as TerriaJSON).

Which values in aspects do you need to override, is it just the `terria.type` property? 
If so, I'd prefer adding another trait (called `overrideType` or something)

For example (from your test)

Instead of this

```ts
// The aspects will be added to the record and used.
const theOverriddenAspects = {
  aspects: {
    terria: {
      type: theType,
      definition: {
        name: theCatalogItemName,
        url: theDataUrl
      },
      id: theMagdaItemId
    }
  }
};

// Simulate a catalog item of magda type.
const theCatalogItem = {
  id: theMagdaItemId,
  name: theCatalogItemName,
  recordId: theRecordId,
  url: "https://a.magda.portal", // ok not being used in the test
  override: theOverriddenAspects,
  type: "magda" // ok not being used in the test
};
```

You could do this

```ts
const theCatalogItem = {
  id: theMagdaItemId,
  name: theCatalogItemName,
  recordId: theRecordId,
  url: "https://a.magda.portal", // ok not being used in the test
  override: {
    name: theCatalogItemName,
    url: theDataUrl
  },
  overrideType: theType,
  type: "magda" // ok not being used in the test
};
```

lib/Models/Definition/updateModelFromJson.ts
```diff
@@ -25,7 +25,8 @@ export default function updateModelFromJson(
         propertyName === "id" ||
         propertyName === "type" ||
         propertyName === "localId" ||
-        propertyName === "shareKeys"
+        propertyName === "shareKeys" ||
+        propertyName === "aspects"
```
nf-s: I don't `updateModelFromJson` should have `Magda` specific things like this - it should only handle Terria model JSON.

Removing this would follow from my previous comment suggestion

nf-s (changes requested):

lib/Models/Definition/updateModelFromJson.ts
mwu2018: Removed.

lib/Models/Catalog/CatalogReferences/MagdaReference.ts
mwu2018: Changed to use `addOrOverrideAspects` trait to pass on `terria` aspect.

lib/ReactViewModels/ViewState.ts
mwu2018: Bug fixed.

CHANGES.md
```diff
@@ -2,7 +2,7 @@ Change Log
 ==========
 
 #### next release (8.1.18)
-
+* MagdaReference can use override attribute to add or override "terria" aspect of target.
```
nf-s: Can you please change this to mention `addOrOverrideAspects`?

lib/Models/Catalog/CatalogReferences/MagdaReference.ts
```diff
@@ -179,7 +187,8 @@ export default class MagdaReference extends AccessControlMixin(
         this.uniqueId,
         existingRecord,
         override,
-        previousTarget
+        previousTarget,
+        addOrOverrideAspects
```
nf-s, line 191: At some point I think we should refactor all these arguments, there are too many of them and they are quite hard to follow

nf-s: But I think that can be left for another day

nf-s (approved):

PR #6099, opened 2022-01-20
Wc print2

wibily:
## Features:
- React portal into new window allows usage of styled components
- Scale is scaled from parent map
- Will correctly show which SelectableDimension is selected and displayed on the map
- Collapsable/Expandable datasets

## To test:
Go to http://ci.terria.io/wc-print2/
Add in map data from explore map data button
Go to Share/Print button in the top right
Click on Show print preview

### Checklist

-   [ x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [ x] I've updated relevant documentation in `doc/`. N/A
-   [x ] I've provided instructions in the PR description on how to test this PR.


CHANGES.md
```diff
@@ -174,6 +175,7 @@ Change Log
 * Fix `MapNavigation` getter/setter `visible` bug.
   * Replace `CompositeBarItemController` `visible` setter with `setVisible` function
 * Use `yarn` in CI scripts (and upgrade node to v14)
+* TSXified PrintView
```
nf-s: This should be under `next release (8.1.18)`

lib/Models/SelectableDimensions.ts
```diff
@@ -31,22 +33,19 @@ interface Base extends Dimension {
...
 
+export type option<A> = {
+  readonly id?: A;
+  readonly name?: string;
+};
+
 // The default type for SelectableDimension is Select (dropdown menu)
 export interface SelectableDimensionSelect extends Base {
   type?: undefined | "select";
+  readonly options?: readonly option<any>[];
```
nf-s: I think this can be `string`. If not, maybe create generic type in `SelectableDimensionSelect`

```suggestion
  readonly options?: readonly option<string>[];
```

lib/Models/SelectableDimensions.ts
```diff
@@ -66,4 +65,47 @@ namespace SelectableDimensions {
...
+  dim.placement ? dim.placement === placement : placement === DEFAULT_PLACEMENT;
+const isEnabled = (dim: SelectableDimension) => !dim.disable;
+const hasValidOptions = (dim: SelectableDimension) => {
+  const minLength = dim.allowUndefined ? 1 : 2; // Filter out dimensions with only 1 option (unless they have 1 option and allow undefined - which is 2 total options)
+  return (
+    isDefined(dim.options) &&
+    dim.options.length >= minLength &&
+    dim.options.length < MAX_SELECTABLE_DIMENSION_OPTIONS
+  );
+};
```
nf-s, line 84: I like these 👍

lib/ReactViews/Map/Panels/SharePanel/Print/PrintDatasets.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import CatalogMemeberMixin from "../../../../../ModelMixins/CatalogMemberMixin";
+import { BaseModel } from "../../../../../Models/Definition/Model";
+import Description from "../../../../Preview/Description";
+
+interface Props {
+  items: readonly (CatalogMemeberMixin.Instance & BaseModel & any)[];
```
nf-s:
```suggestion
  items: readonly BaseModel[];
```

lib/ReactViews/Map/Panels/SharePanel/Print/PrintDatasets.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+const PrintDatasets = (props: Props) => {
+  return (
+    <>
+      {props.items.map((item, index) => (
+        <details key={index} open>
+          <summary>{item.name}</summary>
```
nf-s: There is a `getName` function which takes `BaseModel` and tries to find best name property.

```suggestion
          <summary>{getName(item)}</summary>
```

lib/ReactViews/Map/Panels/SharePanel/Print/PrintDatasets.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import CatalogMemeberMixin from "../../../../../ModelMixins/CatalogMemberMixin";
```
nf-s:
```suggestion
import { getName } from "../../../../../ModelMixins/CatalogMemberMixin";
```

nf-s: Needed for change below

lib/ReactViews/Map/Panels/SharePanel/Print/PrintView.tsx
```diff
@@ -0,0 +1,179 @@
...
+                  scale={getScale(props.terria.currentViewer.getContainer())}
+                />
+              </PrintViewMap>
+            ) : (
+              <div>loading</div>
+            )}
+          </div>
+        </section>
+        <section className="PrintView__source">
+          <PrintSource link={buildShareLink(props.terria, props.viewState)} />
```
nf-s: Here we should use `buildShortShareLink` if possible.

For example see

https://github.com/TerriaJS/terriajs/blob/7e0f5f594a5c5bcf57f1490792e22c1b20bda221/lib/Models/sendFeedback.ts#L32-L36

package.json
```diff
@@ -79,7 +79,6 @@
     "classnames": "^2.3.1",
     "clipboard": "^2.0.8",
     "core-js": "^3.1.4",
-    "create-react-class": "^15.6.2",
```
nf-s, line 82: 🎉

nf-s: Bonus points for removing dep

lib/ReactViews/Map/Panels/SharePanel/Print/PrintWorkbench.tsx
```diff
@@ -0,0 +1,76 @@
...
+  filterSelectableDimensions,
+  findSelectedValueName
+} from "../../../../../Models/SelectableDimensions";
+import { useTranslation } from "react-i18next";
+
+interface Props {
+  workbench: Workbench;
+}
+
+const renderDisplayVariables = (catalogItem: BaseModel & any) => {
```
nf-s: Can we get rid of the `any` here?

```suggestion
const renderDisplayVariables = (catalogItem: BaseModel) => {
```

lib/ReactViews/Map/Panels/SharePanel/Print/PrintWorkbench.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <div className="layer-title">{catalogItem.name}</div>
+      {catalogItem.currentTime && (
+        <div className="layer-time">Time: {catalogItem.currentTime}</div>
+      )}
+      <Legend forPrint={true} item={catalogItem} />
+    </div>
+  );
+};
+
+const WorkbenchItem = ({ item }: { item: BaseModel | any }) => {
```
nf-s:
```suggestion
const WorkbenchItem = ({ item }: { item: BaseModel }) => {
```

lib/ReactViews/Map/Panels/SharePanel/Print/PrintWorkbench.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+  return null;
+};
+
+const renderLegend = (
+  catalogItem: MappableMixin.Instance &
+    DiscretelyTimeVaryingMixin.Instance &
+    CatalogMemeberMixin.Instance &
+    BaseModel &
+    any
```
nf-s:
```suggestion
  catalogItem: BaseModel
```

lib/ReactViews/Map/Panels/SharePanel/Print/PrintWorkbench.tsx
```diff
@@ -0,0 +1,76 @@
...
+) => {
+  if (!catalogItem.isMappable) {
```
nf-s:
```suggestion
  if (!MappableMixin.isMixedInto(catalogItem)) {
```

lib/ReactViews/Map/Panels/SharePanel/Print/PrintWorkbench.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return null;
+  }
+
+  return (
+    <div key={catalogItem.uniqueId} className="layer-legends">
+      <div className="layer-title">{catalogItem.name}</div>
+      {catalogItem.currentTime && (
```
nf-s:
```suggestion
      {DiscretelyTimeVaryingMixin.isMixedInto(catalogItem) && (
```

nf-s (changes requested):
Looking good!

Just a few type-related suggestions.

Also, the map screenshot doesn't update if there is an existing window.

To recreate:
- Click Share/Print
- Leave tab open
- Click Share/Print again

lib/ReactViews/Map/Panels/SharePanel/Print/PrintDatasets.tsx
wibily: Thanks, i didn't know about that function. It makes tightening up the types a lot easier

lib/ReactViews/Map/Panels/SharePanel/Print/PrintWorkbench.tsx
wibily: yup, will do

nf-s (approved):
Looks good!

A few small things - I realise this is not the final version - so feel free to respond to them in another PR:

1. Time format is not great
2. Dealing with scale in 3D mode is hard, not sure how best to do it. In this example the camera angle is quite low, so scale will change a lot from bottom of the map to the top. Maybe it is better to not show a scale if there is large variation across the map?
3. There are some `Collapsible` caret icon things missing here
4. The share link doesn't update when you re-generate the Print view if there is an existing window

Screen Shot 2022-01-25 at 1 38 23 pm

Other things 

- general typography things (multiple fonts, spacing...) - not my area of expertise

wibily:
thanks for the detailed review:
1) Yeah, agreed. I'll make a note of it in the main ticket
2) I'm using the same component for the scale as the one in the main viewer so i think we'll need to figure out that problem holistically. I'm not sure what the best solution is either :/
3) Yup, the whole expandable items section needs a bit more work. Right now i'm just using the default HTML <summary> and <detail> elements
4. Nice catch, i missed that
5. Yeah, I've kinda fudged the CSS a bit by importing a style directly into the page. Since the parent element is a ThemeProvider, I really should be using the standard <Text>/<Box> components

PR #6101, opened 2022-01-20
Fix bad ESRI features in featureDataToGeoJson

nf-s:
#### Before

- http://ci.terria.io/main/#configUrl=https://terria-catalogs-public.storage.googleapis.com/pacificmap/map-config/dev.json&share=s-kfCRm4NveldHaWZ2v5UMR7hdZAx

#### After

- http://ci.terria.io/get-esri-feature-fix/#configUrl=https://terria-catalogs-public.storage.googleapis.com/pacificmap/map-config/dev.json&share=s-kfCRm4NveldHaWZ2v5UMR7hdZAx

### Checklist


wibily:
ehhh it seems fine although our checks for "does this json blob expect our shape of the object" should really be holistically looked at. I would suggest io-ts :)

wibily (approved):

PR #6106, opened 2022-01-24
Only show `TableMixin.legends` if there is actually data on the map

nf-s:
### Test me

After vs Before

image

#### Before

- Load GeoJSON dataset which doesn't use Table style:
- http://ci.terria.io/main/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/de-australia/map-config/dev.json&share=s-gITOdwiNoc56yRynRhPsXoF47Mr
- Note mis-matched legends

#### After

- Repeat with http://ci.terria.io/tablemixin-legend-rows/#configUrl=https://raw.githubusercontent.com/TerriaJS/saas-catalogs-public/main/de-australia/map-config/dev.json&share=s-gITOdwiNoc56yRynRhPsXoF47Mr
- Note no legends - which is correct

### Checklist

### DIFF
--- a/test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
+++ b/test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
@@ -25,873 +25,908 @@ import CommonStrata from "../../../../lib/Models/Definition/CommonStrata";
 import updateModelFromJson from "../../../../lib/Models/Definition/updateModelFromJson";
 import Terria from "../../../../lib/Models/Terria";
 
-describe("GeoJsonCatalogItem - with cesium primitives", function() {
-  let terria: Terria;
-  let geojson: GeoJsonCatalogItem;
+describe("GeoJsonCatalogItemSpec", () => {
+  describe("- with cesium primitives", function() {
+    let terria: Terria;
+    let geojson: GeoJsonCatalogItem;
 
-  beforeEach(function() {
-    terria = new Terria({
-      baseUrl: "./"
+    beforeEach(function() {
+      terria = new Terria({
+        baseUrl: "./"
+      });
+      geojson = new GeoJsonCatalogItem("test-geojson", terria);
+      geojson.setTrait(CommonStrata.user, "forceCesiumPrimitives", true);
     });
-    geojson = new GeoJsonCatalogItem("test-geojson", terria);
-    geojson.setTrait(CommonStrata.user, "forceCesiumPrimitives", true);
+
+    describe("GeoJsonCatalogItem", function() {
+      it("reloads when the URL is changed", async function() {
+        geojson.setTrait(
+          CommonStrata.user,
+          "url",
+          "data:application/json;base64,eyJ0eXBlIjoiRmVhdHVyZUNvbGxlY3Rpb24iLCJmZWF0dXJlcyI6W3sidHlwZSI6IkZlYXR1cmUiLCJwcm9wZXJ0aWVzIjp7fSwiZ2VvbWV0cnkiOnsidHlwZSI6IlBvaW50IiwiY29vcmRpbmF0ZXMiOlsxNDgsLTMxLjNdfX1dfQ=="
+        );
+        await geojson.loadMapItems();
+        expect(geojson.mapItems.length).toEqual(1);
+        expect(
+          (geojson.mapItems[0] as GeoJsonDataSource).entities.values.length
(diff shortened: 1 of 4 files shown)


test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
```diff
@@ -25,873 +25,908 @@ import CommonStrata from "../../../../lib/Models/Definition/CommonStrata";
...
 import Terria from "../../../../lib/Models/Terria";
 
-describe("GeoJsonCatalogItem - with cesium primitives", function() {
-  let terria: Terria;
-  let geojson: GeoJsonCatalogItem;
-
-  beforeEach(function() {
-    terria = new Terria({
-      baseUrl: "./"
+describe("GeoJsonCatalogItemSpec", () => {
```
nf-s, line 28: Note, the reason for large diff, is that I have wrapped all specs in 

```
describe("GeoJsonCatalogItemSpec", () => {
```

test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
```diff
@@ -25,873 +25,908 @@ import CommonStrata from "../../../../lib/Models/Definition/CommonStrata";
...
-    );
-    await geojson.loadMapItems();
-    expect(geojson.mapItems[0] instanceof GeoJsonDataSource).toBeTruthy();
-    expect(geojson.useMvt).toBeFalsy();
+      expect(geojson.legends.length).toBe(0);
```
nf-s, line 863: This line is a new change to the `"Disabled protomaps - More than 50% features detected"` test:

```ts
expect(geojson.legends.length).toBe(0);
```

test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
```diff
@@ -25,873 +25,908 @@ import CommonStrata from "../../../../lib/Models/Definition/CommonStrata";
...
+      );
+      await geojson.loadMapItems();
+      expect(geojson.mapItems[0] instanceof GeoJsonDataSource).toBeFalsy();
+      expect(geojson.useMvt).toBeTruthy();
+      expect(geojson.legends.length).toBe(1);
```
nf-s, line 849: This line is a new change to the `"Unchanged - less than 50% features detected"` test:

```ts
expect(geojson.legends.length).toBe(1);
```

mwu2018:
I have a naïve question about the styling: Is it possible that there are different colours polygons so that showing the colour legends is helpful?

nf-s:
Sorry @mwu2018 I should have clarified the edge case.

`GeoJsonMixin` has 3 render modes:
- CZML:
  - if `czmlTemplate` is defined (see `GeoJsonTraits.czmlTemplate`)
- Mapbox vector tiles (through geojson-vt and protomaps.js)
  - Will be used by default, if not using unsupported traits (see below)
- Cesium primitives if:
  - `GeoJsonTraits.forceCesiumPrimitives = true`
  - Using `timeProperty` or `heightProperty` or `perPropertyStyles` or simple-style `marker-symbol`
  - More than 50% of GeoJSON features have simply-style properties

The `TableMixin.legends` are only relevant to the "Mapbox vector tiles" rendering mode.

The example dataset has more than 50% of GeoJSON features have simply-style properties - so Cesium primitives are used.

The `GeoJsonMixin` - `TableMixin` interaction is a bit weird, but the whole point of it is that we use the `TableStyle` system for Mapbox vector tiles.

So if we aren't using table stying, then we set `TableMixin.dataColumnMajor` to be empty [here](https://github.com/TerriaJS/terriajs//blob/93a220f573a73818152a7aec6e02802d3ab087fa/lib/ModelMixins/GeojsonMixin.ts#L980-L981)

```ts
get dataColumnMajor() {
      if (!this.readyData || !this.useMvt || this.disableTableStyle) return [];
```

Which now gets picked up in `TableMixin.legends`

```ts
    get legends(): Model<LegendTraits>[] {
      // Only return legends if we have rows in dataColumnMajor and mapItems to show
      if (this.dataColumnMajor?.length !== 0 && this.mapItems.length > 0) {
```

Whereas before, `TableMixin.legends` was returning legends even if table styling wasn't used

mwu2018 (approved):

mwu2018:
Good clarification!

PR #6108, opened 2022-01-25
Add `WebMapServiceCatalogGroup.perLayerLinkedWcs`

nf-s:
### Add `WebMapServiceCatalogGroup.perLayerLinkedWcs`

Add `WebMapServiceCatalogGroup.perLayerLinkedWcs`, this can be used to enable `ExportWebCoverageService` for **all** WMS layers. `item.linkedWcsCoverage` will be set to the WMS layer `Name`

#### Example config

```json
{
  "workbench": [],
  "catalog": [
    {
      "id": "deafrica_terria_prod",
      "type": "wms-group",
      "name": "Digital Earth Africa data catalogue",
      "url": "https://ows.digitalearth.africa/",
      "itemProperties": {
        "opacity": 1.0
      },
      "excludeMembers": [],
      "isOpen": true,
      "perLayerLinkedWcs": {
        "linkedWcsUrl": "https://ows.digitalearth.africa/"
      }
    }
  ],
  "corsDomains": [
    "digitalearthafrica.org",
    "digitalearth.africa"
  ]
}

```

#### Before

- http://ci.terria.io/main/#clean&https://gist.githubusercontent.com/nf-s/7bb9d210cff55cdb6b5cffde12ff4e96/raw/38b6c37b8d6e4436bfa3c1025a0f1f47cbd5f339/de-africa-wms-wcs.json
- Ignore error about `perLayerLinkedWcs`
- Note lack of `Export` option in workbench

#### After

- http://ci.terria.io/wms-group-wcs/#clean&https://gist.githubusercontent.com/nf-s/7bb9d210cff55cdb6b5cffde12ff4e96/raw/38b6c37b8d6e4436bfa3c1025a0f1f47cbd5f339/de-africa-wms-wcs.json
- All items will have `Export` functionality

### Checklist


lib/Models/Catalog/Ows/WebMapServiceCatalogGroup.ts
```diff
@@ -276,6 +277,21 @@ class GetCapabilitiesStratum extends LoadableStratum(
...
+    // Copy over ExportWebCoverageTraits if `linkedWcsUrl` has been set
+    // See WebMapServiceCatalogGroupTraits.perLayerLinkedWcs for more info
+    if (this.catalogGroup.perLayerLinkedWcs?.linkedWcsUrl) {
+      updateModelFromJson(model, stratum, {
+        // Copy over all perLayerLinkedWcs objects
+        ...this.catalogGroup.traits.perLayerLinkedWcs.toJson(
+          this.catalogGroup.perLayerLinkedWcs
+        ),
+        // Override linkedWcsCoverage with layer.Name (or Title if Name is undefined)
+        linkedWcsCoverage: layer.Name ?? layer.Title
```
wibily: wait why is it sometimes layer.Name and othertimes layer.name for different catalog types?
Also, it may be worth creating a helper function `getName: (layer: CapabilitiesLayer) => string` that does the resolution as there are other calls to just layer.Name and i'm not sure if they need to fall back to layer.Title. E.g lib/Models/Catalog/Ows/WebMapServiceCatalogGroup.ts:260

wibily (approved):

lib/Models/Catalog/Ows/WebMapServiceCatalogGroup.ts
nf-s: Sorry I should have just used `layer.Name` here. I have clarified the difference between the two [here](https://github.com/TerriaJS/terriajs//blob/4453b2a553699853614c156f9137e7038f6bf083/lib/Models/Catalog/Ows/WebMapServiceCapabilities.ts#L55-L69)

```ts
export interface CapabilitiesLayer {
  readonly _parent?: CapabilitiesLayer;
  /** ### Adapted from WMS 1.3.0 spec:
   *
   * #### 7.2.4.2
   *
   * A number of elements have both a `<Name>` and a `<Title>`.
   * The `Name` is a text string used for machine-to-machine communication while the `Title` is for the benefit of humans.
   *
   * #### 7.2.4.6.3
   *
   * If, and only if, a layer has a `<Name>`, then it is a map layer that can be requested by using that `Name` in the `LAYERS` parameter of a GetMap request.
   * If the layer has a `Title` but no `Name`, then that layer is only a category title for all the layers nested within.
   * A client shall not attempt to request a layer that has a `Title` but no `Name`.
   */
  readonly Name?: string;
  readonly Title: string;
```

PR #6147, opened 2022-02-17
Clipping box

na9da:
### What this PR does

- Implements clipping box.
- Adds `group` and `checkbox-group` dimension selectors.

TODOs:
- [ ] Workbench clipping UI changes requested by client
- [ ] Sync with latest `master`
  
### Test me
  
1. Open this [CI link](http://ci.terria.io/clipping-box/#clean&https://gist.githubusercontent.com/na9da/3031b76246f3d5f6802b3ec1d56afd05/raw/f43383f9bb57f77d15ec082520fa122aa2f02b04/clipping-box-tilesets.json).
2. Add any of the 2 3dtiles datasets.
3. Zoom to the dataset and enable the clipping box for the item from the workbench
4. Drag, scale and rotate the clipping box to select the clipping volume
5. Create a share link and open and verify that the clipping is retained.

### Checklist


lib/Sass/global/_normalize.scss
```diff
@@ -1,152 +1,125 @@
-/*! normalize.css v3.0.3 | MIT License | github.com/necolas/normalize.css */
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
+
+/* Document
+   ========================================================================== */
```
nf-s, line 4: These changes have impacted a few styles:

Screen Shot 2022-02-21 at 2 36 37 pm

lib/ReactViews/Workbench/Controls/DimensionSelectorSection.tsx
```diff
@@ -47,72 +42,219 @@ class DimensionSelectorSection extends React.Component<PropsType> {
...
+        isDefined(dim.options) &&
+        dim.options.length < MAX_SELECTABLE_DIMENSION_OPTIONS &&
+        dim.options.length + (dim.allowUndefined ? 1 : 0) > 1;
+      if (!hasValidOptions) {
+        return false;
+      }
+    }
+    return true;
+  });
+}
```
nf-s: This `filterValidSelectableDimensions` function is now in `SelectableDimensions.ts` (called `filterSelectableDimensions`)

https://github.com/TerriaJS/terriajs//blob/6cdfab184e284cd4d8c3a2837164e54fd93fbada/lib/Models/SelectableDimensions.ts#L119-L128

lib/Models/SelectableDimensions.ts
```diff
@@ -49,9 +54,34 @@ export interface SelectableDimensionCheckbox extends Base {
...
+  extends Omit<Base, "setDimensionValue"> {
+  type: "group";
+
+  // We don't allow nested groups for now to keep the UI simple
+  readonly selectableDimensions: Exclude<
+    SelectableDimension,
+    SelectableDimensionGroup | SelectableDimensionCheckboxGroup
+  >[];
+}
+
```
nf-s, line 79: I find the CheckboxGroup a bit confusing. Depending on where you click, sometimes the checkbox is hit (which changes boolean value), and sometimes the group expands/collapses but the value doesn't change.

I think I would rather a normal Group, and then add another checkbox inside the group to toggle clipping

lib/ModelMixins/Cesium3dTilesMixin.ts
```diff
@@ -267,11 +283,21 @@ function Cesium3dTilesMixin<T extends Constructor<Model<Cesium3dTilesTraits>>>(
...
+
+      return [this.tileset, ...this.clippingMapItems];
     }
 
     @computed get selectableDimensions(): SelectableDimension[] {
-      return [...super.selectableDimensions, this.shadowDimension];
+      return filterOutUndefined([
+        ...super.selectableDimensions,
+        this.shadowDimension,
+        this.clippingDimension
```
nf-s: I think `shadowDimension` and `clippingDimension` should be in `ShadowMixin`/`ClippingMixin`.`selectableDimensions`. Just so it is consistent with other usage of `SelectableDimension` - eg `TableMixin` etc.

lib/ModelMixins/ClippingMixin.ts
```diff
@@ -0,0 +1,355 @@
...
+type InstanceType = BaseType & {
+  clippingPlanesOriginMatrix(): Matrix4;
+  clippingPlaneCollection: ClippingPlaneCollection | undefined;
+  clippingDimension: SelectableDimension | undefined;
+  clippingMapItems: CustomDataSource[];
+};
+
+function ClippingMixin<T extends Constructor<BaseType>>(
+  Base: T
+): T & Constructor<InstanceType> {
```
nf-s: Should `ClippingMixin` just extend `MappableMixin`?

lib/ModelMixins/ClippingMixin.ts
```diff
@@ -0,0 +1,355 @@
...
+            setDimensionValue: (stratumId, value) => {
+              this.clippingBox.setTrait(stratumId, "clipDirection", value);
+            }
+          }
+        ]
+      };
+    }
+  }
+
+  return MixedClass as any;
```
nf-s: Does this need to be `any` here?

nf-s (changes requested):
Awesome feature @na9da ! I love it

A few minor things and one more general issue:

Sometimes I can't click on/interact with the clipping box. It happens when I rotate the camera around the box, and try to click on different sides on the box. I think it happens when the camera is quite low to the ground?

https://user-images.githubusercontent.com/6187649/154887770-fe9a6012-9156-44c8-bf93-2d6af9d3737f.mov

na9da:
> Awesome feature @na9da ! I love it
> 
> A few minor things and one more general issue:
> 
> Sometimes I can't click on/interact with the clipping box. It happens when I rotate the camera around the box, and try to click on different sides on the box. I think it happens when the camera is quite low to the ground?
> Screen.Recording.2022-02-21.at.3.06.09.pm.480.mov

Good catch. I have some code that disables interaction when the box is out of camera view so that performance is not impacted when the box is not in view. There must be something wrong with that code. I'll see if I can fix it or else I'll disable it.

lib/Models/SelectableDimensions.ts
na9da: Good catch!

I'll see if I can fix that behavior.

I had previously used simple groups, but the client specifically requested for a checkbox group. It is also closer to the design @philipgrimmett made for collapsible groups.

na9da: I have fixed the behavior so that clicking anywhere on the length of the checkbox group element should toggle it.

lib/ModelMixins/Cesium3dTilesMixin.ts
na9da: Done.

lib/ModelMixins/ClippingMixin.ts
na9da: The `MappableTraits` was redundant so I removed it.

lib/ModelMixins/ClippingMixin.ts
na9da: I think my editor was briefly using TS 4.x for type checking which throws an error about not being able to convert an `abstract` class to a concrete one. I think we will have to deal with it when we migrate to TS4.x. I have removed the cast.

lib/Sass/global/_normalize.scss
na9da: Good catch!

I decided to upgrade normalize.css as the styling for some of the elements like `details` and `summary` didn't match the standard behavior in HTML. Newer versions don't have this problem.

I have fixed the styling changes you have pointed out. My opinion is we should keep the upgrade and fix any other inconsistencies we come across. Let me know if you think otherwise.

nf-s: Yep totally agree - we are also going to encounter subtle styling changes with the new TSX workbench PR

nf-s (changes requested):
All your changes look good to me!

One other issue - the "Clamp to ground" button doesn't work for me

http://ci.terria.io/clipping-box/#clean&https://gist.githubusercontent.com/na9da/3031b76246f3d5f6802b3ec1d56afd05/raw/f43383f9bb57f77d15ec082520fa122aa2f02b04/clipping-box-tilesets.json&share=s-hanHpGlOtgClqIG1hsCcdh0zy3l

na9da:
> All your changes look good to me!
> 
> One other issue - the "Clamp to ground" button doesn't work for me
> 
> http://ci.terria.io/clipping-box/#clean&https://gist.githubusercontent.com/na9da/3031b76246f3d5f6802b3ec1d56afd05/raw/f43383f9bb57f77d15ec082520fa122aa2f02b04/clipping-box-tilesets.json&share=s-hanHpGlOtgClqIG1hsCcdh0zy3l

Ok, how did you make that happen? Did you drag the top surface to lift the box up, that is allowed by design. "Clamping to ground" only prevents the box from going underground, it doesn't really make the box stick to the ground. So maybe I need to modify the text to make it clear.

https://github.com/TerriaJS/terriajs/blob/41beb04aa41e091cd8ccddea4b1f5fb5a7fa7eee/lib/Models/BoxDrawing.ts#L280-L283

na9da:
@nf-s  Have changed the text from "Clamp box to ground" to "Keep above ground". Hope that makes it clear.

nf-s (approved):
Looks good!

PR #6151, opened 2022-02-18
Add story routes & catalog routes

tephenavies:
### What this PR does

Replaces #4961.

Tasks remaining:
- [x] Parameterise location for fetching JSON for story routes
- [x] Merge TerriaMap changes needed
- [x] Add tests for story route
- [ ] Add tests for catalog route
- [x] Update CHANGES.md
- [x] Update relevant documentation (probably including ADRs)
  
### Test me

**Test catalog routes**. For these tests the routes will bring up different previews and the url will always rewrite to the base url, retaining all query string & hash parameters:
* http://ci.terria.io/story-routes/catalog/DL4Av6wY5h - preview "GeoJSON Test"
* http://ci.terria.io/story-routes/catalog/xrD8v0H4cn - preview "Train Stations"

**Test story routes**. These should load a story and start the story playing, or show an error if the story JSON couldn't be fetched. There should be no url rewrites on success or failure to load a story. Cesium Ion key has been removed so you'll have to switch to 3d smooth or 2d:
* http://localhost/test-slug/story/DEACoastlines#configUrl=https://gist.githubusercontent.com/steve9164/9ccc8e521453457cbf5214eb7a576a07/raw/1d1a60b3652b04e5dd5ef0b03776b9182d9f9ffc/dea-map-config-prod.json (configUrl hash parameter needed for this to load the DEA catalog. The final url would be `https://maps.dea.ga.gov.au/story/DEACoastlines`)

### Checklist


buildprocess/ci-deploy.sh
```diff
@@ -21,7 +21,7 @@ npm install -g yarn@^1.19.0
 
 # Clone and build TerriaMap, using this version of TerriaJS
 TERRIAJS_COMMIT_HASH=$(git rev-parse HEAD)
-git clone -b main https://github.com/TerriaJS/TerriaMap.git
+git clone -b 6151-spa-routing https://github.com/TerriaJS/TerriaMap.git
```
tephenavies: Only for CI. This can be reverted when https://github.com/TerriaJS/TerriaMap/pull/578 is merged.

buildprocess/ci-deploy.sh
```diff
@@ -33,7 +33,7 @@ git tag -a "TerriaMap-$TERRIAMAP_COMMIT_HASH--TerriaJS-$TERRIAJS_COMMIT_HASH" -m
 rm yarn.lock # because TerriaMap's yarn.lock won't reflect terriajs dependencies
 yarn install
 yarn add -W moment@2.24.0
-yarn gulp build
+yarn gulp build --baseHref="/${SAFE_BRANCH_NAME}/"
```
tephenavies, line 36: [Passing `baseHref` as a parameter to TerriaMap's `gulp` build tasks](https://github.com/TerriaJS/TerriaMap/pull/578/files#diff-25789e3ba4c2adf4a68996260eb693a441b4a834c38b76167a120f0b51b969f7) is IMO less confusing than pulling `baseHref` out of `devserverconfig.json` at build time when the rest of the configuration in that file is runtime configuration - which was the solution [the previous time `baseHref` was proposed](https://github.com/TerriaJS/TerriaMap/pull/498/files#diff-25789e3ba4c2adf4a68996260eb693a441b4a834c38b76167a120f0b51b969f7R46-R51).

lib/Models/Terria.ts
```diff
@@ -1006,6 +1013,62 @@ export default class Terria {
...
+          segments[0] === "story" &&
+          segments[1].length > 0 &&
+          isDefined(this.configParameters.storyRouteUrlPrefix)
+        ) {
+          let storyJson;
+          try {
+            storyJson = await loadJson(
+              `${this.configParameters.storyRouteUrlPrefix}${encodeURIComponent(
+                this.appName
+              )}/${segments[1]}`
```
tephenavies: Share JSON is downloaded in the client, rather than using terriajs-server. It must come from a server supporting CORS (even if the domain is in the proxyable domains list). My concerns with allowing proxying is that this is outside the initSources and url traits system, so there's some configuration of the proxying behaviour that can't be done here. Though maybe we should allow it and use default proxy settings.

lib/Models/Terria.ts
```diff
@@ -1006,6 +1013,62 @@ export default class Terria {
...
+          this.initSources.push({
+            name: `Go to ${pageUrl.pathname}`,
+            errorSeverity: TerriaErrorSeverity.Error,
+            data: {
+              previewedItemId: decodeURIComponent(segments[1])
+            }
+          });
+          const replaceUrl = new URL(newUrl);
+          replaceUrl.pathname = new URL(document.baseURI).pathname;
+          history.replaceState({}, "", replaceUrl.href);
```
tephenavies, line 1072: Go immediately back to the baseURI after parsing. Uses `replaceState` (rather than `pushState`) to not create a history entry.

lib/Models/Terria.ts
tephenavies: Also, should we use `appName` as part of the URL? It means that e.g. nationalmap and DEA can share the same stories bucket and have slightly different stories for the same story id (which I think is essential, I don't think we can't get away with globally unique story ids). The alternative is use a slightly different `storyRouteUrlPrefix` for each map, which may be a better idea.

na9da: For whether to use `appName`. I think I prefer the second option of changing the `storyRouteUrlPrefix` per map because it is less confusing and avoids having to explicitly communicate everywhere that the share json URL has to be `storyRouteUrlPrefix + appName`.

na9da: For proxying the story route url, the user could configure it as:

`"storyRouteUrlPrefix": "proxy/https://tile-test.terria.io/stories/"` and add it to the proxy list.

So, shouldn't be a problem, I guess.

lib/Models/Terria.ts
```diff
@@ -1006,6 +1031,63 @@ export default class Terria {
...
+        // then split into url segments
+        // e.g. "http://ci.terria.io/main/story/1#map=2d" -> ["story", "1"]
+        const segments = (pageUrl.origin + pageUrl.pathname)
+          .slice(this.appBaseHref.length)
+          .split("/");
+        if (
+          segments.length === 2 &&
+          segments[0] === "catalog" &&
+          segments[1].length > 0
+        ) {
```
na9da: One small problem is that the test fails if I have a `/` after `catalog` or `story`. Should we be a bit more permissive?

lib/Models/Terria.ts
na9da: One other option is to make `storyRouteUrlPrefix` a template string that'll accept `{appName}`

na9da:
It'll be nice if we extracted the new code in `Terria` to a different `Routing` class to not add to the ever growing `Terria.ts`. But I leave that decision to you (maybe just leave it as a comment to come back in the future).

test/Models/TerriaSpec.ts
```diff
@@ -324,157 +303,196 @@ describe("Terria", function() {
...
+        await terria.updateApplicationUrl(
+          new URL("story/my-story", document.baseURI).toString()
+        );
+        expect(terria.initSources.length).toBe(1);
+        expect(terria.initSources[0].name).toMatch(/my-story/);
+        if (!isInitData(terria.initSources[0]))
+          throw new Error("Expected initSource to be InitData from my-story");
+
+        expect(toJS(terria.initSources[0].data)).toEqual(
+          (await (await fetch("test/stories/TerriaJS%20App/my-story")).json())
```
na9da, line 507: nitpick: one other way to do this might be to `require()` the json file.

na9da (approved):
Works great @steve9164! 

I have just left a few comments.

lib/Models/Terria.ts
tephenavies: Maybe `/catalog/` should be a route that just opens the catalog, but `/story/` should not be a route IMO. It should only support `/story/:story-id`.

test/Models/TerriaSpec.ts
tephenavies: Yeah, I was going to do that, but I couldn't work out how to get webpack to import `my-story` (which lacks the `.json` file extension) as a JSON file

lib/Models/Terria.ts
na9da: Oh sorry, I meant `/` at the end like in `/story/foo/`

lib/Models/Terria.ts
tephenavies: Thanks for your opinion. I agree @na9da. I've changed the behaviour to leave out appName from URL construction. At the moment I won't implement storyRouteUrlPrefix as a template string, but I'll keep it in mind if requirements change.

lib/Models/Terria.ts
tephenavies: Yeah, this is probably a requirement at least for story routes, since I think they're intended to be sometimes typed out by humans, not just linked to from media releases etc.

architecture/0009-story-and-catalog-routes.md
```diff
@@ -0,0 +1,45 @@
+# 9. Story and catalog routes
```
tephenavies, line 1: New ADR for routes

doc/deploying/running-through-reverse-proxy.md
```diff
@@ -0,0 +1,13 @@
+To configure TerriaMap to run correctly through a reverse proxy that prefixes a path to URLs you should pass a `baseHref` parameter to any `gulp` "build" task that you run.
```
tephenavies, line 1: New doc page for gulp task `baseHref` parameter

lib/Language/en/translation.json
```diff
@@ -865,7 +865,7 @@
       "missingUrlMessage": "Could not load the ArcGis FeatureServer endpoint because the catalog item does not have a `url`.",
       "invalidServiceTitle": "Invalid ArcGIS Feature Service layer",
       "invalidServiceMessage": "An error occurred while invoking the ArcGIS Feature Service.  The server's response does not appear to be a valid Feature Service layer.",
-      "reachedMaxFeatureLimit": "Warning: This layer has reached the FeatureService feature limit ({{maxFeatures}})."
+      "reachedMaxFeatureLimit": "Warning: This layer has reached the FeatureService feature limit."
```
tephenavies: Bad merge somewhere
```suggestion
      "reachedMaxFeatureLimit": "Warning: This layer has reached the FeatureService feature limit ({{maxFeatures}})."
```

lib/Models/Terria.ts
```diff
@@ -1009,6 +1034,64 @@ export default class Terria {
...
+      if (newUrl.startsWith(this.appBaseHref)) {
+        function checkSegments(urlSegments: string[], customRoute: string) {
+          // Accept /${customRoute}/:some-id/ or /${customRoute}/:some-id
+          return (
+            ((urlSegments.length === 3 && urlSegments[2] === "") ||
+              urlSegments.length === 2) &&
+            urlSegments[0] === customRoute &&
+            urlSegments[1].length > 0
+          );
+        }
```
tephenavies, line 1054: Now accepts trailing slash

tephenavies:
Thanks for your suggestions @na9da. Do you mind taking another quick look now?

PR #6157, opened 2022-02-22
Customised ideal zoom

### LINKED ISSUE #6158: Better support of customised ideal zoom
It is not easy to come up with correct parameters when using `idealZoom` property in `MappableTraits` to customise ideal zoom. Need a better way.

mwu2018:
### What this PR does

Fixes #6158 

It is much easier to obtain a set of parameters for  `initialCamera` than `idealZoom` property in `MappableTraits` when setting customised ideal zoom.

The required parameters are easily obtained via shared link function.  

### Test me
The Ballarat 3D Photomesh dataset in `dev` has been using this new feature. Please see
https://github.com/TerriaJS/vic-digital-twin-catalog-next/blob/493b02f7b3feb051f6309f26e2a3412260931acf/dev/c884463c-b699-44ac-822c-34299d1f6d96.json#L43 and
https://dev.vic.digitaltwin.terria.io/#share=s-bqEgCgznEfFvHr9RJ0lU4U96N0i

### Checklist


tephenavies (commented):
I think this should instead modify the `idealZoom` trait(s). My thoughts are that idealZoom should act very similarly to `initialCamera` from init files: https://docs.terria.io/guide/customizing/initialization-files/#cameraposition

I.e. it can have:
* `{west,south,east,north}`
* `{west,south,east,north,position,direction,up}`
* `{positionHeading}`; or
* `{lookAt}`

mwu2018:
@steve9164
I've revised the code based on your comment. However, there are some differences, I think.
- Use `initialCamera` trait name. It is the same as the key name from the shared link api. I am open to other options.
- Can specify either `lookAt` or `initialCamera` traits or both. The `lookAt` takes precedence.
- With `initialCamera` trait, the bounding box values are required, which cover both `{west,south,east,north}` and `{west,south,east,north,position,direction,up}` cases.
- The `{positionHeading}` case is not added at this stage.

mwu2018:
I will need to update the catalog once this PR is approved and deployed.

tephenavies:
Yeah I think we don't need to add positionHeading. Why not have the following?

```
item.idealZoom?.lookAt?.targetLongitude
item.idealZoom?.lookAt?.targetLatitude
item.idealZoom?.lookAt?.range

item?.idealZoom?.west
item?.idealZoom?.east
item?.idealZoom?.north
item?.idealZoom?.south
item?.idealZoom?.position
item?.idealZoom?.direction
item?.idealZoom?.up
```

Then `idealZoom` would match `initialCamera`

mwu2018:
> Yeah I think we don't need to add positionHeading. Why not have the following?
> 
> ```
> item.idealZoom?.lookAt?.targetLongitude
> item.idealZoom?.lookAt?.targetLatitude
> item.idealZoom?.lookAt?.range
> 
> item?.idealZoom?.west
> item?.idealZoom?.east
> item?.idealZoom?.north
> item?.idealZoom?.south
> item?.idealZoom?.position
> item?.idealZoom?.direction
> item?.idealZoom?.up
> ```
> 
> Then `idealZoom` would match `initialCamera`

@steve9164  That will work but it might confuse the users. Using two different traits `lookAt` and `initialCamera` (or a better name) makes it clear that they are different things.

tephenavies:
Ok. What about?
```
item.idealZoom?.lookAt?.targetLongitude
item.idealZoom?.lookAt?.targetLatitude
item.idealZoom?.lookAt?.range

item?.idealZoom?.cameraPosition?.west
item?.idealZoom?.cameraPosition?.east
item?.idealZoom?.cameraPosition?.north
item?.idealZoom?.cameraPosition?.south
item?.idealZoom?.cameraPosition?.position
item?.idealZoom?.cameraPosition?.direction
item?.idealZoom?.cameraPosition?.up
```

mwu2018:
The only advantage of my original code is that it is easy to communicate to users. E.g. in the `CHANGES.md`, we can state 
```
`IDEAL ZOOM` can be customised by providing `lookAt` or `initialCamera` for `idealZoom` in `MappableTraits`. The `lookAt` takes precedence of `initialCamera` if both exist. The values for `initialCamera` can be easily obtained from shared link api.
```

tephenavies:
I don't think we should use the name `initialCamera`. We could do:
```
item.idealZoom?.lookAt?.targetLongitude
item.idealZoom?.lookAt?.targetLatitude
item.idealZoom?.lookAt?.range

item?.idealZoom?.camera?.west
item?.idealZoom?.camera?.east
item?.idealZoom?.camera?.north
item?.idealZoom?.camera?.south
item?.idealZoom?.camera?.position
item?.idealZoom?.camera?.direction
item?.idealZoom?.camera?.up
```

tephenavies (approved):

PR #6163, opened 2022-02-24
Wc story v2

### LINKED ISSUE #5427: Story v2.0 - Mobile - improve interaction on mobile devices
Example: https://maps.dea.ga.gov.au/#share=s-DEACoastlines&playStory=1 

Another issue we had with the DEA Coastlines story was that Terria currently seems to show different zoom level depending on the size of the screen the story is viewed on. For example, the DEA Coastlines layers have zoom-dependent symbology that transition at certain zoom levels. If a story is created on a large screen, but then viewed on a smaller screen (i.e. laptop or mobile), the difference in zoom levels causes different layers to show up than was originally intended (e.g. points showing when annual coastlines are described in the story scene text).

Check: 

- [x] missing arrow to go to next story scene
- [x] text modal covering the map almost entirely

wibily:
### What this PR does

Fixes #5427 
Cleans up the view story experience a bit.

Features implemented:
- General design refresh
- Ability to collapse down the story body text
- Clamped title (Title bar expands to allow 2 lines but will be ellipsed after that)

Features missing from Figma design:
- Pop out modal? to jump to a specific scene
- Colors
- Alignment issues
  
### Test me
http://ci.terria.io/wc-story-v2/#share=s-yDfpqsoYzZhkLHye1YmVTAF69Kf

  
### Checklist


lib/ReactViews/Story/StoryPanel/StoryBody.tsx
```diff
@@ -0,0 +1,54 @@
...
+import styled from "styled-components";
+
+const StoryContainer = styled.div<{ isCollapsed: boolean }>`
+  max-height: ${props => (props.isCollapsed ? 0 : "100px")};
+  @media (min-height: 700px) {
+    max-height: ${props => (props.isCollapsed ? 0 : "200px")};
+  }
+  @media (min-height: 900px) {
+    max-height: ${props => (props.isCollapsed ? 0 : "400px")};
+  }
```
na9da, line 13: For the media query, can we somehow use `${p => p.theme.sm}` (from variables [defined here](https://github.com/TerriaJS/terriajs/blob/edbf5deab37cd7adf6c957b76ae34c4a1e39c909/lib/Sass/common/_variables.scss#L158-L161))?

lib/Core/animation.ts
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+      const onEnd = () => {
+        element.removeEventListener("transitionend", onEnd);
+        resolve();
+      };
+      element.addEventListener("transitionend", onEnd);
+    }
+  });
+
+let timeoutID: ReturnType<typeof setTimeout>;
```
na9da: I'm not entirely sure about the scoping rules here. Won't `timeoutID` be set to the id of the last call which means `animateEnd` will clear only the last timeout event? So if `animationTimeout` is called multiple times, won't the earlier calls always result in a promise rejection?

Also is this used anywhere right now?

lib/ReactViews/Story/StoryPanel/StoryBody.tsx
wibily: thanks, will change

na9da:
Hey @wibily looks great so far, the code looks cleaner and good effort turning it to `tsx`.

Not a new bug, but I think the story title in the story side panel also needs to be truncated, but probably that's a different PR.

lib/Core/animation.ts
wibily: hmm i think you are right, I'll try and create some tests for it although it may need a testing library like enzyme to properly emulate the events, This is currently only used in StoryPanel.exitStory() however I would like to see this replace every instance where we are using a timeout to perform some sort of state change whilst waiting for an animation

na9da: > I would like to see this replace every instance where we are using a timeout to perform some sort of state change whilst waiting for an animation

That's a good idea. In that case it makes sense to document the purpose in the file header.

wibily: yup will do

lib/ReactViews/Story/StoryPanel/StoryBody.tsx
wibily: actually no we can't, this is min-height not max-width. Ie it is a special use case to figure out how high we want the story to be based on the vertical real estate of the screen

lib/Core/animation.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+/**
+ *
+ * @param element - HTML element that will fire the transitionend event due to a CSS animation
+ * @returns Promise that will resolve when the element finishes its animation or reject if it is not detected within ANIMATION_DURATION ms
+ */
+
+export const animateEnd = (element: Element | null) => {
+  let timeoutID: ReturnType<typeof setTimeout> | undefined = undefined;
```
wibily, line 51: note: this is only a partial solution. A proper solution would generalise the solution of cleaning up resources created inside Promises. E.g. if the timeout wins the Promise.race, the element never gets the transitionend event listener removed as we will need someway of referencing the onEnd function which is only definned inside the transitionEnd promise.

lib/ReactViews/Story/StoryPanel/StoryBody.tsx
na9da: ah ok got it :+1:

lib/Core/animation.ts
na9da: One simple solution might be to define `onTransitionEnd` and `oneEnd` as `const`s inside `animateEnd`, then `animateEnd` and `onTransitionEnd` can refer to the same `onEnd`.

na9da (approved):
Thanks @wibily, I have a left a suggestion for the animate method, but you can decide whether to implement it.

lib/Core/animation.ts
wibily: ahh i don't think i've got time to implement that so i'm going to merge it in as is for now

PR #6188, opened 2022-03-12
User styling (after split Compare)

nf-s:
### What this PR does

Fixes https://github.com/terriajs/product/issues/116 https://github.com/TerriaJS/terriajs/issues/961 https://github.com/TerriaJS/terriajs/issues/6169

### Changes

* Increase `ForkTsCheckerWebpackPlugin` memoryLimit to 4GB
* Add `renderInline` option to markdownToHtml/React + TSify files
* Organise `lib/Map` into folder structure
* Add `modelDimensionCallLoadMapItems` trait to `CatalogMemberTraits` - if `true` then `loadMapItems()` will be called when dimension values are changed.
* Add `featureCounts` to `GeoJsonMixin` - this tracks number of GeoJSON Features by type
* Add `polygon-stroke`, `polyline-stroke` and `marker-stroke` to GeoJSON `StyleTraits` - these are only applied to geojson-vt features (not Cesium Primitives)
* TableMixin manual region mapping dimensions are now in a `SelectableDimensionGroup`
* Fix misc font/color styles
* Create reusable `StyledTextArea` component
* `Collapsible` improvements:
  * Add `"checkbox"` `btnStyle`
  * `onToggle` can now stop event propagation
  * `title` now supports custom markdown
* Add `rightIcon` and `textLight` props to `Button`
* New `addTerriaScrollbarStyles` scss mixin
* `TableAutomaticStylesStratum` now creates `styles` for every column - but will hide columns depending on `TableColumnType`
* `TableAutomaticStylesStratum.numberFormatOptions` is now `TableStyle.numberFormatOptions`
* Implement `TableColorStyleTraits.legendTicks` - this will determine number of ticks for `ContinuousColorMap` legends
* `DiscreteColorMap` will now use `minimumValue`/`maximumValue` to calculate bins
* `SelectableDimensions` improvements
  * Add `color`, `text`, `numeric` and `button` types
  * Add `onToggle` function to `SelectableDimensionGroup`
  * `Group` and `CheckboxGroup` now share the same UI and use `Collapsible`
  * `enum` (previously `select`) now uses `react-select` component
  * `color` uses `react-color` component
  * `DimensionSelectorSection` / `DimensionSelector*` are now named the same as the model - eg `SelectableDimension`
* Create `Portal`, `PortalContainer`,`SidePanelContainer` and `WorkflowPanelContainer`. There are used by `WorkflowPanel`.
* Create `WorkflowPanel` - a basic building block for creating Workflows that sit on top of the workbench
  * It has three reusable components, `Panel`, `PanelButton`, `PanelMenu`
* Create `selectableDimensionWorkflow` - This uses `WorkflowPanel` to show `SelectableDimensions` in a separate side panel.
  * `TableStylingWorkflow` - set styling options for TableMixin models
  * `VectorStylingWorkflow` - this extends `TableStylingWorkflow` - used to set styling options for GeoJsonMixin models (for Protomaps/geojson-vt only)
* Create `viewingControls` concept. This can be used to add menu items to workbench items menu (eg "Remove", "Export", ...)
  * TSXify `ViewingControls`
* Add temporary `legendButton` property - this is used to show a "Custom" button above the Legend if custom styling has been applied
  * This uses new `TableStyle.isCustom` property
* Move workbench item controls from `WorkbenchItem.jsx` `WorkbenchItemControls.tsx`
  
#### Table styling

- Region mapping - continuous - http://ci.terria.io/user-styling-split-compare/#share=s-kuMzzVGzLgJTpWAM9pjxmKKA4cl
- Points - categorical - http://ci.terria.io/user-styling-split-compare/#share=s-tMT3SPdc3rDokzdgrW7PtxG6maJ

#### Vector styling

- GeoJSON points (continuous) http://ci.terria.io/user-styling-split-compare/#share=s-rOUf7XlgeZmzwFM6gqJEXnT0kz2
- GeoJSON points (categorical) http://ci.terria.io/user-styling-split-compare/#share=s-a0YTu1Sn0UVsizYg8iSmvYsLpOr
- GeoJSON lines (continuous - diverging) http://ci.terria.io/user-styling-split-compare/#share=s-rq5HBxSVpuDKrlkaX8AeYsCJqh7&clean&https://terria-catalogs-public.storage.googleapis.com/common/aus-gov-open-data/vic/prod.json
- GeoJSON polygons (discrete) http://ci.terria.io/user-styling-split-compare/#share=s-uwxllGCsLnQp3M7ZsfeIHgQYfXf&clean&https://terria-catalogs-public.storage.googleapis.com/common/aus-gov-open-data/vic/prod.json
  
#### Test link for clipping planes

Not related to this PR - but just to test CheckboxGroup is still working

http://ci.terria.io/user-styling-split-compare/#clean&https://gist.githubusercontent.com/na9da/3031b76246f3d5f6802b3ec1d56afd05/raw/f43383f9bb57f77d15ec082520fa122aa2f02b04/clipping-box-tilesets.json

### For another day

- [ ] Internationalisation
- [ ] Add more tests for `SelectableDimensions`
- [ ] Add tests for `Workflow...`
- [ ] Fix transparent circles in Protomaps

### Checklist

-   [x] Not many tests


lib/ModelMixins/CatalogMemberMixin.ts
```diff
@@ -180,16 +181,32 @@ function CatalogMemberMixin<T extends Constructor<CatalogMember>>(Base: T) {
...
             }
           }
         })) ?? []
       );
     }
 
+    @computed get viewingControls(): ViewingControl[] {
+      return [];
+    }
+
```
na9da, line 205: I love this addition.

Not part of this PR but one other consumer for viewing controls could be external plugins. External plugins might want to show viewing controls but for specific items or items that match some arbitrary condition, eg, it has a certain mixin.

One way of supporting that use case might be by adding a ~`Terria`~ `ViewState` observable called `viewingControlsMenu` which is an array of `(item) => ViewingControl | undefined` and it gets imported here.

lib/Models/SelectableDimensions/SelectableDimensionWorkflow.ts
```diff
@@ -0,0 +1,37 @@
...
+} from "./SelectableDimensions";
+
+/**
+ * Model for SelectableDimensionWorkflow. Basically just includes a bunch of selectableDimensions, name, icon and a catalog member.
+ * This can be used to create "workflows" that display SelectableDimensions in a side panel above the Workbench
+ *
+ * See `lib/ReactViews/Workflow/Workflows/SelectableDimension/SelectableDimensionWorkflow.tsx` for more info and usage
+ */
+export default interface SelectableDimensionWorkflow {
+  readonly type: string;
```
na9da: Doubt: what's the purpose of `type`?

lib/Models/Terria.ts
```diff
@@ -481,6 +482,12 @@ export default class Terria {
   @observable
   mapInteractionModeStack: MapInteractionMode[] = [];
 
+  @observable isWorkflowPanelActive = false;
+
+  /** Gets or sets the active SelectableDimensionWorkflow, if defined, then the workflow will be displayed using `WorkflowPanel` */
+  @observable
+  selectableDimensionWorkflow?: SelectableDimensionWorkflow;
+
```
na9da, line 512: Suggestion: I was thinking about how an external plugin component might launch a new workflow. It could set this observable directly but that seems like reaching into internal Terria state. Another option might be to declare `SelecatableDimensionsWorkflow` also as a namespace (just like `ViewingControls`) and add a function to the namespace called `runWorkflow(viewState: ViewState, workflow: ...)` or something. That way a plugin could call `SelectableDimensionsWorkflow.runWorkflow(viewState, workflow)`. We could also migrate this setting to `ViewState` which seems more appropriate.

lib/ModelMixins/CatalogMemberMixin.ts
```diff
@@ -180,16 +181,32 @@ function CatalogMemberMixin<T extends Constructor<CatalogMember>>(Base: T) {
...
+              // If no error and modelDimensionCallLoadMapItems is true, then call loadMapItems
+              if (
+                !result.error &&
+                this.modelDimensionCallLoadMapItems &&
+                MappableMixin.isMixedInto(this)
+              ) {
+                this.loadMapItems().then(loadMapItemsResult => {
+                  loadMapItemsResult.raiseError(this.terria);
+                });
+              }
```
na9da: Doubt: shouldn't this be called always? Considering in most cases the reload will happen only if any traits that affect loading has changed :thinking:

na9da:
Hey @nf-s, this looks fantastic. I love how customizable the whole workflow thing is now. Just made a few comments to look at.

lib/ModelMixins/CatalogMemberMixin.ts
nf-s: Yeah I love that idea! that would be super nice

lib/Models/SelectableDimensions/SelectableDimensionWorkflow.ts
nf-s: Nothing! I have removed it

lib/ModelMixins/CatalogMemberMixin.ts
nf-s: Yeah good point

lib/Models/Terria.ts
nf-s: Yep that sounds good too!
I'm not attached to this being in `Terria.ts` - I just put it in the easiest place for the moment 🙂

na9da (approved):
Looks good to merge!

PR #6190, opened 2022-03-14
Viewer change events

### LINKED ISSUE #4752: ViewerChanged events missing in next
V8 is missing the `afterViewerChanged` and `beforeViewerChanged` events

Is there maybe some replacement for them, [L58-L60](https://github.com/TerriaJS/terriajs/blob/next/lib/ViewModels/TerriaViewer.ts#L58-L60)

zoran995:
### What this PR does

Fixes #4752 and #6186

Hook viewer change events so they are raised after new cesium or leaflet viewer is created and before they are destroyed. Changed home camera to getter/setter so it will properly preserve home camera. Without that, a new viewer will be created on initialization while the camera view is still not read from the init file, and it won't be preserved.

Be aware that there is a workaround for this in #6186.

### Test me

Check if changes between viewers are working correctly. To confirm that events are raised correctly propper way is to debug some of the event listeners.
There is also another way to check this, start measuring in 3D and switch to 2D and try moving some point, points move without problem. Test same on the `main`, and notice that it doesn't work.

### Checklist

-   [x] I've updated relevant documentation in `doc/`. (not sure this is applicable)


tephenavies:
Thanks for the PR @zoran995!

tephenavies (approved):

PR #6198, opened 2022-03-16
Fixed Polyline position not being updated correctly in Leaflet mode

### LINKED ISSUE #6189: Polylines are not updated in Leaflet
in the _updatePolyline method of LeafletVisualizer.ts there is a section updating existing polyline positions:
``` 
const curLatLngs = polyline.getLatLngs();
let bPosChange = latlngs.length !== curLatLngs.length;
for (let i = 0; i < curLatLngs.length && !bPosChange; i++) {
  const latlng = curLatLngs[i];
  if (latlng instanceof L.LatLng && latlng.equals(latlngs[i])) {
    bPosChange = true;
  }
}
if (bPosChange) {
  polyline.setLatLngs(latlngs);
}
```
In v7 Js version of the same code:

```
var curLatLngs = polyline.getLatLngs();
var bPosChange = latlngs.length !== curLatLngs.length;
for (var i = 0; i < curLatLngs.length && !bPosChange; i++) {
  if (!curLatLngs[i].equals(latlngs[i])) {
    bPosChange = true;
  }
}
if (bPosChange) {
  polyline.setLatLngs(latlngs);
}
```

If I add a Not-operator to   **if (latlng instanceof L.LatLng && !latlng.equals(latlngs[i]))**  in the v8-version the polylines seems to update correctly. I'm just wondering if this is just a blunder during port to TS? or is there another reason behind this?

himby:
### Fixed Polyline position not being updated correctly in Leaflet mode

Fixes #6189

**LeafletVisualizer.ts**
Added a Not-operator to the condition that checks for position equality during update of polylines. This was probably missed during port to TypeScript.
  
### Test me
 
Add this CZML through MyData/Add Local Data
 [CzmlSpeedVectorTest.czml](https://gist.github.com/himby/0e6f3f68d8116f8a48ebfa449f0aabc0)

Before PR speed vectors are just drawn but not moving.
After PR the speed vector lines shall move together with point (billboard without image in this case)

### Checklist

    - Cannot find any unit tests for LeafletVisualizer at all.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Change Log
 * Added experimental routing system - there may be breaking changes to this system in subsequent patch releases for a short time. The routes currently include:
   * `/story/:share-id` ➡ loads share JSON from a URL `${configParameters.storyRouteUrlPrefix}:share-id` (`configParameters.storyRouteUrlPrefix` must have a trailing slash)
   * `/catalog/:id` ➡ opens the data catalogue to the specified member
+* Fixed a polyline position update bug in `LeafletVisualizer`. Polylines with time varying position will now correctly animate in leaflet mode.
 * [The next improvement]
 
 #### 8.1.25 - 2022-03-16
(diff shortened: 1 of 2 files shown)


na9da (approved):
Good catch @himby!

Thanks for the fix.

PR #6200, opened 2022-03-16
Fix #6180 - button cursor

### LINKED ISSUE #6180: Button cursor: default
Something I just noticed while updating #5344, is it desired to have buttons with `cursor: default` instead of `cursor: pointer` 
i.e. navigation buttons, plus button in data catalog, ideal zoom, about data. Most of those are using RawButton.
 image    image    image  
image  

This is introduced with an upgrade of normalize.css, normalize previously defined `cursor: pointer` for buttons (https://github.com/TerriaJS/terriajs/commit/20d5c316e4c3b3f55928f04c3a7e233f687034d3#diff-7ed9bbf182b4ef4dc6ef6c041d64dd629ee18efee578030877640f0db7b275f9L280-L286)

zoran995:
### What this PR does

Fixes #6180

Changes the cursor for buttons to be pointer again.

discussed https://github.com/TerriaJS/terriajs/pull/5344#issuecomment-1067402317
> Yep, pls make whatever changes are required to communicate that active buttons are clickable. We should be more consistent for hover states across the app (that's on me), but as a general rule I would suggest toggling brightness +/- 10.
  
### Test me
  
### Checklist


nf-s (approved):
Thanks!

PR #6211, opened 2022-03-27
Upgrade to CesiumJS v1.91

kring:
Requires https://github.com/TerriaJS/cesium/pull/83, which I've already published to npm as terriajs-cesium 1.91.0.

I haven't gone crazy testing this, but most things I've tried work well. Some known problems:

* [x] KMZ files can't be decompressed because the zip.js web worker isn't found correctly.
* [x] 3D Tiles crashing on `const hasOutlines = hasExtension(model.gltf, "TERRIA_solid_outlines");` in `Model.js`.
* [x] I haven't tried it due to the above, but I wouldn't be surprised to see problems resolving the Draco webassembly module.
* [x] There are some test failures.

Don't merge this until the above are fixed.


kring:
This should be in good shape now, ready for review.
Next up: upgrade from Cesium 1.91 to 1.92.

nf-s (approved):
Thanks @kring !

PR #6217, opened 2022-03-29
fix data preview error messages + loading

nf-s:
### Fix data preview error messages + loading

Related to https://github.com/TerriaJS/terriajs/issues/6214 https://github.com/TerriaJS/terriajs/issues/6197
  
### Test me

After vs Before

image

#### Before

- [Go to `main` CI
](http://ci.terria.io/main/)
- Click "Magda Item"
- See no error message

#### After
  
- [Go to branch CI](http://ci.terria.io/fix-data-preview-error-loading/)
- Click "Magda Item"
- See error message

### Checklist

### DIFF
--- a/lib/ModelMixins/ReferenceMixin.ts
+++ b/lib/ModelMixins/ReferenceMixin.ts
@@ -32,32 +32,30 @@ function ReferenceMixin<T extends Constructor<Model<ReferenceTraits>>>(
     @observable
     private _target: BaseModel | undefined;
 
-    private _referenceLoader = new AsyncLoader(() => {
+    private _referenceLoader = new AsyncLoader(async () => {
       const previousTarget = untracked(() => this._target);
-      return this.forceLoadReference(previousTarget).then(target => {
-        if (!target) {
-          throw new DeveloperError("Failed to create reference");
-        }
+      const target = await this.forceLoadReference(previousTarget);
 
-        if (target?.uniqueId !== this.uniqueId) {
-          throw new DeveloperError(
-            "The model returned by `forceLoadReference` must be constructed with its `uniqueId` set to the same value as the Reference model."
-          );
-        }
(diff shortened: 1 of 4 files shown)


lib/ModelMixins/ReferenceMixin.ts
```diff
@@ -32,32 +32,30 @@ function ReferenceMixin<T extends Constructor<Model<ReferenceTraits>>>(
...
+          "The model returned by `forceLoadReference` must be constructed with its `sourceReference` set to the Reference model."
+        );
+      }
+      if (this.weakReference && target?.sourceReference) {
+        throw new DeveloperError(
+          'This is a "weak" reference, so the model returned by `forceLoadReference` must not have a `sourceReference` set.'
+        );
+      }
+      runInAction(() => {
+        this._target = target;
```
nf-s, line 58: Just converted to async function

lib/ReactViewModels/ViewState.ts
```diff
@@ -562,16 +565,11 @@ export default class ViewState {
           return Result.error(`Could not view catalog member ${getName(item)}`);
         }
       }
-      const theItem: BaseModel =
-        ReferenceMixin.isMixedInto(item) && item.target ? item.target : item;
```
na9da, line 566: Doubt: why don't we need to resolve the target any more? Is it because we are already recursively calling `viewCatalogMember` when item is a reference?

na9da (approved):
Looks good!

lib/ReactViewModels/ViewState.ts
nf-s: Yep, it should be handled by

https://github.com/TerriaJS/terriajs/blob/e6615c39b89eb209330008f0b824ab15da106d26/lib/ReactViewModels/ViewState.ts#L560-L567

PR #6218, opened 2022-03-29
Add `itemProperties`, `itemPropertiesByType` and `itemPropertiesByIds`

### LINKED ISSUE #1052: `itemProperties` is not applied to groups inside a CkanCatalogGroup
When all the data for a map comes from one CKAN, it makes sense to have the CkanCatalogGroup open by default and immediately start populating. Requested for City of Sydney ESP.

nf-s:
### Add `itemProperties`, `itemPropertiesByType` and `itemPropertiesByIds`

Fixes https://github.com/TerriaJS/terriajs/issues/1052 https://github.com/TerriaJS/terriajs/issues/5631 https://github.com/TerriaJS/terriajs/issues/5155 https://github.com/TerriaJS/terriajs/issues/1408 https://github.com/terriajs/nationalmap/issues/1129

* Add `itemProperties`, `itemPropertiesByType` and `itemPropertiesByIds` to `GroupTraits` and `ReferenceTraits`.
  * Properties set `override` strata
  * Item properties will be set in the following order (highest to lowest priority) `itemPropertiesByIds`, `itemPropertiesByType`, `itemProperties`.
  * If a parent group has `itemProperties`, `itemPropertiesByType` or `itemPropertiesByIds` - then child groups will have these values copied to `underride` when the parent group is loaded
  * Similarly with references.
* All dynamic groups (eg `WebMapServiceCatalogGroup`) will create members and set `definition` strata (instead of `underride`)

### New docs

These are a bit rough

- [traits-in-depth.md](https://github.com/TerriaJS/terriajs/blob/2a32ee6935d03ed039976b4fa3d11e9b22b96d3f/doc/contributing/traits-in-depth.md)
- [strata-examples.md](https://github.com/TerriaJS/terriajs/blob/2a32ee6935d03ed039976b4fa3d11e9b22b96d3f/doc/contributing/strata-examples.md)
  
### Background

For the [SIP dataset styling issue](https://github.com/terriajs/nationalmap/issues/1129) (in NationalMap) - I’ve been looking into fleshing out `itemProperties` so I can attach custom styling to each dataset.
The SIP datasets are harvested from CKAN (data.gov.au) - so we can’t use itemProperties in its current form.

### Test me

A bunch of CSVs loaded from CKAN - with overridden styles
  
http://ci.terria.io/item-props/#clean&https://gist.githubusercontent.com/nf-s/67a545b72eef135a9d4903e2a65b9d9b/raw/02fcfcabfc37ad6eedb2908aa0c52ee39bcf6c0c/item-props-by-id.json&share=s-aAu23kmvBQxDNGwoVU3zQxjiGry

#### Example

https://gist.githubusercontent.com/nf-s/67a545b72eef135a9d4903e2a65b9d9b/raw/02fcfcabfc37ad6eedb2908aa0c52ee39bcf6c0c/item-props-by-id.json

```json
"catalog": [
    {
      "type": "ckan-group",
      "name": "Data.gov.au",
      "url": "https://data.gov.au",
      "itemPropertiesByIds": [
        {
          "ids": ["6TbYz2Jj/01809262-b20a-4fa2-96d2-73114722f364/aa8d7eea-9eed-42e5-a322-1abee690c49a"],
          "itemProperties": {
            "activeStyle": "",
            "defaultStyle": {
              "color": {
                "nullColor": "#F2F3F4"
              }
            }
          }
        }
```

### Checklist


lib/Traits/TraitsClasses/ItemPropertiesTraits.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export class ItemPropertiesByIdTraits extends ModelTraits {
+  @primitiveTrait({
+    name: "ID of model",
+    description:
+      "The ID of model to apply `itemProperties` to. This must be defined.",
+    type: "string"
+  })
+  id?: string;
```
tephenavies: Should this be an array (`ids`)? I think many times we'll want to apply to multiple items.

nf-s: Good idea! done

tephenavies (approved):
This is really good! I've wanted something like this for a while.

PR #6219, opened 2022-03-29
Load translations instead of bundling

tephenavies:
### What this PR does

Trying out my idea from https://github.com/TerriaJS/terriajs/pull/5891#issuecomment-948598146:

> Move translations in terriajs repo from `lib/Language/en/translation.json` to `wwwroot/languages/en/translation.json` (or similar in wwwroot) and then (somehow) pull translations from the either the URL `languages/{{lng}}/translation.json` or `build/TerriaJS/languages/{{lng}}/translation.json` based on an override system (TerriaMap translations override TerriaJS translations)
> 
> Maybe the override system can be simple. Translations from the `translation` namespace could always come from `build/TerriaJS/languages/{{lng}}/translation.json` (i.e. terriajs repo) and `languageOverrides` could always come from `languages/{{lng}}/languageOverrides.json`

### How does this improve i18n?

1. **It's much easier to add a language to your app.**

To add a weblate language in TerriaMap config.json:
```json
{
  "parameters": {
    "languageConfiguration": {
      "enabled": true,
      "debug": false,
      "languages": {
        "en": "English",
        "fr": "Français",
        "de": "Deutsch"
      },
      "fallbackLanguage": "en"
    },
  }
}
```

2. **JSON files for languages are loaded as you switch to that language (previously both en & fr were included in the bundle)**
3. **There's a clearer distinction between namespaces**
    * "translation":translations that come with terriajs
    * "languageOverrides": translations that you include in your TerriaMap

The translation file for a language is loaded when you switch to the language

This may improve the situation discussed in https://github.com/TerriaJS/terriajs/discussions/6213

  
### Test me

Try http://ci.terria.io/improve-i18n-tjs/#configUrl=https://gist.githubusercontent.com/steve9164/1ea036a6cba44a63636fe8f1cc9689aa/raw/f9e3df70e7e0b25ed47fa00f0e54874da8d18f76/custom-lngs.json

See [config gist](https://gist.githubusercontent.com/steve9164/1ea036a6cba44a63636fe8f1cc9689aa/raw/f9e3df70e7e0b25ed47fa00f0e54874da8d18f76/custom-lngs.json)
  
Or build this locally against https://github.com/TerriaJS/TerriaMap/tree/improve-i18n. Use config pasted above to test loading multiple languages.

### Checklist

-   [x] No unit tests for loading defined languages
- [x] Revert TerriaJS/terriajs@0901bcd
- [x] Ensure that the latest translations are synced on merge
- [x] Test changing directory for translation files in weblate
- [x] Change the directory for translations in weblate - in [settings](https://hosted.weblate.org/settings/terriajs/terriajsnext/)


arky:
+1 
Please kindly merge this PR that improves localization support in Terriajs project.

arky:
Thank you @steve9164 for your work. Kudos!

tephenavies:
@zoran995 Could you take a look at this and let me know what you think?

lib/Models/Internationalization.ts
```diff
@@ -121,18 +120,13 @@ class Internationalization {
...
-        },
-
         backend: Object.assign(
           {
-            loadPath: "/languages/{{lng}}/{{ns}}.json",
+            loadPath: function loadPath(lngs: string[], namespaces: string[]) {
+              return namespaces[0] === "translation"
+                ? `${terriajsResourcesBaseUrl}languages/{{lng}}/{{ns}}.json`
+                : "languages/{{lng}}/{{ns}}.json";
+            },
```
tephenavies: Switch translation loading from importing JSON files (including translation in bundle) to loading on demand.

zoran995:
@steve9164 I am perfectly fine with this change 

Just might consider leaving English bundled, so users don't end up with something like this
image

it shouldn't happen in normal cases, this is with 10s request throttle, and it won't happen when English is fallback as it is already bundled

zoran995:
@steve9164 an idea, what about using weblate API for retrieving translation files
i.e. https://hosted.weblate.org/api/translations/terriajs/terriajsnext/en/file/
this might be a fallback when translation file is not found in `wwwroot/languages`, and also a flag to not use it at all (i.e. not used by default and then activate it for ci). It might simplify translation and changes can be easily checked in a live version of the app

p.s. haven't read weblate TOC

tephenavies:
I tried various throttling to get raw translation strings displaying, but never got it. I think if it's possible to get raw translation strings we can fix it through a `<link rel="preload" href="build/TerriaJS/languages/en/translation.json" as="fetch">` that we add to the DOM in `entry.js` (so it won't be loaded before `entry.js` is rendered, but will be preloaded before `index.js`). We could set the preload href in a paramter to `gulp build|release` (default to the standard en path, allow setting your own preload path, or disabling the preload).

tephenavies:
I don't want to point out to external services for UI strings by default. Possibly we could include a docs section about how to do that.

zoran995 (approved):
Maybe provide a config option to add an additional translation source, leaving that to you to decide if it's needed, other than that LGTM

arky:
I guess the default language is detected by checking the user agents locale. 

Is it possible to force a language switch by passing a variable such 'language: language-code_variant (km, mm, zh)' in TerriaJS config or passing URL arguments? The documentation doesn't go into much detail. 

https://docs.terria.io/guide/customizing/client-side-config/#languageconfiguration

zoran995:
@arky You should be able to use lng query parameter as per
https://github.com/TerriaJS/terriajs/blob/0901bcdfc21e9aec9a1458da3a8c4444efe9d7de/lib/Models/Internationalization.ts#L47-L52
and
https://github.com/TerriaJS/terriajs/blob/0901bcdfc21e9aec9a1458da3a8c4444efe9d7de/lib/Models/Internationalization.ts#L140
i.e. `?lng=fr` should open app with French language, but haven't tested it, writing from phone

You can provide changeLanguageOnStartWhen​ from config, if I recall correctly

I don't think there is a other way to force start language as we use language detector, AFAIK

lib/Models/Internationalization.ts
```diff
@@ -121,18 +120,13 @@ class Internationalization {
...
-          }
-        },
-
         backend: Object.assign(
           {
-            loadPath: "/languages/{{lng}}/{{ns}}.json",
+            loadPath: function loadPath(lngs: string[], namespaces: string[]) {
+              return namespaces[0] === "translation"
+                ? `${terriajsResourcesBaseUrl}languages/{{lng}}/{{ns}}.json`
+                : "languages/{{lng}}/{{ns}}.json";
```
na9da, line 133: Maybe add a comment on what other namespace we expect here? Eg: `languageOverrides`

na9da (approved):
Looks good to me @steve9164!

PR #6236, opened 2022-04-04
Api changes

na9da:
### What this PR does

Simplifies some of the APIs and fixes a few bugs.

* Added [ViewingControlsMenu.addMenuItem](https://github.com/TerriaJS/terriajs/blob/4c56f1227657bed4ac16d874b1d9d8622138fc9b/lib/ViewModels/ViewingControlsMenu.ts#L15) for making catalog wide extensions to viewing controls options.
* Added [MapToolbar](https://github.com/TerriaJS/terriajs/blob/1ddc0521c84764bfdb004939e87f6c62ecfc67c3/lib/ViewModels/MapNavigation/MapToolbar.ts#L86), a simpler API for adding buttons to the map navigation menu for the most common uses cases. It uses the MapNavigationModel, which is very customizable but simplifies it for common use cases by removing some of the boilerplate.
* Added `BoxDrawing` creation methods `fromTransform` and `fromTranslationRotationScale`.
* Fixed a bug where `zoom` hangs for catalog items with trait named `position` ([commit](https://github.com/TerriaJS/terriajs/commit/e9bf325044792296572a35d45067df0967a16316))
* Moved workflows to `Models/Workflows` and added helper method [runWorkflow](https://github.com/TerriaJS/terriajs/blob/1ddc0521c84764bfdb004939e87f6c62ecfc67c3/lib/Models/Workflows/SelectableDimensionWorkflow.ts#L41) to invoke a workflow.

### Test me
  
~How should reviewers test this? (Hint: If you want to provide a test catalog item, [create a Gist](https://gist.github.com/) of its catalog JSON, add its url and your branch name to this url: `http://ci.terria.io/<branch name>/#clean&<raw url of your gist>` , and then paste that in the body of this PR)~

Mostly just API changes.

### Checklist


lib/Models/Cesium.ts
```diff
@@ -709,7 +709,7 @@ export default class Cesium extends GlobeOrMap {
             });
           }
         });
-      } else if (target.position !== undefined) {
+      } else if (target.position instanceof Cartesian3) {
```
na9da, line 712: When `target` is a catalog item with `position` trait, it resulted in zoom hanging forever. This more specific check prevents it.

lib/Models/Workflows/SelectableDimensionWorkflow.ts
```diff
@@ -25,6 +31,29 @@ export default interface SelectableDimensionWorkflow {
...
 
+namespace SelectableDimensionWorkflow {
+  /**
+   * Runs a selectable dimension workflow which is a workflow for a workbench item.
+   *
+   * @param viewStateOrTerria - The {@link ViewState} or {@link Terria} instance
+   * @param workflow - A {@link SelectableDimensionWorkflow} instance
+   */
+  export function runWorkflow(
+    viewStateOrTerria: ViewState | Terria,
```
na9da: Why `ViewState | Terria`? It can be confusing for an end developer to decide when to use `viewState` and when to use `terria`. So I think, as much as possible, our APIs must accept `viewState`. But models do not have access to `viewState` so for this method to be usable from model layer we have to also accept `terria`.

lib/ModelMixins/ClippingMixin.ts
```diff
@@ -200,50 +200,54 @@ function ClippingMixin<T extends Constructor<BaseType>>(
...
-          keepBoxAboveGround: this.clippingBox.keepBoxAboveGround,
-          onChange: action(({ modelMatrix, isFinished }) => {
-            Matrix4.multiply(
-              this.inverseClippingPlanesOriginMatrix,
-              modelMatrix,
-              this.clippingPlaneModelMatrix
-            );
-            if (isFinished) {
-              const position = Matrix4.getTranslation(
+        this._clippingBoxDrawing = BoxDrawing.fromTransform(
```
na9da, line 203: The real change here is replacing `new BoxDrawing` with `BoxDrawing.fromTransform`, rest is indentation change.

lib/Core/types.ts
```diff
@@ -1 +1,15 @@
...
+ * @example
+ * ```ts
+ *   Optional<Button, "id" | "name">
+ * ``
+ */
+export type Optional<T extends object, K extends keyof T = keyof T> = Omit<
+  T,
+  K
+> &
+  Partial<Pick<T, K>>;
```
nf-s: This file probs doesn't need to exist -  as we have [lib/Core/TypeModifiers.ts](https://github.com/TerriaJS/terriajs//blob/2b482eba0802e7f5f2100aeaa0d485187478aeed/lib/Core/TypeModifiers.ts)

lib/ReactViewModels/ViewState.ts
```diff
@@ -145,7 +158,13 @@ export default class ViewState {
...
   }
 
-  @observable workbenchWithOpenControls: string | undefined = undefined;
+  /**
+   * ID of the workbench item which currently has an open ViewingControls menu.
+   *
+   * @private
+   */
+  @observable
+  workbenchItemWithOpenControls: string | undefined = undefined;
```
nf-s: Is this supposed to be `@private` here?

lib/ReactViewModels/ViewState.ts
```diff
@@ -104,6 +105,18 @@ export default class ViewState {
...
+   * for the given catalog item instance.  This is useful for plugins to extend
+   * the viewing control menu across catalog items.
+   *
+   * Use {@link ViewingControlsMenu.addMenuOption} instead of updating directly.
+   */
+  @observable
+  readonly globalViewingControlOptions: ((
+    item: CatalogMemberMixin.Instance
+  ) => ViewingControl | undefined)[] = [];
+
```
nf-s: ❤️

nf-s (changes requested):
Love it!
Just a few small things to clarify - and our discussion about use of `namespace` in Slack

lib/Core/types.ts
na9da: I missed that somehow. Moved `Optional` to TypeModifiers.ts

nf-s (approved):
💟

PR #6241, opened 2022-04-06
GTFS Coloured Models

tephenavies:
### What this PR does

* Adds the ability to colour models by line in GTFS
* Fixes a bug with cross-origin billboard images in Leaflet
  
### Test me
  
See http://ci.terria.io/vic-gtfs/#clean&https://gist.githubusercontent.com/steve9164/bc8413f73e44094f02de8a9f72a9c2f6/raw/4e3c00ee3a6b02af40051c3f6cd63133eddf6182/vic-gtfs.json

### Checklist

-   [x] Added unit tests


lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
```diff
@@ -475,3 +529,19 @@ function updateBbox(lat: number, lon: number, rectangle: RectangleExtent) {
...
+ */
+function jsonPathLike(data: JsonObject, path: string): any {
+  if (data === undefined) return undefined;
+  const index = path.indexOf(".");
+  if (index === -1) return data[path];
+  return jsonPathLike(
+    data[path.slice(0, index)] as JsonObject,
+    path.slice(index + 1)
+  );
+}
```
na9da: Suggestion: There's also the [lodash.get](https://lodash.com/docs/4.17.15#get) function we could use.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
```diff
@@ -163,8 +167,31 @@ export default class GtfsCatalogItem extends MappableMixin(
...
+          const gtfsEntity: FeedEntity = data.featureInfo?.get("entity");
+          const value = jsonPathLike(
+            gtfsEntity as JsonObject,
+            this.model.colorModelsByProperty.property!
+          );
+          if (value !== undefined) {
+            let breakLoop = false;
+            this.model.colorModelsByProperty.colorGroups.forEach(
+              (colorGroup, i) => {
+                if (breakLoop || colorGroup.regExp === undefined) return;
```
na9da: Possible bug: `breakLoop` here might not serve its intended purpose of early returning once we hit a matching regexp. The `return` only returns from the arrow function but `forEach` will continue iterating the full array. We might have to rewrite using `for` or `find()`.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
```diff
@@ -279,19 +306,44 @@ export default class GtfsCatalogItem extends MappableMixin(
...
+      maximumScale: new ConstantProperty(this.model.maximumScale),
+      minimumPixelSize: new ConstantProperty(this.model.minimumPixelSize ?? 0),
+      shadows: ShadowMode.DISABLED
     };
 
     return new ModelGraphics(options);
   }
 
+  @computed
+  private get _coloredModels() {
```
na9da, line 321: We might want this to return `undefined` when in 2d mode. Right now the map freezes with some LeafletVisualizer errors when switching to 2d.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
```diff
@@ -279,19 +306,44 @@ export default class GtfsCatalogItem extends MappableMixin(
...
+      return undefined;
+    }
+    return colorGroups.map(({ color }) => {
+      const coloredModel = model.clone();
+      coloredModel.color = new ConstantProperty(
+        Color.fromCssColorString(color ?? "white")
+      );
+      coloredModel.colorBlendMode = new ConstantProperty(
+        ColorBlendMode.REPLACE
+      );
```
na9da: Just a suggestion: We could use `ColorBlendMode.MIX` and change the gltf material to something like:

```
"pbrMetallicRoughness": {
        "baseColorFactor": [
          1.0,
          1.0,
          1.0,
          0.5
        ],
        "metallicFactor": 1.0,
        "roughnessFactor": 0
      }
```

which in my opinion looks more "3d" :) 

image

na9da (commented):
Looks great @steve9164. I liked the overall implementation of it. Have left a few comments to consider.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: How do I change the gltf material?

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: Yeah good point. I probably want `findIndex`.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: I'll try to debug those LeafletVisualizer errors.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
na9da: You can use https://github.com/CesiumGS/gltf-pipeline CLI to convert the glb to gltf, edit json and then back to glb.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: The LeafletVisualizer errors are because the image in cross-origin. Maybe billboards in LeafletVisualizer need a cross origin attribute?

tephenavies: The LeafletVisualizer errors occur inside `recolorBillboard` when the billboard `color !== Color.WHITE`. This is becuase the image is fetched without `crossOrigin="anonymous"`.
One funny thing is that the layer works in 2d if you change the opacity to 100% (opacity is dealt with through the `color` attribute of the billboard).

tephenavies: Fixed

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: This does look good. I've updated my model and GtfsCatalogItem

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: Fixed

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: Ahh, I thought there must be an easy way to do this, but forgot to check lodash. Fixed.

lib/Models/Catalog/Gtfs/GtfsCatalogItem.ts
tephenavies: Just FYI we shouldn't return different things from mapItems based on `currentViewer` (or any other way of distinguishing Cesium/Leaflet) because models are meant to be independent of any viewer, and sometimes can be used simultaneously in more than 1 viewer (e.g. preview map running Leaflet and main map running Cesium).

na9da: Good point!

na9da (approved):
Great effort @steve9164 !

PR #6254, opened 2022-04-13
Upgrade to CesiumJS v1.92

kring:
This is a PR into #6211 for ease of review.

Requires TerriaJS/cesium#84.

This looks like a lot of changes, but most of them are just to remove `makeRealPromise`, because CesiumJS uses real promises now.

PR #6258, opened 2022-04-19
Add support for using the splitter with 3D Tiles

### LINKED ISSUE #4965: Support splitting 3d tiles
As an end user of Terria,
I want to be able to use the splitter for 3D tiles, not just 2D imagery,
So that I can compare 2 models side by side.

---

The "proper" way to do this is to use a similar approach to the one taken with terrain rather than use Cesium's higher level clipping plane API (for code cleanliness and performance reasons). Basically, you need to add an extra parameter to the 3D tiles fragment shaders, and extend Cesium's API to allow setting that parameter from Terria. I believe the source for the 3dtiles shaders mostly live in the js source files, like https://github.com/CesiumGS/cesium/blob/0482770ae8e7e90f8bdfbb56fa194981f800d156/Source/Scene/Cesium3DTileBatchTable.js

@kring did something similar for terrain already, see [this commit](https://github.com/CesiumGS/cesium/commit/2916a07f23e4eed36be6972378a8de0e94ac7e9c)

Ideally, these changes should be made in upstream CesiumGS and not just our fork because it could be useful to other Cesium users too, and not just TerriaJS folks.

kring:
This is a PR into #6254, so merge that first.

Fixes #4965

image


nf-s (approved):
Thanks a lot @kring for all your work getting this feature in - I love it!

PR #6272, opened 2022-04-28
Progress bar visibility #6144

### LINKED ISSUE #6144: Progress bar visibility
Progress bar is not visible on light color base maps (such as positron) making user wonder what is going on
image
If I didn't know it is there I wouldn't see it. We could use base map `contrastColor` for defining a dynamic color of the progress bar

staffordsmith83:
### What this PR does

Fixes #6144 

When rendering `ProgressBar` get the current `baseMapContrastColor` and use that for the ProgressBar `background-color`.

As this is a computed value solution required wrapping the return value of `createReactClass()` in `observer()`

Screen Shot 2022-04-28 at 11 37 44 am (2)

  
### Test me
  
Manually toggle the basemap to different settings in Map Settings dialog. If necessary enable throttling on Network tab of dev tools to slow down loading and observe the changes.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 
 #### next release (8.2.4)
 
+* ProgressBar colour now depends on baseMapContrastColor - improves visibility on light map backgrounds
 * [The next improvement]
 
 #### 8.2.3 - 2022-04-22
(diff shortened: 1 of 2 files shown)


staffordsmith83:
Will open new PR with different branch name with no `#` character

PR #6282, opened 2022-05-06
Exclude story from share

### LINKED ISSUE #5756: Allow users to exclude story from sharelink
## Issue

An increasingly common use case for Terria stories is to provide an easy way to access a preloaded set of layers or products (e.g. https://maps.dea.ga.gov.au/#share=s-DEACoastlines&playStory=1). 

Once a user completes a story and starts to explore the layers/product, they often wish to share examples of interesting locations of interest they have found in the data. Currently, a user can achieve this by creating a share link via "Share/print". However, these share links currently include a link to the original story too. This means that when a user visits this new share link, they get this result with the story dialogue ("This map contains a story"):

image

For new users who are not familiar with Terria, the natural response to seeing this pop-up is to click "Yes", which takes them away from the specific share link scene captured by the original user. Currently, the only way to exclude the story from a share link is to delete the entire story manually then copy the link, which is not intuitive.

## Possible solution

It would be fantastic to have a "Include story" checkbox in the "Share/print" popup so that users can choose to exclude the story when creating a share link. This would allow share links to be created from a shared story link without including the original story itself:

image

staffordsmith83:
### What this PR does

Fixes #5756

"Allow users to exclude story from sharelink"

This allows users to indicate whether stories should be included in a share.

STORIES ARE INCLUDED BY DEFAULT. But users can exclude them:
When clicking the Share/Print button, and going to Advanced Options, the user has an option to toggle:
image

The state of this choice is remembered for a browser session.

When sharing from the Story Editor panel, include stories is forced to true and the checkbox is not rendered:
image

To facilitate this I have added a new property and setter to ViewState:
```ts
...
  @observable includeStoryInShare: boolean = false;
...
  @action
  setIncludeStoryInShare(bool: boolean) {
    this.includeStoryInShare = bool;
  }
```

This means that a user choice will persist for a browser session.

Some notes:
- Pass includeStories option to buildShareLink depending on storyShare prop.
- When sharing from Story editor, stories will be automatically included, otherwise will check viewState.includeStoryInShare property.
  
### Test me
  
Test in the browser.

### Checklist


staffordsmith83:
Rebased onto current main.

staffordsmith83:
This PR works in isolation, but when merged with the latest main there is a problem: When sharing a map with stories, the prompt 'This map includes a story, would you like to play it now?' no longer appears. This may be related to recent work TSIfying the StoryBuilder?

Maybe @na9da do you know about this functionality?

tephenavies:
Looks like the issue is with `main`: http://ci.terria.io/main/#share=s-wgs7rCoDmYIadOx66cB94UyLqq3

staffordsmith83:
@steve9164 yep I think it may be to do with the conversion of StandardUserInterface to typescript. There was that other bug  the other day and fixed it adding empty array as second param to useEffect()... Do we need to add `props.viewState.storyShown` to the dependency array or something?

staffordsmith83:
The issue with the Story Prompt modal is to do with changes to TerriaMap repo.
This error is independent of this PR - this PR tested against previous version of TerriaMap and works.

image

CHANGES.md
```diff
@@ -3,6 +3,8 @@ Change Log
 
 #### next release (8.2.5)
 
+* Allow users to choose whether Story is included in Share
+* Fix `CatalogIndex` types
```
tephenavies:
```suggestion
```

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.ts
```diff
@@ -110,7 +110,7 @@ export async function buildShortShareLink(
 export function getShareData(
   terria: Terria,
   viewState?: ViewState,
-  options = { includeStories: true }
+  options = { includeStories: viewState?.includeStoryInShare ?? true }
```
tephenavies: Each of the share link/share data functions should have the same defaults. And after I think you should revert all changes made to this file.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -320,8 +341,8 @@ const SharePanel = observer(
                   )
                 }
               />
-              {this.renderWarning()}
```
tephenavies: Why has this been removed?

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -320,8 +341,8 @@ const SharePanel = observer(
...
             </div>
+            <div></div>
```
tephenavies:
```suggestion
```

tephenavies (changes requested):

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
staffordsmith83: Ah mistake!

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.ts
staffordsmith83: Implemented

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
staffordsmith83: Reinstated

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
staffordsmith83: Implemented

tephenavies (approved):

PR #6297, opened 2022-05-20
Editor tinymce

### LINKED ISSUE #3828: Ability to resize images in Story text
At the moment we have no control of the size of images in Story text. We should enable control of the size of the image.

### LINKED ISSUE #5595: Story v3.0 - Improve editing stories
- Once formatted text is pasted into the body of the scene text editor, it was very difficult to revert to the original default formatting (e.g. font and font colour). I was pasting content from a word document and accidentally forgot to paste as plain text, resulting in non-matching formatting. To make the text consistent with my other scenes, I had to copy existing text from another scene, then plain text paste my new text into that to restore the default formatting (simply clearing the badly formatted text then pasting as plain text still retained the incorrect formatting). Proposed solution: include a "restore default formatting"/"convert to raw text"/"remove formatting" button that would restore all formatting to the defaults.

- Email links don't currently seem to be supported. I tried both "dea@ga.gov.au" (which was converted into a relative link), and "mailto:dea@ga.gov.au" (which was converted to plain text)

- In addition to Header 1 and Header 2 options, I would have liked an option for smaller caption-style text too. This would allow us to include a figure caption or reference below any included images in a less distracting/obvious way

### LINKED ISSUE #6309: Story Prompt bug: shows up immediately after adding Capture New Scene
Story Prompt Error

Since Wednesday 1st of June there is a bug present where immediately after adding a new story scene, and saving, the Story Prompt ("This map contains a story,  Would you like to view it now?") is displayed. Clicking 'Maybe Later' on the prompt does nothing, while clicking 'Yes' closes the prompt but does not play the story.

This erroneous behaviour only occurs when adding the first scene to a story. Adding subsequent scenes does not trigger the prompt.

staffordsmith83:
### What this PR does

Fixes #5595 
Fixes #3828
Fixes #6309 

Replaced Pell WYSIWYG editor library with TinyMCE library.
This is used to create the HTML content for each scene in a story.

The main improvements implemented with this upgrade are the following:

To address issue https://github.com/TerriaJS/terriajs/issues/3828
Set a maximum height for inserted images, so that an image will be visible in its entirety in the viewer without vertical scrolling. Aspect ratio will be preserved.

Allow image resizing with drag handles on corners.

image

To address issue # Story v3.0 - Improve editing stories #5595:
1. If pasting formatted text from Word etc, font and colour etc will be brought across. This formatting can be removed and reset to the default content style of the editor with a new button 'Clear formatting':
Pasted image 20220602100000

2. Email links: I have used the tinyMCE autolink plugin to enable auto linking of email addresses.

3. The full range of headings is now available:
image

This PR also fixes a bug whereby the 'Would You Like to View It Now?' Story Prompt is incorrelty shown after creating the first story scene.

### Test me
  
Go to the Story Builder, then click Capture Scene button.
Try adding images and specifying size in the dialog. Try resizing images with the provided handles.

### Checklist

-   [y] UI feature only.


staffordsmith83:
All working, except problem with image resize drag handles not visible. Apparently this is a problem with self hosting TinyMCE and some of the required css not loading. Issue here : https://github.com/tinymce/tinymce-react/issues/56

nf-s (changes requested):
Looking good - I love the new editor!

I think we should match the style of the old editor which doesn't have the rounded rectangle border, and has a solid border below the header.

image

Messing around with CSS in my broswer I was able to get this with the following CSS rules

image

```css

.tox:not(.tox-tinymce-inline) .tox-editor-header {
  border-bottom: 1px solid #ddd;
  box-shadow: none;
}

.tox-tinymce {
  border:none;
}
```

A few other small things:

- Should we remove the un-minified CSS files? Or do you think they should be left there for reference?
  - `content.css`
  - `content.inline.css`
  - `skin.css`
  - `skin.shadowdom.css`
- It would be nice if `StoryEditor.jsx` and `Editor.jsx` were TypeScript - but not absolutely necessary.
- I wouldn't worry about your comments about coupling, we can refactor later if needed
- Lazy loading works perfectly - thanks for including that!
- It might be nice to theme modals and stuff to match Terria UI - but I think that is another day's problem

image

nf-s (approved):

PR #6299, opened 2022-05-24
Add progress of 3DTiles data source loading to Progress Bar

### LINKED ISSUE #6130: Enable 3DTiles to affect the loading progress bar
It would be great if 3DTiles could be included in the loading progress bar along the top of the screen. Apparently it's as simple as:

> yeah should be able to just hook into this event: https://github.com/CesiumGS/cesium/blob/main/Source/Scene/Cesium3DTileset.js#L486

### LINKED ISSUE #6144: Progress bar visibility

staffordsmith83:
### What this PR does

Fixes #6130 

Adds an animated progress bar to show that 3DTileset sources are loading. As there is no way of knowing the progress of these sources, this is an indeterminate progress bar, and could also be used for other loading scenarios where progress is indeterminate.

Add a new `CesiumEvent()` called indeterminateTileLoadProgressEvent to Terria.ts:
```ts
  readonly tileLoadProgressEvent = new CesiumEvent();
  readonly indeterminateTileLoadProgressEvent = new CesiumEvent();
```

This new event just hold a boolean for whether 3DTiles datasources are loading.
We set this event by setting event listeners on 3dTilset sources:

For each added 3DTiles data source, we listen to `tileset.tileLoad` and 'tileset.allTilesLoaded' events. 

In `ProgressBar.jsx` a new indeterminate progress bar div has been added. This is only visible when `terria.indeterminateProgressEvent` emits true, and is set to visible=false when `terria.indeterminateProgressEvent` emits a false value.

In practise, this means that both progress bars can be visible at the same time in superimposition, which looks fine from a UI perspective and limits complex logic.

When 3DTileset sources are removed from the workbench, `terria.indeterminateTileLoadProgressEvent` emits false, so we are not left with any unresolved progress indicators. If there are other 3DTileset sources loading, these will quickly restart the progress bar so that use case is accounted for.

When 3DTileset sources are removed from the workbench the event listeners are also removed. To achieve this, a reference to the event listeners is saved in an array of objects at `Cesium._3dTilesetEventListeners`. 
  
### Test me
  
To test, please add a 3DTiles data source to your catalog and add the data.
For example, add the following to the `catalog` array in `TerriaMap/wwwroot/init/simple.json` in TerriaMap:

```json
    {
      "type": "3d-tiles",
      "name": "Fort Denison",
      "url": "https://nsw.digitaltwin.terria.io/api/v0/data/a99c8280-dbfe-475a-8a4e-3c2d2ca87513/3dtiles/tileset.json",
      "id": "adsfsadfds"
    },
```

Try adding the datasource, then zooming to it, and observe the progress bar at the top of the screen.

### Checklist


lib/Models/Cesium.ts
```diff
@@ -614,6 +632,23 @@ export default class Cesium extends GlobeOrMap {
...
+          const startingListener = tileset.tileLoad.addEventListener(() => {
+            this._updateTilesLoadingIndeterminate(true);
+          });
+
+          //Add event listener for when tiles finished loading for current view. Infrequent.
+          const finishedListener = tileset.allTilesLoaded.addEventListener(
+            () => {
+              this._updateTilesLoadingIndeterminate(false);
+            }
+          );
```
tephenavies: As discussed, remove these event listeners in `Cesium.destroy` and set indeterminate loading to finished

staffordsmith83: Implemented

lib/Models/Cesium.ts
```diff
@@ -1,6 +1,7 @@
 import i18next from "i18next";
 import { autorun, computed, runInAction } from "mobx";
 import { computedFn } from "mobx-utils";
+import { Event } from "terriajs-cesium";
```
tephenavies: This should be 
```suggestion
 import Event from "terriajs-cesium/Source/Core/Event ";
```

lib/Models/Cesium.ts
```diff
@@ -608,24 +617,65 @@ export default class Cesium extends GlobeOrMap {
...
       const primitives = this.scene.primitives;
 
       // Remove deleted primitives
-      prevPrimitives.forEach(primitive => {
+      prevPrimitives.forEach((primitive, i) => {
         if (!allPrimitives.includes(primitive)) {
           primitives.remove(primitive);
+          if (
+            isCesium3DTileset(primitive) &&
+            allCesium3DTilesets.indexOf(primitive) === -1
```
tephenavies: This condition is always true. If `allPrimitives.includes(primitive)` is false then `allCesium3DTilesets.indexOf(primitive) === -1` because `isCesium3DTileset` implies `isPrimitive`.

lib/Models/Cesium.ts
```diff
@@ -608,24 +617,65 @@ export default class Cesium extends GlobeOrMap {
...
+          ) {
+            try {
+              // Remove all event listeners from any Cesium3DTilesets by running stored remover functions
+              const fnArray = this._3dTilesetEventListeners[i].removeFns;
+              for (let j = 0; j < fnArray.length; j++) {
+                fnArray[j](); // run the function
+              }
+              this._3dTilesetEventListeners.splice(i, 1); // Remove the item for this tileset from our eventListener reference storage array
```
tephenavies: I think this has the problem of deleting while iterating, even though we're not iterating over this array. If we remove two 3d tilesets in one "tick", the second one may throw an error or remove the wrong event listeners.

Before you fix this, look at the suggestion below, which solves this and the reordering problem.

lib/Models/Cesium.ts
```diff
@@ -608,24 +617,65 @@ export default class Cesium extends GlobeOrMap {
...
+            const finishedListener = this._eventHelper.add(
+              primitive.allTilesLoaded,
+              () => {
+                this._updateTilesLoadingIndeterminate(false);
+              }
+            );
+
+            // Push new item to eventListener reference storage
+            try {
+              this._3dTilesetEventListeners.push({
```
tephenavies: It's not guaranteed that new 3d tilesets will be at the end of `allPrimitives`. I think to make this work you could replace `this._3dTilesetEventListeners` with a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), like `Map<Cesium3DTileset, EventListenerRemover>`. It's also make removing easier.

--------

#### Old ideas
You could change `allPrimitives.forEach` above to `this._3dTilesetEventListeners = allPrimitives.map` and return a new `EventListenerRemover` for a newly added dataset, and for an existing 3d tileset return the old `EventListenerRemover` from `this._3dTilesetEventListeners` (get the index from `prevPrimitives.indexOf(primitive)`. Recreating the array every time also deals with dropping old `EventListenerRemover` (and the above `prevPrimitives.forEach` would have already removed the event listeners).


Another option is to store the event handler remove functions on the 3d tilesets:
```ts
(primitive as any)._terriaProgressEventsRemoveFunctions = [start,finished];
```
But this is not as good as the first solution using Map.

tephenavies (changes requested):

staffordsmith83:
Some errors shown in the console when adding and removing 3DTilesets but these are in main and not related to this PR. Issue noted separately at https://github.com/TerriaJS/terriajs/issues/6321

lib/Models/Cesium.ts
staffordsmith83: Changes from @steve9164 have been incorporated. thanks for the useful review!

I used 
```
private _3dTilesetEventListeners = new Map<
    Cesium3DTileset,
    Event.RemoveCallback[]
  >(); // eventListener reference storage
```

No longer using custom Interface EventListenerRemover.
Event listener removers are accessed using the primitive as the key, instead of iterating.

Fixed the import of `Event`.

Some further simplifications of code for readability.

`3dTilesetEventListeners` is not recreated every time I dont think... would this be better? You mentioned in your change suggestions as an old idea.

Please note some errors unrelated to this PR, started a separate issue at https://github.com/TerriaJS/terriajs/issues/6321

lib/Models/Cesium.ts
```diff
@@ -608,13 +616,24 @@ export default class Cesium extends GlobeOrMap {
         }
       }
 
+      const allCesium3DTilesets = this._allMapItems.filter(isCesium3DTileset);
```
tephenavies:
```suggestion
```
Unused code.

tephenavies (approved):
This looks great. Sorry for suggesting the array method of tracking removers only to realise after you'd implemented it that it wouldn't work.

staffordsmith83:
> This looks great. Sorry for suggesting the array method of tracking removers only to realise after you'd implemented it that it wouldn't work.

No worries @steve9164 thanks for your careful review much appreciated

PR #6302, opened 2022-05-25
Fix docs build

tephenavies:
Docs have not built successfully since mid-March. This branch:
* Fixes out of memory build error (set NODE_OPTIONS)
* Points to userguide
* Upgrades to node 16 for building docs
* Fixes trying to access `document` while running TerriaJS in node
* Fixes a bug with doc generation to do with SensorObservationServiceCatalogItem

Also I've added doc generation success/failure messages to our Slack, so we won't be oblivious to doc build errors in future.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 
 #### next release (8.2.5)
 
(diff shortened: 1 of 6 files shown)


na9da (approved):
Looks good!

PR #6326, opened 2022-06-07
Fix Story Prompt

### LINKED ISSUE #6325: Story Prompt does not show when sharing a map that includes stories
image
This prompt should show when loading a map that includes stories.

staffordsmith83:
### What this PR does

Fixes #6325

When sharing a map with stories included, a 'stroy prompt' should show:
image

  
### Test me
  
Create a story in a map, and share it.
Open the share link.
The story prompt should show.

### Checklist

### DIFF
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -533,7 +533,7 @@ export default class Terria {
   @observable depthTestAgainstTerrainEnabled = false;
 
   @observable stories: StoryData[] = [];
-  @observable storyPromptShown: number = 0;
+  @observable storyPromptShown: number = 0; // Story Prompt modal will be rendered when this property changes. See StandardUserInterface, section regarding sui.notifications. Ideally move this to ViewState.
 
   /**
    * Gets or sets the ID of the catalog member that is currently being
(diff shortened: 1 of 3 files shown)


lib/Models/Terria.ts
```diff
@@ -533,6 +533,7 @@ export default class Terria {
   @observable depthTestAgainstTerrainEnabled = false;
 
   @observable stories: StoryData[] = [];
+  @observable storyPromptShown: number = 0;
```
na9da: Might be good to leave a comment here on the purpose of this property.

tephenavies: Definitely needs a comment. This sort of property should be on `ViewState` instead of `Terria`, but currently the origin of the scenes in `terria.stories` cannot be determined after it's been set.

staffordsmith83: Added comment for this line. Cannot move to ViewState without a bigger fix I think

tephenavies: Maybe sometime we'll fix story scene provenance (probably when we next extend stories, either with having a catalogue of stories for a map, or logging on to author official stories) and then we can get rid of this and detect loaded stories from outside Terria.ts.

tephenavies (approved):

PR #6334, opened 2022-06-09
Fix 3dtiles styling

na9da:
### What this PR does

Fixed default 3d-tiles styling to add a workaround for a Cesium bug which resulted in wrong translucency value for point clouds.

Caveat: this fix ignores the original opacity of the tileset feature and only uses the opacity value from the trait.
  
### Test me
  
- Load any point cloud (without this fix)
- Zoom out a bit, the whole tileset turns white
- Zoom in close and you can see that the features are translucent even when the opacity setting is full
- Now load the point cloud with this fix, both the above problems should be fixed

Before: http://ci.terria.io/main/#share=s-gkvNLmOhvGTK9oY9cSZcySA896F
After: http://ci.terria.io/fix-3dtiles-styling/#share=s-gkvNLmOhvGTK9oY9cSZcySA896F

### Checklist


lib/ModelMixins/Cesium3dTilesMixin.ts
```diff
@@ -393,17 +393,31 @@ function Cesium3dTilesMixin<T extends Constructor<Model<Cesium3dTilesTraits>>>(
...
+        // translucency values making the tileset translucent even when the
+        // computed opacity is 1.0. It also makes the whole of the point cloud
+        // appear white when zoomed out to some distance.  So for now, the only
+        // solution is to discard the opacity from the tileset and only use the
+        // value from the opacity trait.
         style.color =
-          "rgba((${COLOR}.r === undefined ? 1 : ${COLOR}.r) * 255, " +
-          "(${COLOR}.g === undefined ? 1 : ${COLOR}.g) * 255, " +
-          "(${COLOR}.b === undefined ? 1 : ${COLOR}.b) * 255, " +
-          "${COLOR}.a === undefined ? ${opacity} : ${COLOR}.a * ${opacity})";
```
na9da, line 406: Using any expression (eg `0.0 + 1.0`) in the alpha component triggers a Cesium bug, so we'll have to ignore the opacity from the feature and use only the opacity trait value for now.

staffordsmith83 (approved):
Looks good, tested and working.

PR #6338, opened 2022-06-11
feat: implement cesium and leaflet credits as react components

zoran995:
### What this PR does

Implement Cesium and Leafet credits as react components so they can properly utilise translation and it is easier to maintain them. This fixes part of #6336, the credits are not being updated on language change.

- [x] Show Terria logo in the 2d viewer
- [x] Show extra credits links in the 2d viewer
- [ ] Implement a modal as react component to show data attribution in both cesium and leaflet viewers so they are consistent. ~Is it better to leave this as a `Good first issue`?~ - better not, turns out this was a little ... to implement reactively. Done in #6340

### Test me

Try changing language in leaflet or cesium viewer, and observe that credits are being translated  

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [x] I've updated relevant documentation in `doc/`. - not relevant
-   [x] I've updated CHANGES.md with what I changed.
-   [x] I've provided instructions in the PR description on how to test this PR.

Issue #6336, opened 2022-06-10
Map credits not translated and rendered as translation key


zoran995:
I just noticed that credits at the map footer are rendered as a translation key instead of translation. Also changing of language won't update them.
![image](https://user-images.githubusercontent.com/19208948/173154265-70028270-0839-4504-ba57-9a3e0ebc2057.png)
Rendering as a translation key started happening after the merge of #6219, and the reason behind this is that the translation is not available when initializing cesium, and this is not in react context so it is not updated later.
https://github.com/TerriaJS/terriajs/blob/bd6198e66b9cf4314a95464acba8ab5d40516ef8/lib/Models/Cesium.ts#L270-L278
Not updating the value on a language change is an older issue.

We probably should go with this approach https://github.com/TerriaJS/terriajs/pull/6219#issuecomment-1084101382, as it sometimes happens to see a glitch with translation load, less than a half of a second; without any throttle, don't know what is different from a few months ago (sorry for bad image quality)
![image](https://user-images.githubusercontent.com/19208948/173154023-53430834-9e45-47c5-a25b-1530c0e06096.png)

Steps to reproduce

- Open http://ci.terria.io/main/
- Notice the credits being a translation keys
- Change the view to 2D and then back to 3D
- Now they are translated into the chosen language since the translation file is already loaded.

tephenavies:
Using a preload link doesn't seem to fix this. It only warms the cache so downloading translation.json is fast. I think we need to wait on the promise of `i18next.init({...})` before rendering to avoid no-language flash.

PR #6340, opened 2022-06-13
Data attributions modal

zoran995:
### What this PR does

Supersedes #6338

Implement Cesium and Leafet credits as react components so they can properly utilise translation and it is easier to maintain them. This fixes part of https://github.com/TerriaJS/terriajs/issues/6336, the credits are not being updated on language change.

image
image

- [x] Show Terria logo in the 2d viewer
- [x] Show extra credits links in the 2d viewer
- [x] Implement a modal as react component to show data attribution in both cesium and leaflet viewers so they are consistent.
  
### Test me
  
Try checking language in leaflet or cesium viewer, and observe that credits are being translated
Try clicking Basemap in credits, data attribution model should appear on screen showing current data credits

### Checklist

-   [x] I've updated relevant documentation in `doc/`. - not relevant


lib/Models/Cesium.ts
```diff
@@ -461,6 +419,83 @@ export default class Cesium extends GlobeOrMap {
...
+        const creditDisplayElements: { credit: Credit; count: number }[] =
+          creditDisplay._currentFrameCredits.lightboxCredits.values;
+
+        const credits = creditDisplayElements
+          .sort((credit1, credit2) => {
+            return credit2.count - credit1.count;
+          })
+          .map(({ credit }) => credit.html);
+
+        if (this.arrayEquals(credits, this.cesiumDataAttributions.toJS()))
```
tephenavies: This should use `import { isEqual } from "lodash-es";`

lib/Models/Cesium.ts
```diff
@@ -461,6 +419,83 @@ export default class Cesium extends GlobeOrMap {
...
+          return;
+        // first remove ones that are not on the map anymore
+        for (const attribution of this.cesiumDataAttributions.toJS()) {
+          if (!credits.includes(attribution)) {
+            this.cesiumDataAttributions.remove(attribution);
+          }
+        }
```
tephenavies: This only works in all cases due to the shallow copy inside `toJS()`. I think this would be improved by either iterating backwards (as usually done when iterating and removing items), or making the shallow copy more explicit using `this.cesiumDataAttributions.slice()`.

lib/Map/Cesium/CesiumSelectionIndicator.ts
```diff
@@ -17,6 +17,7 @@ import isDefined from "../../Core/isDefined";
 declare module "terriajs-cesium/Source/Scene/Scene" {
   export default interface Scene {
     tweens: TweenCollection;
+    frameState: any;
```
staffordsmith83: Can we define the type properly here or not possible?

lib/Models/Cesium.ts
```diff
@@ -461,6 +419,83 @@ export default class Cesium extends GlobeOrMap {
...
+
+        const credits = creditDisplayElements
+          .sort((credit1, credit2) => {
+            return credit2.count - credit1.count;
+          })
+          .map(({ credit }) => credit.html);
+
+        if (this.arrayEquals(credits, this.cesiumDataAttributions.toJS()))
+          return;
+        // first remove ones that are not on the map anymore
```
staffordsmith83, line 479: Perhaps follow this up with a few more comments

lib/Models/GlobeOrMap.ts
```diff
@@ -110,6 +110,10 @@ export default abstract class GlobeOrMap {
 
   abstract notifyRepaintRequired(): void;
 
+  @computed
```
staffordsmith83, line 116: A comment here might be useful?

lib/Models/LeafletAttribution.ts
```diff
@@ -0,0 +1,96 @@
+import { IObservableArray, runInAction, observable } from "mobx";
+import L from "leaflet";
+import Terria from "./Terria";
+
+export class LeafletAttribution extends L.Control.Attribution {
```
staffordsmith83, line 5: I like how this functionality is in its own class and not in `Leaflet.ts`. Do you think this could be done for the `updateCredits()` method that has been added to Cesium.ts too, just so the architecture is similar for Cesium and Leaflet modes? Im sure you have your reasons for doing it differently for Cesium and Leaflet, but what do you think?

lib/ReactViews/Credits/Credit.type.ts
```diff
@@ -0,0 +1,4 @@
+export interface ICredit {
+  url: string;
+  text: string;
+}
```
staffordsmith83, line 4: In terriajs looks like we dont have any other files named `****.type.ts` . How is it done in the rest of the repo, could you just export this interface and also `ICreditsProps` from `Credits.tsx`? Fine if you have your reasons just good to stick to the patterns used elsewhere in the project.

lib/ReactViews/Credits/Credits.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useTheme } from "styled-components";
+
+export const Spacer = () => {
+  const theme = useTheme();
+  return (
+    <span aria-hidden="true" css={{ color: theme.textLight }}>
+      |
+    </span>
+  );
+};
```
staffordsmith83: Do you think `Spacer` should be defined in `Credit.tsx` instead? I feel this pattern is a bit circular, where `Credits` imports `Credit`, and `Credit` imports `Spacer` which is defined in the `Credits.tsx` file. Once again, fine if you have your reasons.

lib/ReactViews/Credits/MapCredits/MapCredits.props.ts
```diff
@@ -0,0 +1,8 @@
+import { ICredit } from "../Credit.type";
+import GlobeOrMap from "../../../Models/GlobeOrMap";
+
+export interface IMapCreditsProps {
```
staffordsmith83: In terriajs we dont really use this pattern of defining interfaces for props in separate files named `{componentName}.props.ts`. Do you think this is better practise and we should adopt throughout the repo? If so you could make a ticket suggesting we follow this pattern? Otherwise we are defining the type of Props within the component file itself, or naming it `{componentName}Types.ts`. I think we should choose one way and stick with it, thoughts welcome.

lib/ReactViews/Credits/MapCredits/MapCredits.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { useTranslation } from "react-i18next";
+import { useDataAttributionContext } from "../../../ReactContexts/DataAttributionContext";
+import { Credits, Spacer } from "../Credits";
+import { CreditsContainer } from "../CreditsContainer";
+import { DataAttributionModal } from "../DataAttribution/DataAttributionModal";
+import { TerriaLogo } from "../TerriaLogo";
+import { MapCreditLogo } from "./MapCreditLogo";
+import { IMapCreditsProps } from "./MapCredits.props";
+
+export const MapCredits: FC<IMapCreditsProps> = observer(
```
staffordsmith83, line 20: Throughout this PR you define functional components as type `: FC<props>`. Small point, but for consistency, I think you should define the type as `React.FC<props>`.

Occurs in 6 or 7 components in this PR.

For example how its done elsewhere in the terriajs repo:

```
export const StoryButton: React.FC<StoryButtonProps> = props => {
```

lib/ReactViews/Credits/index.ts
```diff
@@ -0,0 +1 @@
+export { MapCredits } from "./MapCredits/MapCredits";
```
staffordsmith83, line 2: We dont use this pattern much in terriajs of having another index.ts to reexport something. Is it worth us adopting more widely? Is it to simplify imports?

lib/ReactContexts/DataAttributionContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const context = useContext<IDataAttributionContext>(
+    Context as ReactContext<IDataAttributionContext>
+  );
+  if (!context) {
+    throw new Error(
+      "useDataAttributionContext must be used within a DataAttributionContextProvider"
+    );
+  }
+  return context;
+};
```
staffordsmith83: I dont think we are using ReactContext anywhere else in Terriajs. Could this functionality (dataAttributionVisible, showDataAttribution, hideDataAttribution) put in one of the global state classes e.g. ViewState instead? Once again, let us know if there are benefits to adopting this pattern instead, i.e. avoiding prop drilling. I appreciate it works well but just for consistency sake we should use the same patterns throughout the repo as much as possible.

lib/ReactViews/SelectableDimensions/Color.tsx
```diff
@@ -90,7 +90,7 @@ export const SelectableDimensionColor: React.FC<{
         <div
           css={{
             position: "absolute",
-            zIndex: "2"
+            zIndex: 2
```
staffordsmith83, line 93: Good catch!

staffordsmith83 (changes requested):
Lots of great work here @zoran995 thanks so much! I have made a few comments, mostly regarding consistency with the existing code base. Open to adopting your patterns if you can justify why they are better, see individual suggestions.

Specifically, I note that this PR is intended to fix the attributions being translated on language change, but not the other problems in issue #6336 correct? 

Thanks once again for your awesome work!

lib/ReactViews/Credits/MapCredits/MapCredits.tsx
zoran995: `import React from "react";` should not be used in react 17 and later it will deprecated in future version of React. We still can't upgrade to React 17 or 18, see https://github.com/TerriaJS/terriajs/pull/5549/files#issuecomment-922506060 for details, but IMHO it would be good to start adopting new pattern

lib/ReactViews/Credits/MapCredits/MapCredits.props.ts
zoran995: perfectly fine for me, just got used to this on another project it is really easy to find props. In Terria most of the react files have props, updated styled components with custom styling needed for it, and then file starts to grow. But it would be probably too much change to go with this

I don't see this either `or naming it {componentName}Types.ts`

lib/ReactViews/Credits/Credits.tsx
zoran995: lets move it to its own file 😄

lib/ReactViews/Credits/Credit.type.ts
zoran995: I think that the `Checkbox/types.ts` and `Styled.types.ts` are the only files of this type (and guess who added them 🤣). It won't be bad to separate types from components when they are used in multiple places. In the end, this will be in traits in a future version of terria since it is used in config.

lib/ReactViews/Credits/index.ts
zoran995: Yes to simplify imports and to differentiate what is intended to be used in other places of the codebase (just like if we were using an external library for it). Basically, we know this should be used in other places and nothing else, if something else is needed then move it and use as a separate component. I think the Checkbox is the only component with `index.ts`

lib/ReactContexts/DataAttributionContext.tsx
zoran995: Yeah prop drilling and using react state for what it is meant to be used as there is no reason to make this observables and IMHO, the ViewState is toooooo big (the same as `Terria.ts`)
But I can put it in viewState, if that is team call. There was also a proposal to start using context but no one started implementing it so I don't know what is decision on it (i.e. provide terria and viewState through context for start)

lib/Map/Cesium/CesiumSelectionIndicator.ts
zoran995: Having type for entire frameState would be really hard to maintain (https://github.com/CesiumGS/cesium/blob/1.94/Source/Scene/FrameState.js) but I will add type for what is used in our code

lib/Models/Cesium.ts
zoran995: done

lib/Models/GlobeOrMap.ts
zoran995: done

lib/Models/LeafletAttribution.ts
zoran995: For cesium this is just a modification of the approach that was used for a long time in Terria. I am not sure that it can be done for cesium since most of the API is private, we just hide it and provide our implementation with the minimum usage of private API. Changing to a leaflet approach will require looking into private API and copying it which will make it harder to maintain in future.

staffordsmith83:
Thanks for all your responses Zoran! Ill get onto this PR review tomorrow and hopefully we can merge it, thanks again.

lib/Models/LeafletAttribution.ts
staffordsmith83: OK sounds like you have taken the best approach, thanks.

lib/ReactViews/Credits/Credit.type.ts
staffordsmith83: OK agreed

lib/ReactViews/Credits/MapCredits/MapCredits.tsx
staffordsmith83: Yes I agree with you, thanks for the warning.

lib/ReactContexts/DataAttributionContext.tsx
staffordsmith83: @steve9164 what are your thoughts on this point?

tephenavies: Could it be replaced with local `useState` in `MapCredits` component? I think that'd be the simplest. I don't think we'll need to change or access visibility of credits outside that component. Maybe if map credits gets added to the tour we'd need it?

If it can't be done with `useState` it should use observables on ViewState.

As for providing terria and viewState through context, I think we should. It'd add a `<TerriaGlobals.Provider>` or something to a lot of test cases, but all our components would look a lot nicer.

lib/ReactViews/Credits/index.ts
staffordsmith83: Ok, fine if written in typescript

lib/ReactContexts/DataAttributionContext.tsx
zoran995: @steve9164 great catch that it can be replaced with local state 🙌 🚀

lib/ReactViews/Credits/MapCredits/MapCredits.tsx
```diff
@@ -0,0 +1,71 @@
...
+  credits?: ICredit[];
+  currentViewer: GlobeOrMap;
+}
+
+export const MapCredits: FC<IMapCreditsProps> = observer(
+  ({ currentViewer, hideTerriaLogo, credits }) => {
+    const { t } = useTranslation();
+    const [dataAttributionVisible, setDataAttributionVisible] = useState(false);
+
+    const showDataAttribution = useCallback(() => {
```
staffordsmith83, line 25: I see you are the only contributor to use useCallback() perhaps we could make more use of it to make things more efficient?

staffordsmith83 (approved):
Great, thanks for your hard work Zoran, approved!

PR #6342, opened 2022-06-15
Fixes for plugin init.

### LINKED ISSUE #6337: Viewer mode not persisted
Setting viewer mode in the init file will force the viewer mode which will override the value persisted in local storage. This is a change from behaviour implemented in #4947 and I don't see documented in changes.md. 

On first look, it seems that the issue is introduced with plugins in TerriaMap (https://github.com/TerriaJS/TerriaMap/pull/582)
https://github.com/TerriaJS/TerriaMap/blame/75f2489f3bb50096a59caeabe4ce8ca00d02e20b/index.js#L78-L88
this changed the order of initialization and persisted map settings are loaded before init data

na9da:
### What this PR does

Fixes #6337 

Also see - https://github.com/TerriaJS/TerriaMap/pull/597

* Added `beforeRestoreAppState` hook for call to `Terria.start()` which gets called before state is restored from share data.
* Made `order` optional for `ICompositeBarItem`.
  
### Test me

Check https://github.com/TerriaJS/TerriaMap/pull/597

### Checklist


mwu2018 (approved):
I've verified the changes via https://github.com/TerriaJS/TerriaMap/pull/597.

PR #6343, opened 2022-06-15
`ChartPanelDownloadButton` will now only export columns which are visible in chart

### LINKED ISSUE #6324: v7 to v8 migration - CSV download issue - investigate
Difference in CSV download from point drill between v7 and v8
When the user selects a point (or polygon?) drill from the analysis tools this generates (a call to GSKY) and Terria plots the timeseries of the four variables: PV, NPV, Bare and Rain. If the user unselects a subset of variables these are removed from the timeseries plot in Terria. If the user then selects the Download CSV button:
•	In v7 the CSV includes only the selected variables, excluding the unselected variables.
•	In v8 the CSV includes all the variables, and does not respect the unselected variables.
The v7 behaviour is important for users as the Rain variable has more time steps than for the FC variables, which makes use of the CSV file (in v8) unnecessarily complicated particularly if Rain was deselected for this reason.

Steps to reproduce: 

**v7**
- go to https://map.geo-rapp.org/
- go to Add data; select Analysis tools - Mean for point or region (monthly) 

image

- select a polygon and run the analysis
- the analysis will result in a chart with 4 variables 
- deselect a nr of variables (2-3) and then click Download 

image
- once downloaded. check the CSV; it will only have the remaining variables 

image

**v8**
- go to https://geo-rapp.prod.saas.terria.io/
- take the same steps as v7 
- the downloaded CSV will include all variables, including the un-selected ones

image

nf-s:
### `ChartPanelDownloadButton` will now only export columns which are visible in chart

Fixes https://github.com/terriajs/terriajs/issues/6324

* TSXify `ChartPanelDownloadButton`
* `ChartPanelDownloadButton` will now only export columns which are visible in chart
* Cleanup `Mixin` and `Traits` inheritance
* Wrap the following components in `observer` - `ChartItem`, `LineChart`, (chart) `Legends`, `ChartPanelDownloadButton`

  
### Test me
  
- http://ci.terria.io/fix-chart-download/#configUrl=https://terria-catalogs-public.storage.googleapis.com/geo-rapp/map-config/prod.json
- go to Add data; select Analysis tools - Mean for point or region (monthly)
- select a polygon and run the analysis
- the analysis will result in a chart with 4 variables
- deselect a nr of variables (2-3) and then click Download

### Checklist


lib/ModelMixins/TableMixin.ts
```diff
@@ -742,8 +742,10 @@ function TableMixin<T extends Constructor<Model<TableTraits>>>(Base: T) {
       return this.tableColumns.find(column => column.type === type);
     }
 
-    findColumnByName(name: string): TableColumn | undefined {
-      return this.tableColumns.find(column => column.name === name);
+    findColumnByName(name: string | undefined): TableColumn | undefined {
+      return isDefined(name)
+        ? this.tableColumns.find(column => column.name === name)
+        : undefined;
```
na9da, line 741: nitpick: It might be better to filter out undefined `name` at the point where it gets called rather than here?

nf-s: I prefer allowing `undefined` argument, as most calls to `findColumnByName` use traits with optional values. It makes using optional chaining possible - otherwise you need to use if statement or ternary expressions before calling `findColumnByName`

For example

image

na9da (approved):
@nf-s , works well in my testing.

Charts are messy and good effort cleaning up some of it :)

PR #6345, opened 2022-06-15
Save sharelink helper

staffordsmith83:
### What this PR does

Fixes [#568](https://github.com/TerriaJS/nsw-digital-twin/issues/568)

Adds a button `How to Bookmark Sharelinks` to the SharePanel that will open the HelpPanel and an item with `itemName === 'bookmarkHelp'`

Button is only rendered if this item exists in the help items for the current map.
Help item has just been added to NSW Digital Twin prod environment for testing:
https://github.com/TerriaJS/nsw-digital-twin-catalog/pull/19

As part of this PR I have changed the styling of the SharePanel as specified by @philipgrimmett 

image.

The trickiest part is that when opening the help panel via the button on SharePanel, we want the SharePanel to stay open underneath the HelpPanel so that when a user has read what to do and closes the HelpPanel, they will be taken back to the SharePanel. To implement this a new boolean property was added to `BaseOuterPanel` called `disableCloseOnFocusLoss` and this is passed as a prop down the chain.

A new property was added to ViewState called `retainSharePanel`. When we open a help item from the SharePanel, we set `retainSharePanel = true` with an @action.

Closing the HelpPanel, the SharePanel is still visible, but we should reset `retainSharePanel = false` so that the panel closes when anything else is opened (e.g. the StoryEditor or any other popup). Currently this is implemented on ViewState:
```js
hideHelpPanel() {
    this.showHelpMenu = false;
    setTimeout(() => {
      this.setRetainSharePanel(false);
    }, 500); // TODO: this is a bit hacky, but need to re-enable closing of share panel when loses focus.
  }
```
This feels a bit hacky and I am open to suggestions if its needs improvement.

  
### Test me

I have created a config.json gist with the new help item to test here:

http://ci.terria.io/save-sharelink-helper/#configUrl=https://gist.githubusercontent.com/staffordsmith83/d2f61102e9d2751f46022b9a1faa934c/raw/5bdebf7217e4ecf534f9f237050d39e92ff942a9/config.json
  
Click Share / Print , and verify that How to Bookmark Sharelinks button is present.
Click the button and see that you are taken to the correct help item.
Close the HelpPanel and verify that the SharePanel is still open and behaves correctly.

Also, test with [How to Bookmark Sharelinks](http://ci.terria.io/save-sharelink-helper/) (with no bookmarkHelp item in the config.json) and verify that the button is no longer rendered in SharePanel.

The Help Item has also been added to NSW Digital Twin Catalog:
image

Go to https://dev.nsw.digitaltwin.terria.io/ and open help to verify existence.

### Checklist

-   ~~[ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~~
-   ~~[ ] I've updated relevant documentation in `doc/`.~~


staffordsmith83:
Should I refactor this https://github.com/TerriaJS/terriajs/blob/7d2418f83c2c43cda1b7fa6196f1dee9f5a40477/lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx#L411-L417 into an action?

lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
```diff
@@ -186,3 +187,15 @@
...
+  text-decoration: underline;
+  padding-top: 0.8em;
+  cursor: help;
+}
+
+hr.thin-line-divider {
+  border: 0;
+  height: 0;
+  border-top: 0.5px solid rgba(255, 255, 255, 0.15);
+}
```
na9da, line 201: Normally I would say, we should refrain from writing new `scss` code, in this case though it might make sense to revisit it a later point and rewrite it using `tsx` and `styled-components`.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -397,8 +398,37 @@ const SharePanel = observer(
...
+                medium
+                textLight
+                isLink
+                onClick={evt => {
+                  evt.preventDefault();
+                  evt.stopPropagation();
+                  viewState.setRetainSharePanel(true);
+                  viewState.showHelpPanel();
+                  viewState.selectHelpMenuItem("bookmarkHelp");
+                }}
```
na9da: Makes sense to extract it out into new function.

lib/ReactViewModels/ViewState.ts
```diff
@@ -311,7 +311,8 @@ export default class ViewState {
    * Gets or sets a value indicating whether the catalog's modal share panel
    * is currently visible.
    */
-  @observable shareModalIsVisible: boolean = false;
+  @observable shareModalIsVisible: boolean = false; // Small share modal inside StoryEditor
+  @observable retainSharePanel: boolean = false;
```
na9da: Might be good to add a line of comment documenting the observable.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
```diff
@@ -397,8 +398,37 @@ const SharePanel = observer(
                 )
               }
             />
+            {/* Following code block dependent on existence of "bookmarkHelp" Help Menu Item */}
+            <If
+              condition={this.props.terria.configParameters.helpContent.some(
+                e => e.itemName === "bookmarkHelp"
+              )}
+            >
```
na9da: We use less of `If` nowadays because it is not available for `tsx`. The standard react way of doing this is `{condition && <Text>...</Text>}`

lib/ReactViews/Map/Panels/InnerPanel.jsx
```diff
@@ -16,6 +16,8 @@ const InnerPanel = createReactClass({
      * panel to close.
      */
     doNotCloseFlag: PropTypes.string,
+    /** Disable closing on loss of focus and only allow with close button */
+    disableCloseOnFocusLoss: PropTypes.boolean,
```
na9da: I think this should be `PropTypes.bool` instead. It might be causing this warning in the console:

```
Warning: Failed prop type: InnerPanel: prop type `disableCloseOnFocusLoss` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.This often happens because of typos such as `PropTypes.function` instead of `PropTypes.func`.
```

na9da (changes requested):
Hey @staffordsmith83, great work and great effort navigating the hacky state dependencies :) 

I have left a few suggestions. I also found a scenario where it is possible to end up like this:
image

eg: `Open share panel > click bookmark help link > *help panel opens* > click anywhere on the workbench > *help panel disappears* > open map settings`

I think it happens because the `<HelpPanel>` can close itself without calling `hideHelpPanel` when it is no longer the top-element:

https://github.com/TerriaJS/terriajs/blob/63c8facb2970e4cd6a9bf3ceb1634b0ec48d14c0/lib/ReactViews/StandardUserInterface/StandardUserInterface.tsx#L401

I think one solution for that problem might be to try moving the following code to [componentWillUnmount](https://reactjs.org/docs/react-component.html#componentwillunmount) hook of `HelpPanel`, which will ensure that it is always run when the help panel is disposed.

https://github.com/TerriaJS/terriajs/blob/63c8facb2970e4cd6a9bf3ceb1634b0ec48d14c0/lib/ReactViewModels/ViewState.ts#L660-L662

lib/ReactViewModels/ViewState.ts
staffordsmith83: Done

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
staffordsmith83: Resolved

lib/ReactViews/Map/Panels/InnerPanel.jsx
staffordsmith83: Resolved, yes that error no longer occurs with your change thanks!

lib/ReactViews/Map/Panels/SharePanel/SharePanel.jsx
staffordsmith83: Resolved, extracted out to an action on ViewState

lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
staffordsmith83: OK Im getting my head around styled components Ill raise a ticket to do this in next sprint

staffordsmith83:
@na9da Ive implemented your changes, Ill make another ticket to update with styled-components in next sprint.

na9da (approved):
Looks great @staffordsmith83!

PR #6362, opened 2022-06-27
Limit workbench item title to 2 lines

### LINKED ISSUE #6341: Improve workbench title padding
http://ci.terria.io/main/#share=s-14QuWsALwB1WNovomZAz2RINYB6

image

If the title is super long, it overlaps onto other items

image

zoran995:
### What this PR does

Fixes #6341

Limit the number of lines in workbench item title to 2, and set `min-height` instead of `height` so padding is correct

image

  
### Test me
  
### Checklist

-  ~[ ] I've provided instructions in the PR description on how to test this PR.~


nf-s:
Thanks for the PR @zoran995 !
I fixed this using a different approach in https://github.com/TerriaJS/terriajs/pull/6357
image

But I think limiting 2 lines might be better. What do you think @philipgrimmett ?

nf-s:
Hi @zoran995 

We think it is best to have the full Workbench item title visible (which is in `main` now from https://github.com/TerriaJS/terriajs/pull/6357) - but think that it would be nice if the title is limited to 2 lines when the item is collapsed

What do you think?

zoran995:
Yeah that is better than always limiting the title, I also added a title attribute which will be useful when the item is collapsed so that the user can get an entire title in a tooltip
image
image

nf-s (approved):
Thanks @zoran995 !

PR #6365, opened 2022-06-29
Extend recognised file types to include ifc

KeyboardSounds:
**Note that terriajs DOES NOT support IFC without a closed source plugin that isn't publicly available at this time.**

### What this PR does

When the terriajs-ifc plugin is present, users will be able to add data to a map from local storage and web sources.
  
### Test me
  
1. Try uploading any terria-compatible file to the map. It should work.
2. Try upload an IFC file. It should fail gracefully.

### Checklist

-   [X] ~I've updated relevant documentation in `doc/`.~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Change Log
 * Wrap the following components in `observer` - `ChartItem`, `LineChart`, (chart) `Legends`, `ChartPanelDownloadButton`
 * Improve TerriaReference error logging
 * Fix handling GeoJSON if features have null geometry
+* Allow IFC files to be added to a map from local or web data (Requires non-open source plugin) 
 * [The next improvement]
 
 #### 8.2.6 - 2022-06-17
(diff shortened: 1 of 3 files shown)


zoran995:
hi @KeyboardSounds, can we expose getDataType in a different way so it can be extended from plugins i.e. dataType object on Terria and plugins can then extend it. Don't know if there is time to do it now, or leave it as a separate issue.

KeyboardSounds:
Ahaha that's the exact reason this PR is still in draft. We do need to do exactly that, because my current solution is bad, but alas, you are also right about time constraints. I'll merge this and make a new issue.

mwu2018 (approved):

KeyboardSounds:
> hi @KeyboardSounds, can we expose getDataType in a different way so it can be extended from plugins i.e. dataType object on Terria and plugins can then extend it. Don't know if there is time to do it now, or leave it as a separate issue.

@zoran995 here's the new issue: https://github.com/TerriaJS/terriajs/issues/6367

PR #6372, opened 2022-06-30
Use undefined default value and add changelog entry

tephenavies:
Fix a couple of default values and add a changelog entry for https://github.com/TerriaJS/terriajs/pull/5706.


staffordsmith83 (approved):
Thanks for improving this @steve9164

PR #6375, opened 2022-06-30
Await Internationalisation initialisation in `Terria.start`

### LINKED ISSUE #6336: Map credits not translated and rendered as translation key

nf-s:
### Await Internationalisation initialisation in `Terria.start`

Fixes #6336 #6352

image

Also - `UserDrawing.messageHeader` can now also be `() => string`.  
I've done this instead of use `applyTranslationIfExists` as it follows how `onMakeDialogMessage` works
  
### Test me
  
- http://ci.terria.io/main/
- http://ci.terria.io/international-await/
- Measure Tool

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 
 #### next release (8.2.8)
 
+* Await Internationalisation initialisation in `Terria.start`
 * [The next improvement]
 
 #### 8.2.7 - 2022-06-30
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Change Log
 #### next release (8.2.8)
 
 * Await Internationalisation initialisation in `Terria.start`
+* `UserDrawing.messageHeader` can now also be `() => string`
 * [The next improvement]
 
 #### 8.2.7 - 2022-06-30
--- a/lib/Models/Internationalization.ts
+++ b/lib/Models/Internationalization.ts
@@ -62,7 +62,7 @@ class Internationalization {
      */
     i18StartOptions: I18nStartOptions | undefined,
     terriajsResourcesBaseUrl: string
-  ): void {
+  ) {
     const languageConfig = Object.assign(
       defaultLanguageConfiguration,
       languageConfiguration
@@ -77,7 +77,7 @@ class Internationalization {
      * @param {Array} languageConfiguration.changeLanguageOnStartWhen
      */
 
-    i18next
+    return i18next
       .use(HttpApi)
       .use(LanguageDetector)
       .use(initReactI18next)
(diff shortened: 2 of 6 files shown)


staffordsmith83 (commented):
I managed to get a situation where the measure tool text is in French but the language set to English:
image

And vice versa:
image


It appears that `currentInteractionMode.message()`on `MapInteractionWindowWrapper` is not reactive to language changes

staffordsmith83 (changes requested):
Need to make `currentInteractionMode.message()` on `MapInteractionWindowWrapper` reactive to language changes

staffordsmith83 (approved):
Looks good, I was thinking to only initialise the ui when the `i18next.init `promise resolved by using `.then()` but wasnt sure what the sequence of events are in initialising terria. 

I notice we dont need to use `require` anymore, that was something to do with Cesium types? I know you must be across this.

One point, even though we have removed most MobX stuff from `MeasureTool` we still get `[mobx] Computed value UserDrawing@1089.svgPoint is being read outside a reactive context. Doing a full recompute` . This is also present in main so its not like something new introduced with this PR.

Have tested the measure tool quite a bit and looks good to go!

nf-s:
> I notice we dont need to use `require` anymore, that was something to do with Cesium types? I know you must be across this.

I prefer to use `import` - but as the cesium types are incomplete, sometimes we need to add model declarations. 

For example https://github.com/TerriaJS/terriajs/blob/6a3cce9eaf710a021783eb05e2eb3b8685ac2182/lib/ThirdParty/terriajs-cesium-extra/index.d.ts#L43

> One point, even though we have removed most MobX stuff from `MeasureTool` we still get `[mobx] Computed value UserDrawing@1089.svgPoint is being read outside a reactive context. Doing a full recompute` . This is also present in main so its not like something new introduced with this PR.

Good find!
I removed the `@computed` from `get svgPoint` - https://github.com/TerriaJS/terriajs/pull/6375/commits/fdb240c816348108cb3201fc9c4494b9d14f84a4

PR #6395, opened 2022-07-06
Add `initFragmentPaths` support for hostnames different to `configUrl`/`applicationUrl`

nf-s:
### For example

**In Config JSON** https://gist.githubusercontent.com/nf-s/167f4749a4911288c54641589f417ecd/raw/55fd4f27c28dfe4636e119d84b2b1503117624a7/geo-rapp-fragment-path.json

```json
"parameters": {
    "initFragmentPaths": [
          "https://terria-catalogs-public.storage.googleapis.com/geo-rapp/dev/"
      ],
}
```
  
**Go to URL** with `australia` init fragment in hash

- http://ci.terria.io/init-fragment-hostname/#configUrl=https://gist.githubusercontent.com/nf-s/167f4749a4911288c54641589f417ecd/raw/55fd4f27c28dfe4636e119d84b2b1503117624a7/geo-rapp-fragment-path.json&australia

**After PR**

It will load

- https://terria-catalogs-public.storage.googleapis.com/geo-rapp/dev/australia.json

**Before PR** 

`australia` would be resolved to `applicationUrl`

-  https://ci.terria.io/init-fragment-hostname/geo-rapp/dev/australia.json

Which seems weird

### Checklist


test/Models/TerriaSpec.ts
```diff
@@ -141,6 +142,45 @@ describe("Terria", function() {
...
+
+      if (!isInitFromOptions(initSource))
+        throw "Init source is not from options";
+
+      expect(
+        initSource.options.map(source =>
+          isInitFromUrl(source) ? source.initUrl : ""
+        )
+      ).toEqual([
+        "path/to/config/path/to/init/something.json",
```
mwu2018, line 181: This looks odd to me. Shouldn't it be "path/to/init/something.json"? Need to document how a path is constructed.

mwu2018: I think, the initFragmentPath should be relative to baseUrl specified by configUrl. Then by setting
```
initFragmentPaths: [
              "../init/",
              "https://hostname.com/some/other/path/"
            ]
```
the path will be constructed as "path/to/init/something.json".

test/Models/TerriaSpec.ts
```diff
@@ -332,6 +374,57 @@ describe("Terria", function() {
...
+
+      if (!isInitFromOptions(initSource))
+        throw "Init source is not from options";
+
+      expect(
+        initSource.options.map(source =>
+          isInitFromUrl(source) ? source.initUrl : ""
+        )
+      ).toEqual([
+        "https://application.url/path/to/init/someInitHash.json",
```
mwu2018, line 425: Need to explain why it is not "https://application.url/path/to/config/path/to/init/someInitHash.json" when initFragmentPaths[0] is set to "path/to/init/", which is different from line 179.  The reason is that the path "path/to/init" is relative to the path "/" specified by applicationUrl: "https://application.url/". This is because terria.updateApplicationUrl has been called.

nf-s: I have added comments to how paths are constructed

I also added more docs

- https://github.com/TerriaJS/terriajs/blob/31b8384535768d958b2d653486bcaed5b7a7dc85/doc/deploying/controlling-with-url-parameters.md#catalog-files-init-files
- https://github.com/TerriaJS/terriajs/blob/31b8384535768d958b2d653486bcaed5b7a7dc85/doc/customizing/client-side-config.md#intializationurls

test/Models/TerriaSpec.ts
nf-s: I have added comments to how paths are constructed

mwu2018 (approved):

PR #6396, opened 2022-07-06
Apply all `InitSources` sequentially instead of asynchronously

### LINKED ISSUE #5949: Overwrite elments present in workbench
In v.8.1.4, differently from what was able to do in v.8.1.0, it is no more possible to overwrite the element present in the `workbench` into a `.json` file.

As a practical example, I have some layers that are uploaded by the file `terria.json`, some of them are shown at the beginning to the user putting their `id` into the  `workbench`.
Then, I have more advanced layers for advanced analysis that could be enabled by the user clicking on a button that links to the same website adding `#advance_fun`. This adds the layers presented in the file `advance_fun.json` where the new layers are described and the `workbench` list the id of some of them.

Now, when the user reloads the page asking for the extra layers, the layers visible still remain the ones listed in the `workbench` of the file `terria.json`, and not anymore thelayers for the advance analysis.

nf-s:
### Apply all `InitSources` sequentially instead of asynchronously

Fixes #5949

`InitSources` should be applied on top of each other - in order specified by `terria.initSources` array.

I have added an `init-sources` doc - https://github.com/TerriaJS/terriajs/blob/abac123141f60541fb5b383e60355c2f26ec7a5b/doc/contributing/init-sources.md

For example - these should be applied in order
- `config.initializationUrls`
  - these are added in `terria.start() -> terria.setupInitializationUrls()`
- `updateApplicatonUrl`
  - `interpretHash`
    - init fragments - eg `#simple` or `#https://some.url/config.json`
    - `#start`
    - `#share`
  - `/catalog/` routes
  - `/story/` routes

When `InitSources` are applied (see `terria.applyInitData()`) there are two async tasks
- `loadModelStratum`
- Loading models and adding them to the workbench

Depending on these tasks, `InitSources` may be applied out of order, at random times - and may get interlaced

This PR makes sure `InitSources` are applied in order

### Test me

The following URL has 
- `configUrl`
  - https://gist.githubusercontent.com/nf-s/167f4749a4911288c54641589f417ecd/raw/55fd4f27c28dfe4636e119d84b2b1503117624a7/geo-rapp-fragment-path.json
- This add the following `initializationUrl`
  - https://terria-catalogs-public.storage.googleapis.com/geo-rapp/dev.json
    - Which set `workbench: ["tN2nrg", "LfaYaM", "idvaI5"]`
- Init fragment https://terria-catalogs-public.storage.googleapis.com/geo-rapp/dev/australia.json
  - Which **should** override  `workbench: ["nrm-regions-2017"]`
 
**Before** - http://ci.terria.io/main/#configUrl=https://gist.githubusercontent.com/nf-s/167f4749a4911288c54641589f417ecd/raw/55fd4f27c28dfe4636e119d84b2b1503117624a7/geo-rapp-fragment-path.json&https://terria-catalogs-public.storage.googleapis.com/geo-rapp/dev/australia.json

**After** - http://ci.terria.io/sync-init-source/#configUrl=https://gist.githubusercontent.com/nf-s/167f4749a4911288c54641589f417ecd/raw/55fd4f27c28dfe4636e119d84b2b1503117624a7/geo-rapp-fragment-path.json&https://terria-catalogs-public.storage.googleapis.com/geo-rapp/dev/australia.json

### Checklist


lib/Models/Terria.ts
```diff
@@ -1239,29 +1242,28 @@ export default class Terria {
...
-              severity: initSource?.errorSeverity,
-              message: {
-                key: "models.terria.loadingInitSourceError2Message",
-                parameters: {
-                  loadSource: initSource!.name ?? "Unknown source"
-                }
+    // Sequentially apply all InitSources
+    for (let i = 0; i < loadedInitSources.length; i++) {
+      const initSource = loadedInitSources[i];
+      if (!isDefined(initSource?.data)) return;
```
tephenavies:
```suggestion
      if (!isDefined(initSource?.data)) continue;
```

test/Models/TerriaSpec.ts
```diff
@@ -33,6 +33,7 @@ import ViewState from "../../lib/ReactViewModels/ViewState";
 import { buildShareLink } from "../../lib/ReactViews/Map/Panels/SharePanel/BuildShareLink";
 import SimpleCatalogItem from "../Helpers/SimpleCatalogItem";
 import { defaultBaseMaps } from "./../../lib/Models/BaseMaps/defaultBaseMaps";
+import { exp } from "protomaps";
```
tephenavies:
```suggestion
```

tephenavies (changes requested):

tephenavies (approved):

PR #6401, opened 2022-07-08
Add arrow key navigation to StoryPanel

### LINKED ISSUE #6387: Keyboard interactivity for stories
Please can we add keyboard controls to stories:
- key up or left goes to previous
- key down or right goes to next

staffordsmith83:
### What this PR does

Fixes #6387

A simple PR to add more logic to the "keydown" event listener in Story Panel.

Variable is renamed from `escKeyListener` to `keydownListener`.
Now handles arrow keys for navigation too.
  
### Test me
  
Create a story with multiple scenes. 
Play Story.
Jump forwards or backwards with arrow keys:
Behaviour should be:
- key up or left goes to previous
- key down or right goes to next

### Checklist

~-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~
~-   [ ] I've updated relevant documentation in `doc/`.~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ Change Log
 
 * Pin `html-to-react` to `1.4.5` due to ESM module in dependency (`parse5`) breaking webpack
 * Add step to `"Deploy TerriaMap"` action to save `yarn.lock` after `sync-dependencies` (for debug purposes)
+* Implement ability to navigate between scenes in StoryPanel using keyboard arrows
 * [The next improvement]
 
 #### 8.2.8 - 2022-07-04
--- a/lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
+++ b/lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
@@ -87,7 +87,7 @@ interface State {
 
 @observer
 class StoryPanel extends React.Component<Props, State> {
-  escKeyListener: EventListener | undefined;
+  keydownListener: EventListener | undefined;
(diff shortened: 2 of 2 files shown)


nf-s (approved):
Looks good!

One possible change - but i'm not sure if this is the best behavior - is to limit key presses when at the start or end of a story. For example

- If you press up or left at the first scene - then nothing happens
- if you press down or right at the last scene - then nothing happens

staffordsmith83:
> Looks good!
> 
> One possible change - but i'm not sure if this is the best behavior - is to limit key presses when at the start or end of a story. For example
> 
> * If you press up or left at the first scene - then nothing happens
> * if you press down or right at the last scene - then nothing happens

Yeah I wondered about that too, thoughts @meh9 ?

meh9:
@staffordsmith83 I've always been of the opinion that Stories revolve, so up/left at start moves you to the end, and right/down at the end move you to the start. However, trumping this is that we should do whatever happens when you click prev/next links with your mouse.

staffordsmith83:
Ah so there is more logic with the prev/next mouse clicks. I can replicate this if you prefer, or change the behaviour of the prev/next buttons to revolve mode too! Preference @meh9 ?

meh9:
Yes I think there is more logic from memory so we should try to replicate it.

I am uncomfortable with overriding every single decision made by others on the spur of the moment because they have probably thought about this harder than I have, so let's keep current functionality for now. :)

staffordsmith83:
OK will do!

PR #6402, opened 2022-07-08
Fix support for `initUrls` in `startData`

### LINKED ISSUE #6403: Share intializationUrls are broken
### Discussed in https://github.com/TerriaJS/terriajs/discussions/6398

<div type='discussions-op-text'>

<sup>Originally posted by **glughi** July  7, 2022</sup>
Hello,
there is a way to share (and load with `#start`) the JSON catalog used?
Thanks in advance,
G.
</div>

nf-s:
### Fix support for `initUrls` in `startData`

Fixes https://github.com/TerriaJS/terriajs/issues/6403

You are supposed to be able to do this in `#start=`

For example

```json
{
  "version": "8.0.0",
  "initSources": [
    "http://gist.githubusercontent.com/nf-s/67e44309245f3caca1cee5cdf2eee3eb/raw/75c50871fd972f8d6f35039c6e5278f3f7437a69/colorscale-minmax.json"
  ]
}
```
  
But it isn't working

### Test me
  
- Before http://ci.terria.io/main/#clean&start={"version":"8.0.0","initSources":["http://gist.githubusercontent.com/nf-s/67e44309245f3caca1cee5cdf2eee3eb/raw/75c50871fd972f8d6f35039c6e5278f3f7437a69/colorscale-minmax.json"]}
- After http://ci.terria.io/fix-initurl-startdata/#clean&start={"version":"8.0.0","initSources":["http://gist.githubusercontent.com/nf-s/67e44309245f3caca1cee5cdf2eee3eb/raw/75c50871fd972f8d6f35039c6e5278f3f7437a69/colorscale-minmax.json"]}

### Checklist


doc/customizing/initialization-files.md
```diff
@@ -49,25 +51,25 @@ Key points:
 
 Most of the other properties of each layer depend on the specific type.  See the links above for details of each type.
 
-### Using a catalog file
+### Using an init file
 
-There are four ways to load a catalog file into a TerriaJS application:
+There are four ways to load a init file into a TerriaJS application:
```
staffordsmith83: 'an init file'

doc/deploying/controlling-with-url-parameters.md
```diff
@@ -20,7 +21,7 @@ Parameter      | Meaning
 `mode=preview`   | Operate in "preview mode", which mostly means not showing a warning if the screen is `small`  
 `share=`...      | Load a map view previously saved using the "Share" function with URL shortening.
 `start=`...      | Load a map view previously saved without URL shortening. The argument is a URL-encoded JSON structure defined using an internal format described below.
-`<initfile>`     | Load catalog file as described below.
+`<initfile>`     | Load [init/catalog file](../customizing/initialization-files.md) as described below.
```
staffordsmith83: Maybe need to reiterate this bit here for the people who skim read:
NB: You do not need to specify <initfile> in the url, as any unrecognised parameter (eg `foo`) is treated as the name of a catalog file

doc/deploying/controlling-with-url-parameters.md
```diff
@@ -48,34 +53,12 @@ The `start=` parameter essentially embeds an entire catalog file in the URL. The
             ]
         },
         {
-            "initialCamera": {
```
staffordsmith83, line 57: Can you not set initialCamera any more? Just homeCamera which is less specific?

nf-s: You can still set `initialCamera` - I just thought it is cluttering up the doc. Full properties supported is listed in `ShareData` link above

https://github.com/TerriaJS/terriajs/blob/d4dbbadf34054b512d98d5608c62b30c56294f6b/doc/deploying/controlling-with-url-parameters.md#L39

staffordsmith83: Sounds good.

lib/Models/Terria.ts
```diff
@@ -2072,8 +2072,10 @@ async function interpretStartData(
   errorSeverity?: TerriaErrorSeverity,
   showConversionWarning = true
 ) {
-  const containsStory = (initSource: InitSourceData) =>
-    Array.isArray(initSource.stories) && initSource.stories.length;
+  const containsStory = (initSource: InitSourceData | string) =>
```
staffordsmith83: Why does this need to accept strings too but then we check if its json? Just for my learning. Is this the best place to have this union type or could we define InitSourceData as a union with string? Whats happening here?

lib/Models/Terria.ts
```diff
@@ -2102,22 +2104,32 @@ async function interpretStartData(
...
-                name,
-                data: isJsonObject(initSource, false) ? initSource : {},
-                errorSeverity
-              };
-            })
+            ...startDataV8!.initSources.map(initSource =>
+              isJsonString(initSource)
+                ? {
+                    name,
+                    initUrl: initSource,
```
staffordsmith83, line 2168: Whats the difference here, why are we storing an `initUrl` when its a json string, and a `data` when its not? So I can see we store an initUrl when you use the start url param to pass json object with `version` and `initSources` properties . When do you store a data value instead?

Ah no worries I can see in test suite, you can pass settings directly: `initSources: ["http://something/init.json", { splitPosition: 0.3 }]`

staffordsmith83 (approved):
Looks good to me man, a couple of conflicts need resolving thats all, sorry for leaving it so long.
Also, could you give me a quick answer to a couple of questions I put on the code just so I see how things work better. Cheers.

lib/Models/Terria.ts
nf-s: I have removed the function as it was a bit weird.

It is checking init sources from here

https://github.com/TerriaJS/terriajs/blob/bef8fb2ecea8a3f14ca7be28c1e0da0c1e302a12/lib/Models/InitSource.ts#L43-L47

which supports both string and `InitSourceData`. No other usage of `InitSourceData` has a union with `string` like this

I have added some comments to make it more clear

https://github.com/TerriaJS/terriajs/blob/8a94ac1002bfe29024092be3ccd4c0885c29581a/lib/Models/Terria.ts#L2180-L2191

lib/Models/Terria.ts
nf-s: I have added some comments to make it more clear

https://github.com/TerriaJS/terriajs/blob/ced86ee39229bfc1f867d72adbea5c0f78e4d66c/lib/Models/Terria.ts#L2157-L2179

The `InitSource` stuff is a bit weird - and most of the property names are coming from our JSON config - so we can't really change them

Issue #5420, opened 2021-04-12
Use `import type ".."` syntax for type only imports


na9da:
In some places we inadvertently import cesium modules like `Scene` although it is only used for its type by TS. Webpack sometimes fails to infer this and includes the whole module in the bundle. We can avoid this by using the TS `import type Scene from "..."` syntax. By turning on the lint rule [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint)

zoran995:
@na9da Just to be sure, when importing types and components/functions from the same module, is it better to separate that into 2 imports, i.e.
```
import type { ChartPoint } from "./ChartData";
import { ChartData } from "./ChartData";
```

na9da:
Hi @zoran995, I think you have to separate them out as 2 different imports like you have shown, otherwise TS would given an error. 

I created this issue as one of the things we could explore to reduce the load time as observed here #5422. I haven't really got the chance to test it out fully. We might end up using this only for few imports where we want to optimize the bundle size.

zoran995:
This will require an upgrade of prettier to at least `2.23`, otherwise, it will throw 
```
SyntaxError: '=' expected.
```

PR #6408, opened 2022-07-09
Prettier v2

zoran995:
### What this PR does

Enables fix of #5420

This is pretty big change for terriajs codebase but version 2 of prettier came out 2 and half years and it would be pretty good to upgrade to it. It also allows us to use import just type and some other modern syntax that was erroring on old versions.

If decided to merge it should be done the same way as the last time, by creating two new tags pre-prettier-v2 and post-prettier-v2

### Checklist

-   ~[ ] I've provided instructions in the PR description on how to test this PR.~


buildprocess/generateCatalogIndex.ts
```diff
@@ -136,13 +136,15 @@ export default async function generateCatalogIndex(
...
+            const result = await (
+              member as ReferenceMixin.Instance
+            ).loadReference();
             result.logError(`FAILED to load Reference ${name} (${path})`);
             result.pushErrorTo(
               errors,
               `FAILED to load Reference ${name} (${path})`
             );
-            result.catchError(e => console.error(e.toError().message));
+            result.catchError((e) => console.error(e.toError().message));
```
nf-s, line 147: Is it possible to configure prettier to keep single argument anon functions without brackets?

nf-s (changes requested):
Thanks @zoran995 !
Just one comment - otherwise looks good!

buildprocess/generateCatalogIndex.ts
zoran995: Yes setting `"arrowParens": "avoid"` in `.prettierrc` will do that, I personally prefer to always add parens as it is easier to add type annotation, another param, consistency and bunch of other options, I also believe most of the developers are used to this new syntax. but in the end prettier handles them so perfectly fine with both

https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev

nf-s: Yep fair enough, I think you're right.

tephenavies:
We should extend this to also apply prettier to more files in the repo (e.g. gulpfile.js, everything in buildprocess, markdown files). Maybe even change our prettierignore file to ignore certain files rather than only include some directories.

zoran995:
@steve9164 I have updated prettierignore to include more files

tephenavies:
That's better. I think we should also format:
* `*.yml` files
* `architecture`
* plenty of things in `wwwroot/` (probably most things that aren't `.gitignore`d, third party code or generated files that are never partially modified)
* `.github`

I was thinking of changing `.prettierignore` to almost match `.gitignore`, plus add a few extra lines.

`.prettierignore`:
```.gitignore

# =====================
# From .gitignore. Update this list when .gitignore changes

docs/
node_modules/
.idea/
build/
error.log
output.log
npm-debug.log
.DS_Store
terria.lib.js
terriajs-server.log
terriajs.pid
/wwwroot/doc/
/wwwroot/schema/
/wwwroot/user-guide/
/coverage/
package-lock.json
*.swp
dist
.history/

# Catalog index generation files
catalog-index*.json

# End of .gitignore copy
# =====================

# These are auto-generated by css-modules-typescript-loader
**/*.scss.d.ts

# Generated files
/wwwroot/data/regionids

/wwwroot/third_party

# Reignore some files
doc/acknowledgements/attributions.md

```

And change the package.json scripts to  `"prettier --write ."` and `"prettier --check ."`. That'll circumvent the problem where prettier is trying to format files it doesn't recognise when the ignore file is reversed.

tephenavies:
Merging on 22nd August.

PR #6409, opened 2022-07-09
upgrade husky to v8

zoran995:
### What this PR does

husky in v7 changed the way it works internally and it should provide more performant, also the v4 version wasn't running always on my machine most of the times it just skipped it like it wasn't there

### Test me
  
- checkout the branch
- run yarn install
- try some dummy commit that should trigger prettier run and observe that pretty-quick fixed that file

### Checklist


nf-s (changes requested):
Thanks @zoran995 !

I'm getting `The '.husky/pre-commit' hook was ignored because it's not set as executable` 

is there an easy way to set cross-platform file permissions automatically?

zoran995:
According to the husky docs this should be handled automatically https://typicode.github.io/husky/#/?id=hooks-not-running

Only things that occur to me is that since the pre-commit file is coming from windows it don't have those set up correctly (but if that is case than it is really weird behavior) 
Can you try just to see if above can be source of the problem
```sh
rm .husky/pre-commit
npx husky add .husky/pre-commit "npx pretty-quick --staged"
```

Also one important note for husky v8 is that it requires node 14 or later

zoran995:
@nf-s this should work now (tested on mac)
the fix was to update git index with chmod
```sh
git update-index --chmod=+x .husky/pre-commit
```

force push was just a committer name change so CLAassistant doesn't complain

nf-s (approved):
Perfect, thanks!

PR #6414, opened 2022-07-11
Strongly type `item` in WorkbenchItem and remove `show` toggle for non `Mappable` items

nf-s:
### Test me

- http://ci.terria.io/workbench-type-item/
- "Explore map data"
- Add "Magda Item" to the workbench
- Notice that item doesn't have the "show" toggle box while it is loading
  - This is because the `MagdaReference` isn't mappable - until it has loaded the reference
- **Note** WMS item will throw errors

### Checklist


lib/ReactViews/Workbench/WorkbenchItem.tsx
```diff
@@ -52,11 +57,17 @@ class WorkbenchItemRaw extends React.Component<IProps> {
...
+      this.props.item.setTrait(
+        CommonStrata.user,
+        "show",
+        !this.props.item.show
+      );
   }
 
   @computed
   get isOpen(): boolean {
+    if (!CatalogMemberMixin.isMixedInto(this.props.item)) return true;
```
mwu2018, line 73: Why is this test needed? It seems unnecessary.

mwu2018:
A referenced item can be anything. E.g. it can be a group:
```
    {
      "type": "magda",
      "url": "https://nsw.digitaltwin.terria.io",
      "name": "Magda Item 1",
      "recordId": "gp-SydneyWest_CESIUM_3DTiles_CENTRE1"
    },
```
Adding it to the workbench by clicking on "+" will cause error. Do you think we need to handle this in this PR?

lib/ReactViews/Workbench/WorkbenchItem.tsx
mwu2018: Ok, it is type checking for the next line. But why returning `true`?

nf-s: Type checking for next line.
If item doesn't have `CatalogMemberMixin` then there is no way of toggling if the item is open in the workbench or not - so I think it should be open by default

nf-s:
> A referenced item can be anything. E.g. it can be a group:
> 
>     {
>       "type": "magda",
>       "url": "https://nsw.digitaltwin.terria.io",
>       "name": "Magda Item 1",
>       "recordId": "gp-SydneyWest_CESIUM_3DTiles_CENTRE1"
>     },
> 
> Adding it to the workbench by clicking on "+" will cause error. Do you think we need to handle this in this PR?

I don't think this is relevant to this PR - the purpose of this PR is to add stricter types to `WorkbenchItem.tsx`

If a `Reference` is capable of being an item or a group (like `MagdaReference`) - you should add `isGroup: true`

```json
{
      "type": "magda",
      "url": "https://nsw.digitaltwin.terria.io",
      "name": "Magda Item 1",
      "recordId": "gp-SydneyWest_CESIUM_3DTiles_CENTRE1",
      "isGroup": true
}
```

lib/ReactViews/Workbench/WorkbenchItem.tsx
nf-s: I have added comments in the file

mwu2018: If an item does not have `CatalogMemberMixing`, it will never be added to the workbench. Therefore it does not matter if it is true or false.

mwu2018 (approved):

lib/ReactViews/Workbench/WorkbenchItem.tsx
nf-s: You're probably right - but `workbench.items` has type `BaseModel[]`

PR #6418, opened 2022-07-12
Fix table regions short report and showInChartPanel

nf-s:
### Fix table regions short report and showInChartPanel

* Fix "Regions: xxx" short report showing for non region mapped items
* Fix `showInChartPanel` default for mappable items (which is showing "moments" chart by default)

Before vs After

image

### Test

- **Before** http://ci.terria.io/main/#share=s-2yHZCIquhOOEFUHTNIQLTRYnlJz
- **After** http://ci.terria.io/table-regions-chart-fix/#share=s-2yHZCIquhOOEFUHTNIQLTRYnlJz

### Checklist


staffordsmith83 (approved):
Great, awesome that you add unit tests even for this, approved

PR #6425, opened 2022-07-14
Check if at start or end of stories before key nav

### LINKED ISSUE #6423: Modify arrow key navigation behaviour in Story Panel
Relates to PR: Add arrow key navigation to StoryPanel #6401
Related to issue: [#6387](https://github.com/TerriaJS/terriajs/issues/6387)

Modify the behaviour of the key navigation to be the same as that offered by the next/prev clickable buttons. When at end of story, right/down key should not return to start and vice versa.

staffordsmith83:
Change useCapture param on removeEventListener to make sure properly removed### What this PR does

Fixes #6423

Modify the behaviour of arrow key navigation for StoryPanel, so that cannot loop back to start, or go from start to end.
Also, needed to change `window.removeEventListener("keydown", this.keydownListener, false); to `window.removeEventListener("keydown", this.keydownListener, true)` so that event listeners properly removed when component unmounted.
  
### Test me
  
Test with this [link](http://ci.terria.io/modify-story-key-nav/#share=s-qbwqXlS25XBimInqef7cheXsITK)

Play story, navigate though scenes. Observe no looping.
Close StoryPanel, make sure arrow keys do not continue to navigate between scene locations.

### Checklist

~~-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~~
~~-   [ ] I've updated relevant documentation in `doc/`.~~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@ Change Log
   * Supports over 40 formats - including Collada, obj, Blender, DXF - [full list](https://github.com/assimp/assimp/blob/master/doc/Fileformats.md)
 * Add `description` to `getDataType` - this will be displayed between Step 1 and Step 2
 * Add warning message to `GltfMixin` when showing in 2D mode (Leaflet)
+* Prevent looping when navigating between scenes in StoryPanel using keyboard arrows
+* Fix bug where StoryPanel keyboard navigation persists after closing StoryPanel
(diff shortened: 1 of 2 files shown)


na9da (approved):
Looks good to me @staffordsmith83!

PR #6427, opened 2022-07-15
Move rendering of ShareUrl  to functional component

### LINKED ISSUE #6420: Extract share url logic out of SharePanel into separate functional component
Also fix styling to be in line with @philipgrimmett suggestions:
image

Also investigate implementing this idea from @zoran995 , moving includeStoryInShare into the share URL component:
"And also, I don't see a need for includeStoryInShare in viewState, it can be used as a sharePanel internal state; it even can go into the share URL component and setter exposed through ref using useImperativeHandle"

staffordsmith83:
Should we also move shareUrl property out of SharePanel?

### What this PR does

Extract the conditionally formatted rendering of ShareUrl Clipboard to a separate functional component.

At the moment the functionality for shortening the url etc is still in SharePanel. Should we also move `SharePanel.shareUrl` to `ShareUrlClipboard.shareUrl`?
Thoughts @zoran995 ?

### Test me
  
### Checklist

### DIFF
--- a/lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
+++ b/lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
@@ -5,9 +5,7 @@ import { observer } from "mobx-react";
 import React, { ChangeEvent } from "react";
 import { Trans, WithTranslation, withTranslation } from "react-i18next";
 import defined from "terriajs-cesium/Source/Core/defined";
-import Clipboard from "../../../Clipboard";
 import Icon from "../../../../Styled/Icon";
-import Loader from "../../../Loader";
 import Input from "../../../../Styled/Input";
 import { TextSpan } from "../../../../Styled/Text";
 import DropdownStyles from "../panel.scss";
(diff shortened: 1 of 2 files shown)


zoran995:
Hi @staffordsmith83 This is a good starting point for `ShareUrl` component that is fully encapsulated
Basically, all logic related to building share URL and shortening should go away from SharePanel to ShareUrl, the only logic that should stay in `SharePanel` is `includeStory` and `shortenWithService` which should be passed through props (even those can be in `ShareURL` and then exposed). Going this way `ShareUrl` has all logic encapsulated and later it is easy to reuse it wherever needed without `SharePanel`.

I would leave `shareUrl` in `SharePanel` and pass it along with the update function to `ShareUrl` and then useEffect that tracks those `includeStory` and `shortenWithService` and does the build share URL logic. `shareUrl` could probably go into `ShareUrl` component and exposed with `useImperativeHandle` but it could be tricky to force rerender.
```
useImperativeHandle(
  forwardRef,
  () => ({
    url: shareUrl
  }),
  [forwardRef, shareUrl]
);
```

I was free to play with it last night https://github.com/TerriaJS/terriajs/compare/main...zoran995:terriajs-i18n:share-url-fc in #6428
when working with functional components everything that even seems like a simple component it actually can and most probably should be (I know that it can be odd at the start to have components used in only one place). But it simplifies the understanding of what happens in each component and what it does, i.e. `ShareUrl` is building URL, rendering clipboard, warning, children and that's it, cognitive load smallest possible (this can go further with a custom hook for URL build). And lastly, this really simplifies writing tests, you have only a few cases to test for each component (at first looks like many of them but not much than what should be written for existing components)

staffordsmith83:
Ah awesome thanks for your contribution @zoran995! Ill take a look at your draft PR, great. I think that approach makes sense, passing `shareUrl` and the updater down in props. Actually, you dont want to put `shareUrl` in `viewState` instead? :-)

zoran995:
> Actually, you dont want to put shareUrl in viewState instead? :-)

Are we using it anywhere outside `SharePanel`? I would really like to expose it using imperativeHandle from `ShareUrl`, and that passing it around and rendering on update works without any issues. Maybe I just didn't play enough around it or it is not meant to do that, the last commit in #6428 changed from imperative handle to state

Passing state and updater requires a component which uses `ShareUrl` to have `shareUrl` as state

staffordsmith83:
OK that sounds good. No I was joking about `viewState` I know you think its getting way too big! Not used anywhere outside `SharePanel`.

zoran995:
> Maybe I just didn't play enough around it or it is not meant to do that

Actually I didn't know how standard refs work in this segment, check https://github.com/TerriaJS/terriajs/pull/6428#issuecomment-1185957816 for details

PR #6428, opened 2022-07-15
Share url fc

### LINKED ISSUE #6420: Extract share url logic out of SharePanel into separate functional component

### LINKED ISSUE #6433: Text contrast broken
The contrast for text in some features is currently not working i.e. text is illegible (e.g white text over cream background)  

image

To test: 
 - go to https://nationalmap.gov.au/
 - select and add any dataset or group and click Share from the About this data page 
 - check the text in the window showing the Share url

zoran995:
### What this PR does

Fixes #6420
Fixes #6433

Rewrite of share panel content to use functional component. Share URL now fully encapsulate logic around creating shareUrl so it can be reused if needed without the rest of the share panel
There are parts of code not needed anymore in sharePanel or not working correctly, see comment below.

This is how it looks now
image
image

Catalog share
image

Story share
image

  
### Test me
  
### Checklist

~-   [ ] I've updated relevant documentation in `doc/`.~


lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
```diff
@@ -51,521 +33,104 @@ interface PropTypes extends WithTranslation {
...
-        this._unsubscribeFromPrintMediaChange = function() {
-          matcher.removeEventListener("change", handlePrintMediaChange);
-        };
-      }
-
-      this._oldPrint = window.print;
-      window.print = () => {
-        this.print();
-      };
-    }
```
zoran995, line 116: I don't see anything changed after removing this, do we need this anymore? Also, this was run for `storyShare` and `catalogShare` and I don't see the point in doing that. It's easy to re-add it just in a different place

zoran995:
To get notified about `ref.current` changes we need to callback refs instead of standard refs (https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node). `useCallbackRef` is just a wrapper around that which allows providing multiple child values to parent without lifting the entire state

staffordsmith83 (commented):
Hi Zoran looks awesome, a couple of points:
To me looks a little bit like there is a higher degree of modularisation compared to the rest of the way things are done in Terria. As is splitting things out into separate files more than we usually do. For example: 

EmbedSection - has two files, should just be merged into one I think. is this to prepare for any use case where we need to use `EmbedInput` separately? 

`AdvancedOptions` - could be called `AdvancedOptionsSection` to fit with other names? This component includes:
- `IncludeStoryCheckbox`
- `ShortenWithServiceCheckbox` - I like that this shows in local mode now but deactivated
	- Do these really need to be separate components? Could we include in `AdvancedOptions`?
	- I feel we should only move functionality to separate files when it is being used in some other component, or will be soon.

Another thing, SharePanel has
	SharePanelContent (only used for Print Embed mode)
	ShareUrl (used in all three. The two simple modes (Catalog Share and Story Share) just use this and a title wrapped in a box). I dont really like that SharePanelContent has ShareUrl inside it but the others just render ShareUrl directly. I think that SharePanelContent could be put back inside SharePanel. Is it more efficient your way? Just to be clearest for maintainability.

But I do see how it simplifies SharePanel itself greatly... And that its a functional component using hooks instead of a class which is probably better...

staffordsmith83:
I have made a new branch https://github.com/TerriaJS/terriajs/tree/extract-sharepanel-modified with some changes but just reduced the modularisation a bit.

lib/ReactViews/Map/Panels/SharePanel/SharePanelContent.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { PrintSection } from "./Print/PrintSection";
+import { useCallbackRef } from "../../../useCallbackRef";
+
+interface ISharePanelContentProps {
+  terria: Terria;
+  viewState: ViewState;
+  closePanel: () => void;
+}
+
+export const SharePanelContent: FC<ISharePanelContentProps> = ({
```
staffordsmith83, line 24: I think its confusing to have this separate to SharePanel. Especially as the content for StoryShare and CatalogShare are rendered directly in SharePanel and dont have separate components. Im sure there is a good reason though, just not the clearest structure for maintainability going forward

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
staffordsmith83: This part is before my time, I dont see it being used anywhere either and probably safe to remove

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
```diff
@@ -50,522 +32,89 @@ interface PropTypes extends WithTranslation {
...
+          viewState={viewState}
+          closePanel={this.closePanel}
+        />
+      );
     }
-    this.changeOpenState(true);
-    this.renderContentForStoryShare();
   }
 
   render() {
```
staffordsmith83, line 120: I can see logically whats happening, but SharePanel is still a bit confusing with three different paths in `renderContent()` and only two different paths in `render()`.

I kind of think we should focus it more explicitly around the three modes `catalogShare`, `storyShare`, and `printEmbedShare` so that it is clearly structured for maintainability moving forward... What do you think?

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
```diff
@@ -50,522 +32,89 @@ interface PropTypes extends WithTranslation {
...
-  }
-
-  openWithUserClick() {
-    if (this.props.onUserClick) {
-      this.props.onUserClick();
+        </Box>
+      );
+    } else {
+      return (
+        <SharePanelContent
```
staffordsmith83, line 111: While this reduces the lines of code in the SharePanel class, it is a bit confusing. If we think about SharePanel having three modes `catalogShare`, `storyShare`, and `printEmbedShare`, it is confusing for someone looking at this class that `primtEmbedShare` (the default one) is created in such a different way to the other two. My thinking was to just have `SharePanel` and `ShareUrl`, and you could move a lot of functionality out of `SharePanel` into `ShareUrl`. But having so much logic in `SharePanelContent` is confusing structurally. Can you refactor so that the code in `SharePanelContent` is contained within `SharePanel` and `ShareUrl`?

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrlWarning.tsx
```diff
@@ -0,0 +1,85 @@
...
+  color: #552800;
+`;
+
+const WarningLink = styled.a`
+  color: #552800;
+  text-decoration: underline;
+  cursor: pointer;
+`;
+
+export const ShareUrlWarning: FC<IShareUrlWarningProps> = observer(
```
staffordsmith83, line 38: Good to be separate component, nice

lib/ReactViews/Map/Panels/SharePanel/share-panel.scss
```diff
@@ -57,127 +57,3 @@
     color: $color-primary;
   }
 }
-
```
staffordsmith83, line 60: Great thanks for getting rid of all this scss

lib/ReactViews/Map/Panels/SharePanel/AdvancedOptions/AdvancedOptions.tsx
```diff
@@ -0,0 +1,93 @@
...
+        ) : (
+          <AdvanceOptionsIcon glyph={GLYPHS.closed} />
+        )}
+      </RawButton>
+
+      {advancedOptions && (
+        <>
+          <StyledHr />
+          <Box column>
+            <IncludeStoryCheckbox
```
staffordsmith83: I think you should just put code for IncludeStoryCheckbox back inside AdvancedOptions

lib/ReactViews/Map/Panels/SharePanel/AdvancedOptions/AdvancedOptions.tsx
```diff
@@ -0,0 +1,93 @@
...
+              includeStory={includeStoryInShare}
+              onChange={includeStoryInShareOnChange}
+            ></IncludeStoryCheckbox>
+            <Spacing bottom={1} />
+            <ShortenWithServiceCheckbox
```
staffordsmith83: I think you should just put code for ShortenWithServiceCheckbox back inside AdvancedOptions

lib/ReactViews/Map/Panels/SharePanel/AdvancedOptions/Embed/EmbedInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useTranslation } from "react-i18next";
+
+import Input from "../../../../../../Styled/Input";
+import { IShareUrlRef } from "../../ShareUrl";
+
+interface IEmbedInputProps {
+  shareUrl: IShareUrlRef | null;
+}
+
+export const EmbedInput: FC<IEmbedInputProps> = ({ shareUrl }) => {
```
staffordsmith83: Put back inside EmbedSection to reduce modularisation a bit

lib/ReactViews/Map/Panels/SharePanel/Print/PrintSection.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Box from "../../../../../Styled/Box";
+import { TextSpan } from "../../../../../Styled/Text";
+import Button from "../../../../../Styled/Button";
+import { downloadImg } from "./PrintView";
+
+interface IPrintSectionProps {
+  viewState: ViewState;
+}
+
+export const PrintSection: FC<IPrintSectionProps> = ({ viewState }) => {
```
staffordsmith83, line 17: Yeah makes sense to have this as separate component, nice

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrlBookmark.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import Text from "../../../../../Styled/Text";
+
+interface IShareUrlBookmarkProps {
+  viewState: ViewState;
+}
+
+const bookMarkHelpItemName = "bookmarkHelp";
+
+export const ShareUrlBookmark: FC<IShareUrlBookmarkProps> = ({ viewState }) => {
```
staffordsmith83, line 14: Would we ever use this component outside of `SharePanel` (or `SharePanelContent` as it is currently)? I initially has this as a separate component too but may unnecessarily complicate the structure.

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrlInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Input from "../../../../../Styled/Input";
+import { useTranslation } from "react-i18next";
+
+interface IStoryShareInput {
+  theme?: "light" | "dark";
+  shareUrl: string;
+  placeholder?: string;
+}
+
+export const ShareUrlInput: FC<IStoryShareInput> = ({
```
staffordsmith83: This could almost certainly be added into `ShareUrlClipboard`. Would we ever use `ShareUrlInput` in another component do you think?

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrlClipboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface IShareUrlClipboardProps {
+  terria: Terria;
+  theme: "dark" | "light";
+  inputTheme?: "dark" | "light";
+  rounded: boolean;
+  shareUrl: string;
+  placeholder?: string;
+}
+
+export const ShareUrlClipboard: FC<IShareUrlClipboardProps> = ({
```
staffordsmith83: Would we ever use this component outside `ShareUrl` component? Could it go back inside there to reduce modularisation a little bit?

staffordsmith83 (changes requested):
Awesome work here man thankyou very much, please feel free to dispute any suggestions with justification, thanks again.

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrlClipboard.tsx
zoran995: as I worked on this I thought I would need to have a separate clipboard and input field for different cases of share URL usage, but it proved otherwise, but I left it separated

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrlBookmark.tsx
zoran995: I would leave it like this to separate logic that gets the bookmark help button from help items from the share panel, someone who forks the repo then can just remove the usage of bookmark but functionality is there or decide to add something else (fewer conflicts on future merge his change don't depend on something else i.e. we changed how help items work and it is no more item name but itemId)

staffordsmith83: Yeah that makes sense, thanks @zoran995

zoran995:
your changes regarding modularisation look fine to me, I think you can commit them directly to this branch so we don't pollute the PR list. 
I recently changed to doing things this way as I found it is easier to quickly scan through code and figure things around, don't have to read what i.e. checkbox does as it is already in the component name, and easier to test. Of course most of the times it goes out of control, but don't think much about it as navigation between files is really easy in modern tools (vscode now works in browser from github)

SharePanelContent is there to move logic to the functional component as it is easier to work with them, haven't tried to do move the entire SharePanel as it will be changed further. Neither story nor catalog share has to be `SharePanel` instances anymore, they can have their own panels that render only ShareUrl and it will make it pretty straightforward to work and maintain. After that with the new menu implementation just delete `SharePanel` and use content directly :D

staffordsmith83:
> your changes regarding modularisation look fine to me, I think you can commit them directly to this branch so we don't pollute the PR list. I recently changed to doing things this way as I found it is easier to quickly scan through code and figure things around, don't have to read what i.e. checkbox does as it is already in the component name, and easier to test. Of course most of the times it goes out of control, but don't think much about it as navigation between files is really easy in modern tools (vscode now works in browser from github)
> 

Yep this makes sense, just personal preference and I can see the advantages in your approach! But I think maybe we should keep it a bit less modular, to be in line with the rest of the repo.

> SharePanelContent is there to move logic to the functional component as it is easier to work with them, haven't tried to do move the entire SharePanel as it will be changed further. Neither story nor catalog share has to be `SharePanel` instances anymore, they can have their own panels that render only ShareUrl and it will make it pretty straightforward to work and maintain. After that with the new menu implementation just delete `SharePanel` and use content directly :D

Ah, I see your motivation. OK, let me see what @steve9164 thinks about this, but I like that you are moving towards the new panel implementation. Thanks again for your hardwork, one more commit coming to reduce modularisation a bit...

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrl.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+    const buildUnshortenderUrl = (
+      terria: Terria,
+      viewState: ViewState,
+      includeStories: boolean
+    ) => {
+      return buildShareLink(terria, viewState, {
+        includeStories
+      });
+    };
```
tephenavies: I think this function should be removed, and `buildShareLink` should be called directly

tephenavies (changes requested):
Everything else looks good. Just 1 change needed.

lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrl.tsx
staffordsmith83: Good catch, implemented.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
staffordsmith83: All good for @steve9164 so all good for me!

staffordsmith83 (approved):
All resolved

tephenavies (approved):

PR #6460, opened 2022-07-26
Fix terrain load order in Cesium viewer.

na9da:
### What this PR does

1. Fixes terrain load order so that when a `CesiumTerrainCatalogItem` is added to the workbench it gets used. Previously the terrain setting from `config` used to override the terrain added to workbench.

   Now the load order priority is:
   ```
   a. 3d-smooth when `useTerrain` is `false`
   b. Terrain added to workbench
   c. Terrain specified in `config.cesiumTerrainAssetId`
   d. Terrain specified in `config.cesiumTerrainUrl`
   e. Cesium world terrain when `config.useCesiumIonTerrain` is true
   f. Fallback to 3d-smooth (EllipsoidTerrainProvider) 
   ```

2. Setting `show` to false will correctly disable the terrain item added to workbench

3. Disable `Ideal Zoom` for terrain item

4. Workbench terrain item will show a status message `Terrain status: In use` when it is in use. When switching to `3d-smooth` you can observe it change to `Terrain status: Not in use` 

5. Improves `CesiumTerrainCatalogItem` loading so that an error is thrown when there is a network issue.

### Test me
  
1. Load this before link - [before link](http://ci.terria.io/main/#configUrl=https%3A%2F%2Fgist.githubusercontent.com%2Fna9da%2Fcd3f2cda07d0ac40004704c280de00b9%2Fraw%2Fb5bb30e74e2c7309249557543cd85510c97ae1c7%2Fterrain-test.json&share=s-t7ozfQ3Igdu6Rurx1Lnv9dq8TjM).
2. Observe how the terrains added to workbench has no effect. You can check this by opening the Network tab and searching for "terrain" and then checking the URL. It should be something like `https://assets.cesium.com/1/*` where the `1` is the `ionAssetId` set in this [config.json](https://gist.github.com/na9da/cd3f2cda07d0ac40004704c280de00b9#file-terrain-test-json-L14)
3. Now load the link for this PR - [after link](http://ci.terria.io/fix-terrain-load-order/#configUrl=https%3A%2F%2Fgist.githubusercontent.com%2Fna9da%2Fcd3f2cda07d0ac40004704c280de00b9%2Fraw%2Fb5bb30e74e2c7309249557543cd85510c97ae1c7%2Fterrain-test.json&share=s-aodgF8D7R6puctYH5nSXhm3y5Sm).
4. Observe that the terrain added to workbench gets loaded and applied to the map (you can verify this by searching for the asset id `3957` in the network tab).
5. Switch to 3d-smooth mode from the map settings. The workbench terrain should update its status to `Not in use`.

### Checklist


lib/Models/Catalog/CatalogItems/CesiumTerrainCatalogItem.ts
```diff
@@ -12,32 +12,89 @@ export default class CesiumTerrainCatalogItem extends UrlMixin(
...
+      terrainProvider.readyPromise
+    ])
+      .catch(() => false)
+      .finally(() =>
+        terrainProvider.errorEvent.removeEventListener(networkErrorListener)
+      );
+
+    return isReady
+      ? terrainProvider
+      : Promise.reject("Failed to load terrain provider");
```
tephenavies: Please wrap this string in a `new Error` or `TerriaError`.

lib/Models/Catalog/CatalogItems/CesiumTerrainCatalogItem.ts
```diff
@@ -12,32 +12,89 @@ export default class CesiumTerrainCatalogItem extends UrlMixin(
...
+    return `Terrain status: ${status}`;
+  }
+
+  /**
+   * Returns a Promise to load the terrain provider
+   */
+  @action
+  private async loadTerrainProvider(): Promise<
+    CesiumTerrainProvider | undefined
+  > {
```
tephenavies: Please remove the action decorator so that property access is tracked.

tephenavies (changes requested):
Looks good. Just a couple of changes needed

lib/Models/Catalog/CatalogItems/CesiumTerrainCatalogItem.ts
na9da: Good catch! I always forget that `forceLoadMapItems` runs inside a `computed`!

na9da:
@steve9164 - Thanks for the feedback. I resolved your change requests and made one additional change to `shortReport` so that it doesn't override a trait setting.

tephenavies (approved):

PR #6465, opened 2022-07-29
Set JulianDate.toIso8601 second precision to nanosecond

nf-s:
### Set JulianDate.toIso8601 second precision to nanosecond

When animating timeline we can get dates with way too precise seconds

image

Then when trying to pass these date strings through `JulianDate.fromIso8601` we get 

image

This PR changes precision to nanoseconds

### Checklist

-   [ ] ~I've provided instructions in the PR description on how to test this PR.~ Use seals data from vic-dt


na9da (approved):
Looks good. That was :zap: fast

PR #6472, opened 2022-08-02
Fix a style bug caused by not recalculating if mobile interface should be used on resize

tephenavies:
### What this PR does

Fix a bug introduced in translating StandardUserInterface.jsx to tsx.
  
### Test me
  
On both http://ci.terria.io/main/ and http://ci.terria.io/fix-resize-mobile/ check styling of the top bar of menu buttons and the items in the mobile hamburger menu when you:
* load the page in a small window ( < 768px) and resize to a large
* reload the page in a large window and resize to small

Old:
image
image

New:
image
image

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 
 #### next release (8.2.11)
 
+* Fixed a bug where menu items were rendered in the wrong style if the window was resized from small to large, or large to small.
 * [The next improvement]
 
 #### release 8.2.10 - 2022-08-02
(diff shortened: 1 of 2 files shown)


nf-s (approved):

zoran995 (approved):

na9da:
The UI is fixed but the globe still appears a bit stretched out, so do the stars.

Probably the canvas size is not getting updated?

image

na9da:
I only see that in firefox, can't repro in chrome.

PR #6480, opened 2022-08-05
Help video changes

na9da:
### What this PR does

* Add `videoCoverImageOpacity` option to `HelpContentItem` so that we can fade the background of help video panels.
* Fix a bug where all `HelpVideoPanel`s were being rendered simultaneously resulting in autoplayed videos playing at random.

### Checklist

### DIFF
--- a/lib/ReactViewModels/defaultHelpContent.ts
+++ b/lib/ReactViewModels/defaultHelpContent.ts
@@ -23,7 +23,7 @@ export interface HelpContentItem {
   placeholderImage?: string;
 
   // The `placeholderImage` is also used as background cover image for the container that embeds the video. This setting allows us to control the opacity of the cover image.
-  videoCoverImageOpacity?: boolean;
+  videoCoverImageOpacity?: number;
 
   paneMode?: PaneMode;
   trainerItems?: TrainerItem[];
--- a/lib/ReactViewModels/defaultHelpContent.ts
+++ b/lib/ReactViewModels/defaultHelpContent.ts
@@ -22,6 +22,9 @@ export interface HelpContentItem {
   videoUrl?: string;
   placeholderImage?: string;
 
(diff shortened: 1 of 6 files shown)


lib/ReactViews/Map/Panels/HelpPanel/HelpVideoPanel.jsx
```diff
@@ -49,15 +50,15 @@ class HelpVideoPanel extends React.Component {
       [Styles.shiftedToRight]:
         !isExpanded ||
         !this.props.viewState.showHelpMenu ||
-        this.props.viewState.topElement !== "HelpPanel",
-      [Styles.isHidden]: !itemSelected // when the item isn't selected
+        this.props.viewState.topElement !== "HelpPanel"
     });
     return (
-      <div className={className}>
+      itemSelected && <div className={className}>
```
na9da: Render this component only when it is selected. Not really sure why it was implemented this way. In my testing with videos that have autoplay enabled, this resulted in the user clicking one video but another video gets played.

lib/ReactViewModels/defaultHelpContent.ts
```diff
@@ -22,6 +22,9 @@ export interface HelpContentItem {
   videoUrl?: string;
   placeholderImage?: string;
 
+  // The `placeholderImage` is also used as background cover image for the container that embeds the video. This setting allows us to control the opacity of the cover image.
+  videoCoverImageOpacity?: boolean;
```
tephenavies: Should be `number`

tephenavies (dismissed):

lib/ReactViewModels/defaultHelpContent.ts
na9da: Ah yes!

mwu2018:
I've verified it with `dev` site.

mwu2018 (approved):

PR #6482, opened 2022-08-05
Fix `shouldShorten` property in catalog and story `ShareUrl`

### LINKED ISSUE #6478: Story share url not shortening
the Story Share function does not shorten the share URL, but the Share function does
To test: 
- using TerriaJS 8.2.10 try creating a story with 3 scenes and then share the url; it's too long

nf-s:
### Fix `shouldShorten` property in catalog and story `ShareUrl`

Fixes https://github.com/TerriaJS/terriajs/issues/6478

Catalog and Story share URL was using `terria.getLocalProperty("shortenShareUrls")` to determine if share URL can be shortened - but this is not defined for most users.

It now uses `canShorten(terria)` if `terria.getLocalProperty("shortenShareUrls")` is undefined.

### Test me
  
- http://ci.terria.io/fix-share-shorten/
- Create story
- Click share in story panel
- See shortened URL

Then go into main share panel and toggle "Shorten the share URL using a web service"

The setting should be reflected in all share panels (eg from story or catalog) - and setting value should be preserved in local storage (eg after refreshing the page)

### Checklist

### DIFF
--- a/lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
+++ b/lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
@@ -3,16 +3,15 @@ import { TFunction } from "i18next";
 import { observer } from "mobx-react";
 import React from "react";
 import { WithTranslation, withTranslation } from "react-i18next";
-
 import Terria from "../../../../Models/Terria";
 import ViewState from "../../../../ReactViewModels/ViewState";
-
+import Box from "../../../../Styled/Box";
+import Spacing from "../../../../Styled/Spacing";
+import Text from "../../../../Styled/Text";
+import { canShorten } from "./BuildShareLink";
 import Styles from "./share-panel.scss";
 import { SharePanelContent } from "./SharePanelContent";
 import { ShareUrl } from "./ShareUrl";
-import Box from "../../../../Styled/Box";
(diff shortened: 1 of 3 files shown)


lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
```diff
@@ -80,9 +79,8 @@ class SharePanel extends React.Component<PropTypes, SharePanelState> {
             terria={terria}
             viewState={viewState}
             includeStories={true}
-            shouldShorten={
-              !!terria.getLocalProperty("shortenShareUrls") ?? true
-            }
+            // Note we don't use terria.getLocalProperty("shortenShareUrls") here - as it is only configurable in SharePanelContent
+            shouldShorten={!!canShorten(terria)}
```
zoran995: Maybe check if local property is undefined and then use `canShorten` otherwise use local property value, so user choice doesn't get overridden?

tephenavies (approved):

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
nf-s: Done!

mwu2018 (approved):

PR #6485, opened 2022-08-05
Fix `generateCatalogIndex` for nested references

nf-s:
### Fix `generateCatalogIndex` for nested references

This PR adds support for loading nested references in the `generateCatalogIndex` script - for example

- `MagdaReference` -> `TerriaReference`

### Checklist


na9da (approved):
Looks good @nf-s !

When you revisit this in the future, it might make sense to extract some of the inner functions like `loadReference` as top-level function like `recursivelyLoadReference` to break up the function a bit.

PR #6488, opened 2022-08-08
Fix search debounce

### LINKED ISSUE #6405: `CatalogIndex` search sometimes hangs - to investigate
- https://nationalmap.gov.au/
- Explore map data
- Type in "Tree" (pretty quickly - but not too quickly!)
- ...nothing will happen
- add "s" (so search field is "trees")
- see results appear

Seems like a race condition.

nf-s:
###  Fix `SearchBox` handling of `searchWithDebounce` when `debounceDuration` prop changes. It now fushes instead of cancels.

Fixes https://github.com/TerriaJS/terriajs/issues/6405

Before this change, when `SearchBox.props.debounceDuration` was changing, it was cancelling `searchWithDebounce` instead of flushing - which means searches sometimes weren't getting triggered properly after the `CatalogIndex` started loading
  
#### With CatalogIndex
  
- http://ci.terria.io/fix-search-debounce/#configUrl=https://nationalmap.gov.au/config.json
- Open catalog search
- Type "Tree" quite fast
- Search results now appear

#### With vanilla search

- http://ci.terria.io/fix-search-debounce/
- Open catalog search
- Type "Test"

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ Change Log
 * Add `getFeatureInfoUrl` and `getFeatureInfoParameters` to `WebMapServiceCatalogItemTraits`
 * Fix `SearchBoxAndResults` Trans values
 * Fix `generateCatalogIndex` for nested references
+* Fix `SearchBox` handling of `searchWithDebounce` when `debounceDuration` prop changes. It now fushes instead of cancels.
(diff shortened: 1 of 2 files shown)


mwu2018 (approved):
Verified.

PR #6495, opened 2022-08-09
Change the prompt text for uploading local/web file.

na9da:
### What this PR does

- [x] Changes the prompt text for uploading local/web file
   * Dropped "optional" from the prompt text in file upload modal for both local and web data.
   * Changed the text for the first file upload option from "Auto-detect (recommended)" to simply "File type".
    Note:  we are only changing the text, the auto-detect option still works but we downplay it a bit. This is because the Auto-detect feature does not work reliably. In the future we can hopefully improve the auto-detection.
- [x] Automatically suffix supported extension list to the entries in file type dropdown to improve clarity.
- [x] Removed IFC from upload file type (until further testing) 

**Before:**
image
**After:**
image

  
### Test me
 
- CI link - http://ci.terria.io/upload-type-changes/
- 
### Checklist

### DIFF
--- a/lib/Core/getDataType.ts
+++ b/lib/Core/getDataType.ts
@@ -169,12 +169,6 @@ export default function(): GetDataTypes {
         value: "shp",
         name: i18next.t("core.dataType.shp"),
         extensions: ["zip"]
-      },
-      {
-        // NOTE: will only work if non open-source terriajs-ifc plugin is added to the map
-        value: "ifc",
-        name: i18next.t("core.dataType.ifc"),
-        extensions: ["ifc"]
       }
     ]
   };
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,11 @@ Change Log
 ==========
 
 #### next release (8.2.12)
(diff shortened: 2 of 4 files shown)


lib/Core/getDataType.ts
```diff
@@ -169,12 +169,6 @@ export default function(): GetDataTypes {
         value: "shp",
         name: i18next.t("core.dataType.shp"),
         extensions: ["zip"]
-      },
-      {
-        // NOTE: will only work if non open-source terriajs-ifc plugin is added to the map
-        value: "ifc",
-        name: i18next.t("core.dataType.ifc"),
-        extensions: ["ifc"]
```
na9da, line 177: Removed until further testing

nf-s (approved):
Looks good!

Please add entry to `CHANGES.md` before merging.

Also, in "Add web data" is it worth changing "Step 1: Select file type" to "Step 1: Select file or web service type"

image

na9da:
Good point! I have made the suggested changes.

PR #6502, opened 2022-08-10
Propagate `knownContainerUniqueIds` across references and their target.

nf-s:
### Propagate `knownContainerUniqueIds` across references and their target.

This is needed to fix `generateCatalogIndex` for nested references. But it is also important to have complete `knownContainerUniqueIds` when creating share links with nested references.

An example of a nested reference: `TerriaReference` -> `CkanItemReference` -> `CSVCatalogItem`

This PR makes sure that all three models have the same `knownContainerUniqueIds`

### Checklist


mwu2018 (approved):

mwu2018 (approved):

mwu2018:
@nf-s I've run prettier and pushed to your branch so the CI build would not fail. Could you please add description to this PR as why this PR is needed?

PR #6506, opened 2022-08-15
Set pedestrian heading to the camera heading when dropped.

### LINKED ISSUE #6426: When enabling Pedestrian View do not adjust view direction
When we first enable Pedestrian View and click on the map we also spin the viewport around to face North. This is disorienting to the user - we shouldn't change things they haven't asked for. Instead just enable Pedestrian View and take the viewport down to the ground, but if the user was already looking in a direction (i.e. they were not looking straight down) then do not change the horizontal direction they were already looking in.

E.g. the user is looking at this with Pedestrian View off, and they click on the bridge by the X, expecting to get a view along the river:
Screen Shot 2022-07-15 at 12 57 37

What they expect to happen:
Screen Shot 2022-07-15 at 12 59 29

What actually happens:
Screen Shot 2022-07-15 at 12 58 04

na9da:
### What this PR does

Fixes #6426 

Previously, when dropping the pedestrian to ground, we set the camera heading to 0 (= north). This puts the pedestrian in a different direction than the one they were facing before the drop. 

This PR sets the pedestiran heading to the camera heading so that the user faces the same direction after the drop.
  
### Test me
  
1. Open the current main branch CI - http://ci.terria.io/main/#share=s-rbo9PmJairGOHghvdwJ59Ml15Dh
2. Observe that the compass `N` is at the bottom 
3. Drop the pedestrian to ground 
4. Observe that the compass `N` is now at the top
5. Repeat the same for this branch - http://ci.terria.io/fix-pedestrian-heading/#share=s-rbo9PmJairGOHghvdwJ59Ml15Dh
6. Observe that dropping the pedestrian does not change the compass heading

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Change Log
 
 #### next release (8.2.13)
 
+* Fix pedestrian drop behaviour so that the camera heading stays unchanged even after the drop
 * [The next improvement]
 
 #### release 8.2.12 - 2022-08-10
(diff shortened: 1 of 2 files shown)


meh9:
Awesome!

mwu2018 (approved):
Tested, as expected.

PR #6511, opened 2022-08-17
Fix loading of EPSG:4326, WMS 1.3.0 services

### LINKED ISSUE #6510: WMS 1.3.0 relies on EPSG to specify lat/lon order resulting in Terria failing to load EPSG:4326 layers
The WMS spec says that for version 1.3.0 it relies on the lat/lon / x/y order to be specified by the EPSG database. Summary:

https://docs.geoserver.org/latest/en/user/services/wms/basics.html#axis-ordering

In the case of EPSG:4326 this means that the lat/lon order changes from WMS 1.1 to 1.3.0. However, Terria does not comply with this.

However, because the vast majority of WMS servers will also have EPSG:3857 available we usually don't encounter this problem since Terria will always prefer EPSG:3857.

Here are two examples of servers that do have EPSG:4326 but do not have EPSG:3857:

https://services.slip.wa.gov.au/public/services/SLIP_Public_Services/Environment/MapServer/WMSServer?service=WMS&version=1.3.0&request=GetCapabilities

https://www.asris.csiro.au/arcgis/services/TERN/AWC_ACLEP_AU_NAT_C/MapServer/WMSServer?request=GetCapabilities&service=WMS

The result is that we cannot view any layers from these servers.

Here are example tile URL's showing the issue:
- original terria request with incorrect lon/lat bbox: https://www.asris.csiro.au/arcgis/services/TERN/AWC_ACLEP_AU_NAT_C/MapServer/WMSServer?transparent=true&format=image/png&exceptions=XML&styles=default&version=1.3.0&crs=EPSG:4326&service=WMS&request=GetMap&layers=AWC_000_005_05_N_P_AU_NAT_C&bbox=129.37500000000003,-33.749999999999986,135,-28.125000000000007&width=256&height=256
- modified request with lat/lon bbox: https://www.asris.csiro.au/arcgis/services/TERN/AWC_ACLEP_AU_NAT_C/MapServer/WMSServer?transparent=true&format=image/png&exceptions=XML&styles=default&version=1.3.0&crs=EPSG:4326&service=WMS&request=GetMap&layers=AWC_000_005_05_N_P_AU_NAT_C&bbox=-33.749999999999986,129.37500000000003,-28.125000000000007,135&width=256&height=256

It is likely that the bug is that Terria does not look at the EPSG specification to determine lat/lon order for the datasets. The correct fix for this is to make Terria do that. However, if I remember correctly Terria only loads EPSG 3857 and 4326 and no other EPSG codes, so an acceptable fix may be to reverse the order for WMS 1.3.0 EPSG:4326 only.

na9da:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/6510

Currently we pass `crs` and `srs` as request parameters to `WebMapServiceImageryProvider` instead of passing them as constructor options. This has 2 problems:

- Cesium ~overrides the query parameters~ sets `srs` and `crs` query parameter from `options` [here](https://github.com/CesiumGS/cesium/blob/fceb03ade47a0b994581c8f0c2727129b3221945/Source/Scene/WebMapServiceImageryProvider.js#L263) and [here](https://github.com/CesiumGS/cesium/blob/fceb03ade47a0b994581c8f0c2727129b3221945/Source/Scene/WebMapServiceImageryProvider.js#L291) using values from the constructor `options`.
- When not setting `options.crs`, Cesium chooses a default value of `CRS:84` instead of `EPSG:4326` (in 3d mode), and so skips the `bbox` transformations that are required for correctly loading `EPSG:4326` in WMS version v1.3.0. This causes the above bug (inverted bbox for GetMap queries resulting in server returning transparent tiles)

This PR fixes it by correctly passing `crs` and `srs` as constructor `options` instead of as query parameters.

### Test me
  
- Open this [main branch link](http://ci.terria.io/main/#share=s-o66LYvxOLND41CGwHPuUKlfIBG3) that loads a WMS endpoint which only supports EPSG:4326
- Observe that nothing is rendered on the map, `GetMap` requests are send but returns transparent images. This is because WMS 1.3.0 expects a reversed `bbox` for `EPSG:4326` but the reversing does not happen
- Now open the [link for this branch](http://ci.terria.io/pass-wms-crs/#share=s-o66LYvxOLND41CGwHPuUKlfIBG3)
- Observe that the dataset is rendered correctly on the map
- Test other WMS endpoints to ensure nothing else is broken

### Checklist


lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -511,18 +511,6 @@ class WebMapServiceCatalogItem
...
-      if (this.crs) {
-        if (this.useWmsVersion130) {
-          parameters.crs = this.crs;
-          getFeatureInfoParameters.crs = this.crs;
-        } else {
-          parameters.srs = this.crs;
-          getFeatureInfoParameters.srs = this.crs;
-        }
-      }
-
```
na9da, line 523: Instead of passing `crs` `srs` as `parameters`, pass them as constructor options to `WebMapServiceImageryProvider` below.

tephenavies (approved):

PR #6532, opened 2022-09-06
Update a prompt text in DataPreview

ykiu:
### What this PR does

The functionality of adding multiple datasets at once by pressing ctrl/shift is great, but is hard to discover.

This commit improves the message displayed in the DataPreview to make the feature more discoverable.

### Test me

Launch TerriaMap and see if the message is appropriate.

image

### Checklist

  - unit tests are not applicable because there're no changes to the code


nf-s (changes requested):
Hi @ykiu 

Thanks for making this PR!
Sorry it has taken us so long to review it.

I have a few comments

- Can you please replace the `"+"` with the plus icon?
  - image
  - You should be able to use `<Icon glyph={Icon.GLYPHS.add} />` 
- Can you please change "Ctrl or Shift" to just "Shift". Ctrl doesn't work on mac, and Shift should work on all platforms

Let me know what you think, thanks 🙂

ykiu:
Thanks @nf-s , that way it looks nicer. I've updated the PR and also reorganized the relevant parts of the dictionaries so that they are easier to work with for translators. By the way I'm not a native speaker of English so please feel free to correct the English message I added, if it sounds weird.

image

nf-s (approved):
Thanks so much @ykiu !

Sorry for the delay in review. This is a great improvement 🙂

ykiu:
@nf-s Thank you!

PR #6536, opened 2022-09-08
Make WFS Request params more dynamic

### LINKED ISSUE #6518: Provide an option to override WFS `srsName` parameter by passing it as URL parameter?
Our current WFS implementation hardcodes `srsName` parameter as `urn:ogc:def:crs:EPSG::4326` ([here](https://github.com/TerriaJS/terriajs/blob/ddc258a132c7252c5aa20d689b31ba372e00737a/lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts#L363)).

WFS servers accepts different formats for `srsName`:
 -  `urn:` is a good default setting because it has a well known [spec enforced](https://docs.geoserver.org/latest/en/user/services/wfs/axis_order.html) lat,lon ordering for the returned WFS geometries.  
 - `EPSG:XXXX` is not reliable because different servers return different ordering (lat,lon or lon,lat) and it is hard to tell.

There are however some servers that return the correct lat,lon ordering that Terria epxects for `EPSG:4326` but does not accept the `urn:` format. Currently, it is not possible for a user to use the `Add web data` option to add such a WFS endpoint to Terria. 
eg: https://webmap.yvw.com.au/YVWassets/service.svc/get?request=GetFeature&service=WFS&typeName=yvw:NDWHYDRANTS&version=1.1.0&srsName=EPSG:4326

This issue is to discuss whether we should read the `srsName` parameter from the WFS input url to override our default `urn:` setting.

staffordsmith83:
### What this PR does

Fixes #6518

`srsName` and `outputFormat` were hard coded for WFS Requests.
This PR makes both of these param values intelligent, now pulled from the GetCapabilities response.

This resolves two issues, servers that cannot handle `urn` type `srsName` values, and servers that do not accept `gml3` as a value for `outputFormat` but would prefer the string `text/xml; subtype=gml/3.1.1`

### Test me

Go to before and after links and add 'NDW Hydrants' item from map catalog provided by gist.

### Before
http://ci.terria.io/main/#clean&https://gist.githubusercontent.com/staffordsmith83/ee13108b303c599e33f3c7b89950ca45/raw/abd173b89f26897643995a25531d8282faf9c00c/wfsTestCat.json

Request fails for 2 reasons:
"Value gml3 provided for a parameter OUTPUTFORMAT is invalid."
"An unexpected error occurred. Original message: 'Value cannot be null.
Parameter name: key'" - this is due to using `srsName=urn:ogc:def:crs:EPSG::4326`and works if we change the request to use `srsName=EPSG:4326`

### After
http://ci.terria.io/wfs-params-dynamic/#clean&https://gist.githubusercontent.com/staffordsmith83/ee13108b303c599e33f3c7b89950ca45/raw/abd173b89f26897643995a25531d8282faf9c00c/wfsTestCat.json

Now both `srsName` and `outputFormat` params are filled from the getCapabilities response for the WFS service.

## UPDATED: Test user supplied `srsName`

Go to Explore Map Data>My Data>Add Web Data
Try adding these WFS getFeature requests:

http://geossdi.dmp.wa.gov.au/services/wfs?service=WFS&version=1.1.0&request=GetFeature&typeName=mt:MineralTenement&outputFormat=JSON&srsName=urn:ogc:def:crs:EPSG::4326

http://geossdi.dmp.wa.gov.au/services/wfs?service=WFS&version=1.1.0&request=GetFeature&typeName=mt:MineralTenement&outputFormat=JSON&srsName=EPSG:4326

http://geossdi.dmp.wa.gov.au/services/wfs?service=WFS&version=1.1.0&request=GetFeature&typeName=mt:MineralTenement&outputFormat=JSON&srsName=CRS:84

http://geossdi.dmp.wa.gov.au/services/wfs?service=WFS&version=1.1.0&request=GetFeature&typeName=mt:MineralTenement&outputFormat=JSON&srsName=EPSG:4283

Note in the network tab that the request is made with the user specified srsName. This is because each of these srsNames is in the terria supported list.

Now try with 
http://geossdi.dmp.wa.gov.au/services/wfs?service=WFS&version=1.1.0&request=GetFeature&typeName=mt:MineralTenement&outputFormat=JSON&srsName=EPSG:9999

Note that the request defaults to use `srsName=urn:ogc:def:crs:EPSG::4326` because `EPSG:9999` is not in our supported list of CRS.

### Checklist

~~- [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~~
- [y] I've updated CHANGES.md with what I changed.
- [y] I've provided instructions in the PR description on how to test this PR.


lib/Models/Catalog/Ows/WebFeatureServiceCapabilities.ts
```diff
@@ -134,6 +134,41 @@ function getOutputTypes(json: any): string[] | undefined {
...
+interface SrsNamesForLayer {
+  layerName: string;
+  srsArray: string[]; // First element is DefaultSRS
+}
+
+/**
+ * Get the coordinate systems (srsName) supported by the WFS service for each layer.
+ * @param json
+ * returns an object with an array of srsNames for each layer. The first element is the defaultSRS as specified by the WFS service.
+ * TODO: For catalog items that specify which layer we are interested in, why build the array describing the srsNames for all the other layers too?
```
staffordsmith83, line 146: Does this introduce inefficiency in the code?

staffordsmith83: RE: "TODO: For catalog items that specify which layer we are interested in, why build the array describing the srsNames for all the other layers too?"

nf-s: I wouldn't worry. The main issue is that we have to load these `GetCapabilities` documents - which can be multiple MB - usually there is more overhead in load time than processing them in JS.

staffordsmith83: Yeah its not a complex operation quick as. As long as its in the right place Im happy cheers

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
```diff
@@ -332,21 +332,47 @@ class WebFeatureServiceCatalogItem extends GetCapabilitiesMixin(
...
+        (layer) => layer.layerName === layerNamesArray[0]
+      );
+
+      const searchValue = new RegExp("4326");
+      const result =
+        layerSrsArray?.srsArray.find((srsName) => searchValue.test(srsName)) ??
+        "urn:ogc:def:crs:EPSG::4326"; // Default to urn identifier for WGS84 if we cant find something better.
+      return result;
+    };
+
```
nf-s: I think it would be better to create two new `Traits` in `WebFeatureServiceCatalogItemTraits` - something like `srsName` and `outputFormat`.

Then you can add default values for both in `GetCapabilitiesStratum` (in `lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts`). It would also simplify the `forceLoadGeojsonData` function - as you can move all of that logic into `GetCapabilitiesStratum`.

This would allow the values to be overridden in JSON catalog.

**A similar example is how `crs` is handled in WMS:**

Where we have a `crs` trait in `WebMapServiceCatalogItemTraits`

https://github.com/TerriaJS/terriajs/blob/588eb6a63bc1c23d1629d5fc0a4920ffff4b5411/lib/Traits/TraitsClasses/WebMapServiceCatalogItemTraits.ts#L195-L202

Which is given a default value in `WebMapServiceCapabilitiesStratum`

https://github.com/TerriaJS/terriajs/blob/cad62a45cbee98c7561625458bec3a48510f6cbc/lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts#L299-L314

This value is then accessed in `WebMapServiceCatalogItem` like so

https://github.com/TerriaJS/terriajs/blob/ce07007dc068e737c8bc8d5ee8b666ab590108d3/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts#L549-L569

nf-s (changes requested):

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
staffordsmith83: Great, thanks for the advice Nick! Just what I was looking for, will dig in in this direction.

staffordsmith83: OK @nf-s Ive stripped back from where I got too complex and implemented the important parts in a simple neat way that can be overridden with the traits system, I think.
One thing I havent done, in contrast to `WebMapServiceCatalogItem`, is to check for `srsName` in the provided url. I dont think this is important, we just override with the best one anyway.

Let me know if this is more on track, thanks for your help.

staffordsmith83:
Also @nf-s where is a good place to put the `SUPPORTED_CRS_4326` array? I got this from WebMapServiceCatalogItemTraits.ts but now relevant to WFS too... Its also different, as I added the urn identifier to the first element of the array for WFS.

nf-s:
@staffordsmith83 If it is different from WMS - I would just put a full copy in `WebFeatureServiceCatalogItemTraits`

staffordsmith83:
Done

staffordsmith83:
I see that the issue this PR stems from actually says :"Provide an option to override WFS `srsName` parameter by passing it as URL parameter" - i need to implement that logic similar to WMS I think! But is this in general more on the right track for the way Terria works
?

nf-s:
> But is this in general more on the right track for the way Terria works

Yeah it would be really nice if WFS picks up on `srsName` in the URL provided - especially for when a user adds their own WFS to an existing map.

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
```diff
@@ -359,8 +387,8 @@ class WebFeatureServiceCatalogItem extends GetCapabilitiesMixin(
             request: "GetFeature",
             typeName: this.typeNames,
             version: "1.1.0",
-            outputFormat: supportsGeojson ? "JSON" : "gml3",
-            srsName: "urn:ogc:def:crs:EPSG::4326", // srsName must be formatted like this for correct lat/long order  >:(
+            outputFormat: supportsGeojson ? "JSON" : this.outputFormat, // Will choose best option from GetCapabilities response TODO: should we move JSON conditional stuff to where default is set too?
```
nf-s: See https://github.com/TerriaJS/terriajs/pull/6536#discussion_r973856595

```suggestion
         outputFormat: this.outputFormat,
```

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
```diff
@@ -341,15 +369,15 @@ class WebFeatureServiceCatalogItem extends GetCapabilitiesMixin(
...
       [
         ...getCapabilitiesStratum.capabilitiesFeatureTypes.values()
       ].reduce<boolean>(
         (hasGeojson, current) =>
-          hasGeojson && hasOutputFormat(current?.OutputFormats),
+          hasGeojson && hasJsonOutputFormat(current?.OutputFormats),
         true
       );
-
+    ``;
```
nf-s: I think all of this logic should be in your new `@computed get outputFormat() ...` in `GetCapabilitiesStratum`

https://github.com/TerriaJS/terriajs/blob/7959181bae353621a34dfbbcd4946b6f00c457de/lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts#L241-L250

Then you could just have `outputFormat: this.outputFormat` (see https://github.com/TerriaJS/terriajs/pull/6536#discussion_r973856575).

This means that the outputFormat can be overridden in a JSON catalog file if needed.

lib/Traits/TraitsClasses/WebFeatureServiceCatalogItemTraits.ts
```diff
@@ -25,6 +32,22 @@ export default class WebFeatureServiceCatalogItemTraits extends mixTraits(
...
+    description: `Spatial Reference System to use. For WFS we prefer WGS 84 (${SUPPORTED_CRS_4326.join(
+      ", "
+    )}). With WFS requests it is best to use the urn identifier for the srsName, to enforce lat,long order in returned results.`
+  })
+  srsName?: string;
+
+  @primitiveTrait({
+    type: "string",
+    name: "Output Format",
+    description: `Output format to request for WFS requests. We support gml3 and gml3.1.1.`
```
nf-s: It might be worth adding that we support GeoJSON - and saying we only support GML in 4326 - to the `description`

nf-s (changes requested):

staffordsmith83:
I think last thing to do is to provide a mechanism to check if the axis order from a WFS server is coming back wrong. EPSG:4326 can be either y,x or x,y it seems. That is why we prefer `urn...4326...` so if we fall back on using `EPSG:4326` we should check the returned coords axis order I suppose...

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
```diff
@@ -235,6 +238,65 @@ class GetCapabilitiesStratum extends LoadableStratum(
...
+  };
+
+  // Helper function for get outputFormat()
+  supportsGeojson =
+    this.hasJsonOutputFormat(this.capabilities.outputTypes) ||
+    [...this.capabilitiesFeatureTypes.values()].reduce<boolean>(
+      (hasGeojson, current) =>
+        hasGeojson && this.hasJsonOutputFormat(current?.OutputFormats),
+      true
+    );
```
nf-s: This is non-observable - which may cause issues as `this.capabilitiesFeatureTypes` is `@computed`.

To fix this you can put it inside `get outputFormat` or create a new `@computed` property

nf-s (changes requested):

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
staffordsmith83: Ah great, yes I was playing around with that, one approach was a had a new trait `supportsGeojson` - I dont think we need that. If I make a new computed property should it also be a trait? I would prefer to put inside `get outputFormat` rather than that I reckon

nf-s: I agree, I don't think we need a `supportsGeojson` trait.

It's totally fine to create `@computed` properties that aren't in the traits system.

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
staffordsmith83: Done

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
staffordsmith83: Ah I was thrown by this error when I had it marked as `observable`: 
```
Catalog item could not be added to workbench.: [mobx] Trying to declare a computed value for unspecified getter 'supportsGeojson'
  Failed to load `NDW Hydrants` metadata
```

staffordsmith83: Done, put inside `get outputFormat`

staffordsmith83:
> I think last thing to do is to provide a mechanism to check if the axis order from a WFS server is coming back wrong. EPSG:4326 can be either y,x or x,y it seems. That is why we prefer `urn...4326...` so if we fall back on using `EPSG:4326` we should check the returned coords axis order I suppose...

I think this issue is now resolved for all spec compliant WFS services. We only fall back on `EPSG:4326` if `urn:ogc:def:crs:EPSG::4326` is not available. If WFS services do not support `urn:ogc:def:crs:EPSG::4326`, then their `EPSG:4326` response should be in the correct order. 

See https://github.com/qgis/QGIS/issues/36584#issuecomment-631722599

Further work on this would be to support non complaint WFS servers which we are not going to spend too much time on now.

lib/Models/Catalog/Ows/WebFeatureServiceCatalogItem.ts
```diff
@@ -235,6 +238,64 @@ class GetCapabilitiesStratum extends LoadableStratum(
...
+    // First check to see if URL has CRS or SRS
+    const supportedCrs = [...SUPPORTED_CRS_3857, ...SUPPORTED_CRS_4326];
+    const queryParams: any = this.catalogItem.uri?.query(true) ?? {};
+    const urlCrs =
+      queryParams.srsName ??
+      queryParams.crs ??
+      queryParams.CRS ??
+      queryParams.srs ??
+      queryParams.SRS;
+    if (urlCrs && supportedCrs.includes(urlCrs)) return urlCrs;
```
staffordsmith83, line 290: One more question @nf-s  do we need to check if this user specified `srsName` is supported by the WFS in the getCapabilities?
Otherwise, if we try adding a url that IS in our list of supported CRS, but where its not supported by the addede WFS service, we get an error, but it's not very descriptive.

E.g. try to add this through "Add Web Data" : 
https://webmap.yvw.com.au/YVWassets/service.svc/get?request=GetFeature&service=WFS&typeName=yvw:NDWHYDRANTS&version=1.1.0&srsName=EPSG:4283

Screen Shot 2022-09-21 at 8 09 58 am


Problem here is that WFS services dont necessarily specify ALL the srsNames they support in the getCapabilities response. I am starting to see some downsides to WFS!

My thoughts are: We should not dictate what a user can add too much - if they want to specify an `srsName` for their added WFS service, they can. If we support it, we will try to use it. But if their WFS service does not support it, thats their problem. We cant really enforce the check against getCapabilities, because it does not always return the full list of supported CRS.

nf-s: Yep I agree. If a `srsName` is provided by the user (either through URL or JSON catalog file), we should just use it even if it isn't advertised.

nf-s (approved):

PR #6542, opened 2022-09-14
Carto API v3 (Map Table + Query) support

### LINKED ISSUE #6323: Basic support for Carto API v3
- example of Carto 3.0 vector tiles/api docs  - https://api-docs.carto.com/

nf-s:
### Carto API v3 (Map Table + Query) support

Fixes https://github.com/terriajs/terriajs/issues/6323

### Todo

- [x] Handle error messages from Carto API

### Test me

See https://github.com/TerriaJS/terriajs/issues/6323#issuecomment-1247866595 for example configs and test links

### Checklist


staffordsmith83 (changes requested):
I managed to get an error clicking around with both of the example datasets loaded from https://github.com/TerriaJS/terriajs/issues/6323#issuecomment-1154747680

```bash
Uncaught Error: [mobx] Computed values are not allowed to cause side effects by changing observables that are already being observed. Tried to modify: FeatureInfoSection@4843.featureChangedCounter
    at invariant (mobx.module.js?daf9:29:1)
    at fail (mobx.module.js?daf9:24:1)
    at checkIfStateModificationsAreAllowed (mobx.module.js?daf9:763:1)
    at ObservableValue.prepareNewValue (mobx.module.js?daf9:1085:1)
    at ObservableObjectAdministration.write (mobx.module.js?daf9:4163:1)
    at FeatureInfoSection.set [as featureChangedCounter] (mobx.module.js?daf9:4327:1)
    at eval (FeatureInfoSection.tsx?490b:131:1)
    at executeAction (mobx.module.js?daf9:951:1)
    at runInAction (mobx.module.js?daf9:2086:1)
    at FeatureInfoSection.eval (FeatureInfoSection.tsx?490b:130:1)
```

To replicate, load both the table and query item with this sharelink:
http://ci.terria.io/carto-v3/#clean&https://gist.githubusercontent.com/nf-s/c2485b3d2f05cbededdaa50e85a64db9/raw/87f132a401682098599ee911383d0c82543cbd19/carto-v3.json&share=s-pG3vSJFltAEvdLB0WiZIlNdjJFX

Click on a feature to get FeatureInfo to show up,
then toggle the visibility of that layer to off in the workbench.
Application crashes. Occurs for either table or query layer. Dont have to have both loaded for crashes to occur.

image

nf-s:
> I managed to get an error clicking around with both of the example datasets loaded from [#6323 (comment)](https://github.com/TerriaJS/terriajs/issues/6323#issuecomment-1154747680)
> 
> ```shell
> Uncaught Error: [mobx] Computed values are not allowed to cause side effects by changing observables that are already being observed. Tried to modify: FeatureInfoSection@4843.featureChangedCounter
>     at invariant (mobx.module.js?daf9:29:1)
>     at fail (mobx.module.js?daf9:24:1)
>     at checkIfStateModificationsAreAllowed (mobx.module.js?daf9:763:1)
>     at ObservableValue.prepareNewValue (mobx.module.js?daf9:1085:1)
>     at ObservableObjectAdministration.write (mobx.module.js?daf9:4163:1)
>     at FeatureInfoSection.set [as featureChangedCounter] (mobx.module.js?daf9:4327:1)
>     at eval (FeatureInfoSection.tsx?490b:131:1)
>     at executeAction (mobx.module.js?daf9:951:1)
>     at runInAction (mobx.module.js?daf9:2086:1)
>     at FeatureInfoSection.eval (FeatureInfoSection.tsx?490b:130:1)
> ```
> 
> To replicate, load both the table and query item with this sharelink: http://ci.terria.io/carto-v3/#clean&https://gist.githubusercontent.com/nf-s/c2485b3d2f05cbededdaa50e85a64db9/raw/87f132a401682098599ee911383d0c82543cbd19/carto-v3.json&share=s-pG3vSJFltAEvdLB0WiZIlNdjJFX
> 
> Click on a feature to get FeatureInfo to show up, then toggle the visibility of that layer to off in the workbench. Application crashes. Occurs for either table or query layer. Dont have to have both loaded for crashes to occur.
> 
> image

Good find @staffordsmith83 - this issue relates to all GeoJSON isn't cause by this PR so I have created a separate issue - https://github.com/TerriaJS/terriajs/issues/6576

staffordsmith83:
OK @nf-s  Ill review this as is!

lib/Traits/TraitsClasses/CartoMapV1CatalogItemTraits.ts
```diff
@@ -23,6 +23,7 @@ export default class CartoMapCatalogItemTraits extends mixTraits(
   })
   config?: JsonObject | string;
 
+  /** TODO: Make this camel case please */
```
staffordsmith83, line 30: Can we remove this comment now? Or are you referring to `auth_token`?

wwwroot/languages/en/translation.json
```diff
@@ -716,7 +716,8 @@
       "czml": "CZML",
       "gpx": "GPX",
       "json": "Terria Catalog",
-      "carto": "Carto",
+      "carto": "Carto V1",
```
staffordsmith83, line 719: does the key "carto" need to be renamed "carto-v1" throughout? Especially if carto-v3 becomes the standard...

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+        this.strata.set(
+          CartoMapV3Stratum.stratumName,
+          new CartoMapV3Stratum(this)
+        );
+      });
+    }
+  }
+
+  get type() {
+    return CartoMapV3CatalogItem.type;
```
staffordsmith83, line 88: I assume this always returns "carto-v3" because of line 59

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  get typeName() {
+    return i18next.t("models.carto-v3.name");
```
staffordsmith83, line 92: So should this be 
```ts
return i18next.t(`models.${this.type}.name`)
```
or something like that. In case we change the type name

lib/Traits/TraitsClasses/CartoMapV3CatalogItemTraits.ts
```diff
@@ -0,0 +1,62 @@
...
+  })
+  connectionName = "carto_dw";
+
+  @primitiveTrait({
+    type: "string",
+    name: "Base URL",
+    description:
+      'Base URL for Carto API (eg "https://gcp-us-east1.api.carto.com/")'
+  })
+  baseUrl = "https://gcp-us-east1.api.carto.com/";
```
staffordsmith83: I suppose this is the default if a baseUrl trait is not provided in a catalog item definition? So is `gcp-us-east1` the best or maybe `gcp-australia-southeast1.api.carto.com` like in the examples? More of our users in Aus or not?

lib/Traits/TraitsClasses/CartoMapV3CatalogItemTraits.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  @primitiveTrait({
+    type: "string",
+    name: "Geo column name",
```
staffordsmith83, line 46: Should be "Geom column name"?

lib/Traits/TraitsClasses/CartoMapV3CatalogItemTraits.ts
```diff
@@ -0,0 +1,62 @@
...
+    description:
+      "Column name of the geom at the table (used for Table and Query API)"
+  })
+  cartoGeoColumn = "geom";
+
+  @primitiveTrait({
+    type: "string",
+    name: "Geo column name",
```
staffordsmith83: Wrong value for name

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  protected async forceLoadMetadata() {
+    let response: JsonObject | undefined;
+
+    /** If cartoQuery is defined - use Query API (https://api-docs.carto.com/#8f2020d9-edf3-4b50-ae58-9edeaa34613c)
```
staffordsmith83: Different type of comment here and at line 111. /** */ vs //

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    if (
+      response &&
+      isJsonObject(response.tilejson) &&
+      isJsonStringArray(response.tilejson.url)
+    ) {
+      mvtTileJsonUrls = response.tilejson.url;
+    }
+
+    if (geoJsonUrls.length === 0 /*&& mvtTileJsonUrls.length === 0*/) {
```
staffordsmith83, line 153: Can we remove commented out bit now?

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+      ).throwIfError();
+    } else {
+      throw new TerriaError({
+        title: "Invalid Carto V3 config",
+        message: "`cartoQuery` or `cartoTableName` must be defined"
+      });
+    }
+
+    let geoJsonUrls: string[] = [];
+    let mvtTileJsonUrls: string[] = [];
```
staffordsmith83: Comment out until we add support?

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+export default class CartoMapV3CatalogItem extends GeoJsonMixin(
+  CreateModel(CartoMapV3CatalogItemTraits)
+) {
+  static readonly type = "carto-v3";
+
+  @observable
+  geoJsonUrls: string[] = [];
+
+  @observable
+  mvtTileJsonUrls: string[] = [];
```
staffordsmith83: Comment out until we add support? Otherwise implies that its supported already to a casual code reviewer...

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
```diff
@@ -0,0 +1,290 @@
...
+    //       title: "Failed to load GeoJSON",
+    //       message: `Invalid response from GeoJSON URL ${url}:\n\n
+    //       \`\`\`
+    //       ${JSON.stringify(jsonData)}
+    //       \`\`\``
+    //     });
+    //   })
+    // );
+
+    // Merge all geojson responses into a combined feature collection
```
staffordsmith83, line 233: This should always be fine hey? Any situation where we would not want to combine all the geojson into one Feature collection? Would sometimes a request to Carto return different version of the same data, and therefore combining it would lead to duplicates?

staffordsmith83 (changes requested):
All looks really good, found a few things to check, in individual comments.

One thing: How is the documentation generated e.g. https://docs.terria.io/guide/connecting-to-data/catalog-type-details/carto/ ? Does there need to be a bit more added to the V3 one about 'table API' vs 'query API'?

lib/Traits/TraitsClasses/CartoMapV1CatalogItemTraits.ts
nf-s: I'm referring to `auth_token`. I don't want to change it at the moment as it will break the existing API.

wwwroot/languages/en/translation.json
nf-s: I have kept it `"carto"` so it matches `CartoMapV1CatalogItem.type`

nf-s: At some point we will want to swap them over - but that will break existing API

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: Yep. 

We need to implement the following for all `CatalogMemberMixin` items

https://github.com/TerriaJS/terriajs/blob/cad62a45cbee98c7561625458bec3a48510f6cbc/lib/ModelMixins/CatalogMemberMixin.ts#L29

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: It could be, but there isn't an explicit link between catalog member `type` and our translation strings - we just try to use the same type name for consistency.

I find `i18next.t("models.carto-v3.name")` more readable

lib/Traits/TraitsClasses/CartoMapV3CatalogItemTraits.ts
nf-s: Good idea, I'll change it

lib/Traits/TraitsClasses/CartoMapV3CatalogItemTraits.ts
nf-s: The API call to Carto uses `geo_column` - so I would prefer to keep it is close to that as possible.

lib/Traits/TraitsClasses/CartoMapV3CatalogItemTraits.ts
nf-s: fixed!

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: fixed!

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: Done!

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: We need to leave it as a comment until we add tile JSON support

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: Done!

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
nf-s: I actually couldn't come across any examples of the Carto API returning geoJSON URLs - but it does return an array of URLs.

I figured the best default behaviour would be to merge them. Alternatively, I can just use the first geoJSON URL - or maybe throw an error if multiple URLs are returned

nf-s:
> How is the documentation generated

At the moment documentation is generated from `Traits` using the `generateDocs.ts`.

I have added a description to both carto v1 and v3 - https://github.com/TerriaJS/terriajs/commit/dfaeed3fa2daaea8bd66a796712a4a46464b69f7

lib/Models/Catalog/CatalogItems/CartoMapV3CatalogItem.ts
staffordsmith83: I think merge is probably best actually. Cheers

staffordsmith83 (approved):
Cheers for the changes, looks great!

imakihi:
Thank you @nf-s !! Great work.

PR #6547, opened 2022-09-16
Add `excludeInactiveDatasets` to `CkanCatalogGroup`

### LINKED ISSUE #1143: DGA CKAN filter out "data_state" = "inactive"

nf-s:
### Add `excludeInactiveDatasets` to `CkanCatalogGroup`

Fixes https://github.com/TerriaJS/nationalmap/issues/1143

Added `excludeInactiveDatasets` to `CkanCatalogGroup` (`true` by default). This will filter out CKAN Datasets which have `state` or `data_state` (data.gov.au specific) **not** set to `"active"`.

### After vs Before

In NationalMap data.gov.au CKAN - lots of problematic organisations have been excluded

image

### Test me

- **Before** http://ci.terria.io/main/#configUrl=https://terria-catalogs-public.storage.googleapis.com/nationalmap/config/dev.json
- **After** http://ci.terria.io/ckan-filter-datasets/#configUrl=https://terria-catalogs-public.storage.googleapis.com/nationalmap/config/dev.json

### Checklist

- [ ] ~I've updated relevant documentation in `doc/`.~


lib/Models/Catalog/Ckan/CkanCatalogGroup.ts
```diff
@@ -261,6 +261,15 @@ export class CkanServerStratum extends LoadableStratum(CkanCatalogGroupTraits) {
       return;
     }
 
+    // Return if excluding inactive dataset and dataset is inactive
+    if (
```
staffordsmith83, line 269: Is it possible to have a case where only one of `state` or `data_state` is active and that means it's still a valid working datasource? Like state=active and dataState=inactive? Probably not just checking.

Also, is the only allowable value for `dataState` or `state` "active"? Is this a ckan standard? I see in the docs:

"state (string) – the current state of the dataset, e.g. 'active' or 'deleted', only active datasets show up in search results and other lists of datasets, this parameter will be ignored if you are not authorized to change the state of the dataset (optional, default: 'active')"

But it does say "e.g. 'active' or 'deleted'" implying that other values can be used here... Is it possible we need to account for a wider range of values? Current logic just says that if the flag is present, and it has any other value than active, dont show it... Reckon thats OK?

lib/Traits/TraitsClasses/CkanCatalogGroupTraits.ts
```diff
@@ -60,4 +60,11 @@ export default class CkanCatalogGroupTraits extends mixTraits(
     description: `True to allow entire WMS servers (that is, WMS resources without a clearly-defined layer) to be added to the catalog; otherwise, false.`
   })
   allowEntireWmsServers: boolean = true;
+
+  @primitiveTrait({
+    type: "boolean",
+    name: "Exclude inactive datasets",
+    description: `True to remove inactive datasets (\`state\` is not \`active\` OR )\`data_state\` is not \`active\`.`
```
staffordsmith83: How come the backticks around the entire description? Should I be doing this?

staffordsmith83 (changes requested):
Just to make sure you have considered the other possible values for `state` - I dont know much about ckan so happy for a quick response and Ill approve.

I can see how its cleaned up heaps good idea!

lib/Traits/TraitsClasses/CkanCatalogGroupTraits.ts
nf-s: Nah I don't know why I did that. I have changed to double quotes.

nf-s: Sometimes I prefer backticks if I have nested quotes and things in the description - as the field is rendered as markdown in our docs

lib/Models/Catalog/Ckan/CkanCatalogGroup.ts
nf-s: From flicking through [ckan](https://github.com/ckan/ckan) code base - I can see the following values for `state`
- "active" 
- "deleted" 
- "draft"

As for `data_state` - we have the schema here https://github.com/datagovau/ckanext-datagovau/blob/966e045f8bfe0cc1453d9eaf6511e8c476d59c26/ckanext/datagovau/schemas/dataset.yaml#L91-L100

- "active" 
- "inactive" 

As for why data.gov.au has an additional property - I have no idea. All of the datasets we harvest through data.gov.au have `state = "active"`.

I have added some comments to clarify both properties. 

https://github.com/TerriaJS/terriajs/blob/9f9b414d1f7149eadcdbe39f8842378404a341de/lib/Models/Catalog/Ckan/CkanDefinitions.ts#L55-L62

I have changed logic to check for values other than `"active"`:

https://github.com/TerriaJS/terriajs/blob/0f6700a1eda14ea293a070b589cc65f35abe8e3d/lib/Models/Catalog/Ckan/CkanCatalogGroup.ts#L264-L276

staffordsmith83: Oh yeah that looks solid!

staffordsmith83 (approved):
Looks good man!

PR #6554, opened 2022-09-21
Add `.logError()` to all usage of `updateModelFromJson` where the `Result` object is ignored

nf-s:
### Add `.logError()` to all usage of `updateModelFromJson` where the `Result` object is ignored

Quite a few calls to `updateModelFromJson` were ignoring the `Result` object - which means if an error occurs then nothing is shown to the user **or** printed to console. This makes debugging harder

### Checklist


staffordsmith83 (commented):
Looks good. I noticed there is another occurence where error is not handled, in `ApiTableCustomComponent` - could probably be removed:
https://github.com/TerriaJS/terriajs/blob/e2cc3b6706377bd39ccecb561b130833a2a55187/lib/ReactViews/Custom/ApiTableCustomComponent.ts#L54-L59

staffordsmith83 (approved):

nf-s:
> Looks good. I noticed there is another occurence where error is not handled, in `ApiTableCustomComponent` - could probably be removed:
> 
> https://github.com/TerriaJS/terriajs/blob/e2cc3b6706377bd39ccecb561b130833a2a55187/lib/ReactViews/Custom/ApiTableCustomComponent.ts#L54-L59

Good pickup - thanks!

PR #6594, opened 2022-10-18
Mobx update attempts

### LINKED ISSUE #6588: Upgrade to MobX 6.6.2
As part of upgrading to React v18, we should really upgrade mobx from v4 to v6 major version.
The main reason is that we get build errors from the mobx4 dependency when we upgrade to React v18, as mobx v4 is using the now deprecated `RefForwardingComponent`

As part of this we should also upgrade packages `mobx-react` and `mobx-react-lite`

please see draft PR #6594

staffordsmith83:
### What this PR does
Fixes #6588

Update to MobX v6

Followed this guide:  
[https://mobx.js.org/migrating-from-4-or-5.html](https://mobx.js.org/migrating-from-4-or-5.html)

Got up to running this:  
`npx mobx-undecorate --keepDecorators`

Need to go in and manually update the constructor of some classes, wherever the codemod tool has inserted the following:  
`// TODO: [mobx-undecorate] verify the constructor arguments and the arguments of this automatically generated super call`

UPDATE:
The codemod tool `mobx-undecorate` adds constructors into classes so that it can call `makeObservable(this)`.
In many cases, e.g. `GeoJsonCatalogItem` this is a subclass that inherits from a parent class (`CreateModel`) which already has a constructor defined. Because we are overriding the parent class constructor, we need to specify the constructor arguments again, and the arguments for the `super()` call. Whilst in many cases is is sufficent to have `constructor(...args: any[])` and then `super(args)` in some cases this is not correct and the proper arguments can be tricky to figure out. It **may be possible** to just ass `makeObservable(this)` to the parent class! This is untested.

### Test me

### Checklist


zoran995:
Terriajs was upgraded to mobx v6 in #6687

PR #6605, opened 2022-10-25
Add mapbox GL viewier :trollface: 

nf-s:


nf-s:
Not going to happen.

Created PR in case people change their mind 😁

PR #6612, opened 2022-11-02
Update favicon dynamically if specified

### LINKED ISSUE #1196: UI for filtering CSV data by column
Add UI controls to the Now Viewing tab to allow the user to filter the data in a CSV file.  For example, if one column in the CSV file is `Year`, the UI might allow the user to only show the points/regions where the `Year` is within a specified range.  For numeric values, a double-sided slider is probably a good choice.  For enums, checkboxes like the ones we see with ABS datasets could work.

This is related to our [GovHack2014 project](http://2014.hackerspace.govhack.org/content/data-region-comparator).

### LINKED ISSUE #6613: Allow different Favicon for different maps
National map would like Australian Coat of Arms or the Australian Polygon:
https://github.com/TerriaJS/nationalmap/issues/1196

DTV would possibly like a Vic related logo: https://github.com/terriajs/vic-digital-twin/issues/398

These could be stored with map config.

staffordsmith83:
### What this PR does

UPDATE:
All functionality moved to `plugins-private/favicon-plugin`
This PR now just adds a property to `configParams` to allow plugin configuration.
Property is `pluginConf` Its a flexible object with any key value pairs.

Fixes #6613

Also related to:
https://github.com/TerriaJS/nationalmap/issues/1196
https://github.com/terriajs/vic-digital-twin/issues/398

Can now specify a `faviconPath` parameter in map config params.
This should be the path to a folder of favicons, with `favicon.ico` present at a minimum.
Initialisation code will see if this is specified, if so, will use favicon from specified folder.
To test, have put files [here](https://github.com/TerriaJS/saas-catalogs-public/tree/main/nationalmap/favicons), in which will get pushed to the same cloud storage bucket as the map config:
https://terria-catalogs-public.storage.googleapis.com/nationalmap/favicons

### Test me
NO LONGER RELEVANT, this functionality moved to plugin will be separate PR.
Go to:

http://ci.terria.io/dynamic-favicon/#configUrl=https://terria-catalogs-public.storage.googleapis.com/nationalmap/config/dev.json

Favicon should update during application load

### Checklist


lib/Models/Terria.ts
```diff
@@ -321,6 +321,11 @@ interface ConfigParameters {
 
   relatedMaps?: RelatedMap[];
 
+  /**
+   * Optional plugin configuration
+   */
+  pluginsConf?: Record<string, any>;
```
tephenavies: I think the name of this should be either `plugins` (it'll basically always be used as `terria.configParameters.plugins`) or `pluginsConfig`, as config is our standard abbreviation of configuration

tephenavies (changes requested):

lib/Models/Terria.ts
staffordsmith83: Given the usage as `terria.configParameters.plugins` I like `plugins`, update coming

staffordsmith83: Resolved in new commit

tephenavies (approved):

PR #6616, opened 2022-11-03
Fix netlify docs build

tephenavies:
Upgrade Python version to 3.8 to match new netlify build image.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,4 +1,8 @@
 [build]
   command = "yarn gulp docs"
   publish = "wwwroot/doc"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -1,4 +1,8 @@
 [build]
   command = "yarn gulp docs"
   publish = "wwwroot/doc"
-  environment = { NODE_VERSION = "16", NODE_OPTIONS = "--max_old_space_size=4096", NETLIFY_USE_YARN = "true", PYTHON_VERSION = "3.7" }
```
tephenavies, line 4: Only change is PYTHON_VERSION "3.7" to "3.8", necessary because the new ubuntu image neflify has upgraded us to doesn't have Python 3.7

nf-s (approved):

PR #6638, opened 2022-11-30
Modify GltfMixin to use stable instances of Cesium DataSource and Entity

na9da:
### What this PR does

* Modifies `GltfMixin` to use stable instances of `DataSource` and `Entity` objects instead of regenerating them each time the traits change. This should improve the performance of Gltf items without breaking any existing behaviours. It is also useful if child classes or other tools want a stable reference to these objects to perform other computations on the Model.
* Adds a new `transformationJson` property that returns the `position`, `rotation` and `scale` as an easy to use JSON object.

#### General changes
 There should be no observable change when loading gltf models [before](http://ci.terria.io/main/#share=s-telGt1G1A5dSIprCefEVudQDjSG) and [after](http://ci.terria.io/gltf-stable-resources/#share=s-telGt1G1A5dSIprCefEVudQDjSG) links.

#### Zoom to behaviour
- Try uploading a [sample gltf](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Duck/glTF-Binary/Duck.glb) to the [main branch](http://ci.terria.io/main/) and this branch.
- Notice that in main branch, the map zooms to empty space because the model position is unknown and assumed to be at the center of the earth. 
- In the [new branch](http://ci.terria.io/gltf-stable-resources/), notice that the zooming doesn't happen at all.

### Checklist


lib/ModelMixins/GltfMixin.ts
```diff
@@ -25,24 +25,54 @@ const Axis: Axis = require("terriajs-cesium/Source/Scene/Axis").default;
...
     @computed
-    private get cesiumUpAxis() {
+    get disableZoomTo() {
+      const { latitude, longitude, height } = this.origin;
+      return (
+        latitude === undefined ||
+        longitude === undefined ||
+        height === undefined
+      );
+    }
```
na9da, line 66: Disable zoom to when model position is not known.

na9da:
I have added one more change to this in the latest commit which disables zoom to when the model position is not known.

staffordsmith83 (commented):
~~deleted comment~~

staffordsmith83 (changes requested):
~~deleted comment~~

lib/ModelMixins/GltfMixin.ts
```diff
@@ -127,23 +170,33 @@ function GltfMixin<T extends Constructor<GltfModel>>(Base: T) {
       return super.shortReport;
     }
 
+    @computed
+    get modelEntity(): Entity {
+      const entity = this._modelEntity;
+      entity.position = new ConstantPositionProperty(this.cesiumPosition);
+      entity.orientation = new ConstantProperty(this.cesiumRotation);
+      entity.model = this.modelGraphics;
+      return entity;
```
staffordsmith83, line 181: Maybe need to trigger `notifyRepaintRequired` here?

lib/ModelMixins/GltfMixin.ts
```diff
@@ -25,24 +25,54 @@ const Axis: Axis = require("terriajs-cesium/Source/Scene/Axis").default;
...
+}
+
 function GltfMixin<T extends Constructor<GltfModel>>(Base: T) {
   abstract class GltfMixin extends ShadowMixin(
     CatalogMemberMixin(MappableMixin(Base))
   ) {
+    // Create stable instances of DataSource and Entity instead
+    // of generating a new one each time the traits change and mobx recomputes.
+    // This vastly improves the performance.
+    private readonly _dataSource = new CustomDataSource("glTF Model");
```
staffordsmith83, line 51: Maybe add a comment that `_dataSource` and `_modelEntity` should not be modified outside of their computeds, `private` is nice but could still modify properties...

staffordsmith83 (changes requested):
Nice one Nanda, couple of comments to resolve!

lib/ModelMixins/GltfMixin.ts
na9da: I think the `notifyRepaintRequired()` will automatically happen when `mobx` recomputes the `mapItems` triggering a chain of events that [reaches here](https://github.com/TerriaJS/terriajs/blob/6c44081460f857a44f2daf84a42e5bae0871a924/lib/Models/Cesium.ts#L743).

Also i think the general rule is that we shouldn't access any other Terria resources from within the Catalog items because they are like data providers and must be independent of the rest of the infra(?)

na9da: Thoughts @staffordsmith83?

staffordsmith83: "Also i think the general rule is that we shouldn't access any other Terria resources from within the Catalog items because they are like data providers and must be independent of the rest of the infra(?)" oh yeah thats critical I think. Looked through that chain of events right you are!

staffordsmith83 (approved):
This looks good Nanda!

AnaBelgun:
Update 22 Dec 2022: 
- ready to merge

Issue #5335, opened 2021-03-17
Leaflet rendering issue on multiple geojsons with same source 


zoran995:
When having multiple GeoJSON items with the same URL, there is a rendering issue in Leaflet that GeoJson will be visible on the map when hidden or removed. Sometimes it happens that it doesn't render anything when only one item is shown. Everything works correctly in Cesium viewer
Steps to reproduce: 
- open http://ci.terria.io/next/#clean&https://gist.githubusercontent.com/zoran995/a8f0df9fdd22551e59d3270986051576/raw/f29c43bb69b0eff2b311337797609d05063a6217/simple.json
- change to leaflet viewer
- hide both layers
- show first layer (nothing renders - if the order of removing was the first layer and the second layer)
- show the second layer - the first one is rendered on the map
- try hiding or removing layers (points are rendered no matter what you do)

![image](https://user-images.githubusercontent.com/19208948/111549343-4ea40a80-877c-11eb-973f-16b259ccd46c.png)
![image](https://user-images.githubusercontent.com/19208948/111549303-3e8c2b00-877c-11eb-8eda-8fbfef9fb012.png)

Haven't checked with other layer types. I was sure that something similar was happening with GeoJSON layers (and resolved) but can't find it anywhere _edit: I was thinking that this is similar to #4342 but it is not_

himby:
I can confirm similar behavior with multiple CZML datasources, it seems related to visualizers not being deleted correctly when disabling a catalog item. They seem to add up when enabling it again.

![image](https://user-images.githubusercontent.com/77962945/158369535-9795df92-33a0-4412-9447-5c727eaac1a1.png)

himby:
Still been struggling with this issue:
Using entities.id check instead of contains when adding datasources to leaflet and removing the raiseToTop re stacking seems to fix the problem. 

Maybe others can confirm my findings.

(Leaflet.ts)

![image](https://user-images.githubusercontent.com/77962945/189646012-ef27705f-d9ce-42bb-adc5-d87a8c1582b8.png)

nf-s:
This might be related to 

- https://github.com/TerriaJS/terriajs/issues/6399
- https://github.com/TerriaJS/terriajs/issues/6344

zoran995:
This is fixed as part of #6639

PR #6639, opened 2022-11-30
Fixes a timing bug in Cesium map which resulted in some DataSources not getting completely removed.

### LINKED ISSUE #6344: Datasources might not be added / removed correctly from Cesium and Leaflet maps in some rare cases
After Cesiums' switch to real `Promise` api from `when.js`, the `DataSourceCollection.add()` api is fully asynchronous all the time. Previously, when passing a concrete `DataSource` as argument, the `add()` used to happen instantly without any delay. But after the switch, `add()` is [always deferred](https://github.com/CesiumGS/cesium/blob/958e74bdab1e76767ff2ab88aef0b2b05f3c956e/Source/DataSources/DataSourceCollection.js#L90). 

This has some consequence to our code for adding datasources to cesium/leaflet maps. Our implementation (incorrectly) depended on this "instant" add effect to work correctly.

- The subsequent call to [raiseToTop()](https://github.com/TerriaJS/terriajs/blob/bd6198e66b9cf4314a95464acba8ab5d40516ef8/lib/Models/Cesium.ts#L578) might silently delete a datasource (because calling `raiseToTop` on an item that does not exist has an undefined behavior)
- Datasources might get addded multiple times to the `DataSourceCollection`

This is rare and difficult to repro by adding items from catalog, but might happen when programatically adding data sources.

na9da:
### What this PR does

Partially fixes https://github.com/TerriaJS/terriajs/issues/6344 (for Cesium map).

I tried the same fix for Leaflet and it only seems to partially fix it, which hints at an additional bug somewhere in LeafletDataSourceDisplay. I can re-visit Leaflet later when I have time.

### Test me

It is slightly hard to replicate since this is a timing bug. But you can try the following a few times to get it.
1. Open the [main branch CI](ci.terria.io/main/) - make sure you are in 3d mode
2. Add all of these items under the `CSV` group by pressing the `+` button
image
3. Once all the items are loaded, click the `Remove all` workbench button.
4. You should sometimes see some left over points on the map, like below:
    
image
5. Now repeat the same a few times for [this branch](http://ci.terria.io/fix-cesium-ds-remove/).
6. The bug should hopefully be gone.

### Checklist

- [ ] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ There already exists tests for this, but it is not triggered because it is a timing bug.


lib/Models/Cesium.ts
```diff
@@ -627,32 +627,59 @@ export default class Cesium extends GlobeOrMap {
...
+    // Ensure stacking order matches order in allDataSources - first item appears on top.
+    runInAction(() =>
+      availableDataSources.forEach((d) => dataSources.raiseToTop(d))
+    );
+  }
+
   private observeModelLayer() {
     let prevMapItems: MapItem[] = [];
-    return autorun(() => {
+    return autorun(async () => {
```
tephenavies: Please remove this async annotation

tephenavies (commented):
This is a good improvement. I think this will work in almost all cases, but I think it's possible to have edge cases if we get two calls of syncDataSourceCollection running at once.

This may be overkill, but I think it's possible to ensure that the datasource syncing doesn't run twice simultaneously by:
* Splitting the `autorun` into sync parts and a separate reaction for the datasource syncing
* ~Use a `scheduler` to ensure the reaction function doesn't run until the last instance finishes~ Note: below I've basically recreated the `scheduler` feature of `reaction`, because I think the intent is more clear that way
* Maybe even expose the all-datasources-added promise as a public property of TerriaViewer (or GlobeOrMap)?

Some code to show the above:
```
private observeModelLayer() {
  // Closure variables
  let prevMapItems = [];
  // Instead of closure variable, this variable might be on TerriaViewer 
  let allDatasourcesAdded = Promise.resolve();
  // Reaction clean up code omitted
  autorun(() => {
    // Existing autorun that deals with Primitives and ImageryLayers
  });
  reaction(() => this.availableDataSources, () => {
    allDatasourcesAdded = allDatasourcesAdded.then(
      () => this.syncDataSourceCollection(this.availableDataSources, this.dataSources)
    );
  });
  return cleanupFunc;
}
```

It's possible that calls of the syncDSC function will be delayed so they end up running multiple times, but I think it won't be able to run simultaneously to itself (which I think could cause problems).

tephenavies:
What do you think @na9da?

tephenavies:
Actually I think my suggestion has a problem that the allDatasourcesAdded promise is not set synchronously when a change to mapItems is made, so you'd still have to do e.g.:

```
// Manipulate workbench
workbench.add(item)
// Wait 0ms for reaction inside Cesium.ts to run
await Promise.resolve();
// Wait potentially for an async datasource
await terriaViewer.allDatasourcesAdded;
// Do something with cesium.dataSources (e.g. check for a dataset in a spec)
```

na9da:
Nice catch @steve9164. For a moment I forgot the root cause of the bug which is that the `.contains()` check and the[ final push() to the collection](https://github.com/TerriaJS/cesium/blob/160d6f108c8bf606fd608df04d375fec5c3f7cb7/Source/DataSources/DataSourceCollection.js#L90-L94) does not happen synchronously which leaves a gap for concurrent calls to slip through and do multiple inserts. 

If I understood correctly, in your solution, there is still a chance that different concurrent runs get the same closure value for the promise and so fire concurrently when the previous promise finishes? 

Another possible solution is to maintain a `Set<DataSource>` of active "syncs" as a guard against double inserts.
```
class Cesium {
   // 
   private activeDataSourceSyncs: Set<DataSource> = new Set();

   syncDataSourceCollection(...) {
        ...
        for (let ds of availableDataSources) {
           if (!dataSources.contains(ds) && !this.activeDataSourceSyncs.has(ds)) {
               try {
                 activeDataSourceSyncs.add(ds);
                 await dataSources.add(ds);
               } finally {
                 activeDataSourceSyncs.remove(ds);
               }
           }
        }
   }
}
```

Longer term solution would be to contribute an `addIfDoesNotContain()` or `addSync()` method upstream as it looks like this could be a common source of bugs. Otherwise, we could create a child class of `DataSourceCollection` to add them.

However, I would prefer something similar to your solution as it could ensure that the whole syncing thing happens only 1 at at time.

na9da:
Another solution:

- Keep a sync count with a maximum value of 2
- Inisde the sync() method, skip execution if the current sync count > 0 (meaning a sync is already running)
- When the current active run finishes, it decrements the sync count and re-runs sync() if the sync count is still `> 0` (with setTimeout so that we don't enter a spin loop)

The max value 2 is because even if we get 10s of re-runs we only need to sync with the latest `availableDataSources`.

na9da:
Hey @steve9164, I have implemented my 2nd solution of using a flag/count to skip syncing when another sync is already running. 

I think I have also figured out the problem with syncing for Leaflet. The `raiseToTop(ds)` call silently deletes the last item in the collection if `ds` does not exist. For Leaflet, when we set `show=false` for a workbench item, we are actually removing the DS from the collection. After that, when we do `raiseToTop(ds)` if we don't check for `contains(ds)` first, it will not only add `ds` but also remove the last element from the collection!

The latest commit should also hopefully fix #5335

tephenavies:
@na9da 
> If I understood correctly, in your solution, there is still a chance that different concurrent runs get the same closure value for the promise and so fire concurrently when the previous promise finishes?

I don't think that's true. In my snippet when `allDatasourcesAdded` is used, it's immediately updated with a new promise that won't resolve until the new async task is complete. If the reaction fires twice, execution 1 will update the promise and execution 2 will immediately attach to the new promise updated by 1 (which resolves after all async work in 1 is done).

na9da:
@steve9164, I have updated Cesium & Leaflet to use your solution and tested it. It looks more elegant now.

lib/Models/Leaflet.ts
```diff
@@ -403,37 +450,55 @@ export default class Leaflet extends GlobeOrMap {
...
+      if (!dataSources.contains(ds) && ds.show) {
+        await dataSources.add(ds);
+      }
+    }
 
-      // Ensure stacking order matches order in allDataSources - first item appears on top.
-      allDataSources.forEach((d) => dataSources.raiseToTop(d));
-    });
+    await new Promise((resolve) =>
+      window.setTimeout(resolve, Math.random() * 1000)
```
tephenavies: Is this a remnant of testing?

na9da: Ooops! removed.

tephenavies (approved):

PR #6640, opened 2022-11-30
AssimpCatalogItem fixes

na9da:
### What this PR does

Following fixes for `AssimpCatalogItem`:
- Correctly handle zipped files with a root directory, i.e individual files are all under a `somedir/*`. Previously assimp was not able to correctly locate these files when referenced relatively within the source package.
- Fix logic for finding unsupported textures. Previously, it used to fail for images with URLs like '//texture.png` which appears in some blender files.

##### a. ZIP files with a root dir
 
1. Open this [copy of main branch](http://ci.terria.io/main-with-assimp/) with assimp upload type enabled
2. Go to local file upload and choose the last option 
       
image

   3. Download this[ zip file](https://github.com/TerriaJS/terriajs/blob/assimp-fixes/wwwroot/test/AssImp/zip-test.zip) locally and upload it to the CI site. If you inspect the zip file you'll see that the assets are inside a root dir called `zip-tests/`
   4. Now open the console, you'll see an error like: 
image
   5. From the console run this code to manually position the object:
         ```
            viewState.terria.workbench.items[0].setTrait("user", "origin", {
        "latitude": -42.8826,
        "longitude": 147.3257,
        "height": 10
      }) 
         ```
   6. Now zoom to the object, you should see a box like below with the texture missing:
      
image

##### b. Texture image URLs starting with `//`
1. Follow the same steps for this [zip file](https://github.com/mrdoob/three.js/files/1352842/aircraft_blend_and_gltf.zip) containing a blender file. The blender file references the texture file as `//texture.png`
2. Open console to see a warning like: 
image
3. The texture loads correctly in spite of the warning

#### After: 
Repeat the same  tests for [this branch](http://ci.terria.io/assimp-fixes/).
 
### Checklist

- [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - Partial tests, I have just updated the existing `.zip` to have a root directory


lib/Core/getDataType.ts
```diff
@@ -169,6 +169,12 @@ export default function (): GetDataTypes {
         value: "shp",
         name: i18next.t("core.dataType.shp"),
         extensions: ["zip"]
+      },
+      {
+        value: "assimp",
+        name: i18next.t("core.dataType.assimp-local"),
+        description: i18next.t("core.dataType.assimp-local-description"),
+        extensions: ["zip"]
```
na9da, line 177: Will remove before merging.

lib/Models/Catalog/Gltf/AssImpCatalogItem.ts
```diff
@@ -197,22 +210,22 @@ export default class AssImpCatalogItem
         gltfJson.images?.forEach((image) => {
           if (!image.uri) return;
 
+          // Replace back slashes with forward slash
+          let newUrl = image.uri.replace(/\\/g, "/");
+          // Remove start "./" or "//" from uri
+          if (newUrl.startsWith("//") || newUrl.startsWith("./")) {
+            newUrl = newUrl.slice(2);
+          }
+
```
na9da, line 219: We were already doing this rewrite, but after the supported extensions check. This resulted in incorrect warning about unsupported extensions because `URI('\\texture.png')` treats `texture.png` as domain name instead of path.

lib/Core/getDataType.ts
mwu2018: Why will this part be removed?

mwu2018 (commented):
I have done the tests and they look good. Left one comment.

lib/Core/getDataType.ts
na9da: That's because assimp converted models are not geo-referenced and will be placed automatically at the center of the earth, which is not very useful.

You can still use it from a catalog and specify an `origin`, `rotation` and `scale` for the model.

mwu2018: Why is it here then?

mwu2018 (approved):
Good to merge after removing the new code in getDataType.ts.

lib/Core/getDataType.ts
na9da: So that we can test on the CI site. The bugs apparently manifests only when doing file-uploads not when loaded from a catalog.

na9da:
Thanks @mwu2018, for the reiview. I'll merge after updating CHANGES.md.

PR #6644, opened 2022-12-02
BoxDrawing improvements.

na9da:
### What this PR does

- Added option `drawNonUniformScaleGrips` to enable/disable uniform-scaling
- Set limit on the size of scaling grips
- Somewhat improve move interaction
- Restore modified map state when interaction stops

### Test me

The changes are hard to test from CI because the only place we use `BoxDrawing` is for 3d-tiles clipping box but that's not available on CI. You could however load up a tileset in your local, enable clipping box to verify these results.

#### Scale grip sizing
Keep zooming in and shrinking the box as small as you can. As the box gets smaller (less than the size of an average house) you can see that the scale grips become huge relative to the box size. With this PR the scale grips should be sized relative to the size of the whole box.
Before:
image
After:
image

#### Move interaction
Try moving the box from a viewing angle almost horizontal to the earth surface. Previously, the box wont move. This is because  we couldn't find an intersection point from the camera ray to the globe surface when the camera is parallel to the surface. We somewhat improve by falling back to finding an intersection with the ellipsoid instead of the terrain. This is better than doing nothing.

image

### Checklist

- [ ] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ Specs are not yet ready for `BoxDrawing`.

### DIFF
--- a/lib/Models/BoxDrawing.ts
+++ b/lib/Models/BoxDrawing.ts
@@ -123,10 +123,17 @@ type ScalePointStyle = {
  */
 type InteractionState =
   | { is: "none" }
-  | { is: "picked"; entity: Entity & Interactable }
+  | {
+      is: "picked";
+      entity: Entity & Interactable;
+      beforePickState: {
+        isFeaturePickingPaused: boolean;
+        enableInputs: boolean;
+      };
+    }
   | { is: "hovering"; entity: Entity & Interactable };
 
-type OnChangeParams = {
+export type BoxDrawingChangeParams = {
   /**
    * The modelMatrix of the box
    */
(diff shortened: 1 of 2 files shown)


lib/Models/BoxDrawing.ts
```diff
@@ -519,22 +537,31 @@ export default class BoxDrawing {
...
     };
 
     const handleRelease = (click: MouseClick) => {
       if (state.is === "picked") {
-        this.cesium.isFeaturePickingPaused = false;
-        scene.screenSpaceCameraController.enableInputs = true;
+        this.cesium.isFeaturePickingPaused =
+          state.beforePickState.isFeaturePickingPaused;
+        scene.screenSpaceCameraController.enableInputs =
+          state.beforePickState.enableInputs;
```
na9da, line 564: Restore previous map state (like feature picking) when a picked object is released instead of blindly setting it to true/false.

mwu2018 (approved):
- Please update CHANGES.md

For future improvement
- There are still some discrepancies in horizontal view angle where the clipping box scaling and moving speed suddenly becomes very fast, sometimes it get lost.
- If the clipping box is moved far away from the target, it is difficult to be moved back, especially in top viewing angle.

PR #6645, opened 2022-12-02
Add ActionBar.

na9da:
### What this PR does

- A new floating menu at the bottom of the map that looks like:
  image
  This will be used by the SceneEditor, but other plugins can also make use of it.
- Refactors some of the Portal code in Terria

### Test me

The `ActionBar` is currently not used by any feature in core Terria, so it cannot be tested from the CI. 

### Checklist


lib/Styled/Button.tsx
```diff
@@ -188,13 +188,21 @@ export const Button: React.FC<ButtonProps> = (props) => {
...
+          <IconSpan
             css={iconProps && iconProps.css}
+            margin={
+              // Apply left or right margin only when the button content is not empty
+              props.children
+                ? props.rightIcon
+                  ? "margin-left: 8px"
+                  : "margin-right: 8px"
+                : null
+            }
```
na9da, line 200: Previously, a margin used to be shown even when the contents where empty, this bloated the buttons that only had an Icon and no text. This fixes it.

lib/ReactViews/ActionBar/ActionBarPortal.tsx
```diff
@@ -0,0 +1,43 @@
...
+  left: 0;
+  right: 0;
+  margin: auto;
+
+  /* Animate slide in-out */
+  transition: all 0.2s;
+  transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+`;
+
+export default withViewState(ActionBarPortal);
```
tephenavies: Prefer use of `useViewState` and reduce passing viewState through props to reduce noise. Could you update StandardUserInterface/Portal to use `useViewState`? Then I think you won't need viewState here.

lib/ReactViews/Workflow/WorkflowPanelPortal.tsx
```diff
@@ -37,4 +34,4 @@ const Container = styled.div<{ show: boolean }>`
   opacity: ${(p) => (p.show ? 1 : 0)};
 `;
 
-export default withViewState(WorkflowPanelContainer);
+export default withViewState(WorkflowPanelPortal);
```
tephenavies: I think this `withViewState` also can be removed when Portal uses useViewState.

tephenavies (changes requested):

na9da:
Thanks for the feedback @steve9164! I have updated the code to replace explicit `viewState` props with `useViewState()` hook.

tephenavies (approved):
Thanks @na9da!

tephenavies:
Actually, should there be a changelog entry in this PR, or is it included in a different PR?

na9da:
Thanks @steve9164! 

> Actually, should there be a changelog entry in this PR, or is it included in a different PR?

I have updated the changelog in the latest commit. Kept it for last to avoid frequent merge conflicts.

PR #6646, opened 2022-12-02
Add MapToolbar.addToolButton() API.

na9da:
### What this PR does

Adds an easy to use API for creating buttons that launch Tool UIs. This is just a wrapper around existing functionality that simplifies the usage.
eg usage:
```
const toolButton = MapToolbar.addToolButton<SceneEditorProps>(
      viewState,
      () => ({
        text: pluginName,
        icon: iconSceneEditor,
        tool: {
          name: pluginName,
          component: () => import("./Views/App").then((m) => m.default)
        }
      })
    );
```

This adds the below button to the map navigation menu, that when clicked launches the tool.
image

### Checklist


tephenavies:
@steve9164 Check both methods of adding map toolbar buttons against this new proposed method.

lib/ViewModels/MapNavigation/MapToolbar.ts
```diff
@@ -278,3 +278,5 @@ class ModalButtonController extends MapNavigationItemController {
     super.deactivate();
   }
 }
+
+export { addToolButton, ToolButton, ToolButtonOptions } from "./ToolButton";
```
tephenavies: If you want one file to import from plugins, I think it'd be better to restructure to something like this:

```
MapToolbar/
├─ index.ts
├─ addButton.ts (ex MapToolbar.ts)
├─ ToolButton.ts
```

where index.ts only has `export { x } from "./y"` lines.

lib/ViewModels/MapNavigation/ToolButton.ts
```diff
@@ -0,0 +1,152 @@
...
+  const toolButton = {
+    removeButton: () => terria.mapNavigationModel.remove(id),
+    closeTool: () => controller?.closeTool(),
+    openTool: (props?: Partial<ToolProps>) => {
+      const navigationItem = terria.mapNavigationModel.findItem(id);
+      const controller = navigationItem?.controller;
+      if (controller instanceof ToolButtonController) {
+        controller.openTool(props);
+      }
+    }
```
tephenavies: This lack of symmetry between open and close methods is odd. Should openTool use the `controller` in the closure? If not, why not? If `openTool` uses a different controller, should `closeTool` then use that too?

lib/ViewModels/MapNavigation/ToolButton.ts
```diff
@@ -0,0 +1,152 @@
+import { action, computed } from "mobx";
```
tephenavies: I'm not sure what this file is for. Is it to do with MapNavigation buttons, or buttons that form a tool along the bottom of the screen?

Is there any clash with the 2 existing methods of creating MapNavigation items, and if so, should we deprecate one of them?

lib/ViewModels/MapNavigation/MapToolbar.ts
tephenavies: I'd prefer not to import -> export things, except in an index.ts file of a folder. If you think we should export everything from `lib/ViewModels/MapNavigation/MapToolbar`, we should use the following structure:
```
MapToolbar/
├── index.ts (just exports things from other files in the dir)
├── Button.ts (move MapToolbar.ts to this, or a different name, or split into smaller files)
└── ToolButton.ts
```

lib/ViewModels/MapNavigation/ToolButton.ts
na9da: The only reason I can think of now why I wrote it that way is a bit of defensive coding. If something else has removed the item from `mapNavigationModel` you still want `closeTool()` to work, so I use the closure value which will always exist. But, you want `openTool()` to fail if something else removed the button from `mapNavigationModel`, so I use the queried result. 

If it makes the code a bit simpler, I'm happy to make it symmetric and do a query in both cases (and maybe extract the query into a helper function)

lib/ViewModels/MapNavigation/ToolButton.ts
na9da: This file defines a convenience function for adding a button that is specifically for opening and closing a Tool, to the right-hand-side map menu. A [Tool](https://github.com/TerriaJS/terriajs/blob/main/lib/ReactViews/Tools/Tool.tsx#L30) is basically how a plugin injects UI code into the DOM.

We currently do have a way of adding Tool buttons which is used here:
https://github.com/TerriaJS/terriajs/blob/6c44081460f857a44f2daf84a42e5bae0871a924/lib/ReactViews/Map/Navigation/registerMapNavigations.tsx#L110-L125

The `MapNavigationModel` is quiet a powerful mechanism for extending the Map menu with different kinds of buttons using custom controllers. However I felt there is a need to expose a simpler API for plugins to easily add Tool buttons.

It gets called like this:

```
    const toolButton = MapToolbar.addToolButton<MyToolProps>(
      viewState,
      () => ({
        text: toolName,
        icon,
        tool: {
          name: toolName,
          component: () => import("./MyTool").then((m) => m.default)
        }
      })
    );
```

In the plugin work, I have been writing similar wrapper functions that change some internal state without having to expose those objects directly. Eg: [function for adding new data-types](https://github.com/TerriaJS/terriajs/blob/main/lib/ViewModels/UploadDataTypes.ts), [function for interacting with feature info panel](https://github.com/TerriaJS/terriajs/blob/6c44081460f857a44f2daf84a42e5bae0871a924/lib/ViewModels/FeatureInfoPanel.ts#L38), [function for extending viewing controls menu](https://github.com/TerriaJS/terriajs/blob/6c44081460f857a44f2daf84a42e5bae0871a924/lib/ViewModels/ViewingControlsMenu.ts#L28). Not exposing the internal objects means we can refactor them without breaking the external API.

Let me know if any of that makes sense. Also, in my mind I don't think of it as a set in stone thing but as an experimental interface that is evolving and we could change it if something feels broken.

lib/ViewModels/MapNavigation/MapToolbar.ts
na9da: Good point. 

Like I noted above, all of this is very experimental, so for eg, now that we have `addToolButton()` we can get rid of the `addModeButton()` function that `MapToolbar.ts` implements. I have to update a few plugins before we can remove it. I'll revisit this post xmas break and do another clean up.

na9da: I have made the changes here - https://github.com/TerriaJS/terriajs/pull/6646/commits/72090a02e39d93e398ce7bbf09db1ba763db441d

lib/ViewModels/MapNavigation/ToolButton.ts
na9da: I have made the changes here - https://github.com/TerriaJS/terriajs/pull/6646/commits/72090a02e39d93e398ce7bbf09db1ba763db441d

tephenavies: > The only reason I can think of now why I wrote it that way is a bit of defensive coding. If something else has removed the item from `mapNavigationModel` you still want `closeTool()` to work, so I use the closure value which will always exist. But, you want `openTool()` to fail if something else removed the button from `mapNavigationModel`, so I use the queried result.

Defensive coding is good, I think in this case a one line comment along the lines of:
```
// Access controller through mapNavigationModel to ensure the button hasn't been removed 
```

I think a comment similar to this would still be useful in `getController`.

mwu2018:
to discuss with @na9da

tephenavies (approved):

na9da:
Thanks for the review @steve9164. I have updated the code with the comment line.

PR #6647, opened 2022-12-02
Adds an API to extend workbench item controls from plugins.

na9da:
### What this PR does

This adds an API that can be used by plugins to extend the workbench with new input controls.

eg Usage:

```
    WorkbenchItem.Inputs.addInputviewState, (item) => {
      if (!GeoJsonMixin.isMixedInto(item)) {
        return;
      }

      // Return a control of type SelectableDimension
      return 
        {
          type: "button",
          text: "Count shapes",
          onClick: () => console.log(shapeCounts(item)),
          icon: countIcon
        }'
    })
```

This adds a button for all GeoJson items that counts and logs the number of primitives shapes to the console. 

### Checklist


lib/ViewModels/WorkbenchItemControls.ts
```diff
@@ -0,0 +1,27 @@
...
+import { SelectableDimension } from "../Models/SelectableDimensions/SelectableDimensions";
+import ViewState from "../ReactViewModels/ViewState";
+
+type WorkbenchItemControlsGenerator = (
+  item: BaseModel
+) => SelectableDimension | undefined;
+
+type WorkbenchItemControlDisposer = () => void;
+
+export const addControl = action(
```
staffordsmith83: Need jsdoc for this, to help intellisense be clear on what the generator function should return. When testing it was not obvious that my generator Fn had to return something like 
```
{
          type: "button",
          value: "Test Button",
          setDimensionValue: () => console.log('xxxxxxxx'),
          buttonStyle: "secondary"
        }
        
```

staffordsmith83: In fact, I find `setDimensionValue` to be not a clear description for what is essentially the function that gets called onClick for the button thats added... I suppose this is defined in SelectableDimensions and that is a core Terria concept. Perhaps it is just because Im not familiar with the Selectable Dimensions concept yet, but is there any way we could ovveride this, and when people use `WorkbenchItemControls.addControl` in their plugins they specify `onClick` function instead or something that is less specific to Terria in house nomenclature?

staffordsmith83: I also found that it was not immediately obvious that `addControl` would add a control for every item added to the workbench (subject to conditions of course). Perhaps this funciton could be renamed `addControlForItem` to make that clear?

lib/ViewModels/WorkbenchItemControls.ts
```diff
@@ -0,0 +1,27 @@
+import remove from "lodash-es/remove";
+import { action } from "mobx";
+import { BaseModel } from "../Models/Definition/Model";
+import { SelectableDimension } from "../Models/SelectableDimensions/SelectableDimensions";
+import ViewState from "../ReactViewModels/ViewState";
+
+type WorkbenchItemControlsGenerator = (
+  item: BaseModel
+) => SelectableDimension | undefined;
```
staffordsmith83: Could we extend `SelectableDimension` type here to make it more specific to what we want people to provide in the return of their generator function? See other comments for this file. This could help us to provide specific intellisense hints for what people should be providing for this object: 
```
{
          type: "button",
          value: "Test Button",
          setDimensionValue: () => console.log('xxxxxxxx'),
          buttonStyle: "primary"
        }

```

Actually this is probably not good practise, but givent he complex inheritance in `SelectableDimensions.ts` we need a guide for how users of the plugin API can use `addControl` . How could we make this clear for people given the wide variety of controls that they could add?

staffordsmith83 (changes requested):
Is it normal to have both a `WorkbenchItemControls.tsx` and `.ts` file? Is there a benefit for that pattern that outweights the potential confusion? Otherwise, this looks great!

na9da:
Hey @staffordsmith83, thanks for the very valuable feedback pointing out the pain points in using this API. 

When I initially wrote this API, I was trying to stick to existing concepts like the  `SelectableDimension`. 

I think the name `SelectableDimension` came from the original use-case which was for selecting some [dimension](https://www.mapserver.org/ogc/wms_dimension.html) of a WMS or similar catalog item. But the declarative input mechanism was widely useful so its scope broadened but the name stuck on. Ideally at some point we should rename it to something like `TerriaInput` that can be considered wherever we have the pattern of extensible collection of inputs.

To address your points, if we consider what this API is actually trying to do - it is trying to add a new type of input for a workbench item. Maybe we could structure it as `WorkbenchItem.Inputs.addButton()`. Where the `addButton()` expects a more predictable type that takes an `onClick` handler instead of a `setDimensionValue`.

The reason for structuring it as `WorkbenchItem.Inputs` is because we have another related module called [ViewingControlsMenu](https://github.com/TerriaJS/terriajs/blob/main/lib/ViewModels/ViewingControlsMenu.ts) which we can fold under `WorkbenchItem.Menu`.

na9da:
Hey @staffordsmith83 , I have refactored the code based on your feedback.

The function for adding input has been changed to `WorkbenchItem.Inputs.addInput()` which accepts an input generator (only buttons for now, but can be easily extended to support other SelectableDimension types).

Let me know whether you think this is a better approach.

staffordsmith83 (approved):
Hey Nanda, I like this much better, exactly what I was hoping for, and can easily be extended. Approved!

staffordsmith83:
I tested with this plugin code: 
```
import {
  TerriaPlugin,
  TerriaPluginContext,
} from "terriajs-plugin-api";

import { addInput } from "terriajs/lib/ViewModels/WorkbenchItem/Inputs";

const plugin: TerriaPlugin = {
  name: "MODIFIED",
  description: "MODIFIED",
  version: "0.0.1",
  register({ viewState, terria, workbench }: TerriaPluginContext) {
    console.log("Registering Plugin")
   
    addInput(viewState, (item) => {
      return {
          type: "button",
          text: "Test Button",
          onClick: () => console.log('xxxxxxxx'),
          icon: "secondary"
        }
    })
  }
};


export default plugin;
```

PR #6664, opened 2022-12-15
6503 upload terria web data

### LINKED ISSUE #6503: Handle Terria catalog JSON in Upload > Web data
Currently we can add a local terria catalog file from `Upload > Local data` option. 

It should be possible to do the same for remote terria catalog files via `Upload > Web data`. This is so that the catalog file becomes part of the share URL.

sixlighthouses:
### What this PR does

Fixes #6503

Add ability to upload Terrria catalog json from url

### Test me

Add the following via web data upload 

https://gist.githubusercontent.com/na9da/ed0c4495ccfbfa76a17505505d06caa2/raw/b99d1f81703ffc927d302e2f3a2cf2fd0bbf93b7/delta.json


lib/ReactViews/ExplorerWindow/Tabs.jsx
```diff
@@ -36,6 +36,10 @@ const Tabs = observer(
       this.props.viewState.myDataIsUploadView = false;
     },
 
+    async onUrlAddFinished() {
+      await this.props.viewState.viewCatalogMember();
```
na9da: Although this call doesn't err out, I think the expectation is we pass a catalog member to [viewCatalogMember](https://github.com/TerriaJS/terriajs/blob/c465cf1288316f5b82043db8e9f33f634c31eb87/lib/ReactViewModels/ViewState.ts#L590) as a parameter and the method will focus the catalog member in the explorer view. 

What we could do instead is call [openAddData()](https://github.com/TerriaJS/terriajs/blob/main/lib/ReactViewModels/ViewState.ts#L540) which will just open the Data Catalogue.

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
```diff
@@ -117,6 +119,22 @@ const AddData = createReactClass({
...
+      promise = loadJson(this.state.remoteUrl).then((data) => {
+        if (data.error) {
+          return Promise.reject(data.error);
+        }
+        this.props.terria.catalog.group
+          .addMembersFromJson(CommonStrata.user, data.catalog)
+          .raiseError(this.props.terria, "Failed to load catalog from file");
+
+        return Promise.resolve();
+      });
```
na9da: It might be good if the `raiseError()` can cover `loadJson()` call so that the user gets an error pop-up if the URL failed to load. One common scenario is user copies a URL that cannot be loaded due to CORS error. It will be good to show some feedback when that happens.

lib/ReactViews/RelatedMaps/related-maps.scss
```diff
@@ -1,7 +1,13 @@
...
+
 .dropdown-inner {
-  width: 600px;
   padding: 10px;
   font-size: 0.9rem;
+  width: 600px;
+
+  @media (max-width: $mobile) {
+    max-width: fit-content;
+  }
```
na9da: I had a look but couldn't figure out the change, a screenshot would be good.

na9da (commented):
@sixlighthouses - great work navigating the messy file handling code. I have made a couple of comments, let me know what you think. Also, please update the CHANGES.md when done.

lib/ReactViews/RelatedMaps/related-maps.scss
sixlighthouses: This was a git fail on my behalf this commit relates to this issue https://github.com/TerriaJS/terriajs/issues/6633, when I created the branch I branched off the feature branch not main

lib/ReactViews/ExplorerWindow/Tabs.jsx
sixlighthouses: updated, this is a nicer solution

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/AddData.jsx
na9da: That could work, but I think a simpler option might be something like:

```js
      promise = loadJson(this.state.remoteUrl)
        .then((data) => {
          if (data.error) {
            return Promise.reject(data.error);
          }
          this.props.terria.catalog.group
            .addMembersFromJson(CommonStrata.user, data.catalog)
            .raiseError(this.props.terria, "Failed to load catalog from file");
        })
        .then(() => {
          this.props.onUrlAddFinished();
        })
        .catch((error) =>
          TerriaError.from(error).raiseError(
            this.props.terria,
            `An error occurred trying to add data from URL: ${this.state.remoteUrl}`
          )
        )
        .finally(() => {
          this.setState({
            isLoading: false
          });
        });

```

which is basically `catch`ing the `loadJson()` response and raising the error to user. We will be showing two separate errors depending on whether it is a URL failure or `addMembersFromJson` failure, but that's a good thing, I think.

Also note that, I moved the `onUrlAddFinished()` to the `then()` so that the user stays in the current upload view if an error occurred. We could also maybe use `await` style to rewrite my block of code and simplify it.

Thoughts @sixlighthouses ?

sixlighthouses: @na9da , this is a nice solution -- TerriaError was what I needed but couldn't find.  I think the .then .catch style is fine here as the rest of the file uses it and I assume it may all get a tidy up when it gets TSified

na9da (approved):
Looks good @sixlighthouses. I'll leave it to you to merge it in.

PR #6672, opened 2023-01-06
6610 ga share url

### LINKED ISSUE #6610: Layer names are not recorded in Google Analytics when user enters via Share URL
When a user goes to a Terria instance and adds a layer using the catalogue we record an Event in Google Analytics. However, it appears that we do not record an event for the specific layer(s) when they view a layer using a Share URL (we do however record it when using a Story URL). We do record the Share URL itself which is good and potentially helpful in certain situations, but we should also record an event for the layer.

We have been asked by a stakeholder if we record an event when they come in using a Share URL. We will need to fix this.

What we want is for a layer that has been shown to a user using a Share URL to be recorded as a Google Analytics event, so we can find it in the list of layers being accessed in Google Analytics, and accurately report on it, e.g. the Transmission Lines in NationalMap:

Screen Shot 2022-11-01 at 17 02 42

When you go to NatMap and navigate the catalogue and add it through the catalogue (you can do it from a share to the metadata page in the catalogue, just click Add To The Map here: https://nationalmap.gov.au/#share=s-9el4phg2QZzUfbkBdTUHCm4vIfC), we record this Google Analytics Event:

Screen Shot 2022-11-01 at 17 55 41

That is:
> Event Category: Data Source
> Event Action: Add from preview button
> Event Label: National Datasets/Energy/Electricity Infrastructure/Transmission/Electricity Transmission Lines

For the purposes of telling our stakeholders how often their layer has been shown, it's the `Event Label` that is key to this, and what we need to also record when the user comes in through a Share URL.

To solve this issue, a Google Analytics Event should be recorded when you visit this Share URL that has the Transmission Lines layer added: https://nationalmap.gov.au/#share=s-l7HlrvjWRTWqLgOY67Q1NkK4wm6

Currently we record an Event with an `Event Label` set to the Share URL only.

The event details should likely be along the lines of:
> Event Category: Data Source
> Event Action: Add from share URL
> Event Label: National Datasets/Energy/Electricity Infrastructure/Transmission/Electricity Transmission Lines

sixlighthouses:
### What this PR does

Fixes #6610 

A layer that is shown via a Share URL is logged to a GA event

### Test me

Open the dev tools on your browser and use the following URL

http://ci.terria.io/6610-ga-share-url/#configUrl=https%3A%2F%2Fga-testing-bw.storage.googleapis.com%2Fconfig.json&share=s-9AxYHpzxyt4HZ07zqrwdvL7iGHW

Observe the events in the console

Pasted image 20230105193016


lib/Models/Terria.ts
```diff
@@ -2165,6 +2178,7 @@ async function interpretStartData(
         // Note startData.initSources can be an initUrl (string) or initData (InitDataSource/JsonObject)
         // Terria.initSources are different to startData.initSources
         // They need to be transformed into appropriate `InitSource`
+
```
tephenavies:
```suggestion
```

Remove empty line.

lib/Core/AnalyticEvents/analyticEvents.ts
```diff
@@ -51,7 +51,8 @@ export enum DataSourceAction {
   removeFromPreviewButton = "Remove from preview button",
   removeFromWorkbench = "Remove from workbench",
   removeAllFromWorkbench = "Remove all from workbench",
-  addFromDragAndDrop = "Add from Drag and Drop"
+  addFromDragAndDrop = "Add from Drag and Drop",
+  addFromShareUrl = "Add from share URL"
```
tephenavies: This event is fired when a dataset is added from share URL, but also when it's added from an initialisation file. Maybe this action should be "Add from share or init source"

tephenavies (changes requested):
This is great. There's just a couple of things to fix

lib/Core/AnalyticEvents/analyticEvents.ts
tephenavies: And change `addFromShareUrl` to `addFromShareOrInit`

tephenavies (approved):

PR #6675, opened 2023-01-11
Cesium token error handling

### LINKED ISSUE #4739: Show error message if Cesium Ion access token is invalid

staffordsmith83:
### What this PR does

Fixes #4739 

Adds similar error handling to that implemented in `CesiumTerrainCatalogItem` to catch errors when first trying to load any Cesium Terrain.

Also adapts code in `Cesium.ts` that appeared to be from v7 but was commented out.

### Test me

Load the [branch with a config.json that specifies an incorrect `cesiumIonAccessToken`](http://ci.terria.io/cesium-token-error-handling/#clean&configUrl=https://gist.githubusercontent.com/staffordsmith83/9400ab1111c6b714275b5a7a2c8e2c97/raw/55287e70f422a7c3baecdcd66627ba2d681e5ade/badCesiumKeyConfig.json). Switch to 3D Terrain mode (if not already). The map should automatically switch back to 3D Smooth mode and display an error.

Load the [branch with a config.json that specifies an incorrect `cesiumIonAccessToken` and a valid `cesiumTerrainAssetId`](http://ci.terria.io/cesium-token-error-handling/#clean&configUrl=https://gist.githubusercontent.com/staffordsmith83/a722991a75d37d26a1616410cba72677/raw/50b1936ae402d2a9d6bda5f02341c4312f2c80b5/badCesiumIonKeyCustomTerrain.json). Switch to 3D Terrain mode (if not already). The map should automatically switch back to 3D Smooth mode and display an error.

Load the [branch with a config.json that specifies a valid `cesiumIonAccessToken` and an **invalid `cesiumTerrainUrl`** ](http://ci.terria.io/cesium-token-error-handling/#clean&configUrl=https://gist.githubusercontent.com/staffordsmith83/c9eb19481ce3516c6d26b6fe15fd3524/raw/289d6b51775f905b9520b1d254ca80ac8a74ea88/badCesiumTerrainUrl.json)
This should also throw an error - I have included in this PR as it seems like the best place to handle this edge case.
See https://github.com/TerriaJS/terriajs/blob/ead4eac60a0d1c0eb431f30567f41be1be0ffe9b/lib/Models/Cesium.ts#L430-L435

### Checklist

### DIFF
--- a/lib/Models/Cesium.ts
+++ b/lib/Models/Cesium.ts
@@ -87,6 +87,7 @@ import TerriaFeature from "./Feature/Feature";
 import GlobeOrMap from "./GlobeOrMap";
 import Terria from "./Terria";
 import UserDrawing from "./UserDrawing";
+import { setViewerMode } from "./ViewerMode";
 
 //import Cesium3DTilesInspector from "terriajs-cesium/Source/Widgets/Cesium3DTilesInspector/Cesium3DTilesInspector";
 
@@ -149,6 +150,7 @@ export default class Cesium extends GlobeOrMap {
     });
   });
 
+  private _terrainMessageViewed: boolean = false;
   constructor(terriaViewer: TerriaViewer, container: string | HTMLElement) {
     super();
 
@@ -229,26 +231,6 @@ export default class Cesium extends GlobeOrMap {
     (<any>this.scene).highDynamicRange = false;
 
     this.scene.imageryLayers.removeAll();
-
-    //catch Cesium terrain provider down and switch to Ellipsoid
-    //     terrainProvider.errorEvent.addEventListener(function(err) {
-    //         console.log('Terrain provider error.  ', err.message);
-    //         if (viewer.scene.terrainProvider instanceof CesiumTerrainProvider) {
-    //             console.log('Switching to EllipsoidTerrainProvider.');
-    //             that.terria.viewerMode = ViewerMode.CesiumEllipsoid;
-    //             if (!defined(that.TerrainMessageViewed)) {
-    //                 that.terria.raiseErrorToUser({
-    //                     title : 'Terrain Server Not Responding',
-    //                     message : '\
-    // The terrain server is not responding at the moment.  You can still use all the features of '+that.terria.appName+' \
-    // but there will be no terrain detail in 3D mode.  We\'re sorry for the inconvenience.  Please try \
-    // again later and the terrain server should be responding as expected.  If the issue persists, please contact \
-    // us via email at '+that.terria.supportEmail+'.'
-    //                 });
-    //                 that.TerrainMessageViewed = true;
-    //             }
-    //         }
-    //     });
 
     this.updateCredits(container);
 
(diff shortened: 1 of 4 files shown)


lib/Models/Cesium.ts
```diff
@@ -428,6 +410,59 @@ export default class Cesium extends GlobeOrMap {
...
+              message:
+                "\
+The terrain server is not responding at the moment.  You can still use all the features of " +
+                this.terria.appName +
+                " \
+but there will be no terrain detail in 3D mode.  We're sorry for the inconvenience.  Please try \
+again later and the terrain server should be responding as expected.  If the issue persists, please contact \
+us via email at " +
+                this.terria.supportEmail +
+                "."
```
na9da: Might be good to move these strings to the translation file? 

Example:
https://github.com/TerriaJS/terriajs/blob/8a051a7845c6625c53bc50e76920060ccdf22d35/lib/Models/Cesium.ts#L583-L588

test/Models/CesiumSpec.ts
```diff
@@ -296,6 +296,120 @@ describeIfSupported("Cesium Model", function () {
...
+        terria2.workbench.removeAll();
+        // Set an invalid token for the test
+        terria2.configParameters.cesiumIonAccessToken = "expired_token";
+      });
+      // Instantiate Cesium object with the invalid token
+      cesium2 = new Cesium(terriaViewer2, container2);
+
+      await cesium2.terrainProvider.readyPromise.catch(() => {});
+      // Wait a few ticks to allow for delay in adding event listener to terrainProvider in Cesium.ts
+      await runLater(() => {}, 5);
```
na9da, line 335: Can we get rid of this if we pass `fireImmediately: true` as an option to the [autorun](https://github.com/TerriaJS/terriajs/blob/8a051a7845c6625c53bc50e76920060ccdf22d35/lib/Models/Cesium.ts#L391-L402) call? 

It is better not to use delays in test where we can to avoid slowing down tests.

test/Models/CesiumSpec.ts
```diff
@@ -296,6 +296,120 @@ describeIfSupported("Cesium Model", function () {
...
+    // declare different variables here, we need new instances to test
+    let terria2: Terria;
+    let terriaViewer2: TerriaViewer;
+    let container2: HTMLElement;
+    let cesium2: Cesium;
+
+    beforeEach(function () {
+      terria2 = new Terria({
+        baseUrl: "./"
+      });
```
na9da, line 310: Any reason not to use `terria`, `terriaViewer` and other top level instances?

test/Models/CesiumSpec.ts
```diff
@@ -296,6 +296,120 @@ describeIfSupported("Cesium Model", function () {
...
+    });
+
+    afterEach(function () {
+      cesium2?.destroy();
+      document.body.removeChild(container2);
+    });
+
+    it("should thow an warning when cesiumIonAccessToken is invalid", async function () {
+      runInAction(() => {
+        terria2.workbench.removeAll();
```
na9da: Is the `removeAll` necessary?

na9da (changes requested):
Looks great and good effort on the tests @staffordsmith83 

I have left a few comments for you to review.

test/Models/CesiumSpec.ts
tephenavies: The problem is that the token error handler is added to the terrain provider after `readyPromise` resolves, but somehow awaiting the readyPromise in this test resolves first. The proper fix would be to expose a promise within the Cesium class for the terrain provider ([fromPromise](https://github.com/mobxjs/mobx-utils/tree/5.6.2#frompromise) could be useful for this).

staffordsmith83: Yes thats the same conclusion @na9da and I got to yesterday. Shall I go ahead and implement "expose a promise within the Cesium class for the terrain provider" do you think @steve9164 ?

lib/Models/Cesium.ts
staffordsmith83: Done

test/Models/CesiumSpec.ts
staffordsmith83: The `fireImmediately: true` option doesnt work in this case

staffordsmith83: I am making a separate ticket for "The proper fix would be to expose a promise within the Cesium class for the terrain provider ([fromPromise](https://github.com/mobxjs/mobx-utils/tree/5.6.2#frompromise) could be useful for this)."

staffordsmith83: Here: https://github.com/TerriaJS/terriajs/issues/6690

test/Models/CesiumSpec.ts
staffordsmith83: No, not necessary, blindly copied from other tests pattern. I have removed now.

test/Models/CesiumSpec.ts
staffordsmith83: Yes, the reason is that we need to set some incorrect config parameters and then instatiate those top level instances. To simulate having a bad `cesiumionAccessToken` etc.

I have tried going back and using the top level instances, but it is not a neat test - and I cannot get the tests to reliably fail because there is a pretty complex initialisation sequence in Terria.. I prefer to have these instances we are running the tests on sandboxed to be sure. Does that sound OK @na9da ? Otherwise, with more familiarity with the initialisation sequence of terria there might be a way to operate on those top level instances.

na9da (approved):
The changes look good to me!

PR #6678, opened 2023-01-13
6648 share compare

### LINKED ISSUE #6648: Key Share URL use case broken: share URL for Compare datasets forgets 2nd date
A key Share URL use case has a bug: when viewing a time-based dataset such as satellite imagery, clicking Compare to get a `(copy)` of the dataset to set a second date to compare with *and moving the `(copy)` dataset below the original in the workbench*, when creating a Share URL and going to it, the `(copy)` dataset has been forgotten and instead we only have two instances of the original dataset with the same date - the 2nd date has been forgotten.

Steps to reproduce:
1. Load a time-based dataset like Sentinel and set some date: https://nationalmap.gov.au/#share=s-imLZhQIEKV3RSPCoXzeiRW6qStR
2. Click the 3 dots, then Compare to get a `(copy)` instance of the dataset. Set some other date for this dataset.
3. Move the `(copy)` dataset below the original in the workbench.
4. Get a Share URL, and load it in a new tab.

Expected:
The loaded Share URL has the original dataset with its date set, and the new `(copy)` dataset with its different date.

Actual:
There are two identical instances of the first dataset, there is no `(copy)` dataset, and the 2nd date has been forgotten.

Expected:
expected

Actual:
actual

sixlighthouses:
### What this PR does

Fixes #6648 
Added a check in SplitItemReference to see if we have a previousTarget [BaseModel] that has the same id as its source model.  If it does return it

### Test me

The following link was created using the steps outlined in the ticket

[Test Link](http://ci.terria.io/6648-share-compare/#configUrl=https%3A%2F%2Fterria-catalogs-public.storage.googleapis.com%2Fde-australia%2Fmap-config%2Fprod.json&share=s-nfi6H48GNn3W1lTBGIJLnlel7Il)

Required result with the 2nd dataset being the Copy with the changed date

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 #### next release (8.2.24)
 
+- Fix `Key Share URL use case broken: share URL for Compare datasets forgets 2nd date`
 - [The next improvement]
 
 #### 8.2.23 - 2023-01-06
--- a/lib/Models/Catalog/CatalogReferences/SplitItemReference.ts
+++ b/lib/Models/Catalog/CatalogReferences/SplitItemReference.ts
@@ -50,10 +50,13 @@ export default class SplitItemReference extends ReferenceMixin(
       });
     }
 
-
     try {
-      // Ensure the target we create is a concrete item
-      if (previousTarget !== undefined && sourceItem.uniqueId === this.splitSourceItemId) return previousTarget;
+      // Ensure the target we create is a concrete item and check if we can return previousTarget
+      if (
+        previousTarget !== undefined &&
+        sourceItem.uniqueId === this.splitSourceItemId
+      )
+        return previousTarget;
       sourceItem = getDereferencedIfExists(sourceItem);
       return sourceItem.duplicateModel(this.uniqueId, this);
     } catch (e) {
--- a/lib/Models/Catalog/CatalogReferences/SplitItemReference.ts
+++ b/lib/Models/Catalog/CatalogReferences/SplitItemReference.ts
@@ -50,8 +50,10 @@ export default class SplitItemReference extends ReferenceMixin(
       });
     }
 
+
     try {
       // Ensure the target we create is a concrete item
+      if (previousTarget !== undefined && sourceItem.uniqueId === this.splitSourceItemId) return previousTarget;
       sourceItem = getDereferencedIfExists(sourceItem);
       return sourceItem.duplicateModel(this.uniqueId, this);
     } catch (e) {


lib/Models/Catalog/CatalogReferences/SplitItemReference.ts
```diff
@@ -51,7 +51,12 @@ export default class SplitItemReference extends ReferenceMixin(
     }
 
     try {
-      // Ensure the target we create is a concrete item
+      // Ensure the target we create is a concrete item and check if we can return previousTarget
+      if (
+        previousTarget !== undefined &&
+        sourceItem.uniqueId === this.splitSourceItemId
```
tephenavies, line 57: What does this check guarantee? 

Also, it's probably not a standard use-case of SplitItemReference, but if `this.splitSourceItemId` changes between calls to `loadReference` this change will return the old model.

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 #### next release (8.2.24)
 
+- Fix `Key Share URL use case broken: share URL for Compare datasets forgets 2nd date`
```
tephenavies, line 5: It's better to note the bug that was fixed rather than a specific case of it (although it can be good list some specific use cases in a sub-list under the main bug fix entry). Sorry, I don't have a suggestion of a better changelog entry until I've understood the code change in SplitItemReference.ts.

tephenavies (changes requested):

lib/Models/Catalog/CatalogReferences/SplitItemReference.ts
sixlighthouses: This check is needed to prevent the second item (dragged from first to second in workbench) from stripping out the required dereferenced details, which is the cause of the bug.  i.e.  if there is a previousTarget passed in and the id's match return the previousTarget

AnaBelgun:
@steve9164 to talk to @sixlighthouses to clarify the approach

tephenavies:
@na9da to investigate solution to write more documentation and deal with some corner cases (potentially including id of reference changing).

na9da:
@steve9164  - in the init cycle we call `loadReference()` in two places - [1 here](https://github.com/TerriaJS/terriajs/blob/92da14a5b56fb54a62209f64753bbc8989ae0f84/lib/Models/Terria.ts#L1467-L1482), and then [2 here](https://github.com/TerriaJS/terriajs/blob/92da14a5b56fb54a62209f64753bbc8989ae0f84/lib/Models/Terria.ts#L1529). The first call will definitely trigger `forceLoadReference()`. The second call will trigger `forceLoadReference()` only if the copy item comes after the source item in the workbench (because mobx thinks the first item has changed so it triggers a recompute).

After the first call, we have a bit of code that explicitly sets the user-stratum for the `target` item of the reference. This ensures that the copy item has the correct date and whatever changes user made to it:
https://github.com/TerriaJS/terriajs/blob/92da14a5b56fb54a62209f64753bbc8989ae0f84/lib/Models/Terria.ts#L1472-L1482

However, if the 2nd call to `loadReference()` results in a `forceLoadReference()` call, then it will again create a new `target`. But after this new `target` is created, the user-stratum is not updated which causes the bug.

What the change in this PR does is to avoid creating a new `target` if the source item has not changed, which fixes it for `SplitItemReference`

~But I think this bug might also affect other references (?).~ Maybe not, because the other references do not refer to a workbench item.

What are other possible solutions here? Do we avoid the second call to `loadReference()`?

PR #6687, opened 2023-02-02
Upgrade to TS4 and mobx6 (Traits interface solution)

na9da:
### What this PR does

Traits interface solution for #5219 #6588 

### Test me

### Checklist


tephenavies:
Let's go ahead with this solution. Overriding traits in mixins (or classes) should be discouraged (use computed strata where possible), and this solution works for places where it can't be avoided.

BTW why are there no changes needed in DiscretelyTimeVaryingTraits? Is it because the interface is only needed if properties get overridden twice?

Next steps are to:
* debug the type error on node build (we need to allow some way of skipping the node build on `yarn install` so it doesn't error before saving the yarn.lock file that would help us debug problems with the node build)
* Write this solution as an ADR
* Write the other solution as a rejected ADR, and note that the implementation could be used in future
* Decide what version number to use for terriajs release (ts & mobx upgrades will mean changes to DTs, does this warrant a v9 release? We don't follow semver, because terriajs doesn't have a nicely exported interface yet, so maybe just a minor version?)
* ?

tephenavies:
- [x] @steve9164 to do a more thorough review (incl. React component method binding)

lib/ModelMixins/CatalogFunctionJobMixin.ts
```diff
@@ -111,11 +117,13 @@ function CatalogFunctionJobMixin<
   T extends Constructor<CatalogFunctionJobMixin>
 >(Base: T) {
   abstract class CatalogFunctionJobMixin extends GroupMixin(
-    AutoRefreshingMixin(CatalogMemberMixin(Base))
+    AutoRefreshingMixin(MappableMixin(CatalogMemberMixin(Base)))
```
steve9164, line 120: I'm not sure if this class should have MappableMixin

lib/Models/Definition/addModelStrataView.ts
```diff
@@ -49,3 +49,12 @@ export default function addModelStrataView<
 
   return model;
 }
+
+function decorate(
+  target: any,
+  decorators: { [id: string]: PropertyDecorator }
+) {
```
tephenavies, line 65: Could you add a comment why we shouldn't use a mobx function for this?

lib/Models/Definition/createCombinedModel.ts
```diff
@@ -234,6 +236,15 @@ function createCombinedStratum<T extends ModelTraits>(
   return <StratumFromTraits<T>>(<unknown>result);
 }
 
+function decorate(
+  target: any,
+  decorators: { [id: string]: PropertyDecorator }
+) {
```
tephenavies, line 243: Also should have a comment

lib/ReactViews/FeatureInfo/FeatureInfoSection.tsx
```diff
@@ -472,7 +485,9 @@ export class FeatureInfoSection extends React.Component<FeatureInfoProps> {
                 </>
               ) : (
                 // Show templated feature info
-                this.templatedFeatureInfoReactNode
+                // Calling toJS() because when passing a parsed node to react,
```
tephenavies: Should use observable.ref instead. I've made the change in https://github.com/TerriaJS/terriajs/pull/6687/commits/b60040e9b7146c3e71a5843e1d2c4f82df954960

lib/ReactViews/Search/SearchBox.jsx
```diff
@@ -50,6 +50,9 @@ export const SearchBox = createReactClass({
     theme: PropTypes.object
   },
 
+  /**
```
tephenavies, line 53: Does this file need updating from createReactClass?

lib/ViewModels/CompositeBar/CompositeBarItemController.ts
```diff
@@ -116,14 +116,14 @@ export abstract class CompositeBarItemController
...
    */
-  @computed
   get visible(): boolean {
     return this._visible;
   }
 
-  @action
   setVisible(v: boolean) {
-    this._visible = v;
+    runInAction(() => {
```
tephenavies, line 124: Why change this to use `runInAction`?

lib/ViewModels/MapNavigation/MapNavigationItemController.ts
```diff
@@ -12,22 +12,31 @@ export interface IMapNavigationItemController
 }
 
 export default abstract class MapNavigationItemController extends CompositeBarItemController {
+  constructor() {
+    // TODO: [mobx-undecorate] verify the constructor arguments and the arguments of this automatically generated super call
```
tephenavies: Remove this comment

test/ModelMixins/DiffableMixinSpec.ts
```diff
@@ -39,8 +41,17 @@ describe("DiffableMixin", function () {
 });
 
 class TestDiffableItem extends DiffableMixin(
-  CreateModel(mixTraits(DiffableTraits, CatalogMemberTraits, SplitterTraits))
+  MappableMixin(
+    CreateModel(mixTraits(DiffableTraits, CatalogMemberTraits, SplitterTraits))
+  )
 ) {
+  constructor(...args: ModelConstructorParameters) {
+    // TODO: [mobx-undecorate] verify the constructor arguments and the arguments of this automatically generated super call
```
tephenavies: Remove comment

test/ModelMixins/TimeFilterMixinSpec.ts
```diff
@@ -34,8 +34,15 @@ describe("TimeFilterMixin", function () {
 });
 
 class TestTimeFilterableItem extends TimeFilterMixin(
-  CreateModel(mixTraits(TimeFilterTraits))
+  MappableMixin(CreateModel(mixTraits(TimeFilterTraits)))
 ) {
+  constructor(...args: ModelConstructorParameters) {
+    // TODO: [mobx-undecorate] verify the constructor arguments and the arguments of this automatically generated super call
```
tephenavies: Another undecorate comment

package.json
```diff
@@ -163,34 +163,38 @@
...
     "ts-loader": "^5.3.3",
     "ts-node": "^5.0.1",
-    "typescript": "^3.9.2",
+    "typescript": "^4.9.4",
     "urijs": "^1.18.12",
     "url-loader": "^1.1.2",
-    "webpack": "~4.39.0",
+    "webpack": "~4.46.0",
+    "webpack-cli": "^3.3.11",
+    "webpack-dev-server": "^3.1.14",
```
tephenavies, line 179: These probably don't need to be dependencies

tephenavies (changes requested):

lib/ReactViews/Search/SearchBox.jsx
tephenavies: It doesn't at the moment because it's not wrapped with `@observer`

lib/Models/Definition/addModelStrataView.ts
tephenavies: I've had a look into this, and I think this only works because the caller of this (CreateModel) puts `makeObservable` in the constructor. This is fine to rely on as long as we include a comment in this file explaining that.

lib/Models/Definition/createCombinedModel.ts
tephenavies: It seems you've changed this away from using `makeObservable`. Could you explain why?

lib/Models/Definition/addModelStrataView.ts
na9da: The mobx4/5 decorate utility was [removed in mobx6](https://mobx.js.org/migrating-from-4-or-5.html#upgrading-classes-to-use-makeobservable).

> this only works because the caller of this (CreateModel) puts makeObservable in the constructor

Yeah, as you might already know, in mobx6 the role of decorators is [simply to note](https://github.com/mobxjs/mobx/blob/58bb052ca41b8592e5bd5c3003b68ec52da53f33/packages/mobx/src/api/computed.ts#L38-L41) the properties that need to be enhanced. The actual enhancement is done by the call to [makeObservable](https://github.com/mobxjs/mobx/blob/58bb052ca41b8592e5bd5c3003b68ec52da53f33/packages/mobx/src/api/makeObservable.ts#L43) in the constructor. So yes, this wouldn't work without a call to `makeObservable`.

lib/Models/Definition/createCombinedModel.ts
na9da: > It seems you've changed this away from using makeObservable. Could you explain why?

Not sure what you meant. The only difference from the other `decorate()` definition is that here we are decorating [a plain object](https://github.com/TerriaJS/terriajs/blob/1a2295507cd064e7d14ba9f9c31d9bfce4bb58f8/lib/Models/Definition/createCombinedModel.ts#L189-L193) instead of a [class prototype](https://github.com/TerriaJS/terriajs/blob/1a2295507cd064e7d14ba9f9c31d9bfce4bb58f8/lib/Models/Definition/addModelStrataView.ts#L58).

na9da: Actually, good catch. No one is calling `makeObservable` on the [result](https://github.com/TerriaJS/terriajs/blob/1a2295507cd064e7d14ba9f9c31d9bfce4bb58f8/lib/Models/Definition/createCombinedModel.ts#L189-L193). 
So in reality, the trait  wasn't getting memoized.

I'll add a `makeObservable(result)` after decorating it.

lib/ViewModels/CompositeBar/CompositeBarItemController.ts
na9da: The method was previously decorated using `@action`. When child classes override this method, they'll have to remember to use `@override` - otherwise it causes an app crash. This is hard to enforce especially if community users have extended this class. By switching to `runInAction` instead, child classes can freely override the method.

Couple of things to note for the future: 
  - Mixing mobx decorators and inheritance is tricky.
  - Inheritance as an API mechanism is brittle because when overriding a method, the dependency between child class and parent class is implicit. This makes it difficult for upstream library to make changes without breaking downstream behaviour in non-obvious ways.

lib/Models/Definition/createCombinedModel.ts
na9da: [Done](https://github.com/TerriaJS/terriajs/blob/6f8a680ba2d1d5f3188c24e9d0ce2586fb573881/lib/Models/Definition/createCombinedModel.ts#L235).

package.json
na9da: Removed.

na9da:
> BTW why are there no changes needed in DiscretelyTimeVaryingTraits? Is it because the interface is only needed if properties get overridden twice?

@steve9164  - that's right. We have to use the interface trick when a trait definition creeps in more than once in the signature. To see this in action:
 - Get rid of this [cacheDuration](https://github.com/TerriaJS/terriajs/blob/76a586cc4b0af14d794350c3d938b3c2f5b91e14/lib/Traits/TraitsClasses/UrlTraits.ts#L32) definition.
 - Observe TS raise the dreaded 2611 errors.
 - Now replace this [UrlTraits](https://github.com/TerriaJS/terriajs/blob/76a586cc4b0af14d794350c3d938b3c2f5b91e14/lib/ModelMixins/UrlMixin.ts#L7) with `Omit<UrlTraits, "cacheDuration">` 
 - Observe the errors disappear.
 
What that demonstrates is  when the trait definition appears in the type hierarchy more than once it triggers the error. In this case it is appearing in the signature of `UrlMixin` (`Model<UrlTraits>`) and the signature of `CreateModel(SomeCatalogItemWithUrlTraits)`.

zoran995:
@na9da I wonder don't we need to set a `useDefineForClassFields` as stated here https://mobx.js.org/installation.html#use-spec-compliant-transpilation-for-class-properties ?

na9da:
> 

Good find @zoran995. I had missed that. I'll check how turning on `useDefineForClassFields` affects our code. Thanks!

na9da:
@zoran995 @steve9164 - I have made a [new commit to](https://github.com/TerriaJS/terriajs/pull/6687/commits/4006433874a994313fa41d17550ca14a9bbbf9c0) enable [useDefineForClassFields](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier) and fix associated errors.

One change (as you can see in the commit) is that when a trait-class extends from another trait-class and re-defines a trait, we have to explicitly initialize the trait. TS will raise a type error otherwise. The reason I think is because with the new behaviour, an uninitialized class field is always set to `undefined`. So if the base class sets a value for the class field, then re-defining the field in a child class will silently make it `undefined`. TS will raise an error to force us to explicitly handle this.

At least in my limited testing, this doesn't seem to cause any other problem.

[TS playground](https://www.typescriptlang.org/play?useDefineForClassFields=true#code/MYGwhgzhAEBiD29oG8BQ0NgPwC5oQBcAnASwDsBzaAXmgCIw6BuVAX1VVEhgCEwjoAUwAeBQWQAmMBEjQZMufMXIUm0APTroAOXgFoAeQDS6eZomCuRQdGx5CpSms2GT8jertLHVWg2YermwcwPBkEPAgggB0IPAUABRkggDu0HxECQCU0WBZLEA) highlighting the error.

na9da:
I have upgraded typescript to 5.x. Things seem to work ok. Just need to continue using `experimentalDecorators` flag until some point when mobx decorators are migrated.

I feel we should make the jump to 5.x but can revert if for some reason we should wait.

zoran995:
Personally I would defer upgrade to ts5 for follow up pull request to avoid potential edge cases of upgrade. Basically one release with ts4 and mobx 6 and then next release ts5, so if something go bad community can stay on ts4 version till issue is resolved, since upgrade to ts4 is pretty major step forward

na9da:
No problem! Will revert to TS4 :+1:

tephenavies:
Looks good. Just needs type errors on node build fixed.

na9da:
There are following type errors on node build:
image

The styled-component error is easy to fix by updating to the latest versions.

The `AbortSignal` error can be fixed by upgrading to the latest `@types/nodes` but due to complicated dependencies it is hard to trace which packages are depending on the older `@types/node`. We will have to pin it to the latest version on `terriamap`.

na9da:
To fix the CI build errors I had to do the following things:
- Pin `@types/node` to `^18.15.11` in both [terriajs](https://github.com/TerriaJS/terriajs/blob/4757b9e04397ab3177fa5bd00f7a767fc4daae4e/package.json#L15) and [terriamap](https://github.com/TerriaJS/TerriaMap/blob/e4cd1970cb48c130ccdc9baf51e7f5b9fb6e0929/package.json#L40).
- [Move some deps](https://github.com/TerriaJS/terriajs/compare/90f8cda..4757b9e) in terriajs from `devDependencies` to `dependencies`

zoran995:
@steve9164 @na9da I wonder how would you advise the community on merging this into older branches/custom code? should there be pre/post tag?

na9da:
@zoran995 - yeah I think it is a good idea to have a pre and post tag when we merge.

We'll also plan to make a community announcement on the discussions forum and provide some level of instructions on upgrading.

In general you might need manual fixes to your code if you have model layer changes in your fork of terriamap/terriajs. These fall into the following types:
 - Error when using mobx decorators in your classes (`mobx-undecorate` provides some degree of automation to fix this).
 - Errors when overriding some mobx decorated method from parent class. You'll need to use `@override` to do that.
 - TS2611 error when overriding traits in your custom catalog item
 - Other typescript errors relating to your code

na9da:
BTW we have an [8.3.0-rc.2](https://www.npmjs.com/package/terriajs/v/8.3.0-rc.2) release, which is one way to try it out. [Some of our maps](https://nationalmap.gov.au/) have been upgraded to `rc` version so that we can be confident about stability of the build before we merge it in.

na9da:
To investigate: Mobx throws a weird error when wrapping `componentDidMount` in `action`. I have made a change to fix it in one place https://github.com/TerriaJS/terriajs/pull/6687/commits/17464b042b1926f20ecacbf3a89d6a11a7bc5b13.

Error is something like: `action can only be used on properties with a function value.`

Can be reproduced by opening the `Difference` tool using a version of this branch prior to the commit https://github.com/TerriaJS/terriajs/pull/6687/commits/17464b042b1926f20ecacbf3a89d6a11a7bc5b13

na9da:
Can repro the above error by wrapping any `componentDidMount()` definition with an `@action` decorator. Pretty sure this has to do with how the new `mobx-react` `observer` function patches the `componentDidMount` method. Not sure if this is intentional or a bug. 

We have just a couple of places where we do this, so it is easy to workaround by using `runInAction()` within the method.

tephenavies (approved):

PR #6691, opened 2023-02-07
Display group 1

### LINKED ISSUE #6684: Allow users to easily select all data layers under a group/sub-group
Allow users to enable all layers under a group or sub-group in the catalogue to the workbench Post adding, it should allow users to disable layers. 
Triggered by the need to create a complete 'image' of an area (made up from different reality meshes, buildings, or other data, including 2D). 

Options for implementation: 
- [ ] enable a button under group / subgroup to allow users to "select all" 
- [ ] use a 'composite' layer made of different other layers (not necessarily in the same group)

staffordsmith83:
### What this PR does

Fixes #6684

Allows a Catalog Group to have an Add All button.
This is optional, defined by a new GroupTrait `displayGroup: boolean`.

### Test me

Use this link to test.
http://ci.terria.io/display-group-1/#clean&https://gist.githubusercontent.com/staffordsmith83/81893a01b7709328c55fc332b7307bd3/raw/0cc673b2dc954c6dd5bf099e48b1d1b70ed796e0/displayGroupsCatalog.json

There should be only two group items in the catalog. One is a displayGroup.
Try to add the items to the catalog with Add All button.
Add All button should change to Remove All.

Try adding all, then removing one item from the workbench.
Add All button should appear in the catalog again - Remove All is only active when all group items exist in the workbench.

Also see the icons next to the tree in the catalog - should have an add all small icon that changes to a minus symbol with the same behaviour at the Add/Remove All button..

Screen Shot 2023-02-14 at 2 59 26 pm

### Checklist


staffordsmith83:
Dear @philipgrimmett  I have implemented your changes. Another review please?

DONE -   _plus-list-20_ icon right-aligned next to group title in catalogue list
DONE -   _Add all_ text toggles to _Remove all_
DONE -    _Add all_ button right-aligned in description header
DONE  It doesn't match existing behaviour, but I find myself expecting to see the workbench items appear in the same order that they appeared in the catalogue
DONE -   Doesn't look like we're currently using rounded buttons in the catalogue, if there a rect style we can use instead?
DONE (I think? Please clarify if not)  Ensure group accordion doesn't collapse/expand when users quick-add groups to the map

philipgrimmett:
Thanks @staffordsmith83 ! Notes below:

- I noticed that we're not using consistent icons for quick removing groups and individual datasets. Probably best if we use the minus-circle icon across both and reserve close-circle for dismissing modal windows, clearing text fields etc.
Screenshot 2023-02-14 at 1 22 12 pm

- The 'add all' button for groups is effectively the same as the 'add to map' button for individual datasets, but their padding is different. Not sure if this PR is the right place to make that charge (see https://github.com/TerriaJS/product/issues/149), but if we're hard coding styles anyway it would be great if primary buttons have a minimum of 8px padding.

staffordsmith83:
Hi @philipgrimmett regarding the close circle, I thought you might want to provide a min-list icons similar to your plus list?
plus-list-20 (1)

I can adjust that padding, but it does not sit very well with the other stuff on the row... Ill have a play. Cheers!

lib/ReactViews/DataCatalog/CatalogGroup.jsx
```diff
@@ -135,7 +161,10 @@ CatalogGroup.propTypes = {
   ]),
   selected: PropTypes.bool,
   removable: PropTypes.bool,
-  removeUserAddedData: PropTypes.func
+  removeUserAddedData: PropTypes.func,
+  displayGroup: PropTypes.bool,
+  checkAllItemsLoadedFunction: PropTypes.func,
+  addRemoveButtonFunction: PropTypes.func
```
na9da: Doubt: Is `checkAllItemsLoadedFunction` a function or boolean? Based on the the usage it appears to be boolean to me. If so can we rename it for clarity?

lib/ReactViews/DataCatalog/DisplayGroupService.ts
```diff
@@ -0,0 +1,85 @@
...
+  // Check if all the mappable items from our group are already loaded in the workbench
+  const checkAllMappablesInWorkbench = (
+    workbenchArr: readonly string[],
+    groupItemsArray: string[]
+  ) =>
+    groupItemsArray.every(
+      (member: any) =>
+        !MappableMixin.isMixedInto(terria.getModelById(BaseModel, member)) ||
+        workbenchArr.includes(member)
+    );
```
na9da: Nitpick: Any reason to wrap this inside a function? (possibly happened during a refactor or something?)

lib/ReactViews/DataCatalog/DisplayGroupService.ts
```diff
@@ -0,0 +1,85 @@
...
+      terria
+    );
+    previewedGroup.loadMembers().then(() => {
+      for (
+        let index = previewedGroup.memberModels.length - 1;
+        index >= 0;
+        index--
+      ) {
+        const memberModel = previewedGroup.memberModels[index];
+        addOrRemoveMember(memberModel, viewState, forceState, keepCatalogOpen);
```
na9da: Doubt: Can we avoid introducing `forceState` if we only call `addOrRemoveMember()` on a member if it is not in the state that we want it to be. i.e If user clicked "Add all" we call it only if the member is not already added to the workbench (and vice-versa).

lib/ReactViews/DataCatalog/DisplayGroupService.ts
```diff
@@ -0,0 +1,85 @@
+import { runInAction } from "mobx";
```
na9da: Is splitting into `*Service` file a pattern you have used before? Personally, I have been writing them as loose (outer functions) inside the component file itself. I don't hold a strong opinion at this point though.

na9da (changes requested):
Hey @staffordsmith83, I have tested the feature and it is working great. Just left a few comments to consider.

lib/ReactViews/DataCatalog/CatalogGroup.jsx
staffordsmith83: Good catch, done

lib/ReactViews/DataCatalog/DisplayGroupService.ts
staffordsmith83: Yes, happened in a refactor, done!

lib/ReactViews/DataCatalog/DisplayGroupService.ts
staffordsmith83: Yes, i think the logic I have added now should do it, good idea.

staffordsmith83:
Dear @na9da Ive implemented all your ideas, but one question: I get quite a few errors `Computed value CsvCatalogItem@642.zoomOnAddToWorkbench is being read outside a reactive context. Doing a full recompute` - I think these are present in `main` too, but do you know what might be causing it?

lib/ReactViews/DataCatalog/DisplayGroupService.ts
staffordsmith83: I have split it into a service file because these funtions are imported into two different components. 
Yeah its a pattern i used all the time, but now I think the term might be used more by Angular devs... Should I rename?

lib/ReactViews/DataCatalog/DisplayGroupService.ts
staffordsmith83: it might be worth checking my logic here, second pair of eyes, but I think its sound

lib/ReactViews/DataCatalog/DisplayGroupService.ts
na9da: Sounds good, although I like `*Helper` better. Service sounds like some daemon or process to me, but probably that's just me?

lib/ReactViews/DataCatalog/DisplayGroupService.ts
na9da: I feel I have to pause a bit  to understand the logic. The problem could be that we only have a `toggleItemOnMapFromCatalog` to work with, so it is hard to express whether we want add or remove the item. 

One solution might be to add separate functions in the `DispalyGroupService` file for adding and removing: `addMemberToMap` and `removeMemberFromMap`. Each would internally call `toggleItemOnMapFromCatalog` after checking `viewState.terria.workbench.contains(item)`.

Then we could write something like:
```
 if (allItemsLoaded) {
          // toggle action should remove all group members from map
          removeMemberFromMap(memberModel, viewState, keepCatalogOpen);
  } else if (!itemLoaded) {
           // toggle action should add missing members to the map
          addMemberToMap(memberModel, viewState, keepCatalogOpen);
  }
```

What say @staffordsmith83 ?

lib/ReactViews/DataCatalog/DisplayGroupService.ts
staffordsmith83: Yeah, I think I must have got 'Service' from using Angular so much. Helper is a better generic term I will change

lib/ReactViews/DataCatalog/DisplayGroupService.ts
staffordsmith83: So I have fixed it with the new logic - thinking that it was best to not introduce another parameter into `toggleItemOnMapFromCatalog` function, and that it would also be best to keep using that function rather than write another one.

The logic is sound:
We check if all items are already loaded
We check if our current item of interest is already loaded
If all items are loaded, then we can just call `toggle` for all, and then they will all be removed, this is what we want.
Net we check if `!allItemsLoaded && !itemLoaded` - if we have a partially loaded group, and our current item of interest is not loaded, then we toggle it, which will load it - this is what we want.

All other cases we do nothing. I think there is only one other case not accounted for:
`!allItemsLoaded && itemLoaded` - thats fine, do nothing


I.e. like this:

```
const allItemsLoaded = allMappableMembersInWorkbench(
      previewedGroup.members,
      terria
    );
    previewedGroup.loadMembers().then(() => {
      for (
        let index = previewedGroup.memberModels.length - 1;
        index >= 0;
        index--
      ) {
        const memberModel = previewedGroup.memberModels[index];
        const itemLoaded = terria.workbench.contains(memberModel);
        if (allItemsLoaded) {
          addOrRemoveMember(memberModel, viewState, keepCatalogOpen);
        } else if (!allItemsLoaded && !itemLoaded) {
          addOrRemoveMember(memberModel, viewState, keepCatalogOpen);
        }
      }
    });
```

In this way the logic achieves what we want, without writing a duplicate function with similar role, and without introduing more parameters like `forceState`.

However, I get your point that rewriting it could make it clearer? What do you think? I think given the comments in the code its clear enough personally?

na9da: Got it.

I'll probably also leave a comment inside the branches of the `if` condition just to be extra clear.

staffordsmith83: OK will do!

na9da (approved):
Looks great @staffordsmith83!

As for the `computed` warning it's been there for sometime but we need to fix it.

Chrome lets you expand the warning message to see the trace:
image


Looks like it is happening in [Wokrbench.ts](https://github.com/TerriaJS/terriajs/blob/65d5f5985bc06725f31f58428dd0ac9d95d98f86/lib/Models/Workbench.ts#L211-L213) because we are not wrapping the code after the `await` inside an action.

PR #6703, opened 2023-02-20
Revert reaction to autorun

### LINKED ISSUE #6702: Terrain Hides Underground Features not working
This was a bug introduced in https://github.com/TerriaJS/terriajs/pull/6675

When we changed autorun to reaction  [in this code](https://github.com/TerriaJS/terriajs/blob/df3636bdc307c4d9b8c48a9db807094e8cafe4cb/lib/Models/Cesium.ts#L393-L408) it no longer responds to changes in those settings. (The reaction doesn't run because according to mobx docs the first parameter must return all the dependency values). 

Now that we have solved the test issues differently, can we revert to using autorun?

staffordsmith83:
### What this PR does

Fixes #6702

Changed reaction back to autorun in `Cesium.ts`.

### Test me

Try to set terrain to hide underground features.

### Checklist

- [y] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)

### DIFF
--- a/lib/Models/Cesium.ts
+++ b/lib/Models/Cesium.ts
@@ -6,8 +6,7 @@ import {
   IObservableArray,
   observable,
   reaction,
-  runInAction,
-  IAutorunOptions
+  runInAction
 } from "mobx";
 import { computedFn } from "mobx-utils";
 import AssociativeArray from "terriajs-cesium/Source/Core/AssociativeArray";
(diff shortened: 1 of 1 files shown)


na9da (approved):
Looks good!

Tested by toggling `Terrain hides underground features` and checking it correctly sets `viewState.terria.scene.globe.depthTestAgainstTerrain`.

PR #6708, opened 2023-02-24
Box drawing fixes and update

na9da:
### What this PR does

- Adds a new option called [disableVerticalMovement](https://github.com/TerriaJS/terriajs/blob/b6502f7a70c76a34980833734396dad381ac826e/lib/Models/BoxDrawing.ts#L177) to `BoxDrawing` which if set to `true` disables up/down motion of the box when dragging the top/bottom sides of the box. 
- Keeps height (mostly) steady when moving the box laterally on the map. Previously the height of the box used to change wrt to the ellipsoid/surface.
- Fixes a bug that caused map panning and zooming to break when interacting with multiple active BoxDrawings.
- Removed some code that was causing too much drift between mouse cursor and model when moving the model laterally on the map.

### Test me

- Visit this [CI link for main branch](http://ci.terria.io/main/#share=s-krVENrjdcWkasKZA56f59ZRWpEC).
- You should be able to pan and zoom the map
- Try clicking on the box and then try panning and zooming. It will not work.
- Now try it with [this branch](http://ci.terria.io/restrict-vert-box-movement/#clean&share=s-krVENrjdcWkasKZA56f59ZRWpEC), the bug should be fixed.

### Checklist

### DIFF
--- a/lib/Models/BoxDrawing.ts
+++ b/lib/Models/BoxDrawing.ts
@@ -1577,7 +1577,13 @@ function intersectRayEllipsoid(
     return;
   }
 
-  const intersectionPoint = Ray.getPoint(ray, interval.start, result);
+  // start=0 means ray origin is inside the ellipsoid
+  // in which case there is only a single intersection
+  // which is given by stop.
+  // This can happen, for eg, when moving the box while the camera
+  // is below the box.
+  const t = interval.start !== 0 ? interval.start : interval.stop;
+  const intersectionPoint = Ray.getPoint(ray, t, result);
   return intersectionPoint;
 }
 
--- a/lib/Models/BoxDrawing.ts
+++ b/lib/Models/BoxDrawing.ts
@@ -244,6 +244,8 @@ export default class BoxDrawing {
 
   public disableVerticalMovement: boolean;
 
+  public keepHeightSteadyWhenMovingLaterally = true;
+
   public onChange?: (params: BoxDrawingChangeParams) => void;
 
   // An external transform to convert the box in local coordinates to world coordinates
@@ -788,7 +790,9 @@ export default class BoxDrawing {
     const scratchDirection = new Cartesian3();
     const scratchMoveVector = new Cartesian3();
     const scratchEllipsoid = new Ellipsoid();
+    const scratchRay = new Ray();
     const scratchCartographic = new Cartographic();
+    const scratchPreviousPosition = new Cartesian3();
     const scratchCartesian = new Cartesian3();
     const scratchMoveStep = new Cartesian3();
 
@@ -828,7 +832,7 @@ export default class BoxDrawing {
         );
 
         moveStep = moveVector;
-      } else {
+      } else if (this.keepHeightSteadyWhenMovingLaterally) {
         // Move the box laterally on the globe while keeping the height
         // steady. To do this we derive a new ellipsoid from the globe's
         // ellipsoid by enlarging it so that the current box center falls on
@@ -840,7 +844,7 @@ export default class BoxDrawing {
 
         const cameraRay = scene.camera.getPickRay(
           mouseMove.endPosition,
-          new Ray()
+          scratchRay
         );
         if (!cameraRay) {
           return;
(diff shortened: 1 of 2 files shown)


lib/Models/BoxDrawing.ts
```diff
@@ -533,7 +543,11 @@ export default class BoxDrawing {
     const handlePick = (click: MouseClick) => {
       const pick = scene.pick(click.position);
       const entity = pick?.id;
-      if (entity === undefined || !isInteractable(entity)) {
+      if (
+        entity === undefined ||
+        !isInteractable(entity) ||
+        !this.dataSource.entities.contains(entity)
```
na9da, line 553: Without this extra check, the `handlePick` function is run even when the picked entity does not belong to this box. As a result, when multiple BoxDrawings are active on the map, we capture the map state multiple times and restore it back multiple times resulting in breaking of the panning and zooming of the map:

https://github.com/TerriaJS/terriajs/blob/b6502f7a70c76a34980833734396dad381ac826e/lib/Models/BoxDrawing.ts#L563-L566

lib/Models/BoxDrawing.ts
```diff
@@ -776,6 +790,38 @@ export default class BoxDrawing {
     const scratchMoveStep = new Cartesian3();
     const scratchSurfacePoint = new Cartesian3();
     const scratchSurfacePoint2d = new Cartesian2();
+    const scratchNextStep = new Cartesian3();
+    const scratchNextCartographic = new Cartographic();
+
+    const keepHeightSteady = (
+      currentPosition: Cartesian3,
+      moveStep: Cartesian3
```
na9da, line 798: 1. Find the next cartesian position (x,y,z wrt center of earth) from the `moveStep`
2. Compute its cartographic (lat, lon, height) position. 
3. Then set the cartographic height of the next position to the previous cartographic height and convert back to cartesian.

lib/Models/BoxDrawing.ts
```diff
@@ -948,11 +1003,25 @@ export default class BoxDrawing {
...
+    // Call enabledFn only if movement is is allowed for this side, otherwise call disabledFn
+    const ifActionEnabled = (
+      enabledFn: (...args: any[]) => any,
+      disabledFn?: (...args: any[]) => any
+    ) => {
+      return (...args: any[]) => {
+        return this.disableVerticalMovement && isTopOrBottomSide
+          ? disabledFn?.apply(this, args)
+          : enabledFn.apply(this, args);
+      };
```
na9da, line 1022: A helper function that performs the `enabledFn`  action if vertical movement is enabled, otherwise it performs the `disabledFn` action.

lib/Models/BoxDrawing.ts
```diff
@@ -818,34 +864,7 @@ export default class BoxDrawing {
...
-        if (!surfacePoint2d) {
-          // cartesianToCanvasCoordinates can unexpectedly return undefined.
-          return;
-        }
-
-        // Floor the startPosition and endPosition above the ellipsoid.
-        const yDiff = mouseMove.endPosition.y - mouseMove.startPosition.y;
-        mouseMove.startPosition.y = surfacePoint2d.y;
-        mouseMove.endPosition.y = surfacePoint2d.y + yDiff;
-
```
na9da: Remove some experimental code that was causing too much drift between the mouse cursor and the model.

lib/Models/BoxDrawing.ts
```diff
@@ -814,72 +829,72 @@ export default class BoxDrawing {
...
-            mouseMove.endPosition,
-            scratchEndPosition
-          ) ??
-          scene.camera.pickEllipsoid(
-            mouseMove.endPosition,
-            undefined,
-            scratchEndPosition
-          );
-
-        if (!previousPosition || !endPosition) {
```
na9da, line 879: We have improved the lateral movement of the box by keeping its height steady as we move it on the globe surface. 

To do this we derive a new ellipsoid from `globe.ellipsoid` which passes through the center of the box. This ellipsoid will roughly have the same shape as that of the `globe.ellipsoid`. Then we find the intersection point between the camera ray (ray from camera origin to location of users mouse cursor) and this new enlarged ellipsoid. This new point will lie on the ray and will have roughly the same height as the current box position. 

Because the new point lies on the camera ray, it makes it look like the box is following the mouse cursor.

lib/Models/BoxDrawing.ts
```diff
@@ -813,45 +832,76 @@ export default class BoxDrawing {
...
-        // Box origin projected on the ellipsoid surface.
-        const surfacePoint = projectPointToSurface(origin, scratchSurfacePoint);
-        // Surface point in screen coordinates
-        const surfacePoint2d = scene.cartesianToCanvasCoordinates(
-          surfacePoint,
-          scratchSurfacePoint2d
+      } else if (this.keepHeightSteadyWhenMovingLaterally) {
+        // Move the box laterally on the globe while keeping the height
+        // steady. To do this we derive a new ellipsoid from the globe's
+        // ellipsoid by enlarging it so that the current box center falls on
```
staffordsmith83, line 838: Clever approach, nice

staffordsmith83 (commented):
This looks great Nanda, have tested a lot and cant find any problems. Clever approach!

staffordsmith83 (approved):
This all looks great

PR #6715, opened 2023-02-28
Change UploadTypes API to support adding multiple entries for the same catalog item type

na9da:
### What this PR does

Replaces `addRemoteUploadType` and `addLocalUploadType` with `addOrReplaceRemoteFileUploadType` and `addOrReplaceLocalFileUploadType`.

Previously the upload type entries were keyed using the catalog item `type` which allowed only 1 entry per catalog item. But sometimes we want to add multiple entries for a catalog item. For eg: we want to add separate entries for OBJ & Collada types, both handled by "assimp" catalog item.

The new API function allows this possibility by accepting a `key` param which is used in place of `type` value to key the entries. 

### Test me

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 #### next release (8.2.25)
 
+- Replaces addRemoteUploadType and addLocalUploadType with addOrReplaceRemoteFileUploadType and addOrReplaceLocalFileUploadType.
 - [The next improvement]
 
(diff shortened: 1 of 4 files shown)


sixlighthouses (approved):
Spec tests pass and functionality is visible and correct in https://test.nsw.digitaltwin.terria.io/ 
Approved

PR #6724, opened 2023-03-06
Issue 6716

### LINKED ISSUE #6716: Location/Locate Me functionality Ideal Zoom zooms in too far and places you under terrain
When using the `Location` button (what I call Locate Me) it adds `My Location` as an item on the workbench. Clicking the `Ideal Zoom` button on this item zooms you in too far, far enough to place the viewport below terrain.

Screenshot 2023-02-28 at 16 12 27

Screenshot 2023-02-28 at 16 12 46

See video of bug:

https://user-images.githubusercontent.com/5283003/221760489-22bd9677-6d33-47d8-b43a-50066364cd87.mov

We should not zoom in to such an extreme degree, and instead place the viewport e.g. just when the scale switches to be `100m`:

Screenshot 2023-02-28 at 16 16 59

mwu2018:
### What this PR does

Fixes #6716 

~The solution is to add a customised ideal zoom to the location point, taking terrain height into account.~

An ideal zoom depends on an item's `rectangle` trait (default) or customised `idealZoom` trait. If none of them are well defined, the behaviour will not be as expected.

This bug fix is to specify a `rectangle` ranges, as a close-up view, to the location mark that is programmatically created as `geojson` type.

This principle should be applied to all map items, whether they are `geojson` type or not, single point or not.

There is another [attempt](https://github.com/TerriaJS/terriajs/pull/6754) to automatically specify rectangle ranges to a single point of `geojson` type. However, that approach is not practical if the geojson data is provided via url. 

On balance, the solution provided by this fix is a good choice.

@na9da suggested that [this code path](https://github.com/TerriaJS/terriajs/blob/e3b155dd0b8c03b060de1f17907ace62df109d0a/lib/Models/Cesium.ts#L925) be modified to be similar to [this code path](https://github.com/TerriaJS/terriajs/blob/e3b155dd0b8c03b060de1f17907ace62df109d0a/lib/Models/Cesium.ts#L841) where it  takes terrain height into account when zooming in. However, the ideal zoom of an added location point will take [this code path](https://github.com/TerriaJS/terriajs/blob/e3b155dd0b8c03b060de1f17907ace62df109d0a/lib/Models/Cesium.ts#L915).

### Test me
Note: Must use `https` protocol!

Can be verified by comparing the results between [before](https://ci.terria.io/main) and [after](https://ci.terria.io/issue-6716).
- Choose `3D Terrain` or `3D Smooth` mode.
- Click on `Location` button on the right.
- Click on `IDEAL ZOOM` button on the left.
- No difference if choosing 2D mode. They work well both before and after.

The above tests will use real location point of a host computer.

### Checklist

- [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) The above test procedure probably is better than a unit test.


lib/ReactViews/Map/Navigation/Items/MyLocation.ts
```diff
@@ -14,6 +14,12 @@ import ViewerMode from "../../../../Models/ViewerMode";
 import { GLYPHS } from "../../../../Styled/Icon";
 import StyleTraits from "../../../../Traits/TraitsClasses/StyleTraits";
 import MapNavigationItemController from "../../../../ViewModels/MapNavigation/MapNavigationItemController";
+import {
+  LookAtTraits,
+  IdealZoomTraits
+} from "../../../../Traits/TraitsClasses/MappableTraits";
+import Cartographic from "terriajs-cesium/Source/Core/Cartographic";
+import { sampleTerrain } from "terriajs-cesium";
```
tephenavies: @mwu2018 I think this is the source of some of your problems. This should be imported from the correct file within terriajs-cesium. terriajs-cesium does not support importing without file paths.

tephenavies: You can track it back through the webpack build error:
```
ERROR in ./node_modules/terriajs-cesium/Source/Widgets/BaseLayerPicker/createDefaultImageryProviderViewModels.js
Module not found: Error: Can't resolve '/home/runner/work/terriajs/terriajs/node_modules/terriajs-cesium/Source/Assets/Textures/NaturalEarthII' in '/home/runner/work/terriajs/terriajs/node_modules/terriajs-cesium/Source/Widgets/BaseLayerPicker'
 @ ./node_modules/terriajs-cesium/Source/Widgets/BaseLayerPicker/createDefaultImageryProviderViewModels.js 237:15-128
 @ ./node_modules/terriajs-cesium/Source/Widgets/Viewer/Viewer.js
 @ ./node_modules/terriajs-cesium/Source/Cesium.js
 @ ./lib/ReactViews/Map/Navigation/Items/MyLocation.ts
 @ ./lib/ReactViews/Map/Navigation/registerMapNavigations.tsx
 @ ./lib/ReactViews/Map/Navigation/MapNavigation.tsx
 @ ./lib/ReactViews/StandardUserInterface/MapColumn.jsx
 @ ./lib/ReactViews/StandardUserInterface/StandardUserInterface.tsx
 @ ./lib/ReactViewModels/ViewState.ts
 @ ./test/Models/MapNavigation/MapNavigationModelSpec.ts
 @ multi ./test/SpecMain.ts ./test/Core/addedByUserSpec.ts ...
 ```
 
 The first line in the "stack trace" that is in terriajs code is MyLocation.ts importing Cesium.js

mwu2018: You are spot-on. I used vscode quick-fix for automatically import, which did not raise any warnings. The error only shows up in the CI build.

AnaBelgun:
Is it possible to prioritise this PR review and approval please? *before Easter*

tephenavies:
I'd prefer if we could solve this in a way that ideal zoom for all single point datasets would function correctly. E.g. try adding https://gist.githubusercontent.com/steve9164/7e8c75c5586883a69ee4776ccebe4353/raw/74b78acc58d59b1961f3d277a7a4eeebfa9c97e8/point.geojson as web data, and using ideal zoom.

mwu2018:
> I'd prefer if we could solve this in a way that ideal zoom for all single point datasets would function correctly. E.g. try adding https://gist.githubusercontent.com/steve9164/7e8c75c5586883a69ee4776ccebe4353/raw/74b78acc58d59b1961f3d277a7a4eeebfa9c97e8/point.geojson as web data, and using ideal zoom.

@steve9164 See the revised description.

tephenavies (approved):

PR #6730, opened 2023-03-17
Issue 6726

### LINKED ISSUE #6726: Magda v2 auth upgrade: Update padlock icon UI display logic
This ticket is created for [DTV Magda v2 Auth Upgrade](https://github.com/TerriaJS/vic-digital-twin-app/issues/63)

The required logic has been specified in section 3.2.2:

https://github.com/TerriaJS/vic-digital-twin-app/blob/17b1729823757282f5a10e8b58dab4167ba4d71e/doc/access-control-in-magda-v1-and-migration-plan.md#322-padlock-ui-logic-changes-in-terriajs

### Technical Notes

- For ease of version management & deployment, we should make the changes backward compatible.
e.g. We can check if aspect "esri-access-control" exists (i.e. it's a Magda v1 backend) to tell whether we should apply the new logic (for Magda v2)
Another option to check the backend version is to check the `authnReadPolicyId` field of the registry record --- the `authnReadPolicyId` field has been removed since Magda v2

- Pseudocode of the desired logic

```
if (`esri-access-control` exists) {
  if(!isEmpty(`esri-access-control`.`access`) && `esri-access-control`.`access` != 'public') {
    show padlock
  } else {
    hide padlock
  }
} else if(`access-control` exists) {
  // the logic here should be updated according to our migration plan setup
  if(isUuid(access-control`.`orgUnitId `)) {
    show padlock
  } else {
    hide padlock
  }
} else {
  hide padlock
}
```
p.s. the link to `access-control` aspect definition
https://github.com/magda-io/magda/blob/master/magda-registry-aspects/access-control.schema.json

mwu2018:
### What this PR does

Fixes #6726 

### Test me
Can only be tested with DT. 
Unit tests alone are good enough.

### Checklist


lib/Models/Catalog/CatalogReferences/MagdaReference.ts
```diff
@@ -863,8 +863,13 @@ const prepareDistributionFormat = createTransformer(
...
 function getAccessTypeFromMagdaRecord(magdaRecord: any): string {
   const record = toJS(magdaRecord);
-  const accessControl: any =
-    record && record.aspects && record.aspects["esri-access-control"];
-  const access = accessControl && accessControl.access;
-  return access;
+
+  // Magda V2 access control has higher priority.
+  if (record?.aspects?.["access-control"]) {
+    return record.aspects["access-control"].orgUnitId ? "non-public" : "public";
```
t83714: should be `record.aspects["access-control"]?.orgUnitId`?

mwu2018: Has tested `record.aspects["access-control"]` in the if statement.

t83714: yes you are right 👍

t83714:
@mwu2018  This PR I had a look. It looks good to me :100: 

One more thing, Magda's side has a planned feature (which will not be released recently) that might impact the padlock:

https://github.com/magda-io/magda/issues/3455

It is a feature I believe is closer to the meaning that the padlock tried to represent.

Do you think it’s still possible to make some adjustments to cover that as well?

Or it’s too late, and we can look at it in future?

So the code might look like this now:
```
if (record?.aspects?.["access-control"]) {
  return record.aspects["access-control"].orgUnitId
    ? "non-public"
    : record.aspects["access-control"].constraintExemption
    ? "public"
    : "non-public";
} else if (record?.aspects?.["esri-access-control"]) {
  return record.aspects["esri-access-control"].access;
} else {
  return "public";
}
```

t83714 (approved):
Thanks @mwu2018 
Looks good to me 🎉

PR #6743, opened 2023-03-30
Mousemove perf improvements

na9da:
### What this PR does

Optimizes the `Compass` &  `LocationBar` components to reduce the number of re-renderings in each frame or when the user moves the mouse.

Currently, each mouse move triggers re-rendering of `LocationBar` component which is unnecessary. There is also unessecary re-rendering of the Compass component in each frame. 

When using an interactive tool like the scene-editor, this results in to much work being done in each frame, affecting the overall performance. The optimizations in this PR should at least slightly improve the performance without any noticeable change to the functionality.

Here is relevant snip from the flamegraph generated using firefox performance tool after a few seconds of interaction with the map.

image

This PR significantly reduces both hot paths.

### Test me

Both the compass and location bar features should continue to work as expected.

### Checklist


lib/ReactViews/Map/Navigation/Items/Compass.tsx
```diff
@@ -150,7 +152,7 @@ type IStateTypes = {
 };
 
 // the compass on map
-class Compass extends React.Component<PropTypes, IStateTypes> {
+class Compass extends React.PureComponent<PropTypes, IStateTypes> {
```
na9da, line 155: By using a `PureComponent`, react will re-render only if the state and prop values change.

lib/ReactViews/Map/Navigation/Items/Compass.tsx
```diff
@@ -757,13 +759,21 @@ function viewerChange(viewModel: Compass) {
...
+                });
+              }
+            },
+            200,
+            {
+              maxWait: 200,
+              leading: true,
+              trailing: true
+            }
+          )
```
na9da, line 776: debounce to reduce number of updates

lib/ReactViews/Map/Legend/LocationBar.jsx
na9da, line 1: Converted to `.tsx`

lib/ReactViewModels/MouseCoords.ts
```diff
@@ -37,15 +48,18 @@ export default class MouseCoords {
     Cancelable;
   tileRequestInFlight?: unknown;
 
-  @observable elevation?: string;
-  @observable utmZone?: unknown;
-  @observable latitude?: string;
-  @observable longitude?: string;
-  @observable north?: unknown;
-  @observable east?: unknown;
-  @observable cartographic?: Cartographic;
```
na9da, line 46: From a performance perspective, it is perhaps a wrong idea to use `@observable` for very frequently updated values because of the mobx overhead. We instead introduce a new `updateEvent` to subscribe to changes.

lib/Models/GlobeOrMap.ts
```diff
@@ -60,7 +60,7 @@ export default abstract class GlobeOrMap {
   // Avoid duplicate mousemove events.  Why would we get duplicate mousemove events?  I'm glad you asked:
   // http://stackoverflow.com/questions/17818493/mousemove-event-repeating-every-second/17819113
   // I (Kevin Ring) see this consistently on my laptop when Windows Media Player is running.
-  @observable mouseCoords: MouseCoords = new MouseCoords();
+  mouseCoords: MouseCoords = new MouseCoords();
```
staffordsmith83, line 63: MouseCoords no longer observable, nice

lib/ReactViews/Map/Legend/LocationBar.tsx
```diff
@@ -0,0 +1,91 @@
...
+import MouseCoords from "../../../ReactViewModels/MouseCoords";
+import Styles from "./legend.scss";
+
+interface PropsType {
+  terria: Terria;
+  showUtmZone: boolean;
+  mouseCoords: MouseCoords;
+}
+
+const LocationBar: React.FC<PropsType> = observer(({ mouseCoords }) => {
```
staffordsmith83, line 16: And now mouseCoords is no longer an observable so should cut out overhead, and still rerender based on props being updated?

lib/ReactViews/Map/Navigation/Items/Compass.tsx
staffordsmith83: good idea

lib/ReactViews/Map/Navigation/Items/Compass.tsx
staffordsmith83: Because it only does a shallow comparison right?

staffordsmith83 (approved):
Looks really good and well thought out Nanda, no criticisms just a few comments for the record

lib/ReactViews/Map/Legend/LocationBar.tsx
na9da: The update is triggered by a [new event listener](https://github.com/TerriaJS/terriajs/blob/24d5ecb6084ec47b72440bfc5f7edb1be0bf9cac/lib/ReactViews/Map/Legend/LocationBar.tsx#L26). This is much more efficient because it avoids the mobx overhead. I'm also using react refs to update the DOM which [avoids the re-rendering](https://github.com/TerriaJS/terriajs/blob/24d5ecb6084ec47b72440bfc5f7edb1be0bf9cac/lib/ReactViews/Map/Legend/LocationBar.tsx#L27-L32) altogether.

lib/ReactViews/Map/Navigation/Items/Compass.tsx
na9da: Yes [PureComponent](https://react.dev/reference/react/PureComponent) does shallow comparison. Plain `Component` does only basic `===` equality check and is not effective when the object changes.

na9da:
Thanks Staf!

PR #6744, opened 2023-03-30
New BoxDrawing movement fixes

na9da:
### What this PR does

- Improves move behaviour for `BoxDrawing` 
- Fixes a bug that resulted in the map losing interactivity when the user releases the box while have a modifier key (eg ctrl/alt) pressed.

### Tetsting

Ping @na9da for details.

### Checklist


lib/Models/BoxDrawing.ts
```diff
@@ -619,6 +626,19 @@ export default class BoxDrawing {
...
+      // Bind the release action to all LEFT_UP + any modifier. This is
+      // required because we want the release to happen even if the user is by
+      // chance pressing down on some other key. In such cases Cesium will not
+      // trigger a LEFT_UP event unless we explicitly pass a modifier.
+      (modifier) =>
+        eventHandler.setInputAction(
+          handleRelease,
+          ScreenSpaceEventType.LEFT_UP,
+          modifier as any
+        )
```
na9da, line 639: Bind the release event + all available keyboard modifiers so that the release callback is  triggered even if the user is holding down a key.

lib/Models/BoxDrawing.ts
```diff
@@ -833,75 +861,94 @@ export default class BoxDrawing {
...
+        // Move the box laterally on the globe while keeping its height (almost) steady.
+        // To do this:
+        //   1. Find the exact point on the box surface mouse pick landed.
+        //   2. Derive a new ellipsoid such that the picked point on the box
+        //      lies on the new ellipsoid surface.
+        //   3. Find the point where the camera ray intersects this new
+        //      ellipsoid. This intersection point will have the same height as
+        //      that of the picked point while also visually coinciding with
+        //      the mouse cursor.
+        //   4. Use this intersection point to compute the box moveStep
```
na9da, line 873: Previously, we were not considering the exact point where the user clicked on the box. This resulted in a jumpy start to the interaction when the box center aligns itself with the mouse cursor. Now we account for the clicked point which rectifies this issue.

lib/Models/BoxDrawing.ts
staffordsmith83: This works well. Modifiers SHIFT, ALT and CTRL will stop box dragging, but do not prevent click release from having intended effect. No problem in my mind.

lib/Models/BoxDrawing.ts
staffordsmith83: Works well, great solution

lib/Models/BoxDrawing.ts
```diff
@@ -833,75 +861,94 @@ export default class BoxDrawing {
...
+        const cameraRay = scene.camera.getPickRay(
+          mouseMove.endPosition,
+          scratchRay
+        );
+
+        if (!cameraRay) {
+          return;
+        }
+
+        // Get the intersection between the camera ray and the derived ellipsoid
```
staffordsmith83, line 931: Ooh yes I see this is nice. I wondered how this might work!

lib/Models/BoxDrawing.ts
```diff
@@ -926,15 +973,23 @@ export default class BoxDrawing {
             scratchCartesian
           );
 
-        if (!previousPosition || !nextPosition) {
+        if (!nextPosition || !previousPosition) {
+          // We couldn't resolve a globe position, maybe because the mouse
+          // cursor is pointing up in the sky. Reset the box position when we
+          // can find the globe position again.
+          resetPosition = true;
```
staffordsmith83, line 980: Works well

staffordsmith83 (approved):
This looks great Nanda, good approach, logical, and I could follow with great comments. Could perhaps be refactored but I can follow it. No criticisms, as I am not very experienced with these Cesium patterns but tested and looks good.

PR #6745, opened 2023-03-30
API for listening to file drag-n-drop events

na9da:
### What this PR does

Adds a new method `addFileDragDropListener` to get callback after user drags-n-drops a file and it is loaded successfully. 

### Checklist


staffordsmith83 (approved):
I assume this is mainly to have access to the `mouseCoordinates` for the drop event? Or the addedItem parameters too? Not sure of reason for the PR so bit hard to assess :-) E.g. I cant see where `raiseFileDragDropEvent()` is called anywhere so I cant see if its being cleaned up. I suppose it will be used in plugins. 

Code looks good though, and handy for sure. Approved, but let me know of the purpose for future reference!

na9da:
Thanks for the review @staffordsmith83 . The function is mostly for plugins that want to listen to drag-n-drop events (eg: scene-editor).

PR #6753, opened 2023-04-05
Update the 3Dtiles init example.

meh9:
### What this PR does

Update the 3DTiles init file example to v8 init format, and use a dataset that works.

### Test me

Go to:

https://nationalmap.gov.au/#clean&https://raw.githubusercontent.com/TerriaJS/terriajs/3dtiles-example-update/wwwroot/test/init/cesium3dtiles.json

Open the catalogue and add the dataset.

Before it didn't work, now it does.

### Checklist

    There are no unit tests for this as far as I know.
    There's no relevant docs to update.
    This small change doesn't require a change entry as it's not part of the library functionality.

### DIFF
--- a/wwwroot/test/init/cesium3dtiles.json
+++ b/wwwroot/test/init/cesium3dtiles.json
@@ -1,13 +1,16 @@
 {
   "catalog": [
     {
-      "name": "Cesium 3D Tiles",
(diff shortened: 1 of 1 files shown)


na9da (approved):
Looks good. Thanks @meh9!

PR #6778, opened 2023-05-12
Fix magda access type resolution

na9da:
### What this PR does

This PR fixes `accessType` resolution for MagdaReference when `magdaRecord` is undefined.

Currently, when `magdaRecord` is set, MagdaReference correctly resolves the `acessType` for the model from the magda access headers. But when `magdaRecord` is undefined it always resolves the `acessType` as `public`. 

This PR fixes it so that when `magdaRecord` is unset, the `acessType` is resolved using [default terria strategy](https://github.com/TerriaJS/terriajs/blob/aa70e64f629881897ef3e6c091bb233365ef0e86/lib/ModelMixins/AccessControlMixin.ts#L30-L37) implemented by AcessControlMixin which may return `private` if a parent is private.

### Test me

Please test from DT (ping me), I have also added test cases.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 #### next release (8.2.29)
 
 - Fix app crash when rendering feature info with a custom title.
+- Fix `acessType` resolution for `MagdaReference` so that it uses the default terria resolution strategy when `magdaRecord` is not defined.
 - [The next improvement]
 
 #### 8.2.28 - 2023-04-28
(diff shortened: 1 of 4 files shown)


t83714:
@na9da do we have testing release terriajs that we can use to build & test it in dev vic cluster?

na9da:
Not right now @t83714 , can release to dev next week.

mwu2018 (approved):
I built an image based on this branch and deployed it to dev. It can be verified that all non-public groups have padlocks next to them. See https://dev.vic.digitaltwin.terria.io/ Need signing in.

na9da:
Awesome. Thanks @mwu2018 - that was so quick :zap:

PR #6779, opened 2023-05-12
Improve-calc-2d-view

### LINKED ISSUE #6777: Improve calculation of 2D view extents from upwards facing 3D camera views
See `Cesium.getCurrentCameraView()` specifically [this section](https://github.com/TerriaJS/TerriaMap/blob/e043699682c4aeb49a9a924e6bb0a474f57c97de/lib/Models/Cesium.ts#L1010-L1024)

```ts
 if (!center) {
      /** TODO: binary search to find the horizon point and use that as the center. 
      This will give a real `center` for camera views where the horizon is visible.
      But this will not be possible if camera is facing upwards and the horizon is not visible. 
      So we need to return a useful CameraView that works in 3D mode and 2D mode.
      In this case return the correct definition for the cesium camera, with position, direction, and up, 
      but for the required `rectangle` property, that will only be used when switching to 2D mode, provide the homeCamera view extent. 
      **/
      return new CameraView(
        this.terriaViewer.homeCamera.rectangle,
        camera.positionWC,
        camera.directionWC,
        camera.upWC
      );
    }
```

The problem I think is that our CameraView class requires rectangle property, which is view extent, so that 2D mode can work. But when camera is facing upwards, there is no view extent that can be defined on the globe surface. So we cannot supply a rectangle. The problem also exists when setting a view in 3D mode where the camera faces up, and then switching to 2D mode - the map defaults back to the homeCamera view, which could be considered a bug. 

Currently we have made sharelinks work by returning a CameraView object that has the homeCamera.rectangle  but the proper camera position to be used in 3d mode... So the sharelinks will work, but when switching to 2d mode the map will go back to the home position (which is no worse than current behaviour).

CameraView - Holds a camera's view parameters, expressed as a rectangular extent and/or as a camera position, direction, and up vector. - Rectangular extent is required. So we now return this:
```  
return new CameraView(
        this.terriaViewer.homeCamera.rectangle,
        camera.positionWC,
        camera.directionWC,
        camera.upWC
```

This issue was partially fixed in https://github.com/TerriaJS/terriajs/pull/6776 but could do with more improvements.

There is this gist that finds the horizon similar to what Kevin suggested but using linear search instead of binary search:
https://gist.github.com/mramato/a5e403379c4989313545#file-drawhorizonpoint-js-L17-L20
So this will be good for some of the edge cases, but not those where the horizon is not visible at all. For those I suggest to try implement something like cesium rectangular sensor? This library might help: https://github.com/jlouns/cesium-sensor-volumes

Stafford Smith
:house_with_garden:  [11:25 AM](https://terria.slack.com/archives/C4NCJT799/p1683775539174139)
Perhaps the best long term solution is to provide an extent rectangle (CameraView.rectangle) that represents the area of ground surface underneath the camera, projected with a 60deg field of view or similar. I will make a separate PR for this and tackle after DTV bugs.

staffordsmith83:
### What this PR does

Fixes #6777 and https://github.com/TerriaJS/vic-digital-twin/issues/487

Improved calculation of 2D view from 3D view. Previously the homeCamera was used when no horizon was visible to reliably calculate extent rectangle.

### Test me

Before: [Open this link, create a sharelink, and open that](http://ci.terria.io/main/#share=s-j2Ij3nAq3ufaTKUjCzkYtasTYpS)

When switching to 2D view, the map should lose its position and default to the homeCamera

After: [Open this link, create a sharelink, and open that](http://ci.terria.io/improve-calc-2d-view/#share=s-j2Ij3nAq3ufaTKUjCzkYtasTYpS)

When switching to 2D view, the map should be facing straight down from the previous 3D camera position

### Checklist

~~- [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~~
- [y] I've updated CHANGES.md with what I changed.
- [y] I've provided instructions in the PR description on how to test this PR.


lib/Models/Cesium.ts
```diff
@@ -994,6 +995,22 @@ export default class Cesium extends GlobeOrMap {
     };
   }
 
+  /**
+   * Helper method to clone a camera object
+   * @param camera
+   * @returns Camera
+   */
+  cloneCamera(camera: Camera): Camera {
```
na9da: This could be `private`

lib/Models/Cesium.ts
```diff
@@ -1004,13 +1021,43 @@ export default class Cesium extends GlobeOrMap {
...
     const center = isDefined(pickRay)
-      ? scene.globe.pick(pickRay, scene)
+      ? scene.globe.pick(pickRay, scene) // will be undefined if we are facing above the horizon
       : undefined;
 
     if (!center) {
-      // TODO: binary search to find the horizon point and use that as the center.
-      return this.terriaViewer.homeCamera; // This is just a random rectangle. Replace it when there's a home view available
-      // return this.terria.homeView.rectangle;
+      debugger;
```
na9da: Stray `debugger`

na9da (changes requested):
Hey @staffordsmith83,

The changes look great. I have just couple of comments.
Also please merge from `main` and remember to move the line in `CHANGES.md` to `8.3.1`.

na9da (approved):
Looks good @staffordsmith83 !

PR #6794, opened 2023-05-31
Fix search results for items inside deeply nested references

na9da:
### What this PR does

Fixes search results for items that are inside deeply nested references.

Currently if we have a `CatalogIndexReference` that points to a leaf item and the leaf item is inside the following catalog tree:
`MagdaReference -> TerriaReference -> Group -> LeafItem` then `CatalogIndexReference` fails to correctly load the `LeafItem`. 

This happens because the `CatalogIndexReference` only resolves references [one level deep](https://github.com/TerriaJS/terriajs/blob/ccd229665794dcd91f023cb0a01e8bc6a4bd6a3d/lib/Models/Catalog/CatalogReferences/CatalogIndexReference.ts#L73-L79). This PR fixes it by [recursively resolving references](https://github.com/TerriaJS/terriajs/blob/b78a716283f7b25921f3dcb7e0e2428c16adbebc/lib/Models/Catalog/CatalogReferences/CatalogIndexReference.ts#L73-L83) so that the whole chain until the `Group` container is fully expanded.

### Test me

- Open this link for [main](http://ci.terria.io/main/#configUrl=https://gist.githubusercontent.com/na9da/8716edc8207463233e1dac8454f031ac/raw/dcac31d1031e327bdbd6234a9c67e323dbe76f98/test-config.json) branch.
- Search for "ptv"
- Add any search result 
image

- Observe that an error is thrown
image

- Now repeat the same for [this branch](http://ci.terria.io/fix-index-ref-loading/#configUrl=https://gist.githubusercontent.com/na9da/8716edc8207463233e1dac8454f031ac/raw/dcac31d1031e327bdbd6234a9c67e323dbe76f98/test-config.json)
- Observe that the catalog item is correctly added to the workbench without errors

### Checklist

### DIFF
--- a/lib/ModelMixins/ReferenceMixin.ts
+++ b/lib/ModelMixins/ReferenceMixin.ts
@@ -145,13 +145,13 @@ function ReferenceMixin<T extends AbstractConstructor<BaseType>>(Base: T) {
      * @param maxDepth The maximum depth up to which the references should be resolved.
      * @returns A promise that is fulfilled with a Result value when all the references have been loaded.
      */
-    async recursivelyLoadReference(depth: number): Promise<Result<void>> {
+    async recursivelyLoadReference(maxDepth: number): Promise<Result<void>> {
       let currentTarget: BaseModel | undefined = this;
       const errors: TerriaError[] = [];
-      while (depth > 0 && ReferenceMixin.isMixedInto(currentTarget)) {
+      while (maxDepth > 0 && ReferenceMixin.isMixedInto(currentTarget)) {
         (await currentTarget.loadReference()).pushErrorTo(errors);
         currentTarget = currentTarget.target;
-        depth -= 1;
+        maxDepth -= 1;
       }
       const result = TerriaError.combine(
         errors,
--- a/lib/ModelMixins/ReferenceMixin.ts
+++ b/lib/ModelMixins/ReferenceMixin.ts
@@ -9,6 +9,7 @@ import DeveloperError from "terriajs-cesium/Source/Core/DeveloperError";
 import AbstractConstructor from "../Core/AbstractConstructor";
 import AsyncLoader from "../Core/AsyncLoader";
 import Result from "../Core/Result";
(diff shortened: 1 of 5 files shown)


lib/ModelMixins/ReferenceMixin.ts
```diff
@@ -135,6 +136,30 @@ function ReferenceMixin<T extends AbstractConstructor<BaseType>>(Base: T) {
...
+      while (maxDepth > 0 && ReferenceMixin.isMixedInto(currentTarget)) {
+        (await currentTarget.loadReference()).pushErrorTo(errors);
+        currentTarget = currentTarget.target;
+        maxDepth -= 1;
+      }
+      const result = TerriaError.combine(
+        errors,
+        "Failed to recursively load reference `${this.uniqueId}`"
+      );
+      return Result.error(result);
```
tephenavies:
```suggestion
      const maybeError = TerriaError.combine(
        errors,
        "Failed to recursively load reference `${this.uniqueId}`"
      );
      return Result.none(maybeError);
```

Use `Result.none` to convey that the result is normally an empty successful result. I'm not sure if the name `maybeError` is a bit too haskell'ish.

tephenavies (changes requested):

lib/ModelMixins/ReferenceMixin.ts
na9da: Good catch @steve9164!

I have changed it to the suggested pattern.

`Result.none` sounds a bit counter-intuitive to me though - from the name it sounds like it should always return an empty `Result`. Maybe it should be called `errorOrNone`?

tephenavies (approved):

lib/ModelMixins/ReferenceMixin.ts
tephenavies: It's not the best name. At some point I'd like to refactor Result a bit. I'll make a ticket and add that point

PR #6806, opened 2023-06-16
Add option `focusWorkbenchItems` to `initialCamera` setting.

na9da:
### What this PR does

Currently when the map has preloaded workbench items, there is no way to automatically focus the map on the workbench items. This PR adds a new option `focusWorkbenchItems` to `initialCamera` setting that will automatically zoom the map to show the workbench items.

If a subsequent init source or the share URL specifies a different `initialCamera` setting, that will take precedence and this option will be ignored.

Caveat: The current implementation focuses the first workbench item. This will be fixed to focus all workbench items when we refactor the zoom code path. [Issue here](https://github.com/TerriaJS/terriajs/issues/6807).

[Example init file](https://gist.githubusercontent.com/na9da/70dee24e06e994d340bfb3c1eb45140f/raw/b12543c00ea8dd12668fe5da4c7390b34e8c5498/focus-workbench-items.json).

### Test me

- Open this [CI link](http://ci.terria.io/focus-workbench-items/#https://gist.githubusercontent.com/na9da/2c78674c66903a932106983683411339/raw/c750222ba5bf77959fdf56cd8f23380128872e10/no-focus-workbench-items.json) for this branch using a config that does NOT enable `focusWorkbenchItems`
- Observe that the map loads the home camera view
- Now open this [CI link](http://ci.terria.io/focus-workbench-items/#https://gist.githubusercontent.com/na9da/70dee24e06e994d340bfb3c1eb45140f/raw/b12543c00ea8dd12668fe5da4c7390b34e8c5498/focus-workbench-items.json) with [focusWorkbenchItems](https://gist.github.com/na9da/70dee24e06e994d340bfb3c1eb45140f#file-focus-workbench-items-json-L3) enabled
- Observe that the map automatically zooms to the workbench dataset
- Ensure this works for both Cesium and Leaflet viewers

### Checklist


lib/Models/Terria.ts
```diff
@@ -1053,6 +1063,38 @@ export default class Terria {
...
+    ) as MappableMixin.Instance | undefined;
+    if (firstMappableItem) {
+      // When the app loads, Cesium/Leaflet viewers are loaded
+      // asynchronously. Until they become available, a stub viewer called
+      // `NoViewer` is used. `NoViewer` does not implement zooming to mappable
+      // items. So here wait for a valid viewer to become available before
+      // attempting to zoom to the mappable item.
+      const isViewerAvailable = () => this.currentViewer.type !== NoViewer.type;
+      await when(isViewerAvailable);
+      await this.currentViewer.zoomTo(firstMappableItem, 0.0);
```
tephenavies: Should this also check whether a viewer other than NoViewer is loading before calling `when`?

na9da: Not sure I understand why that check is required.

If a viewer is currently loading, then `when` will re-run the `isViewerAvailable` test whenever `currentViewer` is updated. Or is there something else I'm missing?

tephenavies: If `mainViewer.viewerMode` is undefined, then this.currentViewer.type will be permanently NoViewer and we'll have a hanging promise.

I think you could wait for `mainViewer.viewerLoadPromise`, check that the viewer isn't NoViewer afterwards then run zoomTo

na9da: @steve9164 - as discussed offline, I'm going with the option of keeping this code and leaving a note on hanging promise. 

Discussed reason: Because this piece of code runs before react renders, `viewerLoadPromise` will always resolve successfully to a `NoViewer` instance. So the suggested workaround will fail.

tephenavies (approved):
Great!

PR #6815, opened 2023-07-03
Refactor zoom for Cesium and Leaflet maps.

### LINKED ISSUE #6761: The behaviour when clicking the Ideal zoom button is different from the config field "zoomOnAddToWorkbench": true
### The behaviour when clicking the Ideal zoom button is different from the config field "zoomOnAddToWorkbench": true

Currently, the behaviour when clicking the Ideal zoom button is different from the config field "zoomOnAddToWorkbench": true (triggered by calling viewer.zoomTo()).

Consequently, when we set config `idealZoom.camera` and `"zoomOnAddToWorkbench": true` for a dataset, the camera will not be able to auto-move to the position that is specified by `idealZoom.camera` when adding it to the workbench. Instead, the camera will be moved to a fixed position regardless of the value set to `idealZoom.camera`.

The user will have to click the "Ideal Zoom" button to move the camera position to the position specified by `idealZoom.camera`.

#### Expected Behaviour

When config `idealZoom.camera` and `"zoomOnAddToWorkbench": true`, the camera should be moved to the position that is specified by `idealZoom.camera`, when the dataset is added to the workbench (same as Ideal zoom button).

Related ticket: https://github.com/TerriaJS/vic-digital-twin/issues/439

na9da:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/6761

Changes:
- Refactor zoom code path by moving the implementation of zoom behaviour for Cesium and Leaflet maps to [Cesium/Zooming.ts](https://github.com/TerriaJS/terriajs/blob/refactor-zoom/lib/Map/Cesium/Zooming.ts) and [Leaflet/Zooming.ts](https://github.com/TerriaJS/terriajs/blob/refactor-zoom/lib/Map/Leaflet/Zooming.ts). We could use a similar pattern to refactor picking code path by separating it to `Map/Leaflet/Picking.ts` and `Map/Cesium/Picking.ts`.
- Separate code for 3D/2D zoom view computation from the actual zooming mechanism. We now  have [compute3dZoomView](https://github.com/TerriaJS/terriajs/blob/0c31385e7c1879e19dcb96bd9f5dce3f1b52bcf2/lib/Map/Cesium/compute3dZoomView.ts#L39-L57) and [compute2dZoomView](https://github.com/TerriaJS/terriajs/blob/0c31385e7c1879e19dcb96bd9f5dce3f1b52bcf2/lib/Map/Leaflet/compute2dZoomView.ts#L29-L42) functions that computes the zoom views for [different target types](https://github.com/TerriaJS/terriajs/blob/88cd768dc4269a43242758cde118399a51a2bb78/lib/Map/ZoomTarget.ts#L9-L13). This makes it easier to understand and also improves extensibility without touching other parts of the code.
- Improved zooming to clamped DataSource entities:
  https://github.com/TerriaJS/terriajs/blob/0ba6212b0f5ffce40c8a1e85edd9272316b836f8/lib/Map/Cesium/compute3dZoomView.ts#L239-L250
 This replicates a similar change to the [Cesium Viewer](https://github.com/CesiumGS/cesium/issues/10555#issuecomment-1204045063). And here is a [before](http://ci.terria.io/main/#share=s-zmnGfbmvgLW9lXEOFJhdLE9DucE&clean&https://gist.githubusercontent.com/na9da/fb1bf736532ad280314c9bc438653fe0/raw/9b26b185418b58a6d03185a15fe9fac5719944d4/ducky.json) and [after](http://ci.terria.io/refactor-zoom/#share=s-zmnGfbmvgLW9lXEOFJhdLE9DucE&clean&https://gist.githubusercontent.com/na9da/fb1bf736532ad280314c9bc438653fe0/raw/9b26b185418b58a6d03185a15fe9fac5719944d4/ducky.json) example. See how when zooming to the Ducky, in the before example it is nowhere to be seen.
- Add option to zoom to an array of `MappableMixin.Instance` types.
- More spec coverage

### Test me

- Open this [CI link](http://ci.terria.io/refactor-zoom/) and randomly test different types of items to see if they work as expected

### Checklist


lib/ReactViews/Workbench/Controls/ViewingControls.tsx
```diff
@@ -146,76 +146,12 @@ class ViewingControls extends React.Component<
...
-      };
-
-      zoomToView = CameraView.fromJson(camera);
-    } else if (
-      item.rectangle?.east !== undefined &&
-      item.rectangle?.west !== undefined &&
-      item.rectangle.east - item.rectangle.west >= 360
-    ) {
-      zoomToView = this.props.viewState.terria.mainViewer.homeCamera;
-      console.log("Extent is wider than world so using homeCamera.");
```
na9da, line 214: The `idealZoom` behaviour has been folded into the rest of the zoom code path.

lib/Models/Terria.ts
```diff
@@ -1094,7 +1094,7 @@ export default class Terria {
       // a hanging promise if a valid viewer never becomes available,
       // for eg: when react is not rendered - `currentViewer` will always be `NoViewer`.
       await when(isViewerAvailable);
-      await this.currentViewer.zoomTo(firstMappableItem, 0.0);
+      await this.currentViewer.zoomTo(mappables, 0.0);
```
na9da, line 1097: Instead of zooming to first mappable in the workbench, we zoom to a view showing all the mappables.

nf-s:
This PR also fixes

- https://github.com/TerriaJS/terriajs/issues/6303

lib/Models/Cesium.ts
```diff
@@ -842,124 +836,7 @@ export default class Cesium extends GlobeOrMap {
   }
 
   doZoomTo(target: any, flightDurationSeconds = 3.0): Promise<void> {
```
nf-s, line 838: Can you please type `target` here.

Also, can you please use `ZoomTarget` here 

https://github.com/TerriaJS/terriajs/blob/bdb0b606b1e59b82731c9afbc7c2e5d8530562f7/lib/Models/NoViewer.ts#L28-L31

lib/Map/Cesium/compute3dZoomView.ts
```diff
@@ -0,0 +1,343 @@
...
+import CameraView from "../../Models/CameraView";
+import Cesium from "../../Models/Cesium";
+import { ZoomTarget } from "../ZoomTarget";
+
+/**
+ * Result of {@link compute3dZoomView} function. Something which can be zoomed to in 3D.
+ *
+ * TODO: Can we can simplify by reducing everything to just CameraView?
+ */
+export type ZoomView3d = BoundingSphere | Cartesian3 | CameraView;
```
nf-s, line 32: I think keeping `ZoomView3d` flexible will might help us in the long run - as then it will be easier to implement zoom behaviour for new catalog items/mappable items.

But there is a trade off in therms of maintenance costs. What do you think?

nf-s (changes requested):
Looks great! This is so much cleaner than before. I just have two small comments

I tested with a bunch of catalog items and could only find one issue with this:

> Improved zooming to clamped DataSource entities:
> here is a [before](http://ci.terria.io/main/#share=s-zmnGfbmvgLW9lXEOFJhdLE9DucE&clean&https://gist.githubusercontent.com/na9da/fb1bf736532ad280314c9bc438653fe0/raw/9b26b185418b58a6d03185a15fe9fac5719944d4/ducky.json) and [after](http://ci.terria.io/refactor-zoom/#share=s-zmnGfbmvgLW9lXEOFJhdLE9DucE&clean&https://gist.githubusercontent.com/na9da/fb1bf736532ad280314c9bc438653fe0/raw/9b26b185418b58a6d03185a15fe9fac5719944d4/ducky.json) example. See how when zooming to the Ducky, in the before example it is nowhere to be seen.

This works for me in "3d terrain" mode, but in "3d smooth" mode I still get put under the globe. This may be out-of-scope for this PR, if so let me know and I can make a ticket

PR #6826, opened 2023-07-26
Wps results fix

### LINKED ISSUE #6392: WPS results are showing columns twice after sharing
https://beta.geo-rapp.terria.io/#share=s-eLoi6D4gSxt6RWMxWgRlxJD0D0s

image

sixlighthouses:
### Fix WPS duplicate 

Fixes #6392

Workbench items were being duplicated when loaded from WPS results in share URL.  
Added a check in models/Terria.ts when loading items to ensure duplicates are not added

### Test me

Load this share URL http://ci.terria.io/wps-results-fix/#share=s-uCLMJuXtXBpFXvmJbBI4gE4rGUz

Two items should be loaded to the workbench 
image

### Checklist


na9da:
Hey @sixlighthouses, that's some good sleuthing :detective: job.

The root cause is that our workbench implementation is not protected from duplicate items - that might need a rethink at some point.

The fix you made works, but  I would suggest changing [newItems](https://github.com/TerriaJS/terriajs/blob/e46799735edc134acf49856995e2f263ea0556f3/lib/Models/Terria.ts#L1799) to a [Set&lt;BaseModel&gt;](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead? It'll automatically guard against similar bugs in other code paths. You can then convert the set to array using `[...newItems]`.

sixlighthouses:
thanks @na9da a set is a nice solution will make that change

na9da:
Hey @sixlighthouses, on another thought the better fix would be to make the set method for [workbench.items](https://github.com/TerriaJS/terriajs/blob/e46799735edc134acf49856995e2f263ea0556f3/lib/Models/Workbench.ts#L36-L38)  sanitize its input by de-duplicating the input array. You could use the lodash `uniq` method to do that.

Sorry if you already started on the fix.

sixlighthouses:
thanks @na9da updated the workbenches items set method

lib/Models/Terria.ts
```diff
@@ -1802,7 +1802,10 @@ export default class Terria {
     while (true) {
       const model = newItemsRaw.shift();
       if (model) {
-        await this.pushAndLoadMapItems(model, newItems, errors);
+        // check if model is already in the newItems list by uniqueId
+        if (!newItems.find((item) => item.uniqueId === model.uniqueId)) {
+          await this.pushAndLoadMapItems(model, newItems, errors);
+        }
```
na9da: I guess we could remove this check now that our workbench will weed out the duplicates.

lib/Models/Workbench.ts
```diff
@@ -34,7 +34,12 @@ export default class Workbench {
     return this._items.map(dereferenceModel);
   }
   set items(items: readonly BaseModel[]) {
-    this._items.spliceWithArray(0, this._items.length, items.slice());
+    const setItems = new Set(items);
+    this._items.spliceWithArray(
```
na9da, line 39: Probably add a comment here to say that we want the workbench items to be unique.

na9da:
Looks good @sixlighthouses. I have just made a couple of comments.

CHANGES.md
```diff
@@ -1,10 +1,14 @@
...
-#### next release (8.3.2)
+#### next release (8.3.3)
 
 - Fixed a bug when restoring timefilter from a share link having more than one imagery item with the same base URL (but different layer names).
 - [The next improvement]
 
+#### 8.3.2
+
+- Fix WPS duplicate display of analysis results when loaded through a share URL
+
```
na9da: I think the correct next release for this change is `8.3.2` (so after the timefilter bug fix).

https://github.com/TerriaJS/terriajs/blob/15d468656bee43b0f872075cbf867475993e2536/CHANGES.md?plain=1#L3-L7

lib/Models/Terria.ts
sixlighthouses: reverted back to previous method

lib/Models/Workbench.ts
sixlighthouses: good idea, commented

na9da (approved):
Good to go :rocket:

PR #6838, opened 2023-08-08
Clipping box interim redesign

### LINKED ISSUE #6836: Investigate short term solution for locating clipping box

na9da:
### What this PR does

Fixes #6836 

This PR introduces the following temporary UX fixes for clipping box:

* An option to zoom to clipping box
* An option to re-position the clipping box
* Trigger repositioning of clipping box when the user enables clipping box for the first time
* Cursor and scale point handle changes (makes it much easier to grasp)
* More robust interaction with the box

Note that this is a temporary fix which will be removed in a later re-design, so I have cut corners in a few places by accessing `private` members to get things working.

I have also tested with scene-editor to make sure the changes are compatible.

Changes:
image

image

### Test me

- Open this [CI link](http://ci.terria.io/zoom-to-clipping-box/#share=s-8gsJuhul1IrAfitEIvu3BCeZA7Z&clean&https://gist.githubusercontent.com/na9da/737c818d6ca2ee0d22147181ae6eba0c/raw/9bd838001fd5820e65791a8e0892d29fa6005f92/melb-central-ion.json)
- Enable clipping box for the workbench item
- Observe that the reposition workflow is triggered
- Move the mouse and click on map to place it
- Navigate away from the box
- Click the `Zoom to clipping box` button to return to the clipping box
- Ensure the other features for clipping box continue to work

### Checklist

- [ ] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ I have not added tests because this will be a temporary fix an will be overridden in the future.


lib/ReactViews/SelectableDimensions/Button.tsx
```diff
@@ -6,29 +6,27 @@ import { StyledIcon } from "../../Styled/Icon";
...
-            styledWidth="16px"
-            styledHeight="16px"
-          />
-        ))
+      renderIcon={() =>
+        iconGlyph === "spinner" ? (
+          <AnimatedSpinnerIcon {...iconProps} />
+        ) : iconGlyph ? (
+          <StyledIcon glyph={iconGlyph} {...iconProps} />
+        ) : undefined
```
na9da, line 29: Added support for showing a regular icon glyph or an animated spinner icon.

lib/Traits/TraitsClasses/ClippingPlanesTraits.ts
```diff
@@ -125,7 +125,7 @@ export class ClippingBoxTraits extends mixTraits(BoxDrawingTraits) {
     name: "Keep clipping box above ground",
     description: "When true, prevents the box from going underground."
   })
-  keepBoxAboveGround = true;
+  keepBoxAboveGround = false;
```
na9da: Switch default option to allow box to move below ground.

lib/ModelMixins/ClippingMixin.ts
```diff
@@ -257,107 +274,194 @@ function ClippingMixin<T extends AbstractConstructor<BaseType>>(Base: T) {
...
               id: "true",
-              name: i18next.t("models.clippingBox.clipModel")
+              name: `${i18next.t("models.clippingBox.clipModel")}`
             },
             {
               id: "false",
               name: i18next.t("models.clippingBox.clipModel")
             }
           ],
-          setDimensionValue: (stratumId, value) => {
```
na9da, line 276: The below change is just from moving code around.

philipgrimmett:
## General clipping suggestions

- [x] Display help text in workbench when selecting box location: '**_Click on map to position clipping box_**' (hallway testing indicated that users may need the prompt and don't immediately think to look at the map)
- [x] Display matching help text in mouse tooltip
- [x] Make mouse tooltip fully opaque to increase legibility
- [x] Hide workbench config options until clipping box is positioned on map and while box is being repositioned (show help text until new position is confirmed or cancelled)
- [x] Hide workbench options while clipping box is hidden ie. clipping mode, zoom to, reposition
- [x] Increase handle size and click area
- [x] If possible, add subtle hover state to box planes eg. +0.1 opacity

## PR-specific suggestions

- 'Zoom to' is a generic function that can appear in multiple locations (workbench, stories, maybe feature info?). Ideally the zoom target should be communicated through visual hierarchy, reducing the need to explicitly state the parent name eg. Zoom to 'clipping box' (although the target should be referenced in alt text/tooltip). Same reasoning applies to 'reposition'. Slightly worse UX, but more sustainable for workspace real estate. Happy to discuss.

## Bugs
- Activating clipping seems to break feature info. Disabling clipping doesn't seem to fix the bug, refresh required (tested in Arc/Chrome, Safari)

## Notes

- If clipping box is positioned partially below ground while 'above ground only' is disabled, enabling 'above ground only' does not immediately update the vertical position (snaps above ground when user interacts with any plane)
- Pressing ESC to cancel also exits full screen mode (tested in Arc)

## Figma prototype link

https://www.figma.com/proto/jtdgEiY0gxtZlGtxSJTzcZ/Terria---2023?page-id=11635%3A2936&type=design&node-id=11635-7700&viewport=-4945%2C-8742%2C0.9&t=uwIEL9EPHKiGpJPW-1&scaling=min-zoom&starting-point-node-id=11635%3A2937&mode=design

sixlighthouses:
@na9da this is a really nice piece of work, I added a minor commit to cleanup the simpleClippingPlaneCollection() function in ClippingMixin.ts

sixlighthouses (approved):

PR #6844, opened 2023-08-17
Allow story body to use html tag iframe.

### LINKED ISSUE #6843: Fix broken embedded video link in story panel
A user is able to create a story with embedded video. However, when the story is loaded, the video part is dropped.

Related issue https://github.com/TerriaJS/vic-digital-twin/issues/541.

mwu2018:
### What this PR does

Fixes #6843 
The `dompurity` module thinks `iframe` tag is unsafe and is ignored by its default configuration. This PR add `iframe` as a safe tag in `StoryBody` if only the video clips to be embedded are from the following three media source domains with specified paths:
- https://www.youtube.com/embed/
- https://www.youtube-nocookie.com/embed/
- https://player.vimeo.com/video/

**Note**
The story editor allows user to insert a picture url but it will not be played back -- a different issue.

### Test me

Try to create and **SAVE** a story with embedded video, then **PLAY** the story, using [before](http://ci.terria.io/main) and [after](http://ci.terria.io/issue-6843) the fix.

### Checklist


na9da:
Some security context on allowing users to embed arbitrary `iframe` tags:
 - https://github.com/cure53/DOMPurify/issues/566
 - https://stackoverflow.com/questions/7289139/why-are-iframes-considered-dangerous-and-a-security-risk

To support embedding videos, one safer solution might be to deny embedding `iframes` by default and safelist iframes from a few known sites like youtube or vimeo. Additional config options could be provided for the map owner to allow other sites.

cc @steve9164

na9da:
[Example](https://github.com/cure53/DOMPurify/issues/340#issuecomment-670758980) of how to restrict `iframe` usage by `src`.

mwu2018:
@na9da @steve9164 Could you please take another look? The iframe tag is allowed only if the source is from Youtube.

nf-s:
@steve9164 to check iframe attributes...

lib/ReactViews/Story/StoryPanel/StoryBody.tsx
```diff
@@ -45,6 +45,48 @@ const StoryContainer = styled(Box).attrs((props: { isCollapsed: boolean }) => ({
...
+
+function sourceBasedParse(story: Story) {
+  if (shouldAddIframeTag(story)) {
+    return parseCustomHtmlToReact(
+      story.text,
+      { showExternalLinkWarning: true },
+      false,
+      {
+        ADD_TAGS: ["iframe"],
+        ALLOWED_ATTR: ["src", "width", "height"]
```
tephenavies: This will restrict attributes for all tags to only these 3. This will stop e.g. embedded svgs from rendering, but also many other examples. See second scene of http://ci.terria.io/issue-6843/#share=s-ejPZyk2EJitArJp0T2naTpnok5Y&playStory=1 compared to the second scene of http://ci.terria.io/main/#share=s-ejPZyk2EJitArJp0T2naTpnok5Y&playStory=1 (the blue svg is not rendered any more).

tephenavies: Other examples include text color added using scene editor being ignored

mwu2018: Not sure why `svg` tag is removed. Needs investigation.

mwu2018: This restriction `ALLOWED_ATTR: ["src", "width", "height"]` does not behave as expected. It did not remove the svg tag. Instead, it removes all properties of `svg.g.path`.

Removing this restriction will work.

tephenavies (approved):

PR #6846, opened 2023-08-18
Timezones fix

### LINKED ISSUE #6817: Add timezone override trait

sixlighthouses:
### What this PR does

Fixes #6817

Add the ability to set a timeZone offset, dateFormat and isStaticDate

isStaticDate can be used to force a datetime to not be transformed to users local time

#### item example

```
    { 
      "id": "IUpoiafh",
      "name": "Landsat 5-9 Collection 2",
      "layers": "landsat5_c2l2_sr",
      "url": "https://ows.datacubechile.cl/wms",
      "type": "wms",
      "opacity": 1,
      "timeZone": "-3:00",
      "dateFormat": "isoDateTime",
      "isStaticDate": false,
    }
```

Created new trait class which extends ModelTraits
```DateTimeTraits```

has 3 optional traits

```timeZone``` which takes an UTC offset with the format "+10:00" 
```dateFormat``` see useage at -- https://github.com/felixge/node-dateformat
```isStaticDate``` boolean -- false or not set datetimes will be transformed to users locale (default browser behaviour).  true datetimes will be displayed as per source data (no time transformation)

### Test me

http://ci.terria.io/timezones/#clean&https://gist.githubusercontent.com/sixlighthouses/a9d1e72fba0002caa068dd90418be5bc/raw/d39d3ff17e3a615277cbfdaf4bc2b728a71e1732/test.json

View the layers 

- Landsat 5-9 IsoDate with timezone -03:00
- Landsat 5-9 IsoDateTime no timezone
- Landsat 5-9 String format no timezone
- Landsat 5-9 String format with timezone

The display of each of the layers datetime data in the workbench and timeline are defined by the following config

```
{ 
      "id": "IUpoiafh",
      "name": "Landsat 5-9 IsoDate with timezone -03:00",
      "layers": "landsat5_c2l2_sr",
      "url": "https://ows.datacubechile.cl/wms",
      "type": "wms",
      "opacity": 1,
      "dateFormat": "isoDate",
      "timeZone": "-03:00"
    },
    { 
      "id": "e43c635f-aa1e-4d16-9777-b8484ea10854",
      "name": "Landsat 5-9 IsoDateTime no timezone",
      "layers": "landsat5_c2l2_sr",
      "url": "https://ows.datacubechile.cl/wms",
      "type": "wms",
      "opacity": 1,
      "dateFormat": "isoDateTime"
    },
    { 
      "id": "5281d54a-5d05-4033-aa4a-63a0cbde402f",
      "name": "Landsat 5-9 String format no timezone",
      "layers": "landsat5_c2l2_sr",
      "url": "https://ows.datacubechile.cl/wms",
      "type": "wms",
      "opacity": 1,
      "dateFormat": "dddd, mmmm dS, yyyy, h:MM:ss TT"
    },
    { 
      "id": "a5649b8e-e114-44fd-bb84-a0fad1f63b25",
      "name": "Landsat 5-9 String format with timezone",
      "layers": "landsat5_c2l2_sr",
      "url": "https://ows.datacubechile.cl/wms",
      "type": "wms",
      "opacity": 1,
      "dateFormat": "dddd, mmmm dS, yyyy, h:MM:ss TT",
      "timeZone": "-03:00"
    },
```

### Checklist

### CODE REVIEW

**Review by nf-s — CHANGES_REQUESTED** [review]
Thanks for the PR!

I think we may need to approach this problem in a more holistic way. The way Terria handles time is a bit of a mess, so it makes it really hard to add new features like this.

For example - here there are dates with inconsistent formatting

- `Timeline` label is showing correct format
- `CesiumTimeline` tick labels is showing old format
- `DateTimePicker` label is showing correct format
- `DateTimePicker` items is showing old format
- Chart times are showing old format

image

I also think that if we are applying timezone transformations per catalog item, we need to add the timezone to the date string by default (eg `Monday, April 2nd, 2001, 7:00:00 AM (AEST)`).

As for the motivation behind this PR, I'm not sure we should be making changes due to incorrect handling of dates service-side. I added a comment about the Chile Opendatacube here - https://github.com/TerriaJS/product/issues/285#issuecomment-1742334077

**Comment by nf-s** [comment]
I'm not sure of the best way to move forward with this, so it would be good for us to have a meeting with @steve9164 and @na9da

**Comment by nf-s** [comment]
I think I have figured this out.

The `dateformat` package has an option to print out times in UTC - which can be configured in `TimeVaryingTraits` our current `main` branch - for example

```json
{
    "dateFormat": "UTC:dd/mm/yyyy, HH:MM:ss Z"
{
```

This means that regardless of your timezone - you will see the UTC time that is stored on the server.

This example shows the same layer twice - but the first (left) layer is showing UTC time - the second layer is showing in the local timezone on my computer (GMT-3 for Santiago, Chile)

image

Share link to test - http://ci.terria.io/main/#share=s-6VZEvzzo3KFpaW3C78d84DZEY3z

nf-s:
I'm going to close this - as the problem this was trying to solve ended up being a opendatacube-ows problem

Context - https://github.com/TerriaJS/terriajs/discussions/6968

PR #6857, opened 2023-08-31
Adds EPSG:7855 to known projections.

na9da:
### What this PR does

Adds EPSG:7855 to known projections.
This should fix ideal zoom for some layers.

### Test me

Check ideal zoom [before](http://ci.terria.io/main/#share=s-ofGTlN9sg2JsYR9J0Y2IgOzeWbO) and [after](http://ci.terria.io/add-proj-7855/#share=s-ofGTlN9sg2JsYR9J0Y2IgOzeWbO)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 #### next release (8.3.3)
 
 - Fixes broken point dragging interaction for user drawing in 3d-mode.
+- Added EPSG:7855 to `Proj4Definitions`.
 - [The next improvement]
 
 #### 8.3.2 - 2023-08-11
--- a/lib/Map/Vector/Proj4Definitions.ts
+++ b/lib/Map/Vector/Proj4Definitions.ts
@@ -57,7 +57,9 @@ const Proj4Definitions: Record<string, string> = {
   "EPSG:102100": "EPSG:3857",
   "EPSG:27700":
     "+proj=tmerc +lat_0=49 +lon_0=-2 +k=0.9996012717 +x_0=400000 +y_0=-100000 +ellps=airy +towgs84=446.448,-125.157,542.06,0.15,0.247,0.842,-20.489 +units=m +no_defs",
-  "EPSG:7844": "+proj=longlat +ellps=GRS80 +no_defs +type=crs"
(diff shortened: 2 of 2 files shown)


nf-s (approved):
LGTM!

I can't remember much about this stuff, but is there a reason why some use of [`Reproject` ](https://github.com/TerriaJS/terriajs/blob/main/lib/Map/Vector/Reproject.ts) uses the terriajs [`Proj4Definitions`](https://github.com/TerriaJS/terriajs/blob/main/lib/Map/Vector/Proj4Definitions.ts) instead of the terriajs-server [`proj4lookup`](https://github.com/TerriaJS/terriajs-server/blob/master/lib/controllers/proj4lookup.js) service?

For example - in `GeoJsonMixin` the CRS needs to be in the `proj4lookup` service in order to reproject features

https://github.com/TerriaJS/terriajs/blob/ed7c97fc722a1719e3b24f661bdde41c6bf7edce/lib/ModelMixins/GeojsonMixin.ts#L1436-L1458

It might be worth adding `EPSG:7855` to terriajs-server too

Also, this ties into the general issues with GDA2020 and GDA94 that we haven't resolved - in that the transformation from GDA2020 to WGS84, and GDA94 to WGS84 is null - but transformation from GDA94 to GDA2020 is not null.

But that is definitely for another day 🙂

na9da:
Good observations @nf-s. 

I think the `Proj4Definitions` is a [carry over from v7](https://github.com/TerriaJS/terriajs/blob/terriajs7/lib/Models/ArcGisMapServerCatalogItem.js). It is possible that the `Reproject` module was added later. We could look at switching to Reproject everywhere in code. In this case though, it might involve an `async` step to call [Reproject.checkProjection](https://github.com/TerriaJS/terriajs/blob/5d100803807df188b66764b82acac8985043b6cc/lib/Map/Vector/Reproject.ts#L81) so that a projection not available locally gets imported from the server. 

Also I think, the Reproject API could be rewritten as `reproject` and `reprojectAsync` or something to avoid a `checkProjection` step. We could even split the projection code into a new package shared between the server and client. I feel it is best done as part of some other bigger refactoring work.

na9da:
@nf-s terriajs-server change is here - https://github.com/TerriaJS/terriajs-server/pull/150

PR #6872, opened 2023-09-05
Support time window query

### LINKED ISSUE #6873: Support time window query
Some services, e.g. esri map service, support time interval query but the current code does not have this capability.

mwu2018:
### What this PR does

Fixes #6873 
Support time window query for esri-mapServer type.

- Add `timeWindowDuration` and `timeWindowUnit` traits to type `esri-mapServer` to support time query with window.
- A positive `timeWindowDuration` indicate forward time window while negative value backward window.
- The `timeWindowUnit` can be any units recognised by `moment` module. If an invalid value is given, will query time without window.

### Test me
[before](http://ci.terria.io/main/#clean&https://gist.githubusercontent.com/mwu2018/a8dbf3d7a165d96633713d9e90b0d214/raw/fd17acb7dea34a788be9dfe42751da79a59462bf/test-esri-mapserver-without-time-interval.json)

[after](http://ci.terria.io/support-time-interval/#clean&https://gist.githubusercontent.com/mwu2018/0d48fa21ddb596dd1562e39f1e385280/raw/48523cedaf84dc02a1174c1b89ad9580c3e7b3de/test-esri-mapserver-time-interval.json)

### Checklist


lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
```diff
@@ -393,14 +400,62 @@ export default class ArcGisMapServerCatalogItem extends UrlMixin(
...
+    if (
+      this.modelDimensions.length < 1 ||
+      this.modelDimensions[0].selectedId === undefined
+    ) {
+      return undefined;
+    }
+    const selectedId = this.modelDimensions[0].selectedId;
+    if (this.modelDimensions.length < 1 && selectedId === undefined) {
+      return undefined;
+    } else {
```
na9da: When using `modelDimensions` you don't have to parse it manually. When the user selects an option from the drop down menu, the [traits get set automatically](https://github.com/TerriaJS/terriajs/blob/b19ebd44715854237c1708f844d0b1ebb86edddd/lib/ModelMixins/CatalogMemberMixin.ts#L200-L204). So I think you can remove this function.

lib/Traits/TraitsClasses/TimeVaryingTraits.ts
```diff
@@ -71,4 +71,34 @@ export default class TimeVaryingTraits extends mixTraits(CatalogMemberTraits) {
...
+      " * `minute` <br/>" +
+      " * `second` <br/>",
+    type: "string"
+  })
+  timeUnit?: string;
+
+  @primitiveTrait({
+    name: "Is Forward",
+    description:
+      "If true, the interval is current time forward. Otherwise backward. Default to true (forward).",
```
na9da: `TimeVaryingTraits` gets pulled into a lot of items. Since we are only implementing this functionality for `ArcGisMapServerCatalogItem`, it is better if the new traits are added to just the [ArcGisMapServerCatalogItemTraits](https://github.com/TerriaJS/terriajs/blob/main/lib/Traits/TraitsClasses/ArcGisMapServerCatalogItemTraits.ts) class.

na9da (approved):
@mwu2018 - I have partially reviewed the PR and think it is a great start and good work on the specs.

In the context of ArcGis MapServer, `intervals` actually mean the time step between two dates in the date selector (or the timeline). Whereas the concept we are implementing here - the window of data rendered on the map - is called a [time window](https://desktop.arcgis.com/en/arcmap/latest/map/time/setting-a-time-window.htm). So ideally we should rename the traits to something that aligns with the original terms. 

Another thing we could do is combine the units and time window setting into one. We already use momentjs in a few places so you could use that to [parse](https://momentjs.com/docs/#/durations/creating/) a string like `10s` or `1M`. 

If we do that you could name the trait as `timeWindowDuration` and accept strings like `1M` `1Y`. The direction trait could be named as `timeWindowDirection`.

Let me know what you think.

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
mwu2018: This function will be used to check if we want to do time window query. It is not to set traits.

lib/Traits/TraitsClasses/TimeVaryingTraits.ts
mwu2018: I think these optional time window traits are still useful for other time based data type. Don't you think so?

mwu2018:
> momentjs

The doc says momentjs is not recommended for new projects.

mwu2018:
@na9da I've revised partially based on your comments. Happy to discuss.

lib/Traits/TraitsClasses/TimeVaryingTraits.ts
nf-s: I think it is better to not add `Traits` to items unless they are implemented. By adding these to `TimeTraits`, it means they will show up in docs.terria.io for most Catalog Items - even though they are only functional for `ArcGisMapServerCatalogItem`

mwu2018: This is a good point. How about putting them into a `TimeWindowTraits` to be used by any types that have its implementation?

na9da: Agree with @nf-s. 

I think it is best to define the traits in `ArcGisMapServerCatalogItemTraits` for now - mainly because you never know when or if we are going to implement time window for other time varying items. 

When we get to that point it is pretty easy to pull these traits into a common class like you suggested.

na9da:
> The doc says momentjs is not recommended for new projects.

That is correct, though the project hasn't been deprecated and we [already](https://github.com/TerriaJS/terriajs/blob/4f419c5f23203e023683532b67057082358bd666/lib/Core/createDiscreteTimes.ts#L1) [use](https://github.com/TerriaJS/terriajs/blob/4f419c5f23203e023683532b67057082358bd666/lib/ReactViews/BottomDock/Timeline/DateTimePicker.tsx#L10) it in terriajs. 

I think it should be ok to use it for now until we switch to something else - https://github.com/TerriaJS/terriajs/issues/3350

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
na9da: I probably didn't explain it well enough. The model dimensions are used for automatically setting trait values when the user selects a value from the dropdown. It is not meant to be parsed in code. There is also no need to parse it because you can directly access the `this.timeWindowDuration` trait value and expect it to be the value set by the user from the dropdown.

So I would remove this function and instead rewrite `getTimeParams` as something like:
```
private getTimeParams(): TimeParams {
    ...
    return {currentTime, timeWindowDuration: this.timeWindowDuration}
  }
```

Let me know if that makes sense, happy to chat.

mwu2018: In the unit test, after updating model with this [json](https://github.com/TerriaJS/terriajs/blob/0d1f71b3f91fca42f1901d4c1423a98e11b5944b/test/Models/Catalog/esri/ArcGisMapServerCatalogItemSpec.ts#L514), the model has `modelDimensions` trait set only. How do I have `timeWindowDuration` and `timeUnit` set automatically without `modelDimensions` being parsed?

test/Models/Catalog/esri/ArcGisMapServerCatalogItemSpec.ts
```diff
@@ -467,6 +468,115 @@ describe("ArcGisMapServerCatalogItem", function () {
...
+                value: {
+                  timeWindowDuration: 2,
+                  timeUnit: "week"
+                }
+              }
+            ],
+            selectedId: "2 weeks"
+          }
+        ]
+      };
```
na9da: You don't need to test `modelDimensions` specifically. Instead you can change the `modelJson` to simply `{"timeDurationWindow": "..."}` or simply do `item.setTrait(CommonStrata.user, "timeWindowDuration", "...")`. Same for the other specs.

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
na9da: See comment here - https://github.com/TerriaJS/terriajs/pull/6872#discussion_r1320925646

test/Models/Catalog/esri/ArcGisMapServerCatalogItemSpec.ts
mwu2018: It appears that if an item has `selectedId` set in `modelDimensions`  in the catalog json, the item's traits `timeWindowDuration` and `timeUnit` are still undefined after being added to the workbench. The user has to explicitly select an option from the drop-down menu to update the model. Another problem is that if an option is selected, it can not be unselected by choosing an empty option value. E.g. the following option will do nothing

```
{
   "id": "no time window",
    "value": { }
}
```

na9da: You are right. That is because the `modelDimensions` is meant as an easy way to generate a UI for updating traits. So it'll change the trait value only when a user interacts with it. (Without `modelDimensions` you'll have to write a new React component for every custom configuration)

If you want to configure a default time window configuration, you can do like you normally configure any trait value - by specifying it in the item catalog definition. eg:
```json
{
   "type": "esri-mapServer",
   "url": "...",
   "timeWindowDuration": "4d"
}
```

> Another problem is that if an option is selected, it can not be unselected by choosing an empty option value. E.g. the following option will do nothing

You should be able to use `null` values to unset a trait.
eg: 
```json
{
   "id": "no time window",
    "value": { "timeWindowDuration": null }
}
```

mwu2018: It can not be `null` because its type is `number` or `undefined`. I have to do this:
```
{
  "id": "no time frame",
  "value": {
    "timeWindowDuration": 0,
    "timeUnit": "year"
  }
}
```

na9da: You could try defining the trait with `isNullable: true` to allow `null` values:

```js
@primitiveTrait({
    name: "Time Window Duration",
    description:
      "Specify a time window duration from the current time that this dataset is available. Positive: forward; negative: backward.",
    type: "number",
    isNullable: true,
  })
  timeWindowDuration?: number;
```

mwu2018: This approach seems not working as it will not try to update the model if the trait is set null in the `modelDimenstions`.

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
```diff
@@ -452,9 +475,52 @@ export default class ArcGisMapServerCatalogItem extends UrlMixin(
...
+        windowDuration: number | undefined,
+        timeUnit: string | undefined
+      ): number | undefined {
+        if (
+          windowDuration === undefined ||
+          windowDuration === 0 ||
+          timeUnit === undefined
+        ) {
+          return undefined;
+        }
```
na9da: Just a nitpick. I think we can avoid checks like these, if we make `TimeParams` stricter like:

```ts
interface TimeParams {
  currentTime: number;
  timeWindowDuration: number;
  timeWindowUnit: number;
}```

and make `getTimeParams()` return `getTimeParams(): TimeParams | undefined` if any one of the field is invalid.

File: lib/Traits/TraitsClasses/ArcGisMapServerCatalogItemTraits.ts [hunk 5]
```diff
@@ -57,4 +57,20 @@ export default class ArcGisMapServerCatalogItemTraits extends mixTraits(
...
+  })
+  timeWindowDuration?: number;
+
+  @primitiveTrait({
+    name: "Time Unit",
+    description:
+      "The time unit for the `Time Interval`. Any units supported by `moment` module are valid, such as, `year`, `month`, `week`, `day`, `hour`, etc.",
+    type: "string"
+  })
+  timeUnit?: string;
```
na9da: Ideally I would prefer if it was possible to merge `timeUnit` and `timeWindowDuration` into a single trait. But otherwise, please rename this as `timeWindowUnit` to match the other trait name.

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
mwu2018: Can't do that because we need to support the case where only `currentTime` is defined.

lib/Traits/TraitsClasses/ArcGisMapServerCatalogItemTraits.ts
mwu2018: If merging number and unit into a single trait like `timeWindowDuration: "1d"`, we still need an extra trait indicating forward or backward window. My approach is to use positive or negative sign for that purpose.

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
```diff
@@ -452,9 +475,52 @@ export default class ArcGisMapServerCatalogItem extends UrlMixin(
...
+        currentTime: number,
+        windowDuration: number
+      ) {
+        const timeBound = Number(currentTime) + windowDuration;
+        if (windowDuration >= 0) {
+          return currentTime + "," + timeBound;
+        } else {
+          return "" + timeBound + "," + currentTime;
+        }
+      }
```
na9da: Could you move these functions outside the class but in the same file? Should make the method easier to read.

lib/Traits/TraitsClasses/ArcGisMapServerCatalogItemTraits.ts
na9da: Ok, let's make them all separate traits then, i.e `timeWindow`, `timeWindowUnits` and `timeWindowDirection`.

Reasoning:
- It aligns more with the original ArcGis parameters. It also requires less thought from the user when configuring, like how do I map forward/backward attribute to a `timeWindow` value etc.
- In the future, when using an editor UI to configure it is easier to enter a `number` and then pick the units and direction from drop-downs.

na9da (changes requested):

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
```diff
@@ -393,14 +401,32 @@ export default class ArcGisMapServerCatalogItem extends UrlMixin(
...
+      isForwardTimeWindow
+    } as TimeParams;
+    return timeParams;
+  }
 
-    const imageryProvider = this._createImageryProvider(dateAsUnix);
+  @computed
+  private get _currentImageryParts(): ImageryParts | undefined {
+    const timeParams = this.getTimeParams();
+    const imageryProvider = this._createImageryProvider(timeParams);
```
na9da: The `getTimeParams()` needs to be a `@computed`. This is because we are passing it as parameter to `_createImageryProvider` which itself is a computed (memoized) object. But here, `getTimeParams` returns a new object each time which will result in re-computing of `_createImageryProvider`. This is an expensive operation. You can avoid this by converting it to something like `@computed get timeParams() {...}`.

To see this in action, open the network tab for both the before and after links and try changing the opacity. For the after link, see that changing the opacity results in network fetches, whereas the network tab is silent when doing this for the before link.

mwu2018: Is it due to refactoring (functions being moved out at the class level)?

na9da (approved):
Great work @mwu2018 - looks good to go :rocket:

PR #6874, opened 2023-09-07
Mapserver Groups

sixlighthouses:
### What this PR does

Fixes [#536](https://github.com/TerriaJS/vic-digital-twin/issues/536)

Currently ArcGIS Mapservers group  can only handle nesting groups 1 level deep.  This PR adds a recursive check which will allow nesting to be as deep as defined by the Mapservice config.

### Test me

http://ci.terria.io/issue-536/#clean&https://gist.githubusercontent.com/sixlighthouses/a4fd27b75f41638efe902fa29ddf2825/raw/9167cbe29e558c4be7a269d37fd8298464278866/test.json

image

Open Data Catalog and Click through the multi level groups under Planning Overlays -- as per image above

prior to this change no layers are displayed under the second level groups i.e Environment and Landscape

### Checklist

### DIFF
--- a/lib/Models/Catalog/Esri/ArcGisMapServerCatalogGroup.ts
+++ b/lib/Models/Catalog/Esri/ArcGisMapServerCatalogGroup.ts
@@ -49,6 +49,7 @@ export class MapServerStratum extends LoadableStratum(
     ) as this;
   }
 
+  /** returns an array of the parent layers id's */
   findParentLayers(layerId: number): number[] {
     const parentLayerIds: number[] = [];
     const layer = this.layers.find((l) => l.id === layerId);
@@ -201,12 +202,13 @@ export class MapServerStratum extends LoadableStratum(
       return;
     }
     const id = this._catalogGroup.uniqueId;
-    //if parent layer is not -1 then this is sublayer so we define its ID like that
-    let layerId =
-      id +
-      "/" +
-      (layer.parentLayerId !== -1 ? layer.parentLayerId + "/" : "") +
-      layer.id;
+    let layerId = id + "/" + layer.id;
+
+    const parentLayers = this.findParentLayers(layer.id);
+
+    if (parentLayers.length > 0) {
+      layerId = id + "/" + parentLayers.reverse().join("/");
+    }
 
     let model: ArcGisMapServerCatalogItem | ArcGisMapServerCatalogGroup;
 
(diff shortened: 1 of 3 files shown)


lib/Models/Catalog/Esri/ArcGisMapServerCatalogGroup.ts
```diff
@@ -220,7 +233,14 @@ export class MapServerStratum extends LoadableStratum(
...
         layerId
       );
+
       if (existingModel === undefined) {
+        const parentLayers = this.findParentLayers(layer.id);
+
+        if (parentLayers.length > 0) {
+          layerId = id + "/" + parentLayers.reverse().join("/");
+        }
+
```
nf-s: I think it would be better to put your new `layerId` at the top of `createMemberFromLayer`. At the the moment your new correct `layerId` is only being used for new items - but the old incorrect id is being used for existing items, new groups or existing groups.

I tested another MapServer group with 4 nested layers, and it is missing the layers that are in nested sub groups 

- https://plan-gis.mapshare.vic.gov.au/arcgis/rest/services/Planning/VicPlan_PlanningSchemeHistory/MapServer

lib/Models/Catalog/Esri/ArcGisMapServerCatalogGroup.ts
```diff
@@ -49,6 +49,18 @@ export class MapServerStratum extends LoadableStratum(
     ) as this;
   }
 
+  findParentLayers(layerId: number): number[] {
```
nf-s, line 53: Can you please add a comment to highlight that it also returns the `layerId`, not just parentlayers?

nf-s (changes requested):
Thanks for working on this - it has been needed for a while!
I have two small comments

lib/Models/Catalog/Esri/ArcGisMapServerCatalogGroup.ts
sixlighthouses: Added comment to make that clearer

lib/Models/Catalog/Esri/ArcGisMapServerCatalogGroup.ts
sixlighthouses: Thanks Nick, moving the check for parents to the top of the function means it does correctly pick up the nested groups

nf-s (approved):
Looks good, thanks!

Issue #6538, opened 2022-09-09
Upgrade to React v18


AnaBelgun:
Upgrade to React V18. Details  tbc.

tephenavies:
@zoran995 do you think your React v17 PR would still be useful?

Resources for @staffordsmith83:
* New JSX transform Babel steps: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#manual-babel-setup
* React v17 release blog: https://reactjs.org/blog/2020/10/20/react-v17.html
* React v18 release blog: https://reactjs.org/blog/2022/03/29/react-v18.html
* React v18 how to upgrade guide: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html

tephenavies:
BTW @zoran995 the reason we're upgrading to React v18 is to use [Chakra v2](https://chakra-ui.com/getting-started). The intention is to replace our <Box>, <Text>, <StyledButton> etc. with equivalent Chakra components, and use other components available too.

zoran995:
Looking at changes for react 18 it would be good to pick up on react 17 PR and update it for react 18 since it has some fixes not directly related to upgrading, although, I would first push an upgrade to typescript v4, so we don't have to update react imports later and we also might run into an undocumented issue with react 18 and typescript 3.x (like the one with imports).

If going with chakra migration I would propose creating a wrapper at least for the most commonly used components, the best is to do that for all, i.e. TerriaButton that internally renders the chakra button. This provides us with a simple update of rendering logic without having to update each usage.

For picking up the react 17 PR basic steps would be, the same needs to be done on terriamap:
- merge the latest changes into PR
- use codemod to update imports
- update to react 18
- use codemod for updating the types 
- follow the migration guide and fix issues that arise
- pay attention to dependency incompatibility when migrating to the new version, if not possible to upgrade immediately (working like that) we should create an issue for each one to be updated later
- test that everything works correctly, especially documentation generation

If there is large amount of merge conflicts then it is easier to do the above steps on new PR and just cherry-pick the changes from react 17 PR  - **Actually we can also drop [this commit](https://github.com/TerriaJS/terriajs/pull/5549/commits/0849e48ac044be0e1c117912487559dbf132810c) and continue with just config from that pr**

For TerriaMap, consider removing `redbox-react` from dependencies, it is unmaintained, and use suspense instead, I don't recall ever seeing it in action

staffordsmith83:
[...]

PR #6902, opened 2023-09-22
React 18

tephenavies:
### Before merging
- [ ] Revert [ec350bb](https://github.com/TerriaJS/terriajs/pull/6902/commits/ec350bbf261355bb73a281d7de67859f53ccae3d)

### What this PR does

Part of #6538.

Upgrade React to 18 and Typescript to 5.

### Test me

Can be tested at http://ci.terria.io/react18-sd/

### Checklist


tephenavies:
Tests passing!

zoran995:
Can we also remove `import * as React from "react";` to stick to more modern syntax? It also might reduce bundle size and build speed. We can also do it as a follow-up ticket

lib/ReactViews/Map/Panels/InnerPanel.jsx
```diff
@@ -151,7 +159,7 @@ const InnerPanel = createReactClass({
               }
             }
             ${(p) =>
-              p.showDropdownAsModal &&
+              showDropdownAsModal &&
```
zoran995: This change is causing a white screen when opening map settings and other dropdowns.
Screenshot 2023-10-04 at 13 35 40

lib/ReactViews/Map/Panels/DropdownPanel.jsx
```diff
@@ -1,18 +1,25 @@
...
-import defined from "terriajs-cesium/Source/Core/defined";
+const buttonStyle = (p, panelIsOpen) => css`
+  ${panelIsOpen &&
+  `&:not(.foo) {
+    background: ${p.theme.colorPrimary};
+    svg {
+      fill: ${p.theme.textLight};
+    }
+  `};
+`;
```
nf-s: I think it might be better to make a `button` styled component here to make it more consistent with other components

lib/ReactViews/StandardUserInterface/SidePanelContainer.tsx
```diff
@@ -35,4 +35,9 @@ const SidePanelContainer = styled.div.attrs<PropsType>(({ viewState }) => ({
   margin-left: ${(p) => (p.show ? "0px" : `-${p.theme.workbenchWidth}px`)};
 `;
 
-export default withViewState(SidePanelContainer);
+// Styled components seems to make children prop not be visible through withViewState. Explicitly type the component to make it work.
+export default withViewState(
+  SidePanelContainer as React.ComponentType<
+    PropsType & React.HTMLAttributes<HTMLDivElement>
+  >
+);
```
nf-s: It might be better to add explicit compatibility for `StyledComponents` in `withViewState`

I had a play and this seems to satisfy types

```ts
export const withViewState =
  <P extends WithViewState>(
    Component:
      | React.ComponentType<P>
      | StyledComponent<
          keyof JSX.IntrinsicElements | React.ComponentType<any>,
          WithViewState
        >
  ): React.FC<Omit<P, "viewState">> =>
  (props) =>
    (
      <ViewStateContext.Consumer>
        {(viewState) => {
          if (!viewState)
            throw new TerriaError({ message: "ViewState is not defined!" });
          return <Component {...(props as P)} viewState={viewState} />;
        }}
      </ViewStateContext.Consumer>
    );
```

package.json
```diff
@@ -148,18 +149,18 @@
...
+    "react-dom": "^18.2.0",
     "react-ga4": "^2.1.0",
     "react-i18next": "^11.18.0",
     "react-responsive": "^5.0.0",
     "react-select": "^3.1.1",
     "react-swipeable": "^5.1.0",
     "react-transition-group": "^4.3.0",
-    "react-uid": "^2.3.0",
+    "react-uid": "^2.3.3",
     "react-virtual": "~2.3.2",
```
nf-s, line 163: Should we update any of these other react-based libraries?

package.json
```diff
@@ -178,7 +179,7 @@
     "ts-essentials": "^5.0.0",
     "ts-loader": "^5.3.3",
     "ts-node": "^5.0.1",
-    "typescript": "^4.9.5",
+    "typescript": "^5.2.2",
```
nf-s, line 181: 🎉

nf-s (changes requested):
Thanks for taking this on @steve9164 ! 

A few things I noticed that aren't working correctly - 

Chart expand/download buttons

- Open "All generation types"
- Click on point
- Click on "expand" or "download" buttons in feature info

A few other issues caused by what @zoran995 commented on - `showDropdownAsModal is not defined`

I agree with @zoran995 on reducing use of `import * as React from "react"` - it would be great to at least remove all of these from `.tsx` files

I think we also need to update to the new React Root API - https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apis

> Until you switch to the new API, your app will behave as if it's running React 17

I believe there is only two usages of ReactDOM.render in https://github.com/TerriaJS/TerriaMap/blob/main/lib/Views/render.jsx#L9

package.json
zoran995: IMO there are some dependencies that probably should be upgraded as they list older versions of react as peer dependencies (quick check can be done via `npx check-peer-dependencies --yarn`)
- `@visx/*` should be upgraded to _at least_ v2.10.0, it is the first version that lists react 18 as dependency
- `html-to-react` version used has peer dependency on react 16/17, newer versions are on react 18
- `react-anything-sortable`
- `react-datepicker`
- `react-responsive`
- `react-select`
- `react-swipeable`
- `react-virtual` (I think the only issue with the upgrade is the wrong usage of `createRef` in functional component and useRef should be used instead https://github.com/TerriaJS/terriajs/pull/5549/commits/b6390c5cf4249af1c1e8e5ce6efad00d4dc6a495)
- `react-shallow-testutils`

lib/ReactViews/StandardUserInterface/SidePanelContainer.tsx
tephenavies: I removed the explicit typing that I added on SidePanelContainer and everything still seems to work, so 🤷

lib/ReactViews/Map/Panels/InnerPanel.jsx
tephenavies: Good catch. I have re-written these as proper styled components with regression tests (simple render specs)

lib/ReactViews/Map/Panels/DropdownPanel.jsx
tephenavies: You're right. I don't know why I was avoiding writing as a component

package.json
tephenavies: It's unlikely that the libraries are incompatible with React 18, and more likely that library authors are using version strings such as
```
"react": "^16.0 || ^17.0"
```

when they should use the standard:
```
"react": "^0.13.0 || ^0.14.0 || >=15"
```

but I'll check the usage of those libraries

tephenavies: A library that definitely needs updating is [mobx-react](https://github.com/mobxjs/mobx/tree/main/packages/mobx-react#readme)

zoran995: @steve9164 there will be issues with libraries listing react in peerDependencies as 
```
"react": "^16.0 || ^17.0"
```
if we are on react 18 after upgrading to the new version of the node which comes with a new npm that automatically installs and checks peerDependencies. You can see the error occurring in https://github.com/TerriaJS/terriajs/pull/6986

tephenavies:
Breaking into smaller PRs. Starting with #7050

PR #6903, opened 2023-09-22
Add `culture` to `BingMapsCatalogItem`

ykiu:
### What this PR does

The Bing Maps API has a parameter called `culture` (https://learn.microsoft.com/en-us/bingmaps/rest-services/common-parameters-and-types/supported-culture-codes), which can be used for localizing labels in the Bing base maps. Cesium's `BingMapsImageryProvider` accepts the parameter, but our BingMapsCatalogItem previously did not.

This PR adds the `culture` parameter to BingMapsCatalogItem, which will be passed to `BingMapsImageryProvider`.
This PR also adds some basic tests for the catalog item.

### Test me

> How should reviewers test this? (Hint: If you want to provide a test catalog item, [create a Gist](https://gist.github.com/) of its catalog JSON, add its url and your branch name to this url: `http://ci.terria.io/<branch name>/#clean&<raw url of your gist>` , and then paste that in the body of this PR)

http://ci.terria.io/bing-maps-culture/#clean&https://gist.githubusercontent.com/ykiu/4ab8f504823351c5b3bbd5999f0b620f/raw/3e509c6615ab4f636fb9d51c6ec16bdc6b2e3f69/bing.json

- `BingMaps Road (default)` displays map labels in Bing's default language (English). 
  - image
- `BingMaps Road (ja)` displays map labels in Japanese.
  - image

### Checklist

  - not applicable, since docs are auto-generated for catalog items.


tephenavies (approved):
Thanks for the contribution! This is great.

nf-s (changes requested):
Hey @steve9164 

There seems to be a test error

https://github.com/TerriaJS/terriajs/actions/runs/6320724251/job/17502169425#step:7:3810

image

It passes when I run it locally though 🤔 
Do you have any ideas?

ykiu:
Hi @steve9164 and @nf-s, thanks for reviewing/testing this!

I was able to reproduce the `Script error. thrown` error locally, although it only occurred occationally.

I looked up the error briefly and came up with an hypothetical cause. You have far more experience with the codebase of terriajs than I do, so maybe you are already aware of the things I've found out, but I'd like to share my thoughts anyway just in case it helps.

**tl;dr**: the error seems to be due to the flakiness of some tests which are unrelated to the failing test.

According to this blog post https://sentry.io/answers/script-error/ `Script error. thrown` occurs when an error from a JavaScript file served from an external origin is caught by the window.onerror event handler.

In my local testing (`yarn gulp release && yarn start` then accessing http://localhost:3002/SpecRunner.html), `Script error. thrown` occurred kind of randomly with different tests failing each time I reload the browser.

image

So here's my hypothesis: there's an async process started by a script from an external origin and the process errors from time to time. Because the process is async the timing of the error is non-deterministic, hence the random test failures.

I dumped the network requests made during the test runs and searched for scripts from external origins. A possible culprit might be the following JSONP response from bing. It is certainly "a script from an external origin" but I'm not yet sure if it's doing any async stuff or if it is in fact causing an error.

`http://dev.virtualearth.net/REST/v1/Imagery/Metadata/Aerial?jsonp=loadJsonp839604&incl=ImageryProviders&key=XXXXX&uriScheme=http`

If my hypothesis is right the CI error is false positive, and re-running the CI would be the short term solution for this PR. The long term solution would be to investigate exact place where the error is occuring (which may be the JSONP mentioned above or something else) and fixing it.

Thank you!

zoran995:
I will drop my 2 cents, the solution ~~should~~(is) **never** ~~be~~ to just rerun tests as it suggests that test definition is fragile and wrong to some extent. Tests have to provide the same result every time they are run, be independent of each other and not be affected by the environment they are executed on. If anything above is not true it might cause someone else to spend time investigating why the test is failing when changes don't affect the tested code. 

`BingMapsImageryProvider` is using `Resource.fetchJsonp` to load the metadata so putting the code below in beforeEach should resolve the issue
```
spyOn(Resource.prototype, "fetchJsonp").and.returnValue(
  Promise.resolve({})
);
```
I am not sure if there is a better proper fix to this as this is something handled by Cesium internally.

p.s. I have just created #6961 where I wrote a little about other issues in a unit testing environment

ykiu:
Thank you very much for your suggestion, @zoran995. I'll take a look hopefully this weekend.

ykiu:
Oh yeah as @zoran995 pointed out the JSONP request is actually made by BingMapsCatalogItem, so I needed to mock it. I updated the tests as suggested by @zoran995 and they are working locally on my end. Could someone please test them out?

nf-s (approved):
Thanks @ykiu 

Another great improvement, thanks a lot!

ykiu:
Thanks @nf-s @steve9164 and @zoran995  for the great feedback and the improvements!

PR #6931, opened 2023-10-05
RFC: 6923 deprecate use of jsx control statements

### LINKED ISSUE #6923: Deprecate use of jsx-control-statements
Currently we are using `jsx-control-statements` which depends on babel. Replacing the usages with regular jsx expressions will help transition to [esbuild](https://github.com/TerriaJS/terriajs/issues/6624)

ljowen:
### What this PR does
RFC - DO NOT MERGE YET!

Fixes #6923

Replace usages of jsx-control-statements, this will allows us to remove babel as a dependency
- [x] If
- [x] For
- [ ] When 

### Test me

http://ci.terria.io/6923-deprecate-use-of-jsx-control-statements

Check 

### Checklist

### DIFF
--- a/lib/ReactViews/Preview/DataPreviewSections.jsx
+++ b/lib/ReactViews/Preview/DataPreviewSections.jsx
@@ -100,16 +100,13 @@ class DataPreviewSections extends React.Component {
               }
               bodyTextProps={{ medium: true }}
             >
-              <Choose>
-                <When condition={item.content?.length > 0}>
-                  {renderSection(item)}
-                </When>
-                <When condition={item.contentAsObject !== undefined}>
-                  <Box paddedVertically={3} fullWidth>
-                    <MetadataTable metadataItem={item.contentAsObject} />
-                  </Box>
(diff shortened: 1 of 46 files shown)


lib/ReactViews/Preview/DataPreviewSections.jsx
ljowen, line 1: AFAICT this isn't actually used (and eslint picks it up as such) since there's no `className={item}`

nf-s (commented):
This is great! 
Thanks for taking this on @ljowen 

I am happy with your approach for replacing `If` and `For`

nf-s:
@ljowen to fix the CI you need to rename your branch as it is too long - I'm not sure of the actual character limit.
It is something due to how Kubernetes `ConfigMap`s get created for each branch

> I think there are 2 configmaps (something like terriajs-branch-terriamap-config and terriajs-branch-terriamap-serverconfig) and they’re both getting truncated to the same name [because of the length of the branch name]

PR #6937, opened 2023-10-12
Deprecate jsx control statements

### LINKED ISSUE #6923: Deprecate use of jsx-control-statements

ljowen:
### What this PR does

Fixes #6923

Replace usages of jsx-control-statements, this will allows us to remove babel as a dependency
- [x] If
- [x] For
- [x] When 

### Test me

http://ci.terria.io/deprecate-jsx-control-statements

Check 
Review code changes, navigate through site

### Checklist

- [x] There are unit tests to verify my changes are correct or unit tests aren't applicable. (Unit test added for more complex case, most changes simple logic)


ljowen:
Noticed the following on CI `Warning: Failed prop type: Invalid prop `children` supplied to `CatalogGroup`.` This is from changes to DataCatalogGroup 

- [x] Fix invalid prop warning

architecture/0001-babel-and-ts-loader.md
```diff
@@ -24,7 +24,7 @@ support for it covers everything we need to use it in terriajs, including
 build times.
 
 - Using _only_ ts-loader is out of the question, as we rely on tools inside the
-  babel ecosystem including jsx-control-statements & styled-components to name a
+  babel ecosystem including styled-components to name a
```
tephenavies: It's best not to remove information from old Architecture Decision Records. In many cases it's best to create a new ADR with updated details, but here we could put a strikethrough of "jsx-control-statements" and a short explanation in parentheses. 

```suggestion
  babel ecosystem including ~jsx-control-statements~ (removed 2023-10-16) & styled-components to name a
```

Removing jsx-control-statements is also a candidate for its own ADR. If you do write a new ADR, link it from the edit to this one.

lib/ReactViews/DataCatalog/CatalogGroup.jsx
```diff
@@ -113,35 +113,36 @@ function CatalogGroup(props) {
...
-                <Loader />
-              </li>
-            </When>
-            <When condition={props.children.length === 0 && props.emptyMessage}>
+          {props.loading && (
+            <li key="loader">
+              <Loader />
+            </li>
+          )}
+          {props.loading === false &&
```
tephenavies: I think it's best to use `!props.loading` here because it's the true logical opposite of `loading`, and also because `loading` could be undefined. Looks like usage of `<CatalogGroup>` always specifies `loading`, but it's not guaranteed by defaultProps or PropTypes.

tephenavies (approved):
Thanks for making those changes!

PR #6944, opened 2023-10-22
Upgrade to Cesium 1.113.0

kring:
The only problem I know about is that WMTS feature picking most likely doesn't work because it needs a CesiumJS change. I don't see a WMTS dataset to test with in TerriaMap, though. There may very well be other problems.

I've already published `terriajs-cesium` 5.0.0 (mirroring `@cesium/engine` 5.0.0) to npm, as well as the new `terriajs-cesium-widgets` 4.1.0 (mirroring `@cesium/widgets` 4.1.0). I've tagged them`kring-test` instead of `latest` for now, though. To publish new versions to npm in the future, do:

```
npm run prepare
npm run release
npm publish -w terriajs-cesium
npm publish -w terriajs-cesium-widgets
```


kring:
Try it out here:
http://ci.terria.io/simple-cesium-upgrade/

na9da:
Thanks @kring for the massive effort. 

I have merged in the latest `terriajs#main` and updated `terriajs-cesium` and `terriajs-cesium-widget` packages to match the [latest upstream](https://github.com/CesiumGS/cesium/releases/tag/1.113).

Have also upgraded our internal plugin libraries to work with the new packages.

I believe once ready to merge we should make `terriajs-new` branch of [terriajs/cesium](https://github.com/terriajs/cesium) repository as the new default branch (or rename it as `main`). Making a PR to the current default branch doesn't make sense.

Ideally we should test deploy to one of our instances before a final merge.

na9da (approved):

PR #6957, opened 2023-11-01
Use cesium ion geocoder

### LINKED ISSUE #5107: Geocoder - update
We're currently using Bing geocoder along GNAF. 
Options: 
- keep Bing = changing arrangements for platforms we manage using Bing enterprise
- Pelias
- Here
- Esri 
- other?

ljowen:
### What this PR does

Fixes #5107

Implement CesiumIonGeocoder which calls: https://api.cesium.com/v1/geocode/search

### Test me

Visit http://ci.terria.io/use-cesium-ion-geocoder search for an address

### Checklist

### Issues
- Each result only has a `label` property which doesn't include the state, country, or any other metadata
```
            "properties": {
                "label": "Annerley, Australia"
            },
            "bbox": [
                153.0238800048828,
                -27.525028228759766,
                153.04367065429688,
                -27.50009536743164
            ]
```
This makes it hard to differentiate the locations with the same name, for example searching "Annerley" returns two results with "Annerley, Australia". The same problem can be observed in Cesium itself: https://sandcastle.cesium.com/standalone.html#c=bY9Pa8MwDMW/ivGpgyGx6+qGQq+FHQY7+eI6Wmuq2MFSUtJPvyS9jK26vaf3059YsqgZE92omp3JdDMHkjR08LV6G2/jqg8la0iZqrcvW5/tq3WiE1Pjs1lrn7q+VDVD5Q0AKnU9ByXB0xCvpBBFFnCJOvyNujaNJrW7J5tM5CAyd74H5s90J28bh3P+H8oltCmfP0aqHKYldnlrjg8TABzO8jmppfAp1D+T5/8uqr28I0rIbQyiTPA4EGLp8ByYqU74Aw

### DIFF
--- a/buildprocess/ci-deploy.sh
+++ b/buildprocess/ci-deploy.sh
@@ -23,7 +23,7 @@ npm install -g yarn@^1.19.0
 
 # Clone and build TerriaMap, using this version of TerriaJS
 TERRIAJS_COMMIT_HASH=$(git rev-parse HEAD)
-git clone -b main https://github.com/TerriaJS/TerriaMap.git
+git clone -b use-cesium-ion-geocoder https://github.com/TerriaJS/TerriaMap.git
 cd TerriaMap
 TERRIAMAP_COMMIT_HASH=$(git rev-parse HEAD)
 sed -i -e 's@"terriajs": ".*"@"terriajs": "'$GITHUB_REPOSITORY'#'${GITHUB_BRANCH}'"@g' package.json
--- a/lib/Models/SearchProviders/CesiumIonSearchProvider.ts
+++ b/lib/Models/SearchProviders/CesiumIonSearchProvider.ts
@@ -7,11 +7,7 @@ import Scene from "terriajs-cesium/Source/Scene/Scene";
 import SearchProviderResults from "./SearchProviderResults";
 import SearchResult from "./SearchResult";
 import Rectangle from "terriajs-cesium/Source/Core/Rectangle";
-import Cartesian3 from "terriajs-cesium/Source/Core/Cartesian3";
-import Cartographic from "terriajs-cesium/Source/Core/Cartographic";
-import GeocoderService from "terriajs-cesium/Source/Core/GeocoderService";
 import loadJson from "../../Core/loadJson";
-import bbox from "@turf/bbox";
 interface CesiumIonSearchProviderOptions {
   terria: Terria;
   url?: string;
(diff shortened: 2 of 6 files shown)


nf-s:
I see what you mean, the API is terrible!
And there’s basically no documentation!
Are there any other parameters we can put in requests?
It looks like cesium is using the PeliasGeocoderService under the hood https://github.com/CesiumGS/cesium/blob/2bd8dadacc8f429ff0e286d3cd688d197449e96e/packages/engine/Source/Core/IonGeocoderService.js#L51

buildprocess/ci-deploy.sh
```diff
@@ -23,7 +23,7 @@ npm install -g yarn@^1.19.0
 
 # Clone and build TerriaMap, using this version of TerriaJS
 TERRIAJS_COMMIT_HASH=$(git rev-parse HEAD)
-git clone -b main https://github.com/TerriaJS/TerriaMap.git
+git clone -b use-cesium-ion-geocoder https://github.com/TerriaJS/TerriaMap.git
```
ljowen, line 26: - [ ] revert before merge

nf-s (changes requested):
One small comment, other than that is looks great - thanks!

lib/Models/SearchProviders/CesiumIonSearchProvider.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (!response.features) {
+      searchResults.message = i18next.t("viewModels.searchNoLocations");
+      return;
+    }
+
+    if (response.features.length === 0) {
+      searchResults.message = i18next.t("viewModels.searchNoLocations");
+    }
+
+    searchResults.results = response.features.map<SearchResult>((feature) => {
```
nf-s: Can you please wrap the modifications of `searchResults.message` and `searchResults.results` in a mobX `runInAction`?

This will get rid of the warnings in the console

```
[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: SearchProviderResults@1082.results
```

ljowen: Thanks for picking this up!

nf-s (approved):
Looks good, thanks!

irbian:
It's been a while, has the situation changed? cesium still returns only bbox and label. We mainly miss that the center point is not included so in places like Spain you get strange things:

image


Compare with a selfhosted pelias geocoder
image

PR #6972, opened 2023-11-14
Add `ApiColumnTraits` (`responseDataPath`) to `ApiTableCatalogItem`

nf-s:
### Add `ApiColumnTraits` (`responseDataPath`) to `ApiTableCatalogItem`

Adds new `apiColumns` traits to `ApiTableCatalogItem`. It only adds one extra property for `responseDataPath` - this allows us to support APIs with nested properties.

I had to create a new `apiColumns` trait - as you can't add new properties to existing traits (eg `columns: TableColumnTraits[]`

### Test me

This example has nested lat/lon

```json
{
  "catalog": [
    {
      "name": "EPA API Integration test",
      "description": "something",
      "type": "api-table",
      "columns": [
        { "name": "siteID" },
        { "name": "lat" },
        { "name": "lon" }
      ],
      "apiColumns": [
        { "name": "lat", "responseDataPath": "geometry.coordinates.0" },
        { "name": "lon", "responseDataPath": "geometry.coordinates.1" }
      ],
      "apis": [
        {
          "url": "https://www.epa.vic.gov.au/api/envmonitoring/sites?environmentalSegment=air",
          "responseDataPath": "Model.records",
          "kind": "PER_ROW"
        }
      ]
    }
  ]
}
```

[Test link](http://ci.terria.io/api-table-columntraits/#clean&start=%7B%22initSources%22%3A%20%5B%7B%22workbench%22%3A%20%5B%22test%22%5D%2C%22catalog%22%3A%20%5B%7B%22id%22%3A%20%22test%22%2C%22name%22%3A%20%22EPA%20API%20Integration%20test%22%2C%22description%22%3A%20%22something%22%2C%22type%22%3A%20%22api-table%22%2C%22columns%22%3A%20%5B%7B%20%22name%22%3A%20%22siteID%22%20%7D%2C%7B%20%22name%22%3A%20%22lat%22%20%7D%2C%7B%20%22name%22%3A%20%22lon%22%20%7D%5D%2C%22apiColumns%22%3A%20%5B%7B%20%22name%22%3A%20%22lat%22%2C%20%22responseDataPath%22%3A%20%22geometry.coordinates.0%22%20%7D%2C%7B%20%22name%22%3A%20%22lon%22%2C%20%22responseDataPath%22%3A%20%22geometry.coordinates.1%22%20%7D%5D%2C%22apis%22%3A%20%5B%7B%22url%22%3A%20%22https%3A%2F%2Fwww.epa.vic.gov.au%2Fapi%2Fenvmonitoring%2Fsites%3FenvironmentalSegment%3Dair%22%2C%22responseDataPath%22%3A%20%22Model.records%22%2C%22kind%22%3A%20%22PER_ROW%22%7D%5D%7D%5D%7D%5D%7D)

### Checklist

- [ ] ~I've updated relevant documentation in `doc/`.~


lib/Traits/TraitsClasses/ApiTableCatalogItemTraits.ts
```diff
@@ -29,12 +30,28 @@ export class ApiTableRequestTraits extends mixTraits(ApiRequestTraits) {
...
+  })
+  responseDataPath?: string;
+}
+
 export default class ApiTableCatalogItemTraits extends mixTraits(
   TableTraits,
   CatalogMemberTraits,
   LegendOwnerTraits,
-  AutoRefreshingTraits,
-  ApiRequestTraits
```
nf-s, line 37: Note I have removed `ApiRequestTraits` here because `ApiTableCatalogItem` only supported `queryParameters` and `updateQueryParameters` - I have re-added this manually

na9da (approved):
@nf-s - looks good to me.

I think it would have been nice if we could somehow override the table `columns` API. The traits system allows redefining the `columns` trait inside `ApiTableCatalogItemTraits` but the inclusion of `TableMixin` type masks that (and forces the type back to table columns).

I think your solution is the next best.

nf-s:
@na9da I totally agree. Do you have any ideas about how we could achieve this? 
We could add a function called `excludeTraits<T extends ModelTraits>(trait:T, keys: (keyof T)[])` which could be used in `mixTraits`

For example

```ts
export default class ApiTableCatalogItemTraits extends mixTraits(
  excludeTraits(TableTraits, ["columns"]),
  ...
) {
  @objectArrayTrait({
    name: "Columns",
    type: ApiTableColumnTraits,
    ...
  })
  columns: ApiTableColumnTraits[] = [];
```

I believe we have a ticket for this!

- https://github.com/TerriaJS/terriajs/issues/5207

PR #6985, opened 2023-11-27
Add Typescript eslint

### LINKED ISSUE #3303: Model architecture: Enable eslint for .ts files

ljowen:
### What this PR does

Fixes #3303
Special thanks to @zoran995 for their earlier PR!
- Enable eslint for .ts and .tsx files 
- Run `yarn gulp lint --fix`
- Manually fix remaining problems. Where possible I've been conservative about changing typings or changes that could have unseen side effects.

More rules can be enable progressively

### Test me

Given the blast radius exploratory testing is likely best approach along with careful review: http://ci.terria.io/typescript-eslint

### Checklist

- ~[] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~


.eslintrc
```diff
@@ -25,221 +31,145 @@
...
-    /* ECMAScript 6 */
-    "constructor-super": 1,
-    "generator-star-spacing": 0, // see Babel section
-    "no-this-before-super": 1,
-    "no-var": 0,
-    "object-shorthand": 0, // see Babel section
-    "prefer-const": 1,
-    "no-useless-escape": 0
-  }
+        // @TODO: enable these rules
```
zoran995: May I suggest putting this as `warn` instead of turning them `off`, and ditching the rule of max-warnings 0 when running linter. Having them as `warn` will allow developers to fix them incrementally during development, if the rules are turned off then we can't get any meaningful information that those even exist during development. 
I agree that treating warnings as errors in linter might be beneficial so more rules can be enforced, but for me, it is pretty much the same as having it error by default. 

What do you think about this? @ljowen @nf-s

.eslintrc
```diff
@@ -1,6 +1,11 @@
...
+  "root": true,
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended",
+    "plugin:react-hooks/recommended"
+  ],
+  "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "requireConfigFile": false,
     "ecmaVersion": 2018,
```
zoran995: Should this be in sync with the es version configured in tsconfig?

.eslintrc
```diff
@@ -1,6 +1,11 @@
 {
-  "extends": ["eslint:recommended", "plugin:react/recommended"],
-  "parser": "@babel/eslint-parser",
+  "root": true,
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended",
+    "plugin:react-hooks/recommended"
```
zoran995: we should also add `plugin:@typescript-eslint/eslint-recommended` here as it effectively disables the conflicting eslint rules and replaces them with typescript aware implementations

zoran995: also, do you maybe have warnings and error numbers with `plugin:@typescript-eslint/recommended` extended just to have a rough idea of what is needed to enforce a more standard configuration

ljowen: without the .ts,.tsx overrides we get:
✖ 353 problems (40 errors, 313 warnings)

.eslintrc
ljowen: I think that's a good idea, I'd like to resolve these as a follow up PR (and change these back to the default "error"). So I've set --warn-limit=341 in gulpfile. To me it doesn't make sense to have warnings=0, since that just means warnings become errors.

doc/js/rewrite-links.js
```diff
@@ -11,5 +11,7 @@ document.querySelectorAll("a").forEach(function (a) {
       a.target = "_blank";
       a.rel = "noreferrer noopener";
     }
-  } catch {}
+  } catch {
+    /* eslint-disable-line no-empty */
```
ljowen, line 15: We could also use this option:  https://eslint.org/docs/latest/rules/no-empty#allowemptycatch

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -1751,7 +1751,7 @@ function isPolygonOnTerrain(polygon: PolygonGraphics, now: JulianDate) {
   return isClamped || (!hasPerPositionHeight && !hasPolygonHeight);
 }
 
-export function getColor(color: String | string | Color): Color {
+export function getColor(color: string | string | Color): Color {
```
ljowen:
```suggestion
export function getColor(color: string | Color): Color {
```

lib/Models/Catalog/CatalogItems/SenapsLocationsCatalogItem.ts
```diff
@@ -88,14 +88,27 @@ export class SenapsLocationsStratum extends LoadableStratum(
 
   static async load(senapsLocationsCatalogItem: SenapsLocationsCatalogItem) {
     const locationsUrl = senapsLocationsCatalogItem._constructLocationsUrl();
+
+    function addStreamIds(f: SenapsFeature, index: number, streamData: any) {
```
ljowen, line 92: Added streamData parameter instead of relying on closure

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
ljowen, line 1: Conflict with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol

lib/Models/Definition/addModelStrataView.ts
```diff
@@ -19,6 +19,8 @@ export default function addModelStrataView<
 ): ModelPropertiesFromTraits<InstanceType<T>>;
 export default function addModelStrataView<
   T extends TraitsConstructor<ModelTraits>
+  /* TODO: Fix this overload type */
+  /* eslint-disable-next-line @typescript-eslint/ban-types */
 >(model: Function, Traits: T): ModelPropertiesFromTraits<InstanceType<T>>;
```
ljowen, line 24: I couldn't easily find an appropriate type for `model` that would satisfy tsc since it seems to get treated as both a function and a class constructor

lib/ReactViews/Context/ViewStateContext.tsx
```diff
@@ -20,12 +20,11 @@ export const useViewState = () => {
...
-export const withViewState =
-  <P extends WithViewState>(
-    Component: React.ComponentType<P>
-  ): React.FC<Omit<P, "viewState">> =>
-  (props) =>
-    (
+export const withViewState = <P extends WithViewState>(
+  Component: React.ComponentType<P>
+): React.FC<Omit<P, "viewState">> =>
+  function withViewState(props) {
```
ljowen, line 26: this feels a bit hacky but means we get a `displayName`

lib/ReactViews/Map/TerriaViewerWrapper/Splitter/dragHook.ts
```diff
@@ -23,7 +23,9 @@ try {
 
   window.addEventListener("test", callback, options);
   window.removeEventListener("test", callback, options);
-} catch (err) {}
+} catch (err) {
+  /* ignore error */
```
ljowen:
```suggestion
  /* eslint-disable-line no-empty */
```

.eslintrc
```diff
@@ -25,221 +32,152 @@
     }
   },
   "rules": {
-    "react/jsx-no-undef": 0,
+    "react-hooks/exhaustive-deps": "error",
+    "react/jsx-boolean-value": ["error", "never", { "always": [] }],
+    "react/jsx-closing-bracket-location": ["error", "line-aligned"],
+    "react/jsx-closing-tag-location": "error",
+    "react/jsx-curly-spacing": ["error", "never", { "allowMultiline": true }],
```
zoran995: It would be better to install `eslint-config-prettier` and let it handle this, I am a bit worried that those rules might cause conflicts with prettier formatting

.eslintrc
```diff
@@ -25,221 +32,152 @@
...
+      {
+        "allowAllCaps": true,
+        "ignore": []
+      }
+    ],
+    "react/no-danger": "warn",
+    "react/no-did-update-set-state": "error",
+    "react/no-will-update-set-state": "error",
+    "react/self-closing-comp": "error",
+    "react/jsx-no-undef": ["error", { "allowGlobals": true }],
```
zoran995, line 55: I would also consider `react/jsx-no-leaked-render` and `react/no-object-type-as-default-prop` as they can early catch some pretty nasty stuff. In the past, there were cases when 0 was rendered due to some programmatic scroll or some weird behaviour

.eslintrc
```diff
@@ -25,221 +32,152 @@
...
-    "default-case": 0,
-    "dot-notation": [
-      0,
-      {
-        "allowKeywords": true,
-        "allowPattern": ""
-      }
-    ],
-    "dot-location": [1, "property"],
-    "eqeqeq": 1,
```
zoran995, line 75: eqeqeq won't actually be activated by default :(

.eslintrc
```diff
@@ -25,221 +32,152 @@
...
+    "prefer-const": "error"
+  },
+  "overrides": [
+    {
+      "files": ["**/*.ts", "**/*.tsx"],
+      "extends": [
+        "eslint:recommended",
+        "plugin:react/recommended",
+        "plugin:react-hooks/recommended",
+        "plugin:@typescript-eslint/recommended"
```
zoran995: here you also need a `plugin:@typescript-eslint/eslint-recommended` so js eslint rules are converted to ts version

zoran995: actually, my bad `plugin:@typescript-eslint/recommended` will activate it automatically but IMO doesn't hurt to add it explicitly, your call

.eslintrc
```diff
@@ -25,221 +32,152 @@
...
-    "spaced-comment": [0, "always"],
-    "wrap-regex": 0,
+    "no-unneeded-ternary": 1 /* ECMAScript 6 */,
+    "prefer-const": "error"
+  },
+  "overrides": [
+    {
+      "files": ["**/*.ts", "**/*.tsx"],
+      "extends": [
+        "eslint:recommended",
```
zoran995: To check if this was intentional, having all those extends here will force you to specify every overriding config option in 2 places, above for .js and here for typescript files, i.e. `eslint:recommended` sets `rule-1` to error, you disable it or change it and that here again specify `eslint:recommended` forcing you to specify config in two places

.eslintrc
tephenavies: Prettier recommends using `eslint-config-prettier`, so we should include this

ljowen: Thanks @zoran995  I've added that and removed any rules reported by the `eslint-config-prettier` cli tool: https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#cli-helper-tool

.eslintrc
ljowen: Thanks for picking this up! I've added it back in and fixed in .ts,.tsx files

.eslintrc
ljowen: I've simplified the config so the `overrides` is just changing to `warn` errors still be to fixed in ts,tsx files

.eslintrc
ljowen: I'd prefer to be explicit, will add it

.eslintrc
ljowen: These require an update to `eslint-plugin-react` and introduce ~100 new errors. I'll create an issue to enable and fix these rules in the interest of keeping the scope of this PR down

tephenavies (approved):

PR #6993, opened 2023-12-01
update readme node versions

nf-s:
We now no longer support NodeJS 14 - and we have added support for 18 and 20.

I have updated docs to reflect this

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,8 @@ Sites we're aware of that are using TerriaJS. These are not endorsements or test
 
 ### Technical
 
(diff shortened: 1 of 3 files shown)


ljowen:
Nice one! Since node 16 is EoL shall we update .nvmrc to 18 or dare I say 20?

tephenavies (approved):

nf-s:
We will keep Node 16 as default version until we upgrade webpack to remove need for `--openssl-legacy-provider` in `NODE_OPTIONS`

PR #6997, opened 2023-12-03
Upgrade typescript to 5.2

tephenavies:
### What this PR does

Offshoot of #6902.

* Upgrades TypeScript to ~5.3~ 5.2
* Improve tsconfig files
* Switch to new JSX transform (added in React 17 and backported to React 16.14)

### Test me

http://ci.terria.io/ts5/

### Checklist


tephenavies:
☹️ TS 5.2 -> 5.3 upgrade breaks our cursed super access: https://github.com/microsoft/TypeScript/issues/35314

tsconfig-node.json
```diff
@@ -1,25 +1,9 @@
 // This tsconfig is used to build the model layer only for use in a node.js app.
 // Build with ./node_modules/.bin/tsc -p tsconfig-models.json
```
ljowen:
```suggestion
// Build with ./node_modules/.bin/tsc -p tsconfig-node.json
```

tephenavies: I've never properly read that comment. Good fix.

tsconfig.json
```diff
@@ -7,13 +7,23 @@
...
+    "typeRoots": [
+      "./lib/ThirdParty/",
+      "./node_modules/",
+      "./node_modules/@types/",
+      "../node_modules/",
+      "../node_modules/@types/",
+      "../../node_modules/",
+      "../../node_modules/@types/",
+      "../../../node_modules/",
+      "../../../node_modules/@types/"
```
ljowen, line 25: Curious why need need 3 levels of parent here? Also I believe `./node_modules` makes `./node_modules/@types/` redundant.

ljowen (approved):

tsconfig.json
tephenavies: I included 3 levels because I have an odd workflow using [git worktree](https://git-scm.com/docs/git-worktree) and yarn packages. I end up with many terriajs folders inside 1 TerriaMap at `packages/tjs/<branch-name>` (for WIP, reviews, testing), and then control which one to link, re-run `yarn install` and build. I could use `packages/<branch-name>` but I like to have all my terriajs folders in their own namespace of sorts. I'm trying to think of a disadvantage of having this config compared to running TS4 which did it's own walking

I tried with only `node_modules/` entries and only `node_modules/@types/` entries and neither is sufficient (types for `react` require `node_modules/@types/` and types for `terriajs-cesium` require `node_modules/`).

The reason this change to typeRoots is needed at all is because of a change in TS 5.1 https://devblogs.microsoft.com/typescript/announcing-typescript-5-1/#explicit-typeroots-disables-upward-walks-for-node_modules-types

ljowen: Thanks, I didn't realise scanning upwards was a behaviour pre 5.1.

nf-s:
Ready to merge in now

PR #7016, opened 2023-12-19
Fix: draw marker on first load

### LINKED ISSUE #6854: Location search marker not shown when opening from share link.
Open this [share link](https://nationalmap.gov.au/#share=s-v1ODZS9Vp9mYhqJfwWl0TloAq5N) I created after searching for Melbourne airport. Observe that the location marker is not shown when opening the share link although it was correctly shown after searching.

ljowen:
### What this PR does

Fixes #6854 

When a location marker is stored in a share or start url, react to it's addition to the catalog once and draw the marker

### Test me

Visit http://ci.terria.io/bug-search-marker/#share=s-kzAOTeO49a1mrKMNum3v7zrVzO7 the map marker should be shown

### Checklist

### DIFF
--- a/lib/Models/LocationMarkerUtils.ts
+++ b/lib/Models/LocationMarkerUtils.ts
@@ -110,3 +110,7 @@ export function getMarkerLocation(terria: Terria): LatLonHeight | undefined {
   }
   return undefined;
 }
+
+export function getMarkerCatalogItem(terria: Terria) {
+  return terria.getModelById(CzmlCatalogItem, MARKER_UNIQUE_ID);
+}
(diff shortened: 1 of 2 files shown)


na9da (approved):
Looks good @ljowen!

I think the root cause here is that we don't serialize the `overlays` as part of the share link like we do `workbench`. I'm not sure if that was deliberate design decision - maybe overlays are meant to be ephemeral and anything permanent must be added to the workbench(?). If we decide to serialize `overlays` in the future, then we could probably remove the special handling for the marker item.

Issue #6895, opened 2023-09-18
WPS: default value for time widgets?


sppigot:
It would be nice if WPS datetime inputs would use the current value of the date-time slider e.g. Here is an datetime input to a WPS from a DescribeProcess WPS call:

```
<Input minOccurs="1" maxOccurs="1">
                <ows:Identifier>ReportTime</ows:Identifier>
                <ows:Title>Surf report time</ows:Title>
                <ows:Abstract>The time at which to produce the surf report</ows:Abstract>
                <ComplexData maximumMegabytes="1024">
                    <Default>
                        <Format>
                            <MimeType>application/vnd.geo+json</MimeType>
                            <Schema>http://www.w3.org/TR/xmlschema-2/#dateTime</Schema>
                        </Format>
                    </Default>
                    <Supported>
                        <Format>
                            <MimeType>application/vnd.geo+json</MimeType>
                            <Schema>http://www.w3.org/TR/xmlschema-2/#dateTime</Schema>
                        </Format>
                    </Supported>
                </ComplexData>
</Input>
```

At the moment, when collecting info for the WPS inputs via the auto generated form, Terria will use a widget for this input with no values set: 

![terria-wps](https://github.com/TerriaJS/terriajs/assets/1842016/84035d25-0e69-4fc8-8e5e-9ac1c0d603bc)

Yet it is often the case with a time series layer (say from THREDDS) that a date-time has been selected in the time slider. Could this be used as a default value for the date-time widget?

sixlighthouses:
merged with PR #7026

PR #7026, opened 2024-01-10
Wps default date

### LINKED ISSUE #6896: WPS: Date or time widget reset after typing in values is unexpected (chrome)
When manually typing in either a date to the date time widgets in chrome (116.0.5845.188 or 117.0.5938.89 but this has been around in earlier versions too) the browser will reset and lose the value in the time widget. The opposite also happens in that, if you type in a time in the time widget, then your previously entered date will be reset and lost. e.g.

Typed in a date to the date widget:

typing-in-date

Then typing in a time to the time widget and the date widget gets reset: and my previously entered date is lost:

typing-in-a-time

This is a bit frustrating for power users who don't use the click helpers.

Doesn't seem to happen with firefox.

sixlighthouses:
### What this PR does

Fixes #6896 & #6895

Refactored the DateTimeParameterEditor  to a functional component and added the terria object to the props, this allows the component to load with a default dateTime -- either the current time or if a timeline is loaded it takes the currently loaded time from that

### Test me

Browse to 
http://ci.terria.io/wps-default-date/#share=s-bRTdoGD1E1hSA95nLCVwIEjNXa9

From the My Data tab select the AWAVEA WPS service and then the Wave: Annual Bivariate Probability Distribution service

Select a point location,  manually edit the start and end times.  Update the times manually several times to test that the other unedited values do not change.

image

Ensure the debugging console is open on the Network tab, hit Run Analysis

confirm the Payload of the request has the Date Time correctly formatted  as YYYY-MM-DDThh:mm

image

Upon loading the Analysis Tools, if the current Terria session has timeline loaded that time in the will become the default time in the dateTime inputs of the WPS process

image

### Checklist

- [x] There are unit tests to verify my changes are correct


sppigot:
@sixlighthouses - Thanks for your work on this!

Testing with no timeline:

- Default date/time does get set to current date/time in wps parameter form
- Manually changing dates or times works correctly

However, when if I set dates/times and then click on the map to get a location, the dates and times seem to be reset to the current date/time when the form reappears. Changing them back to what I had before seems to work but when I submit the form the start date/time value is messed up in the request that goes to the WPS:

e.g. 
messedupdatetime

Testing with timeline shortly.

sppigot:
@sixlighthouses - testing with a timeline - the default date does get set to the currently selected value in the timeline but seems to be off by 1 day for some reason? (Am I missing something obvious here?). Check the following image:

offby1day

Thanks for your work on this @sixlighthouses !

sixlighthouses:
@sppigot thanks for test and the feedback.  I have made some updates so when you get a chance please review again

CHANGES.md
```diff
@@ -55,6 +58,8 @@
 - `WebMapServiceCatalogItem` will drop problematic query parameters from `url` when calling `GetCapabilities` (eg `"styles","srs","crs","format"`)
 - Fixed regression causing explorer window not to display instructions when first opened.
 - Enable eslint for typescript: plugin:@typescript-eslint/eslint-recommended
+- Fix WPS date parameter reset bug
+- [The next improvement]
```
na9da: I think these need to go.

lib/ReactViews/Analytics/DateTimeParameterEditor.tsx
```diff
@@ -0,0 +1,97 @@
...
+          .format("YYYY-MM-DD");
+        newTimeValue = moment.utc(ct.toISOString()).local().format("HH:mm");
+      }
+    } else {
+      const ct = new Date(parameter.value);
+      newDateValue = moment.utc(ct.toISOString()).local().format("YYYY-MM-DD");
+      newTimeValue = moment.utc(ct.toISOString()).local().format("HH:mm");
+    }
+    parameter.setValue(CommonStrata.user, `${newDateValue}T${newTimeValue}`);
+  }, [parameter, terria.timelineStack.clock.currentTime]);
```
na9da, line 47: It is more correct to depend on `parameter.value` instead?

na9da (changes requested):
@sixlighthouses I think these changes look great. Thanks for taking time to refactor the component as `tsx`. 

I have just left a couple of comments.

lib/ReactViews/Analytics/DateTimeParameterEditor.tsx
na9da: On further thought, that could lead to an infinite loop.

na9da (approved):
Looks good!

PR #7031, opened 2024-01-19
Can enable/disable shorten share urls from initSources.

### LINKED ISSUE #6852: Add a url/init option to deselect share URL shortening for any map
This came up in the [discussion forum](https://github.com/TerriaJS/terriajs/discussions/6848).

Currently when a map has a URL shortening service enabled, the default behaviour is to shorten the URLs when creating a share/story link. This is the correct default behaviour. 

However, in some situations, we might want a user embedding or linking to an external map to be able to deselect URL shortening by default. This is to limit leaking of metadata from any private datasets (like an intranet service) that maybe added by other users who follow the link.

Adding a url option similar to `hideWorkbench=1` will not work because, currently url options are not copied when creating a new share link. 

The other solution is to make it an  [elements](https://github.com/TerriaJS/terriajs/blob/b14835b75232eb8be722c6255f9c6e1789b1f5cb/lib/Models/Terria.ts#L443) option [similar to this one](https://gist.githubusercontent.com/steve9164/5cef5f3de4e84f1a4f91766a218cdd65/raw/90c16f7f1d21885974ad7e1cf3067426495b9a1e/hide-elements.json) and then enable it via `#start` parameter.

This has to work both in story and global share URL context.

mwu2018:
### What this PR does

Fixes #6852 
When opening a map with `initSources` that has `settings.shortenShareUrls` property, the map will 
- Disable shortening shared URL function" if `shortenShareUrls` is `false`;  
- Enable shortening shared URL function if `shortenShareUrls` is `true`  or `undefined`. 
- The status of shortening shared URL function remains unchanged if `shortenShareUrls` is `undefined`. 

The behaviour will persist in window local storage. E.g. `initSources`:

```
{
  "initSources": [
    {
      "settings": {
        "shortenShareUrls": false,
        ...
      }
      ...
    }
  ]
}
```

### Test me
- An external app that opens the map normally will make request to URL with `#start=` similar to https://ci.terria.io/issue-6852/#start={%22initSources%22:[{%22settings%22:{%22shortenShareUrls%22:false},%22catalog%22:[]}]}. It will disable the shorten share URLs function.
- Alternatively, open https://ci.terria.io/issue-6852/#clean&https://gist.githubusercontent.com/mwu2018/edaf87bdccdbe9d6ddb4daab88ae7605/raw/9dcf8497ea6080625d521e9aa66f7f275886a079/wms-group-add-all-test.json
- Verify `Share/Print -> Advanced options -> Shorten the share URL using a web service` is always unchecked.

### Checklist


nf-s (changes requested):
Looks good @mwu2018 !

Can I please request one small change - can you move `shortenShareUrls` into `settings.shortenShareUrls`?

Thanks!

nf-s (changes requested):
Can you please also add the new property to the docs 

https://github.com/TerriaJS/terriajs/blob/853223acceeafad3af129b918e46013d5f8f8e50/doc/customizing/initialization-files.md#L306-L317

mwu2018:
@nf-s Could you please take another look? Thanks.

nf-s (approved):
Looks good, thanks Mike!

PR #7043, opened 2024-02-02
Fixed a bug in ArcGisMapServerCatalogItem

### LINKED ISSUE #7041: ArcGisMapServerCatalogItem should not use "pre-cached tiles" if a layer ID is specified in URL path.
Upload data to http://ci.terria.io/main from URL https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/MapServer/8 will result in error. 

***Note***
The logic is wrong at https://github.com/TerriaJS/terriajs/blob/b1b8753dac2b4fdfa4ef5fbe73bf4af6f0b8fb76/lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts#L588

It should be something like

```
  usePreCachedTilesIfAvailable:
    !this.layers &&
    ( this.layersArray.length === 0 ||
      setsAreEqual(
        this.layersArray.map((l) => l.id),
        stratum.allLayers.map((l) => l.id)
      )
    ),
```

mwu2018:
### What this PR does

Fixes #7041 

### Test me
Open [before](http://ci.terria.io/main) or [after](http://ci.terria.io/issue-7041) in browser and upload data from URL https://portal.spatial.nsw.gov.au/server/rest/services/NSW_Administrative_Boundaries_Theme/MapServer/8
then compare the results.
-  Before: Will not show boundaries in the map and will get many 404 errors in the dev tool panel.
-  After: Will show boundaries properly.

### Checklist

### DIFF
--- a/lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
+++ b/lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
@@ -102,15 +102,8 @@ class MapServerStratum extends LoadableStratum(
       token = await getToken(item.terria, item.tokenUrl, item.url);
     }
 
-    let layerId;
-    const lastSegment = item.uri.segment(-1);
-    if (lastSegment && lastSegment.match(/\d+/)) {
-      // URL is a single REST layer, like .../arcgis/rest/services/Society/Society_SCRC/MapServer/16
-      layerId = lastSegment;
-    }
-
     let serviceUri = getBaseURI(item);
-    let layersUri = getBaseURI(item).segment(layerId || "layers"); // either 'layers' or a number
+    let layersUri = getBaseURI(item).segment("layers");
     let legendUri = getBaseURI(item).segment("legend");
 
     if (isDefined(token)) {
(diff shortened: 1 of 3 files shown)


lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
```diff
@@ -153,10 +153,17 @@ class MapServerStratum extends LoadableStratum(
...
+      // The following conditional code is commented out becasue
+      // the test will be true when the requested url path ends with
+      // layer ID. Overwriting "layers" will make "stratum.allLayers"
+      // contain single layer instead of all layers that were set
+      // previously from the base URI, causing unexpected errors.
+      //
+      // If layersMetadata is only a single layer -> shove into an array
+      // else if (isDefined(layersMetadataResponse?.id)) {
+      // layers = [layersMetadataResponse];
+      // }
```
nf-s: It might be better to remove the case where layers metadata is returned for a single layer.

This could be done by changing

https://github.com/TerriaJS/terriajs/blob/48f705486fde742dd8a9fcc743890b2179f7c590/lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts#L113

to just

```ts
let layersUri = getBaseURI(item).segment("layers");
```

Then you can remove the entire block highlighted by this comment. 

This would mean that we get "detailed" layer metadata for all layers regardless of if layers have been specified (in the `url` or the `layers` trait) - which should fix the bug you found where `allLayers` only contained speciried layers

nf-s (changes requested):
You have fixed the bug, but I wonder if it might be a better approach to entirely remove the case where layer metadata is fetched for a single layer.

Please see my comment below.

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
nf-s: I am not entirely sure of the difference between "basic" and "detailed" layer metadata.

The "basic" layer metadata is the `layers` object in the MapServer service metadata. The "detailed" layer metadata is fetched by calling the `layers` MapServer endpoint.

https://github.com/TerriaJS/terriajs/blob/48f705486fde742dd8a9fcc743890b2179f7c590/lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts#L149-L156

mwu2018:
@nf-s Based on your comment, more lines are removed.

nf-s (approved):
Looks good, thanks Mike!

PR #7046, opened 2024-02-07
Replace leaflet with cesium 2d

### LINKED ISSUE #6550: Investigate replacing Leaflet viewer with Cesium in 2D scene mode
Example Cesium 2D map - https://sandcastle.cesium.com/?src=Rotatable%202D%20Map.html

Cesium can be put in 2D mode with one line of code (from a quick test).

This would greatly simplify some core code and reduce maintenance costs. 

While Leaflet is very light-weight - I'm not sure it makes sense when Terria is so heavy.
Leaflet (default) vector performance is also very poor.

Switching to cesium 2D mode would give us a few things for free
- Vector support/performance that is equal to 3D mode (probably better even)
- support for 4326 imagery services
- ...

Possible complications
- How does splitter work in 2D mode

nf-s:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/6550 https://github.com/TerriaJS/terriajs/issues/6739

Todo

- Look into @glughi [comment](https://github.com/TerriaJS/terriajs/issues/6550#issuecomment-1920810191) about `mapProjection` https://github.com/TerriaJS/terriajs/issues/6550#issuecomment-1920810191
- Try replacing preview map and pedestrian mode map to cesium

### Test me

### Checklist

### DIFF
--- a/lib/Map/DragPoints/DragPoints.js
+++ b/lib/Map/DragPoints/DragPoints.js
@@ -86,17 +86,11 @@ DragPoints.prototype._createDragPointsHelper = function (pointMovedCallback) {
   if (defined(this._dragPointsHelper)) {
     this._dragPointsHelper.destroy();
   }
-  if (this._terria.mainViewer.viewerMode === ViewerMode.Leaflet) {
-    this._dragPointsHelper = new LeafletDragPoints(
-      this._terria,
-      pointMovedCallback
-    );
-  } else {
-    this._dragPointsHelper = new CesiumDragPoints(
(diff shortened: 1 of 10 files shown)


nf-s:
We have decided against this for the following reasons

- Leaflet is useful to keep for low-performance devices
- We would need to replace the Leaflet preview map (for the catalog and pedestrian mode) with Cesium - which would have performance overhead
- Leaflet is more reliable and well-maintained than Cesium's 2D mode

PR #7059, opened 2024-02-23
Thredds palettes

sixlighthouses:
### What this PR does

Adds the functionality to allow users to make use of Thredds palettes for styling.

### Test me
Browse to
url: `http://ci.terria.io/thredds-palettes/`

Add a thredds server  -- example https://thredds.ereefs.aims.gov.au/thredds/wms/GBR4_H2p0_B3p1_Cq3P_Dhnd/monthly.nc?service=WMS&version=1.3.0&request=GetCapabilities

Add a layer

Select a style from the workbench control

You will now see the Palettes drop down

select a palette to see the layers style change n.b some layers will need to have the Color Scale Range edited to correctly use the palettes

### Checklist


sixlighthouses:
On first load of the Thredds server no style is selected, so no palettes are displayed
image


Once the user selects a style, we fetch the list of palettes associated with that style

image

sppigot:
Thanks @sixlighthouses - this is a really useful addition to the THREDDS WMS support! Please merge ASAP

lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
```diff
@@ -410,6 +466,29 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
       });
     }
 
+    if (this.catalogItem.isThredds) {
+      //   const paletteResult: StratumFromTraits<WebMapServiceAvailableLayerStylesTraits>[] =
+      // [];
```
ljowen:
```suggestion
```

lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
```diff
@@ -250,6 +307,8 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
           legendUri.setQuery("transparent", "true");
         }
 
+        i;
+
```
ljowen:
```suggestion
```

lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
```diff
@@ -410,6 +466,29 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
       });
     }
 
+    if (this.catalogItem.isThredds) {
+      //   const paletteResult: StratumFromTraits<WebMapServiceAvailableLayerStylesTraits>[] =
+      // [];
+
+      async () => {
```
ljowen: How does this anonymous function get called?

lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
```diff
@@ -410,6 +466,29 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
...
+        const res = result[0];
+        const styles = res.styles ?? null;
+        if (styles && styles[0].abstract) {
+          const palettes = await this.fetchPalettes(
+            styles[0].abstract,
+            styles[0].legend
+          );
+          console.log(palettes);
```
ljowen:
```suggestion
```

lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
sixlighthouses: That one is an IIFE -- I used it so I can call the async fetchPalettes without having to make the parent function async

sixlighthouses: Thanks for picking that one up Lawrence, it was never getting called, removed that code block

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -619,6 +620,72 @@ class WebMapServiceCatalogItem
...
+      const urlRegex = /(https?:\/\/[^\s]+)/g;
+      const selectedStyle = this.availableStyles[0].styles.filter(
+        (style) => style.name === this.styles
+      );
+      const abstract = selectedStyle[0]?.abstract;
+      const urls = abstract?.match(urlRegex);
+      const extractedUrl = urls?.[0];
+
+      if (!extractedUrl) {
+        reject("No URL found in abstract");
```
ljowen: Unless we also return here the fetch below will still execute and make a request to `<ur>/undefined` 

Should No url in the abstract be treated as an error? My understanding is it's an option property for thredds wms's so it might instead be more appropriate to `resolve([])`

sixlighthouses: Updated to resolve, thanks Lawrence

ljowen (approved):

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -671,6 +738,7 @@ class WebMapServiceCatalogItem
             );
             styles[layerIndex] = newStyle;
             this.setTrait(stratumId, "styles", styles.join(","));
+            this.fetchPalettes();
```
nf-s: Can you please add a condition here to `fetchPalettes` is only called if the server supports it? 

Currently, if you select a style for a non-THREDDs WMS - this causes an uncaught error

nf-s (changes requested):
I think there are a few architectural issues with this PR. 

The ncWMS `GetMetadata` with `item=layerDetails` (which is currently being extracted from the selected WMS style abstract) - actually has nothing to do with specific styles, it returns extra metadata for a given layer.

For example

- https://thredds.ereefs.aims.gov.au/thredds/wms/GBR4_H2p0_B3p1_Cq3P_Dhnd/monthly.nc?request=GetMetadata&item=layerDetails&layerName=CS_bleach

The JSON response includes the following things related to palettes

```json
{
  "scaleRange": [-50, 50],
  "palettes": [
    "default",
    "default-inv",
    "div-BrBG",
    ...
  ],
  "defaultPalette": "psu-viridis",
  "noPaletteStyles": ["contours"]
}
```

This means that the palettes should be visible if any `style` is selected that isn't in `noPaletteStyles` (styles that don't support palettes). The default palette should also be set to `defaultPalette`, and the default `colorScaleMinimum` and `colorScaleMaximum` should be set using `scaleRange`.

This `GetMetadata` JSON should be fetched in `WebMapServiceCatalogItem.forceLoadMetadata` - and should be in it's own LoadableStratum (eg something like `NcMWSGetMetadataStratum`). From that LoadableStratum, you can define values for `threddsPalette` and `threddsPalettes` (and also `colorScaleMinimum` and `colorScaleMaximum`).

- You can also check the current layer there, and return undefined if the layer name is in `noPaletteStyles`.

The GetMetadata URL can be constructed using something like

```ts
this.uri.clone()
  .setSearch({
    service: "WMS",
    version: this.useWmsVersion130 ? "1.3.0" : "1.1.1",
    request: "GetMetadata",
    item: "layerDetails",
    layerName: this.layersArray[0],
  })
  .toString()
```

While WMS does support multiple layers, I think we should just fetch `GetMetadata` for the first layer at the moment.

How you have setup `threddsPaletteDimensions` works well! 
So only minimal changes should be needed for that.

Finally, can you please add a `supportsGetMetadata` and `supportsPalette` traits? You can just copy [`supportsColorScaleRange`](https://github.com/TerriaJS/terriajs/blob/3e23262f6c30048c1379cd35993a8c5528419b03/lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts#L809-811). This will allow us to override the default behaviour - eg turn it off or on manually.

Also, sorry to nitpick, but all of this is ncWMS not THREDDs specific - so can you please rename the `WebMapServiceCatalogItem` traits to `palette` and `availablePalettes` (just to match the existing pattern with `availableStyles` etc) - and other thredds name properties/objects

Please let me know if you need any further info/clarification - thanks!

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
nf-s: Sorry this comment doesn't make sense with all the other changes I have suggested

sixlighthouses (commented):
added traits for
- colorScaleMinimum
- colorScaleMaximum
- defaultPalette
- noPaletteStyles

added checks for current style is included in noPaletteStyles

sixlighthouses (commented):
Todo
- [x] GetLegendGraphic failing when not using default
- [x] GetLegendGraphic being reset when changing styles

sixlighthouses:
Ready for re-review

lib/Traits/TraitsClasses/WebMapServiceCatalogItemTraits.ts
```diff
@@ -316,6 +326,45 @@ export default class WebMapServiceCatalogItemTraits extends mixTraits(
...
+  })
+  palette?: string = "default";
+
+  @primitiveArrayTrait({
+    type: "string",
+    name: "Palettes",
+    description:
+      "Used to store the palettes available for a layer. This is a non-standard property supported by THREDDS servers. This property is ignored unless WebMapServiceCatalogItem's isThredds is true. The default value is ['default']."
+  })
+  palettes?: string[] = [];
```
nf-s: Can you please rename this to `availablePalettes` (just to match the existing pattern with `availableStyles` etc)

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -619,6 +623,104 @@ class WebMapServiceCatalogItem
...
+            "defaultPalette",
+            data.defaultPalette
+          );
+          this.setTrait(
+            CommonStrata.user,
+            "noPaletteStyles",
+            data.noPaletteStyles
+          );
+          this.setTrait(CommonStrata.user, "showPalettes", true);
+          resolve(data.palettes);
```
nf-s: The `user` stratum should only be used for user-driven changes to traits. The current `fetchPalettes` implementation makes changes to the `user` stratum - this means that when a share-link is generated, the palettes will be included. 

Can you please move the `fetchPalettes` logic into a new `LoadableStratum` (you can call it `NcMWSGetMetadataStratum`) ?

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -665,12 +767,26 @@ class WebMapServiceCatalogItem
...
+              const newStyles = currentStyle + "/" + currentPalette;
+              this.setTrait(stratumId, "styles", newStyles);
+              if (this.noPaletteStyles.includes(newStyle)) {
+                this.setTrait(stratumId, "showPalettes", false);
+              } else {
+                this.setTrait(stratumId, "showPalettes", true);
+              }
+            } else {
+              this.setTrait(stratumId, "styles", styles.join(","));
+            }
```
nf-s: Same [here](https://github.com/TerriaJS/terriajs/pull/7059#discussion_r1588826518), I think you should keep `styles` and `palette` separate. After creating the `NcWMSLoadableStratum`, you can change `showPalette` there without explicitly calling `setTrait`.

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -225,12 +226,16 @@ class WebMapServiceCatalogItem
   }
 
   protected async forceLoadMetadata(): Promise<void> {
+    if (this.palettes.length === 0) {
+      this.fetchPalettes();
+    }
```
nf-s: This should be loaded after `GetCapabilitiesStratum`

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -619,6 +623,104 @@ class WebMapServiceCatalogItem
...
+          newPalette: string | undefined
+        ) => {
+          //if (this.stylesArray[0]) {
+          runInAction(() => {
+            this.setTrait(stratumId, "palette", newPalette);
+            const currentStyle = this.stylesArray[0]
+              ? this.stylesArray[0]
+              : "default";
+            const newStyles = currentStyle.split("/")[0] + "/" + newPalette;
+            this.setTrait(stratumId, "styles", newStyles);
```
nf-s: I think there are some issues with updating `styles` like this - you should keep `palette` and `styles` separate.

With the current implementation, `palette` is not used unless the user makes changes - for example - if `palette` was defined in a catalog JSON file, it wouldn't be used.

You can apply the palette to the `styles` parameter here instead. You should only apply the palette to the first style - as this maps to the first layers style.

https://github.com/TerriaJS/terriajs/blob/bacefe0a8ac9e69aceff4138f4eb4ce2b0714c9e/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts#L566

nf-s (changes requested):
Thanks @sixlighthouses this is definitely a good step forward. There are still a few architectural issues with this PR. 

It is important to follow the Terria model layer patterns that I mentioned [previously](https://github.com/TerriaJS/terriajs/pull/7059#pullrequestreview-1965192798).

If you are unsure how to apply the Terria model layer patterns, there is some documentation here:

- https://docs.terria.io/guide/contributing/model-layer/
- https://docs.terria.io/guide/contributing/traits-in-depth/
- https://docs.terria.io/guide/contributing/strata-examples/

Please let me know if you want any further info, we can setup a meeting to discuss this.

sixlighthouses:
Updated as per feedback, ready for review

nf-s:
Todo

- [ ] Write tests

PR #7064, opened 2024-03-05
Fix magda group reload

tephenavies:
### What this PR does

- Improved `ChartableMixin.isMixedInto` to ensure there are no false positive matches when testing References.
- Fixed a bug in `MagdaReference` where members of a group would not be updated/created correctly when a group is reloaded.

### Test me

Tested on dev eap. Unfortunately I haven't had time to write/update unit tests. I will do that in a subsequent PR.

### Checklist


na9da (approved):
Approving. Couldn't find any immediate issues with this change.

PR #7084, opened 2024-03-20
Fix Cesium resolving relative baseUrl wrongly by passing in absolute URL

### LINKED ISSUE #7081: Latest TerriaJS version broke `/story/` prefix path
Test link

- http://ci.terria.io/main/story/DEACoastlines#configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json

Notice in network tab that a few cesium resources will incorrectly use `http://ci.terria.io/main/story/build` as the baseHref

image

This is most likely because of the latest cesium upgrade

tephenavies:
### What this PR does

Fixes https://github.com/TerriaJS/terriajs/issues/7081

### Checklist


nf-s (changes requested):
This seems to break the `/story/` path


If you go to

- http://ci.terria.io/fix-cesium-baseUrl/story/DEACoastlines#configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json

Cesuim correctly loads assets, but the story doesn't load

tephenavies:
Interesting find. This is actually because of case sensitivity. CI builds this deployment for http://ci.terria.io/fix-cesium-baseurl/, but our nginx also supports navigating to it at http://ci.terria.io/fix-cesium-baseUrl/ (something I didn't know until now).

Case insensitivity of our CI cannot be trivially disabled (use of regex in `path` automatically uses nginx case insensitive regex, and our rewrites that remove the branch name to make routes work with the docker images rely on nginx ingress regex paths).

And I'm not sure that TerriaJS has any reason to support case insensitive base href matching (IMO URLs should be case sensitive).

An argument could be made that we should support (and maybe redirect) `/Story/`, but I'd move that to a new issue.

TL;DR: issue is to do with case sensitivity, there is no change to this behaviour in this PR, and is no different to fetching e.g. http://ci.terria.io/MAIN/story/xxxx

nf-s (approved):
Oh interesting - I didn't mean to do that.

I agree supporting case insensitive base href matching is unnecessary - I reckon this is good to go!

PR #7087, opened 2024-03-22
Fix ArcGisMapServerImageryProvider tileInfo bug

nf-s:
### What this PR does

There is a bug in Ceisum's code here - there is an `await` missing

https://github.com/TerriaJS/cesium/blob/33d48c5987ff3b6176bff315a32c693550696c2c/packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js#L767

which means that when `mapServerData` is provided, it doesn't parse metadata correctly.

For example here - `useTiles` is supposed to be set to false

https://github.com/TerriaJS/cesium/blob/33d48c5987ff3b6176bff315a32c693550696c2c/packages/engine/Source/Scene/ArcGisMapServerImageryProvider.js#L138-L141

But because of the missing `await` - the imagery provider is created before this value is changed.

To fix this in the short-term I have added our own check for `tileInfo`

### Test me

- Add the following web data
- `https://go-vicmap-age.shared.landusevictoria.systems/arcgis/rest/services/dtv_foi_poly_melb/MapServer`
- Before http://ci.terria.io/main/
- After http://ci.terria.io/arcgismapserver-tileinfo/
- Before you will not see the layer on the map (and also a bunch of 404 erros)
- After you will see the layer

### Checklist

- [x] ~I've updated relevant documentation in `doc/`.~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
 - Fixed exception thrown from `objectArrayTrait` when a model has 0 strata and a `MergeStrategy` of `topStratum`.
 - Fix `generateCatalogIndex` after `searchProvider` changes
 - Fix bug with relative URLs being ignored in `generateCatalogIndex`
+- Fix bug with ArcGisMapServerImageryProvider not correctly identifying if the `tile` endpoint can be used
 - [The next improvement]
 
 #### 8.6.1 - 2024-03-14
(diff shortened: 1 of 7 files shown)


staffordsmith83 (approved):
Looks good, I would maybe have indicated that the additions to `ArcGisInterfaces` and `ArcGisMapServerCatalogItem` were temporary, but we dont know how long it will take to fix this bug in Cesium so lets assume these changes to be permanent

PR #7114, opened 2024-04-26
Provide ambient types or TSify some js[x] files.

na9da:
### What this PR does

Currently, running `tsc` from an external library like a plugin fails with many `... element implicitly has type "any"` errors. This is because `tsc` will not type check js/jsx modules from external libraries in `node_modules`. These errors can be suppressed by turning off [noImplicitAny option](https://www.typescriptlang.org/tsconfig/noImplicitAny.html), however this is pretty unsafe.

This PR tries not to change any functionality, instead we typify just enough files so that `tsc` will succeed without having to add any extra configuration. We make the following specific modifications:
- If it is trivial, we convert `.js[x]` files to `.ts[x]` and make necessary changes so that it type checks
- For non-trivial `.js[x]` files, we create an ambient `.d.ts` file with type signatures for all the exports
- Upgrades `markdown-it` and `linkify-it` so that we can `import` their ESM versions and typings. 

### Test me

Please just do a random test adding different catalog items, via catalog explorer, web url and drag-n-drop, open feature info panel, charts etc.

### Checklist

- [ ] ~There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~ Fixed broken specs.


package.json
```diff
@@ -132,9 +135,9 @@
     "javascript-natural-sort": "^0.7.1",
     "json5": "^2.1.0",
     "leaflet": "^1.8.0",
-    "linkify-it": "^2.0.0",
+    "linkify-it": "^5.0.0",
     "lodash-es": "^4.17.11",
-    "markdown-it": "^11.0.0",
+    "markdown-it": "^14.1.0",
```
pjonsson, line 141: I had this in my pile of security fixes, this fixes CVE-2022-21670.

lib/Core/arrayContains.ts
```diff
@@ -0,0 +1,8 @@
+export default function arrayContains<T>(array: readonly T[], value: T) {
+  for (let i = 0; i < array.length; ++i) {
+    if (array[i] === value) {
+      return true;
+    }
+  }
+  return false;
+}
```
tephenavies: I've removed this (only used in 1 place in terriajs) and replaced with `Array.prototype.includes`.

lib/ReactViewModels/ViewState.ts
```diff
@@ -449,8 +447,6 @@ export default class ViewState {
       }
     );
 
-    this._disclaimerHandler = new DisclaimerHandler(terria, this);
```
tephenavies, line 452: Is this not needed?

lib/ReactViews/StandardUserInterface/StandardUserInterface.tsx
```diff
@@ -68,10 +67,7 @@ const StandardUserInterfaceBase: React.FC<StandardUserInterfaceProps> =
     });
 
     const handleDragOver = (e: React.DragEvent<HTMLDivElement>) => {
-      if (
-        !e.dataTransfer.types ||
-        !arrayContains(e.dataTransfer.types, "Files")
-      ) {
+      if (!e.dataTransfer.types || !e.dataTransfer.types.includes("Files")) {
```
tephenavies, line 70: I've made this change to remove `arrayContains`

tephenavies (commented):
Looks good. Just need to confirm the DisclaimerHandler changes.

lib/ReactViewModels/ViewState.ts
na9da: This was used in v7 for showing catalog item initial message. In v8 this is handled here:

https://github.com/TerriaJS/terriajs/blob/d4ccee3f8476f7737f615ee06e8806f3d43ddd2b/lib/ModelMixins/MappableMixin.ts#L175-L179

lib/Models/LocationMarkerUtils.ts
```diff
@@ -1,12 +1,14 @@
 import i18next from "i18next";
-import markerIcon from "./markerIcon";
 import prettifyCoordinates from "../Map/Vector/prettifyCoordinates";
 import CommonStrata from "./Definition/CommonStrata";
 import CzmlCatalogItem from "./Catalog/CatalogItems/CzmlCatalogItem";
 import Terria from "./Terria";
 import LatLonHeight from "../Core/LatLonHeight";
 import { toJS } from "mobx";
 
+//@ts-ignore
```
pjonsson: I think the ts-ignore adds a lint warning, should this be `@ts-expect-error <string that explains why>`?

tephenavies (approved):

na9da:
Merging. Thanks @steve9164 @pjonsson.

PR #7127, opened 2024-05-02
Remove `string-replace-webpack-plugin` to fix critical security warning.

na9da:
### What this PR does

Removes [string-replace-webpack-plugin](https://www.npmjs.com/package/string-replace-webpack-plugin) which generates a [security warning](https://github.com/advisories/GHSA-76p3-8jx3-jpfq) due to one of its stale dependency.

The plugin was mainly used for replacing babel generated getters with `superGet` in mobx 4. But after mobx upgrade this is no longer required.

It is currently only used for [updating](https://github.com/TerriaJS/terriajs/blob/b21e457fcda8ff4ce0aa1c95eb3348b69f72513e/buildprocess/configureWebpack.js#L77-L88) the [credit string in Ion.js](https://github.com/CesiumGS/cesium/blob/c82102b8eda2f86e4e90acd023893296de57bd2e/packages/engine/Source/Core/Ion.js#L46-L50) that is shown when using the default Ion token.

It looks like this:
image

I'm not sure if it is a very useful customization to have, so I have dropped it in favor of keeping the build script simpler. However if the consensus is to add it back, I can use a more updated webpack plugin like [string-replace-loader](https://www.npmjs.com/package/string-replace-loader) or see if we can customize it run time instead of build time.

### Test me

Shouldn't affect the app except it will now show Cesium's default credit line when a default access token is used.

### Checklist

### DIFF
--- a/buildprocess/configureWebpack.js
+++ b/buildprocess/configureWebpack.js
@@ -1,6 +1,5 @@
 const path = require("path");
 const CopyPlugin = require("copy-webpack-plugin");
-const StringReplacePlugin = require("string-replace-webpack-plugin");
 const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const ForkTsCheckerNotifierWebpackPlugin = require("fork-ts-checker-notifier-webpack-plugin");
 const webpack = require("webpack");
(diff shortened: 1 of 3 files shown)


buildprocess/configureWebpack.js
```diff
@@ -51,66 +50,6 @@ function configureWebpack(
...
-          pattern: /\/\*[\S\s]*?\*\//g, // find multi-line comments
-          replacement: function (match) {
-            // replace http:// and https:// with a spelling-out of it.
-            return match.replace(/(https?):\/\//g, "$1-colon-slashslash ");
-          }
-        }
-      ]
-    })
-  });
-
```
na9da, line 113: The `sprintf` credits this was trying to replace is [no longer a dependency of Cesium](https://github.com/CesiumGS/cesium/issues/9796#issuecomment-922063923).

buildprocess/configureWebpack.js
```diff
@@ -51,66 +50,6 @@ function configureWebpack(
...
-          replacement: function () {
-            return 'Please set "cesiumIonAccessToken" in config.json';
-          }
-        },
-        {
-          pattern: / before making any Cesium API calls/g,
-          replacement: function () {
-            return "";
-          }
-        }
```
na9da, line 88: Do we **really** need this?

tephenavies: I think it's pretty important. The default message is flat out wrong.

tephenavies:
I think substituting the Cesium Ion key default message is important to us

buildprocess/configureWebpack.js
na9da: Ok, I'll work out a way to include it - I thought it was too small to read.

pjonsson:
This dependency was removed in #7351, so this PR should probably be closed.

zoran995:
this was moved to terriajs-cesium

PR #7133, opened 2024-05-03
Show feature info panel chart error inline

na9da:
### What this PR does

When feature info panel chart loading fails, show an inline error, instead of a popup.

Current behaviour is to pop an error message to the user, but  the load function being connected to the react life-cycle methods,  can be called several times. This results in as many pop ups  to the user. Also if every feature has an error, then user will have to dismiss the errors for each one of them. I think a better solution might be to show the error inline instead of popping it to the user.

eg:
image

### Test me

Compare behaviour in before and after links.

Before:
[Test link for main branch](http://ci.terria.io/main/#share=s-xXVKBF4Gg8Aei4HBsHqtT2pTgJk&clean&https://gist.githubusercontent.com/na9da/a3df09bf3bf08655dba9279e8efb0d67/raw/1ae28eed679a93aeb0380dc8460f18003d0cad40/gistfile1.json).
clicking anywhere on the dataset resulted in the error popup being shown 4-5 times

After:
[Test link for this branch](http://ci.terria.io/show-chart-error-inline/#share=s-xXVKBF4Gg8Aei4HBsHqtT2pTgJk&clean&https://gist.githubusercontent.com/na9da/a3df09bf3bf08655dba9279e8efb0d67/raw/1ae28eed679a93aeb0380dc8460f18003d0cad40/gistfile1.json).
no error popups are shown.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 #### next release (8.7.2)
 
+- When feature info panel chart loading fails, show an inline error, instead of a popup.
 - Add NumberParameterEditor to enable WPS AllowedValues Ranges to be set and use DefaultValue
 
 #### 8.7.1 - 2024-04-16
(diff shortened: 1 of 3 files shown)


na9da:
Closing, fixed in https://github.com/TerriaJS/terriajs/pull/7156 as part of another refactor effort.

PR #7144, opened 2024-05-07
Upgrade protomaps leaflet and remove mvt imagery provider

### LINKED ISSUE #6858: Figure out our future usage of `protomaps.js` - now called `protomaps-leaflet`
The JSON styling module is going to be removed - see

- https://github.com/protomaps/protomaps-leaflet/issues/112

We use this in `MapboxVectorTileCatalogItem` to support a subset of the MapboxGL style specification

https://github.com/TerriaJS/terriajs/blob/ed7c97fc722a1719e3b24f661bdde41c6bf7edce/lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts#L148-L153

This is not a huge issue, we just need to look at how the library API has changed, and move https://github.com/protomaps/protomaps-leaflet/blob/754458357228c5a82e85930812cd461d0e9f7d59/src/compat/json_style.ts into TerriaJS

We should also look into https://www.npmjs.com/package/mapbox-expression - as we may be able to use it to support more complex MapboxGL style expressions

nf-s:
### Upgrade protomaps leaflet

Fixes 

- https://github.com/TerriaJS/terriajs/issues/7240
- #6858  

~To be merged into https://github.com/TerriaJS/terriajs/pull/6725~

Todo

- [x] Finish copying json style to `lib/Map/Vector/MapboxJsonStyleToProtomaps.ts`
- [x] Fix feature picking on load (for share links)

### To test

This PR shouldn't impact usage of region mapping or MapboxVectorTileCatalogItem - so here are some examples to compare

MVT with simple Mapbox style spec
- http://ci.terria.io/main/#share=s-kMGzeYzE1ymufQMkJ0rcdJ3FUdL&clean&https://gist.githubusercontent.com/nf-s/a232b3f39eca70e3cf727cbca259ee0c/raw/7e9113e2135d41fb4162efb19c707cef15c252ba/esri-mvt.json
- http://ci.terria.io/upgrade-protomaps-leaflet/#share=s-kMGzeYzE1ymufQMkJ0rcdJ3FUdL&clean&https://gist.githubusercontent.com/nf-s/a232b3f39eca70e3cf727cbca259ee0c/raw/7e9113e2135d41fb4162efb19c707cef15c252ba/esri-mvt.json

Region mapping (with picked feature)

- http://ci.terria.io/main/#share=s-umjL4Q2NW2FQ4UYYOENMeR9iMtp&configUrl=https://terria-catalogs-public.storage.googleapis.com/pacificmap/map-config/prod.json
- http://ci.terria.io/upgrade-protomaps-leaflet/#share=s-umjL4Q2NW2FQ4UYYOENMeR9iMtp&configUrl=https://terria-catalogs-public.storage.googleapis.com/pacificmap/map-config/prod.json


buildprocess/configureWebpack.js
```diff
@@ -430,6 +434,11 @@ function configureWebpack(
     require.resolve("react-dom")
   );
 
+  config.resolve.alias["protomaps-leaflet"] = path.dirname(
+    require.resolve("protomaps-leaflet")
+  );
+  config.resolve.alias["pmtiles"] = path.dirname(require.resolve("pmtiles"));
+
```
nf-s: I'm not entirely sure why these need to be here - but the build was failing without them

package.json
```diff
@@ -142,12 +141,13 @@
     "mustache": "^2.2.1",
     "node-fetch": "^2.6.1",
     "papaparse": "^5.2.0",
+    "pmtiles": "^3.0.7",
```
pjonsson, line 145: There is a pmtiles 4.0 released now, and that mentions removed pmtiles spec v2 support which significantly reduces bundle size. Is it worth upgrading to v 4.0, or does TerriaMap need pmtiles spec v2 support?

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,190 @@
...
+export class ProtomapsGeojsonSource implements TileSource {
+  /** Data object from Options */
+  private readonly data: string | FeatureCollectionWithCrs;
+
+  /** Resolved geojsonObject (if applicable) */
+  @observable.ref
+  geojsonObject: FeatureCollectionWithCrs | undefined;
+
+  /** Geojson-vt tileIndex (if applicable) */
+  tileIndex: Promise<any> | undefined;
```
pjonsson: Can the `tileIndex` get a more precise type than any, or does that information flow from untyped code?

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  constructor(url: string | FeatureCollectionWithCrs) {
+    makeObservable(this);
+    this.data = url;
+    if (!(typeof url === "string")) {
```
pjonsson: Is the condition the same as `typeof url !== "string"`?

yarn.lock
```diff
@@ -9421,10 +9502,10 @@ protocol-buffers-schema@^3.3.1:
   resolved "https://registry.yarnpkg.com/protocol-buffers-schema/-/protocol-buffers-schema-3.6.0.tgz#77bc75a48b2ff142c1ad5b5b90c94cd0fa2efd03"
   integrity sha512-TdDRD+/QNdrCGCE7v8340QyuXd4kIWIgapsE2+n/SaGiSSbomYl4TjHlvIoCWRpE7wFt02EpB35VVA2ImcBVqw==
 
-protomaps@1.19.0:
-  version "1.19.0"
-  resolved "https://registry.yarnpkg.com/protomaps/-/protomaps-1.19.0.tgz#cfa929b5e2979ca2f040f60ab9bad861845a799c"
-  integrity sha512-S+Nb3s5QR1Skw6cy2uDC+Ih6MX3cNvFwgZFukAujO8nB9yH4MhEtMqlDmO3WOT2mdrBYmQBvx9xx8KeM0Kp36Q==
+protomaps@1.23.0:
```
pjonsson: Shouldn't this entry and its dependencies disappear from yearn.lock after the update to protomaps-leaflet?

package.json
```diff
@@ -31,7 +31,6 @@
     "@mapbox/geojson-merge": "^1.1.1",
     "@mapbox/point-geometry": "^0.1.0",
     "@mapbox/togeojson": "^0.16.2",
-    "@mapbox/vector-tile": "^1.3.0",
```
pjonsson, line 34: Should ` lib/ThirdParty/mapbox__vector-tile` also be deleted if this package is removed?

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,190 @@
...
+    let result: FeatureCollectionWithCrs | undefined;
+    if (typeof this.data === "string") {
+      result = toFeatureCollection(await (await fetch(this.data)).json());
+    } else {
+      result = this.data;
+    }
+
+    runInAction(() => (this.geojsonObject = result));
+
+    return geojsonvt(result, {
```
pjonsson: Result can be `undefined` here, and I'm not sure `geojsonvt()` will handle that well.

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,190 @@
...
+                  bbox.maxY = g1[1];
+                }
+                return new Point(g1[0], g1[1]);
+              })
+            ];
+            numVertices = transformedGeom.length;
+          }
+
+          const feature: ProtomapsFeature = {
+            props: f.tags,
```
pjonsson: The `tags` in f is typed as `any` in this file, but the tag is optional so this will try to assign `something | undefined` to a field of type `JsonObject` (https://github.com/protomaps/protomaps-leaflet/blob/da47e14fe6ba0c4a17c3a8d3dc9921f0f99f39f0/src/tilecache.ts#L13).

package.json
nf-s: The `pmtiles` dependency is linked to our usage of `protomaps-leaflet` - which doesn't support v4 

- https://github.com/protomaps/protomaps-leaflet/issues/177

lib/Map/Vector/ProtomapsGeojsonSource.ts
nf-s: Good point, I have brought in `geojson-vt` types!

lib/Map/Vector/ProtomapsGeojsonSource.ts
nf-s: Yep, that was copy pasted in, thanks for the tip

package.json
nf-s: Good find, i'll remove that!

lib/Map/Vector/ProtomapsGeojsonSource.ts
nf-s: Thanks for spotting this, I've handled this now

lib/Map/Vector/ProtomapsGeojsonSource.ts
nf-s: Thanks, I'll add a fallback here

yarn.lock
nf-s: Ive updated the lockfile

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/** Extent (of coordinates) of tiles generated by geojson-vt */
+const geojsonvtExtent = 4096;
+
+/** Layer name to use with geojson-vt
+ *  This must be used in PaintRules/LabelRules (eg `dataLayer: "layer"`)
+ */
+export const GEOJSON_SOURCE_LAYER_NAME = "layer";
+
+export class ProtomapsGeojsonSource implements TileSource {
```
na9da: Might be good to add a top level doc comment on what this class does.

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,149 @@
...
+  toFeatureCollection
+} from "../../ModelMixins/GeojsonMixin";
+import {
+  PROTOMAPS_DEFAULT_TILE_SIZE,
+  PROTOMAPS_TILE_BUFFER
+} from "../ImageryProvider/ProtomapsImageryProvider";
+import { featureCollection } from "@turf/helpers";
+
+/** Extent (of coordinates) of tiles generated by geojson-vt */
+const geojsonvtExtent = 4096;
```
na9da: nitpick: Maybe this could also be renamed as `GEOJSON_VT_EXTENT` just to be consistent?

lib/Map/Vector/mapboxStyleJsonToProtomaps.ts
```diff
@@ -0,0 +1,319 @@
...
+      return `${style}${weight}${t(z, f)}px ${fontFaces
+        .map((f) => f.face)
+        .join(", ")}`;
+    };
+  }
+  console.log("Can't parse font: ", obj);
+  return (_: number) => "12px sans-serif";
+}
+
+export function mapboxStyleJsonToProtomaps(
```
na9da, line 198: Function name makes it clear but I think a top level doc comment might also be good.

lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts
```diff
@@ -89,22 +84,6 @@ class MapboxVectorTileLoadableStratum extends LoadableStratum(
...
-      const geojsonBbox = bbox(this.item.imageryProvider.source.geojsonObject);
-      return createStratumInstance(RectangleTraits, {
-        west: geojsonBbox[0],
-        south: geojsonBbox[1],
-        east: geojsonBbox[2],
-        north: geojsonBbox[3]
-      });
-    }
-  }
 }
```
na9da, line 87: Was this intentionally removed?

lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts
```diff
@@ -200,7 +173,8 @@ class MapboxVectorTileCatalogItem extends MappableMixin(
 
     return new ProtomapsImageryProvider({
       terria: this.terria,
-      id: this.uniqueId,
+      // Use the URL as the id, this is needed for backward compatibility with MapboxImageryProvider, for when picking features (as it uses the URL as the id)
+      id: this.url,
       data: this.url,
```
na9da: The change to `forceProxy` got me thinking, ideally, shouldn't we wrap `this.url` in a call to `proxyCatalogItemUrl()`? Although, that will be a slightly new behaviour and probably most MVT endpoints are set up to allow CORS anyway.

lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts
```diff
@@ -142,8 +116,9 @@ class MapboxVectorTileCatalogItem extends MappableMixin(
   @computed
   get parsedJsonStyle() {
     if (this.style) {
-      return json_style(this.style, new Map());
+      return mapboxStyleJsonToProtomaps(this.style, {});
```
na9da, line 119: Not something introduced by this PR, but I was debugging why [this catalog definition](https://gist.github.com/na9da/756cc3f6fefcc56ceccaa8232d63d65b) wouldn't work and crashes the app when opening `About data` both in `main` and this branch. 

I think the [styleUrl](https://tiles.arcgis.com/tiles/GB33F62SbDxJjwEL/arcgis/rest/services/Vicmap_Vegetation_Tree_Density/VectorTileServer/resources/styles) returns a `Token required` error as  JSON but with HTTP 200 OK status. This makes our code think it is a valid style object and continues to parse it, but crashes at [this point](https://github.com/TerriaJS/terriajs/blob/5dca224b216c6f785194c1c12fa38caeedc4a2f7/lib/Map/Vector/mapboxStyleJsonToProtomaps.ts#L203) when trying to iterate `layers` which is `undefined`.

I think wrapping this call in a try catch would solve it?

Repro:
- Open [this CI link](http://ci.terria.io/upgrade-protomaps-leaflet/#share=s-y58ZfbCXnUFnyDD1QDEpeF7t6Wa&clean&https://gist.githubusercontent.com/na9da/756cc3f6fefcc56ceccaa8232d63d65b/raw/5de03f00e6c8075ec0f7502290ae18c48a659ee7/tree-density-mvt.json)
- Click about data
- App crashes

na9da (changes requested):
@nf-s Looks great, I have made a few minor comments.

lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts
pjonsson: > I think wrapping this call in a try catch would solve it?

That sounds reasonable, but the type for the first parameter to `mapboxStyleJsonToProtomaps` is `any`, so I think there should be a `if (!obj || !obj.layers) return undefined;` on the first line inside `mapboxStyleJsonToProtomaps` so it doesn't crash on inputs that can be expected.

lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts
nf-s: Yep there shouldn't ever be a geojson source when using `MapboxVectorTileCatalogItem`

lib/Models/Catalog/CatalogItems/MapboxVectorTileCatalogItem.ts
nf-s: 👍

nf-s:
Thanks for the suggestions, I have applied requested changes. This is ready for another review, thanks!

na9da (approved):
Looks good @nf-s. I leave it to you to merge.

PR #7147, opened 2024-05-10
Issue 6758

### LINKED ISSUE #6758: Make the currently selected style (display variable) available in feature info template.
This could be useful to show a feature-info-panel chart relevant to the currently selected style/table column. To achieve this, the catalog maintainer can define a feature info template that looks something like:

```
  <chart x-column="date" y-column="{{terria.activeStyle.id}}"  source="https://example.org/api/{{station_id}}/{{terria.activeStyle.id}}" />
``` 

Few things to note here:
- The `activeStyle.id` should correspond to the column name and the request parameter for fetching data from the API
- This works only for scalar columns. If the selected column is not scalar, the API may not return any relevant data and the chart will be blank.

mwu2018:
### What this PR does

Fixes #6758 

### Test me
- [after](http://ci.terria.io/issue-6758#clean&https://gist.githubusercontent.com/mwu2018/9dc97c35b163ae5af0fc7b84c9806d41/raw/8235cf0302e9e4e28b423eb4ee5e32703d9c34c8/testFeatureInfoSection.json)
  - Select a variable in the workbench.
  - Click on a region.
  - Observe that the "Active style" in the feature info panel should match the selected variable name.
- [before](http://ci.terria.io/main#clean&https://gist.githubusercontent.com/mwu2018/9dc97c35b163ae5af0fc7b84c9806d41/raw/8235cf0302e9e4e28b423eb4ee5e32703d9c34c8/testFeatureInfoSection.json)
  - "Active style" in the feature info panel is always undefined (empty).
 
### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 #### next release (8.7.2)
 
 - Add NumberParameterEditor to enable WPS AllowedValues Ranges to be set and use DefaultValue
+- FeatureInfoSection has access to activeStyle of item having TableTraits.
 
 #### 8.7.1 - 2024-04-16
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 #### next release (8.7.2)
 
 - Add NumberParameterEditor to enable WPS AllowedValues Ranges to be set and use DefaultValue
-- FeatureInfoSection has access to activeStyle of item having TableTraits.
+- Feature info template has access to activeStyle of item having TableTraits.
 
 #### 8.7.1 - 2024-04-16
 
--- a/lib/ReactViews/FeatureInfo/FeatureInfoSection.tsx
+++ b/lib/ReactViews/FeatureInfo/FeatureInfoSection.tsx
@@ -217,7 +217,7 @@ export class FeatureInfoSection extends React.Component<FeatureInfoProps> {
       currentTime?: Date;
       timeSeries?: TimeSeriesContext;
       rawDataTable?: string;
-      activeStyle?: string;
+      activeStyle?: Object;
     } = {
       partialByName: mustacheRenderPartialByName(
         this.props.catalogItem.featureInfoTemplate?.partials ?? {},
@@ -260,8 +260,12 @@ export class FeatureInfoSection extends React.Component<FeatureInfoProps> {
     }
 
     // Add activeStyle property
-    if (TableMixin.isMixedInto(this.props.catalogItem))
-      terria.activeStyle = this.props.catalogItem.activeStyle;
+    if (TableMixin.isMixedInto(this.props.catalogItem)) {
+      const activeStyleId = this.props.catalogItem.activeStyle;
+      terria.activeStyle = this.props.catalogItem.styles.find(
+        (style) => style.id === activeStyleId
+      );
+    }
 
     // If catalog item has featureInfoContext function
     // Merge it into other properties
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -3,6 +3,7 @@
 #### next release (8.7.2)
 
 - Add NumberParameterEditor to enable WPS AllowedValues Ranges to be set and use DefaultValue
+- FeatureInfoSection has access to activeStyle of item having TableTraits.
```
na9da: Nitpick: It would be better to say `Feature info template`.

lib/ReactViews/FeatureInfo/FeatureInfoSection.tsx
```diff
@@ -257,6 +259,10 @@ export class FeatureInfoSection extends React.Component<FeatureInfoProps> {
       );
     }
 
+    // Add activeStyle property
+    if (TableMixin.isMixedInto(this.props.catalogItem))
+      terria.activeStyle = this.props.catalogItem.activeStyle;
```
na9da: In the future we might want to expose other attributes of the `activeStyle` like  the name of the table column this style maps to. So I think it might be better to define `terria.activeStyle` as an object which for now has this shape `{id: this.props.catalogItem.activeStyle}`. 

To see why this might be needed, [check this link](http://ci.terria.io/issue-6758/#share=s-aRONIRxouqIi0YaH7i1URCbVQx9&https://gist.githubusercontent.com/mwu2018/9dc97c35b163ae5af0fc7b84c9806d41/raw/d6c13107236b9fd181b09cc8e8a352be95ad6cdf/testFeatureInfoSection.json) where I have created a new "User Style" which maps to a column called `sta_name`. In the feature info, I might sometimes want access this underlying column instead of the `id`. Making `activeStyle` an object makes it possible to expose the color column as `{id: "User Style", pointColorColumn: "sta_name"}`

na9da (changes requested):
@mwu2018 - looks great, just added couple of comments.

lib/ReactViews/FeatureInfo/FeatureInfoSection.tsx
mwu2018: @na9da I've revised the code based on your comment.
You can also test [this](http://ci.terria.io/issue-6758/#share=s-aRONIRxouqIi0YaH7i1URCbVQx9&https://gist.githubusercontent.com/mwu2018/9dc97c35b163ae5af0fc7b84c9806d41/raw/8235cf0302e9e4e28b423eb4ee5e32703d9c34c8/testFeatureInfoSection.json)

lib/ReactViews/FeatureInfo/FeatureInfoSection.tsx
```diff
@@ -257,6 +259,14 @@ export class FeatureInfoSection extends React.Component<FeatureInfoProps> {
       );
     }
 
+    // Add activeStyle property
+    if (TableMixin.isMixedInto(this.props.catalogItem)) {
+      const activeStyleId = this.props.catalogItem.activeStyle;
+      terria.activeStyle = this.props.catalogItem.styles.find(
+        (style) => style.id === activeStyleId
+      );
```
na9da: @mwu2018 the danger in exposing the entire style object is that the catalog maintainer can now depend on any member of the style object. If we decide to change any of it in the future, it'll silently break the template. So ideally, we should restrict the members we export. I recommend you export just the `id` for now.

mwu2018: Done. Please take another look.

na9da (approved):
Looks good @mwu2018.

PR #7158, opened 2024-05-20
Add i3s catalog item

### LINKED ISSUE #6416: Add cesium native I3S support
See PR - https://github.com/CesiumGS/cesium/pull/9634

Depending on PR outcome - we can always merge it into https://github.com/TerriaJS/cesium

ljowen:
### What this PR does

Fixes #6416

### Test me

Init source:
http://ci.terria.io/add-i3s-support/#clean&https://gist.githubusercontent.com/ljowen/445d4ba31ef521b36875ae7555b8a78b/raw/a821943f2e993ca8ae23fe739e71b8ce258edc1e/frankfurtI3S.json

Other I3S examples:
- Library BIM: https://tiles.arcgis.com/tiles/cFEFS0EWrhfDeVw9/arcgis/rest/services/Turanga_Library/SceneServer
- NYC: https://tiles.arcgis.com/tiles/z2tnIkrLQ2BRzr6P/arcgis/rest/services/NYC_Attributed_v17/SceneServer

### Known Issues
- Highlighting breaks when changing LoD, for example select a feature, zoom / pan away the feature will no longer be highlighted. (Similarly if a feature is selected while the dataset is loading in): See example here in cesium sandcastle: [link](https://sandcastle.cesium.com/#c=lVhrc9s2Fv0rWG23kWYdyokz3Y4sezZWnNazzqOVuv0QZRKIhCQ0IMEFQMtKqv++5wIgRSpKm2YmMglc3OfBwQWHQzYRVlY5+1UblbGZMIbLgq2FXK0ds1vrRM4yLWzxwLGcu3TNnGY3Z1OWccetcPZAFoutzgVTOuVO6sIycScKtpFuzVZCy4ylurgTxmIumRfDIZsKx56a9IebKZtJJTJ2rcSdX9u4szL8TrotM0JxB4muSTi0go5UaSsMZGylnIVuGLKO3UmxwfAFK8QmBpv814/1573Uv0904WBGmHnvhH2aF4y5YHjUXhSd6dM8qweD497vxu0o+NroO5kJkyyNzn8xKi5kbN5bO1fa0XAo6iVnWcJNupI2SXU+DI9DROKGCOlOws2hL1Bj4+zZMJrB003OV2IKQYogWBnQn8EJ/fJC5n7NiC25ssIPOpkLhZDbY9rA25siE6XAT+FmhhdWVako0m1bLrecT3leKmFH7MkJQw1FwRd4ZTlSL62fg10nuZLcymLFltowmZdICcp3J23FFfsfflDVebEbnM8Lj4UX2og9slBYpyvDUl6wBRRmma+9EUkSoDNbC1YZxUpuLaYCMp9heZ17Zquy1Aa6lOYZOcIZ+QP3KNB7LDp7xqYIUbA+1g6Y4lss6yfjtbbucuinQmqHfsoOxzK7HCBZUJVqpUTqoaqXzHo1QeqohgEjKGBda6zBKUVqAVOPv3lvViH7K85u5cJws533RqGwe/A4gM62YeMHhh8fu+Lmg7n96fHVzx/Nd6+Po2nKi+cwkEqb6ndnz14tfkMc9t2jd/86FvIpNgbK5KvkzDb4iPxPjMB+ZPLMMl2G3Y4aUDXiO6faGJ4LRzmprFhWymMhE4tqtfIFKWpAyI94J80hIdD6KiqNWQGWs98q617AqAG2nqpyzV/oDCh2phIeiVCeCuYADE6zLMc0OQX8AFPetiNcwy1AnCgJziG/Xr1d681zhFQZgnajM/VhWoKK9onyagQHGS6DtI+CK6U3PgO8LNWWceeMXFS0cikVXI4x0OR0my+00qttywxNSAjrytHWtGzBCdUAF4VDnBvAaWh7GkFJBG0pG9SmXKUV0eNLbXLs1ZbiFQpqfARLCLAiCDC5JMXwE/7n0vp9CrIlQ4tquUTFSHOoervcNfuzMu6yTs2arXfB+IZLVzPlwc48YEWP/jdHYP/2JAjs0eCZLTq1131VSUUb3KP3lvJ0fV8q0AnRvNNaLbin95aHg31kT7PMJ5mQHJJcx8a4ZRSpzKWTd4GcmNuWglaGsyXxGz9pZLAps6x/3NAEoBPEaDkqQqWlZOoCkJXuQWApke3zmQbxizrQnwE+TvSVhJm2lUTcOwySMRYXJvGgvGBnp6fJ6XkTLe2i2okos9g2MLua3rKgjBYobJuFVT6n0NS2SMnwSgGlfnQxE0uAN+vXS5LAIQP27bfsYCxRolihL7hkp4N6h7csHUi/OX3rbe1qg52D+DOzYBp1HTIC018rm3zM+X04PGuPuqn8Z8uvg2V770LxMjCuLEIn01TwWilZWvRBya8/TL9/kqTcOI3+plzLdKYneIMcr7uMYLtGfAO4ULkEhEadTD862rI3ar/4HbRrgfAVqN2ggGxNnRoYyhHH6BptBIMOQFuWacWVX5DQCJGvSt7h9MmJAZsoA2NOwnCMZVkV4ajsN6k9DGiptjPdbyb/JCT6twt5iXmnlx18pza1jxi1iaaoIBq7xo9Ff97PqJVgG9rgBfMzgemJB+utQJTFaryP2DefvNzufWMtNi43xZJY1cfHF2DwkMXK0CmDs8AKahTA5vHAqE/9ZjwecXF6xKoiojS2ZhKHJVcTnBmm05X6kf6gOZ5peyMa9F7ol+NptQG21mjBlULF0eVLOPtVXoa2DdILfc8yaUvsxEPPr4OpTn8dxvqUIGrwcaoynLLcc6eMxxm0Xul7tD8NiRohimnJU3GNG4P7EchRNC7cTVFW7qkHT79BkS5uxdJNlEw/+ILmaC1zT1mtmu/djMc6/KxHkhjkfl+8hi6QPQXSlKnWU2KuraVL/eRE7UBSaivJxUFDjZ85cXHRVdjsCCPwWtRMEv26WcKruiSE19KIO6kr2yrZCSFG+9SuQVSKyOoLzHyYgEFj/XAGbSXQ1c5ZB4jnx1dBvkFvO5I9Z//twKODXPz++1eIwbcisPtfEk+oS05wiL0EdXWJvi7pIbC7wXyhRj/WSSf4HN3wRxPV8fG81dL4BCe40xbis0Awc/KFkgTMHVmw5+ag+er2l+uI/KZ1o5Qc+nQ0bwHY4TmhnuW5FCqz/UEC/LX26J7pqYkAf6dGlvE8nPde6laLPO+d7yULHAfRudq9pbcQ+g9vFlf+YPW5NtHXg0TFLN9k9RlBAAw3neSD2Np+0Dk41oawrrf7E+nB2NFVl6UKt5yL+gvCQxno7CGQwnEHnpHMvHc5dgudbS8fnNcKiFD7rYheIlKixOgKa5195G0dbJhO1ty+2tCHhVIY8GujYdBZ2PUdDcv7sTNwZX35zadmyW48xMDYZZfvz/9oaVxi3zQr39LS7BI/5mAteXy4CamUvveTlhb/ue/MVx+F/sxux9Zu/9I87o7Vzsc/DHXAXypM4/Zuj4ujfrdci07Ne1XxodCbosHrrkuCgTWSKB6BfESguxlab+d/TEbdgfOmtduhA4ghTA9O0hluK8nt9fPZu8ntzeQ/oRFs9uhBXcr44EvTPkhjRO35xOlbDS8nuKf296ddCIDkE/oKqKgPpOs2LmI0OO8N/AF42hD3UVH0Ln9BGj2fWGm6MEbxDrMHxvYX4n2THgf9dy3fN/ZOemPrtsBHXfR/y5y+HtE3pn6SDJ3IS7ph2+GiAtG4JLV20KDy7/GW2YbzgqcfVkbjBBkxs1rw/pPHJ6z+f5p8P2hhusStEa3niD06Le9b4wv/Xe6h4Zms6INbZxLXjofhZkLrTv/RmkI7YpZKbx5uR4xXTp8fbJPxsB3sOJN3TGYXR76JNlxH952p/OiJbQj5z5bGj2x0v6BOsdckcrx+dHkbJpMkGQ/xWjtxTE9zX+/Y+T8)
- Some features such as searching and clipping box are not yet implemented

### Checklist


lib/Core/getDataType.ts
```diff
@@ -180,6 +180,10 @@ const builtinLocalDataTypes: LocalDataType[] = [
     value: "shp",
     name: "core.dataType.shp",
     extensions: ["zip"]
+  },
+  {
+    value: "i3s",
+    name: "core.dataType.i3s"
```
na9da: This should perhaps go under `builtinRemoteDataTypes` next to 3d-tiles as we only support adding i3s from the `Add web data` tab.

lib/Models/Catalog/CatalogItems/I3SCatalogItem.ts
```diff
@@ -0,0 +1,98 @@
...
+  runInAction,
+  toJS
+} from "mobx";
+import BoundingSphere from "terriajs-cesium/Source/Core/BoundingSphere";
+import Cartesian2 from "terriajs-cesium/Source/Core/Cartesian2";
+import isDefined from "../../../Core/isDefined";
+import I3SCatalogItemTraits from "../../../Traits/TraitsClasses/I3SCatalogItemTraits";
+import CreateModel from "../../Definition/CreateModel";
+import { ModelConstructorParameters } from "../../Definition/Model";
+import { CatalogMemberMixin, MappableMixin } from "terriajs-plugin-api";
```
na9da: This import should be made from the relative path, eg: `../../../../ModelMixins/CatalogMemberMixin`. 

Because the plugin-api is not a dependency of terriajs (but the other way round), imports from plugin-api should only be used from external packages. However TS autoimport will sometimes choose it automatically. I guess we should look at [configuring something like this](https://stackoverflow.com/questions/73505774/option-to-exclude-patterns-from-auto-import-not-working) to avoid this problem.

lib/Models/Catalog/CatalogItems/I3SCatalogItem.ts
```diff
@@ -0,0 +1,98 @@
...
+        .map((layer) => layer.tileset?.boundingSphere)
+        .filter(isDefined)
+    );
+    runInAction(() => {
+      this.dataProvider?.layers.forEach(({ tileset }) => {
+        if (!tileset) {
+          return;
+        }
+        /* Control "lightness" of textures */
+        if (this.lightingFactor) {
```
na9da: Do we need to initialise style and lightingFactor here as we are also doing it in `mapItems` below. I also think doing it in `mapItems` is better since it will recompute on user input whereas forceLoadMapItems won't.

lib/Models/Catalog/CatalogItems/I3SCatalogItem.ts
```diff
@@ -0,0 +1,98 @@
...
+    this.dataProvider = await I3SDataProvider.fromUrl(this.url, {
+      geoidTiledTerrainProvider: this.terrainURL
+        ? await ArcGISTiledElevationTerrainProvider.fromUrl(this.terrainURL)
+        : undefined
+    });
+    this.boundingSphere = BoundingSphere.fromBoundingSpheres(
+      this.dataProvider.layers
+        .map((layer) => layer.tileset?.boundingSphere)
+        .filter(isDefined)
+    );
```
na9da: Nitpick: This could be split off as a `@computed get boundingSphere()` accessor. Not much difference either way, but one pro is that it separates out the logic for computing the bounding sphere.

lib/Traits/TraitsClasses/I3SCatalogItemTraits.ts
```diff
@@ -0,0 +1,17 @@
...
+import Cesium3DTilesCatalogItemTraits from "./Cesium3DTilesCatalogItemTraits";
+
+@traitClass({
+  description: `Creates an I3S item in the catalog from an slpk.`,
+  example: {
+    type: "I3S",
+    name: "CoM Melbourne 3D Photo Mesh",
+    id: "some-unique-id"
+  }
+})
```
na9da, line 13: I didn't know this existed :100:

lib/ModelMixins/Cesium3dTilesMixin.ts
na9da, line 1: I see two alternate approaches:
1. Create a `Cesium3DTilesStyleGenerator` module/class which can then be composed from within the Mixins.
2. To be consistent with the mixin pattern, we could create a `Cesium3dTilesStyleMixin` and extract these functions into it.

I might go with 2 just to be consistent. 

One other problem with simple loose functions is that it is hard to memoize some results using `@computed`. For eg, with this change `showExpressionFromFilters` will always be re-computed when the user changes the opacity, even though opacity does not affect its computation.

na9da (commented):
@ljowen - looking great and good job figuring out the complexities of the model layer. I have just added a few suggestions.

lib/ModelMixins/Cesium3dTilesMixin.ts
ljowen: Thanks I went with option 2.

lib/Models/GlobeOrMap.ts
```diff
@@ -258,9 +263,13 @@ export default abstract class GlobeOrMap {
         this._removeHighlightCallback = function () {
           if (
             isDefined(feature._cesium3DTileFeature) &&
-            !feature._cesium3DTileFeature.tileset.isDestroyed()
+            feature._cesium3DTileFeature.tileset.isDestroyed() === false
           ) {
-            feature._cesium3DTileFeature.color = originalColor;
+            try {
+              feature._cesium3DTileFeature.color = originalColor;
```
ljowen, line 265: This throws an error as internally the features model reference becomes stale when LoD changes

lib/Models/Catalog/CatalogItems/I3SCatalogItem.ts
```diff
@@ -0,0 +1,152 @@
...
+export default class I3SCatalogItem extends Cesium3dTilesStyleMixin(
+  FeatureInfoUrlTemplateMixin(
+    ShadowMixin(
+      MappableMixin(
+        UrlMixin(CatalogMemberMixin(CreateModel(I3SCatalogItemTraits)))
+      )
+    )
+  )
+) {
+  static readonly type = "I3S";
```
ljowen:
```suggestion
  static readonly type = "i3s";
```

test/Models/Catalog/CatalogItems/I3SCatalogItemSpec.ts
```diff
@@ -0,0 +1,173 @@
...
+    item = new I3SCatalogItem("test", new Terria());
+    runInAction(() => {
+      item.setTrait("definition", "url", testUrl);
+      item.setTrait("definition", "allowFeaturePicking", true);
+    });
+  });
+
+  it("should have a type and a typeName", function () {
+    expect(I3SCatalogItem.type).toBe("I3S");
+    expect(item.type).toBe("I3S");
```
ljowen:
```suggestion
    expect(item.type).toBe("i3s");
```

test/Models/Catalog/CatalogItems/I3SCatalogItemSpec.ts
```diff
@@ -0,0 +1,173 @@
...
+  beforeEach(function () {
+    item = new I3SCatalogItem("test", new Terria());
+    runInAction(() => {
+      item.setTrait("definition", "url", testUrl);
+      item.setTrait("definition", "allowFeaturePicking", true);
+    });
+  });
+
+  it("should have a type and a typeName", function () {
+    expect(I3SCatalogItem.type).toBe("I3S");
```
ljowen:
```suggestion
    expect(I3SCatalogItem.type).toBe("i3s");
```

lib/ModelMixins/Cesium3dTilesStyleMixin.ts
```diff
@@ -0,0 +1,139 @@
...
+      if (
+        !isDefined(this.style) &&
+        (!isDefined(this.opacity) || this.opacity === 1) &&
+        !isDefined(this.showExpressionFromFilters)
+      ) {
+        return;
+      }
+
+      const style = clone(toJS(this.style) || {});
+      const opacity = clone(toJS(this.opacity));
```
na9da, line 108: I think for 3D items it makes sense to have a default opacity of 1.0. To do that we'll need to override the default value through the loadable stratum [similar to 3d tiles catalog item](https://github.com/TerriaJS/terriajs/blob/3529651d22262c3d0ebd08750e17c0c241d773d9/lib/ModelMixins/Cesium3dTilesMixin.ts#L58).

lib/ModelMixins/Cesium3dTilesStyleMixin.ts
```diff
@@ -0,0 +1,139 @@
...
+type BaseType = Model<Cesium3dTilesTraits>;
+
+const DEFAULT_HIGHLIGHT_COLOR = "#ff3f00";
+
+function Cesium3dTilesStyleMixin<T extends AbstractConstructor<BaseType>>(
+  Base: T
+) {
+  abstract class Cesium3dTilesStyleMixin extends Base {
+    constructor(...args: any[]) {
+      super(...args);
```
na9da, line 40: I think you need to call makeObservable here in the mixin constructor - https://mobx.js.org/subclassing.html

lib/Models/GlobeOrMap.ts
```diff
@@ -235,7 +237,10 @@ export default abstract class GlobeOrMap {
         // Get the highlight color from the catalogItem trait or default to baseMapContrastColor
         const catalogItem = feature._catalogItem;
         let highlightColor;
-        if (catalogItem instanceof Cesium3DTilesCatalogItem) {
+        if (
+          catalogItem instanceof Cesium3DTilesCatalogItem ||
+          catalogItem instanceof I3SCatalogItem
+        ) {
```
na9da: Maybe now it makes sense to change this test to `hasTrait(HighlightColorTraits, "highlightColor")` to be more generic?

na9da (changes requested):
Looks good @ljowen - i have left some comments, will do another pass to check the feature picking. 

Just noting some things we should  tackle maybe in a separate PR:
- Splitter support (easyish)
- Clipping box (might need some thinking/refactoring)

lib/ModelMixins/Cesium3dTilesStyleMixin.ts
ljowen: Thanks I missed that!

lib/Models/GlobeOrMap.ts
ljowen: Good idea :)

lib/ModelMixins/Cesium3dTilesStyleMixin.ts
ljowen: Agree

ljowen:
> Looks good @ljowen - i have left some comments, will do another pass to check the feature picking.
> 
> Just noting some things we should tackle maybe in a separate PR:
> 
> * Splitter support (easyish)
> * Clipping box (might need some thinking/refactoring)

I've added splitter support, I think Clipping box might have to be a follow up

na9da (approved):
@ljowen - Looks good to go. 

On the feature selection issue, we have the same problem for 3d-tiles so might be good to create a separate issue.

The build error seems to be from prettier.

PR #7172, opened 2024-05-28
Fix position of draggable point after moving

### LINKED ISSUE #7171: Cesium drag point problem
The points drawn on the map by the "_Line Measure_" tool are clamped to the ground. But when a point is moved it is misplaced (because the new position is calculated using the `camera.pickEllipsoid` function) and, if the ground hides features, the points can disappear under the ground.

glughi:
### What this PR does

Fixes #7171 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 #### next release (8.7.4)
 
-- [The next improvement]
+- Fix position of draggable point after moving.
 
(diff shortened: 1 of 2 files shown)


na9da (approved):
I have tested it by measuring on an uneven terrain and it fixes the previous behavior of placing the dragged point on the ellipsoid instead of the terrain. Thanks very much for the PR @glughi

PR #7176, opened 2024-05-31
Add to the doZoomTo function the case of an imagery layer with imageryProvider.rectangle

glughi:
### What this PR does

Adds the "doZoomTo" function for imagery layers that have defined the rectangle property, such as "ion-imagery" layers.

### Checklist


na9da (approved):
Thanks @glughi!

PR #7208, opened 2024-06-20
Add GeoJson clustering

glughi:
### What this PR does

This PR adds the trait `ClusteringTraits` to `GeoJsonTraits`. In this way, it becomes possible to enable the clustering of entities made available by Cesium. Clusterized entities are represented as billboards generated by the `fromText` function of `PinBuilder` and report the number of grouped entities. As a result, clustering only works by using Cesium as a viewer, but there are no drawbacks to switching from Cesium to Leaflet at runtime.

### Test me

Add a GeoJson or WFS layer of points to the catalog and use new parameters
In example:
```
"clustering": {
  "enabled": true
},
```

### Checklist


lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -602,6 +604,33 @@ function GeoJsonMixin<T extends AbstractConstructor<BaseType>>(Base: T) {
           });
         } else {
           const dataSource = await this.loadGeoJsonDataSource(geoJsonWgs84);
+
+          if (this.clustering.enabled) {
+            const pinBackgroundColor = this.clustering.pinBackgroundColor;
```
pjonsson, line 609: I don't know anything about `runInAction(...)`, but a lot of the surrounding code uses that for accessing things in `this`. Does this code not need that for some reason?

glughi: Hi @pjonsson,
`runInAction`, like `action`,  is required when you modify an `observable`. 
This code acts before the modification and the modification is already in a `runInAction` (at line 635).
So, this code not need `runInAction`.

glughi:
Hello, any news?

niktayv:
@glughi Hi, Giovanni. Thank you for your efforts and submission! The code looks good to me, but I could not observe any clustering on my local copy after adding your suggested parameters to the "GeoJSON Test" item from the default `simple.json` in TerriaMap. Here is what I did:
```json
        {
          "id": "DL4Av6wY5h",
          "type": "geojson",
          "name": "GeoJSON Test",
          "url": "test/bike_racks.geojson",
          "clustering": {
            "enabled": true
          }
        },
```
What did I miss?

glughi:
Hello @niktayv, you're right. This clustering is currently only activated if you define`style`.

niktayv:
Ah, thank you @glughi! GeoJSON features should be rendered as Cesium primitives, as you said in the description.

So, I added ` "forceCesiumPrimitives": true,` to the item's configuration:
```json
        {
          "id": "DL4Av6wY5h",
          "type": "geojson",
          "name": "GeoJSON Test",
          "url": "test/bike_racks.geojson",
          "forceCesiumPrimitives": true,
          "clustering": {
            "enabled": true
          }
        },
``` 

After that, clustering for the test dataset in the 3D mode worked:

Screenshot 2025-10-17 at 3 32 24 PM

niktayv (approved):
LGTM. Thank you, @glughi!

PR #7226, opened 2024-06-27
Cog support cesium only 2 warn reprojection

staffordsmith83:
### What this PR does

This is an incomplete PR to add a shortReport warning when adding a COG in a non-native projection.
Not currently working.

### Problem Explanation
The main problem we are trying to solve is ensuring that the shortReport computed property can access imageryProvider.tilingScheme correctly. Here are the key points:

- imageryProvider is defined as a computed property.
- imageryProvider is an asynchronous object that updates its internal state after it is fully loaded. Specifically, imageryProvider.ready becomes true once it is ready.
- The shortReport computed property depends on imageryProvider.tilingScheme, which is only accessible when imageryProvider is ready.
- We need to set up a reaction to imageryProvider.ready so that the shortReport property is recomputed whenever imageryProvider becomes ready.

### Solution Approach
To solve this problem, we could:
Set up a reaction to imageryProvider.ready to update tilingScheme and trigger recomputation of shortReport.

### Implementation Challenge
One potential solution involves extending the constructor of MappableMixin to set up the reaction. However, this approach is tricky because calling super() requires several arguments that are not easily accessible in the context of CogCatalogItem.

### Test me

### Checklist

### DIFF
--- a/lib/Models/Catalog/CatalogItems/CogCatalogItem.ts
+++ b/lib/Models/Catalog/CatalogItems/CogCatalogItem.ts
@@ -35,7 +35,6 @@ export default class CogCatalogItem extends MappableMixin(
   }
 
   @override
-  @computed
   get shortReport(): string | undefined {
     let content = "";
     // Warn for 2D mode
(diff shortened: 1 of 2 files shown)


na9da:
Closing as we have merged this feature into https://github.com/TerriaJS/terriajs/pull/7209

Issue #4250, opened 2020-04-23
Map goes under the surface when zooming to geojson point type


soyarsauce:
Repro
1. http://ci.terria.io/mobx/ (or http://ci.terria.io/master/ as well)
2. drag and drop this geojson (or add web data): https://gist.githubusercontent.com/soyarsauce/3d96c09c930232d878cee544f96c1078/raw/5a741da119807ba4e865c2693906ea2a435333d6/tjs-test.geojson https://gist.github.com/soyarsauce/3d96c09c930232d878cee544f96c1078
3. observe that you're now in space (for mobx), or underground with a mirrored basemap (in master)

soyarsauce:
![image](https://user-images.githubusercontent.com/9134432/80064252-f6b85000-857a-11ea-94db-732b9a2621d7.png)

should just zoom in and show the point above ground

PR #7234, opened 2024-07-10
Cesium2d viewer

glughi:
### What this PR does

Added Cesium in 2D mode (`SCENE2D`) to possible map viewers; _WebMercator_ is set as map projection when using this mode. Cesium 2D is faster than Leaflet and the behavior of the two views (3d and 2d) in this way is homogeneous. Leaflet is not removed from the project and you can use one, the other or both.
Added the parameter `mapViewers` with which in `config.json` you can indicate which map viewers must be enabled and available in the MapSettings panel (the possible values are now "3d", "3dsmooth", "2d" and "cesium2d"). By default it keeps the known behavior.

### Test me

Put `"mapViewers":["3d", "2dcesium"],` in your `config.json` and reload. In _MapSettings_ panel you should see the new option.

### Checklist


na9da:
@glughi - Thanks for this and other PRs you have created. We currently have a bit of a backlog but will hopefully be able to review yours soon :)

lib/ViewModels/TerriaViewer.ts
```diff
@@ -173,6 +179,8 @@ export default class TerriaViewer {
       viewerFromPromise = leafletFromPromise.get();
     } else if (this.attached && this.viewerMode === ViewerMode.Cesium) {
       viewerFromPromise = cesiumFromPromise.get();
+    } else if (this.attached && this.viewerMode === ViewerMode.Cesium2D) {
+      viewerFromPromise = cesium2dFromPromise.get();
```
pjonsson: To me, the body of `cesiumFromPromise` looks identical to the body of `cesium2dFromPromise`. Can't this entire else-if be folded into the previous else if, but with the condition `this.attached && (this.viewerMode === ViewerMode.Cesium || this.viewerMode === ViewerMode.Cesium2D)`?

lib/Models/Terria.ts
```diff
@@ -373,6 +373,11 @@ export interface ConfigParameters {
    */
   searchBarConfig?: ModelPropertiesFromTraits<SearchBarTraits>;
   searchProviders: ModelPropertiesFromTraits<SearchProviderTraits>[];
+
+  /**
+   * List of the enabled MapViewers: 3d, 3dsmooth, 2d, cesium2d
```
pjonsson: The type says this is an array rather than a list. The JSDoc for the other array-options seems to start with just "The", so "The enabled MapViewers .." in this case.

lib/Models/ViewerMode.ts
```diff
@@ -24,6 +25,12 @@ export const MapViewers = Object.seal({
     terrain: false,
     label: "settingPanel.viewerModeLabels.Leaflet",
     available: true
+  },
+  cesium2d: {
+    viewerMode: ViewerMode.Cesium2D,
+    terrain: false,
+    label: "settingPanel.viewerModeLabels.Leaflet",
```
pjonsson, line 33: The other entries have unique labels, so I'm guessing this entry shouldn't use the already used Leaflet-label.

lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -247,15 +253,17 @@ class SettingPanel extends React.Component<PropTypes> {
...
-            {Object.entries(MapViewers).map(([key, viewerMode]) => (
-              <SettingsButton
-                key={key}
-                isActive={key === currentViewer}
-                onClick={(event: any) => this.selectViewer(key as any, event)}
-              >
-                <Text mini>{t(viewerMode.label)}</Text>
-              </SettingsButton>
-            ))}
+            {Object.entries(MapViewers)
```
pjonsson: The previous line contains `elementsNo={3}`, which coincides with `MapViewers.length`. Does this work as intended when the filtering filters out some objects?

And even without filtering, why does the previous line not need to be updated to contain 4 since there is one more item in MapViewers in this PR?

lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -218,14 +218,18 @@ const SettingPanel: FC = observer(() => {
...
-              key={key}
-              isActive={key === currentViewer}
-              onClick={(event: any) => selectViewer(key as any, event)}
-            >
-              <Text mini>{t(viewerMode.label)}</Text>
-            </SettingsButton>
+          {Object.entries(MapViewers)
+            .filter(([key, viewerMode]) =>
+              viewerMode.available &&
+              this.props.terria.configParameters.mapViewers.indexOf(key) >= 0)
```
niktayv: Shoud be `terria.configParameters.mapViewers.indexOf(key) >= 0)`.

niktayv:
@glughi Thanks, Giovanni, for your continued work on this PR!

I approved the CI/build workflow, and from the results it seems that there is not too much left to get the `yarn install`, `yarn gulp build`, and `yarn gulp test-firefox` pass locally.

Well, at least, `tsc -b tsconfig-node.json` should work without errors.

I commented out the duplicated imports and changed the [line 224](https://github.com/TerriaJS/terriajs/pull/7234/files#diff-55e5d7f7c2302a7b3c4a60f84092fb758e0e8a319c648fe6725f39f826b18b68R224) as in my comment.
Running the `yarn gulp test-firefox` command, I got a couple of tests failing.

I wonder why the duplicate imports slipped in. Perhaps, you need to double-check that the `main` branch was at the latest state before merging into your branch.

Please let us know if you need more help.

niktayv:
I've done some more testing on a local TerriaMap instance.

1. The in config.json should be `"mapViewers": ["3d", "2dcesium"],` (I updated the description above).
2. There should be a translation for the button's label, e.g. in `wwwroot/languages/en/translation.json` add `"settingPanel": "viewerModeLabels": { "Cesium2D": "Cesium 2D", }`.
3. The logic for the "3dsmooth" viewer mode appeared to be broken, as with the `"mapViewers": ["3dsmooth", "2dcesium"],` clicking on the "3D Smooth" button did not return the viewer to a 3D mode.
4. Clicking the "Ideal Zoom" button caused the app to crash with some Cesium-related errors (tested on the "GeoJSON Test" and "Simple Test (CZML)" items).

glughi:
Sorry, there had been merge issues with the branch that was left too far behind. Now everything should be ok.

glughi:
Hi @niktayv, what do you think now?

lib/Models/Terria.ts
```diff
@@ -371,6 +371,11 @@ export interface ConfigParameters {
   searchBarConfig?: ModelPropertiesFromTraits<SearchBarTraits>;
   searchProviders: ModelPropertiesFromTraits<SearchProviderTraits>[];
 
+  /**
+   * The enabled MapViewers: 3d, 3dsmooth, 2d, 2dcesium
+   */
+  mapViewers: string[];
```
niktayv: Consider referencing the existing `ViewerMode` type from `./ViewerMode` instead of `string[]` to get compile-time validation of viewer names.

lib/Models/ViewerMode.ts
niktayv: Update documentation to include the new viewer mode.

The @param description lists "3d, 3dsmooth or 2d" but now "2dcesium" is also a valid option.
```ts
 /**
  * Returns the viewer type for the given viewer mode
  *
- * @param viewerMode 3d, 3dsmooth or 2d
+ * @param viewerMode 3d, 3dsmooth, 2dcesium or 2d
  */
```

lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -266,15 +266,21 @@ const SettingPanel: FC = observer(() => {
...
+            )
+            .map(([key, viewerMode]) => (
+              <SettingsButton
+                key={key}
+                isActive={key === currentViewer}
+                onClick={(event: any) => selectViewer(key as any, event)}
+              >
+                <Text mini>{t(viewerMode.label)}</Text>
+              </SettingsButton>
+            ))}
```
niktayv: We need to guard against undefined mapViewers and handle the empty state.

Two concerns with this filtering logic:
- Potential runtime error: If `terria.configParameters.mapViewers` is undefined or null, calling `.indexOf()` will throw. Consider adding a defensive check.
- Empty viewer list: If the filter returns 0 results (e.g., due to a misconfiguration), users will see a "Map View" label with no buttons, potentially confusing. Consider showing a fallback or warning.

Additionally, prefer `.includes(key)` over `.indexOf(key) >= 0` for readability.

Please verify that terria.configParameters.mapViewers is always guaranteed to be a defined array, or add a fallback:
```ts
 {Object.entries(MapViewers)
   .filter(
     ([key, viewerMode]) =>
       viewerMode.available &&
-      terria.configParameters.mapViewers.indexOf(key) >= 0
+      terria.configParameters.mapViewers?.includes(key)
   )
   .map(([key, viewerMode]) => (
```

niktayv (changes requested):
Everything seems to be working well, apart from some UI quirks that may be caused by discrepancies between the old and new logic.

In particular, the MapSettings panel does not retrieve the correct button from local storage. For example, in Local Storage `TerriaMap.viewermode` was `2dcesium`, but after reloading the page, in MapSettings, the highlighted button was "2D", not the "2D Cesium", as expected. Also, in MapSettings, the button corresponding to the current view mode should be highlighted by the white border, but switching between 2D Cesium and 2D viewer modes does not change the button highlighting.

Can you figure out how to restore the expected behaviour of the buttons in MapSettings?

lib/Models/Cesium.ts
```diff
@@ -1034,6 +1041,13 @@ export default class Cesium extends GlobeOrMap {
     const scene = this.scene;
     const camera = scene.camera;
 
+    if (scene.mode === SceneMode.SCENE2D) {
+      const rect = camera.computeViewRectangle();
+      if (rect) {
+        return new CameraView(rect);
+      }
+    }
+
```
niktayv, line 1062: Consider handling the fallback case when computeViewRectangle() returns undefined in 2D mode.

If `camera.computeViewRectangle()` returns undefined (which can happen in edge cases like extreme zoom levels), the code falls through to the 3D logic path, which may produce unexpected results in 2D mode, particularly with `scene.globe.pick()` and frustum calculations that assume 3D perspective.

I was testing the `GeoJSON/Testing stroke styles applied to embedded unstyled point` item in the demo catalogue. The item was not loading, so I clicked "Ideal Zoom", which brought me to a very high zoom level, and the map got black. Clicking the Home View button was causing a crash:  "3D map failed to render".

I tried to add a line for returning a fallback CameraView explicitly for the 2D case:

```ts
 if (scene.mode === SceneMode.SCENE2D) {
   const rect = camera.computeViewRectangle();
   if (rect) {
     return new CameraView(rect);
   }
+  // Fallback for 2D mode when computeViewRectangle returns undefined
+  return new CameraView(this.terriaViewer.homeCamera.rectangle);
 }
```

While this looks plausible, it did not fix the issue that I ran into.

lib/Models/ViewerMode.ts
```diff
@@ -38,6 +48,16 @@ export function setViewerMode(
     if (viewerMode === "3d" || viewerMode === "3dsmooth") {
       viewer.viewerMode = ViewerMode.Cesium;
       viewer.viewerOptions.useTerrain = viewerMode === "3d";
+      if (viewer.terria.cesium) {
+        viewer.terria.cesium.scene.mode =
+          viewerMode === "3d" ? SceneMode.SCENE3D : SceneMode.COLUMBUS_VIEW;
+      }
```
niktayv, line 53: We need to double-check the logic here.

Is using SceneMode.COLUMBUS_VIEW for "3dsmooth" correct?.

Columbus View is a 2.5D flat map projection, not a 3D globe with a smooth ellipsoid. 
The "smooth" in "3dsmooth" refers to the absence of terrain, not a projection change.

Both "3d" and "3dsmooth" should likely use `SceneMode.SCENE3D`, with only the terrain setting differentiating them:

```ts
       if (viewer.terria.cesium) {
-        viewer.terria.cesium.scene.mode =
-          viewerMode === "3d" ? SceneMode.SCENE3D : SceneMode.COLUMBUS_VIEW;
+        viewer.terria.cesium.scene.mode = SceneMode.SCENE3D;
       }
```

Please correct me if I am wrong.

lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -266,7 +275,10 @@ const SettingPanel: FC = observer(() => {
           <Text as="label">{t("settingPanel.mapView")}</Text>
         </Box>
         <FlexGrid gap={1} elementsNo={3}>
-          {Object.entries(MapViewers).map(([key, viewerMode]) => (
+          {(availableMapViewers?.length
+            ? availableMapViewers
+            : Object.entries([MapViewers["2d"]])
+          ).map(([key, viewerMode]) => (
```
niktayv: This looks like a bug: Fallback produces incorrect key "0" instead of "2d".

`Object.entries([MapViewers["2d"]])` wraps the value in an array, so `Object.entries()` returns `[["0", MapViewers["2d"]]]` where the key is `"0"` instead of `"2d"`. This causes:

1. `isActive={key === currentViewer}` to never match (since "0" !== "2d")
2. `selectViewer(key as any, event)` to receive invalid key "0"

A possible fix:
```ts
-          {(availableMapViewers?.length
-            ? availableMapViewers
-            : Object.entries([MapViewers["2d"]])
-          ).map(([key, viewerMode]) => (
+          {(availableMapViewers?.length
+            ? availableMapViewers
+            : [["2d", MapViewers["2d"]] as const]
+          ).map(([key, viewerMode]) => (
```

Please check again.

lib/Models/Cesium.ts
```diff
@@ -90,7 +90,9 @@ import TerriaFeature from "./Feature/Feature";
 import GlobeOrMap from "./GlobeOrMap";
 import Terria from "./Terria";
 import UserDrawing from "./UserDrawing";
-import { setViewerMode } from "./ViewerMode";
+import SceneMode from "terriajs-cesium/Source/Scene/SceneMode";
+import WebMercatorProjection from "terriajs-cesium/Source/Core/WebMercatorProjection";
```
niktayv: The new `SceneMode` and `WebMercatorProjection` imports from `terriajs-cesium` are placed after local imports. For consistency and easier maintenance, consider moving them to the existing Cesium import block (lines 15-61).

lib/ViewModels/TerriaViewer.ts
niktayv, line 286: We need to consider error fallback logic to account for Cesium2D consistently.

The submitted change correctly loads the Cesium viewer for both Cesium and Cesium2D modes. However, the error handling fallback in lines 282-285 only checks for ViewerMode.Cesium:

```ts
this.viewerMode =
  this.viewerMode === ViewerMode.Cesium
    ? ViewerMode.Leaflet
    : undefined;
```

If Cesium2D initialisation fails, viewerMode will be set to undefined instead of falling back to Leaflet, potentially leaving users without a working viewer.

A possible fix:
```ts
 this.viewerMode =
-  this.viewerMode === ViewerMode.Cesium
+  this.viewerMode === ViewerMode.Cesium ||
+  this.viewerMode === ViewerMode.Cesium2D
     ? ViewerMode.Leaflet
     : undefined;
```

CHANGES.md
```diff
@@ -5,6 +5,8 @@
 - Add `token` to `ArcGisMapServerCatalogItem`, `ArcGisMapServerCatalogGroup`, `ArcGisFeatureServerCatalogItem`, `ArcGisFeatureServerCatalogGroup`, `ArcGisImageServerCatalogItem`, `I3SCatalogItem` and `ArcGisCatalogGroup` - if defined, it will be added to the `token` parameter for all ArcGIS Rest API requests.
   - Added `tokenUrl` to `ArcGisImageServerCatalogItem`, and tweaked behaviour in `ArcGisMapServerCatalogItem` and `ArcGisImageServerCatalogItem` so that if both `token` and `tokenUrl` are defined, then `tokenUrl` will be used. This allows the token to be refreshed if needed.
 - WMTS read URL from operations metadata #7371
+- Allow to use Cesium in `SCENE2D` mode (and WebMercator projection) as map viewer.
```
niktayv: “Allow using…”

CHANGES.md
```diff
@@ -5,6 +5,8 @@
...
+- Add a new configuration parameter `mapViewers` to specify which map viewers to show in `SettingPanel`.
```
niktayv: “Add a new mapViewers config parameter…”.
Also, it would be good to clarify that mapViewers controls the _visible viewers_ in SettingsPanel. This improves clarity without changing meaning.

niktayv (changes requested):
I added some more comments related to potential bugs, which need checking, but they may be nitpicks.

Also, I carefully tested if the items in the demo catalogue were loading as expected.
Below is the summary of my testing:

- GeoJSON Test: NO
- CSV/CSV Test: NO
- CSV/Pacific Earthquakes (CSV): NO
- CSV/Region mapping (CSV)/2011 Census AUST (STE): YES
- KML/Simple Test (KML): NO
- CZML Files/Simple Test (CZML): YES
- GeoJSON/Test styled and unstyled features: YES
- GeoJSON/Test overriding styled and unstyled features: YES
- GeoJSON/Testing stroke styles applied to embedded unstyled point.: NO
- Small glTF 3D Models: both YES
- Esri ArcGIS MapServer URL: YES
- WMS Service: YES
- WMTS Service: YES

Presumably, the noticed issues are unrelated to the present PR and need to be raised separately, but it would be good to have a second eye on them.

glughi:
@niktayv, what you've noticed happens in some cases where the data consists of just one point. The rectangle passed to the Cesium `Camera.flyTo` method is a degenerate rectangle (because it is a point, so W=0 and H=0), and the `flyTo` method takes the camera too low. I think the ticket #4250 is about the same.

CHANGES.md
```diff
@@ -5,6 +5,8 @@
 - Add `token` to `ArcGisMapServerCatalogItem`, `ArcGisMapServerCatalogGroup`, `ArcGisFeatureServerCatalogItem`, `ArcGisFeatureServerCatalogGroup`, `ArcGisImageServerCatalogItem`, `I3SCatalogItem` and `ArcGisCatalogGroup` - if defined, it will be added to the `token` parameter for all ArcGIS Rest API requests.
   - Added `tokenUrl` to `ArcGisImageServerCatalogItem`, and tweaked behaviour in `ArcGisMapServerCatalogItem` and `ArcGisImageServerCatalogItem` so that if both `token` and `tokenUrl` are defined, then `tokenUrl` will be used. This allows the token to be refreshed if needed.
 - WMTS read URL from operations metadata #7371
+- Allow using Cesium in `SCENE2D` mode (and WebMercator projection) as map viewer.
+- Add a new configuration parameter `mapViewers` to specify which mapViewers will be visible and usable in in `SettingPanel`.
```
niktayv: There’s a small typo (“in in”).

lib/Models/Cesium.ts
```diff
@@ -1034,6 +1041,15 @@ export default class Cesium extends GlobeOrMap {
     const scene = this.scene;
     const camera = scene.camera;
 
+    if (scene.mode === SceneMode.SCENE2D) {
+      const rect = camera.computeViewRectangle();
+      if (rect) {
+        return new CameraView(rect);
+      }
+      // Fallback for 2D mode when computeViewRectangle returns undefined
+      return new CameraView(this.terriaViewer.homeCamera.rectangle);
```
niktayv: The 2D branch returns `new CameraView(rect)` with only a rectangle, while the 3D branches return CameraView with all four parameters (rectangle, position, direction, up).

If downstream code relies on position, direction, or up properties being present (e.g., when restoring a view or switching modes), this can lead to unexpected behaviour.

Should the 2D CameraView also include camera orientation data (some defaults) for consistency?

Maybe we should document that these properties are undefined in 2D mode.

lib/ReactViews/Map/Panels/SettingPanel.tsx
```diff
@@ -247,6 +250,15 @@ const SettingPanel: FC = observer(() => {
...
 
+  const tmpMapViewers = Object.entries(MapViewers).filter(
+    ([key, viewerMode]) =>
+      viewerMode.available &&
+      terria.configParameters.mapViewers.includes(key as MapViewersKey)
+  );
+  const availableMapViewers = tmpMapViewers?.length
+    ? tmpMapViewers
+    : [["2d", MapViewers["2d"]] as const];
+
```
niktayv: If `terria.configParameters.mapViewers` is `undefined`, calling `.includes()` on it will throw a runtime error. Additionally, `tmpMapViewers?.length` uses optional chaining unnecessarily since `filter()` always returns an array.

We can also make the naming a little more descriptive.

Here is a possible improvement.

```ts
-  const tmpMapViewers = Object.entries(MapViewers).filter(
-    ([key, viewerMode]) =>
-      viewerMode.available &&
-      terria.configParameters.mapViewers.includes(key as MapViewersKey)
-  );
-  const availableMapViewers = tmpMapViewers?.length
-    ? tmpMapViewers
-    : [["2d", MapViewers["2d"]] as const];
+  const configuredViewers = terria.configParameters.mapViewers ?? [];
+  const filteredMapViewers = Object.entries(MapViewers).filter(
+    ([key, viewerMode]) =>
+      viewerMode.available &&
+      configuredViewers.includes(key as MapViewersKey)
+  );
+  const availableMapViewers = filteredMapViewers.length
+    ? filteredMapViewers
+    : [["2d", MapViewers["2d"]] as const];
```

lib/Models/ViewerMode.ts
niktayv: The `@param` correctly lists `2dcesium`, but the explanation on line 78 still references only `Leaflet|Cesium|NoViewer` without mentioning `Cesium2D`.

Update the comment to include Cesium2D:
```ts
-  // ViewerMode can either mean Leaflet|Cesium|NoViewer or 3d|3dsmooth|2d
+  // ViewerMode can either mean Leaflet|Cesium|Cesium2D|NoViewer or 3d|3dsmooth|2d|2dcesium
```

lib/ReactViews/Map/Panels/SettingPanel.tsx
niktayv, line 283: Can we avoid `as any` cast?

The `key as any` cast at line 285 bypasses TypeScript's type checking. Since the key comes from the filtered MapViewers entries, it should be typed properly:

```ts
-          {availableMapViewers.map(([key, viewerMode]) => (
+          {availableMapViewers.map(([key, viewerMode]: [MapViewersKey, typeof MapViewers[MapViewersKey]]) => (
             <SettingsButton
               key={key}
               isActive={key === currentViewer}
-              onClick={(event: any) => selectViewer(key as any, event)}
+              onClick={(event: MouseEvent<HTMLButtonElement>) => selectViewer(key, event)}
             >
```

niktayv: This is a nitpick, but we may consider it as well.

The existing code mutates `MapViewers["3d"].available` on every render, which is a side effect in the render phase. This is a React anti-pattern that can cause issues with concurrent rendering. Since the _new filtering logic_ at lines 253-257 depends on `viewerMode.available`, this mutation directly affects the available viewers shown.

Can we move this initialisation to a one-time setup (e.g., in Terria initialisation) or computing availability inline without mutation?

For instance, we can do:
```ts
-  if (
-    terria.configParameters.useCesiumIonTerrain ||
-    terria.configParameters.cesiumTerrainUrl
-  ) {
-    MapViewers["3d"].available = true;
-  }
-
   const supportsSide = isCesiumWithTerrain;
   // ...
-  const tmpMapViewers = Object.entries(MapViewers).filter(
-    ([key, viewerMode]) =>
-      viewerMode.available &&
-      terria.configParameters.mapViewers.includes(key as MapViewersKey)
+  const is3dAvailable =
+    terria.configParameters.useCesiumIonTerrain ||
+    terria.configParameters.cesiumTerrainUrl;
+
+  const configuredViewers = terria.configParameters.mapViewers ?? [];
+  const filteredMapViewers = Object.entries(MapViewers).filter(
+    ([key, viewerMode]) => {
+      const isAvailable = key === "3d" ? is3dAvailable : viewerMode.available;
+      return isAvailable && configuredViewers.includes(key as MapViewersKey);
+    }
   );
```

niktayv (dismissed):
Grazie, Giovanni! Thank you for all these improvements. I've just pointed out a few minor quirks that may help the other reviewers make their judgment.

lib/ReactViews/Map/Panels/SettingPanel.tsx
glughi: `terria.configParameters.mapViewers` can't be undefined.
The other suggestions look good.

lib/ReactViews/Map/Panels/SettingPanel.tsx
glughi: Couldn't we just remove the available set (`MapViewers["3d"].available = true;`) since there's no other code to change it?

niktayv: Giovanni @glughi, I made this nitpick purely theoretically. Perhaps let's leave it for now before we have more evidence that this really hurts the performance.

niktayv:
Giovanni @glughi, thank you for your hard work. I hope my reviews helped improve your PR, but I think you should wait until the core maintainers confirm whether this PR will be merged. I heard that some time ago they made a specific decision not to use the Cesium 2D view mode, but I don't know the reasons.

Nanda @na9da, what would you suggest?

glughi:
Any news?

na9da:
Hi @glughi, thanks for your work. One reason we have been hesitant to add Cesium2D is that it is less well maintained than 3D mode. There are quiet a few [reported bugs](https://github.com/CesiumGS/cesium/issues?q=state%3Aopen%20label%3A%22category%20-%202d%20%2F%20columbus%20view%22). For instance, as @niktayv pointed out if you add the GeoJSON test item from the default catalog, the points are not visible in Cesium2D mode. This appears to have to something to do with the [heightReference setting](https://github.com/TerriaJS/terriajs/blob/e539e129dfcc9fc764b101bb4bc8fcf2c050f6f5/lib/Table/createLongitudeLatitudeFeaturePerRow.ts#L53) and is fixed if we remove it.

I also noticed a performance drop relative to 3D when panning around with the map zoomed out. Not sure if it is a random one off thing or something else.

We could potentially fix the issues already noted in the above comments and merge this in but disable it by default.

glughi:
Hi @na9da 
I think adding the feature, but disabled, is a good compromise, but obviously, do as you prefer.
I just need to know.

Issue #7237, opened 2024-07-18
Fix legend URL for difference item generated using difference tool


na9da:
Reported in datacube slack.

When diffing two WMS dates, in the difference mode the correct legend for difference output is shown. But when exiting the difference mode, the difference output in the workbench shows the wrong legend.

To generate the correct legend for diff mode, the `time` parameter must be set to `date1,date2`. Example:  https://ows.dev.dea.ga.gov.au/wms?service=WMS&version=1.1.0&request=GetLegendGraphic&format=image%2Fpng&transparent=True&layer=ga_s2ls_intertidal_cyear_3&styles=intertidal_elevation_adaptive&time=2016-01-01%2C2022-01-01

Test dataset: https://terria-cube.terria.io/#share=s-weAnFnIaphPxgUQRnDa1FftkPom

correct legend:
![image](https://github.com/user-attachments/assets/3fc119fa-fd32-485b-bf1c-f0a72bc36aad)

incorrect legend:
![image](https://github.com/user-attachments/assets/8a8d0e40-9b40-403f-81ed-cd304a7120ed)

PR #7243, opened 2024-08-06
Fix legend and feature info for wms difference output.

na9da:
### What this PR does

Makes following changes to WMS difference output:
 - Fixes legend URL for WMS difference outputs - #7237
 - Allow overriding properties of difference output item - https://github.com/TerriaJS/terriajs/issues/7238

### Test me

1. Open [this test link](http://ci.terria.io/wms-diff-fixes/#share=s-2Xp6mjzhyYokbChYJNIbRHM0RsY&https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/dev/terria/terria-cube-v8.json&clean) which has a difference output
2. Ensure the legend shown is for the difference output
image

3. Ensure the feature info for the difference item shows the raw data by default.

### Checklist

### DIFF
--- a/lib/ModelMixins/DiffableMixin.ts
+++ b/lib/ModelMixins/DiffableMixin.ts
@@ -61,7 +61,8 @@ export class DiffStratum extends LoadableStratum(DiffableTraits) {
     // disable export if showing diff
     // currently there is no way to generate export for the difference layer as
     // it requires 2 time parameters which is not supported in standard WCS
-    return this.catalogItem.isShowingDiff ? true : false;
+    const disableExport = this.catalogItem.isShowingDiff;
+    return disableExport;
   }
 
   @computed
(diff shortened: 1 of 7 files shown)


lib/ModelMixins/DiffableMixin.ts
```diff
@@ -56,6 +55,24 @@ class DiffStratum extends LoadableStratum(DiffableTraits) {
   get disableDateTimeSelector() {
     return this.catalogItem.isShowingDiff;
   }
+
+  @computed
+  get disableExport() {
+    // disable export if showing diff
+    // currently there is no way to generate export for the difference layer as
+    // it requires 2 time parameters which is not supported in standard WCS
+    const disableExport = this.catalogItem.isShowingDiff;
```
pjonsson: Is the intermediate variable required for some JavaScript reason, or could the method directly  `return this.catalogItem.isShowingDiff` just as well?

na9da: Fixed. I previously wrote it like `this.catalogItem.isShowingDiff ? true : false` because it reads better, but the linter doesn't like it.

nf-s (approved):
Looks good!

PR #7249, opened 2024-08-13
Fix Deprecation warnings caused by declarations after nested blocks.

ljowen:
See: https://sass-lang.com/documentation/breaking-changes/mixed-decls/

### What this PR does

Reorder declarations to appear above any nested blocks

### Test me

Sass compiles without deprecation warnings

### Checklist

### DIFF
--- a/lib/ReactViews/Workbench/Controls/legend.scss
+++ b/lib/ReactViews/Workbench/Controls/legend.scss
@@ -3,8 +3,7 @@
 .legend {
   composes: clearfix from "../../../Sass/common/_base.scss";
   composes: list-reset from "../../../Sass/common/_base.scss";
-  font-family: $font-mono;
-  font-size: $font-size-small;
+  font-family: $font-mono;  
   // Small font size prevents the font from dictating the table row height.
   font-size: 5px;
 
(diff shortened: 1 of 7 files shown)


lib/ReactViews/Workbench/Controls/legend.scss
```diff
@@ -5,14 +5,13 @@
   composes: list-reset from "../../../Sass/common/_base.scss";
   font-family: $font-mono;
   font-size: $font-size-small;
+  // Small font size prevents the font from dictating the table row height.
+  font-size: 5px;
```
nf-s: How do these two font-size properties interact?

nf-s (approved):
Looks good!

One comment, but its about legacy code

lib/ReactViews/Workbench/Controls/legend.scss
ljowen: The second overrides the first/
Screenshot 2024-08-15 at 3 46 27 pm

The actual font-size of legends gets set by legend-title below, I can't find a way to get text with this rule applied. Given 5px is unreadable my vote would be to either a) remove both, b) keep `$font-size-small`:

Here's what they look like currently with element I'ved hacked in, vs with font-size-small
Screenshot 2024-08-15 at 3 57 08 pm

Screenshot 2024-08-15 at 3 57 51 pm

PR #7268, opened 2024-09-09
Fix `Workbench.collapseAll()` and `Workbench.expandAll()` for References

ykiu:
### What this PR does

Fixes a bug where `Workbench.collapseAll()` and `Workbench.expandAll()` does not collapse/expand References in the workbench

### Test me

Build with TerriaMap and load [this init file](https://gist.github.com/ykiu/fe135b589899970752323af9b7963e66) like so:
http://localhost:3001/#clean&https%3A%2F%2Fgist.githubusercontent.com%2Fykiu%2Ffe135b589899970752323af9b7963e66%2Fraw%2Fbc22d2782df8d77e0fcaff23ad2a3e5de46865d3%2Fterria-reference.json

Then, add "foo" from the catalog to the map.

Before this PR the "Collapse all" button does not actually collapse the item in the workbench.
After this PR the button actually collapses the item.

### Checklist

  - Not applicable.

### DIFF
--- a/test/Models/WorkbenchSpec.ts
+++ b/test/Models/WorkbenchSpec.ts
@@ -6,11 +6,16 @@ import WebMapServiceCatalogItem from "../../lib/Models/Catalog/Ows/WebMapService
 import Workbench from "../../lib/Models/Workbench";
 import Result from "../../lib/Core/Result";
 import TerriaError, { TerriaErrorSeverity } from "../../lib/Core/TerriaError";
+import TerriaReference from "../../lib/Models/Catalog/CatalogReferences/TerriaReference";
 
 describe("Workbench", function () {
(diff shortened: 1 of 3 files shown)


test/Models/WorkbenchSpec.ts
na9da, line 1: @ykiu - could you please remove the import for `BaseModel` as it is not needed any more and adds to the lint errors.

na9da (changes requested):
@ykiu - This looks great, thanks for fixing the buggy behaviour.

I have made just one comment.

ykiu:
@na9da Thank you for pointing that out. I fixed it just now.

na9da (approved):
Merging. Thanks @ykiu.

PR #7270, opened 2024-09-09
Fix sharing when using initfile

### LINKED ISSUE #7269: Share using initfile
If you generate a share URL from a map that uses parameters in the URL, for example, an `initfile`, those parameters are lost. It should be kept, at least in the case of an `initfile`, otherwise, some data could become unreachable.

glughi:
### What this PR does

Fixes #7269 

### Checklist


na9da:
Hi @glughi,

thanks for the PR. 

We are keen to know how you use the init file from URL feature. 

Internally, we were thinking of introducing some controls around how this feature works, because it is not ideal for all maps to allow others to modify the catalog. So maybe we could restrict it by default or disallow init files from URLs from modifying existing catalog items.

Any insights on your specific use case for this feature could help us decide.

Is there a reason you can't have the init URL be part of your [config.json](https://github.com/TerriaJS/TerriaMap/blob/082810cd06dc7528d801d0726798d80575737cd0/wwwroot/config.json#L3)?

Cheers

glughi:
Hi @na9da, we make great use of TerriaJS ability to use different catalogs in addition to the default ones (added through the _config.js_ file), defining them via URL. This way, we don't have just one gigantic catalog, which could be inconvenient for users, but in addition to the standard catalogs, we have some thematic catalogs. For example, the Emilia-Romagna Region has its standard catalog, but it has added one specific for the 2023 flood emergency. Some catalogs are used only by regional technicians and are not available to other users. Working in this way, it becomes necessary to have the ability to share the map without losing a catalog name added by the URL.

na9da:
Thanks for sharing your use case @glughi 

We do have a type of catalog item called [terria-reference](https://docs.terria.io/guide/connecting-to-data/catalog-type-details/terria-reference/) which lets you pull items from another catalog. It doesn't let you modify the catalog dynamically, but allows referencing external catalogs without adding init file to the URL. Not sure if it exactly suits your needs though.

glughi:
Hi, we know _terria-reference_ but it is not what we need in this case. Thanks to URL catalog management, we can put links in web pages all using the same instance of Terria but load a catalog rather than another depending on the page's topic and target.
Little examples:
https://mappe.regione.emilia-romagna.it/#emg202305
https://mappe.regione.emilia-romagna.it/#emg202305rilievi
https://mappe.regione.emilia-romagna.it/#emg202409

glughi:
Hello, if we resolve conflicts do you plan to adopt this PR?

na9da:
Hi @glughi,

Yes we can merge it in, thanks!

glughi:
I close this PR and reopen it as #7727

PR #7272, opened 2024-09-17
Lint fixes

ljowen:
### What this PR does

Fixes 215 remaining lint warnings, mostly unused params, functions and constants.

- Add eslint-disable comments for: 
   - no-unsafe-declaration-merging https://typescript-eslint.io/rules/no-unsafe-declaration-merging/ due to clash with TS2611 type errors (see `UrlTraits.ts` for example)
   - react-hooks/exhaustive-deps as these will most likely have side effects so will need more testing / work to enable / workout which uses are valid

### Test me

- Verify tests pass and app working as expected
- `yarn gulp lint` should return 0 warnings

### Checklist


lib/Models/BoxDrawing.ts
```diff
@@ -207,11 +207,6 @@ const FACE_POINT_VECTORS = [
   new Cartesian3(0.0, 0.0, 0.5)
 ];
 
-// The box has 8 corner points and 6 face points that act as scaling grips.
-// Here we represent them as 7 vectors in local coordinates space.
-// Each vector represents a point and its opposite points can be easily derived from it.
-const SCALE_POINT_VECTORS = [...CORNER_POINT_VECTORS, ...FACE_POINT_VECTORS];
```
ljowen, line 213: unused

lib/Models/BoxDrawing.ts
```diff
@@ -1779,23 +1774,6 @@ export function screenToGlobePosition(
...
-  const cartographic = Cartographic.fromCartesian(
-    position,
-    undefined,
-    scratchCartographic
-  );
-  cartographic.height = 0;
-  return Cartographic.toCartesian(cartographic, undefined, result);
-}
-const scratchCartographic = new Cartographic();
-
```
ljowen, line 1798: unused

lib/ReactViews/Map/MapNavigation/Items/Compass/Compass.tsx
```diff
@@ -420,9 +422,12 @@ class Compass extends React.PureComponent<PropTypes, IStateTypes> {
...
               `}
             >
+              <pre
+                style={{ background: "red", width: "200px", height: "200px" }}
+              >
+                !!!!!!!!
+              </pre>
               <GyroscopeGuidance
-                //@ts-expect-error
-                rightOffset="72px"
```
ljowen, line 425: `GyroscopeGuidance` doesn't have this prop

lib/ReactViews/SelectableDimensions/Color.tsx
```diff
@@ -23,7 +23,7 @@ const debounceSetColorDimensionValue = debounce(
 export const SelectableDimensionColor: React.FC<{
   id: string;
   dim: SelectableDimensionColorModel;
-}> = observer(({ id, dim }) => {
+}> = observer(({ dim }) => {
```
ljowen, line 26: I was unsure of side effects here, whether removing `id` would make the observer lose sensitivity to changes to `id` but that does not seem to be the case see: https://codesandbox.io/p/sandbox/bold-heyrovsky-c2kntd?file=%2Fsrc%2Findex.tsx%3A57%2C53

lib/Core/animation.ts
```diff
@@ -33,10 +33,10 @@ const transitionEnd = (element: Element | null) =>
   });
 
 const animationTimeout = (
-  timeoutID: ReturnType<typeof setTimeout> | undefined
+  _timeoutID: ReturnType<typeof setTimeout> | undefined
```
ljowen, line 36: There's a lot of unused params across the codebase, we could instead just turn off the rule for unused params, open to suggestions (I prefer having to be explicit)

.eslintrc
```diff
@@ -115,18 +115,21 @@
...
+            "destructuredArrayIgnorePattern": "^_",
+            "ignoreRestSiblings": true
           }
         ],
         "@typescript-eslint/ban-ts-comment": "warn",
         "@typescript-eslint/no-loss-of-precision": "warn",
-        "@typescript-eslint/no-unsafe-declaration-merging": "warn",
-        "react-hooks/exhaustive-deps": "warn",
+        "@typescript-eslint/no-unsafe-declaration-merging": "off",
+        "react-hooks/exhaustive-deps": "off",
```
zoran995: I would vote for leaving this turned on, don't how much of them is there but definitely better would be to suppress them using `eslint-disable` comments. If I recall correctly there isn't much of hook usages.
p.s. Yeah it is bad practice to suppress the warning or error but would always choose that instead of disabling the rule for all future code.

lib/Core/animation.ts
pjonsson: As someone who isn't intimately familiar with the code base, it's a one time cost to prepend a "_" to (at most) 215 places, and that prevents future unused leftovers that will confuse me.

.eslintrc
```diff
@@ -87,7 +87,7 @@
     /* Variables */
     "no-label-var": 1,
     "no-unused-vars": [
-      1,
+      "warn",
```
pjonsson, line 90: There's a `--max-warnings` in gulpfile.js for the lint target that you also want to configure.

test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
```diff
@@ -73,12 +73,12 @@ describe("GeoJsonCatalogItemSpec", () => {
                 type: "Polygon",
                 coordinates: [
                   [
-                    [144.80667114257812, -32.96258644191746],
+                    [144.806671142578125, -32.96258644191746],
```
pjonsson: You've already done this now, but I fixed this problem in #7149 back in May (with an identical fix), so I think you could have saved yourselves a bit of effort by first merging the fixes you had laying around already.

.eslintrc
ljowen: My thinking was just having a smaller diff (and I'd like to resolve the `exhaustive-deps` eventually) but you're right better to use disable comments to prevent further instances. Will update

.eslintrc
ljowen: Thanks! Forgot that

test/Models/Catalog/CatalogItems/GeoJsonCatalogItemSpec.ts
ljowen: Thanks Peter, I should have done a scan for existing lint PRs first. I've merged your changes in :)

lib/Table/TableColumn.ts
```diff
@@ -920,6 +920,6 @@ function toNumber(value: string): number | null {
   return null;
 }
 
-function nullFunction(rowIndex: number) {
+function nullFunction() {
```
pjonsson: This function is returned from `scaledValueFunctionForType`, doesn't callers of that function expect the returned function to accept one parameter?

pjonsson:
If you are focusing on lint-related things in the near future, it might be worth considering #7150 after this PR has been merged (I'll fix the conflicts after this is merged).

lib/Table/TableColumn.ts
ljowen: They do but it doesn't look like typescript minds: [example](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAZgJhgXhgCgJQC4ZgK4A2BKAfDAN4BQMNMATgKZR51i6EGUC+lloksOAEYU6bOjQBDOgHMADDnwBbAEYM6GUuyKbkZKrXpMWbRN17DMaBBkpA) . Happy to change it to `(_rowIndex: number)` if you think that's clearer

pjonsson: I had a second look, and wrote something that roughly equals "the type signature states the returned function should take a number and return a number, so I suspect removing the parameter will cause problems in some future upgrade of TypeScript".

As a side note, the return type of `nullFunction` is `null` rather than `number`, so the type is already wrong in the current code base. The function body is just `return null`, so TypeScript must be able to infer that return type as well.

I feel this behavior of a type checker is really strange, so I had a third look. The function `getScaledValueFunctionForType` was added in #3594 and had a caller in `TableMixin.ts`, but that caller no longer seems to be present. Could `getScaledValueFunctionForType` (and its helper functions) be removed instead?

lib/Models/Catalog/CatalogReferences/ThreddsItemReference.ts
```diff
@@ -121,7 +121,7 @@ export default class ThreddsItemReference extends UrlMixin(
     );
     if (model === undefined) return;
     this.setThreddsStrata(model);
-    previousTarget = model;
+    _previousTarget = model;
```
pjonsson, line 124: What do these assignments do, is `previousTarget` some kind of return-value type so the caller gets a hold of a reference to the model?

lib/Models/Definition/LoadableStratum.ts
```diff
@@ -24,7 +24,7 @@ export default function LoadableStratum<
         get: function () {
           return undefined;
         },
-        set: function (value: any) {
+        set: function () {
```
pjonsson: Same comment for this as the nullFunction below: I suspect removing the parameter will cause problems in some future TypeScript upgrade (and a setter without parameter is strange).

lib/Models/Leaflet.ts
```diff
@@ -63,7 +63,7 @@ import Terria from "./Terria";
 // We want TS to look at the type declared in lib/ThirdParty/terriajs-cesium-extra/index.d.ts
 // and import doesn't allows us to do that, so instead we use require + type casting to ensure
 // we still maintain the type checking, without TS screaming with errors
-const FeatureDetection: FeatureDetection =
+const _FeatureDetection: FeatureDetection =
```
pjonsson: This looks like something special that isn't just an unused parameter, can `@ts-expect-error` be used here so this variable doesn't get lost among the noise of unused parameters?

test/Models/Catalog/Ows/WebProcessingServiceCatalogFunctionSpec.ts
```diff
@@ -148,7 +148,7 @@ describe("WebProcessingServiceCatalogFunction", function () {
       let dispose: any;
       job = (await wps.submitJob()) as WebProcessingServiceCatalogFunctionJob;
 
-      await new Promise<void>((resolve, reject) => {
+      await new Promise<void>((resolve, _reject) => {
```
pjonsson: The change in YDYRCatalogFunctionSpec above just removes the parameter. If this isn't somehow different from that, I think it's better to either choose to always keep the unused reject with an underscore or always remove it.

test/Models/Experiment.ts
```diff
@@ -49,7 +49,7 @@ describe("NewStuff", function () {
       );
     });
 
-    autorun((dispose) => {
+    autorun((_dispose) => {
```
pjonsson: Did you intend to remove the console.log in the next line too?

lib/Models/Leaflet.ts
pjonsson: Scratch my previous remark, the comment in the source code is 4+ years old. In #6944, the FeatureDetection variable is removed from CorsProxy.ts since the FeatureDetection is not used in that file any longer, so this line corresponds to an unused import as far as I understand. I'm guessing the variable and the comment can be removed?

lib/Table/TableColumn.ts
ljowen: Nice find thanks I'll remove it

lib/Models/Catalog/CatalogReferences/ThreddsItemReference.ts
ljowen: That's my understanding, it's updating the reference. Although looking where it's call in `ReferenceMixin` that reference doesn't appear to be further used. @na9da do you have insight into this?

lib/Models/Definition/LoadableStratum.ts
ljowen: Good point, will update

test/Models/Catalog/Ows/WebProcessingServiceCatalogFunctionSpec.ts
ljowen: I'll remove the unused rejects since it's not communicating anything extra

test/Models/Experiment.ts
ljowen: This whole file isn't actually asserting anything and hasn't been touched since 2021, I'll check with the team about removing it completely.

ljowen: Deleted

pjonsson (approved):
I had a look again and while I don't speak TypeScript fluently, from what I can see this looks like a solid improvement.

Not having 200+ lint warnings to sift through for every commit will make development easier going forward, so I definitely want this (or some alternative) merged as soon as possible.

lib/Models/Catalog/CatalogReferences/ThreddsItemReference.ts
na9da: I think that line has no effect and can be removed. There's [couple more](https://github.com/TerriaJS/terriajs/blob/a21ecf7a75a66d328c5340fc24522f7c56d17c2d/lib/Models/Catalog/Ckan/CkanItemReference.ts#L391) [of these](https://github.com/TerriaJS/terriajs/blob/a21ecf7a75a66d328c5340fc24522f7c56d17c2d/lib/Models/Catalog/Esri/ArcGisPortalItemReference.ts#L353).

na9da (approved):
@ljowen looks good to me. Just the one comment on removing the `_previousTarget = model` lines. I also concur with @pjonsson and @zoran995 on not turning off the lint rules so that future code is checked. Please update the PR description to reflect that change.

PR #7283, opened 2024-09-28
fix: augment cesium types and use import instead of require in ts files

zoran995:
### What this PR does

Typescript enums can be augmented by using namespace, but namespace augmentation doesn't behave nice when reexported and it seems that it was main reason why the previous attempt to augment cesium enum and namespace has failed. 

### Test me

I have tested this by running the build of terriajs and terriamap which has passed

### Checklist

- ~~[ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~~ - **type level change**

### DIFF
--- a/architecture/0012-do-not-use-require-for-cesium-in-ts.md
+++ b/architecture/0012-do-not-use-require-for-cesium-in-ts.md
@@ -1,0 +1,34 @@
+# 2. Using `imports` instead of `require` + type casting statements
+
+Date: 2024-09-28
+
+## Status
+
+Proposed
+
+## Context
+
+A decision was made in [0002-require-instead-of-import](./0002-require-instead-of-import.md) to use non-exported terriajs-cesium features using `require` + type casting statements due to inability to augment namespace and enum imports.
+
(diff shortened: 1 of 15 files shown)


lib/Map/Leaflet/LeafletVisualizer.ts
```diff
@@ -1059,6 +1056,7 @@ class LeafletGeomVisualizer {
       this
     );
     this.leafletScene.map.removeLayer(this._featureGroup);
+    console.log("destroy");
```
ljowen:
```suggestion
```

ljowen (approved):
Nice discovery @zoran995! Thanks for contributing this

architecture/0012-do-not-use-require-for-cesium-in-ts.md
```diff
@@ -0,0 +1,34 @@
+# 2. Using `imports` instead of `require` + type casting statements
+
+Date: 2024-09-28
+
+## Status
+
+Proposed
```
zoran995, line 7: Just a note we should update this to Accepted

PR #7294, opened 2024-10-22
Add Leaflet maxZoom to configParameters

glughi:
### What this PR does

Related to #7280 

Add "leafletMaxZoom" to configParameters so that the maxZoom of the Leaflet viewer can be changed.

### Test me

### Checklist


lib/Models/Terria.ts
```diff
@@ -328,6 +328,9 @@ export interface ConfigParameters {
    */
   feedbackMinLength?: number;
 
+  /** If undefined, then Leaflet's default attribution will be used */
```
na9da: The comment needs to be updated.

lib/Models/Leaflet.ts
```diff
@@ -103,6 +103,7 @@ export default class Leaflet extends GlobeOrMap {
     clippingRectangle: Rectangle | undefined
   ) => GridLayer = computedFn((ip, clippingRectangle) => {
     const layerOptions = {
+      maxZoom: this.terria.configParameters.leafletMaxZoom,
```
na9da, line 106: I think, because this is a map level config, it is better to pass it to the `Map` instance created here:

https://github.com/TerriaJS/terriajs/blob/676d1d881ad32163f7a136fb80e213285f0da0be/lib/Models/Leaflet.ts#L149-L155

na9da (changes requested):
Thanks for the change @glughi. 

I have made a couple of comments.

lib/Models/Leaflet.ts
glughi: Hi, as you can read in the documentation of Leaflet (https://leafletjs.com/reference.html#map-maxzoom) the `MapOptions.maxZoom` property has no default limitations. The `GridLayerOptions.maxZoom` property is set to 18 by default, so it is the one to change so that zoom greater than 18 can be used.

na9da: @glughi  that makes sense.

But if it is a per layer setting, then it is possible that they are different for different layers, and so it should probably go as an imagery layer trait similar to [leafletUpdateInterval here](https://github.com/TerriaJS/terriajs/blob/c64dd05430f7b1b1a22f687f47226301afb7eb91/lib/Traits/TraitsClasses/ImageryProviderTraits.ts#L53) and then configured individually for each item in the catalog json.

What do you think?

zoran995: Maybe the best option is to have it defined in both places. If I got it correctly, the leaflet will always respect what is defined on the map if the layer `maxZoom` is smaller it won't try to load further images even if the user can zoom in further, is that correct? i.e. `zoomLevel` on the map is set to 20 and layers are set to 18

glughi: @zoran995 @na9da if `MapOptions.maxZoom` is set, the map will not allow zooming beyond that level, regardless of the value defined for layers with `GridLayerOptions.maxZoom`.
If `MapOptions.maxZoom` is greater than `GridLayerOptions.maxZoom`, the user will at some point see the layer disappear (the map will probably turn gray unless there are other visible layers).
Maybe we should set the same value for the Map and all layers.

zoran995: @glughi just looked more in detail at how the leaflet works with this, and I would second @na9da suggestion for having this set on map level and not on per individual layer basis, as those options are mutually exclusive.
Setting the `maxZoom` on the leaflet map is enough to have a zoom level greater than 18. **Only** if the map option is not set it will look at layers level `maxZoom` to determine the `maxZoom` to allow, and each time a new layer is loaded it will recompute the `maxZoom`.
For having this configurable per item basis, as `ImageryProviderTrait`, IMHO there is no need for additional options regarding this on the per layer basis. There are already `maximumLevel` and `minimumLevel` which are mapped to `maxNativeZoom` and `minNativeZoom` for leaflet tile layers which will show less detail layer at higher zoom levels. 

Sorry if I caused confusion with a previous suggestion.

glughi: Hi @zoran995, this is an example of _Leaflet_ with only `MapOptions.maxZoom` set to 23 and no` maxZoom` for layer:
https://jsfiddle.net/0akvu42r/1/
As you can see, after level 18 the map became gray.

Another example in which I set `maxZoom` to 23 also in layer options:
https://jsfiddle.net/0akvu42r/2/

na9da: Thanks @zoran995  for adding your thoughts. 

I think these fiddles make it a bit more clear. Thanks for setting them up @glughi. 

There are two options being discussed:

a) Setting `Map.maxZoom` - This limits the max zoom of the map as a whole but doesn't affect the layers. So if the layers have a lower maxZoom, they'll appear blank **_(not what we want)_**

b) Setting `Layer.maxZoom` - raises the layer `maxZoom` and by default the map `maxZoom` is set to the largest value from the individual layers. From the fiddles, this appears to be the more correct behaviour.

There could be two ways to do `b`, read the maxZoom from (1) map level config or (2) use an item level trait. 

Currently, `maximumLevel` trait [only sets `maxNativeZoom`](https://github.com/TerriaJS/terriajs/blob/727c3047b7be3948847dcfa2b906d2289841d969/lib/Map/Leaflet/ImageryProviderLeafletTileLayer.ts#L271) which leaflet uses only as a base for auto-scaling to higher zoom levels. Maybe it makes sense for `maximumLevel` to also set `maxZoom`?

If the above solution works do we still need map level config? If we do, it should probably be renamed to something like `defaultImageryLayerMaximumLevel` or something. However i'm not convinced of the need for such a map level setting.

also cc @nf-s for opinions.

zoran995: For some reason, setting only on map level worked in my tests on terria code when set on map-only level, but it doesn't work anymore 😕. 

Yeah, option `a` alone is definitely a no-go.

`maximumLevel` shouldn't be used `maxZoom` as it limits requests sent to the backend but the user should be able to zoom in further to the specific layer. If the map can be zoomed in the further leaflet won't send new requests but show the already loaded images. It wouldn't be desirable to limit the zoom to i.e. level 12 as that's the highest layer support but in Cesium you can zoom in further.

It might be best to go with this as is, not supper critical change and it works as it should, maybe just add a comment on the layer level setting so no one has to wonder in future why we have to set this on the layer level instead of the map level.

na9da: Agreed.

na9da (approved):
Approving as this as using map level `leafletMaxZoom` appears to be the most reasonable option at this point.

PR #7302, opened 2024-10-28
Nominatim search provider

glughi:
### What this PR does

Added search provider for Nominatim of OpenStreetMap

### Checklist


lib/Models/SearchProviders/NominatimSearchProvider.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  protected doSearch(
+    searchText: string,
+    searchResults: SearchProviderResults
+  ): Promise<void> {
+    searchResults.results.length = 0;
+    searchResults.message = undefined;
+
+    if (searchText === undefined || /^\s*$/.test(searchText)) {
```
zoran995: this check is unnecessary as searchText is always defined and is longer than min length. It is validated in SearchProviderMixin

lib/Models/SearchProviders/NominatimSearchProvider.ts
```diff
@@ -0,0 +1,140 @@
...
+      CesiumMath.toDegrees(view.rectangle.west) +
+      ", " +
+      CesiumMath.toDegrees(view.rectangle.north) +
+      ", " +
+      CesiumMath.toDegrees(view.rectangle.east) +
+      ", " +
+      CesiumMath.toDegrees(view.rectangle.south);
+
+    const promise = loadJson(
+      this.url +
```
zoran995: might be better to have this as a Resource if possible, it will be more readable than manually constructed string

lib/Traits/SearchProviders/NominatimSearchProviderTraits.ts
```diff
@@ -0,0 +1,33 @@
...
+    description: "The Nominatim key."
+  })
+  key?: string;
+
+  @primitiveTrait({
+    type: "string",
+    name: "Primary country",
+    description: "Name of the country to prioritize the search results."
+  })
+  countryCodes: string = "it";
```
zoran995: I believe that terria defaults should be Australia

lib/Models/SearchProviders/NominatimSearchProvider.ts
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        searchResults.message = {
+          content: "translate#viewModels.searchErrorOccurred"
+        };
+      });
+  }
+}
+
+function createZoomToFunction(model: NominatimSearchProvider, resource: any) {
```
zoran995: Can you also add the proper type definition to the resource here instead of any? It should be Feature<Point> if I am correct.

zoran995 (changes requested):

zoran995:
also, it would be nice to add unit tests for new features

p.s. Not sure what is team's standpoint whether this should reside in the terria codebase or be a plugin for terriajs, so we don't increase the codebase size and implement/maintain every possible search provider out there. It is the main reason the configurable search providers are implemented, and I didn't have time to test the possibility of using a search provider from a plugin but I don't see why it shouldn't work. @na9da what's your opinion on this

lib/Traits/SearchProviders/NominatimSearchProviderTraits.ts
na9da: better to leave it undefined?

na9da:
@zoran995 - thanks for reviewing it. 

It might be possible to write this as a plugin, but will require importing several private Terria classes which is not ideal. I would accept this PR and split it out later when the use case is better supported by the plugin framework.

zoran995:
@na9da LGTM, but I noticed one thing that might need additional consideration. 
When searching for the place it won't prioritise the results from the currently visible region but will show results from the entire world. As you can see in the image although the map shows Australia the first result is for Germany.
image
In v7 this was resolved by sending double requests to the search provider, first for getting only results in the current view box and second for getting results outside and then those were composed by taking a few results from those.

Do we need to handle this somehow, as multiple requests won't play nicely with the Nominatim TOS? Users already can limit the query to specific countries using `countryCodes`. Maybe we should also expose `bounded` as a trait so the user can configure whether to return only results from the viewable part of the map or all results.

@glughi Besides the above this is in a pretty good state from the code perspective, it just needs
- merge main in,
- update changelog,
- adding details for nominatim search provider to docs https://github.com/TerriaJS/terriajs/blob/c0e6d68cc9d5173e4be86d269457cf7fde016c97/doc/customizing/search-providers.md, this part of docs is still manually written
- potentially unit tests

na9da:
@zoran995 

Thanks for bringing up those points. Here's the [terriajs v7 implementation](https://github.com/TerriaJS/terriajs/blob/a6a2e6464cec91578b924728cdb3e3689b8c2794/lib/ViewModels/NominatimSearchProviderViewModel.js#L100-L118) for reference.

I think the changes in the PR works good enough for now? 

In your example when searching for partial name `Melb`, it shows results in Germany instead of Australia, due to how the API prioritizes the results (preferring most complete match). 

The result is as expected when matching for the full name `Melbourne`. With the map focused on Australia, it shows `Melbourne, Australia` and with the map focused on USA, it shows `Melbourne, Florida, USA`. I think that is sufficiently smart.

As you suggested, the map maintainer may also use the `countryCodes` setting for additional filtering.

@glughi - would you be able to merge in the new `main` branch please?

na9da (approved):
Looks good. Thanks @glughi and @zoran995.

na9da:
@glughi - sorry once again, this will need another pull from `main` to be merged.

Our `main` is a protected branch, so I cannot resolve the conflicts myself without creating a new PR.

PR #7304, opened 2024-10-30
Use useRef instead of createRef in functional component

zoran995:
### What this PR does

`createRef` should only be used in class components while in functional components we should use `useRef`, https://react.dev/reference/react/createRef#migrating-from-a-class-with-createref-to-a-function-with-useref.
This is the change mentioned in https://github.com/TerriaJS/terriajs/pull/7213, 

### Test me

Example link from #7290 can be used for testing this

### Checklist

- ~[ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)~


na9da (approved):

na9da:
Good catch @zoran995.

Might be good to replace this in [MouseTooltip](https://github.com/TerriaJS/terriajs/blob/f102e342e51205aed6808379db62452e564c51b1/lib/ReactViews/Tools/PedestrianMode/MouseTooltip.tsx#L14) component for pedestrian mode as well.

PR #7307, opened 2024-11-07
Add option to disable the search in catalog link in Location Search results

aclel:
### What this PR does

This introduces a new configuration option to disable the "Search In the Data Catalog" link in Location Search results.

image

If the config option is set to `false`, the link is not shown. The option defaults to true to show the link.

### Test me

Set `showSearchInCatalogLink` to `false` in config parameters and check that the link does not appear in search results when you search for something.

### Checklist

- [X] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
- [X] I've updated relevant documentation in `doc/`.
- [X] I've updated CHANGES.md with what I changed.
- [X] I've provided instructions in the PR description on how to test this PR.


lib/Models/Terria.ts
```diff
@@ -290,6 +290,10 @@ export interface ConfigParameters {
    * True to display in-app guides.
    */
   showInAppGuides?: boolean;
+  /**
+   * Show the "Search in Catalog" link in Location Search results
+   */
+  showSearchInCatalogLink?: boolean;
```
na9da: Can this option be added to [SearchBarTraits](https://github.com/TerriaJS/terriajs/blob/main/lib/Traits/SearchProviders/SearchBarTraits.ts) instead? That' where other search bar config exists.

Then it can set in config.json as:
```json
{
  "searchBarConfig": {
     "showSearchInCatalog": false
  }
}
```

test/ReactViews/Search/SearchBoxAndResultsSpec.tsx
```diff
@@ -99,6 +99,27 @@ describe("SearchBoxAndResults", function () {
...
+        testRenderer = create(
+          <ThemeProvider theme={terriaTheme}>
+            <SearchBoxAndResults
+              t={() => {}}
+              terria={terria}
+              viewState={viewState}
+            />
+          </ThemeProvider>
+        );
+      });
```
na9da: Thanks for adding specs for your change. I think `it` blocks cannot be nested. So this needs to be extracted out as a new block.

na9da (changes requested):
Hi @aclel, thanks for the contribution. 

I have requested a couple of changes. Let me know what you think.

aclel:
Hi @na9da, thanks for the review! I've made the changes you suggested.

test/ReactViews/Search/SearchBoxAndResultsSpec.tsx
```diff
@@ -106,4 +106,40 @@ describe("SearchBoxAndResults", function () {
...
+      viewState.searchState.locationSearchResults = [];
+
+      Object.defineProperty(
+        viewState.terria.searchBarModel,
+        "showSearchInCatalog",
+        {
+          value: false,
+          writable: true
+        }
+      );
```
na9da: @aclel - sorry one more tiny request. In Terria to correctly set a trait, you need to use this pattern instead of the `defineProperty()` call:

```js
      viewState.terria.searchBarModel.setTrait(
        CommonStrata.user,
        "showSearchInCatalog",
        false
      );
```

na9da: Oops, you'll also need to merge from latest `main` to fix conflicts.

na9da (approved):

test/ReactViews/Search/SearchBoxAndResultsSpec.tsx
aclel: Thanks @na9da! : ) I was wondering if there was another way to set that prop. I'll read the Trait docs.

na9da: no problem, would you be able to make that change  and sync with the latest `main`? I'm not able to merge unless the branch is in sync with main. Thanks @aclel

na9da:
Merged, thanks @aclel.

Issue #6998, opened 2023-12-04
Upgrade to webpack 5


nf-s:
Given we are most likely not going to swap to ESBuild anytime soon, we should upgrade webpack 5.

This may fix the need to use `--openssl-legacy-provider` - which means we can stop using NodeJs 16 as default version

na9da:
Got a hacky version working in my local. Might have to split it up into a few PRs. Also, it might be worth trying to see if we can simplify webpack config overall.

Here's a rough todo list:
- [ ]  https://github.com/TerriaJS/terriajs/issues/7310
- [ ] Remove [babel transform to commonjs](https://github.com/TerriaJS/terriajs/blob/f102e342e51205aed6808379db62452e564c51b1/buildprocess/configureWebpack.js#L154)
- [ ] Review other babel plugins
- [ ] Replace `file-loader`, `url-loader`, `raw-loader` with inbuilt https://webpack.js.org/guides/asset-modules/
  - [ ] Needs to go from package.json
  - [ ] Needs to go from webpack config
  - [ ] Needs to go from [inline imports](https://github.com/search?q=repo%3ATerriaJS%2Fterriajs%20raw-loader!&type=code). We can instead add webpack config to treat imports of csv, xml etc to use the appropriate loader
- [ ] Webpack 5 [no longer](https://webpack.js.org/migrate/5/#:~:text=Level%205%3A%20Runtime%20Errors.) polyfills nodejs modules and globals like `process.env.NODE_ENV` or `Buffer`. Ideally we avoid using those, or configure polyfills.
- [ ] Replace [remaining](https://github.com/TerriaJS/terriajs/pull/7127) use of outdated string-replace-webpack-plugin with an app level support
- [ ] Upgrade sass-loader and css-loader
  - [ ] Running sass-loader automigration requires replacing all `~terriajs-variables` imports in `scss` with relative paths.
  - [ ] Get rid of webpack config for these aliases from webpack
- [ ] Try rewriting webpack config from scratch to see if it simplifies the config by getting rid of redundant config
- [ ] See if we can get hot reloading to work, otherwise remove it.

pjonsson:
@na9da that's fantastic, great work! Bullets 4/5 sound like they are a part of the webpack 5 configuration, but bullets 1/6 sound like they could be merged with separate PRs before going to webpack 5. Could you please indicate which bullets that require webpack 5, and which bullets that could be merged while still using webpack 4?

na9da:
[...]

PR #7311, opened 2024-11-14
CJS to ESM

na9da:
### What this PR does

Convert CJS modules to ESM.

Required for #6998 
Depends on: https://github.com/terriajs/terriamap/tree/cjs-to-esm

TODO:
- [ ] Test node build

What this PR does:
- Removes stale CJS files that are no longer used (https://github.com/TerriaJS/terriajs/pull/7311/commits/395d3b98bc471d54c562f26cb45baa37782fea6b)
- Convert CJS style `module.exports` to ESM style `export default ...` (https://github.com/TerriaJS/terriajs/pull/7311/commits/7129b240e71f3acaa8fbed7bfcfdcd69b0f4026c)
-  Convert CJS require() to ESM import statements. (https://github.com/TerriaJS/terriajs/pull/7311/commits/30654f1a33b1af79afe5843b99855d67cc3c3667)
   This requires, in some places:
     - simply swapping require with import
     - creating a stub .d.ts module definition in lib/ThirdParty
     - updating the library to get new bundled type definitions
       - `file-saver`
       - `react-responsive`
     - or adding @types/x as dependency
        - `@types/proj4`
       
     Remaining require() imports are all for static assets or web workers which can be fixed when upgrading to webpack 5.
- Remove `use strict` declaration for `.tsx?` files.

### Test me

Test http://ci.terria.io/cjs-to-esm

### Checklist


package.json
```diff
@@ -113,7 +113,7 @@
     "dompurify": "^2.3.3",
     "fetch-mock": "^9.11.0",
     "file-loader": "^3.0.1",
-    "file-saver": "^1.3.8",
+    "file-saver": "^2.0.4",
```
pjonsson: There is a file-saver 2.0.5 released 4 years ago, is there a reason to not bump to that version?

package.json
```diff
@@ -195,6 +195,7 @@
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@types/dateformat": "^3.0.1",
     "@types/node": "^18.15.11",
+    "@types/proj4": "^2.5.5",
```
pjonsson, line 192: My understanding is that the type annotations are required by tsc, so without them `yarn gulp release` will fail. Does `yarn gulp release` work for an external repository, like TerriaMap, if `@types/proj4` is a devDependency here?

package.json
na9da: Done. thanks!

package.json
na9da: That was a stray entry, removed.

lib/Core/ConsoleAnalytics.js
```diff
@@ -0,0 +1,50 @@
+import defined from "terriajs-cesium/Source/Core/defined";
```
pjonsson: Is this file restored because the git auto-merge didn't realize it was renamed when being TSified?

package.json
pjonsson: Sorry, the types/proj4 package is still added in the latest revision of the PR?

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -257,7 +256,9 @@ class FeatureServerStratum extends LoadableStratum(
         return undefined;
       }
 
+      // @ts-expect-error @types/proj4 doesn't define a constructor type
       const source = new proj4.Proj((proj4definitions as any)[wkid]);
+      // @ts-expect-error @types/proj4 doesn't define a constructor type
```
pjonsson: Are you sure about this? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a7b4fad09a7a1309d0fa9a1ec22c57720bf5d02a/types/proj4/index.d.ts#L71 and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a7b4fad09a7a1309d0fa9a1ec22c57720bf5d02a/types/proj4/proj4-tests.ts#L65 looks like it should work?

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -257,7 +256,9 @@ class FeatureServerStratum extends LoadableStratum(
         return undefined;
       }
 
+      // @ts-expect-error @types/proj4 doesn't define a constructor type
```
pjonsson: I can see why the type checker might not be able to infer that the argument is a string for the next line. I just threw together #7344 to try to make it slightly less confused.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
```diff
@@ -168,6 +167,7 @@ class SharePanel extends React.Component<PropTypes, SharePanelState> {
         viewState={this.props.viewState}
         btnTitle={btnTitle}
         isOpen={this.state.isOpen}
+        //@ts-expect-error - not yet ready to tackle tsfying StorySharePanel
```
pjonsson, line 170: Should this be 5 lines earlier (before the `<StorySharePanel`) so it is consistent with the placement for the ts-expect-error for MenuPanel?

lib/ReactViews/Tour/TourExplanationBox.tsx
```diff
@@ -31,7 +31,7 @@ export const TourExplanationBox = styled(Box)`
   h1,
   h2,
   h3 {
-    margin-bottom: ${(p) => p.theme.spacing * 3}px;
+    margin-bottom: ${(p) => (p.theme.spacing as any) * 3}px;
```
pjonsson, line 34: Can spacing be `as number` instead of any, or is that wrong?

pjonsson:
I'm not qualified to comment on the removed files, but I found a few minor nits that I commented about, The rest of this PR is "just" mechanical changes and they look consistent to me, so they are either all wrong or they are all right.

Considering how many files this PR touches, I think it would be great to get this merged soon.

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
na9da: `proj4.Proj()` is available but not `new proj4.Proj()`.

I have now modified the code to use only `proj`. Tested by [this spec](https://github.com/TerriaJS/terriajs/blob/baca8d1439eaec1ee46c416399d5f3b7e29e009d/test/Models/Catalog/esri/ArcGisFeatureServerCatalogItemSpec.ts#L109).

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
na9da: Should be fixed by [this commit](https://github.com/TerriaJS/terriajs/pull/7311/commits/24af89ddb16849a9d8b74bab0480f7b1799f010d).

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
na9da: Those are slightly different errors, this one is specifically on the prop `onOpenChanged` and TS requires the expect comment to precede it immediately.

lib/ReactViews/Tour/TourExplanationBox.tsx
na9da: That will require casting to `unknown` first - eg `p.theme.spacing as unknown as number`
```
Conversion of type 'string' to type 'number' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

lib/Core/ConsoleAnalytics.js
na9da: Good catch. Fixed.

lib/ReactViews/Tour/TourExplanationBox.tsx
pjonsson: Yeah, probably better to leave it as any then, so it's easy to find in the future.

lib/ReactViews/Map/Panels/SharePanel/SharePanel.tsx
pjonsson: I misunderstood what triggered it, what you describe is exactly what I wanted, thank you!

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -257,8 +256,8 @@ class FeatureServerStratum extends LoadableStratum(
         return undefined;
       }
 
-      const source = new proj4.Proj((proj4definitions as any)[wkid]);
-      const dest = new proj4.Proj("EPSG:4326");
+      const source = (proj4definitions as any)[wkid];
```
ljowen: I was able to compile without this cast
```suggestion
      const source = proj4definitions[wkid];
```

lib/Models/Catalog/Esri/ArcGisMapServerCatalogItem.ts
```diff
@@ -702,7 +701,9 @@ export function getRectangleFromLayer(
       return;
     }
 
+    // @ts-expect-error @types/proj4 doesn't define a constructor type
     const source = new proj4.Proj(Proj4Definitions[wkid]);
```
ljowen: If source and dest are just projection strings can't we just pass them to proj4 below?
`const source = Proj4Definitions[wkid]`

ljowen (approved):

na9da:
@ljowen - Sorry I forgot, but this also requires approving the terriamap PR https://github.com/TerriaJS/TerriaMap/pull/720. The main change there is removing the call to `polyfill()` which basically [does nothing](https://github.com/TerriaJS/terriajs/blob/baca8d1439eaec1ee46c416399d5f3b7e29e009d/lib/Core/polyfill.js#L8). It previously used to import polyfills from polyfill.io but was [replaced with core-js](https://github.com/TerriaJS/terriajs/commit/1aa18403456488a3324c21b92438f477e6e741ed).

na9da:
Reverted change removing polyfill.js to avoid breaking change in terriamap. It is better to lump this change with [webpack 5 upgrade](https://github.com/TerriaJS/terriajs/pull/7351) which will have several breaking changes.

PR #7342, opened 2024-11-29
Upgrade to TS 5.7+

na9da:
### What this PR does

Upgrades Typescript to TS 5.7+

**Breaking changes**

This PR  implements a workaround for error TS2855 that invalidates access to parent class fields as `super.field` [introduced in TS 5.3](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html#checks-for-super-property-accesses-on-instance-fields). In valid JS, super can only be used to access get/set accessors or methods from parent classes that are declared on the prototype, but not instance fields (declared as `x = "something"`). 

In our case though, traits are [mixed into models as getters](https://github.com/TerriaJS/terriajs/blob/baca8d1439eaec1ee46c416399d5f3b7e29e009d/lib/Models/Definition/addModelStrataView.ts#L32-L36) and not fields, however because we use mapped types to [type the resulting model](https://github.com/TerriaJS/terriajs/blob/baca8d1439eaec1ee46c416399d5f3b7e29e009d/lib/Models/Definition/ModelPropertiesFromTraits.ts#L25-L31) and mapped types does not allow specifying getter/setter variants, TS treats our trait accessors as fields and raises TS2855.

The workaround in this PR re-defines these erroring traits as getters which seems to convince TS (at least for now) that they are valid accessors even though they still are typed using mapped.

Hopefully in the future, we get this TS feature implementing [variant accessors for mapped types](https://github.com/microsoft/TypeScript/issues/43826), then we can get rid of this workaround.

### Test me

Test out - http://ci.terria.io/ts-5-upgrade

### Checklist


lib/Traits/TraitsClasses/CatalogMemberTraits.ts
```diff
@@ -107,15 +107,19 @@ class CatalogMemberTraits extends ModelTraits {
     name: "Name",
     description: "The name of the catalog item."
   })
-  name?: string;
+  get name(): string | undefined {
+    return;
```
pjonsson, line 111: If I read "return" in a getter, I expect some value to be returned, so when skimming code it would be quite easy to miss that this returns undefined despite being "implemented".

Would `get name(): string | undefined {}` give the same effect? That would make these "bogus-getters" more visually similar to the `name?: string` they replace, and different from a "real" getter.

zoran995: In typescript, getters have to return a value, if not TS2378 will be thrown
```
A 'get' accessor must return a value. ts(2378)
```
Maybe it can be made clearer that the default value is undefined by using `return undefined`.

zoran995:
@na9da amazing job making this work
p.s. I tried the same approach a month ago but couldn't make it work (the only difference is I tried replacing all traits with getters).

package.json
```diff
@@ -177,7 +178,7 @@
     "terriajs-tiff-imagery-provider": "2.13.3",
     "thredds-catalog-crawler": "0.0.7",
     "ts-essentials": "^5.0.0",
-    "typescript": "~5.2.0",
+    "typescript": "^5.7.2",
```
pjonsson: TypeScript 5.6 introduces `--noUncheckedSideEffectImports` (https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#the---nouncheckedsideeffectimports-option).

I was hoping that would have found the casing difference between import/export in #7344, but it didn't. It still seems like a nice thing to check though, can that option be enabled in `tsconfig.json`? I'm not sure where the ambient module declaration should go, but adding:
```
declare module "*.css" {}
```
in `lib/ThirdParty/globals.d.ts` makes tsc with `noUncheckedSideEffectImports` enabled pass for me.

pjonsson: Running `yarn gulp lint` prints:
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.3.5 <5.4.0

YOUR TYPESCRIPT VERSION: 5.7.2

Please only submit bug reports when using the officially supported version.

=============
```
So this PR more or less requires the latest typescript-eslint plugin unless we wish to stare at that warning every day. My lint-related PRs have been moderately successful so far, can you upgrade the typescript-eslint plugin along with this change?

Edit: It's not what I wrote, but I didn't mean the typescript-eslint plugin upgrade had to be in this PR, that seems independent and large enough to be in a separate PR that is merged before this.

lib/Traits/TraitsClasses/CatalogMemberTraits.ts
na9da: Yes, @zoran995 is right. A return statement for getter is enforced by TS.

package.json
na9da: Good suggestion, I can check.

lib/Models/Catalog/CatalogItems/OpenDataSoftCatalogItem.ts
```diff
@@ -631,8 +631,8 @@ export default class OpenDataSoftCatalogItem
           ([field, value]) => {
             // geoPoint2dFieldName will return a JSON object - spilt lat/lon columns
             if (field === this.geoPoint2dFieldName && isJsonObject(value)) {
-              cols.lat[index] = `${value.lat}` ?? "";
-              cols.lon[index] = `${value.lon}` ?? "";
+              cols.lat[index] = `${value.lat}`;
```
ljowen, line 634: this changes the behaviour so nullish lat will become "undefined" which we might not want in the table?

To preserve it becoming an empty string can do:
```suggestion
             cols.lat[index] = `${value.lat ?? ""}`;
```

ljowen (approved):

package.json
pjonsson: @na9da Since you've been quiet for a few days I had a quick look at upgrading and got a crash on a `use-at-your-own-risk` when running `yarn gulp lint`.

The error is not present when using ESLint 8.57.0, so I think the easiest way forward for upgrading typescript-eslint plugin is that someone looks at #7323 which should be merged before #7150. There is nothing intricate with #7323, it's just switching to the regular recommended configuration and disabling the new rules from that configuration that fails, and then adding a bunch of underscores to unused variables.

lib/Models/Catalog/CatalogItems/OpenDataSoftCatalogItem.ts
na9da: hmm.. afai understand this preserves the current behaviour.
i.e if `lat` is undefined then, both before and after, the resulting value will be `"undefined"`.

I didn't want to change any behaviour in this PR so I went with the trivial fix.

package.json
na9da: Thanks. I checked upgrading eslint and typescript-eslint. The new versions introduces a bunch of new errors, here's a rough split:

- 34  @typescript-eslint/no-empty-object-type 
  We'll have to disable these rules (at least inside Mixins)
- 326 @typescript-eslint/no-require-imports 
  I'll try to tackle this in the webpack5 PR
- 57 @typescript-eslint/no-unused-expressions 
  Could possibly fix in another PR to be merged before this one

Overall it looks like we should first merge webpack5 and then this one.

pjonsson: I rebased the ESLint 8 PR to latest main after you merged my unused variable fixes last week. I also see you have added the require-to-import conversion commit to the Webpack 5 PR, but I actually had something similar sitting around waiting for the CJS-to-ESM PR to get merged.

I do agree that it's probably better to get the ESLint upgrades merged even if that means some new rules have to be disabled; it's no worse than things are right now, and the disabled rules can be enabled one-by-one at a later stage.

As a general observation, I think you would have saved yourself some effort of having to fix merge conflicts if these other PRs with sweeping changes weren't hanging around in the queue for so long. The CJS-to-ESM PR is an obvious example of what I have in mind, and Zoran's react-imports change is another example. You might also get more help from others, I might have gotten around to upgrading typescript-eslint already if my ESLint 8 PR hadn't gotten stuck in the queue earlier.

lib/Models/Catalog/CatalogItems/OpenDataSoftCatalogItem.ts
ljowen: whoops my bad you're right

package.json
pjonsson: Now that ESLint 8 is merged, I put the update of typescript-eslint to the latest version in #7382.

yarn.lock
```diff
@@ -11619,45 +11466,102 @@ type-is@~1.6.18:
...
   version "0.0.6"
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha512-/aCDEGatGvZ2BIk+HmLf4ifCJFwvKFNb9/JeZPMulfgFracn9QFcAf5GO8B/mweUjSoblS5In0cWhqpfs/5PQA==
 
-typescript@~5.2.0:
-  version "5.2.2"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-5.2.2.tgz#5ebb5e5a5b75f085f22bc3f8460fba308310fa78"
-  integrity sha512-mI4WrpHsbCIcwT9cF4FZvr80QUeKvsUsUvKDoR+X/7XHQH98xYD8YHZg7ANtz2GtZt/CBq2QJ0thkGJMHfqc1w==
+typescript@^5.7.2:
+  version "5.7.2"
```
pjonsson: TypeScript 5.7.3 has been out for a while, so it seems reasonable to update yarn.lock with whatever the latest TypeScript version is just before this is merged.

package.json
pjonsson: Updated #7382 after the Webpack 5 PR was merged. That PR worked on top of its dependencies when I made it, but it's been 2 months and a couple of large PRs have been merged in the meantime, so if you review the PRs that #7382 depends on I will fix any remaining lint errors tomorrow.

pjonsson: On the subject of tsconfig.json settings: can we enable incremental mode so we don't have to wait so long on tsc after making a tiny change to the source code (#7424)?

And if the output from tsc isn't used, we could get an additional performance boost on top of that by not emitting files (#7425).

na9da: For `noUncheckedSideEffectImports` - we could look at that in a separate PR. Ideally we should also remove the remaining non-standard css imports using `style-loader!css-loader!` pattern.

pjonsson: Ok, done in #7464.

Since we are using babel which only operates on a single file when transpiling, I also put up #7465.

PR #7351, opened 2024-12-05
Webpack 5 upgrade

### LINKED ISSUE #6998: Upgrade to webpack 5

na9da:
### What this PR does

Fixes #6998 
Depends on https://github.com/TerriaJS/TerriaMap/pull/718 and https://github.com/TerriaJS/terriajs/pull/7311

Upgrades webpack and other build dependencies to the latest version.

Note: bulk of the changes are from running the saas-migrator that upgrades `.scss` files to modern api.

**Changes summary**
- Packages upgraded
  webpack, ...
- Babel transforms
  - Removed transformation to CJS for the default build. We still use it for node builds.
  - Other transforms removed `plugin-proposal-class-properties`, `proposal-object-rest-spread`, `plugin-syntax-dynamic-import`. These must now be well supported by browsers.
- Cleaned up configureWebpack.js
  - Removed rules that are no longer necessary 
  - Replaced `raw-loader` `file-loader` `url-loader` etc with builtin asset modules
  - Removed use of webpack specific qualifiers for imports like `raw-loader!file-loader` etc to make it more ESM compliant. Added extra config in webpack to correctly handle imports of static files (like csv and xml).
  - Added rules to polyfill nodejs modules that are referenced from some of our dependencies.
  - Removed `hot` reload config. AFAICT this was not being used and required maintaining two different code paths in webpack configuration. If required we can add it back later after testing that it works.
  - Replaced multiple `configureWebpack` parameters with a single options parameter.
  - Removed `string-replace-webpack-plugin` as it was outdated and has a few vulnerabilities. 
    - [x] TODO: Move rewrite of default Cesium API key warning to the app
- Upgraded SASS
  - Migrated SASS files to use modern API (by running the `sass-migrator` script)
  - Replaced webpack aliases `~terriajs-variables` and `~terriajs-mixins` with respective relative path. This was necessary to run the migrator. Also it means simpler webpack config.
  -  `css-modules-typescript-loader` that generates TS types for SASS modules is outdated and no longer works with webpack 5. I have made a [repo](https://github.com/TerriaJS/terriajs-typings-for-css-modules-loader) from a [fork of another similar package](https://github.com/TeamSupercell/typings-for-css-modules-loader/pull/93) that can generate types. The other alternative would be to use this [TS plugin](https://www.npmjs.com/package/typescript-plugin-css-modules) - but not sure if it'll slow down compilation/editing.
- Modified TIFFImageryProvider to work with webpack5 build.
- [ ] TODO: Find a replacement for `svg-sprite-loader`. It appears to be not maintained and has a bunch of audit warnings.
- Rewrote a couple of specs that depended on some CJS feature
- [x] TODO: Test in Windows.

I'll annotate the changes to make review easier.

### Test me

- Specs should pass
- Test everything! - http://ci.terria.io/webpack5

### Checklist


package.json
```diff
@@ -87,18 +87,17 @@
...
     "assimpjs": "^0.0.7",
-    "babel-loader": "^8.2.3",
-    "babel-plugin-lodash": "^3.3.4",
+    "babel-loader": "^9.2.1",
     "bottleneck": "^2.19.5",
     "catalog-converter": "^0.0.9",
     "classnames": "^2.3.1",
     "clipboard": "^2.0.8",
-    "commander": "^11.1.0 ",
+    "commander": "^12.1.0 ",
```
pjonsson, line 89: Commit 0a451d13ea downgraded commander to 11.1.0 so it would work with node 16.

I'm not against bumping the engine version though, I would prefer to bump engines to node >= 20 so we could "easily" get rid of the security issues in terriajs-server as well while we're at it.

package.json
```diff
@@ -165,25 +161,22 @@
     "react-virtual": "~2.3.2",
     "resolve-url-loader": "^5.0.0",
     "retry": "^0.12.0",
-    "sass-loader": "^10",
+    "sass-loader": "^16.0.3",
```
pjonsson, line 158: I'm fairly sure this also requires bumping the engine version.

pjonsson:
I looked through this and besides the things I commented on, the things that are there looks good to me. Would love to get the (approved) CJS to ESM PR merged soon so this PR doesn't end up in limbo.

package.json
na9da: Agree that we should bump the engine version, will look into it. In this case though commander is only used for a couple of CLI things and the CI hasn't tripped on it. So it should be fine i guess.

package.json
```diff
@@ -87,18 +87,17 @@
...
     "clipboard": "^2.0.8",
-    "commander": "^11.1.0 ",
+    "commander": "^12.1.0 ",
     "copy-webpack-plugin": "^6.4.0",
     "core-js": "^3.1.4",
     "create-react-class": "^15.7.0",
-    "css-loader": "^2.1.0",
-    "css-modules-typescript-loader": "^2.0.4",
+    "css-loader": "^7.1.2",
+    "terriajs-typings-for-css-modules-loader": "^2.5.2",
```
pjonsson: Non-alphabetical sorting, so this package will be shuffled down to the other terriajs-packages the next time someone runs `yarn add`.

pjonsson:
@na9da just noticed that the late revert of polyfill.js in CJS-to-ESM PR changed the `export defaults ..` back into the previous `module.exports = ...`.

buildprocess/configureWebpack.js
```diff
@@ -1,42 +1,47 @@
 const path = require("path");
 const CopyPlugin = require("copy-webpack-plugin");
-const StringReplacePlugin = require("string-replace-webpack-plugin");
 const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const ForkTsCheckerNotifierWebpackPlugin = require("fork-ts-checker-notifier-webpack-plugin");
 const webpack = require("webpack");
 
-function configureWebpack(
+/**
```
ljowen, line 7: 👍

pjonsson:
Since this repository was just converted to ESM, should the other terriajs-repositories (that aren't forks of something else) also be converted to ESM?

I had a look at terriajs-server and it seems quite easy, but I had to upgrade jasmine and yargs.

na9da:
> @na9da just noticed that the late revert of polyfill.js in CJS-to-ESM PR changed the export defaults .. back into the previous module.exports = ....

Yep, i'll make those changes in this PR. I didn't want to create two successive breaking changes.

.babelrc
```diff
@@ -11,13 +11,8 @@
...
   ],
   "plugins": [
-    "@babel/plugin-transform-modules-commonjs",
     ["@babel/plugin-proposal-decorators", { "legacy": true }],
-    "@babel/plugin-proposal-class-properties",
-    "@babel/proposal-object-rest-spread",
-    "babel-plugin-styled-components",
-    "@babel/plugin-syntax-dynamic-import",
-    "babel-plugin-lodash"
+    "babel-plugin-styled-components"
```
pjonsson, line 15: Is there anything else controlling babel plugins in the tree, or should the removed babel plugins also be removed from `package.json`/`yarn.lock`?

Edit: (I think removing these plugins from `package.json`/`yarn.lock` would remove a bunch of warnings, and possibly also remove something from `yarn audit` so I would love that.)

package.json
```diff
@@ -165,25 +161,22 @@
...
     "thredds-catalog-crawler": "0.0.7",
     "ts-essentials": "^5.0.0",
     "typescript": "~5.2.0",
     "urijs": "^1.18.12",
-    "url-loader": "^1.1.2",
-    "webpack": "~4.47.0",
-    "webpack-cli": "^4.10.0",
-    "webpack-dev-server": "^4.15.2",
-    "worker-loader": "^2.0.0"
+    "webpack": "^5.96.1",
```
pjonsson, line 172: I've seen in the upgrade guide that webpack 5 contains types, but it was listed under internal improvements.

If there are type annotations in the webpack 5 package, I'm guessing the types/webpack dev dependency for terriajs should be removed, and if there aren't I'm guessing the types/webpack version should also be bumped to 5.x?

buildprocess/configureWebpack.js
```diff
@@ -303,12 +152,46 @@ function configureWebpack(
...
+  // Define NODE_ENV from mode setting - used in a few places in Terria
+  config.plugins.push(
+    new webpack.DefinePlugin({
+      "process.env.NODE_ENV": JSON.stringify(
+        devMode ? "development" : "production"
+      )
+    })
+  );
+
+  // Fork TypeScript type checking to a separate thread
```
pjonsson, line 191: I can see the fork-ts-checker-webpack-plugin 1.0 had support for webpack 5 alpha back in 2019 according to the release notes, but I'm guessing things have changed in webpack 5 since that time. Should the `fork-ts-checker-webpack-plugin` and `fork-ts-checker-notifier-webpack-plugin` be bumped to the latest version at the same time as upgrading to webpack 5?

.babelrc
pjonsson, line 15: I'm not sure where this comment fits, but commit dfaf253adc added a `lib/Core/prerequisites.js` which seems to be about compiling to ES5 (which is also what Babel does?).

Should some things be removed from that file as well?

package.json
```diff
@@ -6,6 +6,7 @@
   "engines": {
     "node": ">= 16.0.0"
   },
+  "sideEffects": false,
```
pjonsson: I read the webpack manual for optimization, and I get the impression this flag helps with reducing the size of the output.  Should other terriajs-packages (terriajs-server, thredds-catalog-crawler, etc.) have a `package.json` that contains `sideEffects: false` as well?

package.json
pjonsson: The previous commander problem that caused the downgrade was most likely fixed by commit 88e587c66e0d4.

test/Models/TerriaSpec.ts
```diff
@@ -36,22 +36,18 @@ import { buildShareLink } from "../../lib/ReactViews/Map/Panels/SharePanel/Build
...
-const mapConfigDereferencedJson = require("../../wwwroot/test/Magda/map-config-dereferenced.json");
-const mapConfigDereferencedString = JSON.stringify(mapConfigDereferencedJson);
-
-const mapConfigDereferencedNewJson = require("../../wwwroot/test/Magda/map-config-dereferenced-new.json");
-const mapConfigDereferencedNewString = JSON.stringify(
-  mapConfigDereferencedNewJson
-);
+import mapConfigBasicJson from "../../wwwroot/test/Magda/map-config-basic.json";
+import mapConfigV7Json from "../../wwwroot/test/Magda/map-config-v7.json";
+import mapConfigInlineInitJson from "../../wwwroot/test/Magda/map-config-inline-init.json";
```
pjonsson, line 41: When I run `yarn gulp build` on this branch, I get this warning just above the prompt after the build completes:
```
WARNING in ./node_modules/terriajs-cesium/Source/Core/buildModuleUrl.js 50:37-48
Critical dependency: Accessing import.meta directly is unsupported (only property access or destructuring is supported)
 @ ./test/Models/TerriaSpec.ts 41:0-72 113:13-27 118:13-27
```

buildprocess/webpack.config.dev.js
```diff
@@ -1 +1 @@
-module.exports = require("./webpack.config.make")(false, true);
+module.exports = require("./webpack.config.make")(false);
```
pjonsson: This config disables devMode, was the wrong parameter removed?

buildprocess/configureWebpack.js
```diff
@@ -181,35 +72,30 @@ function configureWebpack(
...
+    type: "asset/source" // inlines xml as raw text
   });
 
+  // Allow .DAC file to be imported from MouseCoords.ts as URL
   config.module.rules.push({
-    test: /\.json|\.xml$/,
-    include: path.resolve(cesiumDir, "Source", "Assets"),
-    loader: require.resolve("file-loader"),
-    type: "javascript/auto"
+    test: /\.(DAC)$/i,
```
pjonsson, line 84: I think the parenthesis can be omitted, so `/\.DAC$/i` should be enough.

buildprocess/configureWebpack.js
```diff
@@ -181,35 +72,30 @@ function configureWebpack(
...
+    include: [path.resolve(terriaJSBasePath, "wwwroot", "data")]
   });
 
+  // Allow import of .css files from tinymce package
   config.module.rules.push({
-    test: /\.wasm$/,
-    include: path.resolve(cesiumDir, "Source", "ThirdParty"),
-    loader: require.resolve("file-loader"),
-    type: "javascript/auto"
+    test: /\.(css)$/i,
```
pjonsson, line 91: I think this parenthesis can be omitted too, so `/\.css$/i` should be enough.

lib/ReactViews/Generic/Editor.jsx
```diff
@@ -9,8 +9,8 @@ import "tinymce/icons/default";
 import "tinymce/themes/silver";
 import "tinymce/models/dom";
 /* Import a skin (can be a custom skin instead of the default) */
-// import "!!style-loader!css-loader?sourceMap!tinymce/skins/ui/oxide/skin.min.css";
-import "!!style-loader!css-loader?sourceMap!./editor.skin.min.css"; // Custom borderless skin
+// import "!!style-loader!css-loader!tinymce/skins/ui/oxide/skin.min.css";
+import "!!style-loader!css-loader!./editor.skin.min.css"; // Custom borderless skin
```
pjonsson, line 12: The style-loader seems to have been removed from the webpack config, do these imports work as intended despite that?

buildprocess/configureWebpack.js
pjonsson: Running `yarn gulp build` with empty `wwwroot/build` runtimes:

Webpack 4: 22.3s
Webpack 5 (this branch): 43.8s
Webpack 5 (this branch) + fork-ts-* version 9.x: 21s

so I suggest bumping the versions even if it isn't strictly required.

package.json
```diff
@@ -6,6 +6,7 @@
...
   "repository": {
     "type": "git",
     "url": "http://github.com/TerriaJS/terriajs"
```
pjonsson: There is a resolution of `terser-webpack-plugin` to version 4 a few lines below (introduced in commit 5b27cb2435). You might be able to drop the resolution, or at least update the resolution to version 5.

lib/ReactViews/Map/Panels/HelpPanel/help-panel.scss
```diff
@@ -1,15 +1,15 @@
-@import "~terriajs-variables";
-@import "../../../../Sass/common/mixins";
+@use "../../../../Sass/common/_variables";
+@use "../../../../Sass/common/mixins";
```
pjonsson, line 2: Some of these files import `mixins`, and others import `_mixins` (see `guidance-dot.scss` vs this file, both in subdirectories of `ReactViews`), is that intentional?

pjonsson:
>     * Cleaned up configureWebpack.js
>       
>       * Removed rules that are no longer necessary

This is great, navigating around old unused things is confusing for newcomers.

>       * Replaced multiple `configureWebpack` parameters with a single options parameter.

I don't understand this change in general; the parameters are packed into a record at the caller and then immediately unpacked in the arguments of the callee, so the change is effectively adding braces around what was there previously. I also get a complaint that the JSDoc is invalid because it references `options.x` but does not contain the correct `@param options`, and a complaint on the caller that it's missing the babelLoader parameter expected by the function. For languages I know, I usually enable some extra checks in my IDE, but these complaints are with the default settings.

>       * Removed `string-replace-webpack-plugin` as it was outdated and has a few vulnerabilities.
>         
>         * [ ]  TODO: Move rewrite of default Cesium API key warning to the app

I'm all for improved security so I love this, but just updating to Webpack 5 fixes a bunch of vulnerabilities, and this PR blocks a lot of other things. Can the plugin be used with Webpack 5 so this PR can get merged first, and you then do this rewrite in a separate PR?

Edit: I just realized terriajs keeps a fork of Cesium. Isn't it easiest to change the message in that fork, and not rewrite anything at build time?

>     * Upgraded SASS
>       
>       * `css-modules-typescript-loader` that generates TS types for SASS modules is outdated and no longer works with webpack 5. I have made a [repo](https://github.com/TerriaJS/terriajs-typings-for-css-modules-loader) from a [fork of another similar package](https://github.com/TeamSupercell/typings-for-css-modules-loader/pull/93) that can generate types. The other alternative would be to use this [TS plugin](https://www.npmjs.com/package/typescript-plugin-css-modules) - but not sure if it'll slow down compilation/editing.

I had no idea how things were built before looking at this PR and I'm not thrilled about how the generated `scss.d.ts` files are committed to the repository (we've been trying to build things in separate build directories for 20+ years with autoconf/CMake), but I had a look at some sibling package of the linked TS plugin trying to figure out what was wrong with `related-maps.scss` (#7427), and what you've done here is far better than the listed alternatives.

>     * [ ]  TODO: Find a replacement for `svg-sprite-loader`. It appears to be not maintained and has a bunch of audit warnings.

If the plugin works with Webpack 5, I think that should be handled in a separate PR because this PR in its current form brings a lot of improvements already.

package.json
pjonsson: Having read a bit more, I realized a module can be imported for side effects only (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#import_a_module_for_its_side_effects_only), and there seems to be quite a few side effecting imports under `lib/`:
```
$ git grep 'import "' lib
lib/Core/polyfill.js:import "core-js/features/promise";
lib/Core/prerequisites.js:import "core-js/features/symbol";
lib/Core/prerequisites.js:import "core-js/features/promise";
lib/Core/prerequisites.js:import "regenerator-runtime/runtime";
lib/Core/prerequisites.js:import "core-js/features/global-this";
lib/Models/GlobeOrMap.ts:import "./Feature/ImageryLayerFeatureInfo"; // overrides Cesium's prototype.configureDescriptionFromProperties
lib/ReactViews/Generic/Editor.jsx:import "!!style-loader!css-loader!./editor.skin.min.css"; // Custom borderless skin
!! TRIMMED tinymce imports in Editor.jsx
lib/ReactViews/Generic/Timer/drawTimer.js:import "d3-transition";
lib/ReactViews/StandardUserInterface/StandardUserInterface.tsx:import "inobounce";
lib/ReactViews/Workbench/WorkbenchList.tsx:import "!!style-loader!css-loader!./sortable.css";
lib/ThirdParty/urijs/index.d.ts:import "urijs/src/URITemplate";
```
* polyfill/prerequisites: Used by TerriaMap. The point of both polyfill and prerequisites seems to be to "pollute" the global namespace with compatibility functions for browsers that do not support them natively (both "Tip" boxes under https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free), won't that functionality break for TerriaMap when setting `sideEffects: false`?
* css-imports in Editor and WorkbenchList: I'm guessing second tip box from the link in the previous bullet applies to these?
* GlobeOrMap: the only place that overrides `ImageryLayerFeatureInfo`. Can the side-effecting import be eliminated by just putting the 38 lines from the imported file directly into GlobeOrMap?
* Editor: lots of side effecting imports of tinymce modules, and then Editor is imported lazily by StoryEditor.jsx. Don't understand the details, but looks fragile in general.
* drawTimer: commit 9035320a8db added the import in 2019 to "fix the buses layer". I don't know what the "buses layer" is, but the versions of the d3-packages were updated in 2024, so it might be possible to change the import to something more specific now.
* StandardUserInterface: inobounce seems to be a function called from the top level, not sure this can be avoided.
* ThirdParty/urijs: #7435 removes this.

buildprocess/configureWebpack.js
```diff
@@ -181,35 +72,30 @@ function configureWebpack(
...
+    type: "asset/source",
+    include: [path.dirname(require.resolve("tinymce/package.json"))]
   });
 
+  // Remove Cesium debug mode checks in production. This might slightly improve performance.
+  // TODO: It will be good to have it enabled in devMode though, to discover issues with code
+  // however doing so currently breaks a few specs.
```
pjonsson, line 98: #7433 and #7439.

lib/ReactViews/Map/Panels/HelpPanel/help-panel.scss
zoran995: The leading underscore indicates sass partials and indicates to sass that it should not be compiled into its own CSS file. When importing/using these files, it is safe to omit the underscore, as a compiler is smart enough to handle that  but both options works the same

buildprocess/configureWebpack.js
```diff
@@ -219,56 +105,19 @@ function configureWebpack(
...
-  });
-
-  // config.module.loaders.push({
-  //     test: /\.(ttf|eot|svg)(\?.+)?$/,
-  //     include: path.resolve(fontAwesomeDir, 'fonts'),
-  //     loader: require.resolve('file-loader')
-  // });
-
+  // Convert imported svg icons into a sprite
+  // TODO: svg-sprite-loader has several security warnings - we need to find an alternative
```
zoran995, line 117: If it works correctly with this plugin, let's create a ticket and handle icons in a separate PR.

lib/Sass/common/_buttons.scss
```diff
@@ -36,37 +36,38 @@ $btn-setting-size: 35px;
...
 }
 
-.btn-small {
-  padding: 0;
-  &:hover,
-  &:focus {
-    color: $color-primary;
-  }
-}
+// TODO: fix?
```
zoran995: Just marking with a comment as it can be missed easily

package.json
```diff
@@ -87,18 +87,17 @@
...
-    "babel-loader": "^8.2.3",
-    "babel-plugin-lodash": "^3.3.4",
+    "babel-loader": "^9.2.1",
     "bottleneck": "^2.19.5",
     "catalog-converter": "^0.0.9",
     "classnames": "^2.3.1",
     "clipboard": "^2.0.8",
-    "commander": "^11.1.0 ",
+    "commander": "^12.1.0 ",
     "copy-webpack-plugin": "^6.4.0",
```
zoran995: should we also bump this, or leave it for follow-up PR? the latest version is 12.x

package.json
zoran995: good catch `@types/webpack` should be removed https://webpack.js.org/blog/2020-10-10-webpack-5-release/#typescript-typings

.babelrc
zoran995: to make it easier pjonsson is referring to 
- `@babel/plugin-proposal-class-properties`
- `@babel/plugin-proposal-nullish-coalescing-operator`
- `@babel/plugin-proposal-object-rest-spread`
- `@babel/plugin-proposal-optional-chaining`
- `babel/plugin-syntax-dynamic-import`

package.json
zoran995: bumped in #7450

lib/ReactViews/Generic/Editor.jsx
zoran995: `style-loaders` are still used in terria-map and have to stay due to this, @na9da are we safe to remove them from the terriajs part and leave those styles (we might just convert those to scss)

test/Models/TerriaSpec.ts
zoran995: I also see it in the console during the build, and if I understand everything correctly, it is safe to ignore and look into details later. Cesium has safety measures in place around this (https://github.com/CesiumGS/cesium/blob/main/packages/engine/Source/Core/buildModuleUrl.js#L44), and we would actually never reach this line (https://github.com/TerriaJS/terriajs/blob/webpack5/lib/Models/Terria.ts#L747).

lib/ReactViews/Generic/Editor.jsx
pjonsson: I'm more than happy if things work as they are, so no need to change this for me.

(This PR is closing up on 2 months in the queue, and beyond the security fixes this brings this PR is blocking other major updates like TypeScript 5.7, so I suggest we don't extend the scope unless we absolutely have to.)

test/Models/TerriaSpec.ts
pjonsson: Thanks for looking into this. If the warning is harmless, a single 3 line warning is a definite improvement over the current flood of warnings from sass.

na9da: This requires an upstream fix (our in our version of cesium).

There was a [cesium PR](https://github.com/CesiumGS/cesium/pull/11511) from before for fixing it, but I guess more recent webpack doesn't accept that either.

From [webpack documentation](https://webpack.js.org/api/module-variables/#importmeta), only these two references are valid:
```js
// fine to use
console.log(import.meta.url);
const { url } = import.meta;
```

pjonsson: @na9da is that a blocking issue for merging this? If it's just a single 3-line warning but everything still works that is a significant improvement over the current situation.

I think Zoran's PR on top of this fixes all my comments (plus a bunch of other things), what remains to be done on this PR to get merged?

lib/Sass/common/_buttons.scss
```diff
@@ -36,37 +36,29 @@ $btn-setting-size: 35px;
   margin: 0;
 }
 
-.btn-small {
```
zoran995, line 39: This is not used anywhere anymore

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/my-data-tab.scss
```diff
@@ -124,20 +124,18 @@
...
-  margin: $padding;
+  font-family: variables.$font-base;
+  font-size: variables.$font-size-mid-small;
+  margin: variables.$padding;
 }
 
 .btn--back-to-my-data {
   composes: btn from "../../../../Sass/common/_buttons.scss";
-  composes: btn-small from "../../../../Sass/common/_buttons.scss";
-  composes: btn--left from "../../../../Sass/common/_buttons.scss";
```
zoran995, line 140: `btn--left` was an icon definition so it is safe to remove
`btn-small` wouldn't make any point here as we have all its styles defined/overriden here or in styled-component

buildprocess/configureWebpack.js
```diff
@@ -419,12 +258,37 @@ function configureWebpack(
   );
 
   // Alias all lodash imports (including from our dependencies) to lodash-es
+  // This saves close to ~600KB unzipped
+  // Maybe we can remove this when lodash 5 comes out (?)
   config.resolve.alias["lodash"] = "lodash-es";
 
-  // Fix issue with pmtiles TS import (through protomaps-leaflet)
-  config.resolve.alias["pmtiles"] = "pmtiles/dist/index.js";
-
```
zoran995, line 426: @nf-s You added this recently so, can you please just check if everything works correctly without this. Build was failing when this is set

zoran995 (commented):
I have cherry-picked changes from #7450, it should be easier to review now as code-diff overthere got pretty large and we can now handle this in single review cycle

- Fixed review comments
 - Removed unused weback plugins
- upgraded webpack plugins
- I didn't see any reason why we would need `pmtiles` alias in webpack config - marked it with review comment for double check.
- I had to restore worker-loader as webpack native web worker support tends to get pretty complex, I will keep looking for a fix but it might be good to keep worker-loader and handle that as a follow-up. Important note: webpack worker support behaves differently in development and production builds. That's why it works on CI but fails locally. I have run into this issue in the past but still don't have a solution that always works.
- side-effects are removed from package.json we can look into it later

buildprocess/configureWebpack.js
na9da: yeah, i think it works without the aliasing - but could be a webpack 5 thing.

na9da:
Regarding the web worker change, it appears to work if we set [publicPath](https://github.com/TerriaJS/TerriaMap/blob/544e2208f24d1bc68c97743313e6c98f778b1430/buildprocess/webpack.config.js#L17) to `auto` instead of `build/`. 

I think webpack will use `publicPath` + directory-containing-parent-script which  evaluates to `build/build/<downloadWorkerChunkName>.js`. 

We could test it out in a different PR.

lib/Core/polyfill.js
```diff
@@ -8,4 +8,4 @@ function polyfill(callback) {
   callback();
 }
 
-module.exports = polyfill;
+export default polyfill;
```
pjonsson: The TerriaMap PR removes the import of polyfill, can this file be removed completely?

package.json
```diff
@@ -14,24 +14,18 @@
...
     "underscore": "^1.12.1"
   },
   "dependencies": {
     "@babel/core": "^7.23.5",
     "@babel/parser": "^7.23.5",
-    "@babel/plugin-proposal-class-properties": "^7.18.6",
     "@babel/plugin-proposal-decorators": "^7.22.7",
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.18.6",
-    "@babel/plugin-proposal-object-rest-spread": "^7.20.7",
-    "@babel/plugin-proposal-optional-chaining": "^7.21.0",
```
pjonsson, line 27: Should `@babel/plugin-syntax-dynamic-import` also be removed?

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Geospatial data visualization platform.",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">= 16.0.0"
+    "node": ">= 20.0.0"
```
pjonsson, line 7: Should the `types/node` package version be updated to 20.x?

na9da: That follows the typescript version. Will do it when merging https://github.com/TerriaJS/terriajs/pull/7342

pjonsson:
> Regarding the web worker change, [..]
> 
> We could test it out in a different PR.

If this works in the current state I suggest merging it, this PR is already a significant step forward.

I'll rebase the ESLint fixes after this gets merged so the TypeScript upgrade can get merged too.

na9da:
Merging! Thanks everyone, for the contributions.

PS: will merge the [terriamap PR](https://github.com/TerriaJS/TerriaMap/pull/718) after a new terriajs release, but i'll see if I can include the Typescript upgrade PR into this release.

pjonsson:
@na9da since you wrote "Merging!" I was going to rebase the ESLint fixes for TS 5.7 before I go to bed so those could be merged, but I'm guessing you got interrupted by something else. Will this PR get merged soon, or should I go to bed now and rebase things tomorrow?

ljowen (approved):
Tested building on OSX and Windows10 (arm64)

PR #7370, opened 2024-12-13
Tile ArcGis FeatureServer requests

### LINKED ISSUE #3661: Look into tiling ArcGis feature service requests
Currently we just attempt to grab all features from an Esri Feature Service, which 

1. Is very slow because we don't limit requests to the extent of the map 
    - eg if you're looking in Syd you'll get data for Canberra
2. With older versions of ArcGIS server I think you get a subset of features
3. Sometimes fails with newer versions of ArcGIS server if you exceed the feature limit
    - need to specify `returnExceededLimitFeatures: true` as a query param then you at least get something
https://developers.arcgis.com/rest/services-reference/query-feature-service-layer-.htm

We could move to a tiling approach which would enable us to be smarter about which features we request.

Basically you send query params like
````
returnGeometry: true
spatialRel: esriSpatialRelIntersects
maxAllowableOffset: 19567.879241000017 <---- allows simplifying the geom
geometry {
    "xmin":10018754.171389025,
    "ymin":-5009377.08569302,
    "xmax":15028131.257085029,
    "ymax":0.000002983957529067993,
    "spatialReference":{"wkid":102100,"latestWkid":3857}
}
geometryType: esriGeometryEnvelope
inSR: 102100
outFields: *
outSR: 102100
resultType: tile
}
````
And with newer versions of ArcGIS Server you can also use `quantizationParameters` which give you pixel coords rather than geom coords which are typically smaller in data size
````
quantizationParameters: {
   "mode":"view",
   "originPosition":"upperLeft",
   "tolerance":19567.879241000017,
   "extent":{
      "xmin":10018754.171389025,
      "ymin":-10018754.171389025, "xmax":20037508.342781033,
      "ymax":0.000002983957529067993,
       "spatialReference":{"wkid":102100,"latestWkid":3857}
    }
}
````

### PR DESCRIPTION
**Pull request by nf-s**
### Tile ArcGis FeatureServer requests

Fixes https://github.com/TerriaJS/terriajs/issues/3661 https://github.com/TerriaJS/terriajs/issues/7331 https://github.com/TerriaJS/terriajs/issues/6840

~Merge after https://github.com/TerriaJS/terriajs/pull/7144~

### Main breaking changes

- Add tiling support to `ArcGisFeatureServerCatalogItem` - this will be enabled by default if the server supports tiling and unsupported marker/point styles aren't used. See `ArcGisFeatureServerCatalogTraits` `tileRequests`. When enabled, `pbf` tiles will be fetched and drawn using `ProtomapsImageryProvider`

  - This can be disabled by setting `tileRequests` to `false`
  - For point features, only the following `PointSymbolTraits` are available - fill, stroke, radius (which uses height). Custom markers (markers other than `point` or `circle`) are not supported.

### Other changes

- Add `request` parameter to `ArcGisImageServerImageryProvider.buildImageResource` - this enables Cesium to manage requests
- Decrease protomaps tile buffer to 32 pixels (from 64) to increase performance
- Change `ProtomapsImageryProvider` to use a "soft" minimum level, so no tiles will be created below the `minimumZoom` provided.
- Move `ProtomapsImageryProvider.pickFeatures` GeoJSON logic to inside `ProtomapsGeojsonSource.pickFeatures`.
- Fix `FeatureInfoUrlTemplateMixin` reactivity warnings
- Move `GeojsonMixin` protomaps paint/label rules to `tableStyleToProtomaps`.
  - Also create `getStyleReactiveDependencies` that can be used to track (and react to) table styling traits
- Add `dash` to `OutlineStyleTraits` (only supported by line features), and `outlineStyle` to `LegendTraits`.
- Add `MinMaxLevelMixin` to `ArcGisFeatureServerCatalogItem` - only applied when tiling requests
- Tweaked `TableStyleMap` and `TableColorMap` conditions to handle empty `TableColumns` (to support styling `ArcGisFeatureServerCatalogItem` when tiling requests)
- Fix bug with expanding "Developer Details" in the error modal
- Fix regression bug from https://github.com/TerriaJS/terriajs/pull/7144 (not released), GeoJson `MultiPolygon` features were being dropped

### Todo

- [x] Get table styling working (with esri style)
- [x] Check all properties that are relevant to tiling (like is pbf supported, is quantization supported, etc)
- [x] Fix issues along tile border
- [x] Optimise number of features ~can we filter out null features in the request?~
- [x] Feature picking
- [x] Test more layers
- [x] Write unit tests
  - [x] New `ProtomapsArcGisPbfSource`
  - [x] Added `request` to image server
  - [x] `tableStyleToProtomaps` new properties - dash/color/style/width + point features
- [x] ~Make sure requests are cancelled when moving cesium camera around~ This is harder than expected, so will leave unless it seems necessary later
  - [x] ~This is a problem for larger layers, need to at least cancel requests when layers are turned off or removed from the wokrbench~ This is a big issue, but I can't find an easy way to do this. Not for this release, will make a ticket. This ties into larger issues of large network requests crashing Terria. After testing many layers, I haven't been able to crash Terria, but sometimes it takes a while for layer requests to finish, even if you have removed it from the workbench
- [x] Fix performance regression with geojson
- [x] ~Add support for marker icons...~ Not for this release, will make a ticket
- [x] Add min/max scale support
- [x] Fix outline legend
- [x] Add outline dashed styles
- [x] Add `MinMaxLevelMixin` mixin
- [ ] ~Fix bug where tiles are reloaded everytime the style changes (eg edit style)~ leaving this bug due to lack of time - will create a ticket
- [ ] ~Fix share bug with minmaxlevel~ see https://github.com/TerriaJS/terriajs/pull/7370#issuecomment-2632936255 -  leaving this bug due to lack of time - will create a ticket

### Test me

- http://ci.terria.io/tile-arcgis-featureserver/#clean&https://gist.githubusercontent.com/nf-s/1a44b31faa1dd28dc9d23daa0660f50b/raw/ed093cfbae576e5fb0b209fa04711a5c99e55ea7/featuer-server-tile.json/#clean&https://gist.githubusercontent.com/nf-s/1a44b31faa1dd28dc9d23daa0660f50b/raw/ed093cfbae576e5fb0b209fa04711a5c99e55ea7/featuer-server-tile.json
- Test all the layers

### Checklist

### CODE REVIEW

**Comment by zoran995** [comment]
It might be of interest to look into the layer mentioned in #6840. That layer is failing with
> {
    "code": 400,
    "message": "",
    "details": [
        "'Invalid field: objectid' parameter is invalid"
    ]
}

**Comment by nf-s** [comment]
There is a bug I haven't been able to fix. It occurs in the 3D (cesium) viewer, with layers that have a minimum zoom level, when you create a share link with a picked feature

- [Test link](http://ci.terria.io/tile-arcgis-featureserver/#start=%7B%22version%22%3A%228.0.0%22%2C%22initSources%22%3A%5B%7B%22stratum%22%3A%22user%22%2C%22models%22%3A%7B%22__User-Added_Data__%22%3A%7B%22isOpen%22%3Atrue%2C%22members%22%3A%5B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%22%5D%2C%22knownContainerUniqueIds%22%3A%5B%22%2F%22%5D%2C%22type%22%3A%22group%22%7D%2C%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%22%3A%7B%22isOpen%22%3Atrue%2C%22url%22%3A%22https%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%22%2C%22knownContainerUniqueIds%22%3A%5B%22__User-Added_Data__%22%5D%2C%22type%22%3A%22esri-group%22%7D%2C%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Administrative_Boundaries_Theme_multiCRS%2FFeatureServer%22%3A%7B%22isOpen%22%3Afalse%2C%22knownContainerUniqueIds%22%3A%5B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%22%5D%2C%22type%22%3A%22esri-featureServer-group%22%7D%2C%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Water_Theme_multiCRS%2FFeatureServer%22%3A%7B%22isOpen%22%3Atrue%2C%22knownContainerUniqueIds%22%3A%5B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%22%5D%2C%22type%22%3A%22esri-featureServer-group%22%7D%2C%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Water_Theme_multiCRS%2FFeatureServer%2F6%22%3A%7B%22knownContainerUniqueIds%22%3A%5B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Water_Theme_multiCRS%2FFeatureServer%22%5D%2C%22type%22%3A%22esri-featureServer%22%7D%2C%22%2F%22%3A%7B%22type%22%3A%22group%22%7D%7D%2C%22workbench%22%3A%5B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Water_Theme_multiCRS%2FFeatureServer%2F6%22%5D%2C%22timeline%22%3A%5B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Water_Theme_multiCRS%2FFeatureServer%2F6%22%5D%2C%22initialCamera%22%3A%7B%22west%22%3A150.76903928591705%2C%22south%22%3A-33.87578706692171%2C%22east%22%3A150.8090195823401%2C%22north%22%3A-33.82816771331238%2C%22position%22%3A%7B%22x%22%3A-4631421.501641395%2C%22y%22%3A2589577.8893886614%2C%22z%22%3A-3535371.165973835%7D%2C%22direction%22%3A%7B%22x%22%3A0.7248641500756964%2C%22y%22%3A-0.40529504282460255%2C%22z%22%3A0.5570528630182615%7D%2C%22up%22%3A%7B%22x%22%3A-0.4862116981624778%2C%22y%22%3A0.27185672102561875%2C%22z%22%3A0.8304770362888781%7D%7D%2C%22homeCamera%22%3A%7B%22west%22%3A109%2C%22south%22%3A-45%2C%22east%22%3A158%2C%22north%22%3A-8%7D%2C%22viewerMode%22%3A%223dSmooth%22%2C%22showSplitter%22%3Afalse%2C%22splitPosition%22%3A0.4999%2C%22settings%22%3A%7B%22baseMaximumScreenSpaceError%22%3A2%2C%22useNativeResolution%22%3Atrue%2C%22alwaysShowTimeline%22%3Afalse%2C%22baseMapId%22%3A%22basemap-natural-earth-II%22%2C%22terrainSplitDirection%22%3A0%2C%22depthTestAgainstTerrainEnabled%22%3Afalse%7D%2C%22pickedFeatures%22%3A%7B%22providerCoords%22%3A%7B%22%2Fhttps%3A%2F%2Fportal.spatial.nsw.gov.au%2Fserver%2Frest%2Fservices%2FNSW_Water_Theme_multiCRS%2FFeatureServer%2F6%22%3A%7B%22x%22%3A30107%2C%22y%22%3A19661%2C%22level%22%3A15%7D%2C%22https%3A%2F%2Fstorage.googleapis.com%2Fterria-datasets-public%2Fbasemaps%2Fnatural-earth-tiles%2F%7Bz%7D%2F%7Bx%7D%2F%7BreverseY%7D.png%22%3A%7B%22x%22%3A117%2C%22y%22%3A76%2C%22level%22%3A7%7D%7D%2C%22pickCoords%22%3A%7B%22lat%22%3A-33.84944461506681%2C%22lng%22%3A150.77458130421803%2C%22height%22%3A-0.00913791767729486%7D%2C%22current%22%3A%7B%22name%22%3A5373%2C%22hash%22%3A52653753116%7D%2C%22entities%22%3A%5B%5D%7D%2C%22stories%22%3A%5B%5D%7D%5D%7D)

You get this error

image

image

File: lib/Map/ImageryProvider/ProtomapsImageryProvider.ts [hunk 1]
```diff
@@ -160,27 +164,6 @@ export default class ProtomapsImageryProvider
...
-    const tileCount =
-      (Math.abs(neTile.x - swTile.x) + 1) * (Math.abs(neTile.y - swTile.y) + 1);
-    if (tileCount > 4) {
-      throw new DeveloperError(
-        i18next.t("map.mapboxVectorTileImageryProvider.moreThanFourTiles", {
-          tileCount: tileCount
-        })
-      );
-    }
-
```
nf-s, line 183: No longer needed due to `softMinimumZoom`

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -384,89 +390,12 @@ export default class ProtomapsImageryProvider
...
-            f.geometry.coordinates[0],
-            f.geometry.coordinates[1]
-          );
-        }
-
-        featureInfo.configureDescriptionFromProperties(f.properties);
-        featureInfo.configureNameFromProperties(f.properties);
-
-        featureInfos.push(featureInfo);
-      });
```
nf-s, line 469: This has been moved into `lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts`

lib/ModelMixins/FeatureInfoUrlTemplateMixin.ts
```diff
@@ -104,7 +104,22 @@ function FeatureInfoUrlTemplateMixin<T extends AbstractConstructor<BaseType>>(
...
+        "0d"
+      );
+      const showStringIfPropertyValueIsNull = MappableMixin.isMixedInto(this)
+        ? this.showStringIfPropertyValueIsNull
+        : undefined;
+      const maxRequests = this.maxRequests;
+      const currentTime =
+        (TimeVarying.is(this) ? this.currentTimeAsJulianDate : undefined) ??
+        JulianDate.now();
+
```
nf-s: This fixes the "read outside reactive context" warnings we were getting with `featureInfoUrlTemplate`

lib/Table/TableColorMap.ts
```diff
@@ -139,11 +141,11 @@ export default class TableColorMap {
...
-      (!colorTraits.mapType ||
-        colorTraits.mapType === "continuous" ||
-        colorTraits.mapType === "bin") &&
-      colorColumn &&
-      colorColumn.type === TableColumnType.scalar
+      (colorColumn &&
+        !colorTraits.mapType &&
+        colorColumn.type === TableColumnType.scalar) ||
+      colorTraits.mapType === "continuous" ||
+      colorTraits.mapType === "bin"
```
nf-s, line 149: Loosen condition here, as we need to allow Discrete/Continuous color maps when colorColumn is undefined (eg empty)

lib/Table/TableColorMap.ts
```diff
@@ -199,12 +201,12 @@ export default class TableColorMap {
...
     else if (
-      colorColumn &&
-      ((!colorTraits.mapType &&
+      (colorColumn &&
+        !colorTraits.mapType &&
         (colorColumn.type === TableColumnType.enum ||
           colorColumn.type === TableColumnType.region) &&
         this.enumColors.length > 0) ||
-        colorTraits.mapType === "enum")
+      colorTraits.mapType === "enum"
```
nf-s, line 210: Same as above...

lib/Table/TableLegendStratum.ts
```diff
@@ -141,12 +141,12 @@ export class TableAutomaticLegendStratum extends LoadableStratum(
...
+    return (this._item.dataColumnMajor ?? []).length === 0;
+  }
+
   @computed get legends(): StratumFromTraits<LegendTraits>[] {
-    if (
-      this._item.mapItems.length === 0 ||
-      (this._item.dataColumnMajor ?? []).length === 0
-    )
-      return [];
+    if (this._item.mapItems.length === 0) return [];
```
nf-s, line 150: Instead of returning no legend, we now hide the legend - this is so we can override it in `ArcGisFeatureServerCatalogItem`

lib/ModelMixins/FeatureInfoUrlTemplateMixin.ts
```diff
@@ -104,7 +104,22 @@ function FeatureInfoUrlTemplateMixin<T extends AbstractConstructor<BaseType>>(
 
     wrapImageryPickFeatures<T extends ImageryProvider>(imageryProvider: T) {
       const realPickFeatures = imageryProvider.pickFeatures;
-      const catalogItem = this;
+
+      // Get CatalogItem properties, so we aren't reading them outside of a reaction
+      const featureInfoUrlTemplate = this.featureInfoUrlTemplate;
+      const proxiedFeatureInfoUrl = proxyCatalogItemUrl(
+        this,
+        this.featureInfoUrlTemplate ?? "",
```
pjonsson: The statement on the line before this statement defines `featureInofUrlTemplate`. Since that is already defined, is there a reason to not use that here?

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,50 +232,52 @@ export default class ProtomapsImageryProvider
...
+    return await this.requestImageForCanvas(x, y, level, canvas, request);
   }
 
   async requestImageForCanvas(
     x: number,
     y: number,
     level: number,
-    canvas: HTMLCanvasElement
+    canvas: HTMLCanvasElement,
+    request?: Request | undefined
```
pjonsson: What is the convention for ?-variables in TerriaMap, do the type need to include `| undefined`, or is the absence of the variable enough?

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,50 +232,52 @@ export default class ProtomapsImageryProvider
...
   ) {
-    try {
-      await this.renderTile({ x, y, z: level }, canvas);
-    } catch (e) {
-      console.log(e);
-    }
+    if (this.softMinimumLevel && level < this.softMinimumLevel)
+      throw TerriaError.from(`Level: ${level} is below softMinimumLevel`);
```
pjonsson: Can the value of `this.softMinimumLevel` be seen when this error is triggered, or should that be included in the error message?

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,50 +232,52 @@ export default class ProtomapsImageryProvider
...
+          coords.y * PROTOMAPS_DEFAULT_TILE_SIZE
+        ),
         dim: this.tileWidth
       };
     } else if (this.view) {
       tile = await this.view.getDisplayTile(coords);
     }
 
-    if (!tile) return;
+    if (!tile) throw TerriaError.from("Failed to get tile");
```
pjonsson: Is there a tile name that could be included, or is it obvious what tile that failed when this happens?

lib/Map/Vector/Protomaps/ProtomapsArcGisPbfSource.ts
```diff
@@ -0,0 +1,351 @@
...
+    this.maxTiledFeatures = options.maxTiledFeatures;
+    this.enablePickFeatures = options.enablePickFeatures;
+    this.supportsQuantization = options.supportsQuantization;
+  }
+
+  public async get(
+    c: Zxy,
+    tileSizePixels: number,
+    // TODO add support for canceling requests (through cesium)
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
pjonsson: Can this suppression be avoided by naming the parameter `_request?`?

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,285 @@
...
+  switch (type) {
+    case "Point":
+      return GeomType.Point;
+    case "LineString":
+      return GeomType.Line;
+    case "Polygon":
+      return GeomType.Polygon;
+    case "MultiPoint":
+      return GeomType.Point;
+    case "MultiLineString":
```
pjonsson: Are the cases sorted in some way? If not, it should be possible to group them so there's fewer return statements:
```TypeScript
case "Point":
case "MultiPoint":
  return GeomType.Point;
...
case "MultiLineString":
case "MultiPolygon":
  return GeomType.Line;
```

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,285 @@
...
+        minX: Infinity,
+        minY: Infinity,
+        maxX: -Infinity,
+        maxY: -Infinity
+      };
+
+      const geom = f.geometry;
+      transformedGeom = [
+        geom.map((g1) => {
+          g1 = [g1[0] * scale, g1[1] * scale];
```
pjonsson: Is the LHS `g1` conceptually a pair `(x, y)`?  I think it would be easier to read separate  `x` and `y` variables instead of `g1[0]` and `g1[1]`

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
nf-s: It tile is `undefined` then there are bigger problems - there is no appropriate source, or the `view` hasn't been setup. I have modified the message to reflect this

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,319 @@
...
+    return featureInfos;
+  }
+}
+
+export const geomTypeMap = (
+  type: string | null | undefined
+): GeomType | null => {
+  switch (type) {
+    case "Point":
+    case "MultiPoint":
```
pjonsson: Does the input for this function ever come from the code updated in #7432, or is that a separate code path?

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,319 @@
...
+        maxY: -Infinity
+      };
+
+      // Multi-polygon
+      if (Array.isArray(f.geometry[0][0])) {
+        // Note: the type is incorrect here
+        const geom = f.geometry as any as [number, number][][];
+        transformedGeom = geom.map((g1) =>
+          g1.map((g2) => {
+            const x = g2[0];
```
pjonsson: Is this function body identical to the function body in the else-branch on line 271?

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,319 @@
...
+            const y = g2[1];
```
pjonsson: Is `x` and `y` meant to not be scaled?

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,319 @@
...
+            g2 = [x * scale, y * scale];
```
pjonsson: I think this `g2` is no longer used.

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,319 @@
...
+        minX: Infinity,
+        minY: Infinity,
+        maxX: -Infinity,
+        maxY: -Infinity
+      };
+
+      // Multi-polygon
+      if (Array.isArray(f.geometry[0][0])) {
+        // Note: the type is incorrect here
+        const geom = f.geometry as any as [number, number][][];
```
pjonsson: There is a (disabled) ESLint rule in the recommended rules that complains about use of `any`, does `as unknown as [number, number][][]` work here?

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,319 @@
...
+export function geojsonVtTileToProtomapsFeatures(
+  features: geojsonvt.Feature[],
+  tileSize: number
+): ProtomapsFeature[] {
+  const scale = tileSize / GEOJSON_VT_EXTENT;
+
+  return features
+    .map((f) => {
+      let transformedGeom: Point[][] = [];
+      let numVertices = 0;
```
pjonsson: I don't think these two need to be initialized (looks like they are assigned in all code paths below).

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
nf-s: Good pickup, my brain wasn't working yesterday!

nf-s: Fixed

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
nf-s: Yes, but I think it is fine. IMO generally not worth refactoring duplication out unless there are more then 2 instances

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
nf-s: Strictly speaking no, but they make sense as default values, so i'm going to leave them there

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
nf-s: Most likely yes, but depends on a few other factors

nf-s: Just merged your PR in

pjonsson: I'm not sure if your yes means that the input comes from the other code or if it is a separate code path, but thank you for merging the PR.

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
```diff
@@ -0,0 +1,318 @@
...
+        numVertices = transformedGeom.length;
+      }
+
+      if (f.type === 0) return null;
+
+      const geomType = {
+        [1]: GeomType.Point,
+        [2]: GeomType.Line,
+        [3]: GeomType.Polygon
+      }[f.type];
```
pjonsson: I think commit e32e129e7bb addresses the same issue without resorting to comparing to the raw values in the enum.

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -351,18 +431,38 @@ class FeatureServerStratum extends LoadableStratum(
...
       if (uniqueValueRenderer.field2 || uniqueValueRenderer.field3) {
-        console.log(
+        console.warn(
           `WARNING: Terria only supports ArcGisFeatureService UniqueValueRenderers with a single field (\`field1\`), not multiple fields (\`field2\` or \`field3\`)`
         );
       }
 
+      if (uniqueValueRenderer.visualVariables?.length) {
+        console.warn(
+          `WARNING: Terria does not support visual variables in ArcGisFeatureService SimpleRenderers`
```
na9da: Small typo: should this say `UniqueValueRenderers` instead of `SimpleRenderers`? Same for classbreaks

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -315,6 +388,12 @@ class FeatureServerStratum extends LoadableStratum(
       const simpleRenderer = renderer as SimpleRenderer;
       const symbol = simpleRenderer.symbol;
```
na9da: Suggestion: is it worth splitting styling and associated functions into a separate Stratum and file?

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -151,10 +183,47 @@ interface FeatureServer {
...
     supportsPagination: boolean;
   };
+  supportedQueryFormats?: string;
+  maxRecordCount?: number;
+  tileMaxRecordCount?: number;
+  maxRecordCountFactor?: number;
+  supportsCoordinatesQuantization?: boolean;
+  supportsTilesAndBasicQueriesMode?: boolean;
+  objectIdField?: string;
+  fields?: Field[];
```
na9da: Like for the other types, is there a spec link for `FeatureServer` we can point to in comment?

lib/Map/Vector/Protomaps/ProtomapsArcGisPbfSource.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    const arcGisFeatures: Feature[] = [];
+
+    let offset = 0;
+    let fetching = true;
+
+    while (fetching) {
+      const tileResource = this.baseResource.getDerivedResource({
+        // Not sure how to handle request here - as we are making multiple requests
+        // request: request
```
na9da, line 98: Would it work if we clone the `request`? AFAICT, for imagery layers, it only sets some [flags](https://github.com/CesiumGS/cesium/blob/82e59ed00973bd3866d9d175914f7e38dee927dd/packages/engine/Source/Scene/ImageryLayer.js#L1164-L1168) at this stage in the code path. Cancellation might work even without passing `request` as it seems to create a[ new Request by default](https://github.com/CesiumGS/cesium/blob/82e59ed00973bd3866d9d175914f7e38dee927dd/packages/engine/Source/Core/Resource.js#L121).

lib/Map/ColorMap/ColorMap.ts
```diff
@@ -1,7 +1,6 @@
 import Color from "terriajs-cesium/Source/Core/Color";
+import JsonValue from "../../Core/Json";
 
 export default abstract class ColorMap {
-  abstract mapValueToColor(
-    value: string | number | null | undefined
-  ): Readonly<Color>;
+  abstract mapValueToColor(value: JsonValue | undefined): Readonly<Color>;
```
na9da, line 5: Doubt: is this change required? Most implementations of `ColorMap` assumes the param to be `string | ...`. Also the `JsonValue` [imported here](https://github.com/TerriaJS/terriajs/blob/a73f22e2f6c5cb2fd1da023c2dc3654c9d7f9a0e/lib/Map/Vector/Protomaps/tableStyleToProtomaps.ts#L6) comes from `protomap-leaflet`

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,69 +232,91 @@ export default class ProtomapsImageryProvider
     return [];
   }
 
-  async requestImage(x: number, y: number, level: number) {
+  async requestImage(x: number, y: number, level: number, request: Request) {
```
na9da: The `request` param should be optional right?

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,69 +232,91 @@ export default class ProtomapsImageryProvider
...
-    canvas: HTMLCanvasElement
+    canvas: HTMLCanvasElement,
+    request?: Request
   ) {
-    try {
-      await this.renderTile({ x, y, z: level }, canvas);
-    } catch (e) {
-      console.log(e);
-    }
+    if (this.softMinimumLevel && level < this.softMinimumLevel)
```
na9da: Nitpick: Maybe not an issue here, but is it better to test `this.softMinimumLevel !== undefined` as it makes the intent a bit more clear.

lib/ModelMixins/FeatureInfoUrlTemplateMixin.ts
```diff
@@ -156,12 +167,8 @@ function FeatureInfoUrlTemplateMixin<T extends AbstractConstructor<BaseType>>(
               // feature info template url
               feature.description = generateCesiumInfoHTMLFromProperties(
                 feature.properties,
-                (TimeVarying.is(catalogItem)
-                  ? catalogItem.currentTimeAsJulianDate
-                  : undefined) ?? JulianDate.now(),
-                MappableMixin.isMixedInto(catalogItem)
-                  ? catalogItem.showStringIfPropertyValueIsNull
-                  : undefined
+                currentTime,
```
na9da: Doubt: Should we do `currentTime ?? JulianDate.now()` instead so that it is always current and not the time when the closure was generated.

lib/ModelMixins/GeojsonMixin.ts
```diff
@@ -1435,6 +1315,53 @@ export function isMultiPoint(json: any): json is Feature<MultiPoint> {
...
+export function isLine(json: any): json is Feature<LineString> {
+  return (
+    isJsonObject(json, false) &&
+    json.type === "Feature" &&
+    isJsonObject(json.geometry, false) &&
+    json.geometry.type === "LineString"
+  );
+}
+
+export function isMultiLineString(json: any): json is Feature<MultiLineString> {
```
na9da: At some point maybe we should probably extract these functions to a new file. I have found them useful in other contexts.

lib/Table/TableColorMap.ts
na9da: Nitpick: is there a scope to make the intent more clear for the condition? One possibility might be to assign first to a temporarily variable like: `const isScalarColorColumn = colorColumn?.type === TableColumnType.scalar`.

lib/Models/Workflows/TableStylingWorkflow.ts
```diff
@@ -444,29 +445,42 @@ export default class TableStylingWorkflow
...
+    return !!this.item.mapItems.find(
+      (d) => isDataSource(d) && d.entities.values.length > 0
+    );
+  }
+
+  @computed get hasProtomapsImageryProvider() {
+    return !!this.item.mapItems.find(
+      (d) =>
+        "imageryProvider" in d &&
+        d.imageryProvider instanceof ProtomapsImageryProvider
```
na9da: Nitpick: There's also [ImageryParts.is()](https://github.com/TerriaJS/terriajs/blob/aabc08069787f606b223aad09a6983fa74b53544/lib/ModelMixins/MappableMixin.ts#L72) for this test

lib/Map/Vector/Protomaps/ProtomapsGeojsonSource.ts
na9da: In my testing that wasn't working. I think it is a subtle TS gotcha. `FeatureTypes` is an (ambient) `enum` defined in `.d.ts` file. No JS code is generated for it and so `geojsonvt.FeatureTypes`  will be `undefined` at run time. 

Had it been defined in a `.ts` file, it should have worked.

na9da (changes requested):
@nf-s - this is an awesome feature improvement. Great work on the implementation :tada: 

I have left some comments in the code review plus a couple of other observations:

- I noticed in some cases, the column values differ between protomap and geojson rendering modes.  This affects the symbology. Not sure if it is easily fixable though.
protomap:
image
geojson:
image

- Not something from this PR but I think the [second parameter here](https://github.com/TerriaJS/terriajs/blob/aabc08069787f606b223aad09a6983fa74b53544/lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts#L189-L190) should be `this._featureServer`. This for example affects the behaviour when splitting an item for compare.

na9da:
On the cesium error, I think it is happening when the `imageryProvider` gets destroyed while it is being loaded. This happens when the [trait change inside requestImage](https://github.com/TerriaJS/terriajs/blob/a73f22e2f6c5cb2fd1da023c2dc3654c9d7f9a0e/lib/ModelMixins/MinMaxLevelMixin.ts#L75-L85) triggers a chain of events that results in our Cesium.ts removing it from the `scene`.

One fix we could do at our end is to wrap these [setTrait calls](https://github.com/TerriaJS/terriajs/blob/a73f22e2f6c5cb2fd1da023c2dc3654c9d7f9a0e/lib/ModelMixins/MinMaxLevelMixin.ts#L75-L85) within `runLater` - which gives cesium a chance to finish what it is doing before the imagery provider gets destroyed.

lib/Table/TableColorMap.ts
nf-s: I agree it is confusing, I'm not sure adding variables like `isScalarColorColumn` will add clarity, as there will still need to be an explicit check that `colorColumn` is defined too. 

I added a comment above with:

```
TODO: Move all of these default values/behaviours to TableAutomaticStylesStratum - ideally no use of colorColumn is needed here
```

I think `TableColorMap` (amongst other `TableStyle`/`TableColumn` things) go against the pattern of using strata to drive default values. Ideally, `colorTraits.mapType` is set to an appropriate value in `TableAutomaticStylesStratum.styles`, but I thought that would be too far-reaching for this PR.

I just added another comment for later

```
TableAutomaticStylesStratum should also set default values for colorTraits.mapType, colors, etc
```

lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
nf-s: Yep good idea! Done

lib/Map/Vector/Protomaps/ProtomapsArcGisPbfSource.ts
nf-s: I tried this and got some weird behaviour - with requests being dropped. I'm not sure how cesium internals handle this, but I think it doesn't like multiple requests for a single tile.

lib/Map/ColorMap/ColorMap.ts
nf-s: I just opened the type up slightly, to make it easier to use in 

https://github.com/TerriaJS/terriajs/blob/b7e5b425c0296ba249c41079725c93d6ad7d2b98/lib/Map/Vector/Protomaps/tableStyleToProtomaps.ts#L140

nf-s:
> I noticed in some cases, the column values differ between protomap and geojson rendering modes. This affects the symbology. Not sure if it is easily fixable though.

This was due to how `TableColumn` was guessing the type from the values - it was the type to `scalar` which was stripping the comments. `FeatureServerStratum` now sets column types using the `esriFieldType`. This won't solve the general case (for all tabular, geojson data), but at least it fixes it here. I don't particularly want to tweak the `TableColumn` type guessing at this stage.

> On the cesium error, I think it is happening when the imageryProvider gets destroyed while it is being loaded. This happens when the [trait change inside requestImage](https://github.com/TerriaJS/terriajs/blob/a73f22e2f6c5cb2fd1da023c2dc3654c9d7f9a0e/lib/ModelMixins/MinMaxLevelMixin.ts#L75-L85) triggers a chain of events that results in our Cesium.ts removing it from the scene.
> 
> One fix we could do at our end is to wrap these [setTrait calls](https://github.com/TerriaJS/terriajs/blob/a73f22e2f6c5cb2fd1da023c2dc3654c9d7f9a0e/lib/ModelMixins/MinMaxLevelMixin.ts#L75-L85) within runLater - which gives cesium a chance to finish what it is doing before the imagery provider gets destroyed.

I tried adding `runLater`, it did fix the crashing, but the selected features being unselected after the share link loads. I'm also hesitant to add that change without further testing at this stage.

lib/Core/GeoJson.ts
```diff
@@ -0,0 +1,182 @@
...
+  return (
+    isJsonObject(json, false) && json.type === "Feature" && !!json.geometry
+  );
+}
+
+export function isPoint(json: any): json is Feature<Point> {
+  return (
+    isJsonObject(json, false) &&
+    json.type === "Feature" &&
+    isJsonObject(json.geometry, false) &&
```
pjonsson, line 60: These 3 lines are repeated in almost every function, is there a name for what they are checking so that check could be factored out to its own function?

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,69 +230,91 @@ export default class ProtomapsImageryProvider
...
-      maxX: 256 * (coords.x + 1) + PROTOMAPS_TILE_BUFFER,
-      maxY: 256 * (coords.y + 1) + PROTOMAPS_TILE_BUFFER
+      minX: PROTOMAPS_DEFAULT_TILE_SIZE * coords.x - PROTOMAPS_TILE_BUFFER,
+      minY: PROTOMAPS_DEFAULT_TILE_SIZE * coords.y - PROTOMAPS_TILE_BUFFER,
+      maxX:
+        PROTOMAPS_DEFAULT_TILE_SIZE * (coords.x + 1) + PROTOMAPS_TILE_BUFFER,
+      maxY: PROTOMAPS_DEFAULT_TILE_SIZE * (coords.y + 1) + PROTOMAPS_TILE_BUFFER
     };
-    const origin = new Point(256 * coords.x, 256 * coords.y);
+    const origin = new Point(
```
pjonsson, line 297: Can this definition be moved to before the `tile` definition on line 265 and used to define its origin as well?

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -248,69 +230,91 @@ export default class ProtomapsImageryProvider
...
     this.labelers.add(coords.z, tileMap);
 
     const labelData = this.labelers.getIndex(tile.z);
 
     const bbox = {
-      minX: 256 * coords.x - PROTOMAPS_TILE_BUFFER,
-      minY: 256 * coords.y - PROTOMAPS_TILE_BUFFER,
-      maxX: 256 * (coords.x + 1) + PROTOMAPS_TILE_BUFFER,
-      maxY: 256 * (coords.y + 1) + PROTOMAPS_TILE_BUFFER
+      minX: PROTOMAPS_DEFAULT_TILE_SIZE * coords.x - PROTOMAPS_TILE_BUFFER,
```
pjonsson, line 291: Is the RHS here `origin.x - PROTOMAPS_TILE_BUFFER`?

lib/Map/Vector/Protomaps/ProtomapsArcGisPbfSource.ts
```diff
@@ -0,0 +1,350 @@
...
+  };
+
+  let points: Position[][];
+
+  if (isPoint(feature)) {
+    points = [[feature.geometry.coordinates]];
+  } else if (isMultiLineString(feature)) {
+    points = feature.geometry.coordinates;
+  } else if (isPolygon(feature)) {
+    points = feature.geometry.coordinates;
```
pjonsson, line 275: Can this be a single branch with the condition `isMultilineString(feature) || isPolygon(feature)`?

lib/Map/Vector/Protomaps/ProtomapsArcGisPbfSource.ts
```diff
@@ -0,0 +1,350 @@
...
+      )
+    );
+  } else {
+    return [];
+  }
+
+  for (const g1 of points) {
+    const transformedG1: Point[] = [];
+    for (const g2 of g1) {
+      const transformedG2 = [
```
pjonsson, line 312: This code block seems familiar, are the components `x` and `y` this time as well?

lib/Map/Vector/Protomaps/tableStyleToProtomaps.ts
```diff
@@ -0,0 +1,262 @@
...
+  tableStyleMap: TableStyleMap<T>,
+  trait: Key,
+  defaultValue: Value,
+  useRowId = true,
+  transformValue?: (v: Value) => Value
+): (z: number, f?: ProtomapsFeature) => Value {
+  const styleMap = runInAction(() => tableStyleMap.styleMap);
+
+  if (isConstantStyleMap(styleMap)) {
+    return () => {
```
pjonsson, line 41: Why does TypeScript not flag this as a type error, isn't the return type a function that takes a number?

lib/Models/Catalog/Esri/esriStyleToTableStyle.ts
```diff
@@ -0,0 +1,169 @@
...
+    case "esriSMSX":
+      return "cross";
+    case "esriSMSCircle":
+    default:
+      return "point";
+  }
+}
+
+// ESRI uses points for styling while cesium uses pixels
+export function convertEsriPointSizeToPixels(pointSize?: number) {
```
pjonsson, line 91: Does this type signature permit forgetting the parameter, so the caller:
```typescript
const x = convertEsriPointSizeToPixels();
```
will pass the compiler without errors? Is that what you want?

lib/Models/Catalog/Esri/esriStyleToTableStyle.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (!isDefined(pointSize)) return undefined;
+  // 1 px = 0.75 point
+  // 1 point = 4/3 point
+  return (pointSize * 4) / 3;
+}
+
+function convertEsriColorToCesiumColor(
+  esriColor?: null | number[] | undefined
```
pjonsson, line 99: Same comment here about potentially forgetting the parameter without compiler errors.

lib/Table/TableStyleMap.ts
```diff
@@ -97,52 +101,53 @@ export default class TableStyleMap<T extends TableStyleMapSymbolTraits> {
...
       this.traitValues.bin.length > 0
     ) {
+      const mapValueToStyle: MapValueToStyle<T> = (value) => {
+        if (typeof value !== "number") {
+          return this.traitValues.null;
+        }
+
+        const binStyles = this.traitValues.bin ?? [];
+        let i;
+        for (
```
pjonsson, line 114: I think a one line comment explaining what the loop computes would be nice.

Not sure if the continue in the body adds anything.

nf-s:
At this stage, I’d like to keep this PR focused and avoid too many rounds of review. A lot of the changes here are just moving functions between files for better organization, and I generally try to keep those kinds of refactors minimal to reduce the risk of breaking things.

While your suggestions are valuable, I think we need to balance time and effort on TerriaJS. Right now, new features (like this one) and bug fixes are the bigger priorities over broader code improvements. Let’s keep this PR moving, and we can revisit larger refactors when the timing makes sense 🙂

pjonsson:
> While your suggestions are valuable, I think we need to balance time and effort on TerriaJS. Right now, new features (like this one) and bug fixes are the bigger priorities over broader code improvements.

TerriaJS is your code base, so you decide the priorities. I do feel that a PR with a diffstat of +4513-1363 viewed in the Github "Files changed" tab makes it difficult for us to know what parts of the code that have been changed and which parts that have just been moved.

The onToggle fix in this PR seemed to be missing the other place I introduced that bug, so I split out the complete onToggle fix to #7511 and referenced the commit where the problem was introduced.

lib/Map/Vector/Protomaps/tableStyleToProtomaps.ts
na9da: It is valid JS/TS to omit arguments.

lib/Models/Catalog/Esri/esriStyleToTableStyle.ts
na9da: this should be fine, `convertEsriPointSizeToPixels` can be called as `convertEsriPointSizeToPixels(10)` or `convertEsriPointSizeToPixels(undefined)` or `convertEsriPointSizeToPixels()`.

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
na9da: not sure why you think so.

lib/Map/Vector/Protomaps/ProtomapsArcGisPbfSource.ts
na9da: That to me is a pretty trivial difference. I would leave such decisions to the author unless it makes a huge performance or readability improvement.

na9da (approved):
Approving.

> I tried adding runLater, it did fix the crashing, but the selected features being unselected after the share link loads. I'm also hesitant to add that change without further testing at this stage.

For future reference:
`mapItems` is recomputed not because `scaleWorkbenchInfo` changed - there is no direct dependence of mapItems on `scaleWorkbenchInfo`. It is actually a side-effect of inserting a new `"defaults"` stratum which re-computes `strataTopToBottom` and then everything that depends on it (including `mapItems`). When that happens `Cesium.ts` destroys the layer - in the midst of a request, causing some inconsistency in Cesium's internal state.

One other fix might be to pre-emptly add the `"defaults"` stratum  in the [Mixin constructor](https://github.com/TerriaJS/terriajs/blob/ddb88d05b58ba6f80e0120a81b7e9e0a51f0c1c4/lib/ModelMixins/MinMaxLevelMixin.ts#L19). This seems to fix the disappearing feature info panel: 
```
  this.setTrait(CommonStrata.defaults, "scaleWorkbenchInfo", undefined);
```

PR #7371, opened 2024-12-13
Fix 7296 wmts

### LINKED ISSUE #7296: wrong WMTS GetTile requests for some WMTS servers
I'm trying to add tiles from [this WMTS link](https://wmts.marine.copernicus.eu/teroWmts/NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-3D_PT1H-m_202309?service=WMTS&request=GetCapabilities) to the terria instance at [nationalmap.gov.au](https://nationalmap.gov.au/) ([share link](https://nationalmap.gov.au/#share=s-dUOiVzKifX66tZYs0jsb3olxXyC))

While this allows me to see the list of available layers, adding a layer to the map does not actually display anything, and I see tile requests like
```
https://wmts.marine.copernicus.eu/teroWmts/NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-3D_PT1H-m_202309?tilematrix=3&layer=NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-3D_PT1H-m_202309/thetao&style=cmap:thermal&tilerow=2&tilecol=4&tilematrixset=EPSG:3857&format=image/png&service=WMTS&version=1.0.0&request=GetTile
```
failing with status code 400.

This server is a bit special in that in the response to `GetCapabilities` it requests using `https://wmts.marine.copernicus.eu/teroWmts` as the WMTS service endpoint for `GetTile`:
```
https://wmts.marine.copernicus.eu/teroWmts?tilematrix=3&layer=NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-3D_PT1H-m_202309/thetao&style=cmap:thermal&tilerow=2&tilecol=4&tilematrixset=EPSG:3857&format=image/png&service=WMTS&version=1.0.0&request=GetTile
```
works as expected. However, this seems to be supported by the WMTS standard, as libraries like GDAL or `owslib` produce the right requests (I didn't read the standard text, though, so I may be wrong on that).

Would it be possible to fix that? I unfortunately don't know enough about JavaScript / TypeScript to open a PR myself.

cc @rsignell-usgs, @mdsumner

<details><summary>demonstration using the <tt>owslib.wmts</tt> python library</summary>

```python
from owslib.wmts import WebMapTileService

url = "https://wmts.marine.copernicus.eu/teroWmts/NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-3D_PT1H-m_202309?service=WMTS&request=GetCapabilities"

wmts = WebMapTileService(url)
response = wmts.gettile(
    layer="NWSHELF_ANALYSISFORECAST_PHY_004_013/cmems_mod_nws_phy_anfc_0.027deg-3D_PT1H-m_202309/thetao",
    tilematrixset="EPSG:3857",
    tilematrix="5",
    row=9,
    column=14,
    format="image/png",
)
print(response.geturl())
with open("test.png", mode="wb") as f:
    f.write(response.read())
```

</details>

zoran995:
### What this PR does

Fixes #7296, #2927, #1573

Extended OWS common interfaces lists and added support for reading the URL from operations metadata.

### Test me

Try adding data using the init file https://gist.githubusercontent.com/zoran995/10169b47c7c7191f4795b56878c3d9c4/raw/2a27e2165b43f3e920dbebd181d4f6bd14ae4f4a/terriajs-7296-config.json
I have also tested using [New Zealand LINZ data service](https://data.linz.govt.nz/layer/95497-auckland-0075m-urban-aerial-photos-2017/webservices/) mentioned in #2927 

### Checklist


lib/Models/Catalog/Ows/OwsInterfaces.ts
```diff
@@ -6,6 +6,85 @@ export interface OnlineResource {
...
+export interface DomainType {
+  /** Name or identifier of this quantity. */
+  name: string;
+  /** List of all the valid values and/or ranges of values for this quantity. For numeric quantities, signed values should be ordered from negative infinity to positive infinity. */
+  readonly AllowedValues: AllowedValuesType;
+}
+
+export interface RequestMethodType extends OnlineResource {
+  /** Optional unordered list of valid domain constraints on non-parameter quantities that each apply to this request method for this operation. If one of these Constraint elements has the same "name" attribute as a Constraint element in the OperationsMetadata or Operation element, this Constraint element shall override the other one for this operation. The list of required and optional constraints for this request method for this operation shall be specified in the Implementation Specification for this service. */
+  readonly Constraint?: DomainType | DomainType[];
```
pjonsson, line 38: The PR consistently uses `ElemType | ElemType[]` in various places so it must be deliberate, but to me this feels like it's working against the type checker since it will be hard to statically determine if there is a single member or an array of members in an attribute. This seems confirmed by the code performing `x = Array.isArray(y) ? y : [y]` in a number of places.

I did a quick search and it seems possible to define a non-empty array type like this:
```typescript
type NonEmptyArray<T> = [T, ...T[]];

interface X {
  readonly Things?: NonEmptyArray<string>;
}
```
which I'm guessing should be able to represent the expected values?

If the intention is to be able to represent no constraint by not having the member, one constraint with the non-array type, >1 constraint with the array type, and the empty array signaling something different, that should be noted in some comment.

lib/Models/Catalog/Ows/OwsInterfaces.ts
```diff
@@ -6,6 +6,85 @@ export interface OnlineResource {
...
+  /** Connect point URLs for the HTTP Distributed Computing Platform (DCP). Normally, only one Get and/or one Post is included in this element. More than one Get and/or Post is allowed to support including alternative URLs for uses such as load balancing or backup. */
+  readonly HTTP: HTTPType;
+}
+
+interface OperationType {
+  /** Name or identifier of this operation (request) (for example, GetCapabilities). The list of required and optional operations implemented shall be specified in the Implementation Specification for this service. */
+  readonly name: string;
+
+  /** Optional unordered list of valid domain constraints on non-parameter quantities that each apply to this operation. If one of these Constraint elements has the same "name" attribute as a Constraint element in the OperationsMetadata element, this Constraint element shall override the other one for this operation. The list of required and optional constraints for this operation shall be specified in the Implementation Specification for this service. */
+  readonly Constraint?: DomainType | DomainType[];
```
pjonsson, line 58: The JSDoc describes this member as a "list of valid domain constraintS", but the member is  named `Constraint` (without the S). Would it make sense to rename the members that are in lists to have the plural s at the end of their names so it's clear from the name that it's (potentially) multiple things being dealt with?

lib/Models/Catalog/Ows/WebMapTileServiceCapabilities.ts
```diff
@@ -145,45 +128,55 @@ export default class WebMapTileServiceCapabilities {
           });
         }
 
-        return new WebMapTileServiceCapabilities(capabilitiesXml, json);
+        console.log(json);
```
pjonsson: There are a couple of `console.log()` calls, are they forgotten leftovers in this PR as well?

lib/Models/Catalog/Ows/WebMapTileServiceCatalogItem.ts
```diff
@@ -556,6 +537,59 @@ class WebMapTileServiceCatalogItem extends MappableMixin(
...
+      }
+    }
+
+    const resourceUrls: ResourceUrl[] | undefined =
+      !layer.ResourceURL || Array.isArray(layer.ResourceURL)
+        ? layer.ResourceURL
+        : [layer.ResourceURL];
+
+    if (resourceUrls && (this.requestEncoding === "RESTful" || !url)) {
+      for (let i = 0; i < resourceUrls.length; i++) {
```
pjonsson, line 586: Could `for (const resourceUrl of resourceUrls) {` be used to iterate over the content so the next line can be omitted, or is there some JavaScript-reason to always use index-based iterations in TypeScript?

Same comment for the other for-loops where the index is only used to get the element.

lib/Models/Catalog/Ows/OwsInterfaces.ts
zoran995: This is just how the XML conversion library works, as it doesn't have/use schema information. When only one element is available, it will convert it to a single element. When multiple elements are available, it will convert them to an array. So this way, we get proper typing for later usage. 
I don't like those Array.isArray checks, but I don't see any other way besides writing a custom post-processor, and I don't see enough benefits to complicate things

lib/Models/Catalog/Ows/OwsInterfaces.ts
zoran995: The property name and JSdoc are taken from the standard and XSD schema, so we can't change it

lib/Models/Catalog/Ows/OwsInterfaces.ts
pjonsson: Ouch. Well, knowing that, I agree that a custom post-processor isn't worth the effort.

moosepants:
<Edit: sorry I didn't realise this was an unreleased branch that potentially fixes the below behaviour, I will wait for the final push and retest when released>.

I'm not sure if this is related, or whether this is an adjacent issue, however when I add the NZ LINZ WMTS data (via https://data.linz.govt.nz/services;key=_removed_/wmts/1.0.0/WMTSCapabilities.xml) and add a layer, I don't see the tile data, instead a lot of `400 {Bad Requests}` from the tile server. 

After some investigation, the issue appears to be URL Encoding of the TileMatrixSet URL parameter. See below (noting EPSG%3A3857 vs EPSG:3857)

Terria does this (and doesn't work): https://tiles-cdn.koordinates.com/services;key=_removed_/tiles/v4/layer=50767,style%3Dauto/EPSG%3A3857/9/497/327.png
I amended the URL and this works: https://tiles-cdn.koordinates.com/services;key=_removed_/tiles/v4/layer=50767,style%3Dauto/EPSG:3857/8/255/162.png

zoran995:
Hi @moosepants, I just tried loading that dataset using the CI deployment http://ci.terria.io/fix-7296-wmts, but it is still throwing an error due to the encoded tilematrix set value. We are requesting data as 
https://tiles-cdn.koordinates.com/services;key=_removed_/tiles/v4/layer=50767,style%3Dauto/EPSG%3A3857/10/999/651.png
and services expects
https://tiles-cdn.koordinates.com/services;key=463cc81cc5ce40aca4696bcc255f53de/tiles/v4/layer=50767,style%3Dauto/EPSG:3857/10/999/651.png
The difference is only that ESPG:3857 is encoded to EPSG%3A3857, which is a standard method. It would be good to report this to the Koordinates team, as this is an issue on their side (mainly because it won't work even after merging this fix).

zoran995:
Hi @moosepants, I just tried loading that dataset using the CI deployment http://ci.terria.io/fix-7296-wmts, but it is still throwing an error due to the encoded tilematrix set value. We are requesting data as 
https://tiles-cdn.koordinates.com/services;key=_removed_/tiles/v4/layer=50767,style%3Dauto/EPSG%3A3857/10/999/651.png
and services expects
https://tiles-cdn.koordinates.com/services;key=_removed_/tiles/v4/layer=50767,style%3Dauto/EPSG:3857/10/999/651.png
The difference is only that ESPG:3857 is encoded to EPSG%3A3857, which is a standard method. It would be good to report this to the Koordinates team, as this is an issue on their side (mainly because it won't work even after merging this fix).

niktayv:
I've done some thorough testing both in the local environment and with the CI [link](http://ci.terria.io/fix-7296-wmts).
The examples provided in the description did not work for me.
To get a working WMTS server, I spun up a TiTiler server on localhost:8000, following https://developmentseed.org/titiler/user_guide/getting_started/. 
To serve the data, I used a GOG file on my machine, so that my URL looked like 
```
http://127.0.0.1:8000/WebMercatorQuad/WMTSCapabilities.xml?url=file:<path-to-GOG.tif> 
```
Currently, on the `main` branch, this layer fails to load tiles as the ResourceURL is ignored.
With the changes from PR, layer loads the tiles successfully.

niktayv:
I also submitted a ticket to Koordinates about the URL-Encoded TileMatrixSet issue, asking them either to adjust their backend or provide an alias so that Cesium could use the ResourceURL without tweaking.

rsignell:
Would love to see this merged -- this would fix my issues as well

na9da (changes requested):
Looks great @zoran995 - the spec is breaking for invalid tilematrix because previously we defaulted to googlemapscompatible. Should we keep the default?

zoran995:
@na9da, I'm not confident we should define a default tile matrix. WMTS capabilities should advertise all supported CRS/tile matrices, and if we set a default, we’ll likely end up sending requests with a tile matrix that's highly unlikely to work. 
I can restore the default, but in 99% of cases, it would be better to improve the actual parsing of tile matrices. The default acts as a glue to make it look like it works, and we don't know if we are missing something.

na9da (approved):

PR #7373, opened 2024-12-13
fix process error types when obtaining user location

### LINKED ISSUE #3601: Geolocate button is broken is Safari
Attempting to centre the map based on your current location (geolocate) in Safari throws the following error (which isn't super helpful for users):

Screen Shot 2019-07-18 at 10 07 10 am

**Steps to reproduce**
1. https://nationalmap.gov.au
2. Click geolocate button on the map

zoran995:
### What this PR does

Fixes #3601

Process and provide more useful messages to the user. In safari, there is nothing more we can do, as there are numerous bugs reported on the internet, and it is actually sometimes it works and sometimes it doesn't.

### Test me

### Checklist


lib/ReactViews/Map/MapNavigation/Items/MyLocation.ts
```diff
@@ -57,6 +57,9 @@ export class MyLocation extends MapNavigationItemController {
         timeout: 5000,
         maximumAge: 0
       };
+      navigator.permissions.query({ name: "geolocation" }).then((result) => {
+        console.log(result);
```
pjonsson: Does this log the result of each user location query to the console of the server? Aren't there potentially thousands of concurrent users?

zoran995: Forgot to push changes, thanks. It would just log in browser but it is not needed and wouldn't change anything :D

pjonsson: It seems I can't resolve this, but what I commented about seems fixed.

na9da (approved):
LGTM. Thanks @zoran995.

Custom error messages are an improvement over more technical browser messages.

PR #7392, opened 2024-12-18
replace clipboard with native browser support

### LINKED ISSUE #6774: Replace clipboard.js with native clipboard API
Clipboard.js is using a deprecated `document.execCommand` and is no longer recommended for use in new sites. 
It should be relatively easy to replace it with a native clipboard API which is for some time supported in all modern browsers. 
- replace usage of clipboard dependency with clipboard api
- remove clipboard.js from dependencies.

zoran995:
### What this PR does

Fixes #6774

Replace clipboard.js with native browser clipboard support. Browsers support native clipboard events for quite some time, while clipboard.js still uses old deprecated method, which might be removed at any point from browsers.
### Test me

Try copying the share URL in the app, it should properly copy the URL and show success for 3 seconds. I wanted to write unit tests for this, but we are using react-test-renderer, which is deprecated and does not offer convenient ways to test this. We need to see how to get the testing environment to a more modern stack. 

### Checklist

- [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - written above


lib/ReactViews/Clipboard.tsx
```diff
@@ -24,57 +23,51 @@ interface ClipboardProps {
...
+    } catch {
+      setStatus(CopyStatus.Error);
     }
-    function resetTooltipLater() {
-      removeTimeout();
-      timerId = setTimeout(() => {
+  };
+
+  useEffect(() => {
+    // Setup clipboard.js and show a tooltip on copy success or error for 3s
```
pjonsson: If you're removing clipboard.js, I think it makes sense to update the comments so clipboard.js isn't mentioned anywhere.

(There's another mention of clipboard.js on line 13 in this file.)

lib/ReactViews/Clipboard.tsx
```diff
@@ -24,57 +23,51 @@ interface ClipboardProps {
...
-      if (timerId !== null) {
-        clearTimeout(timerId);
-        timerId = null;
+
+  const handleCopy = async () => {
+    try {
+      if (text) {
+        await navigator.clipboard.writeText(text);
+        setStatus(CopyStatus.Success);
+        if (onCopy) onCopy(text);
```
pjonsson, line 36: If onCopy throws an exception, the status will go from `<initial state>` to `CopyStatus.Success` to `CopyStatus.Error` in a short while. Does it matter that some parts might observe the success and other parts might observe the error, or are all three of these state updates batched together so consistency is guaranteed?

test/ReactViews/ClipboardSpec.tsx
```diff
@@ -7,7 +7,7 @@ import Clipboard from "../../lib/ReactViews/Clipboard";
 import Input from "../../lib/Styled/Input";
 import Button from "../../lib/Styled/Button";
 
-describe("Clipboard", function () {
+xdescribe("Clipboard", function () {
```
pjonsson: Perhaps leave a comment here about the test failing because the previous clipboard implementation was replaced with native calls and the tests haven't been updated for that?

zoran995: I already started an internal discussion on the testing environment and will update this after we have actionable steps.

lib/ReactViews/Clipboard.tsx
zoran995: React should batch all those in a single state update

na9da:
@zoran995 - thanks for the improvement.

One thing I realized when testing from our [CI](http://ci.terria.io/native-clipboard) is that `navigator.clipboard` only works in a secure HTTPS context. I'm not sure this is a feature we currently want to break for users in (insecure) HTTP mode. So would it be possible to implement a fallback like the one in this [stackoverflow answer](https://stackoverflow.com/questions/72237719/not-being-able-to-copy-url-to-clipboard-without-adding-the-protocol-https)? We could still remove clipboard.js dependency.

zoran995:
The main idea behind this was not to depend on deprecated APIs such as `document.execCommand`. Perhaps we could implement a fallback with a deprecation notice to give users time to migrate to https, but these days https should be standard (browser probably would never remove such old feature until it blocks them in implementing something modern)

na9da:
I sort of agree.. also it seems to work in `localhost` where we use this feature quite a bit during development. 
I think if we go ahead with this, we should show the `Copy` button only when `navigator.clipboard` is available. Otherwise let the user manually copy it.

zoran995:
Hi @na9da, thank you for the suggestion. I have removed the copy button when the clipboard is unavailable, but would it be better to disable it instead of removing?
I have also added tests using `@testing-library/react`, we already talked about using it in future, and it will be needed for other components as part of the react upgrade

na9da:
@zoran995 - thanks for the changes - it looks great.

>  I have removed the copy button when the clipboard is unavailable, but would it be better to disable it instead of removing?

As discussed offline it might be better to hide the button as you need to provide extra details on why the button is disabled, plus the user might think the feature doesn't work - although they can manually copy using their system copy commands.

I have one other suggestion. In the story panel where the UI is `rounded` it now looks like below when the button is not shown. Ideally it should be rounded on both ends. One approach to fix this might be to remove the `border-radius` from these [two](https://github.com/TerriaJS/terriajs/blob/5745dbe1c0a020a9efadae58741cd1d6bc03097e/lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrl.tsx#L120-L122) [places](https://github.com/TerriaJS/terriajs/blob/5745dbe1c0a020a9efadae58741cd1d6bc03097e/lib/ReactViews/Clipboard.tsx#L73-L74) and instead apply it [on the Box](https://github.com/TerriaJS/terriajs/blob/5745dbe1c0a020a9efadae58741cd1d6bc03097e/lib/ReactViews/Clipboard.tsx#L67) and set `overflow: hidden`. So basically, rounded is applied on the container in one place instead of the children at two different places.
What do you think?

image

zoran995:
Thanks @na9da, good catch I completely forgot about the share URL in the side panel. I have added the border-radius to the parent div

lib/ReactViews/Clipboard.tsx
```diff
@@ -1,86 +1,77 @@
...
-  }, [id]);
+  }, [status]);
 
   const isLightTheme = theme === "light";
+  const canCopy = !!navigator.clipboard;
+
   return (
-    <ClipboardDiv>
+    <ClipboardDiv rounded={!!rounded}>
       <Box>
```
na9da: @zoran995 - thanks for making the change but I think `rounded` needs to be applied to the `<Box>` instead. Otherwise it will clip the status message and looks a bit odd:

image

zoran995: I have updated the PR. I was testing on CI and didn't realise I had forgotten to push the commits to both remotes, so they were out of sync. Later, I also realised that we can't use `overlay: hidden` as it will break the focus outline, so I adapted the solution to work without breaking it. This should be ready for final review

na9da (approved):
Looks good @zoran995. 

The latest change removes the `source` prop which slightly reduces the customizability of the component. I traced back the [history of Clipboard](https://github.com/TerriaJS/terriajs/commit/dd8866e092869e23566eef1cd62cfa827c9c4c0e) to see why it was written like that. but I couldn't find any solid reason. So I guess it is fine to hardcode `<Input>` inside the `Clipboard` component.

package.json
```diff
@@ -26,6 +26,9 @@
     "@mapbox/point-geometry": "^0.1.0",
     "@mapbox/togeojson": "^0.16.2",
     "@opendatasoft/api-client": "^0.1.0",
+    "@testing-library/jasmine-dom": "^1.3.3",
```
pjonsson, line 29: The testing-library dependencies only seem to be used in the tests, so they should be devDependencies I presume?

(I've been trying to move the jasmine type packages to a devDependency in #7489 so we don't increase the size of the TerriaMap image needlessly.)

test/ReactViews/ClipboardSpec.tsx
```diff
@@ -1,27 +1,86 @@
...
+      </ThemeProvider>
+    );
+
+    const button = screen.getByRole("button", { name: "clipboard.copy" });
+    await user.click(button);
+
+    expect(screen.getByText("clipboard.unsuccessful")).toBeVisible();
+    expect(screen.queryByText("clipboard.success")).not.toBeInTheDocument();
+
+    // act(() => {
```
pjonsson, line 77: Are these leftover debug statements that should be removed?

tsconfig.json
```diff
@@ -25,7 +25,7 @@
       "../../../node_modules/",
       "../../../node_modules/@types/"
     ],
-    "types": ["terriajs-cesium"]
+    "types": ["terriajs-cesium", "@testing-library/jasmine-dom"]
```
pjonsson, line 28: Why is this change needed? My understanding is that
```typescript
import JasmineDOM from "@testing-library/jasmine-dom";
```
will make tsc automatically bring in the corresponding types package.

package.json
zoran995: The cd is failing when we move the dependencies to dev, so will need to handle that separately

tsconfig.json
zoran995: Without this code editor would scream about missing types

package.json
pjonsson: Is that the same issue as https://github.com/TerriaJS/TerriaMap/issues/730, or a separate issue?

zoran995: It is not something caused by recent changes; it has actually existed for quite some time. The proper way to address this would be to have terriajs as a standalone package, built separately from TerriaMap, but I don't have a timeline for how long it will take.

pjonsson: I don't understand what you mean by a standalone package, because both terriajs and terriamap seem to be standalone packages already.

The CD failure sounds inconvenient though, is there a log of the failing CD somewhere?

pjonsson:
Running `yarn run tsc -b tsconfig.json` worked for me on main before this PR, but now I get:
```
node_modules/@types/testing-library__jasmine-dom/index.d.ts:327:66 - error TS2304: Cannot find name 'AsymmetricMatcher'.

327                 expectedAccessibleDescription: string | RegExp | AsymmetricMatcher<string>,
                                                                     ~~~~~~~~~~~~~~~~~

node_modules/@types/testing-library__jasmine-dom/index.d.ts:360:76 - error TS2304: Cannot find name 'AsymmetricMatcher'.

360             toHaveAccessibleName(expectedAccessibleName: string | RegExp | AsymmetricMatcher<string>): boolean;
                                                                               ~~~~~~~~~~~~~~~~~

node_modules/@types/testing-library__jasmine-dom/index.d.ts:400:61 - error TS2304: Cannot find name 'AsymmetricMatcher'.

400             toHaveErrorMessage(checkWith: string | RegExp | AsymmetricMatcher<string>): boolean;
                                                                ~~~~~~~~~~~~~~~~~


Found 3 errors.
```
The `AsymmetricMatcher` interface is a part of `@types/jasmine` 3.10.3.

Since Karma will be replaced soon, I tried bumping the types/jasmine package to the next major version as a temporary workaround, but that triggers the problems #7378 fixes.

On top of #7378 with types/jasmine bumped to 3.10.3, I can run `yarn run tsc -b tsconfig.json` again, and I don't need the additions in the `types` directive in `tsconfig.json` that was added in this PR.

package.json
pjonsson: @zoran995 I understand you are busy with other things. Is there anything I can do to help with the issue of your CD?

PR #7228, opened 2024-07-02
Refactor WPS datetime and date parameters.

na9da:
### What this PR does

Follow up to https://github.com/TerriaJS/terriajs/pull/7026, fixes https://github.com/TerriaJS/terriajs/issues/7200

- Slightly modifies the changes introduced in https://github.com/TerriaJS/terriajs/pull/7026 so that the default date/time values are used only if the parameter is required (i.e minOccurs=1)
- The default value will be equal to the timeline value when the WPS form is rendered
- Refactors the implementation of date/time parameters moving most of the validation and logic to model layer, this greatly simplifies the view components

### Test me

- Add the `Date Time WPS` catalog item from [share link](http://ci.terria.io/refactor-wps-datetime/#start=%7B%22version%22%3A%228.0.0%22%2C%22initSources%22%3A%5B%7B%22stratum%22%3A%22user%22%2C%22models%22%3A%7B%7D%2C%22workbench%22%3A%5B%5D%2C%22timeline%22%3A%5B%5D%2C%22initialCamera%22%3A%7B%22west%22%3A102.52441406250001%2C%22south%22%3A-44.84029065139799%2C%22east%22%3A164.48730468750003%2C%22north%22%3A-8.276727101164033%7D%2C%22homeCamera%22%3A%7B%22west%22%3A109%2C%22south%22%3A-45%2C%22east%22%3A158%2C%22north%22%3A-8%7D%2C%22viewerMode%22%3A%222d%22%2C%22showSplitter%22%3Afalse%2C%22splitPosition%22%3A0.4999%2C%22settings%22%3A%7B%22baseMaximumScreenSpaceError%22%3A1%2C%22useNativeResolution%22%3Afalse%2C%22alwaysShowTimeline%22%3Atrue%2C%22baseMapId%22%3A%22basemap-bing-aerial-with-labels%22%2C%22terrainSplitDirection%22%3A0%2C%22depthTestAgainstTerrainEnabled%22%3Afalse%7D%2C%22stories%22%3A%5B%5D%7D%5D%7D&https://gist.githubusercontent.com/na9da/672301fb7f8cb4e6edec658dadc83d05/raw/e40a64080538a0f3a3670e053adf02d02ecfcb5b/wps-catalog.json) for this branch and test the date field - setting new date, clearing etc. Note that submitting won't work because the catalog item points to a static gist ProcessDescription XML.

image

### Checklist

[...]

PR #7396, opened 2024-12-19
explicitly type useCallback arguments

zoran995:
### What this PR does

This is required in a newer versions of react

### Test me

Type only change; there is no benefit of `useCallback` here as it is passed directly to the input component there so we don't need to keep same reference always

### Checklist

### DIFF
--- a/lib/ReactViews/Analytics/DateParameterEditor.tsx
+++ b/lib/ReactViews/Analytics/DateParameterEditor.tsx
@@ -1,5 +1,5 @@
 import { observer } from "mobx-react";
-import React, { useCallback } from "react";
+import React from "react";
 import CommonStrata from "../../Models/Definition/CommonStrata";
 import DateParameter from "../../Models/FunctionParameters/DateParameter";
 import Styles from "./parameter-editors.scss";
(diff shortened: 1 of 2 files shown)


lib/ReactViews/Analytics/DateParameterEditor.tsx
```diff
@@ -15,18 +15,15 @@ const DateParameterEditor: React.FC<DateParameterEditorProps> = observer(
         ? Styles.field
         : Styles.fieldDatePlaceholder;
 
-    const onChangeDate = useCallback(
```
pjonsson, line 18: What is the type error you are getting for the useCallback, and is the error when using React 18?

Do you think a type annotation in the right place could make the type checker happy? I looked in the React 19 documentation and that seems to still support `useCallback`, and considering this code is from July 2024 (#7228), I'm guessing the decision to use useCallback was made deliberately.

zoran995: There is no type error. It is just better to have it explicitly typed, and this is from the react codemod. There is definitely no reason to have `useCallback` for a function passed to the onClick event. I used to do that in the past, but the only real reason to memoize functions is when passing to use effect or a child component is memoized or building a generic hook so the users don't need to worry about rerendering.

na9da (changes requested):
Have tested the change and it works without the callback.

It might make sense to make a similar change here https://github.com/TerriaJS/terriajs/blob/7bf2f921526b287f9dec2dbf5c865f1a97bcbb70/lib/ReactViews/Analytics/DateTimeParameterEditor.tsx#L18-L22

na9da:
[Here's a catalog](https://gist.githubusercontent.com/na9da/576e1d9ff82994a68d9f4458018b23ec/raw/b617b0d8c292bb6f4f22b3b18db3080b7526c232/wps-datetime.json) for testing date and datetime components.

zoran995:
Thanks, @na9da. I removed that one also; I didn't notice it because params were explicitly typed in useCallback, and a new version of react typing throws an error only when not typed

ljowen (approved):

na9da (approved):

PR #7398, opened 2024-12-20
import TFunction type from react-i18next for proper typing

zoran995:
### What this PR does

`TFunction` type should be imported from `react-i18next` instead of `i18next` directly. It will cause issues in new versions of react

### Test me

Those are typeonly changes so it tested by CI

### Checklist


na9da (approved):
LGTM.

Issue #5169, opened 2021-01-22
Styled components remove the ability to finely customise appearance


tephenavies:
Currently we support customising the CSS of specific elements in `globals.scss` of a TerriaMap:
https://github.com/TerriaJS/TerriaMap/blob/100f4f285a880afdf76330905a75dc286269bda4/lib/Views/global.scss#L17-L154

For example AREMI uses this:
https://github.com/TerriaJS/aremi-natmap/blob/3f8c7f4bcf841d60da843e17bda4d440a1893ecd/lib/Views/global.scss#L16-L69

This support will slowly be eroded as we convert components from using CSS modules and Sass to styled-components and they lose their `tjs-...` class names.

We have a couple of options:
* Acknowledge this loss of customisability and add options to styled components theming system to compensate
* Add class names to styled components. If we want to use this for our managed maps we'll have to work out a way to load custom per-map css

tephenavies:
E.g. https://github.com/TerriaJS/terriajs/issues/5814

zoran995:
> * Add class names to styled components. If we want to use this for our managed maps we'll have to work out a way to load custom per-map css

If going this way we need to pay attention to [issues with specificity](https://styled-components.com/docs/advanced#issues-with-specificity).

KeyboardSounds:
I vote for
> Acknowledge this loss of customisability and add options to styled components theming system to compensate

I think the actual real-world impact of this is hopefully minor, and we should be able to provide new themeing options that cover the most common cases

irbian:
We are affected from this

Related, but different, there is a lot of 100vh, IMHO, unneeded, that breaks the possibility of adding elements outside Terria-ui

PR #7443, opened 2025-01-13
UI update

ljowen:
### What this PR does

Major change to default theme and appearance:
May introduce breaking changes with existing maps so will be a major version.

#### Visual changes ####
- Generally increase padding
- Workbench now "floating" with variable height and semi transparent background
- Change to darker colours in default theme to increase contrast
- Font changed from "Nunito" to "Inter", generally increase font sizes
-  More rounded corners, generally decrease border radius to 8px for major panels and 4px for smaller components

**Before**
before
**After**
after

#### Functional changes ####
- Bug fixed where tour would not work correctly if workbench was minimised
- Add config option to keep catalog open when adding / removing datasets `keepCatalogOpen`
- Remove dataset count shown on map
- Hide chart at mobile viewports

### Test me

Before:  http://ci.terria.io/main
After: http://ci.terria.io/ui-update-fonts

### Checklist


lib/ReactViews/SidePanel/FullScreenButton.jsx
```diff
@@ -43,57 +46,73 @@ class FullScreenButton extends React.Component {
...
   renderButtonText() {
     const btnText = this.props.btnText ? this.props.btnText : null;
-    if (this.props.minified) {
-      if (this.props.viewState.isMapFullScreen) {
-        return <Icon glyph={Icon.GLYPHS.right} />;
-      } else {
-        return <Icon glyph={Icon.GLYPHS.closeLight} />;
-      }
+    if (this.props.viewState.isMapFullScreen) {
+      return <span>{btnText}</span>;
```
zoran995, line 50: Do we still need this with the new UI? It will show the text in UI for a brief moment

zoran995:
Quick UI check
1. it is not possible to close the chart panel the close button overlaps with navigation items (15inch MacBook)
image

2. scroll in item search tool is broken, I believe this was fixed in #7290
image

3. story editor capture scene button is not visible 
image

4. Pedestrian tool tooltips overlap with side panel (maybe this is desired). It also overlaps the timeline title
image

5. The gap to about button is large 
image

6. Difference tool overlaps with collapsed workbench header ([test url](http://ci.terria.io/ui-update-fonts/#https://map.terria.io/proxy/https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/master/prod/terria/dea-maps-v8.json))

https://github.com/user-attachments/assets/e8d7575f-4337-4f37-9135-be9f51717553

lib/ReactViews/ExplorerWindow/Tabs.jsx
```diff
@@ -187,18 +185,22 @@ class Tabs extends React.Component {
...
       background: ${props.theme.textLight};
-      color: ${props.theme.colorPrimary};
+      ${props.isCurrent && `border: 1px solid ${props.theme.greyLighter};`}
     }
     ${
       props.isCurrent &&
       `
       background: ${props.theme.textLight};
-      color: ${props.theme.colorPrimary};
+      border: 1px solid ${props.theme.greyLighter};
```
zoran995, line 204: It might be good to use some of the border/outline tricks to avoid the button getting larger/smaller due to the new border applied.

ljowen: fixed thanks

lib/ReactViews/SidePanel/FullScreenButton.jsx
ljowen: We do not

ljowen:
Thanks for the review @zoran995  I've addressed most of these now except for 6), I'm in conversation with our designer about reverting to the small "Show workbench button" which would be my preference

irbian:
Our client love the new changes! there is any ETA of them?

demo.patch
pjonsson, line 1: Not sure this file should be in the PR that gets merged?

demo2.patch
pjonsson, line 1: Not sure this file should be in the PR that gets merged?

lib/Models/Terria.ts
```diff
@@ -373,6 +373,11 @@ export interface ConfigParameters {
    */
   searchBarConfig?: ModelPropertiesFromTraits<SearchBarTraits>;
   searchProviders: ModelPropertiesFromTraits<SearchProviderTraits>[];
+
+  /**
+   * Keep catalog open when adding / removing items
+   */
+  keepCatalogOpen: false;
```
pjonsson: I'm guessing this should have type `boolean`?

lib/ReactViews/Map/MapNavigation/Items/Compass/Compass.tsx
```diff
@@ -346,7 +346,7 @@ class Compass extends React.PureComponent<PropTypes, IStateTypes> {
               // if it's active, show a white circle only, as we need the base layer
               glyph={
                 active
-                  ? Icon.GLYPHS.compassOuterSkeleton
+                  ? null //Icon.GLYPHS.compassOuterSkeleton
```
pjonsson: Accidental change?

If not, remove the old code, and update the comment on line 346.

lib/ReactViews/MapIconButton/MapIconButton.tsx
```diff
@@ -1,3 +1,4 @@
+"use strict";
```
pjonsson: Is this needed for TypeScript files?

lib/ReactViews/Preview/DataPreview.jsx
```diff
@@ -172,4 +175,4 @@ class DataPreview extends React.Component {
   }
 }
 
-export default withTranslation()(DataPreview);
+module.exports = withTranslation()(DataPreview);
```
pjonsson: Accidental change?

lib/ReactViews/Preview/MappablePreview.jsx
```diff
@@ -39,12 +39,16 @@ class MappablePreview extends React.Component {
     t: PropTypes.func.isRequired
   };
 
-  async toggleOnMap(event) {
+  async toggleOnMap(_event) {
```
pjonsson: Line 49/50 use `event`, what outer scope event is picked up by that after this change?

lib/ReactViews/Preview/MappablePreview.jsx
```diff
@@ -39,12 +39,16 @@ class MappablePreview extends React.Component {
...
     if (defined(this.props.viewState.storyShown)) {
       runInAction(() => (this.props.viewState.storyShown = false));
     }
 
-    const keepCatalogOpen = event.shiftKey || event.ctrlKey;
+    const keepCatalogOpen =
+      this.props.viewState.terria.configParameters.keepCatalogOpen ||
+      event.shiftKey ||
+      event.ctrlKey;
+    event.shiftKey || event.ctrlKey;
```
pjonsson: Is this line an expression without any side effects? Should probably be removed in that case?

lib/ReactViews/SelectableDimensions/Group.tsx
```diff
@@ -48,13 +49,11 @@ export const SelectableDimensionGroup: React.FC<{
         onToggle={
           dim.type === "group"
             ? dim.onToggle
-            : (isOpen) => {
+            : (isOpen) =>
```
pjonsson: The change from commit c8b8bf5 is reverted by this change, is that intentional?

lib/ReactViews/SidePanel/Branding.tsx
```diff
@@ -1,9 +1,11 @@
+"use strict";
```
pjonsson: Another "use strict" in a TypeScript file, intentional?

lib/ReactViews/Workbench/Controls/OpacitySection.tsx
```diff
@@ -1,10 +1,12 @@
+"use strict";
```
pjonsson: One more "use strict" in TypeScript file, intentional?

lib/ReactViews/Workbench/WorkbenchButton.tsx
```diff
@@ -1,3 +1,4 @@
+"use strict";
```
pjonsson: Another "use strict" in TypeScript file.

lib/Styled/Button.tsx
```diff
@@ -75,7 +75,11 @@ export const StyledButton = styled.button<IStyledButtonProps>`
     color: #fff;
     background-color: ${props.theme.colorPrimary};
     border: none;
-    border-radius:20px;
+    border-radius:4px;
+        &:hover,
+    &:focus {
+
```
pjonsson: Previous lines in this file put something inside `{ }`, is the intention to have an empty "body" here?

lib/Styled/Checkbox/types.ts
```diff
@@ -50,6 +52,9 @@ export type ICheckboxProps = {
 };
 
 export interface CheckboxIconProps {
+  /* Override svg icon to use switch icon */
+  isSwitch?: boolean;
+  iconIndeterminate?: any;
```
pjonsson: Is this used? Can it get a more precise type than `any`?

wwwroot/images/icons/compass-inner-arrows.svg
pjonsson, line 1: For this file, I get:
```
Unable to render rich display

Invalid image source.
```
in Firefox. I see a preview of the other svg icons.

lib/ReactViews/SelectableDimensions/Group.tsx
ljowen: It is not, thanks!

lib/Styled/Checkbox/types.ts
ljowen: this isnt actually used must've been a merge error. removed

wwwroot/images/icons/compass-inner-arrows.svg
ljowen: Whoops looks like I doubled up on the fill attribute

lib/ReactViews/Map/MapNavigation/Items/Compass/Compass.tsx
ljowen: I've updated the comment, need to revisit compass active state as it's a bit broken at the moment (pre this PR)

lib/ReactViews/SidePanel/FullScreenButton.jsx
```diff
@@ -1,13 +1,16 @@
-import React from "react";
-import PropTypes from "prop-types";
+"use strict";
+const React = require("react");
+const PropTypes = require("prop-types");
```
pjonsson: These import changes looks unintentional.

lib/ReactViewModels/ViewState.ts
```diff
@@ -283,6 +283,8 @@ export default class ViewState {
     // If we're enabling the tour, make sure the trainer is collapsed
     if (bool) {
       this.setTrainerBarExpanded(false);
+      // Ensure workbench is shown
+      this.setIsMapFullScreen(false);
```
zoran995, line 286: This may require additional handling; the animation is set to 500ms by default, and if it is rendered before, it will catch the wrong position of element
image

p.s. this also exists on main so fine to create a ticket and handle it separately not to block this

lib/ReactViews/BadgeBar.tsx
zoran995, line 1: Badge bar is also used in storybuilder and those changes will break styling over there. Might be good to decouple Storybuilder from the badge bar (even moving badge bar next to the workbench is an option)

lib/ReactViews/ExplorerWindow/ExplorerWindow.tsx
```diff
@@ -37,6 +41,18 @@ export default observer<React.FC>(function ExplorerWindow() {
       onStartAnimatingIn={onStartAnimatingIn}
       onDoneAnimatingIn={onDoneAnimatingIn}
     >
+      <Box
+        css={`
+          background: ${theme.dark};
+          border-radius: 6px 6px 0 0;
```
zoran995: Can you also add border-radius to the parent element so the shadow doesn't create an effect a of no border-radius

lib/ReactViews/Map/BottomLeftBar/BottomLeftBar.tsx
```diff
@@ -82,30 +47,6 @@ const BottomLeftBar: FC = observer(() => {
...
-          {screenDataAttributions
-            .flatMap((attributionHtml, i) => [
-              <span key={attributionHtml}>
-                {parseCustomHtmlToReact(attributionHtml)}
-              </span>,
-              <span key={`delimiter-${i}`}> • </span>
-            ])
-            .slice(0, -1)}
-        </AttributionsContainer>
-      )}
```
zoran995: are we sure that we can remove this? It will also hide a warning for using the default ion token

lib/ReactViews/Map/MapNavigation/Items/ZoomControl.tsx
```diff
@@ -218,17 +218,18 @@ class ZoomControlBase extends React.Component<PropTypes> {
...
 const StyledZoomControl = styled(Box).attrs((props) => ({
-  backgroundColor: props.theme.textLight,
+  backgroundColor: props.theme.dark,
   centered: true,
   column: true,
   styledWidth: "32px",
   styledMargin: "7px 0 0 0"
 }))`
   border-radius: 100px;
+  border: 1px solid ${(props) => props.theme.darkLighter};
```
zoran995, line 228: Do we need to add a border to other menu items?

lib/ReactViews/Workbench/Workbench.tsx
```diff
@@ -60,43 +69,80 @@ class Workbench extends React.Component<IProps> {
   render() {
     const { t } = this.props;
     const shouldExpandAll = this.props.terria.workbench.shouldExpandAll;
+
+    // show enable all button if some items are disabled
+    const showEnableAll = this.props.terria.workbench.items
+      .filter((it): it is MappableMixin.Instance =>
+        MappableMixin.isMixedInto(it)
+      )
+      .every((it) => !it.show);
```
zoran995: `enable all` is visible only when all items are disabled, can you update the comment or the implementation

zoran995 (dismissed):
when there is only one story then it the menu won't be visible
image
with more it is visible 
image

lib/ReactViewModels/ViewState.ts
ljowen: There's other instances in `ViewState` where a timeout is set to the animation length so I've followed that pattern

ljowen:
> when there is only one story then it the menu won't be visible image with more it is visible image



> when there is only one story then it the menu won't be visible image with more it is visible image

We need to rework this custom dropdowns as they don't work nicely inside scroll containers (the clipped by first "overflow: auto" parent), in the meantime I've just added a min-height which will give it room when there's only 1 item (note: this is also affecting `main`)

lib/ReactViews/BadgeBar.tsx
ljowen: What's broken? This is what I see
Screenshot 2025-02-04 at 5 17 40 pm

(I'll check the change with our designer)

zoran995: To me, it seems a bit odd to have only `remove all` in the second row, but perhaps that is intentional. When viewed only in the context of the storybuilder, it looks nice; however, in the context of the entire screen, it feels a bit off. It might just be me; let's see what the designer says.

ljowen: I've decoupled it and made them inline, I agree it was a waste of vertical space

lib/ReactViews/Map/BottomLeftBar/BottomLeftBar.tsx
ljowen: Added the attributions back in, thanks for catching this!

lib/ReactViews/Map/MapNavigation/Items/ZoomControl.tsx
ljowen: yes!

test/ReactViewModels/ViewStateSpec.ts
```diff
@@ -114,9 +117,13 @@ describe("ViewState", function () {
       runInAction(() => {
         viewState.setShowTour(true);
       });
+
+      jasmine.clock().tick(animationDuration); // wait for workbench animation
+
       expect(viewState.trainerBarExpanded).toEqual(false);
       expect(viewState.trainerBarShowingAllSteps).toEqual(false);
       expect(viewState.showTour).toEqual(true);
+      jasmine.clock().uninstall();
```
pjonsson: Will the clock uninstall properly if this test crashes half-way through, or will the mock clock remain and cause cascading failures for the tests following this test?

Same comment for the change that adds a similar clock in a TourPortalSpec test.

ljowen: hmm I just tested it and yeah looks like it can affect future tests, I've moved the setup and teardown into `beforeEach` `afterEach` respectively

AmberStandley83:
.

ljowen:
> Our client love the new changes! there is any ETA of them?

Hi @irbian , we aim to get these changes released by the end of this month

zoran995 (approved):
LGTM, I performed another review of the code and UI, and everything seems fine. 
It might be good to have someone else do one more quick review of the code, but it is your call, @ljowen. Maybe it is better to have this merged and further test it on the main before the release

lib/ReactViewModels/ViewState.ts
```diff
@@ -279,10 +279,16 @@ export default class ViewState {
...
   @action
   setShowTour(bool: boolean): void {
-    this.showTour = bool;
     // If we're enabling the tour, make sure the trainer is collapsed
     if (bool) {
       this.setTrainerBarExpanded(false);
+      // Ensure workbench is shown
+      this.setIsMapFullScreen(false);
+      setTimeout(() => {
+        this.showTour = bool;
```
na9da: I think this needs to be wrapped in `runInAction` as we are mutating an `observable`.

lib/ReactViews/BottomDock/Timeline/Timeline.jsx
```diff
@@ -89,6 +89,7 @@ class Timeline extends React.Component {
           className={Styles.textRow}
           css={`
             background: ${(p) => p.theme.dark};
+            min-height: 30px;
```
na9da, line 92: I think this still leaves the timeline item name blocked by the timeline controls and track.

image

Maybe we should increase this to 50px or something. But then we'll need to adjust the story panel (etc.) accordingly.

lib/ReactViews/BottomDock/Timeline/timeline.scss
```diff
@@ -7,26 +7,23 @@
   padding-top: 5px;
   padding-bottom: 5px;
   padding-right: 5px;
+  margin-top: 30px;
```
na9da: Doubt: Is this really required? If so it might be good to document why.

Personally, I feel this adds a bit too much gap between timeline and chart:
image

lib/ReactViews/BottomDock/BottomDock.tsx
```diff
@@ -53,12 +54,18 @@ class BottomDock extends React.Component<PropsType & MeasureElementProps> {
         tabIndex={0}
         onClick={this.handleClick.bind(this)}
         css={`
-          background: ${(p: any) => p.theme.dark};
+          z-index: 10;
+          bottom: 0;
+          right: 0;
```
na9da: this probably needs a `position: relative` as well to take effect - which also seems to fix the gap between timeline issue.

also, can we now remove `bottom-dock.scss`?

lib/ReactViews/BottomDock/BottomDock.tsx
```diff
@@ -42,7 +43,7 @@ class BottomDock extends React.Component<PropsType & MeasureElementProps> {
 
     return (
       <div
-        className={`${Styles.bottomDock} ${
+        className={`${
           this.props.viewState.topElement === "BottomDock" ? "top-element" : ""
```
na9da, line 47: Question: Do we need this change? One issue I noticed is that, previously if chart and story panel overlapped, then it was possible to click one or the other to bring it forward. But now, because bottom dock as whole gets stacked on top, the story panel does not receive the click event to make it the top-element. 

Our UI stacking as a whole needs a re-think.

lib/ReactViews/DataCatalog/data-catalog-group.scss
na9da, line 1: I think we need to consider the alignment of the icon for private datasets. Not sure exactly what change caused the mis-alignment.

before:
image


after:
image

lib/ReactViews/ExplorerWindow/ExplorerWindow.tsx
```diff
@@ -37,6 +41,18 @@ export default observer<React.FC>(function ExplorerWindow() {
...
+      <Box
+        css={`
+          background: ${theme.dark};
+          border-radius: ${theme.radiusXL} ${theme.radiusXL} 0 0;
+        `}
+        paddedRatio={4}
+        fullWidth
+      >
+        <Text extraExtraLarge textLight css={{ fontWeight: 500 }}>
+          {t("addData.dataCatalogue")}
```
na9da, line 55: nitpick: I could be wrong, but I think in most cases it is better to use `const {t} = useTranslation()` instead of global `t`. I think it is because the global `t` may not react correctly to change in translation settings (?).

lib/ReactViews/BottomDock/BottomDock.tsx
na9da: There is also an issue with switching between chart and feature info panel - not sure why.

In [main](http://ci.terria.io/main/#share=s-cQrRi6tq0NFSZ2fuLkTtHSUMICd), I can switch between feature info and chart by clicking their title bar. That doesn't seem to work for[ this branch](http://ci.terria.io/ui-update-fonts/#share=s-cQrRi6tq0NFSZ2fuLkTtHSUMICd). The chart is always in the background.

lib/ReactViews/Map/BottomLeftBar/BottomLeftBar.tsx
```diff
@@ -63,12 +53,7 @@ const BottomLeftBar: FC = observer(() => {
       );
 
   return (
-    <BottomLeftContainer theme={theme}>
-      <MapDataCount
-        terria={viewState.terria}
-        viewState={viewState}
-        elementConfig={viewState.terria.elements.get("map-data-count")}
-      />
+    <Box theme={theme} padded>
```
na9da: nitpick: `theme` is not required for `Box`.

lib/ReactViews/Map/MapNavigation/registerMapNavigations.tsx
```diff
@@ -198,7 +198,7 @@ export const registerMapNavigations = (viewState: ViewState) => {
     id: FEEDBACK_TOOL_ID,
     name: "translate#feedback.feedbackBtnText",
     title: "translate#feedback.feedbackBtnText",
-    location: "BOTTOM",
+    location: "TOP",
```
na9da: I don't have a strong preference here, but when other icons get added like from some plugin, the feedback icon can become harder to spot.

image

lib/ReactViews/Map/TerriaViewerWrapper/Splitter/dragHook.ts
na9da, line 1: Something seems to have broken splitter in mobile mode - not sure what.

lib/ReactViews/Map/Toast.tsx
```diff
@@ -19,8 +19,9 @@ const Container = styled.div`
   left: 35%;
   @media (min-width: ${(props) => props.theme.sm}px) {
     position: absolute;
-    left: 45%; // on larger screens
-    bottom: 40px;
+    left: 50%;
+    transform: translate(-50%, 0);
```
na9da: Is this required? This seems to push the toast slightly to the left of center.

lib/ReactViews/MapIconButton/MapIconButton.tsx
```diff
@@ -41,13 +42,14 @@ const StyledMapIconButton = styled(RawButton)<IStyledMapIconButtonProps>`
     width: 20px;
     margin: 0 auto;
     vertical-align: middle;
-    fill: ${(props) => props.theme.textDarker};
+    fill: #9ca3af;
```
na9da: Doubt: Is this intentionally hardcoded?

lib/Sass/common/_variables.scss
```diff
@@ -1,28 +1,35 @@
...
+$color-secondary: #f59e0b;
 
 $animation-fast: 0.1s;
 
-$dark: #3f4854;
-$dark-with-overlay: #667080;
-$dark-lighter: #595b60;
-$grey: #888888;
-$grey-lighter: #ddd;
+$transparent-dark: #111827bf;
```
na9da: Suggestion: I think ideally we should derive `transparent-dark` from `dark` value. The reason is that most maps customize `$dark` color. This leaves them in an inconsistent state where the workbench and a few components use `$transparent-dark` which could be very different from the `$dark` the map maintainer has set.

[Here's an example](http://ci.terria.io/ui-update-fonts/#configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json) of this for DEA maps.

Ideally, we would use the [new relative color value](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#relative_value_syntax) syntax, like `$transparent-dark: rgb(from $dark r g b / 0.75)`. But looks like it is still not widely supported.

The alternative is to derive the value at run time and inject it as a CSS variable. I was able to do this by setting `$transparent-dark: var(--theme-transparent-dark)` and deriving its value inside `GlobalTerriaStyles.ts` like so:

image

Not sure if there's an easier way to do this, but it would be good if the styling is consistent out of the box. What do you think?

lib/ReactViews/Tools/DiffTool/DiffTool.tsx
```diff
@@ -838,23 +837,19 @@ const DiffAccordionWrapper = styled(Box).attrs({
...
-  top: 70px;
+  top: 120px;
   left: 0px;
-  min-height: 220px;
-  // background: ${(p) => p.theme.dark};
-  margin-left: ${(props) =>
-    props.isMapFullScreen
-      ? 16
-      : parseInt(props.theme.workbenchWidth, 10) + 40}px;
+  margin-left: ${(props) => parseInt(props.theme.workbenchWidth, 10) + 40}px;
```
na9da, line 842: There's some issue with map not entering full screen mode when launching difftool + the positioning wrt to workbench. I am happy to look at it in a different PR (before final release).

na9da (changes requested):
Hi @ljowen, looks great - that's a lot of changes, great work navigating it.

I have made some comments - let me know what you think.

lib/ReactViews/BottomDock/Timeline/Timeline.jsx
ljowen: Looks like this is a firefox specific issue, I'll take a look

ljowen: This is fixed by adding `position: relative` as you've suggested below

lib/ReactViews/BottomDock/Timeline/timeline.scss
ljowen: Since the the title div is absolute positioned, this is here to prevent the chart from overlapping the title. As above this is now fixed for firefox

lib/ReactViews/BottomDock/BottomDock.tsx
ljowen: Adding this fixed the gap issue in firefox. One guess for the difference would be different handling of margin collapse with absolute elements.

lib/ReactViews/BottomDock/BottomDock.tsx
ljowen: Probably not, I figured long term we want to move toward styled components and remove scss, but happy to move the new rules into `bottom-dock.scss` instead.
 

- [x] Investigate `top-element` regression

lib/ReactViews/Map/TerriaViewerWrapper/Splitter/dragHook.ts
ljowen: Looks like it's related to the the map fullscreen logic (it works if the workbench is collapsed)

- [x] Fix splitter at mobile

lib/ReactViews/Tools/DiffTool/DiffTool.tsx
ljowen: This was intentional as with the new collapsed design there isn't sufficient space to fit the diff tool below the Logo / expand button. AFAICT this doesn't break the ability to use the diff tool but I could be missing a step / option.

I've created an issue here to convert it to a workflow:
https://github.com/TerriaJS/terriajs/issues/7457

lib/ReactViews/Map/TerriaViewerWrapper/Splitter/dragHook.ts
ljowen: I've added a check for `useSmallscreenInterface` feels a bit hacky for draghook to be aware of other components, but probably the simplest place for this logic to live

lib/ReactViews/BottomDock/BottomDock.tsx
ljowen: This was caused by setting z-index: 10 in `MapColumn.tsx` which was creating a new stacking context. I believe that was a carry-over from how I was originally laying out the bottom bar but isn't needed now

lib/ReactViews/Map/Toast.tsx
ljowen: This centers it in the viewport (i.e map + workbench width) but not strictly necessary, happy to revert this

lib/Sass/common/_variables.scss
ljowen: Looks relative color value is well supported (TIL thanks!) https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#browser_compatibility 

I'll have a go at implementing that

ljowen: Looking at [caniuse](https://caniuse.com/css-relative-colors) it looks like this support was only added recently (i.e. 2024), however it looks like sass actually transpiles it to a css `color` expression: i.e. `color(srgb 0 0 0.545098 / 0.75)`

lib/ReactViews/MapIconButton/MapIconButton.tsx
ljowen: Fixed, found a few others thanks!

lib/ReactViews/DataCatalog/data-catalog-group.scss
ljowen: - [x] Fix icon alignment

ljowen: I've fixed that now so they line up, I've noticed another issue if "displayGroup" and "private" are both true, but this exists in `main` so will create a new issue for it
Screenshot 2025-02-27 at 4 31 23 pm

lib/ReactViews/Map/Toast.tsx
na9da: IMHO think it looks better when centered wrt to the timeline/attribution UI instead of the viewport. But I leave it to you.

lib/Sass/common/_variables.scss
na9da: I think the problem with doing it in SCSS is that it only has access to the compile time values and not the runtime values coming from config (evidence - the DEA map toolbar is [charocal gray](http://ci.terria.io/ui-update-fonts/#configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json) whereas the workbench is transparent blue). 

Ii'm not sure if even my original suggestion `$transparent-dark: rgb(from $dark r g b / 0.75)` would have worked either. 

It has to read the `dark` value from props eg: `$transparent-dark: rgb(${p => p.props.dark} r g b / 0.75)`. However, I don't think that's valid SCSS. So we might have to resort to injecting via the global

na9da:
Thanks @ljowen - Looks great.
There's just the derivation of transparent color left to resolve.

Also, let me know if you are able to work in the final refinements suggested by @AmberStandley83.

irbian:
I would like to comment on two things that our team has found problematic when trying to make changes or adaptations of the tool:

- The reliance on vh and vw means that we can't add elements freely above or below the frame
- The use of generated classes instead of semantic ones makes difficult to attach external styles to them

This both are probably beyond the scope of your changes, but it would be helpful if you are conscious of them to minimize third party integrations problems

CHANGES.md
```diff
@@ -2,10 +2,30 @@
 
 #### next release (8.9.0)
 
-- Remove unused pmtiles dependency.- **Breaking changes**
-  - Add tiling support to `ArcGisFeatureServerCatalogItem` - this will be enabled by default if the server supports tiling and unsupported marker/point styles aren't used. See `ArcGisFeatureServerCatalogTraits` `tileRequests`. When enabled, `pbf` tiles will be fetched and drawn using `ProtomapsImageryProvider`
-    - This can be disabled by setting `tileRequests` to `false`
-    - For point features, only the following `PointSymbolTraits` are available - fill, stroke, height (which sets circle radius). Custom markers (markers other than `point` or `circle`) are not supported.
+- Remove unused pmtiles dependency.
```
ljowen:
```suggestion
```

na9da:
Hi @irbian, thanks for the feedback. I also see that you have mentioned this before in https://github.com/TerriaJS/terriajs/issues/5169.

To address your points:

> The reliance on vh and vw means that we can't add elements freely above or below the frame

I think this happens when we don't have those specific use cases in mind while making changes. It will be great if you can point out where specifically this has created problems and we will be happy to fix them.

> The use of generated classes instead of semantic ones makes difficult to attach external styles to them

As noted in https://github.com/TerriaJS/terriajs/issues/5169, this is a problem that comes with the switch to styled-components. One solution might be to expose more styling options as theme variables or customizable components. 

One problem I see with using external stylesheets for customizations is that it works like an implicit API that is more prone to accidental breakages when we change things in the UI.

na9da (approved):
Approving.

@ljowen - great job, looking good now :tada:

irbian:
> Hi @irbian, thanks for the feedback. I also see that you have mentioned this before in #5169.
> 
> To address your points:
> 
> > The reliance on vh and vw means that we can't add elements freely above or below the frame
> 
> I think this happens when we don't have those specific use cases in mind while making changes. It will be great if you can point out where specifically this has created problems and we will be happy to fix them.

Our main case was to add a common organization level header to all apps in our domain, with links to other organizations apps. We did implement that one as a plugin that added a div just above the current header, but is something that would happen to any component added outside the frame

> 
> > The use of generated classes instead of semantic ones makes difficult to attach external styles to them
> 
> As noted in #5169, this is a problem that comes with the switch to styled-components. One solution might be to expose more styling options as theme variables or customizable components.
> 
> One problem I see with using external stylesheets for customizations is that it works like an implicit API that is more prone to accidental breakages when we change things in the UI.

I can't speak about others, but for us it wouldn't be as much a problem as much as at least it would be possible to target the element to customization

k5h1tij:
How to revert Charts not showing on Mobile Viewport?

ljowen:
> How to revert Charts not showing on Mobile Viewport?

@k5h1tij This line can be removed to enable chart on mobile https://github.com/TerriaJS/terriajs/blob/b649e7c02f9d53d095de63f266ab59aa30d70a3e/lib/ReactViews/BottomDock/BottomDock.tsx#L63 We plan on restoring it but it needs some design issues addressed first

irbian:
Following this, we have found a possible solution to increase customization and theming

For example, instead of "primaryColor" that is used in lots of cases, we would have iconColor, textColor... initialized as primaryColor, but open to override if necessary

PR #7450, opened 2025-01-29
Webpack5 comments

zoran995:
### What this PR does

- Updates the versions of webpack plugins and cleanup no longer used ones.
- Resolves the Ion default token message issue. Depends on https://github.com/TerriaJS/cesium/pull/88 being merge and release first

### Test me

Locally: 
- sync terriamap-terriajs dependencies 
- install dependencies
- run the project
- check that everything works the same (type checking is still working, assimpjs works correctly (copy-plugin-webpack))

### Checklist


package.json
```diff
@@ -93,11 +88,10 @@
     "classnames": "^2.3.1",
     "clipboard": "^2.0.8",
     "commander": "^12.1.0 ",
-    "copy-webpack-plugin": "^6.4.0",
+    "copy-webpack-plugin": "^12.0.2",
```
pjonsson, line 91: Seems you've updated most comments for package.json, could you remove `@types/webpack` as well so that isn't forgotten?

lib/ReactViews/ExplorerWindow/Tabs/MyDataTab/my-data-tab.scss
```diff
@@ -136,9 +136,6 @@
 
 .btn--back-to-my-data {
   composes: btn from "../../../../Sass/common/_buttons.scss";
-  // TODO: fix?
-  // composes: btn-small from "../../../../Sass/common/_buttons.scss";
-  // composes: btn--left from "../../../../Sass/common/_buttons.scss";
```
zoran995, line 141: `btn--left` was an icon definition so it is safe to remove
`btn-small` wouldn't make any point here as we have all its styles defined/overriden here or in styled-component

lib/Sass/common/_buttons.scss
```diff
@@ -36,15 +36,6 @@ $btn-setting-size: 35px;
   margin: 0;
 }
 
-// TODO: fix?
-// .btn-small {
```
zoran995, line 40: This is not used anywhere anymore

package.json
```diff
@@ -1,12 +1,11 @@
 {
   "name": "terriajs",
-  "version": "8.7.10",
+  "version": "8.7.11",
   "description": "Geospatial data visualization platform.",
   "license": "Apache-2.0",
   "engines": {
     "node": ">= 16.0.0"
   },
-  "sideEffects": false,
```
pjonsson, line 9: This is apparently important for tree shaking. There is a granular version of sideEffects where you provide paths, and I think that would make sense since there are so few imports with side effects in the tree currently. Long-term, I'm guessing you want to get rid of side effecting imports.

pjonsson:
Your change to that other package looks straightforward and should be a matter of just merging, and then getting this moving. Do you know what is blocking the progress with getting this merged?

The diff view is no longer useful for reviewing after merging main, but besides me wishing for not completely eliminating the sideEffects directive I think what was here before merging main into the PR looked good.

package.json
zoran995: It is not always easy to properly setup the list of files, and I don't want to block the webpack upgrade with that

pjonsson: True, it's already been sitting for 2 months, would be nice to get these big PRs into main.

lib/ReactViews/Notification/NotificationWindow.jsx
```diff
@@ -65,14 +65,11 @@ const NotificationWindow = createReactClass({
...
       width: defined(this.props.width) ? this.props.width : "500px"
     };
-    const isStory = type === "story";
 
     return (
       <div className={classNames(Styles.wrapper, `${type}`)}>
         <div
           className={Styles.notification}
-          /* eslint-disable-next-line react/no-unknown-property */
-          isStory={isStory}
```
pjonsson, line 75: Does the comparison on line 78 work if you just remove this?

zoran995: thanks for catching this I restored the eslint-disable comments for those issues and will look in them more when this is merged into webpack5 PR

zoran995:
@na9da I had to restore `worker-loader` as webpack native web worker support tends to get pretty complex, I will keep looking for a fix but it might be good to keep worker-loader and handle that as a follow-up
p. S. Important note: webpack worker support behaves differently in development and production builds. That's why it works on CI but fails locally. I have run into this issue in the past but still don't have a solution that always works

lib/ReactViews/Custom/Chart/ChartPanelDownloadButton.tsx
```diff
@@ -69,9 +69,11 @@ async function download(items: TableMixin.Instance[]) {
   const synthesized = synthesizeNameAndValueArrays(items);
 
   // Could implement this using TaskProcessor, but requires webpack magic.
-  const worker = new Worker(
-    new URL("./downloadHrefWorker.js", import.meta.url)
-  );
+  const worker = await import(
+    //@ts-expect-error: aaa
```
pjonsson: I assume you know the reason you had to put this ts-expect-error in, put that instead of "aaa".

And style wise, no other ts-expect-error in terriajs has a colon after it, and most of them have a space between the `//` and `@ts-expect-error`.

pjonsson:
> p. S. Important note: it behaves differently in development and production builds with native support. That's why it works on CI but fails locally.

I don't know what part of the code this concerns, but it seems worth to put a comment in the code at that place so it doesn't get missed/lost the next time that thing is upgraded.

package.json
na9da: Yeah, we should remove this setting. I was mostly testing to see if it broke anything - it doesn't seem to. I think we can revisit this later.

zoran995:
I have cherry-picked and merged all changes from here into #7351, so it is easier and faster to review
Closing this

Issue #7320, opened 2024-11-20
when adding a layer from a wms, discovered via cssw-catalog, all layers are added at once


pvgenuchten:
I'm trying to find out what happens, so let me describe how I reproduce the issue

At https://rwanda.lsc-hubs.org/map/ we host a recent version of Terria, When opening the CSW catalog in `explore data` and search for 'slope'

![Image](https://github.com/user-attachments/assets/b9ecc532-722f-433b-9173-2b5358d0e41c)

![Image](https://github.com/user-attachments/assets/b4f20e75-4e96-4ac6-b979-743f4e821803)

Then click `+` button on any layer

Expected: Single layer is added to map

[...]

Issue #6866, opened 2023-09-01
Add `WebMapServiceCatalogGroup` support to `CswCatalogGroup` - and tweak handling of `WebMapServiceCatalogItem` `layers`


pvgenuchten:
A common usage scenario in webgis is to reference a wms from a csw record, terria would then offer capability to add the wms to the mapview.

Unfortunately the CSW/Dublin Core spec does not provide any guidance on how to reference a single layer within the WMS, multiple conventions around the globe have been invented to circumvent that limitation. Is it documented which conventions are implemented by terria?

Current behaviour is that Terria, if unable to identify the relevant layer, adds all layers of the WMS (from capabilities) to the request

This behaviour is problematic for the following reasons
- in my case, one of the layers has a problem, and the full request fails
- if the server has many layers, the request takes a long time, is resource intense, and does not provide any usefull information
[...]

Issue #5267, opened 2021-02-24
Enfore WebMapService (WMS) LayerLimit


mwu2018:
If a dataset distribution from magda registry has wms format and more than one layer, all layers are to be loaded at the same time. However, some servers may response with error for such request.

Example:
After request https://nationalmap.dev.saas.terria.io/proxy/_0d/http://neii.bom.gov.au/services/solarclim/wms/data/monClim_gloHorExp1Hou.nc?service=WMS&version=1.3.0&request=GetCapabilities, another request follows, https://nationalmap.dev.saas.terria.io/proxy/_0d/http://neii.bom.gov.au/services/solarclim/wms/data/monClim_gloHorExp1Hou.nc?transparent=true&format=image%2Fpng&exceptions=application%2Fvnd.ogc.se_xml&styles=&tiled=true&time=1993-12-16T15%3A00%3A00.000Z&service=WMS&version=1.1.1&request=GetMap&layers=mean_exposure%2Cland_mask&bbox=10018754.17139462%2C0%2C20037508.342789244%2C10018754.171394622&width=256&height=256&srs=EPSG%3A3857

However, with the second request, the server responds with error `You may only create a map from 1 layer(s) at a time`.

**Technical Note**
We probably need to improve MagdaReference model so that if there are multiple layers, treat the distribution as a group and each layer as a member of the group.

The layer number restriction is probably enforced by the data provider

![image](https://user-images.githubusercontent.com/41275384/109112114-3731bd00-778e-11eb-9478-435ea6d860f9.png)

Related issue https://github.com/TerriaJS/terriajs/issues/5266

PR #7466, opened 2025-02-15
enforce WMS layer limit

sidneygijzen:
### What this PR does

This PR proposes an approach for implementing the WMS layer limit. One of the goals is trying to mitigate the resource-intensive requests - as reported in #6866 and #7320 - by enabling a WMS server administrator to make use of the WMS layer limit.

My initial thought was to modify the `WebMapServiceCatalogItem`  `validLayers` getter, but during manual testing the styles URL param went awry. From what I can tell it is because the `WebMapServiceCatalogItem` `layersArray` is used in the `WebMapServiceCapabilitiesStratum` `legends` getter. Therefore I chose to check for a layer limit in the `layersArray`.

A note on the modified XML test files; during the implementation a couple of unrelated tests errored out. A small number of xml files contained a `<LayerLimit>1</LayerLimit>`. I have opted to remove the `<LayerLimit>1</LayerLimit>` elements from the XML file instead of adjusting the tests.

Finally, I am not sure if it fully addresses #5267, because I haven't looked at the MagdaReference model. 

### Test me

Aside from the added unit test, I tested this manually by configuring a wms-group as a wms. Please see [this gist](https://gist.github.com/sidneygijzen/6332b9a259c47aa50e1df00774a999f9) for a test config. With this "wrong" config I try to simulate what is happening in #7320, which is that the CswCatalogGroup currently has no knowledge of the concept of a WMS-group. The WMS below has a WMS layer limit defined, so only one wms layer is requested in this case.

### Checklist

### DIFF
--- a/lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
+++ b/lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
@@ -113,7 +113,7 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
   @computed
   get layerLimit() {
     if (isDefined(this.capabilities?.Service?.LayerLimit)) {
-      return parseInt(this.capabilities?.Service?.LayerLimit);
+      return parseInt(this.capabilities.Service.LayerLimit, 10);
     }
   }
 
--- a/lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
+++ b/lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
@@ -111,6 +111,13 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
   }
 
   @computed
+  get layerLimit() {
+    if (isDefined(this.capabilities?.Service?.LayerLimit)) {
+      return parseInt(this.capabilities?.Service?.LayerLimit);
+    }
+  }
+
+  @computed
   get layers(): string | undefined {
     let layers: string | undefined;
 
--- a/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
+++ b/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
@@ -246,14 +246,30 @@ class WebMapServiceCatalogItem
   }
 
   @computed
+  get layerLimit() {
+    const gcStratum: WebMapServiceCapabilitiesStratum | undefined =
+      this.strata.get(
+        GetCapabilitiesMixin.getCapabilitiesStratumName
+      ) as WebMapServiceCapabilitiesStratum;
+
+    return gcStratum?.layerLimit;
+  }
+
+  @computed
   get layersArray(): ReadonlyArray<string> {
-    if (Array.isArray(this.layers)) {
-      return this.layers;
-    } else if (this.layers) {
-      return this.layers.split(",");
+    if (!this.layers) return [];
+
+    let layers: string[];
+
+    if (!Array.isArray(this.layers)) {
+      layers = this.layers.split(",");
     } else {
-      return [];
+      layers = this.layers;
     }
+
+    if (this.layerLimit) layers = layers.slice(0, this.layerLimit);
+
+    return layers;
   }
 
   /** LAYERS which are valid (i.e. exist in GetCapabilities).
--- a/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
+++ b/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
@@ -261,10 +261,10 @@ class WebMapServiceCatalogItem
 
     let layers: string[];
 
-    if (!Array.isArray(this.layers)) {
-      layers = this.layers.split(",");
-    } else {
+    if (Array.isArray(this.layers)) {
       layers = this.layers;
+    } else {
+      layers = this.layers.split(",");
     }
 
     if (this.layerLimit) layers = layers.slice(0, this.layerLimit);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 #### next release (8.8.1)
 
+- enforce the WMS layer limit
 - [The next improvement]
 
 #### 8.8.0 - 2025-02-18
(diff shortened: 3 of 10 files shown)


lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -245,15 +245,31 @@ class WebMapServiceCatalogItem
...
   get layersArray(): ReadonlyArray<string> {
-    if (Array.isArray(this.layers)) {
-      return this.layers;
-    } else if (this.layers) {
-      return this.layers.split(",");
+    if (!this.layers) return [];
+
+    let layers: string[];
+
+    if (!Array.isArray(this.layers)) {
```
pjonsson: If you flip the then/else bodies you can avoid the negation in the condition.

lib/Models/Catalog/Ows/WebMapServiceCapabilitiesStratum.ts
```diff
@@ -110,6 +110,13 @@ export default class WebMapServiceCapabilitiesStratum extends LoadableStratum(
       );
   }
 
+  @computed
+  get layerLimit() {
+    if (isDefined(this.capabilities?.Service?.LayerLimit)) {
+      return parseInt(this.capabilities?.Service?.LayerLimit);
```
pjonsson: Doesn't the previous line guarantee that `this.capabilities.Service.LayerLimit` is defined, so the question marks can be omitted?

sidneygijzen:
@pjonsson Thanks for the review! I've updated the PR.

lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -245,15 +245,31 @@ class WebMapServiceCatalogItem
...
+  @computed
+  get layerLimit() {
+    const gcStratum: WebMapServiceCapabilitiesStratum | undefined =
+      this.strata.get(
+        GetCapabilitiesMixin.getCapabilitiesStratumName
+      ) as WebMapServiceCapabilitiesStratum;
+
+    return gcStratum?.layerLimit;
+  }
+
```
nf-s, line 257: Instead of pulling `layerLimit` out of the `WebMapServiceCapabilitiesStratum`, it would be better to create a `layerLimit` trait in `WebMapServiceCatalogItemTraits`.

This would allow the user to configure it (in a JSON catalog), and then you can remove this `get layerLimit` in `WebMapServiceCatalogItem`.

While there are a few instances of using the `WebMapServiceCapabilitiesStratum` directly in `WebMapServiceCatalogItem`, this is not ideal, and should be considered an anti-pattern, at some point we'll refactor it 🙂

nf-s (changes requested):
Thanks @sidneygijzen !

Just one code comment, and can you please add a warning message in `shortReport` if the layer limit has been hit? Just so users know if a WMS item has been configured with too many layers

https://github.com/sidneygijzen/terriajs/blob/c22bf3f09c9f0d190c02599598856e13d2e3d43e/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts#L185-L194

Thanks 🙂

sidneygijzen:
@nf-s thank you for the review! I appreciate it that you're explaining why something should be in a specific spot in the codebase.

I will look into adding a warning message in `shortReport` about the WMS layer limit.

Concerning the code comment, I am afraid I am a bit confused.

A WMS LayerLimit is configured - and visible in the WMS GetCapabilities - by the organisation operating the WMS server. For example: see [this section about  `wms_layerlimit`](https://mapserver.org/ogc/wms_server.html#index-58) in the MapServer docs. This PR aims to read the layer limit configured by the organisation operating the WMS server and use it while making WMS GetMap requests.

Where it gets confusing for me, is that you refer to "allow the user to configure it (in a JSON catalog)". I am assuming that by "the user" you are referring to a person who is deploying TerriaJS. In my honest opinion there is no configuration to be made in this regard by the organisation/person deploying TerriaJS. It is not possible to change the `wms_layerlimit` of the WMS server by configuring it in TerriaJS. If the person deploying TerriaJS would configure a higher value for the layer limit than the WMS server allows, it results in WMS server error messages (see #5267). What would be use cases for setting a lower value than what the WMS server allows?

The aforementioned is the reason why I chose to pull the layer limit out of the `WebMapServiceCapabilitiesStratum`. From my point of view it is simply a constant found in the WMS GetCapabilities. What would be the right place for reading config from the WMS GetCapabilities and using it when making WMS GetMap requests?

What do you think?

pvgenuchten:
In my view both approaches are valid and could both be implemented:
- any software should respect the layerlimit if it is advertised in capabilities, considering fair use of the service
- if a terria admin aims to limit the number of layers requested from a service, it could only be done in a terria configuration (if the service is out of their control)

My suggestion would be to start with this, and then implement the second scenario in a separate PR, unless the approaches are conflicting...

zoran995:
@sidneygijzen, @pvgenuchten, one approach we can try here is to introduce a trait for LayerLimit, allowing users to define their value as part of the JSON configuration. And if the layerLimit is defined as part of WMS Capabilities, use that as the default value if not supplied by the user. But let me check with @nf-s what he thinks about this and get back to this

nf-s:
Sorry for the late reply @sidneygijzen

It is less about being able to manually configure the `layerLimit` - it is more about following existing patterns that are in place. Defining a `layerLimit` trait also provides some documentation (as you will need to specify a description).

Also, if you define a `layerLimit` trait, then the value set in `WebMapServiceCapabilitiesStratum` will automatically be available in `WebMapServiceCatalogItem` (so you can delete your `get layerLimit` accessor in `WebMapServiceCatalogItem`_.

You can look at `isGeoServer` as an example trait.

If you are unsure how the model layer/traits system works - we have some documentation available here

- https://docs.terria.io/guide/contributing/model-layer/
- https://docs.terria.io/guide/contributing/traits-in-depth/
- https://docs.terria.io/guide/contributing/strata-examples/

In short, you should define a `layerLimit` trait in `WebMapServiceCatalogItemTraits` like so

```ts
  @primitiveTrait({
    type: "number",
    name: "Layer limit",
    description: "The maximum number of layers that can be requested simultaneously."
  })
  layerLimit?: number;
```

And then you should delete the `get layerLimit` accessor in `WebMapServiceCatalogItem`

https://github.com/sidneygijzen/terriajs/blob/df7e04cc70e87b2eda71e8fd4c35f28b4b6eb52d/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts#L249-L256

After both of those things, you will be able to use `this.layerLimit` in `WebMapServiceCatalogItem` - and the value from `WebMapServiceCapabilitiesStratum` will automatically flow into the `WebMapServiceCatalogItem` model

Thanks!

sidneygijzen:
Thank you for the detailed response @nf-s. I appreciate it a lot!

I will have a closer look in the coming weeks and update this PR accordingly.

sidneygijzen:
I've thinking about this PR a while now, but the more I work on it, I'm not convinced the current implementation is correct. The current implementation limits the layers array, however I think the limitation is set too early in the process. A better implementation would be when a user would like to view multiple layers simultaneously, not when listing the available layers.

So, following that line of thought, I am thinking of the following: 1) reverting the change to the layers array and 2) checking the layer limit when the user actually wants to add a layer to the map, e.g by checking the layerlimit in the `Workbench` and/or `DataCatalogItem` component. 

Because of the above, I've converted this PR to draft state.

PR #4803, opened 2020-09-29
Port catalogue search from digital twins implementation

reginapramesti:
### What this PR does

Fixes #3657 

Basically took the magda search implemented for digital twins and modified it so that it doesn't just load `MagdaReference` items but loads all groups and references

### Checklist

-   [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
-   [ ] I've updated CHANGES.md with what I changed.

PR #7476, opened 2025-02-19
tsify data catalog tab and convert it to functional component

zoran995:
### What this PR does

This is needed for mobx-react upgrade as part of an upgrade to react 18.

### Test me

### Checklist


lib/ReactViews/ExplorerWindow/Tabs/DataCatalogTab.tsx
```diff
@@ -0,0 +1,94 @@
...
+            {/* TODO2: Implement a more generic MobX DataCatalogSearch */}
+            {searchState.catalogSearchProvider && (
+              <SearchBox
+                searchText={searchState.catalogSearchText}
+                onSearchTextChanged={changeSearchText}
+                onDoSearch={() => search()}
+                placeholder={searchPlaceholder}
+                debounceDuration={
+                  terria.catalogReferencesLoaded &&
+                  searchState.catalogSearchProvider
```
pjonsson: This part of the condition is already checked on line 51.

lib/ReactViews/ExplorerWindow/Tabs/DataCatalogTab.tsx
```diff
@@ -0,0 +1,94 @@
...
+import DataPreview from "../../Preview/DataPreview";
+import Breadcrumbs from "../../Search/Breadcrumbs";
+import SearchBox, { DEBOUNCE_INTERVAL } from "../../Search/SearchBox.jsx";
+import Styles from "./data-catalog-tab.scss";
+import { useViewState } from "../../Context";
+
+interface DataCatalogTabProps {
+  items?: unknown[];
+  searchPlaceholder?: string;
+  onActionButtonClicked?: () => void;
```
pjonsson: This type is `(item: something) => void`.

lib/ReactViews/ExplorerWindow/Tabs/DataCatalogTab.tsx
```diff
@@ -0,0 +1,94 @@
...
+    breadcrumbsShown: showBreadcrumbs,
+    terria
+  } = viewState;
+
+  const searchPlaceholder =
+    props.searchPlaceholder || t("addData.searchPlaceholder");
+
+  const changeSearchText = (newText: string) => {
+    runInAction(() => {
+      viewState.searchState.catalogSearchText = newText;
```
pjonsson, line 38: Should this use `searchState.X` instead of `viewState.searchState.X` just like the return statement a few lines below?

pjonsson:
I did the straightforward (?) conversion to ES6 objects of some of this stuff in #7415, you might be able to cherry-pick the non-overlapping parts on top of this to get more TSification of this part of the code.

lib/ReactViews/ExplorerWindow/Tabs/DataCatalogTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+import SearchBox, { DEBOUNCE_INTERVAL } from "../../Search/SearchBox.jsx";
+import Styles from "./data-catalog-tab.scss";
+
+interface DataCatalogTabProps {
+  items?: unknown[];
+  searchPlaceholder?: string;
+  onActionButtonClicked?: (item: CatalogMemberMixin.Instance) => void;
+}
+
+const DataCatalogTab = observer(function DataCatalogTab(
```
pjonsson, line 20: The `const DataCatalogTab` gets shadowed by the function name inside `observer()`, is that intentional or could the function passed to the observer be without name?

lib/ReactViews/ExplorerWindow/Tabs/DataCatalogTab.tsx
```diff
@@ -0,0 +1,92 @@
...
+                searchText={searchState.catalogSearchText}
+                onSearchTextChanged={changeSearchText}
+                onDoSearch={() => search()}
+                placeholder={searchPlaceholder}
+                debounceDuration={
+                  terria.catalogReferencesLoaded
+                    ? (searchState.catalogSearchProvider as any)
+                        .debounceDurationOnceLoaded
+                    : DEBOUNCE_INTERVAL
+                }
```
na9da: ~Possible error: Previously, if  either `terria.catalogReferencesLoaded` is `false` or if `searchState.catalogSearchProvider` is `undefined` the duration defaults to `DEBOUNCE_INTERVAL`. But now if `searchState.catalogSearchProvider` is `undefined` it could result in a type error?~ Nah, my bad - we are ruling out the `undefined` case above.

Also looking at the `any` cast, I'm thinking whether we can hardcode `debounceDurationOnceLoaded` within `DataCatalogTab` component. I don't see a reason why it needs to be configurable at the `CatalogSearchProvider` level.

zoran995: `debounceDurationOnceLoaded` was added in #4803 don't know the exact reason why it was added. I have improved type cast, but I also don't see why we need that we can just hardcode the value

na9da (approved):

PR #7477, opened 2025-02-19
convert search box and results to functional component

### LINKED ISSUE #7518: Tsify SearchBoxAndResults and convert to functional component
To ensure compatibility with the new `mobx-react`, we need to move SearchBoxAndResults to a functional component as part of this effort this component should be moved to the typescript

p.s. we already have this completed, just need to cherry-pick the changes

zoran995:
### What this PR does

Fixes: #7518 

This is needed for mobx-react upgrade as part of an upgrade to react 18.

### Test me

Verify that search box and results for the same as on main:
- search results are properly renderer and click on search results 
- clearing search box closes the search results panel
- clearing the search box removes the marker from the map

### Checklist


lib/ReactViews/Search/SearchBoxAndResults.tsx
```diff
@@ -67,85 +73,65 @@ const PresentationBox = styled(Box).attrs({
 
 export const LOCATION_SEARCH_INPUT_NAME = "LocationSearchInput";
 
-export class SearchBoxAndResultsRaw extends React.Component {
-  constructor(props) {
-    super(props);
-    this.locationSearchRef = React.createRef();
-  }
+interface SearchBoxAndResultsProps {
+  placeholder?: string;
```
pjonsson: The placeholder is set as `isRequired` in the old code, is the `?` intentional here?

zoran995: It is only used in the `SearchBox`, and it is not required there; furthermore, tests did not provide the placeholder value, so I assumed it is safe to make it optional (can make it required if needed).

zoran995: I have made it required again

lib/ReactViews/Search/SearchBoxAndResults.tsx
```diff
@@ -67,85 +73,65 @@ const PresentationBox = styled(Box).attrs({
...
-    );
-  }
+      // TODO(wing): why is this a reaction here and not in viewState itself?
+      // Close the search results when the Now Viewing changes (so that it's visible).
+      const disposeReaction = reaction(
+        () => viewState.terria.workbench.items,
+        () => {
+          viewState.searchState.showLocationSearchResults = false;
+        }
+      );
```
na9da: Suggestion: Here again we can replace the `reaction` with a `useEffect` like:

```js
    useEffect(() => {
      viewState.searchState.showLocationSearchResults = false;
    }, [viewState.terria.workbench.items]);
```

lib/ReactViews/Search/SearchBoxAndResults.tsx
```diff
@@ -158,10 +144,10 @@ export class SearchBoxAndResultsRaw extends Component {
...
           <PresentationBox highlightBottom={shouldShowResults}>
             <SearchBox
-              ref={this.locationSearchRef}
-              onSearchTextChanged={this.changeSearchText.bind(this)}
-              onDoSearch={this.search.bind(this)}
-              onFocus={this.startLocationSearch.bind(this)}
+              ref={locationSearchRef}
+              onSearchTextChanged={changeSearchText.bind(this)}
+              onDoSearch={search.bind(this)}
+              onFocus={startLocationSearch.bind(this)}
```
na9da: We don't need the `.bind(this)` anymore.

na9da (changes requested):
Looks good @zoran995. I have left a couple of comments.

lib/ReactViews/Search/SearchBoxAndResults.tsx
zoran995: yeah we can do that here as it is a simple value so we will always have new reference

na9da (approved):
LGTM. Please fix CHANGES.md before you merge.

PR #7498, opened 2025-02-25
disable deployment actions job on forks

zoran995:
### What this PR does

### Test me

### Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,6 +4,7 @@ on: push
 
 jobs:
   deploy:
+    if: github.repository_owner == 'TerriaJS'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on: push
 
 jobs:
   deploy:
+    if: github.repository_owner == 'TerriaJS'
```
pjonsson, line 7: Doesn't this need to be `if: ${{ github.repository_owner == 'TerriaJS' }}` like the example at https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#determining-when-to-use-contexts?

zoran995: I used this doc https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/using-conditions-to-control-job-execution, which also references another doc that states 
> The exception to this rule is when you are using expressions in an if clause, where, optionally, you can usually omit ${{ and }}. For more information about if conditionals, see [Workflow syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif).

ljowen (approved):

PR #7504, opened 2025-02-26
Fix GeoJSON regression bug + add tests

nf-s:
### Fix GeoJSON regression bug + add tests

Fix bug introduced in https://github.com/TerriaJS/terriajs/pull/7144

### Test me

There are three geojson files, all three won't work in `main`

- Before http://ci.terria.io/main/#share=s-fR3L6Wrn6XB731RwOjMDNWAe4dJ
- After http://ci.terria.io/fix-geojson-protomaps-bug/#share=s-fR3L6Wrn6XB731RwOjMDNWAe4dJ

### Checklist


lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -82,73 +82,133 @@ export class ProtomapsGeojsonSource implements TileSource {
...
+        numVertices = transformedGeom.length;
+      }
+
+      if (f.type === 0) return null;
+
+      const geomType = {
+        [1]: GeomType.Point,
+        [2]: GeomType.Line,
+        [3]: GeomType.Polygon
+      }[f.type];
```
pjonsson, line 201: Is this the bug fix?

I don't have a strong opinion on whether a PR marked as bug fix should contain minor refactors or not, but PRs are easier to review if the bug fix is in a separate commit from the commit that moves a big chunk of code around.

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -82,73 +82,133 @@ export class ProtomapsGeojsonSource implements TileSource {
...
+        props: { ...(f.tags ?? {}) },
+        bbox,
+        geomType,
+        geom: transformedGeom,
+        numVertices
+      };
+
+      return feature;
+    })
+    .filter((f): f is ProtomapsFeature => f !== null);
```
pjonsson, line 213: I see this has removed a cast, which is nice. Does the return type of the anonymous function passed to filter assert that the input is a ProtomapsFeature? Can't TypeScript figure that out without assertions if the the function passed to map on line 122 gets the return type annotation `ProtomapsFeature | null`? (Or the input type on this function gets a `ProtomapsFeature | null` type annotation.)

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -82,73 +82,133 @@ export class ProtomapsGeojsonSource implements TileSource {
...
-          .filter((f) => f !== null) as ProtomapsFeature[]
+        geojsonVtTileToProtomapsFeatures(tile.features, tileSize)
       );
     }
 
     return result;
   }
 }
+
+export const geomTypeMap = (
```
pjonsson, line 97: This function appears to be unused.

lib/Map/Vector/ProtomapsGeojsonSource.ts
```diff
@@ -82,73 +82,133 @@ export class ProtomapsGeojsonSource implements TileSource {
...
+            }
+
+            if (bbox.minY > y) {
+              bbox.minY = y;
+            }
+
+            if (bbox.maxY < y) {
+              bbox.maxY = y;
+            }
+            return new Point(x, y);
```
zoran995, line 158: Can we extract this to function createPointAndUpdateBbox or two functions, as this is also repeated below?

lib/Map/Vector/ProtomapsGeojsonSource.ts
na9da: I think it is fine because this change is cherry picked from [a bigger PR](https://github.com/TerriaJS/terriajs/pull/7370) for a bug fix release. More refactoring changes will follow in the parent PR.

nf-s:
At this stage I just want to get this PR merged as is, GeoJSON is currently broken in the latest released version of TerriaJS

na9da (approved):
Approving for bug fix release.
Suggested changes could be looked at in the refactor PR.

PR #7522, opened 2025-03-03
Add type cast

ljowen:
### What this PR does

Fixes failing typecheck on `main`

### Test me

Verify CI passes 

### Checklist

### DIFF
--- a/lib/Table/createRegionMappedImageryProvider.ts
+++ b/lib/Table/createRegionMappedImageryProvider.ts
@@ -14,6 +14,7 @@ import ProtomapsImageryProvider from "../Map/ImageryProvider/ProtomapsImageryPro
 import { TerriaFeatureData } from "../Models/Feature/FeatureData";
 import TableStyle from "./TableStyle";
 import { isConstantStyleMap } from "./TableStyleMap";
+import isReadOnlyArray from "../Core/isReadOnlyArray";
 
 export default function createRegionMappedImageryProvider(
   style: TableStyle,
@@ -147,7 +148,7 @@ const getImageryLayerFilteredRow = action(
     if (!isDefined(rowNumbers)) return;
 
     if (!isDefined(currentTimeRows)) {
-      return Array.isArray(rowNumbers) ? rowNumbers[0] : rowNumbers;
+      return isReadOnlyArray(rowNumbers) ? rowNumbers[0] : rowNumbers;
     }
 
     if (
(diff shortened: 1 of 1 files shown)


lib/Table/createRegionMappedImageryProvider.ts
```diff
@@ -147,7 +147,9 @@ const getImageryLayerFilteredRow = action(
     if (!isDefined(rowNumbers)) return;
 
     if (!isDefined(currentTimeRows)) {
-      return Array.isArray(rowNumbers) ? rowNumbers[0] : rowNumbers;
+      return Array.isArray(rowNumbers)
+        ? rowNumbers[0]
+        : (rowNumbers as number | undefined);
```
ljowen: generally not a fan of typecasting like this but the explicit type narrowing looks a bit silly:

```ts
if(typeof rowNumbers === 'number') return rowNumbers;
return Array.isArray(rowNumbers)
        ? rowNumbers[0]
        : (rowNumbers as number | undefined);
```

pjonsson: The error message in the deploy workflow is
```
TS2322: Type 'number | readonly number[]' is not assignable to type 'number | undefined'.
  Type 'readonly number[]' is not assignable to type 'number'.
```
so I think the undefined is already narrowed away by the `if (!isDefined(rowNumbers)) return;` on line 147, and therefore not required in this typecast.

Edit: based on Zoran's explanation that the error is from TypeScript 5.8, I think we can just use `isReadOnlyArray` from `lib/Core` for the check:
```typescript
return isReadOnlyArray(rowNumbers) ? rowNumbers[0] : rowNumbers;
```

zoran995: This appears to be caused by TypeScript 5.8.x. It cannot be reproduced with 5.7.x, not sure why as this has been "broken" in TypeScript for some time (though "broken" might not be the best word).

IMO type casting in this case is good enough and we can revisit if run into it in more places

zoran995 (approved):

lib/Table/createRegionMappedImageryProvider.ts
pjonsson: > It cannot be reproduced with 5.7.x, [..]

TypeScript 5.8 improved the checks/narrowing on conditional returns: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-8.html#granular-checks-for-branches-in-return-expressions

> This appears to be caused by TypeScript 5.8.x. 

That version of TypeScript is not in yarn.lock in any of the repositories, but I guess that is because the deploy job removes yarn.lock before running yarn install?
https://github.com/TerriaJS/terriajs/blob/ddb88d05b58ba6f80e0120a81b7e9e0a51f0c1c4/buildprocess/ci-deploy.sh#L35
Shouldn't it be sufficient with the edit to the package path in the earlier line:
https://github.com/TerriaJS/terriajs/blob/ddb88d05b58ba6f80e0120a81b7e9e0a51f0c1c4/buildprocess/ci-deploy.sh#L29
and then just running `yarn install` without removing `yarn.lock` first so the TypeScript  version doesn't change and cause issues like this?

ljowen:
Fix ts error: Use isReadOnlyArray

PR #7532, opened 2025-03-06
Convert difftool to use workflow panel

### LINKED ISSUE #7457: Convert Diff tool to workflow
With the new UI design the placement of the diff tool is a bit awkward. Change it to use the workflow ui (i.e. replacing the workbench) See style editor for workflow example.

na9da:
### What this PR does

- Fixes https://github.com/terriajs/terriajs/issues/7457
- Removes `showCloseButton` option for `Tool`. This was no longer being used after https://github.com/TerriaJS/terriajs/pull/5062
- Removes `CloseToolButton` component which showed a close button in the map toolbar. This was only being used by the difftool which no longer requires it after the switch to workflow panel.

Previously diff tool used a floating side panel design which was introduced before we had `WorkflowPanel`. This PR updates the difftool to use the `WorkflowPanel` and makes it more consistent with the new UI changes introduced in #7443.

Before:
image

After:
image

### Test me

- Run the difference workflow from [main](http://ci.terria.io/main/#share=s-g8FjgFZeardBY5gnhOgMt9gO9St&configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json) and [this branch](http://ci.terria.io/difftool-workflow/#share=s-g8FjgFZeardBY5gnhOgMt9gO9St&configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json) for the item in the workbench.
image

- Make sure that the functionality and outputs remain unchanged
- Make sure there are no errors

### Checklist


lib/ReactViews/Tools/DiffTool/DiffTool.tsx
```diff
@@ -508,48 +498,59 @@ class Main extends React.Component<MainPropsType> {
...
+      >
+        <div
+          css={`
+            display: flex;
+            flex-direction: column;
+            height: 100%;
+            padding: 5px 10px;
+
+            > * {
+              margin: 10px 0;
```
ljowen: Since you this is flex and column you could use `gap: 20px` to space the children evenly. Might be a bit more concise using `Box` (not sure if we want to move away from that though)

ljowen:
Looks great, that was quick!
Screenshot 2025-03-06 at 2 38 48 pm

Would be good to get @AmberStandley83 's thoughts on wording here, maybe "Back" could be "Edit" or "Close" could be "Save" or "Ok", right now it's not clear which one is taking me to the workbench vs the first panel.

Down the track if we're going to have multistep workflows a "wizard" style UI could be used here but thats for a later date

lib/ReactViews/Tools/DiffTool/DiffTool.tsx
na9da: Thanks, TIL!

AmberStandley83:
@kvw99 this one

kvw99:
As discussed, let's go with option C in the Figma designs: https://www.figma.com/design/5asDpMWtZfa0WcYp5RHgcR/Terria-Digital-Twin-Product----Component-Library?node-id=2805-1529&t=ojMraMpARs43kkB2-1
Frame 2452

lib/ReactViews/Tools/DiffTool/DiffTool.tsx
```diff
@@ -513,48 +500,36 @@ class Main extends Component<MainPropsType> {
...
+          this.resetTool();
+          this.closePanel();
+        })}
+      >
+        <div
+          css={`
+            display: flex;
+            flex-direction: column;
+            height: 100%;
+            padding: 5px 10px;
```
ljowen: nitpick: This would be more consistent with other panels as `padding: 15px`
```suggestion
            padding: 15px;
```

ljowen (approved):

PR #7541, opened 2025-03-10
Add missing children to prop types

### LINKED ISSUE #7515: Implicitly type React children
React 18 types no longer automatically add the children type to functional component props, meaning we will need to either define children in props or use `PropsWithChildren`. This will align us with the types used in React 16 for functional components.
Codemod can be run using `npx types-react-codemod preset-18 ./lib`

zoran995:
### What this PR does

Fixes #7515

We already defined the children in most of the prop types, so making the changes using the codemod is not needed, I simply updated only components that missed those (with the codemod number of changed files was 100+). 

### Test me

Those are type only changes, so if the build and test pass in CI it works.

### Checklist


na9da (approved):
LGTM

Issue #7519, opened 2025-02-28
Upgrade mobx-react to v9


zoran995:
Mobx-react version 7 is not compatible with React 18 strict mode, so we need to upgrade to mobx-react v9. 
New version of mobx-react brings a few breaking changes for class components, so we will need to migrate some components to functional components or manually convert props to observables. See https://github.com/mobxjs/mobx/tree/main/packages/mobx-react#class-components for more details

Components discovered for migration

- [x] SearchBoxAndResults - PR #7477 
- [x] ViewingControls - PR #7574 
- [x] WorkbenchItem - PR #7564
- [x] DragWrapper - PR #7544 
- [x] DataCatalogTab - PR #7476
- [x] DragDropNotification - PR #7543
- [x] DragDropFile - PR #7563 
- [x] MenuButton - PR #7576 
- [x] Dropdown - PR #7577 
- [x] SettingPanel - PR #7589
- [x] DataCatalogGroup - PR #7575 
- [x] FeatureInfoPanelChart - https://github.com/TerriaJS/terriajs/pull/7598
- [x] BottomDockChart and related components - https://github.com/TerriaJS/terriajs/pull/7581
- [x] DateTimePicker - PR #7585
- [ ] FeatureInfoSection
- [ ] MapInteractionWindow

p.s. most of this is already completed and need to be cherry-picked into separate PRs for easier review

PR #7543, opened 2025-03-10
tsify drag drop notification

zoran995:
### What this PR does

Part of #7519 

Convert drag-drop notification to a functional component so it can work properly with modern MobX-React.

### Test me

Try adding the files to the map by drag-and-drop and check notification that data is added to `my data`

### Checklist


lib/ReactViews/DragDropNotification.tsx
```diff
@@ -0,0 +1,88 @@
+import classNames from "classnames";
+import { reaction } from "mobx";
+import { observer } from "mobx-react";
+import { FC, useEffect, useRef, useState } from "react";
+import { Trans } from "react-i18next";
+import Icon from "../Styled/Icon";
+import { useViewState } from "./Context";
+import Styles from "./drag-drop-notification.scss";
```
na9da: Suggestion: Would you consider converting to `styled-component` as part of this PR? It doesn't look overly complicated to me.

lib/ReactViews/DragDropNotification.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const DragDropNotification: FC = observer(() => {
+  const viewState = useViewState();
+  const [showNotification, setShowNotification] = useState(false);
+  const notificationTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(
+    null
+  );
+
+  useEffect(() => {
+    const disposer = reaction(
```
na9da, line 21: Doubt: I know this is coming from the original version but I don't entirely understand why we need the `reaction` - could we simply use `useEffect` and `setState` to achieve the same? 

Another thing I noticed is that if I drag-n-drop a file and not move my mouse, then in `main` the notification will automatically dismiss in 5 secs but in this branch it stays forever. Not sure if it is timing or browser dependent. Are you able to repro it?

na9da (changes requested):

lib/ReactViews/DragDropNotification.tsx
zoran995: The `lastUploadedFiles` is the observable array, and if it is updated without changing the reference `useEffect`, it won't detect that. If we don't have/see any other use case for `lastUploadedFiles`, we might be able to move it to an internal state  and have DragDropNofication nested in DragDropFile

zoran995: The `viewState.lastUploadedFiles` was actually triggering the useEffect cleanup, which cleared the timeout and made it always visible. It is fixed now

lib/ReactViews/DragDropNotification.tsx
zoran995: Done

lib/ReactViews/DragDropNotification.tsx
na9da: hmm.. I tested the following code removing the `reaction` and it still works. Because we have wrapped the component in `observer`, mobx should automatically re-run the component when any referenced observable changes - in this case `lastUploadedFiles`.

Or am I missing something else?

```js
useEffect(() => {
    if (notificationTimeoutRef.current) {
      clearTimeout(notificationTimeoutRef.current);
    }
    // show notification, restart timer
    setShowNotification(true);
    // initialise new time out
    notificationTimeoutRef.current = setTimeout(() => {
      setShowNotification(false);
    }, 5000);

    return () => {
      if (notificationTimeoutRef.current) {
        clearTimeout(notificationTimeoutRef.current);
      }
    };
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [viewState.lastUploadedFiles]);
```

zoran995: Yeah, but the reference will always stay the same and react won't trigger `useEffect`, unless we always reassign the `lastUploadedFiles`. i.e. changing in `DragDropFile` to
```
runInAction(() =>
  props.viewState.lastUploadedFiles.push(
    ...addedCatalogItems.map((item) =>
      CatalogMemberMixin.isMixedInto(item) ? item.name : item.uniqueId
    )
  )
);
```

And since it is exposed through viewState to users I believe we need to be careful about it.

na9da: Yeah, I kind of agree. It might be good to simplify things were possible as we go. So I would suggest rewriting using the local state strategy that you hinted above. Let me know what you think.

zoran995: I will merge this into #7563 and reimplement the logic there

zoran995:
This was merged into #7563

PR #7544, opened 2025-03-10
Tsify drag wrapper and convert to functional component

zoran995:
### What this PR does

Part of #7519 

Rewrite the draggable functionality as a custom hook and reimplement the drag wrapper without `interactjs`. Also, remove interactjs from dependencies as it is no longer used and has a typing issue.

### Test me

Try dragging the feature info panel on the map

### Checklist


lib/ReactViews/DragWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+        restriction: "parent",
+        endOnly: true,
+        elementRect: { left: 0, right: 1, top: 0, bottom: 1 }
+      }
+    });
+
+    const resizeListener = () => {
+      // Use the correct type for the drag event
+      const dragEvent = { name: "drag", axis: "xy" };
+      (interactable as any).reflow(dragEvent);
```
pjonsson: I can't find documentation for the version of interactjs that is currently used, but `reflow` in interactjs 1.10 returns a promise, should there be an await here?

package.json
```diff
@@ -112,7 +112,7 @@
     "i18next-browser-languagedetector": "^6.1.4",
     "i18next-http-backend": "^1.4.1",
     "inobounce": "^0.1.2",
-    "interactjs": "1.4.0-alpha.17",
+    "interactjs": "1.10.27",
```
pjonsson: Should there be a `^` on the version, or does the version of this need to be completely locked?

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "useDefineForClassFields": true, // required for mobx6 - https://mobx.js.org/installation.html#use-spec-compliant-transpilation-for-class-properties
     "outDir": "ts-out",
     "resolveJsonModule": true,
+    "skipLibCheck": true,
```
pjonsson: I'm guessing this is to avoid hitting the
```typescript
import * as Interact from '@interactjs/types/index'

export as namespace Interact
export = Interact
```
in `node_modules/@interactjs/types/typings.d.ts`.

Does there exist some declare-module magic in `lib/ThirdParty` that can avoid the mismatching module format, or whatever the problem with `export = Interact` is caused by?

zoran995: Don't think it is possible, I don't understand why are they generating that file manually instead of having it output by bundler. They changed the app output to esm but haven't updated that file

I have pinned the latest version that doesn't have that issue

pjonsson: With the previous revision of this PR, I added a `lib/ThirdParty/@interactjs/types/typings.d.ts`:
```typescript
declare module "@interactjs/types" {
   export * from '@interactjs/types/index';
}
```
and changed `tsconfig.json`:
```patch
diff --git a/tsconfig.json b/tsconfig.json
index f432dbee0..e68b5262d 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,10 @@
     "useDefineForClassFields": true, // required for mobx6 - https://mobx.js.org/installation.html#use-spec-compliant-transpilation-for-class-properties
     "outDir": "ts-out",
     "resolveJsonModule": true,
-    "skipLibCheck": true,
+    "paths": {
+      "@interactjs/types": ["./lib/ThirdParty/@interactjs/types/typings.d.ts"],
+      "@interactjs/types/*": ["./node_modules/@interactjs/types/*"]
+    },
     "typeRoots": [
       "./lib/ThirdParty/",
       "./node_modules/",
```
and that makes `tsc -b` pass. Not sure if that is better or worse than pinning an older version.

zoran995: It could work, but paths are typically used in monorepos to connect the internal packages, and I wouldn't complicate the TypeScript configuration due to dependency issues. It would be easier to fork interactjs until they resolve it, but I will try a little more radical solution here and completely remove interactjs.

pjonsson: I put up a PR on interactjs that just switched the written file from UMD to ESM format; if he wants something else that gives him the opportunity to say so. And my PR failed the CI for unrelated reasons, so I put up a second PR that made the CI pass.

The interactjs author responded to some ticket about the documentation website being down a couple of days ago, so he's at least still semi-active.

lib/ReactViews/Drag/use-draggable.ts
na9da, line 1: nitpick: It might be good to stick to camel case file names `useDraggable.ts`

lib/ReactViews/Drag/use-draggable.ts
```diff
@@ -0,0 +1,210 @@
+import { useCallback, useEffect, useRef, useState } from "react";
+
+export const useDraggable = (options?: { handleSelector?: string }) => {
```
na9da: nitpick: Should we convert `options` and `handleSelector` to required? AFAICT  `isValidDragHandle` test will fail unless this is set?

na9da (changes requested):
@zoran995 - looks great and we have one less dependency!

One subtle difference I noticed is that in `main`, when hovering over the title, the mouse cursor will switch to `move`. Would it be possible to replicate that behaviour?

image

na9da:
Also, please update the PR description and CHANGES.md to say that the interact dependency was removed.

lib/ReactViews/Drag/use-draggable.ts
zoran995: if we don't specify handleSelector, it will allow the dragging of an entire element. In this case entire feature info panel would cause dragging effect

na9da (commented):
LGTM.

na9da (approved):

PR #7547, opened 2025-03-11
Replace request with fetch in ci-cleanup

ljowen:
### What this PR does

Remove `request` as a dependency, use node's native fetch (introduced node v18)

### Test me

ci-deploy action runs successfully

### Checklist

### DIFF
--- a/buildprocess/ci-deploy.sh
+++ b/buildprocess/ci-deploy.sh
@@ -19,7 +19,6 @@ gh api /repos/${GITHUB_REPOSITORY}/statuses/${GITHUB_SHA} -f state=pending -f co
 # Install some tools we need from npm
 npm install -g https://github.com/terriajs/sync-dependencies
 npm install -g yarn@^1.19.0
-yarn add -W request@2.83.0
 
 # Clone and build TerriaMap, using this version of TerriaJS
 TERRIAJS_COMMIT_HASH=$(git rev-parse HEAD)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -47,6 +47,7 @@
 - Remove unused `arraysAreEqual`, `autoUpdate`, `flattenNested`, `freezeInDebug`, `isPromise`, `loadJsonp`, `OrUndefined`, and `superGet` files from lib/Core.
 - Update to react-virtual 2.10.4.
 - Update types/file-saver to 2.0.7.
+- Remove `request` dependency from CI scripts
 - [The next improvement]
 
 #### 8.8.1 - 2025-02-27
--- a/buildprocess/ci-cleanup.js
+++ b/buildprocess/ci-cleanup.js
@@ -1,18 +1,5 @@
 const childProcess = require("child_process");
 const fs = require("fs");
-const request = require("request");
-
-function requestPromise(options) {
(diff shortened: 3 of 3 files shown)


buildprocess/ci-deploy.sh
```diff
@@ -19,7 +19,6 @@ gh api /repos/${GITHUB_REPOSITORY}/statuses/${GITHUB_SHA} -f state=pending -f co
 # Install some tools we need from npm
 npm install -g https://github.com/terriajs/sync-dependencies
 npm install -g yarn@^1.19.0
-yarn add -W request@2.83.0
```
pjonsson, line 22: The current error looks like the same error that caused #7522.

Can you try to remove the `rm yarn.lock` on line 34 and see if that re-uses the compatible dependencies specified in TerriaMap's `yarn.lock` and only updates the incompatible dependencies?

I've put up a fix to pin the types package to the previous version in #7551, but since we now have an easy way to reproduce the current failure I suggest we first fix the CI problem.

ljowen: Removing `L:34` fixed this particular issue issue, I'm not sure the rational for deleting the lockfile @na9da  @zoran995 any thoughts? Would this cause issues if for example a deploy introduces a new dependency in `terriajs` ?

pjonsson: Ok, so we know keeping `yarn.lock` at least works for this case. I'm guessing the reason it is deleted is because commit 0ce0fde introduced removing the file, but that was for npm.

I've removed the draft status from #7551 so it can be merged.

I have some other commits related to CI in the PR queue, I'm guessing it would be easiest to cherry-pick them into this PR to get the deploy workflow tested. Commit c38759c0 uses the yarn that is pre-installed on Github's CI runners, commit 7c50852 adds `--frozen-lockfile` so `package.json` and `yarn.lock` are kept in sync, and commit 4701385 stops the publish action if it somehow runs with an incompatible Node version.

na9da: I think `rm yarn.lock` is intentional so that we can  discover compatibility issues with newer versions of dependencies sooner (it is ok for the CI to break). When that happens we either fix the code or pin the version for the dependency in package.json (like @pjonsson has done).

pjonsson: If I knew nothing about this project, I would first look for some job in TerriaMap for finding those breakages, but if you want the CI to sometimes go red for reasons unrelated to the PR I guess that's your choice.

More recent versions of flexsearch include types so I guess upgrading to those is the long-term solution, my pin is just to unbreak the CI.

zoran995: I can test if everything works on CI with the react upgrade branch, as it is adding new dev dependencies and upgrading a couple of them. I couldn't make it pass locally but will make some tests on CI. 
@na9da mayde a good point on detecting dependency issues as early as possible but if everything works correctly perhaps we could have a timed action to run that check (i.e. every week or similar) but we can play with it later
p.s. I also noticed we are installing moment 2.24 before build but as we are using newer version in terriajs we might also try removing that also

zoran995 (approved):
LGTM

PR #7563, opened 2025-03-20
covert drag-drop-file to functional component

zoran995:
### What this PR does

Part of #7519 

### Test me

### Checklist


lib/ReactViews/DragDropFile.tsx
```diff
@@ -12,36 +12,27 @@ import Result from "../Core/Result";
...
-class DragDropFile extends Component<PropsType> {
-  target: EventTarget | undefined;
-
-  constructor(props: PropsType) {
-    super(props);
-    makeObservable(this);
-  }
+const DragDropFile: FC = observer(() => {
+  const viewState = useViewState();
+  const targetRef = useRef<EventTarget>();
```
pjonsson, line 58: I think I've seen something that some future React version requires an argument to `useRef()`, is it worth adding the `(null)` argument now since the code is already being changed here?

lib/ReactViews/DragDropFile.tsx
```diff
@@ -97,72 +85,65 @@ class DragDropFile extends Component<PropsType> {
...
+      className={classNames(Styles.dropZone, {
+        [Styles.isActive]: viewState.isDraggingDroppingFile
+      })}
+    >
+      {viewState.isDraggingDroppingFile ? (
+        <div className={Styles.inner}>
+          <Trans i18nKey="dragDrop.text">
+            <h3 className={Styles.heading}>Drag & Drop</h3>
+            <div className={Styles.caption}>
+              Your data anywhere to view on the map
```
pjonsson: Since you seem to be the expert on translation, should this text have a translation?

zoran995: It is translated there is a `Trans` component around it

lib/ReactViews/DragDropFile.tsx
zoran995: yeah it is in react 19, will handle that when we get to upgrading to react 19

zoran995:
Converted to draft as drag and drop notification will be merged into this

na9da (commented):
LGTM. I'll wait for the DragDropNotification refactor to finish review. Would you consider taking out those two `.scss` files and replacing with styled-components if not too difficult?

lib/ReactViews/DragDropFile.tsx
```diff
@@ -97,72 +126,79 @@ class DragDropFile extends Component<PropsType> {
...
-      </div>
-    );
-  }
-}
-
-export default withTranslation()(withViewState(DragDropFile));
+          </Box>
+        ) : null}
+      </DropZone>
+      <DragDropNotification uploadedFiles={lastUploadedFiles} />
```
na9da, line 198: Ideally we would render `DragDropNotification` only if `lastUploadedFiles` is not empty but I guess here we need it so that the CSS slide in transition works.

na9da (approved):
LGTM @zoran995 - thanks for the refactor.

lib/ReactViews/DragDropFile.tsx
zoran995: Yeah, we need it for the styling

PR #7573, opened 2025-03-26
Start lock down stratum work

nf-s:
This is an experiment to see what would be needed in order to lock down the use of "loadable" strata.

It also just an exercise for me to go through and review how loadable strata are used.

In my mind strata should only be used to populate trait values. From the [docs](https://docs.terria.io/guide/contributing/model-layer/#types-of-classes-in-the-model-layer):

> Strata that provide values for a subset of the properties defined on model's traits class, usually by loading them from an external source such as WMS GetCapabilities service

In many cases, we also access loadable stratum properties that don't apply to traits (for example - `WebMapServiceCatalogItem.validLayers` - https://github.com/TerriaJS/terriajs/blob/30654f1a33b1af79afe5843b99855d67cc3c3667/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts#L259-L274)

Most usage of loadable strata like this can be easily refactored by creating extra traits - for example

https://github.com/TerriaJS/terriajs/blob/1cc022d680b1fb61937f93b9ffefbdfe82cb62c2/lib/Traits/TraitsClasses/WebMapServiceCatalogItemTraits.ts#L227-L233

Which means we can now avoid using the stratum directly

https://github.com/TerriaJS/terriajs/blob/1cc022d680b1fb61937f93b9ffefbdfe82cb62c2/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts#L260-L272

There are also a number of patterns that end up using these lodable strata properties directly

- A lot of "dynamic" groups (eg CKAN or OpenDataSoft or WMS group) have a `createMembers` function that sits in a loadable stratum
- Use of `discreteTimes` is also usually from within a loadable stratum

Future work

- I would also be interesting in seeing if we could create a "cachable" loadable stratum - that is - a loadable stratum that has no reactivity.
  - This could then be used to cache expensive operations - for example, loading and parsing WMS `GetCapabilities` on a server with many layers - or loading CKAN datasets

## Todo

- [ ] Fix `WebMapTileServiceCatalogItem`
- [ ] Fix tests that inspect strata values directly
- [ ] Test everything


nf-s:
Closing as this experimental PR is not needed

PR #7574, opened 2025-03-26
convert viewing controls to functional component

zoran995:
### What this PR does

Convert viewing controls to a functional component.

### Test me

### Checklist

### DIFF
--- a/lib/ReactViews/Workbench/Controls/ViewingControls.tsx
+++ b/lib/ReactViews/Workbench/Controls/ViewingControls.tsx
@@ -1,8 +1,8 @@
 import { sortBy, uniqBy } from "lodash-es";
-import { action, computed, runInAction, makeObservable } from "mobx";
+import { runInAction } from "mobx";
 import { observer } from "mobx-react";
-import { Component } from "react";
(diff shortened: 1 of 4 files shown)


lib/ReactViews/Workbench/Controls/ViewingControls.tsx
```diff
@@ -88,59 +88,42 @@ const ViewingControlMenuButton = styled(RawButton).attrs({
...
 
-    makeObservable(this);
-
-    // Set the state directly. Use props if necessary.
-    this.state = {
-      isMapZoomingToCatalogItem: false
+  useEffect(() => {
+    const hideMenu = () => {
+      runInAction(() => {
+        viewState.workbenchItemWithOpenControls = undefined;
```
na9da: Suggestion: If not too difficult are you able to make this a local state instead? `workbenchItemWithOpenControls` doesn't get used anywhere else.

na9da (approved):
Looks good @zoran995. One comment which I leave to you to decide. 

Maybe in the future, we could break this component up. One way of doing that is to define a list similar to `globalViewingControls` with standard viewing controls and `ViewingControls.tsx` component would simply render that list.

PR #7577, opened 2025-03-26
tsify dropdown

zoran995:
### What this PR does

TSify dropdown and convert it to a functional component

### Test me

### Checklist


lib/ReactViews/Generic/Dropdown.tsx
```diff
@@ -0,0 +1,168 @@
...
+  selectOption,
+  textProperty = "name" as P,
+  disabled = false,
+  children
+}: DropdownProps<P, T>) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const scrollListeners = useRef<Element[]>([]);
+
+  const hideList = useCallback(() => {
```
ljowen, line 53: Is this a premature optimization? It prevents a function allocation each render but given the way dropdowns are used I'm not sure the extra complexity is worth it

zoran995: It is used in useEffect for scroll listeners, so we need useCallback to prevent listeners from being removed and added on every render. We could eliminate those two functions and inline them everywhere, but that would mean actions are not separate/isolated.

ljowen (approved):

PR #7585, opened 2025-04-02
Convert date time picker to functional component and split into multiple components

zoran995:
### What this PR does

Refactor the DateTimePicker component to a functional component and split it into multiple smaller components. Haven't updated `react-datepicker` in this PR as it has issues with React types and expects types of the later version. Will update that as part of the react-upgrade PR. 

### Test me

DateTimePicker should work the same as on main. The easiest way to verify is to check the component integration tests as they are first written for old implementation and then new implementation is done
Before: http://ci.terria.io/main/#share=s-a4U7pw4eu6ZdDfCi5omNXtCipIl
After: http://ci.terria.io/date-time-picker/#share=s-a4U7pw4eu6ZdDfCi5omNXtCipIl

### Checklist


lib/ReactViews/BottomDock/Timeline/DatePicker/DateTimePicker.styles.ts
na9da, line 1: Nitpick: `.styles.ts` would be a new convention. I think it might be better to name it simply `DateTimePickerStyles.ts` or  `DateTimePickerElements` or similar.

lib/ReactViews/BottomDock/Timeline/DatePicker/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  "Nov",
+  "Dec"
+];
+
+export const daysInMonth = (month: number, year: number) => {
+  const n = new Date(year, month, 0).getDate();
+  return (Array.apply as any)(null, { length: n }).map(
+    Number.call,
+    Number
+  ) as number[];
```
na9da: Suggestion: I understand this comes from the previous code. We could probably pick an alternative suggested in this [stackoverflow answer](https://stackoverflow.com/questions/3746725/how-to-create-an-array-containing-1-n). I like: `[...Array(n).keys()]`

na9da (changes requested):
Hi @zoran995, the changes look great.

I tested the difference tool workflow and the date picker gets clipped at the bottom. Could you have a look at that please?

Here's an image:
image

You can test it [from this URL](http://ci.terria.io/date-time-picker/#configUrl=https://terria-catalogs-public.storage.googleapis.com/de-australia/map-config/prod.json) and trigger the tool from the workbench item menu.

image

zoran995:
@na9da this is ready for another pass

na9da (approved):
LGTM. Thanks for the refactor @zoran995!

PR #7178, opened 2024-06-04
Extra personalization of internationalization

arte-dev:
### What this PR does

Allow to modify `lookupCookie` for i18next. This is in line with current implementation, but it certainly would be better if all options could be overwritten. This is needed to be able to take advantage of the existing domain cookie that manages the language

### Test me

Use cookie on changeLanguageOnStartWhen and change lookupCookie to the desired value. Change the language and you'll see the proper value on the cookie with name lookupCookie

### Checklist

- [x] Unit tests aren't applicable, 
- [x] I've updated relevant documentation in `doc/`.
- [x] I've updated CHANGES.md with what I changed.
- [x] I've provided instructions in the PR description on how to test this PR.

PR #7593, opened 2025-04-14
I18n lookup cookie

zoran995:
### What this PR does

Superseedes #7178

Allow to modify lookupCookie for i18next. This is in line with current implementation, but it certainly would be better if all options could be overwritten. This is needed to be able to take advantage of the existing domain cookie that manages the language

### Test me

Use cookie on changeLanguageOnStartWhen and change lookupCookie to the desired value. Change the language and you'll see the proper value on the cookie with name lookupCookie

### Checklist


na9da (approved):
Thanks @zoran995. It might be a good practice update the PR description too just in case we come back in the future and want a quick summary.

PR #7598, opened 2025-04-16
Feature info chart

zoran995:
### What this PR does

Part of #7519

Convert `Chart` component of `FeatureInfoPanelChart` to functional component.

### Test me

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,8 @@
 - DOMPurify updated to 3.2.5 to fix CVE-2025-26791.
 - Fix pmtiles rendering issue. #7607
 - Remove unused types/rbush dependency.
+- Convert `Chart` component of `FeatureInfoPanelChart` to functional component.
+- [The next improvement]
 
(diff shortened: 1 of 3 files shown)


niktayv (approved):
LGTM.

I compared the UI "before and after" and it looked identical.
For the future tasks, we should probably improve it a little, as there are minor issues in styling:
Screenshot 2025-04-30 at 8 32 13 PM


I wonder (for my personal enlightenment) why switching to `useMemo` is more beneficial than using `computed`? Is this something that we all agreed on?

zoran995:
> I wonder (for my personal enlightenment) why switching to `useMemo` is more beneficial than using `computed`? Is this something that we all agreed on?

There is a lot of discussion online regarding usage of mobx computed and memoization and there is never a definite answer. Here, unless I have forgot something, the chang is because we don't get any benefit from mobx computed. And if I recall correctly computed used this way will always return new instance of ComputedValue, and without memo it will happen on every render

PR #7608, opened 2025-04-23
Fix 7607 pbf tiles

### LINKED ISSUE #7607: Protomaps basemap configuration in TerriaJS
G'day team,

Hope all is well. The TerriaJS upgrade looks great!

I am currently trying to configure a Protomaps basemap for DTV using a .pmtiles file hosted in a publicly accessible cloudflare R2 bucket. I have verified it is set up correctly by building a quick maplibre powered map using the pmtiles file as a basemap.

I tried the following two methods:
(Method 1 has a giant layer styles json that is too large to print here.)

```
{
  "catalog": [
    {
      "id": "1234",
      "name": "mvt example",
      "type": "mvt",
      "url": "https://<account-id>.r2.dev/australia.pmtiles",
      "keepOnTop": true,
      "layer": "light",
      "opacity": 1,
      "style": {
        "version": 8,
        "glyphs": "https://protomaps.github.io/basemaps-assets/fonts/{fontstack}/{range}.pbf",
        "sprite": "https://protomaps.github.io/basemaps-assets/sprites/v4/light",
        "sources": {
          "protomaps": {
            "type": "vector",
            "url": "pmtiles://https://<account-id>.r2.dev/australia.pmtiles",
            "maxzoom": 14
          }
        },
        "layers": [
          {
            "id": "background",
            "type": "background",
            "paint": {
              "background-color": "#cccccc"
            }
          },
etc...
```
Image

Another method I tried but didn't work was the following:
```
{
  "catalog": [
    {
      "id": "1234",
      "name": "mvt example",
      "type": "mvt",
      "url": "https://<account-id>.r2.dev/australia.pmtiles",
      "keepOnTop": true,
      "layer": "light"
    }
  ]
}
```

I wasn't able to find much documentation of configuring pmtiles into TerriaJS, and wondered if you have any sample config files handy? 

Thanks for your help!
Cheers,
Alex

niktayv:
### What this PR does

Fixes #7607

This PR fixes a regression that occurred within #7370.
Credits to @nf-s for helping to find a quick fix.

### Test me

Use the default catalog in TerriaMap or any other one with the following item types:
- CSV (with Region Mapping)
- MapboxVectorTIleCatalogItem
- GeoJsonCatalogItem
- ArcGisFeatureServer 

See [examples](https://github.com/TerriaJS/terriajs/issues/4937).

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Change Log
 
 #### next release (8.9.4)
+
 - Fix pmtiles rendering issue. #7607
 
 #### next release (8.9.3)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,5 @@
 # Change Log
 
-#### next release (8.9.4)
-
-- Fix pmtiles rendering issue. #7607
-
 #### next release (8.9.3)
 
 - Remove unused d3-array dependency.
@@ -20,6 +16,7 @@
 - TSify FadeIn component. #7590
 - Convert `ViewingControls` to a functional component. #7574
 - DOMPurify updated to 3.2.5 to fix CVE-2025-26791.
+- Fix pmtiles rendering issue. #7607
 - [The next improvement]
 
 #### 8.9.2 - 2025-03-31
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Change Log
 
+#### next release (8.9.4)
+- Fix pmtiles rendering issue.
+
 #### next release (8.9.3)
 
 - Remove unused d3-array dependency.
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # Change Log
 
 #### next release (8.9.4)
-- Fix pmtiles rendering issue. #7606
+- Fix pmtiles rendering issue. #7607
 
 #### next release (8.9.3)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # Change Log
 
 #### next release (8.9.4)
-- Fix pmtiles rendering issue.
+- Fix pmtiles rendering issue. #7606
 
 #### next release (8.9.3)
 
--- a/lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
+++ b/lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
@@ -255,7 +255,7 @@ export default class ProtomapsImageryProvider
     const coords: Coords = { z: level, x, y };
 
     // Adapted from https://github.com/protomaps/protomaps.js/blob/master/src/frontends/leaflet.ts
-    let tile: PreparedTile;
+    let tile!: PreparedTile;
 
     // Get PreparedTile from source or view
     // Here we need a little bit of extra logic for the ProtomapsGeojsonSource
@@ -277,13 +277,20 @@ export default class ProtomapsImageryProvider
         dim: this.tileWidth
       };
     } else if (this.view) {
-      tile = await this.view.getDisplayTile(coords);
+      try {
+        tile = await this.view.getDisplayTile(coords);
+      } catch (e) {
+        if (process.env.NODE_ENV !== 'production') {
+          console.log(e);
+        }
+      }
     } else {
       throw TerriaError.from(
         `Failed to get tile - no view or appropriate source in ProtomapsImageryProvider`
       );
     }
 
+    if (!tile) return canvas;
     const tileMap = new Map<string, PreparedTile[]>().set("", [tile]);
 
     this.labelers.add(coords.z, tileMap);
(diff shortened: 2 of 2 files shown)


lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -274,13 +277,20 @@ export default class ProtomapsImageryProvider
...
+          console.log(e);
+        }
+      }
     } else {
       throw TerriaError.from(
         `Failed to get tile - no view or appropriate source in ProtomapsImageryProvider`
       );
     }
 
+    if (!tile) return canvas;
```
nf-s: I would leave this as is, as we need errors to flow through to how Cesium handles `ImageryProvider` tile errors

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -84,6 +84,9 @@ export const PROTOMAPS_DEFAULT_TILE_SIZE = 256;
 /** Buffer (in pixels) used when rendering (and generating - through geojson-vt) vector tiles */
 export const PROTOMAPS_TILE_BUFFER = 32;
 
+/** Tile cache size for protobuf-leaflet */
```
nf-s: The library here is `protomaps-leaflet`, I'm not sure it is a "tile cache size" - If you look into the library it seems to have something to do with Vector tile layer `extent`.

There is more info on that here- https://github.com/mapbox/vector-tile-spec/blob/master/2.1/README.md

> A layer MUST contain an extent that describes the width and height of the tile in integer coordinates. The geometries within the Vector Tile MAY extend past the bounds of the tile's area as defined by the extent. Geometries that extend past the tile's area as defined by extent are often used as a buffer for rendering features that overlap multiple adjacent tiles.
> 
> For example, if a tile has an extent of 4096, coordinate units within the tile refer to 1/4096th of its square dimensions. A coordinate of 0 is on the top or left edge of the tile, and a coordinate of 4096 is on the bottom or right edge. Coordinates from 1 through 4095 inclusive are fully within the extent of the tile, and coordinates less than 0 or greater than 4096 are fully outside the extent of the tile. A point at (1,10) or (4095,10) is within the extent of the tile. A point at (0,10) or (4096,10) is on the edge of the extent. A point at (-1,10) or (4097,10) is outside the extent of the tile.

You can see the `tileSize` passed into the `TileCache` used here - https://github.com/protomaps/protomaps-leaflet/blob/7f4a3c1c23fa1b0f1ce50a020e4f80c03dfcb4f1/src/tilecache.ts#L115

I think there is something funky going on with the scaling - it seems to be working though, I'll make a ticket to look into this later.

For now, can you call this `TILE_CACHE_TILE_SIZE`

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
```diff
@@ -252,7 +255,7 @@ export default class ProtomapsImageryProvider
     const coords: Coords = { z: level, x, y };
 
     // Adapted from https://github.com/protomaps/protomaps.js/blob/master/src/frontends/leaflet.ts
-    let tile: PreparedTile;
+    let tile!: PreparedTile;
```
nf-s: Also we should avoid using the bang operator like this

nf-s (changes requested):
Thanks @niktayv !

Please see review comments, also can you please add a line to `CHANGES.md`

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
niktayv: Thank you for the clarification! I updated the code.

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
niktayv: The idea was to handle the error shown in the console:
Screenshot 2025-04-24 at 1 54 30 PM
We can address this later.
I reverted this commit.

lib/Map/ImageryProvider/ProtomapsImageryProvider.ts
niktayv: This was a quick attempt to tame TypeScript after wrapping the `await` into a `try ... catch` block.
I'd love to know what would be the best way to so.
Is it okay to leave promises unhandled?

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Change Log
 
+#### next release (8.9.4)
+
+- Fix pmtiles rendering issue. #7607
+
 #### next release (8.9.3)
 
 - Remove unused d3-array dependency.
```
nf-s: You can put the new change log under `next release (8.9.3)` here

nf-s (approved):
Looks good thanks @niktayv 

I just have one small comment about `CHANGES.md`

CHANGES.md
niktayv: Oh, thanks, @nf-s. I did not know what to do here, so it was my wild guess :)

PR #7615, opened 2025-04-30
remove core-js and regenerator-runtime polyfills

zoran995:
### What this PR does

Recent CI-deployment failed due to a dependency issue for regenerator-runtime. 
Facebook archived regenerator-runtime in February, and generators are now widely supported, so we should be safe to drop it. 
Manual imports of core-js polyfills are not needed when `@babel/preset-env` is used with `"useBuiltIns": "usage"`. The comment in the file describes that the only use case was when a dependency utilised some unsupported functionality and Babel couldn't detect its usage. I had a quick run of `core-js-compat` to see what part of polyfills we need and all of them seem no longer relevant (either implemented by major browsers or won't be implemented by spec, or still not used by anyone).

### Test me

Verify that the app functions normally.

### Checklist


ljowen (approved):

ljowen:
- [x] Update CHANGES.md

package.json
```diff
@@ -205,7 +204,6 @@
     "pretty-quick": "^4.0.0",
     "react-shallow-testutils": "^3.0.0",
     "react-test-renderer": "^16.3.2",
-    "regenerator-runtime": "^0.13.2",
```
pjonsson, line 208: Would be great to get these changes into `yarn.lock` as well.

pjonsson (approved):
I've had this patch in my pile of unsubmitted PRs for a while, I think this is a great idea.

PR #7620, opened 2025-05-03
exclude code attributions from docs search

zoran995:
### What this PR does

Stops indexing and searching inside the attributions docs. https://squidfunk.github.io/mkdocs-material/plugins/search/#meta.search.exclude

### Test me

Open the CI deployment of docs and try searching for some code attribution

### Checklist


gulpfile.js
```diff
@@ -132,12 +132,27 @@ function runKarma(configFile, done) {
...
+const attributionTemplate = `"---
+search:
+  exclude: true
+---
+
+# Attributions"`;
+
 gulp.task("code-attribution", function userAttribution(done) {
   var spawnSync = require("child_process").spawnSync;
+  spawnSync(
```
pjonsson: This is missing a return value check (things can always fail for obscure reasons, but it's quite easy to get wrong owner on files when using docker containers and then suddenly one gets a permission denied outside the container, or similar).

I don't know how difficult it is to write a few lines to file in Node.js, but that would eliminate fork+exec and a bunch failure scenarios, and I suspect the error messages for when things do go wrong would be easier to understand. (Not suggesting to do anything with the spawnSync on line 153 since that runs something much more complicated than echoing a few lines that are identical on each invocation.)

ljowen (approved):

PR #7637, opened 2025-05-28
Sharelink size limit error

### LINKED ISSUE #7636: Display a warning when share data exceeds the limit
When the user adds a significant amount of the content to the story (it can be too many scenes or a huge content in one scene), the size of the payload in the `/share` endpoint may exceed the limit set on the server.

Currently, the user experiences a quiet failure of the URL shortening feature, getting an unshortened URL in the clipboard.
No warning or error message is displayed to the user.

Besides that, the SharePanel's UI gets scrambled both in the StoryBuilder and in the main menu's "Share / Print" panel.

Image

Image

We want to let the user know that the limit has been exceeded.

Additionally, we want to warn the user that the size has exceeded 80% of the limit but still stays within the limit, so that the user may review the content of the story in advance.

niktayv:
### What this PR does

Fixes #7636

### Test me

Testing this requires the commits from https://github.com/TerriaJS/terriajs-server/pull/179 to be in the server.
UPDATE: the client will use the default value of `shareMaxRequestSize` if it is not set on the server.

To trigger the error, create a story and add some content to the description which is large enough to exceed the limit.
For example, copy and paste (with formatting) the entire page https://en.wikipedia.org/wiki/New_Zealand which is about 1MB in size.

Before: http://ci.terria.io/main
- paste a large page, e..g. [this one](https://en.wikipedia.org/wiki/New_Zealand) with formatting;
- press the "Share" button in the Story Builder or the "Share / Print" button in the main menu or "Submit" button (with "Share..." checkbox turned on)
- observe the garbled UI as shown in https://github.com/TerriaJS/terriajs/issues/7636

After: http://ci.terria.io/sharelink-size-limit-error
- repeat the steps as above and observe the changes
### Checklist


test/Core/parseStringSizeSpec.ts
```diff
@@ -0,0 +1,54 @@
+import parseStringSize from "../../lib/Core/parseStringSize";
+
+describe("parseStringSize", function () {
+  it("Throws error for non-string input", function () {
+    // @ts-expect-error Testing runtime error for non-string input
```
pjonsson: Is it worth the increased code size/test time to have tests that verify that the typechecker in TypeScript works?

lib/Core/parseStringSize.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param sizeStr - The input string representing the size (e.g., "10KB", "5MB", "1GB").
+ *                  It must be a string containing a numeric value followed by an optional unit (KB, MB, GB, TB).
+ * @returns The size in bytes as a number, or undefined if the input is invalid.
+ * @throws An error if the input is not a string or if the format is invalid.
+ */
+function parseStringSize(sizeStr: string): number | undefined {
+  if (typeof sizeStr !== "string") {
+    throw new Error("Input must be a string");
+  }
+  const sizeUnitRegex = /(\d+)(KB|MB|GB|TB)?$/i;
```
pjonsson: Include `PB` just to be sure nobody bumps into the limit for the next couple of years?

test/Core/parseStringSizeSpec.ts
niktayv: @pjonsson This is a legitimate concern, for sure, but the whole reason I introduced this `parseStringSize` was to avoid pulling another dependency such as the `bytes` library.
Would it make sense to remove this test suite completely?
By the way, I did not notice a significant increase in the test time.

lib/Core/parseStringSize.ts
niktayv: I understand your point, @pjonsson, but my intention was to avoid coupling this helper semantically to the share data size use case.
I agree that it smells like a YAGNI violation, an it make sense to leave only KB and MB here.

test/Core/parseStringSizeSpec.ts
pjonsson: Oh, sorry, I just meant to suggest to remove the tests where you have to `ts-expect-error` since the input is the wrong type and ` tsc` would already catch those kind of problems.

If the other tests are slowing things down, I guess the question is how will performance be after the test tool migration is done next quarter. I don't know what the plans are, so someone else needs to make an educated guess about that.

lib/Core/parseStringSize.ts
pjonsson: I'm guessing files under `Core` should in theory be usable for other things too though?

My take was basically that supporting a couple of extra prefixes costs very little at the time of writing the code, and doing so will save you from later getting an issue where a user has found a clever way to use this code for something else.

test/Core/parseStringSizeSpec.ts
zoran995: I guess this value comes from server side where it isnt validated and server doesn't know about usage of it on frontend, so this is completely valid test

pjonsson: In a compiled language, the checks that don't apply to the specified type would most likely be optimized away, but we're not dealing with a compiled language here. What is the policy for type annotations in these scenarios, should they reflect that the input can be more or less anything?

lib/Core/parseStringSize.ts
niktayv: @pjonsson @zoran995 Should we simply demote that `parseStringSize` from `lib/Core` to a local function in `lib/Models/ShareDataService.ts` because I just use it once there?

When I needed it, I just quickly backed it with some AI help without much thinking.

Now I feel sorry for introducing this hasty abstraction (violating the AHA = Avoid Hasty Abstraction principle).

lib/Models/ShareDataService.ts
```diff
@@ -46,6 +49,18 @@ export default class ShareDataService {
     );
   }
 
+  // readonly shareMaxRequestSize: number = this._serverConfig.maxRequestSize || DEFAULT_MAX_SHARE_SIZE;
+  get shareMaxRequestSize(): number {
```
na9da: If the server doesn't return `shareMaxRequestSize` should we just return `undefined` here and disable the warning? There are situations where map owners might use custom server that does not return `shareMaxRequestSize`.

lib/Core/parseStringSize.ts
na9da, line 1: Suggestion: I think we could avoid parsing the size string client side if we make the server return the value in integer bytes instead. It is easier to do this server side by calling `bytes(sizestr)`. This also makes sense as we are not displaying the human readable string anywhere in the warnings. But even if we have to do that, i think it might be easier to convert from integer to human readable string (or make the server return both).

lib/ReactViews/Story/StoryBuilder.tsx
```diff
@@ -258,6 +259,21 @@ class StoryBuilder extends Component<
...
+  }
+
+  componentDidMount(): void {
+    this.updateShareDataStringSize();
+  }
+
+  componentDidUpdate(): void {
+    this.updateShareDataStringSize();
+  }
+
```
na9da: Instead of doing this in mount/update stages, could you try to `@compute` `shareDataStringSize` automatically when `viewState.terria.stories` change? This could be done by defining a method like:

```
  @computed
  get storiesSize() {
    const terria = this.props.viewState.terria;
    const stories = terria.stories;

    //Note:  We need to do this check to set up a dependency on individual scene data and ensure that storiesSize gets recomputed when any scene data changes.
    const validStories = stories.filter(
      (story) => story.shareData.initSources.length > 0
    ).length;

    return JSON.stringify(
      getShareData(terria, this.props.viewState, {
        includeStories: validStories > 0
      })
    ).length;
  }
```

This also removes the need to add new properties to `viewState` which is usually used to set some shared state for the UI.

wwwroot/languages/en/translation.json
```diff
@@ -354,7 +354,7 @@
   },
   "story": {
     "panelTitle": "Story editor",
-    "panelBody": "Create and share interactive stories directly from your map",
+    "panelBody": "Create and share interactive stories directly from your map. Each story is capped at 2 MB - split large data or many scenes into multiple stories to stay under the limit.",
```
na9da: Is it correct to assume 2MB here?

na9da (changes requested):
Hi @niktayv, great job navigating the complex feature. I have left a couple of recommendations. Let me know what you think.

wwwroot/languages/en/translation.json
niktayv: @na9da No, it was a hasty hard-coding. I am working on parametrising this properly. From your comments on getting the parsed and unparsed values from the server, I see a better solution now. Thanks!

lib/Models/ShareDataService.ts
niktayv: @na9da Yes, it makes a lot of sense. I will update my solution accordingly, taking into account your other comments. Thanks!

lib/ReactViews/Story/StoryBuilder.tsx
niktayv: @na9da Thank you for pointing this out, but I am a bit hesitant to change this because my intention was to share this state with some other parts of the UI, which I eventually decided to postpone. In fact, we will need to know that the size of shareData has exceeded the limit in a few more places, such as the main Share Panel and the Feedback Form. Should we just leave it as is for now and refactor when we return to this in future?

test/Core/parseStringSizeSpec.ts
niktayv: I changed the implementation to parsing on the server, as @na9da suggested: https://github.com/TerriaJS/terriajs-server/pull/179/commits/b80eb6bf25fd119052331dcd1d61b52094ce0d0b

lib/Models/ShareDataService.ts
niktayv: @na9da I've done that and tested.

lib/Core/parseStringSize.ts
niktayv: Done: https://github.com/TerriaJS/terriajs-server/pull/179/commits/b80eb6bf25fd119052331dcd1d61b52094ce0d0b

wwwroot/languages/en/translation.json
niktayv: Fixed.

lib/Core/parseStringSize.ts
niktayv: I removed this `parseStringSize` in favour of the server-side parsing with `bytes`: https://github.com/TerriaJS/terriajs-server/commit/b80eb6bf25fd119052331dcd1d61b52094ce0d0b

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.ts
```diff
@@ -87,11 +87,21 @@ export async function buildShortShareLink(
...
+  try {
+    token = await terria.shareDataService?.getShareToken(
+      getShareData(terria, viewState, options)
+    );
+  } catch (e) {
+    terria.raiseErrorToUser(e, {
+      title: "Error generating share token",
+      severity: TerriaErrorSeverity.Error
+    });
+    return "Error: Share data exceeds limit";
```
zoran995: Can a different error be thrown by the backend? i.e. authentication to s3 is not correct?
Also, if this is shown to the user, it should be i18n. You can use `applyTranslationIfExists` and prefix the key here with `translate#`

lib/ReactViews/Story/StoryBuilder.tsx
```diff
@@ -507,13 +531,23 @@ class StoryBuilder extends Component<
           </Text>
           <Spacing bottom={2} />
           <Text medium color={this.props.theme.textLightDimmed} highlightLinks>
-            {t("story.panelBody")}
+            {t("story.panelBody") +
+              (shareMaxRequestSize
+                ? " " + t("story.panelBodyCapped", { shareMaxRequestSize })
```
zoran995: nit: It might be easier to read this as a template literal string

lib/ReactViews/Story/StoryBuilder.tsx
zoran995: @niktayv, the idea by Nanda makes it more isolated from the StoryBuilder component, and more things might be used to trigger the computation in future. You can put a computed value as a local value in a component, and we can extract it to the viewState when needed

niktayv: @na9da @zoran995 Done. Thanks for your comments and suggestions.

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.ts
niktayv: @zoran995 I've attempted to "parametrize" error messages, but my solution looks quite clumsy to me. Could you please look again?
FYI: going with `applyTranslationIfExists` took me to a massive prop drilling, so I used `i18next.t()` instead.

lib/ReactViews/Story/StoryBuilder.tsx
niktayv: Thanks. Refactored.

lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.ts
```diff
@@ -87,11 +88,36 @@ export async function buildShortShareLink(
...
+          "models.shareData.generateErrorDataExceedsLimitMessage"
+        );
+        terria.raiseErrorToUser(
+          TerriaError.from(error, {
+            message: userMessage
+          }),
+          {
+            severity: TerriaErrorSeverity.Error
+          }
+        );
```
na9da: @niktayv - the new changes look good.

One problem i see here is that this function can be called from context where we don't want to automatically show an error pop up to the user. Eg: print view, feedback modal etc. 

Would it be possible to raise the error in [ShareUrl.tsx](https://github.com/TerriaJS/terriajs/blob/97cc2928b2880cd938cbe3820d755d80390098b9/lib/ReactViews/Map/Panels/SharePanel/ShareUrl/ShareUrl.tsx#L84-L86) instead?

niktayv: @na9da Thank you for pointing this out! I moved the logic to ShareUrl.tsx. Please see if that looks better.

na9da (approved):
LGTM. Thanks @niktayv!

PR #7639, opened 2025-06-02
set scalefactor=undefined when item trait=false

micholm:
### What this PR does

Cesium has a probable bug [here](https://github.com/CesiumGS/cesium/blob/9f116573ee0bfb9ad9c2d3ea7b7abe2bf39f15ec/packages/engine/Source/Scene/MapboxStyleImageryProvider.js#L91) concerning the `MapboxStyleImageryProvider`, where the `scaleFactor` option's null/undefined state is tested but not its boolean value. As a result, tiles are always scaled-up in Cesium, regardless of whether the trait has been set true or false. As a workaround fix, `MapboxStyleCatalogItem` should set the value of `scaleFactor` to `undefined` when false. 

### Test me

Add a `MapboxStyleCatalogItem` to your map's init file and toggle the `scaleFactor` trait between `true`, `false`, and not set, testing in 3D and 2D. Inspect the tile fetches to check that the `@2x` is present when `on` and absent when `off`.

### Checklist

- [x] No tests as test file does not exist for `MapboxStyleCatalogItem` and change is minor.
- [x] no`doc/` changes required--change is not public-facing.

### DIFF
--- a/lib/Models/Catalog/CatalogItems/MapboxStyleCatalogItem.ts
+++ b/lib/Models/Catalog/CatalogItems/MapboxStyleCatalogItem.ts
@@ -39,6 +39,7 @@ export default class MapboxStyleCatalogItem extends MappableMixin(
       styleId,
       accessToken,
       tilesize: this.tilesize,
+      //Cesium only checks if scaleFactor is definied or not.
       scaleFactor: this.scaleFactor ? this.scaleFactor : undefined,
       minimumLevel: this.minimumLevel,
       maximumLevel: this.maximumLevel,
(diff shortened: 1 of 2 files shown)


lib/Models/Catalog/CatalogItems/MapboxStyleCatalogItem.ts
```diff
@@ -39,7 +39,7 @@ export default class MapboxStyleCatalogItem extends MappableMixin(
       styleId,
       accessToken,
       tilesize: this.tilesize,
-      scaleFactor: this.scaleFactor,
+      scaleFactor: this.scaleFactor ? this.scaleFactor : undefined,
```
na9da, line 43: Would be good to leave a comment here stating why we need this.

na9da (approved):
LGTM @micholm . Just one comment.

PR #7646, opened 2025-06-05
Update ExportData.tsx

irbian:
i18n a fixed value

### Test me

Go to dataset preview page and ensure "Export data" changes with each language

### Checklist

- [x] Unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
- [ ] I've updated relevant documentation in `doc/`. (No relevant documentation)
- [ ] I've updated CHANGES.md with what I changed. (No relevant changes)

### DIFF
--- a/lib/ReactViews/Preview/ExportData.tsx
+++ b/lib/ReactViews/Preview/ExportData.tsx
@@ -49,7 +49,7 @@ class ExportData extends Component<PropsType> {
     return (
       <div className={Styles.metadata}>
         <button onClick={this.exportDataClicked.bind(this, catalogItem)}>
-          Export data
+          {this.props.t("workbench.exportData")}
         </button>
       </div>
     );


lib/ReactViews/Preview/ExportData.tsx
```diff
@@ -49,7 +49,7 @@ class ExportData extends Component<PropsType> {
     return (
       <div className={Styles.metadata}>
         <button onClick={this.exportDataClicked.bind(this, catalogItem)}>
-          Export data
+          {this.props.t("workbench.exportData")}
```
zoran995, line 52: It would be better to create a new translation key e.g. `description.exportData` or `preview.exportData` to better reason on where it is used. Also, the value of this translation key is Export, and the expected string in the UI is `Export data`.

zoran995 (changes requested):

PR #7649, opened 2025-06-11
Update pm leaflet

ljowen:
### What this PR does

Update `protomaps-leaflet`

### Test me

View a pmtiles dataset 

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "proj4": "^2.4.4",
     "proj4-fully-loaded": "^0.2.0",
     "prop-types": "^15.6.0",
-    "protomaps-leaflet": "^4.1.1",
+    "protomaps-leaflet": "^5.0.1",
     "rc-slider": "^9.7.2",
     "react": "^16.14.0",
     "react-anything-sortable": "^1.5.2",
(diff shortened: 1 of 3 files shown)


micholm (approved):

package.json
```diff
@@ -24,6 +24,7 @@
     "@mapbox/geojson-merge": "^1.1.1",
     "@mapbox/point-geometry": "^0.1.0",
     "@mapbox/togeojson": "^0.16.2",
+    "@maplibre/maplibre-gl-style-spec": "^23.3.0",
```
zoran995, line 27: We need this because of protomaps-leaflet upgrade, or we will use it directly in some places?

zoran995 (commented):
Can you also update the changelog? Do you know if we need to add a note about breaking changes in protomaps-leaflet, since it now only supports the protomaps basemap tileset v4.0? I'm not familiar with it, but could it break some of the existing maps?

package.json
ljowen: We need this for types used by protomaps leaflet, unfortunately there isn't a DefiniteyTyped package. I've updated the changelog

PR #7653, opened 2025-06-25
Add mapboxsearchprovider

micholm:
### What this PR does

This PR adds a search provider for Mapbox geocoder. This includes forward and reverse geocoding.

### Test me

test by adding this search provider to the config.json and testing that different parameters are working as they should.

For reverse geocoding, you can test by pasting in a "lat, lon", or, open google maps, right click and copy the coords, then paste these in the search bar.

### Checklist


lib/Traits/SearchProviders/MapboxSearchProviderTraits.ts
```diff
@@ -0,0 +1,92 @@
...
+  @primitiveTrait({
+    type: "boolean",
+    name: "Autocomplete",
+    description: `Specify whether to return autocomplete results (true, default) 
+    or not (false). When autocomplete is enabled, results will be included that 
+    start with the requested string, rather than responses that match it exactly. 
+    For example, a query for India might return both India and Indiana with 
+    autocomplete enabled, but only India if it’s disabled.`
+  })
+  autocomplete?: boolean = true;
```
zoran995: Can we name this exactMatch or similar, as we might need autocomplete for different things (i.e. nominatim doesn't allow autocompleting the search results - search as you type)

lib/Models/Terria.ts
```diff
@@ -546,6 +547,7 @@ export default class Terria {
     cesiumIonLoginTokenPersistence: "page",
     cesiumIonAllowSharingAddedAssets: false,
     bingMapsKey: undefined,
+    mapboxSearchProviderAccessToken: undefined,
```
zoran995: I don't think there is a need for this, for Bing Maps we used the one from the config, so users don't need to change it in multiple places

lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -0,0 +1,247 @@
...
+          CommonStrata.defaults,
+          "accessToken",
+          this.terria.configParameters.mapboxSearchProviderAccessToken
+        );
+      }
+      if (this.terria.configParameters.locationSearchBoundingBox) {
+        this.setTrait(
+          CommonStrata.defaults,
+          "bbox",
+          this.terria.configParameters.locationSearchBoundingBox.join(", ")
```
zoran995: This `locationSearchBoundingBox` is not used anywhere else in the codebase and probably should be removed. There is a `boundingBoxLimit` on `searchBarModel`, can we use that?

lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -0,0 +1,247 @@
...
+          return;
+        }
+
+        const resourceSet = result.features[0];
+        if (resourceSet.properties.length === 0) {
+          searchResults.message = {
+            content: "translate#viewModels.searchNoLocations"
+          };
+          return;
+        }
```
zoran995: I don't think this is needed; you take the first result and check its properties length, but properties is an object

lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -0,0 +1,247 @@
...
+
+      searchQuery.appendQueryParameters({
+        proximity: `${mapCenter.longitude}, ${mapCenter.latitude}`
+      });
+    }
+
+    if (this.bbox) {
+      searchQuery.appendQueryParameters({
+        bbox: this.bbox
+      });
```
zoran995: does reverse geocoding support those two options

lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -0,0 +1,247 @@
...
+      /([+-]?\d+\.?\d+)\s*,\s*([+-]?\d+\.?\d+)/
+    );
+    let searchDirection = isCSCoordinatePair.test(searchText)
+      ? MapboxGeocodeDirection.Reverse
+      : MapboxGeocodeDirection.Forward;
+
+    let queryParams = {
+      access_token: this.accessToken,
+      autocomplete: this.autocomplete,
+      language: this.language
```
zoran995, line 80: Can we include the current UI language and then use translations to show the result in the proper language? i.e. the one provided in settings is the default, but we also add the current UI language and then show the translated result.
We can also do it as a future improvement.

zoran995 (changes requested):

zoran995:
Are we required to show Mapbox attribution for geocoding? I can do it as part of the Nominatim improvements

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: I think this would be used if someone wanted to limit all search results to some geographic area of interest. I can remove it if you think it is not going to be used ever. It's probably not super critical to expose it.

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: Good point!

lib/Models/Terria.ts
micholm: One thing to consider is, do we want share it with mapbox basemaps as well? if that is the case then i could rename this to `mapboxAccessToken` and we can add mapbox basemaps as built in options in the future? Also happy to remove it otherwise.

lib/Traits/SearchProviders/MapboxSearchProviderTraits.ts
micholm: sure, maybe ill call it `partialMatch`

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: That would be great--ill take a look.

lib/Models/SearchProviders/MapboxSearchProvider.ts
zoran995: It is better to use `searchBarModel.boundingBoxLimit` as it is already used for that in other places. Here we are filtering results based on it https://github.com/TerriaJS/terriajs/blob/main/lib/ReactViews/Search/LocationSearchResults.tsx#L70. 

p.s. Everything related to search results should be part of the search model, unless it is shared with other places. Even in that case, the value should be applied to the search model and then used. The config parameter `locationSearchBoundingBox` hasn't been used for quite some time; we should be safe to remove it.

lib/Models/Terria.ts
zoran995: Let's discuss it internally, and we can add it in the follow-up if needed, so PR is scoped only to search part :smile:

micholm: awesome, will remove it from this PR

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: Im not sure how this would work at the moment, if you can point me in the right direction, or we can put it in the backlog? cheers.

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: please take a look at change.

micholm:
> Are we required to show Mapbox attribution for geocoding? I can do it as part of the Nominatim improvements

Ill need to take a look, thanks.

lib/Models/SearchProviders/MapboxSearchProvider.ts
zoran995: Let's add it to the backlog; it was more of an idea for better UX.
In theory, the current language can be retrieved as `i18next.language` and then you send it as part of the request `${this.language}, ${i18next.language}`. In that case, the response also contains a translations object, and you check if the current language exists in it with a fallback to the default language specified in the config.

lib/Traits/SearchProviders/MapboxSearchProviderTraits.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  @primitiveTrait({
+    type: "string",
+    name: "BBox",
+    description: `Limit results to only those contained within the supplied 
+  bounding box. Bounding boxes should be supplied as four numbers separated by 
+  commas, in minimum longitude,minimum latitude,maximum longitude,maximum latitude order. 
+  The bounding box cannot cross the 180th meridian.`
+  })
+  bbox?: string;
```
zoran995: This is no longer used after changing to the value from the search bar model. If you see the need, keep this and use the value from the search bar model as the default for it

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: I see, it would be nice; Ok let's put it in the backlog for now.

micholm:
@zoran995 It doesn't specifically mention geocoding in [the docs](https://docs.mapbox.com/help/dive-deeper/attribution/#when-youre-not-using-a-map), but I think to be safe we should add it where you're adding it for `Nominatim`.

Maybe something like 'Search results powered by `<provider>`' for example.

zoran995 (approved):
LGTM

PR #7656, opened 2025-06-30
Nominatim search update

zoran995:
### What this PR does

The Nominatim search provider is not recommended for production use, so warn the user in the console that it is not suitable. To be compliant with the Nominatim TOS, we need to disable the autocomplete search and also show attribution

### Test me

### Checklist


niktayv (approved):
- [x] All code changes LGTM.
- [x] I enabled Nominatim in my local TerriaMap and tested that everything worked as expected.
- [x] Running Jasmine SpecRunner showed no failures.

micholm:
It might be worth updating the example [here](https://github.com/TerriaJS/terriajs/blob/89d1ff91e11f55d6013a5db63d352512da3a9791/doc/customizing/search-providers.md?plain=1#L170) since autocomplete is not available.

lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -201,6 +208,14 @@ export default class MapboxSearchProvider extends LocationSearchProviderMixin(
             content: "translate#viewModels.searchNoLocations"
           };
         }
+        const attribution = result.attribution;
+        if (attribution) {
+          runInAction(() => {
+            this.setTrait(CommonStrata.underride, "attributions", [
```
micholm, line 214: Is setting this every time a query is returned necessary?

micholm:
Nice one @zoran995, Nominatim provider updates are working well, and attribution changes for Nominatim and MapBox look good. I've just left a couple of comments for you to consider, otherwise LGTM.

micholm (changes requested):
Nice one @zoran995, Nominatim provider updates are working well, and attribution changes for Nominatim and MapBox look good. I've just left a couple of comments for you to consider, otherwise LGTM.

lib/Models/SearchProviders/MapboxSearchProvider.ts
zoran995: This is how the Mapbox API is designed. We have no idea what they will return to us as attribution, so we must read it from the response object. i.e. the results for Australia might have a different attribution than the results for the US

micholm: I see what you mean, no problems then.

micholm (approved):
LGTM.  I would remove `"minCharacters": 3` from the json example in search-providers.md. But can be done easily outside of this PR.

nice one!

niktayv:
Zoran, if you decide to update the example in search-providers.md, it might make sense to use `maxResults` instead of `minCharacters`.

zoran995:
I don't think we should remove minCharacters from the example, as it is still enforced, it might be helpful for users who want to allow search with any number of characters. I added attributions to the doc

PR #7659, opened 2025-07-07
FeatureServer fixes

na9da:
### What this PR does

- Apply clipping rectangle in tiled mode. This reduces the number of requests made in tiled mode. 
- Adds `clipToRectangle` trait if we want to turn off the rectangle clipping
- Fix a few reactivity issues that resulted in reloading of layers when unrelated traits change.
- Proxy imageryprovider requests

### Test me

1. Fewer FeatureServer requests when loading layers in this [branch](http://ci.terria.io/fix-feature-server/#share=s-dmICFKSTjt2krfkvmxy6JJyLUqP&clean&https://gist.githubusercontent.com/nf-s/1a44b31faa1dd28dc9d23daa0660f50b/raw/ed093cfbae576e5fb0b209fa04711a5c99e55ea7/featuer-server-tile.json) vs [main](http://ci.terria.io/main/#share=s-dmICFKSTjt2krfkvmxy6JJyLUqP&clean&https://gist.githubusercontent.com/nf-s/1a44b31faa1dd28dc9d23daa0660f50b/raw/ed093cfbae576e5fb0b209fa04711a5c99e55ea7/featuer-server-tile.json). I found this makes some noticeable difference on slower servers.
2. Load [this layer in main](http://ci.terria.io/main/#share=s-dmICFKSTjt2krfkvmxy6JJyLUqP&clean&https://gist.githubusercontent.com/nf-s/1a44b31faa1dd28dc9d23daa0660f50b/raw/ed093cfbae576e5fb0b209fa04711a5c99e55ea7/featuer-server-tile.json), once fully loaded, click collapse all in the workbench, see that the layer gets re-loaded (due to change in unrelated traits). Try the same with [this branch](http://ci.terria.io/fix-feature-server/#share=s-dmICFKSTjt2krfkvmxy6JJyLUqP&clean&https://gist.githubusercontent.com/nf-s/1a44b31faa1dd28dc9d23daa0660f50b/raw/ed093cfbae576e5fb0b209fa04711a5c99e55ea7/featuer-server-tile.json), it shouldn't react to changes in unrelated traits.

### Checklist


lib/Models/Catalog/Esri/ArcGisFeatureServerCatalogItem.ts
```diff
@@ -136,15 +136,19 @@ export default class ArcGisFeatureServerCatalogItem extends MinMaxLevelMixin(
 
   @computed get imageryProvider() {
     // Don't return an imagery provider if we haven't loaded metadata yet
-    if (!this.loadMetadataResult) return undefined;
+    if (!this.strata.has(ArcGisFeatureServerStratum.stratumName)) {
+      return undefined;
+    }
```
na9da, line 141: `loadMetadatResult` returns a new `Result` object after each call to `loadMetadata` which causes this method to recompute unnecessarily. Instead check if the strata exists.

lib/Models/Catalog/Esri/ArcGisFeatureServerStratum.ts
na9da, line 1: Memoize `outFields` etc. This is also required to stop the imageryProvider from re-comping when for example a user strata gets added to the object.

nf-s (approved):
Looks good!

Only issue is that requests to https://services1.arcgis.com are now being proxied through TerriaJS-server - and it is very slow. So we need to add `arcgis.com` in `corsDomains` I think - or remove it from `allowProxyFor`

na9da:
Thanks @nf-s. Good point. I have created a PR here to remove arcgis.com from proxied list - https://github.com/TerriaJS/TerriaMap/pull/756

PR #7666, opened 2025-07-28
Mock TMS metadata to fix spec.

na9da:
### What this PR does

The recently add specs from [TMS PR](https://github.com/TerriaJS/terriajs/pull/7662) was breaking in spec runner when invoked from - http://localhost:3002/SpecRunner.html (but not in CI).

The `TileMapServiceImageryProvider` tries to fetch `tilemapresource.xml` and fails when it is not found. For some reason it works when karma runner is run explicitly from the CLI (like in CI) but not from the SpecRunner URL. This could be due to differences in HTTP error returned by both environment.

This PR adds a mock `tilemapresource.xml` which ensures the request does not fail and the specs succeed.

### Test me

- Build local terriajs checkout
- Make sure http://localhost:3002/SpecRunner.html does not show any errors

### Checklist

### DIFF
--- a/test/Models/Catalog/CatalogItems/TileMapServiceCatalogItemSpec.ts
+++ b/test/Models/Catalog/CatalogItems/TileMapServiceCatalogItemSpec.ts
@@ -10,7 +10,7 @@ describe("TileMapServiceCatalogItem", function () {
 
   beforeEach(function () {
     item = new TileMapServiceCatalogItem("test", new Terria());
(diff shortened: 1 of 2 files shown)


niktayv (approved):
- [x] Code changes LGTM: as far as I understand it, before, there was an `undefined` response; after: some value is provided.
- [x] Running http://localhost:3002/SpecRunner.html in Chrome, both in MacOS and Ubuntu, shows no failure.

PR #7669, opened 2025-07-29
WPS region picker

na9da:
### What this PR does

Port region picker input for WPS from terriajs v7.

The WPS input parameter is described as below by setting a special mime type:
```xml
<Input minOccurs="1" maxOccurs="1">
   <ows:Identifier>region</ows:Identifier>
   <ows:Title>Region to run analysis</ows:Title>
   <ows:Abstract></ows:Abstract>
   <ComplexData maximumMegabytes="10">
     <Default>
       <Format>
         <MimeType>application/vnd.terriajs.region+json</MimeType>
       </Format>
     </Default>
   </ComplexData>
</Input>
```

The `Execute` request will have the parameter filled as below:

```xml
<wps:Input>
  <ows:Identifier>region</ows:Identifier>
  <wps:Data>
    <wps:ComplexData>{"regionType":"SA2","region":"51203"}</wps:ComplexData>
  </wps:Data>
</wps:Input>
```

Note the JSON payload with keys `regionType` and `region`. The `region` corresponds to the value of `regionProp` property specified in the region map definition. For `SA2` it corresponds to `sa2_5dig11`.  Should we change this to property name instead?

This currently requires`WMS` type region mapping servers (or region maps with `analyticsWmsServer` configured).
See [example region mapping definition](https://gist.githubusercontent.com/na9da/ecc74ea91f884e0255b861499c228555/raw/113a85c7d761a2810385b1d38f4c007693d8d01a/nppRegionMaps.json).

### Test me

Steps:
1. Load [this CI link](http://ci.terria.io/wps-region-picking/#clean&https://gist.githubusercontent.com/na9da/f54c0bb049eba391f13421b04d8587bf/raw/f74525257d99ebcdbdc4616403995b8a31f2a238/wps-region-picking.json&configUrl=https://gist.githubusercontent.com/na9da/f54c0bb049eba391f13421b04d8587bf/raw/f93e59e81251c051d5c92d06397ecec0fe7e39ac/config-with-custom-regionmapping.json/) with custom `config.json > parameters > regionMappingDefinitonsUrl`.
2. Open the WPS catalog item from the catalog explorer
3. You should see a parameter with `Select region` button - click it to start the region selection flow.

UI screenshot:
image

### Checklist


lib/Map/Region/RegionProvider.ts
```diff
@@ -289,6 +342,51 @@ export default class RegionProvider {
...
+    let url = baseuri.setQuery("valueReference", nameProp).toString();
+
+    if (this.corsProxy.shouldUseProxy(url)) {
+      url = this.corsProxy.getURL(url);
+    }
+
+    this._loadRegionNamesPromise = loadText(url).then((xml) => {
+      const obj = xml2json(xml);
+
+      if (!defined(obj.member)) {
```
pjonsson: The `xml2json` function can return a string, null, and undefined (#7401), doesn't this line need to check that?

lib/Map/Region/RegionProvider.ts
```diff
@@ -500,3 +680,69 @@ function findVariableForAliases(varNames: string[], aliases: string[]) {
...
+
+  // get the list of IDs that we will attempt to match against for this column
+  const url = regionProvider.corsProxy.getURLProxyIfNecessary(
+    baseuri.setQuery("valueReference", propName).toString()
+  );
+
+  return loadText(url).then(function (xml) {
+    const obj = xml2json(xml);
+
+    if (obj.member === undefined) {
```
pjonsson: Same comment here about checking return value of xml2json.

lib/Map/Region/RegionProvider.ts
```diff
@@ -500,3 +680,69 @@ function findVariableForAliases(varNames: string[], aliases: string[]) {
...
+      console.log(xml);
```
pjonsson: Debug leftovers?

lib/ReactViews/Analytics/RegionParameterEditor.tsx
```diff
@@ -0,0 +1,80 @@
...
+        </button>
+      </div>
+    );
+  }
+);
+
+/**
+ * Prompt user to select/draw on map in order to define parameter.
+ * @param {Object} viewState ViewState.
+ * @param {FunctionParameter} parameter Parameter.
```
pjonsson: The type annotation specifies `RegionParameter`, do these annotations not interfere with the real type annotations?

lib/ReactViews/Analytics/RegionParameterEditor.tsx
```diff
@@ -0,0 +1,80 @@
...
+  terria.pickedFeatures = undefined;
+
+  const pickPointMode = new MapInteractionMode({
+    message: "Select a region on the map",
+    onCancel: function () {
+      terria.mapInteractionModeStack.pop();
+      viewState.openAddData();
+    },
+    buttonText: "Done",
+    customUi: function Done() {
```
pjonsson: The `onCancel` has a nameless function, does that work here as well?

lib/Map/Region/RegionProvider.ts
```diff
@@ -289,6 +342,51 @@ export default class RegionProvider {
...
   }
 
+  loadRegionNames() {
+    if (defined(this._loadRegionNamesPromise)) {
+      return this._loadRegionNamesPromise;
+    }
+
+    const nameProp = this.nameProp || this.regionProp;
+
+    const baseuri = URI(this.analyticsWmsServer).addQuery({
```
zoran995, line 352: Can we use `new URL(this.analyticsWmsServer)` instead of URI? Also, the type suggests that analyticsWmsServer is a string or undefined, this will fail with an undefined URL

lib/Map/Region/RegionProvider.ts
```diff
@@ -473,6 +571,88 @@ export default class RegionProvider {
...
+    ) {
+      regionID = parseInt(regionID, 10);
+    } else if (typeof regionID === "string") {
+      regionID = regionID.toLowerCase();
+      const replacedValue =
+        this._appliedReplacements.serverReplacements[regionID];
+      if (defined(replacedValue)) {
+        regionID = replacedValue;
+      }
+    }
```
zoran995, line 612: This is the same as in `findRegionByID`, can we extract it?

lib/Map/Region/RegionProvider.ts
```diff
@@ -473,6 +571,88 @@ export default class RegionProvider {
...
+  async getRegionFeature(
+    regionId: string | number,
+    terria: Terria
+  ): Promise<Feature | undefined> {
+    let url = this.analyticsWmsServer || this.server;
+    if (terria.corsProxy.shouldUseProxy(url)) {
+      url = terria.corsProxy.getURL(url);
+    }
+
+    url = new URI(url)
```
zoran995, line 636: Same as above, can we use `new URL` instead as a modern alternative

lib/Map/Region/RegionProvider.ts
```diff
@@ -500,3 +680,68 @@ function findVariableForAliases(varNames: string[], aliases: string[]) {
...
+  regionProvider: RegionProvider,
+  propName: string
+): Promise<{ values: (string | number)[] }> {
+  if (regionProvider.serverType !== "WMS") {
+    throw new DeveloperError(
+      "Cannot fetch region ids for region providers that are not WMS"
+    );
+  }
+
+  const baseuri = URI(regionProvider.server).addQuery({
```
zoran995, line 703: same as above for URI usage

lib/Map/Region/RegionProvider.ts
```diff
@@ -289,6 +342,51 @@ export default class RegionProvider {
...
+
+    this._loadRegionNamesPromise = loadText(url).then((xml) => {
+      const obj = xml2json(xml);
+
+      if (!defined(obj?.member)) {
+        const exception = defined(obj.Exception)
+          ? "<br/><br/>" + obj.Exception.ExceptionText
+          : "";
+        throw new TerriaError({
+          title: "CSV region mapping",
```
zoran995, line 375: Can we add translation for this?

lib/Map/Region/RegionProvider.ts
```diff
@@ -500,3 +680,68 @@ function findVariableForAliases(varNames: string[], aliases: string[]) {
...
+/**
+ * Fetch a list of region IDs in feature ID (FID) order by querying a WFS server
+.
+ * This is a slower fall-back method if we don't have a pre-computed JSON list a
+vailable.
+ * Returns a promise which resolves to an object whose 'values' property can be
+used as an argument in processRegionIds.
+ * @private
+ */
+function loadRegionsFromWfs(
```
zoran995, line 693: Should this be marked as an async function?

lib/Map/Region/RegionProvider.ts
```diff
@@ -500,3 +680,68 @@ function findVariableForAliases(varNames: string[], aliases: string[]) {
 
   return undefined;
 }
+
+/**
+ * Fetch a list of region IDs in feature ID (FID) order by querying a WFS server
+.
+ * This is a slower fall-back method if we don't have a pre-computed JSON list a
+vailable.
```
zoran995, line 688:
```suggestion
 * This is a slower fall-back method if we don't have a pre-computed JSON list available.
```

lib/Models/FunctionParameters/RegionTypeParameter.ts
```diff
@@ -1,6 +1,94 @@
...
+  /**
+   * Loads usable region providers
+   */
+  load(): Promise<void> {
+    if (this.loadPromise) {
+      return this.loadPromise;
+    }
+
+    const validRegionTypes = this.validRegionTypes
+      ? new Set(this.validRegionTypes)
```
zoran995, line 56: Why not convert to a set on initialisation?

lib/Models/FunctionParameters/RegionTypeParameter.ts
```diff
@@ -1,6 +1,94 @@
...
+
+  @computed
+  get regionProvider(): RegionProvider | undefined {
+    return this.regionProviders?.find((r) => r.regionType === this.value);
+  }
+
+  /**
+   * Loads usable region providers
+   */
+  load(): Promise<void> {
```
zoran995, line 50: Should this be marked as async?

lib/ReactViews/Analytics/RegionParameterEditor.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <input
+          className={Styles.field}
+          type="text"
+          readOnly={parameter.value === undefined}
+          onChange={() => {
+            // clear value on any attempt to edit the input
+            // ideally we should place an `x` icon to clear the field
+            parameter.setValue(CommonStrata.user, undefined);
+          }}
+          value={getDisplayValue(parameter.value as any, parameter)}
```
zoran995, line 36: `getDisplayValue` already supports undefined can you just extend its type instead of using `as any`

lib/ReactViews/Analytics/RegionTypeParameterEditor.tsx
```diff
@@ -0,0 +1,52 @@
+import { observer } from "mobx-react";
+import React, { useEffect } from "react";
+import ReactSelect from "react-select";
+import CatalogFunctionMixin from "../../ModelMixins/CatalogFunctionMixin";
+import CommonStrata from "../../Models/Definition/CommonStrata";
+import RegionTypeParameter from "../../Models/FunctionParameters/RegionTypeParameter";
+import Loader from "../Loader";
+
+interface PropsType {
+  previewed: CatalogFunctionMixin.Instance;
```
zoran995, line 10: this is not used anywhere

lib/Models/FunctionParameters/RegionParameter.ts
```diff
@@ -1,23 +1,42 @@
...
 }
 
-export default class RegionParameter extends FunctionParameter {
+type RegionValue = string | number;
+
+export default class RegionParameter extends FunctionParameter<RegionValue> {
   static readonly type = "region";
   readonly type = "region";
-  readonly regionProvider: RegionTypeParameter;
+  readonly _regionProviderOrParam: RegionProvider | RegionTypeParameter;
```
zoran995, line 17: Based on the usage, this seems to always be `RegionTypeParameter`

lib/ReactViews/Analytics/RegionPicker.tsx
```diff
@@ -0,0 +1,324 @@
...
+    useEffect(() => {
+      if (loader?.state === "fulfilled") {
+        const mapItem = loader.value.regionMapItem;
+        terria.overlays.add(mapItem);
+
+        return () => {
+          terria.overlays.remove(mapItem);
+        };
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
zoran995, line 102: I wonder why we need this?

lib/ReactViews/Analytics/RegionPicker.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+const DropdownRegionSelector: React.FC<{
+  regionId: RegionId | undefined;
+  regionProvider: RegionProvider;
+  onChange: (region: RegionId | undefined) => void;
+}> = observer(({ regionId, regionProvider, onChange }) => {
+  const regions = regionProvider.regions;
+  const regionNames = regionProvider.regionNames;
+  const [options, setOptions] = useState<{ value: any; label: any }[]>();
+  const [selectedOption, setSelectedOption] = useState<{
```
zoran995, line 156: selectedOption is deferred from the options and `parameter.value` we don't need to store it in state

lib/ReactViews/Analytics/RegionPicker.tsx
```diff
@@ -0,0 +1,324 @@
...
+);
+
+const DropdownRegionSelector: React.FC<{
+  regionId: RegionId | undefined;
+  regionProvider: RegionProvider;
+  onChange: (region: RegionId | undefined) => void;
+}> = observer(({ regionId, regionProvider, onChange }) => {
+  const regions = regionProvider.regions;
+  const regionNames = regionProvider.regionNames;
+  const [options, setOptions] = useState<{ value: any; label: any }[]>();
```
zoran995, line 155: When options are purely deferred, we don't need to store them in state and use an effect to sync them. It's cheaper from the render side to defer value during the rendering cycle. If the list is large, you can use useMemo, but I'm not sure if it is needed.

zoran995 (changes requested):

zoran995:
It wasn't easy to decode the difference between the region parameter, the region type parameter and the region provider and how they are connected (still not sure I got it everything correct)

PR #7674, opened 2025-08-04
Add coordinate result to mapbox search provider

micholm:
### What this PR does

Creates a new coordinate location feature when using the reverse geocoder functionality of MapboxSearchProvider, and adds it to the top of the results list.

Without with feature, a coordinate search will always return the nearest address (if close enough geographically). This is not desirable in a lot of cases where users will be wanting to find an exact location using coordinates.

This is configurable using the new trait parameter `showCoordinatesInReverseGeocodeResult`.

examples:
1. the coordinates `-41.107379944951006, 127.56722178479194` are too far from any known address to show results. However, with this feature the user can still go to the exact position:
image

2. a user has a coordinate for an asset or POI, and wants to see to that position exactly, without this feature, the nearest address is not helpful as it is unlikely to be the same position.

### Test me

* use the MapboxSearchProvider, be sure to supply your API key.
* with `showCoordinatesInReverseGeocodeResult` set to default (true), search for `-9.30738831489543, 134.48433477770556`. You should get a coordinate position and no addresses.
* with `showCoordinatesInReverseGeocodeResult` set to false, the same search should return no results
* with `showCoordinatesInReverseGeocodeResult` set to default (true), search for `-37.62692208450191, 145.54355658542465`. You should get two results, a coordinate and an address which should be spatially different.

### Checklist

- [ X] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
- [ X] I've updated relevant documentation in `doc/`.
- [ X] I've updated CHANGES.md with what I changed.
- [ X] I've provided instructions in the PR description on how to test this PR.

### DIFF
--- a/lib/Models/SearchProviders/MapboxSearchProvider.ts
+++ b/lib/Models/SearchProviders/MapboxSearchProvider.ts
@@ -198,14 +198,13 @@ export default class MapboxSearchProvider extends LocationSearchProviderMixin(
           // A new search has superseded this one, so ignore the result.
           return;
         }
-        let features = result.features;
 
         if (
-          (features.length === 0 &&
+          (result.features.length === 0 &&
             searchDirection === MapboxGeocodeDirection.Forward) ||
           //in the case where coordinate result is true, list is
           //not empty.
-          (features.length === 0 &&
+          (result.features.length === 0 &&
             searchDirection === MapboxGeocodeDirection.Reverse &&
             this.showCoordinatesInReverseGeocodeResult === false)
         ) {
@@ -215,7 +214,7 @@ export default class MapboxSearchProvider extends LocationSearchProviderMixin(
           return;
         }
 
-        const locations: SearchResult[] = features
+        const locations: SearchResult[] = result.features
           .filter(
             (feat) =>
               feat.properties && feat.geometry && feat.properties.full_address
--- a/lib/Models/SearchProviders/MapboxSearchProvider.ts
+++ b/lib/Models/SearchProviders/MapboxSearchProvider.ts
@@ -19,6 +19,7 @@ import SearchResult from "./SearchResult";
 import { Feature, Point } from "@turf/helpers";
 import isDefined from "../../Core/isDefined";
 import CommonStrata from "../Definition/CommonStrata";
+import prettifyCoordinates from "../../Map/Vector/prettifyCoordinates";
 
 enum MapboxGeocodeDirection {
   Forward = "forward",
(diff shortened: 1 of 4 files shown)


lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -95,10 +95,33 @@ export default class MapboxSearchProvider extends LocationSearchProviderMixin(
         isCoordinate.test(lonLat[0]) &&
         isCoordinate.test(lonLat[1])
       ) {
+        const formattedCoordinateString = `${parseFloat(lonLat[0]).toFixed(
```
ljowen: nitpick: since we're using them a few times could we destructure: 
```typescript
const [lon, lat] = lonLat.map(parseFloat);
```

ljowen (approved):

lib/Models/SearchProviders/MapboxSearchProvider.ts
micholm: thanks! I will need to do it after the reverse though, otherwise I think it might be more trouble than it's worth?

lib/Models/SearchProviders/MapboxSearchProvider.ts
```diff
@@ -95,10 +95,33 @@ export default class MapboxSearchProvider extends LocationSearchProviderMixin(
...
+          5
+        )}, ${parseFloat(lonLat[1]).toFixed(5)}`;
+
```
niktayv: A nitpick: this looks like a magic number.
Can the precision (i.e. 5) be made configurable in the traits?
Say, `reverseGeocodeResultPrecision`, like so:
```ts
  @primitiveTrait({
    type: "number",
    name: "reverseGeocodeResultPrecision",
    description: `The number of digits after the dot in the reverse search results.`
  })
  reverseGeocodeResultPrecision?: number = 5;
```

micholm: Thanks @niktayv, however this is for display purposes only, i.e. it doesn't look nice to show such long numeric strings as the name, and 5 points of precision seems to be the standard to display on other platforms (google maps for example). The precision is unaffected by this label.

Happy to discuss further, the label can just be whatever precision the user has entered, i just made a call to limit it so it fits the UI better.

na9da: There's also [prettifyCoordinates.ts](https://github.com/TerriaJS/terriajs/blob/bf564586a4197db1f573eb719725813ef3bfdb1e/lib/Map/Vector/prettifyCoordinates.ts#L31)

micholm: Thanks @na9da I used it :) looks better now and matches the Terria map style.

na9da:
To get CI to build, could you please pin the [typescript version](https://github.com/TerriaJS/terriajs/blob/9f13787801a0c3d2393cb8c9532ead442e288c0d/package.json#L168) to `"5.7.3"`. There's some breaking changes in `5.9.x` which we need to fix.

PR #7676, opened 2025-08-05
feat: change config load to custom function

zoran995:
### What this PR does

### Test me

### Checklist

### DIFF
--- a/lib/ReactViews/Map/Panels/BaseOuterPanel.js
+++ b/lib/ReactViews/Map/Panels/BaseOuterPanel.js
@@ -5,7 +5,7 @@ export default {
     theme: PropTypes.object,
     children: PropTypes.any,
     btnTitle: PropTypes.string,
-    btnText: PropTypes.string,
+    btnText: PropTypes.any,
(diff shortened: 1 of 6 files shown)


lib/ReactViews/Map/Panels/BaseOuterPanel.js
```diff
@@ -5,7 +5,7 @@ export default {
     theme: PropTypes.object,
     children: PropTypes.any,
     btnTitle: PropTypes.string,
-    btnText: PropTypes.string,
+    btnText: PropTypes.any,
```
ljowen, line 8: Is this needed? Could we do a cast instead?

zoran995: I added an option to support the react node along with text and proptypes will throw a warning in the console that a string was expected.
This is the only way with proptypes.

nf-s (approved):
LGTM! It would be good to merge this in soon so I can merge this into `saas-next-ws` branch

PR #7677, opened 2025-08-08
Misc fixes for project TerriaJS version

nf-s:
Note this will not be merged into `main` at this point in time, we may port useful features if they are relevant to OS usage

Major changes

- Tweak `TerriaError.toError`
  - Previously, `toError` was returning a nested error message (that is a rendered version of all errors in the tree)
  - Now, `toError` returns a new error that uses the `title` and `message` of the `highestImportanceError`. And then returns a new `stack` of all nested error `message` and `stack` traces. 
  - This is to keep the error `title` and `message` as concise as possible - otherwise `message` turns into a sprawling mess of all nested error objects
- Add `hideCloseButton` to new `SelectableDimensionWorkflowOptions` interface
- Add the following `ConfigParameters`
  - `disableMobileInterface`
  - `disableSharePanel`
  - `disableShareEmbed`
- Changes to the catalog viewer
  - Add `hideActionButton`, `hideSearch`, `hideActionButton` and `hideBreadcrumbs`
  - Add `onToggleItemOnMap` hook to override item toggle button ("+" or "-")
- `CesiumIonConnector` changes
  - add "+" button to remove "-" if it already exists in the workbench 
  - use `terria.baseUrl` when pointing to `cesium-ion-oauth2.html` - I've added a security check to ensure that the reidrectUri is always the same origin as `window.location`
- Fix MapColumn width when hiding workbench


lib/ReactViewModels/ViewState.ts
```diff
@@ -418,13 +418,15 @@ export default class ViewState {
...
           terria.userProperties.get("hideWorkbench") === "1" ||
           terria.userProperties.get("hideExplorerPanel") === "1";
       }
     );
 
     this._isMapFullScreenSubscription = reaction(
       () =>
+        terria.elements.get("show-workbench")?.visible === false ||
         terria.userProperties.get("hideWorkbench") === "1" ||
         terria.userProperties.get("hideExplorerPanel") === "1",
```
nf-s, line 431: The interaction between `show-workbench` (from elements), and `hideWorkbench` (from hash user properties) is a bit confusing. Ideally this is unified at some point, but I think for the moment we just need to make the behaviours consistent

lib/ReactViews/Workflow/WorkflowPanel.tsx
```diff
@@ -9,6 +9,7 @@ import { scrollBars } from "../../Styled/mixins";
 import Text from "../../Styled/Text";
 import { PortalChild } from "../StandardUserInterface/Portal";
 import { PanelButton } from "./Panel";
+import { SelectableDimensionWorkflowOptions } from "../../Models/Workflows/SelectableDimensionWorkflow";
```
zoran995: selectableDimensionWorkflow is just one specific workflow, I don't think we should depend on its type definition here

lib/ReactViews/Preview/GroupPreview.jsx
```diff
@@ -43,36 +43,38 @@ class GroupPreview extends Component {
...
-              </button>
-            )}
-            <SharePanel
-              catalogShare
-              modalWidth={this.props.widthFromMeasureElementHOC}
-              terria={this.props.terria}
-              viewState={this.props.viewState}
-            />
-          </div>
+          {this.props.terria.configParameters.disableSharePanel && (
```
zoran995: Is it intended to hide `add/remove all` when the share panel is hidden?

lib/Core/TerriaError.ts
```diff
@@ -378,51 +378,50 @@ export default class TerriaError {
   /**
    * Returns a plain error object for this TerriaError instance.
    *
-   * The `message` string for the returned plain error will include the
-   * messages from all the nested `originalError`s for this instance.
+   * Will return the highest importance erorr (as name/message), and then returns a stack trace with the messages from all the nested `originalError`s for this instance.
```
zoran995: nitpick: type `erorr`

zoran995 (commented):
just few small comments

lib/ReactViews/Workflow/WorkflowPanel.tsx
nf-s: Good point! I'll move it to `SelectableDimensionWorkflow`

lib/ReactViews/Preview/GroupPreview.jsx
nf-s: Good pickup! I'll move that one level in

nf-s:
Thanks for your feedback @zoran995 

i also added one more commit

- https://github.com/TerriaJS/terriajs/commit/96104e738f55ecf99341fd0fd9332f641c90a676
- A `disableUserAddedData` config parameter to disable user added data
- `side-panel-add-data` and `side-panel-upload-data` element configurations so I can hide the two buttons in the side panel
  - Is this the correct use of `elements` outside of the `withControlledVisibility` HOC?

zoran995 (approved):
LGTM, I am not sure about the elements usage, we could create a new small component wrapped with a controlled visibility hook to have it follow the same pattern as in the rest of the app

nf-s:
> LGTM, I am not sure about the elements usage, we could create a new small component wrapped with a controlled visibility hook to have it follow the same pattern as in the rest of the app

Yeah I agree, it would be nice to have a small component at some point

PR #7682, opened 2025-08-11
Refactor search providers

zoran995:
### What this PR does

During testing the search providers, I discovered that search debounce doesn't work correctly and that we can achieve a better UX when autocomplete is disabled for some search providers.

Before: 
- Try entering more than three characters in the search box and observe that the search is triggered on each change
- Try searching for mel\ and observe a white screen due to an unescaped regex group

After:
- When autocomplete is disabled for some search providers, an info message for it is shown saying "Press Enter to start searching"; other search providers will search with autocomplete
- Search is debounced correctly, and the logic is moved to the search provider mixin level
- After the enter is pressed, it will start manual search, bypassing the debounce mechanism.
- Search string is escaped for highlight, resolving the hard error thrown due toan  unescaped group
- Fixed a few potential issues with map credits for search providers

image

### Test me

### Checklist


lib/Models/SearchProviders/CatalogSearchProvider.ts
```diff
@@ -170,16 +169,18 @@ export default class CatalogSearchProvider extends CatalogSearchProviderMixin(
...
       } else {
-        await loadAndSearchCatalogRecursively(
+        const t = await loadAndSearchCatalogRecursively(
           this.terria.modelValues,
           searchText.toLowerCase(),
           searchResults,
           resultMap
         );
+
+        console.log(t);
```
ljowen:
```suggestion
```

lib/Traits/SearchProviders/MapboxSearchProviderTraits.ts
```diff
@@ -86,4 +86,8 @@ export default class MapboxSearchProviderTraits extends mixTraits(
     be desired. Defaults to true.`
   })
   showCoordinatesInReverseGeocodeResult?: boolean = true;
+
+  get autocompleteEnabled() {
+    return true;
+  }
```
na9da: Why override this here when it is already part of `LocationSearchProviderTraits` and is defaulting to `true`? Same for the other providers.

lib/ModelMixins/SearchProviders/LocationSearchProviderMixin.ts
```diff
@@ -32,8 +32,18 @@ function LocationSearchProviderMixin<
     @action
     showWarning() {}
 
-    supportsAutocomplete(): boolean {
-      return true;
+    async search(searchText: string, manuallyTriggered?: boolean) {
```
na9da, line 35: Doubt: Could you explain when `manuallyTriggered` is true and why we need to differentiate both?

lib/ReactViews/DataCatalog/DataCatalog.jsx
```diff
@@ -26,24 +26,20 @@ class DataCatalog extends Component {
...
         {isSearching && catalogSearchProvider && (
           <>
             <label className={Styles.label}>{t("search.resultsLabel")}</label>
-            <SearchHeader
-              searchResults={searchState.catalogSearchResults}
-              isWaitingForSearchToStart={
-                searchState.isWaitingToStartCatalogSearch
-              }
-            />
+            <SearchHeader searchResult={catalogSearchProvider.result} />
```
na9da: This should be `catalogSearchProvider.searchResult` instead (currently this results in app crash when searching from catalog explorer modal)

lib/ReactViews/DataCatalog/DataCatalog.jsx
```diff
@@ -26,24 +26,20 @@ class DataCatalog extends Component {
     const unfilteredItems =
       isSearching &&
       catalogSearchProvider &&
-      searchState.catalogSearchResults?.results
-        ? searchState.catalogSearchResults.results.map(
+      searchState.catalogSearchProvider?.result.results
+        ? searchState.catalogSearchProvider.result.results.map(
```
na9da: This should be `catalogSearchProvider.searchResult` instead (currently this results in app crash when searching from catalog explorer modal)

lib/ReactViews/ReactViewHelpers/highlightKeyword.jsx
```diff
@@ -1,6 +1,11 @@
 export default function highlightKeyword(searchResult, keywordToHighlight) {
   if (!keywordToHighlight) return searchResult;
-  const parts = searchResult.split(new RegExp(`(${keywordToHighlight})`, "gi"));
+  const escapedKeyword = keywordToHighlight.replace(
+    /[.*+?^${}()|[\]\\]/g,
+    "\\$&"
+  );
```
na9da, line 6: BTW, there is a new [RegExp.escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape), although it is pretty new.

na9da (changes requested):

na9da:
@zoran995 - Looks good, just a few comments.

na9da (changes requested):
Search from mobile view must also be fixed.

lib/ModelMixins/SearchProviders/LocationSearchProviderMixin.ts
zoran995: `manuallyTriggered` is when the user presses enter to start the search; in that case, we need to cancel the debounce and run the search without waiting

lib/ReactViews/DataCatalog/DataCatalog.jsx
zoran995: This somehow got lost in the changes. I definitely tested this. Fixed

lib/Traits/SearchProviders/MapboxSearchProviderTraits.ts
zoran995: I just wanted to be explicit for all search providers, I can remove it if necessary

lib/ReactViews/ReactViewHelpers/highlightKeyword.jsx
zoran995: I would avoid using it for some time, as it is still pretty new although it would work for most of the users, and this is enough to resolve the issue we are currently seeing

lib/Traits/SearchProviders/MapboxSearchProviderTraits.ts
zoran995: Actually I was running into a bug that the value wasn't propagated correctly

zoran995:
@na9da this is ready for another pass, #7715 builds on top of this

na9da (approved):
Thanks for the fixes @zoran995 - I've tested the changes and they LGTM

PR #7689, opened 2025-08-25
Svg sprite plugin

zoran995:
### What this PR does

`svg-sprite-loader` is no longer maintained and has numerous vulnerable dependencies. After an extensive internal discussion, we decided to replace it with a custom Webpack plugin implementation based on `svg-sprite`. The PR implements a new webpack plugin and loader.
- Loader is responsible for processing the imported icons and replacing them with a module definition. Additionally, the loader registers icons in the shared storage, allowing the plugin to process them. Icons will be registered with a specified namespace, which is necessary for icons coming from plugins, as we don't want them to overwrite terriajs icons. 
- Plugin processes the loaded icons and outputs the SVG-sprite file for each namespace and SVG manifest that contains a list of all output sprites. When loading the sprites there are two options
  - plugin makes a list of sprites available for html-webpack-plugin so they can be injected directly in html
  - svg manifest can be loaded at runtime and then used to load sprites

### Test me

### Checklist


buildprocess/ci-deploy.sh
```diff
@@ -22,7 +22,7 @@ npm install -g yarn@^1.22.22
 
 # Clone and build TerriaMap, using this version of TerriaJS
 TERRIAJS_COMMIT_HASH=$(git rev-parse HEAD)
-git clone -b main https://github.com/TerriaJS/TerriaMap.git
+git clone -b svg-sprite-plugin https://github.com/TerriaJS/TerriaMap.git
```
zoran995: TODO: revert after merge

buildprocess/svgs/SvgSpriteWebpackPlugin.js
```diff
@@ -0,0 +1,271 @@
...
+          `⚠️  Failed to delete old sprite file ${basename(path)}:`,
+          error.message
+        );
+      }
+    });
+  }
+
+  apply(compiler) {
+    console.log("🔨 Initializing SVG Sprite Webpack Plugin...");
+    compiler.hooks.thisCompilation.tap(
```
na9da, line 25: Nitpick: to improve clarity, would it be possible to split this method?

buildprocess/svgs/SvgSpriteWebpackPlugin.js
```diff
@@ -0,0 +1,271 @@
...
+              mainChunk.files.add("svg-sprite.js");
+            }
+
+            console.log("✅ Generate svg sprite injection script");
+          }
+        );
+
+        // Hook into HtmlWebpackPlugin at the compiler level
+        if (compiler.hooks.compilation) {
+          compiler.hooks.compilation.tap(
```
na9da: Do we still need this block now that we are generating the inject script above?

na9da (changes requested):
Looks good @zoran995. Have left a couple of comments.
The CI should be fixed by https://github.com/TerriaJS/terriajs/pull/7699

buildprocess/svgs/SvgSpriteWebpackPlugin.js
zoran995: no, removed

buildprocess/svgs/SvgSpriteWebpackPlugin.js
```diff
@@ -0,0 +1,233 @@
...
+
+              compilation.emitAsset(
+                "svg-sprite.js",
+                new sources.RawSource(script)
+              );
+
+              const mainChunk = compilation.namedChunks.get("main");
+              if (mainChunk) {
+                mainChunk.files.add("svg-sprite.js");
+              }
```
na9da: The re-org seems to have broken this step - `namedChunks` is empty at this stage and so does not include `svg-sprite.js` in the final build. I think it is because this dependency add needs to happen in `compilation.hooks.processAssets` stage which is called after all chunks have been built. Here's [an example](https://github.com/orgs/webpack/discussions/18084).

na9da:
Hi @zoran995 - looks great, could you please merge from the latest `main` when you get a chance? 

Also, please update CHANGES.md.

na9da (approved):
Thanks @zoran995 !

PR #7696, opened 2025-09-03
Tame ESiInt in build scripts

niktayv:
### What this PR does

Similar to https://github.com/TerriaJS/TerriaMap/pull/764, I just added a minimal amount of ESLint rules and annotations to stop the build scripts from screaming with red squiggly lines.

### Test me

No logic should have been changed; only the behaviour in VS Code is affected.

### Checklist


buildprocess/generate-cesium-declarations.js
```diff
@@ -19,7 +19,7 @@ let matches;
 const publicModules = new Set();
 //eslint-disable-next-line no-cond-assign
 while ((matches = regex.exec(declrationFileSource))) {
-  const moduleName = matches[2].match(/([^\s|\(]+)/);
+  const moduleName = matches[2].match(/([^\s|(]+)/);
```
na9da: I think this regex can be safely changed to `/([^\s(]+)/` - which basically means, in a declaration statement, match until the first space or `(` char. This could break for more complex types but it should be easy to catch and fix when it happens.

na9da (changes requested):
Hi @niktayv - thanks for the fixes. I have suggested one change to the regex.

buildprocess/generate-cesium-declarations.js
niktayv: @na9da Nanda, do we need to update CHANGES.md in this case?

na9da: hmm.. It is a minor detail which I don't think needs to be recorded in CHANGES.

na9da (dismissed):
Looks good. Thanks @niktayv.

na9da (approved):
Approved.

PR #7701, opened 2025-09-25
Fix Chart zoom not working #7654

### LINKED ISSUE #7654: Chart zoom not working
Previously scrolling on the chart panel used to scale the chart up or down. This appears to be broken.

Example link: http://ci.terria.io/main/#share=s-jVponG4vZK7kLVK1utmps8Xg9lX

niktayv:
### What this PR does

Fixes #7654

Applied the fix suggested by @hugominas in https://github.com/TerriaJS/terriajs/issues/7654#issuecomment-3296823375.
Reviewed the code and added a cleanup for D3 zoom handlers.

### Test me

### Checklist


niktayv:
@zoran995 Hi, Zoran, I touched some of your code to the best of my understanding.
It would be awesome if you could find time to take a look.

lib/ReactViews/Custom/Chart/BottomDockChart.tsx
```diff
@@ -223,7 +239,7 @@ const Chart: React.FC<ChartProps> = observer(
           [0, 0],
           [Infinity, Infinity]
         ]}
-        onZoom={setZoomedXScale}
+        onZoom={handleZoom}
```
na9da: I think zoom is failing because `XScale` is a function and when passing a function to setState, React treats it as a special [updater function](https://react.dev/reference/react/useState#:~:text=If%20you%20pass%20a%20function%20as%20nextState%2C%20it%20will%20be%20treated%20as%20an%20updater%20function.) instead of as a raw value. So one easy fix might be to change the call to `setZoomedXScale` as:

```js
        // xScale is a function and React setState treats functions as updater functions,
        // so wrap it in another function to actually set the scale.
        onZoom={(xScale) => setZoomedXScale(() => xScale)}
```

na9da (changes requested):
Hi @niktayv, thanks for looking into this issue.

I have suggested a fix that might be easier.

Another thing I noticed which probably existed even before the refactor is that the `<Plot>` component is getting re-rendered on each mouse move, which could be expensive. This can be avoided if we wrap `initialScales` and `zoomedScales` computations in a `useMemo` like below:

```js
    const initialXScale: XScale = useMemo(() => {
      const params = {
        domain: calculateDomainX(processedChartItems),
        range: [0, plotWidth]
      };
      if (xAxis.scale === "linear") return scaleLinear(params);
      else return scaleTime(params);
    }, [xAxis, processedChartItems, plotWidth]);

// and similar for zoomedScales
```

It would be great if you can have a look at it while you are at this. You can test the re-rendering behaviour by putting a [break-point here](https://github.com/TerriaJS/terriajs/blob/bc03021a86d1fac3df24ff914454e576033d9cff/lib/ReactViews/Custom/Chart/utils.tsx#L25) and noticing that the component is getting rendered on each mouse move.

Cheers,
Nanda

lib/ReactViews/Custom/Chart/BottomDockChart.tsx
niktayv: Thank you, @na9da Nanda. That was the case!
I also memoised `initialScales` and `zoomedScales`, as you suggested, and verified that it stopped unnecessary rerenders on `<Plot>`.
Could you please remind me what I should do to prevent `Deploy TerriaMap / deploy (push) Failing ...`?

na9da: Thanks @niktayv 

You can try merging from the latest `main` to fix CI. 
Also, please update the CHANGES.md.

niktayv: Thanks, @na9da Nanda. Merging `main` helped.

na9da (approved):

PR #7709, opened 2025-10-16
Fix preview outside explorer

na9da:
### What this PR does

The preview map is sometimes used outside the explorer context. This fix ensures it is always shown. It was previously hidden if the explorer panel is animating but that check had [to be removed](https://github.com/TerriaJS/terriajs/pull/7691) to fix a race condition. This change makes it so that the preview map is always shown.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 #### next release (8.11.1)
 
+- Fix to show preview map when used outside the explorer panel.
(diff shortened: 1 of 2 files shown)


niktayv (approved):
I see, Nanda, where you are coming from. LGTM.

na9da:
Thanks @niktayv!

Issue #7700, opened 2025-09-21
Zoom into location for satellite imagery is broken


AnaBelgun:
To reproduce issue: 
- go to map.terria.io
- select Australia - National - Satellite imagery 
- add it to the map 
- select "filter by location" 
- once selection is active, click on 'Zoom to' which takes you to a random location, not the filtered/selected one

niktayv:
This issue is related to https://github.com/TerriaJS/terriajs/issues/5057.
See Zoran's @zoran995 comment https://github.com/TerriaJS/terriajs/issues/5057#issuecomment-771862340.
Another way to reproduce the bug is to click at a point covered by satellite imagery, and when the Feature Information panel finishes loading, scroll down to the bottom and click the "Show Change detection..." button. It will use the wrong Lat / Lon values, displayed above the button.

PR #7712, opened 2025-10-24
fix: wms get feature info with projection

### LINKED ISSUE #5057: Incorrect Feature Info panel lat/lon on WMS Point data
Consider https://nationalmap.gov.au/ and add external WMS of https://gs.geoscience.nsw.gov.au/geoserver/ows

Add layer "NSW Drillhole"

I have shared this here https://nationalmap.gov.au/#share=s-ko2gHxva9VKD7RY0UtDSDHOFNxf

Clicking on the points, the lat/lon in the feature panel are not correct.

image

I am not suggesting this is a specific issue with the NSW Geoscience WMS. I needed a WMS vector point source and couldn't find one in the National Map catalogue. We are seeing the same problems on our in-house GeoServer.

http://ci.terria.io/master/ and http://ci.terria.io/next/ similarly afflicted. https://map.terria.io/ (suitably ancient) also not good.

I think .csv and Esri MapServer are ok.

Advice is invited.

zoran995:
### What this PR does

Fixes #5057, #7700

### Test me

### Checklist

### DIFF
--- a/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
+++ b/lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
@@ -5,13 +5,19 @@
 // 3. Observable spaghetti
 //  Solution: think in terms of pipelines with computed observables, document patterns.
 // 4. All code for all catalog item types needs to be loaded before we can do anything.
+import { FeatureCollection } from "geojson";
 import i18next from "i18next";
 import { computed, makeObservable, override, runInAction } from "mobx";
+import Cartesian3 from "terriajs-cesium/Source/Core/Cartesian3";
+import Cartographic from "terriajs-cesium/Source/Core/Cartographic";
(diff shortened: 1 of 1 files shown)


lib/Models/Catalog/Ows/WebMapServiceCatalogItem.ts
```diff
@@ -818,4 +841,40 @@ export function formatDimensionsForOws(
   );
 }
 
+function geoJsonToFeatureInfoWithProject(
```
na9da, line 846: Might be good to add a comment to describe what the function does and link back to [the source](https://github.com/CesiumGS/cesium/blob/5754031f65646bee5f9d0e9a56dec7d3677a8b08/packages/engine/Source/Scene/GetFeatureInfoFormat.js#L74)

na9da (approved):
Looks good @zoran995. Just a minor suggestion

PR #7710, opened 2025-10-23
Add toast message UI.

na9da:
### What this PR does

Part of UX for https://github.com/TerriaJS/terriajs/issues/6845 
Split from  https://github.com/TerriaJS/terriajs/pull/7626

- Adds a new toast message UI.
- Adds new traits to show `initialMessage` as a toast instead of a popup.
 
 ```json
{
  "catalog": [
    {
      "id": "bike-racks-toast",
      "type": "geojson",
      "name": "GeoJSON Test",
      "url": "test/bike_racks.geojson",
      "initialMessage": {
        "title": "Bike racks",
        "content": "This message will self-destruct in 5 seconds",
        "showAsToast": true,
        "toastVisibleDuration": 5
      }
    }
  ]
}
  ```

### Test me

- Open [this CI link](http://ci.terria.io/toast-notification/#clean&https://gist.githubusercontent.com/na9da/9a98f0715d25b86c05d3d6183b4a1951/raw/f247cfbb3bff0ac04050fec31c5b308bcc08e75e/toast.json)
- Add the only item from the catalog
- Observe the notification toast that auto-dismisses
   image

- Remove and add the item again to the workbench
- Observe that the toast is not shown again (only the first time)

### Checklist

- [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
- [ ] I've updated relevant documentation in `doc/`.
- [x] I've updated CHANGES.md with what I changed.
- [x] I've provided instructions in the PR description on how to test this PR.

Issue #6845, opened 2023-08-17
Support for 3031 projections - investigation


AnaBelgun:
Context: 
- some projects focused on the South Pole require a web visualisation using TerriaJS, and hence need data to be projected in 3031 CRS
- currently that's not possible

Options: 
- use 4326 for a 3D view 
- visualise data only in the 2D view based on Leaflet and have a 'South Pole/3031" mode that users can enable 
- Some examples of Leaflet based South Pole web maps: http://www.soosmap.aq/ and https://nilas.org/#0/-72/8

What would option 2 entail?

na9da:
For approach 2 using Leaflet, there is [Proj4Leaflet](https://github.com/kartena/Proj4Leaflet). [Example usage here](https://observablehq.com/@jbaba/antarctica-leaflet).

To support 2 we have to allow configuring additional viewer-modes for a map and the ability to pass CRS parameter for leaflet mode. Currently the viewer-modes are hardcoded. 

The catalog maintainer can use the `shortReport` trait to remind the user to switch to "South Pole" mode.

AnaBelgun:
the task here is to provide a means to extend the map modes, like through a plugin.

SpacemanPaul:
For whole world completeness, you might also consider the north pole stereographic CRS EPSG:3995.

Web Mercator plus 3031 and 3995 would allow full coverage of the entire surface of the Earth.

Presumably a plugin approach would be able to support both, plus any other specialist CRSs that might crop up.

na9da:
Hi @SpacemanPaul - just sharing an early  version of custom projection for 2D leaflet mode. Here's a [CI link for testing](http://ci.terria.io/hacky-proj4leaflet/#share=s-ulC1HOMrFd7d948FCIR3CqCNQZF&https://gist.githubusercontent.com/na9da/b89dbf59c5c2dd62587919cdea9c82d6/raw/5e657d6b53c7a684a057f0c8dcb87013421ca642/antarctica.json&clean). From the map settings select basemap for Antarctica and it should automatically switch to 2D mode with the right projection. 

Some notes:
- This implementation allows configuring  the viewer mode to use for individual basemaps. Terria will instantiate the viewer when switching to the basemap. Here's what the [basemap config looks like](https://gist.github.com/na9da/b89dbf59c5c2dd62587919cdea9c82d6#file-antarctica-json-L14-L48).
- Map extent, max zoom etc is configurable using the viewer setting.
[...]

PR #7714, opened 2025-10-29
Set preferred viewer for base map

na9da:
Split from #7626 
Required for #6845 
Merge after #7710 

### What this PR does

- Adds trait `preferredViewer` to mappable mixin
- Automatically switch to preferred viewer when changing base maps
- Automatically switch base map to a compatible one when changing the viewer mode

Example base map config ([full example](https://gist.github.com/na9da/290ad4694ef4dd2dcafc89cac801af21)):
```json
"baseMaps": {
    "items": [
      {
        "item": {
          "id": "basemap-bing-aerial-with-labels",
          "preferredViewerMode": "2d"
        },
        "image": "build/TerriaJS/images/bing-aerial-labels.png"
      },
      {
        "item": {
          "id": "basemap-bing-aerial",
          "preferredViewerMode": "2d"
        },
        "image": "build/TerriaJS/images/bing-aerial.png"
      }
    ]
  }
```

### Test me

- Open [this CI link](http://ci.terria.io/basemap-options/#https://gist.githubusercontent.com/na9da/290ad4694ef4dd2dcafc89cac801af21/raw/1d1821efc2572dd0029fc87883c0d0cbf434b92c/preferred-viewer-mode.json) with both aerial Bing maps configured to [prefer 2d mode](https://gist.github.com/na9da/290ad4694ef4dd2dcafc89cac801af21#file-preferred-viewer-mode-json-L10)
- Switch from OSM to any bing map
- Observe that the viewer mode is switched to 2d and a toast notification is shown
- Switch viewer mode to 3d, observe that the base map switches to OSM and a toast notification is shown

### Checklist


lib/ViewModels/TerriaViewer.ts
na9da, line 1: Tracks the currently loading base map and sets it so that the last selected base map gets applied. Previously if the user selected multiple base maps then it would be applied in the order they were loaded instead of the order they were selected. This can be surprising to the user when some base maps are slow to load.

niktayv:
I tested the CI link and followed the instructions. Everything worked as described except for the basemap loaded after switching back to 3D mode. It was 'Bing Maps Roads', not OSM. If this is the expected behaviour, it makes sense to reflect it in the description above.
Also,  CHANGES.md have not yet been updated with this feature.

micholm (approved):
Awesome LGTM, works as described.

PR #7716, opened 2025-11-03
Keep camera steady when switching between viewer modes.

na9da:
Split from #7626 
Required for #6845 

### What this PR does

Track initial camera view for viewers and return the same view unless the viewer camera has changed.

Previously, the camera view would drift when switching between viewer modes. This is because each time we switch the viewer, we get the current view from leaflet/cesium instance and the value changes slightly each time depending on how the viewer sets the camera. The drift becomes even more noticeable when using custom projections and the extent used with the custom projection is not compatible with the other map modes. This can be a problem when restoring share links and the viewer is first initialized in 3d/cesium mode and then switches to leaflet/2d mode.

This change makes sure that the view remains stable if the camera hasn't changed after setting the initial view.

### Test me

- Open this main branch [CI link](http://ci.terria.io/main)
- Then from the settings panel, switch between 3D and 2D a few times (7-8 times)
- Observe the camera view drift from home camera (Australia) to the global view.
- Now repeat the same with [this branch](http://ci.terria.io/initial-camera-view)
- Observe that the camera view is stable no matter how many times we switch the viewer modes.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - Update docs for Client-side config: change `searchBar` parameter to `searchBarConfig`
 - Fix to show preview map when used outside the explorer panel.
 - Update `csv-geo-au` support to include the latest Australian Government regions.
+- Keep camera steady when switching between viewer modes.
 - [The next improvement]
 
 #### 8.11.0 - 2025-10-09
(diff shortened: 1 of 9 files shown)


micholm (approved):
LGTM

PR #7727, opened 2025-11-14
Fix sharing when using initfile.

### LINKED ISSUE #7269: Share using initfile

glughi:
### What this PR does

Fixes #7269 
Replace PR #7270 

### Test me
Using, ie, `https://terriamap.com/#emg202305`, you can load some data from `emg202305.json`.
Without this fix, the sharing link from `https://terriamap.com/#emg202305` will be something like 
https://terriamap.com/#share=g-8feface8fefd6fe09ac81f22f21ec5d8
so you can't reload data from `emg202305.json`.
Instead, with this fix, the link will be something like
https://terriamap.com/#emg202305&share=g-8feface8fefd6fe09ac81f22f21ec5d8

### Checklist


lib/ReactViews/Map/Panels/SharePanel/BuildShareLink.ts
```diff
@@ -40,6 +40,16 @@ function buildBaseShareUrl(
 ) {
   const uri = new URI(document.baseURI).fragment("").search("");
 
+  const usefulSubs = new URL(document.URL).hash
```
niktayv: Can we find a more descriptive name for this variable?

niktayv:
It would be helpful to describe how to test this feature.
Is there a minimal way to reproduce a failure?

niktayv:
By the way, I inspected the CI/build failure that occurs at `yarn gulp test-firefox`.
Locally, this test passes, so I asked the core team to help resolve this issue.

glughi:
Updated.

na9da (approved):
LGTM. Thanks @glughi

PR #7731, opened 2025-11-17
Fix disappearing of MyLocation point

### LINKED ISSUE #7725: My Location point disappears
The _My Location_ point can go under the terrain and disappear (partially or completely) in 3D mode if "Terrain hides underground features" is enabled.

https://github.com/user-attachments/assets/4170b7ad-142a-4ee3-a8db-3f9d2055d9bd

glughi:
### What this PR does

Fixes #7725

### Checklist


glughi:
Hello, any news about this PR?

niktayv:
Hi Giovanni. Thanks for the fix and for your interest in improving TerriaJS. 
I tested your solution locally, and it does resolve the issue.

Can I suggest you tweak the implementation to be a bit more robust and generic? 

As far as I understand your idea, we want the app-generated "My Location" item to remain visible, even when the "Terrain hides underground features" checkbox is enabled. Your solution works, but it creates a coupling to i18n and to a very concrete instance of the item, which may be a leaky abstraction.

Add a generic trait `isAlwaysVisible?: boolean = false` to `lib/Traits/TraitsClasses/GeoJsonTraits.ts` with a description for disabling depth testing.
```ts
@primitiveTrait({
  type: "boolean",
  name: "Is Always Visible",
  description:
    "If true, points will render without depth testing so they remain visible through terrain and tiles."
})
isAlwaysVisible?: boolean = false;
```

Set this trait on the app-created My Location marker in `lib/ReactViews/Map/MapNavigation/Items/MyLocation.ts`:
```ts
this._marker.setTrait(CommonStrata.user, "isAlwaysVisible", true);
```

In `lib/ModelMixins/GeojsonMixin.ts`, switch the depth-test override to:
```ts
disableDepthTestDistance: this.isAlwaysVisible
  ? new ConstantProperty(Number.POSITIVE_INFINITY)
  : undefined
```

Hopefully, this will be closer to TerriaJS’s architecture (avoiding name/i18n coupling and making the behaviour reusable).

na9da:
I agree with @niktayv's comment on moving the setting to somewhere that is more specific.

This could be done in [MyLocation.ts](https://github.com/TerriaJS/terriajs/blob/1738f73af352422b58c42c87f8739b2f0a4575b1/lib/ReactViews/Map/MapNavigation/Items/MyLocation.ts#L117) where the marker is created, by adding something like below:

```
// Set depth test distance to a large value to disable it so that the marker is visible even when dept testing is enabled
this._marker.defaultStyle.point.null.setTrait(CommonStrata.user, "disableDepthTestDistance", 9999999);
```

This is using the existing [disableDepthTestDistance trait](https://github.com/TerriaJS/terriajs/blob/1738f73af352422b58c42c87f8739b2f0a4575b1/lib/Traits/TraitsClasses/Table/PointStyleTraits.ts#L61-L67) (which is honestly a bit hard to find)

glughi:
Hi @na9da, I tried the suggestion (and some variations), but it doesn't seem to work to fix the issue.

na9da:
Ah sorry! That's because when using [simple-style for GeoJSON](https://github.com/bioretics/terriajs/blob/11980d518402c8945cf89e56a6fe153516909589/lib/ReactViews/Map/MapNavigation/Items/MyLocation.ts#L145-L148) these other styling traits (like disableDepthTestDistance)  are ignored.

In that case I would use what @niktayv has suggested - to add a new trait and set that from MyLocation.ts - but maybe name the new boolean trait `disableDepthTest`  instead of `isAlwaysVisible` to make it a bit more obvious.

niktayv (approved):
LGTM

na9da:
LGTM. Thanks @glughi @niktayv

PR #7744, opened 2026-01-07
Make StoryPanel draggable

### LINKED ISSUE #7242: Story creation and playback UI needs improvement
If a story contains expanded charts at the bottom of the map, the playback UI will be hidden behind the charts unless the charts are closed. See [example](https://nationalmap.gov.au/#share=s-h8B4vVlKklcetK6tq6rB2pbarYl).

When capturing a scene while a feature panel is open, the "Save" button will be blocked by the panel. ~Therefore there is no way to create a story with feature panel opened.~ Users need to move the panel away.

ljowen:
### What this PR does

Fixes #7242

There's a few places where the Story panel obscures other controls, such as the chart, controls in scene editor and the compare tool.

Make Story playback panel draggable.
- Change breakpoint so story panel and feature info can no longer drag under workbench at screen-md
- Fix non working `css` prop on `DragWrapper`

### Test me
http://ci.terria.io/draggable-story-panel/#share=s-eK4PjihAfNlsrdGumKL3sXBoyx7

- Create a Story and play it, the panel should now be draggable within the map area

### Checklist

### DIFF
--- a/lib/ReactViews/StandardUserInterface/standard-user-interface.scss
+++ b/lib/ReactViews/StandardUserInterface/standard-user-interface.scss
@@ -95,7 +95,7 @@
   pointer-events: none;
   left: 0;
 
-  @media (min-width: variables.$md) {
+  @media (min-width: variables.$sm) {
     left: variables.$work-bench-width + 22px;
     top: 62px;
     right: 96px;
@@ -138,7 +138,7 @@
   pointer-events: none;
   left: 0;
 
-  @media (min-width: variables.$md) {
+  @media (min-width: variables.$sm) {
     left: variables.$work-bench-width + 22px;
     top: 62px;
     right: 96px;
--- a/lib/ReactViews/StandardUserInterface/standard-user-interface.scss
+++ b/lib/ReactViews/StandardUserInterface/standard-user-interface.scss
@@ -125,10 +125,6 @@
   }
 }
 
-.featureInfoFullScreen {
-  left: 0;
-}
-
 .storyPanel {
   z-index: variables.$front-component-z-index - 20;
   position: absolute;
@@ -149,6 +145,11 @@
     pointer-events: auto;
     position: absolute;
   }
+}
+
+.featureInfoFullScreen,
+.storyPanelFullScreen {
+  left: 0;
 }
 
 .showWorkbenchButton {
--- a/lib/ReactViews/StandardUserInterface/standard-user-interface.scss
+++ b/lib/ReactViews/StandardUserInterface/standard-user-interface.scss
@@ -129,6 +129,28 @@
   left: 0;
 }
 
+.storyPanel {
+  z-index: variables.$front-component-z-index - 20;
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  top: 0;
+  pointer-events: none;
+  left: 0;
+
+  @media (min-width: variables.$md) {
+    left: variables.$work-bench-width + 22px;
+    top: 62px;
+    right: 96px;
+    bottom: 0;
+  }
+
+  & > div {
+    pointer-events: auto;
+    position: absolute;
+  }
+}
+
 .showWorkbenchButton {
   position: fixed;
   top: 18px;
--- a/lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
+++ b/lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
@@ -95,7 +95,6 @@ const Swipeable = ({
 class StoryPanel extends Component<Props, State> {
   keydownListener: EventListener | undefined;
   slideRef: RefObject<HTMLElement>;
-  dragWrapperCallbackRef: (node: HTMLDivElement | null) => void;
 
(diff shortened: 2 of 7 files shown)


lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
```diff
@@ -219,112 +220,87 @@ class StoryPanel extends Component<Props, State> {
...
-          onClick={() => this.onClickContainer()}
-          css={`
-            transition: padding, 0.2s;
-            bottom: ${this.props.viewState.terria.timelineStack.top !==
-            undefined
+      <DragWrapper
+        handleSelector=".drag-handle"
+        style={{
+          bottom:
+            this.props.viewState.terria.timelineStack.top !== undefined
```
niktayv, line 231: Potential runtime error if  `story` is undefined.

If `stories` array is empty or `currentStoryId` is out of bounds, `story` will be undefined (line 220), and accessing story.id on line 245 will throw a TypeError. 

Should we add a guard or an early return?
```
   render() {
     const stories = this.props.viewState.terria.stories || [];
     const story = stories[this.props.viewState.currentStoryId];
+
+    if (!story) {
+      return null;
+    }

     return (
       <DragWrapper
```

lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
```diff
@@ -219,112 +220,87 @@ class StoryPanel extends Component<Props, State> {
...
-            }
-          `}
+              : "80px",
+          left: "calc(50% - min(40%, 350px))",
+          width: "min(80%, 700px)"
+        }}
+      >
+        <Swipeable
+          onSwipedLeft={() => this.goToNextStory()}
+          onSwipedRight={() => this.goToPrevStory()}
```
niktayv: Potential gesture conflict between drag and swipe on touch devices.

DragWrapper and Swipeable both handle touch/mouse gestures. On touch devices, horizontal swipes for story navigation may conflict with drag operations, especially when swiping starts on the drag handle. 

I am not sure, but we may want to test this interaction or configure one to take precedence.

lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
ljowen: Sounds good, there's checks already where the panel gets launched but probably good to guard here too

lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
ljowen: Nice find @niktayv ! I've changed it so Swipeable is only on the body and footer (the drag handle is the title bar).

lib/ReactViews/StandardUserInterface/standard-user-interface.scss
```diff
@@ -125,7 +125,30 @@
...
+  }
+
+  & > div {
+    pointer-events: auto;
+    position: absolute;
+  }
+}
+
+.featureInfoFullScreen,
+.storyPanelFullscreen {
```
niktayv: A tiny nitpick. `featureInfoFullScreen` uses capital 'S' in "Screen" while `storyPanelFullscreen` uses lowercase 's'. 
Should we align the casing for consistency (e.g., storyPanelFullScreen)?
Would this require updating the TypeScript declaration file as well?

lib/ReactViews/Story/StoryPanel/StoryPanel.tsx
```diff
@@ -218,47 +219,23 @@ class StoryPanel extends Component<Props, State> {
...
-                padding-right: 50px;
-              `}
-              bottom: ${this.props.viewState.terria.timelineStack.top !==
-              undefined
-                ? "146px"
-                : "80px"};
-            }
-          `}
-        >
+              : "80px",
```
niktayv, line 233: Another little nitpick. 

The bottom positioning values 146px and 80px are magic numbers. 
Should we increase clarity about what they represent (e.g., timeline height offsets)?
Something like this:
```
+const BOTTOM_WITH_TIMELINE = "146px";
+const BOTTOM_WITHOUT_TIMELINE = "80px";
+
 // In render:
 style={{
   bottom:
     this.props.viewState.terria.timelineStack.top !== undefined
-      ? "146px"
-      : "80px",
+      ? BOTTOM_WITH_TIMELINE
+      : BOTTOM_WITHOUT_TIMELINE,
```

ljowen: normally I'd agree if the magic number was used more than once, but in this case it's immediately apparent from the expression what the numbers represent, having to see what the variable value is reduces clarity IMHO

lib/ReactViews/StandardUserInterface/standard-user-interface.scss
ljowen: Good pickup, fixed now

micholm (approved):
LGTM!

PR #7754, opened 2026-02-09
Configurable branding via init file

glughi:
### What this PR does

Added the ability to set `brandBarElements`, `brandBarSmallElements` and `displayOneBrand` in the init files. This way, using the init file in the URL you will be able to see brand elements other than the default ones.

### Test me

Add 
```
"parameters":{
        "brandBarElements": [
            "<a target=\"_blank\" href=\"https://www.bioretics.com\"><img src=\"https://www.bioretics.com/themes/bioretics/assets/images/Bioretics-logo.svg\" alt=\"Bioretics\"  height=\"52\" title=\"Bioretics\" /></a>"
        ]
    }
```
to an init file.

### Checklist


zoran995:
Hi, @glughi, I am not sure we want to support this in the open-source, as it opens the door to presenting the maps with different branding and might give the wrong impression about who owns the map (and the map owner might actually spend a lot of time crafting the catalogue)

glughi:
Hi @zoran995, we could add a parameter in the config to enable/disable this feature. At that point, if the admin then allows it, it is his responsibility.

glughi:
We would also have prepared the change of main theming colours (such as "`colorPrimary`", "`colorSecondary`", "`dark`", etc.) using init files.

PR #7756, opened 2026-02-11
Suppress TileErrors originating from Leaflet's empty img

### LINKED ISSUE #7388: Safari tile error when switching from 2d to 3d
When changing from 2d to 3d viewer in safari we get
```
Tile Error: Got a 200 (OK) response from URL
...
but it could not be loaded as an image.
```

To reproduce
- open https://map.terria.io/#share=s-dlDBYJS3O00Dk72ntDRtKSbkcxW
- open the map in 2d
- ensure the layer is visible
- change to 3D view

For future reference, after a quick look into the code, it seems like an error was thrown from the delayed update of the leaflet map while we have already moved to a 3D view. It looks like a race condition

 _Originally posted by @zoran995 in [#7384](https://github.com/TerriaJS/terriajs/issues/7384#issuecomment-2546367790)_

ljowen:
### What this PR does

Fixes #7388

When unloading imagery tiles, leaflet sets the image src to `data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=` (an transparent 1x1 gif) which errors in Safari.

 `TileErrorMixin` falsely treats this as a tile loading error and retries eventually disabling the layer.

This fixes the issue by suppressing tile errors originating from Leaflet's emtpy image.

See discussion here: https://github.com/Leaflet/Leaflet/issues/9311

When we upgrade to Leaflet 2.0.0 we can drop this check https://github.com/Leaflet/Leaflet/issues/10050

### Test me
View a WMS in 2d mode, change zoom and see error modal. After should show no Errors to user

Before: http://ci.terria.io/main/#share=s-d9E6fSHO0F6Az9ZMDNBNYl4IMiY 
After: http://ci.terria.io/leaflet-img-fix/#share=s-d9E6fSHO0F6Az9ZMDNBNYl4IMiY

### Checklist

### DIFF
--- a/lib/ModelMixins/TileErrorHandlerMixin.ts
+++ b/lib/ModelMixins/TileErrorHandlerMixin.ts
@@ -285,7 +285,7 @@ function TileErrorHandlerMixin<T extends AbstractConstructor<ModelType>>(
                 // This is a failed image element, which means we got a 200 response but
                 // could not load it as an image.
 
-                // If image element is Leaflet's emtpy pixel ignore this error (See: https://github.com/Leaflet/Leaflet/issues/9311)
+                // If image element is Leaflet's empty pixel ignore this error (See: https://github.com/Leaflet/Leaflet/issues/9311)
                 if (e.target.src === LEAFLET_EMPTY_IMAGE_URL) {
                   tellMapToSilentlyGiveUp();
                   return;
--- a/lib/ModelMixins/TileErrorHandlerMixin.ts
+++ b/lib/ModelMixins/TileErrorHandlerMixin.ts
@@ -285,7 +285,7 @@ function TileErrorHandlerMixin<T extends AbstractConstructor<ModelType>>(
 
                 // If image element is Leaflet's emtpy pixel ignore this error (See: https://github.com/Leaflet/Leaflet/issues/9311)
                 if (e.target.src === L.Util.emptyImageUrl) {
-                  tileProviderError.retry = false;
+                  tellMapToSilentlyGiveUp();
                   return;
                 }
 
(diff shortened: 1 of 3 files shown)


ljowen:
@pjonsson FYI this should fix the issues reported with DE SE

lib/ModelMixins/TileErrorHandlerMixin.ts
```diff
@@ -18,6 +18,7 @@ import ImageryProviderTraits from "../Traits/TraitsClasses/ImageryProviderTraits
 import MappableTraits from "../Traits/TraitsClasses/MappableTraits";
 import DiscretelyTimeVaryingMixin from "./DiscretelyTimeVaryingMixin";
 import MappableMixin from "./MappableMixin";
+import L from "leaflet";
```
ljowen: We could avoid importing leaflet by hardcoding the data: uri here

ljowen: Ended up hardcoding so Docs can be built

lib/ModelMixins/TileErrorHandlerMixin.ts
```diff
@@ -271,13 +273,24 @@ function TileErrorHandlerMixin<T extends AbstractConstructor<ModelType>>(
...
+
             if (e.statusCode === undefined) {
               if (runInAction(() => ignoreUnknownTileErrors)) {
                 tellMapToSilentlyGiveUp();
-              } else if ((e as any).target !== undefined) {
+              } else if (e.target !== undefined) {
                 // This is a failed image element, which means we got a 200 response but
                 // could not load it as an image.
+
+                // If image element is Leaflet's emtpy pixel ignore this error (See: https://github.com/Leaflet/Leaflet/issues/9311)
```
pjonsson: "emtpy" => "empty"

pjonsson:
Wonderful, thank you!

micholm (approved):
LGTM

lib/ModelMixins/TileErrorHandlerMixin.ts
ljowen:
```suggestion
                // If image element is Leaflet's empty pixel ignore this error (See: https://github.com/Leaflet/Leaflet/issues/9311)
```

pjonsson:
@ljowen I haven't done the testing but apparently this happens on other browsers too, and it seems to be far more frequent when running "GlobalProtect".

PR #7758, opened 2026-02-12
Share the default timeline settings.

na9da:
### What this PR does

In v7 it used to be possible to share the default timeline state. Re-implement it in v8 so that we can capture scenes showing shadows at a particular time of the day in stories.

Note that the default timeline state is captured in the init source as a model with ID `defaultTimeline` instead of introducing a hardcoded config key for timeline settings.

### Test me

Run through [this story](http://ci.terria.io/share-default-timeline/#share=s-nft9MX2L285f1Ho2gnfapWBQoZP), observe that it captures different times of the day and the timeline playback state.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 #### next release (8.11.4)
 
+- Capture default timeline state in share links including current time,
+  playback etc.
 - [The next improvement]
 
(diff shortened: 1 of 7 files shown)


micholm (approved):
awesome! LGTM

PR #7761, opened 2026-02-13
Fix use of maximumScale in ArcGisMapServer

### LINKED ISSUE #7166: Unused maxScale of ArcGisMapServer
Even if server side has been set a `maxScale`, this is not used and the layer remains visible (badly) even beyond that scale.
Perhaps if `hideLayerAfterMinScaleDenominator` is `true`, the layer should no longer be visible beyond `maxScale`.

An example server where `maxScale` is used (the same in QGIS works as expected):
https://servizigis.regione.emilia-romagna.it/geoags/rest/services/portale/prit2025/MapServer

glughi:
### What this PR does

Fixes #7166

### Test me

See linked issue.

### Checklist


glughi:
Any news?

zoran995 (changes requested):
Hi @glughi, sorry it took a long time to review this. It might be better to fix this on the MapServerStratum level instead of overriding the trait value as we already do for ArcGisFeatureServer in ArcGisFeatureServerStratum

PR #7760, opened 2026-02-13
Added duplicate management for GeoJson, Gpx, and Kml CatalogItems

glughi:
### What this PR does

Fixes #7757

### Test me

See the linked issue.

### Checklist

- [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
- [ ] I've updated relevant documentation in `doc/`.
- [ ] I've updated CHANGES.md with what I changed.
- [x] I've provided instructions in the PR description on how to test this PR.

PR #7762, opened 2026-02-13
fix: splitter fails for local file catalog items

### LINKED ISSUE #7757: Error on compare vector data
Enabling "Compare" on a locally loaded vector layer (in this case, a GeoJSON polygon) gives the error shown below and the _"(copy)"_ layer is added, but without data.

Image

Image

The polygon code:
```
{
  "type": "Feature",
  "geometry": {
    "type": "Polygon",
    "coordinates": [[
      [ -73.958, 40.8003 ],
      [ -73.9498, 40.7968 ],
      [ -73.9737, 40.7648 ],
      [ -73.9814, 40.7681 ],
      [ -73.958, 40.8003 ] 
    ]]
  },
  "properties": {
    "name": "Example Polygon"
  }
}
```

zoran995:
### What this PR does

Fixes #7757

Store local file data as blob URLs in the URL trait instead of private instance fields, and follow the same approach already used for GLTF and Assimp catalogue items. This ensures duplicateModel() preserves the data since traits are copied during duplication. One potential downside of this approach is that we are currently not disposing the catalogue items from memory on remove (here is a ticket to track that https://github.com/TerriaJS/terriajs/issues/4816) 

### Test me

### Checklist

### DIFF
--- a/lib/Models/Catalog/CatalogItems/CsvCatalogItem.ts
+++ b/lib/Models/Catalog/CatalogItems/CsvCatalogItem.ts
@@ -1,5 +1,5 @@
 import i18next from "i18next";
-import { computed, makeObservable, override, runInAction } from "mobx";
+import { action, computed, makeObservable, override, runInAction } from "mobx";
 import isDefined from "../../../Core/isDefined";
 import TerriaError from "../../../Core/TerriaError";
 import AutoRefreshingMixin from "../../../ModelMixins/AutoRefreshingMixin";
@@ -8,6 +8,7 @@ import UrlMixin from "../../../ModelMixins/UrlMixin";
 import Csv from "../../../Table/Csv";
 import TableAutomaticStylesStratum from "../../../Table/TableAutomaticStylesStratum";
 import CsvCatalogItemTraits from "../../../Traits/TraitsClasses/CsvCatalogItemTraits";
+import CommonStrata from "../../Definition/CommonStrata";
 import CreateModel from "../../Definition/CreateModel";
 import { BaseModel } from "../../Definition/Model";
 import StratumOrder from "../../Definition/StratumOrder";
@@ -35,8 +36,6 @@ export default class CsvCatalogItem
     return "csv";
   }
 
-  private _csvFile?: File;
-
   constructor(
     id: string | undefined,
     terria: Terria,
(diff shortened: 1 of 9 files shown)


zoran995:
Deployment is failing due to the branch name

zoran995:
This fixes the same bug as #7760, @ljowen, @na9da can you review and see which approach you think is better

na9da:
Hi @zoran995 - both PRs look good to me.

One thing to check with this approach is whether it causes unnecessary duplication of memory because we are creating an object URL and then `fetching` from it again. My guess is the other strategy of keeping reference to the file might be slightly more memory efficient? You could use the devtools memory profiler to test the theory.

(here's a [sizable geojson](https://data.melbourne.vic.gov.au/explore/dataset/2023-building-footprints/export/) file for testing)

zoran995:
@na9da, both branches have the same memory footprint, which matches my expectations. Both PRs will store the parsed result twice in memory, whereas the source file data is stored only once. URL.createObjectURL creates a pointer to a memory location where the data is stored. 
- In #7760, a duplicate will create a reference from the new item to the old item file and store it like that. Since it is added automatically to the map in `forceLoadMapItems` and `forceLoadGeojsonData`, it will parse that data.
- In this PR, it will create a memory pointer to data on initial load, and in `forceLoadMapItems` and `forceLoadGeojsonData`, it will parse data and store the parsed result as `_dataSource` or `_imageryProvider`.

lib/Models/Catalog/CatalogItems/CsvCatalogItem.ts
```diff
@@ -54,22 +53,23 @@ export default class CsvCatalogItem
     return CsvCatalogItem.type;
   }
 
+  @action
   setFileInput(file: File) {
-    this._csvFile = file;
+    this.setTrait(
+      CommonStrata.user,
+      "url",
+      URL.createObjectURL(file) + "#" + file.name
```
na9da, line 61: Nitpick: any reason to add the URI fragment, could this be a problem later when we implement revoking? Same for the other types.

test/Models/Catalog/CatalogReferences/SplitItemReferenceSpec.ts
```diff
@@ -22,4 +24,39 @@ describe("SplitItemReference", function () {
...
+    const clone = splitRef.target;
+    expect(clone).toBeDefined();
+    expect(clone instanceof GeoJsonCatalogItem).toBe(true);
+
+    const cloneGeoJson = clone as GeoJsonCatalogItem;
+    expect(cloneGeoJson.url).toBe(sourceItem.url);
+    expect(HasLocalData.is(cloneGeoJson)).toBe(true);
+    if (HasLocalData.is(cloneGeoJson)) {
+      expect(cloneGeoJson.hasLocalData).toBe(true);
+    }
```
na9da, line 61: Recommendation: Should we add a manual revoke here?

na9da (approved):
Hi @zoran995 - thanks for testing the memory usage.

I've left a couple of minor comments.

lib/Models/Catalog/CatalogItems/CsvCatalogItem.ts
zoran995: GeoJSON uses it to determine if the file is zip so I used the same pattern

zoran995: To revoke, we would need to split the URL at the hash, but we will need some registry for disposing of data, so when we add that, we can store it without fileName

PR #7766, opened 2026-02-17
 Add ADR for replacing jasmine-ajax with MSW

zoran995:
### What this PR does

Part of https://github.com/TerriaJS/terriajs/issues/7768
Proposes adopting Mock Service Worker (MSW) v2 as the single HTTP mocking solution for all tests, replacing jasmine-ajax, fetchMock, and the loadWithXhr spy pattern.                                                                                                      
                                                                                                                                        
The test suite currently uses three different HTTP mocking approaches, which creates framework coupling (jasmine-ajax is incompatible with Jest/Vitest), incomplete coverage (XHR vs fetch), and a maintenance burden. MSW intercepts at the Service Worker level, handling both XHR and fetch transparently without monkey-patching globals.

The ADR covers:
- Context on current fragmentation and its problems
- Decision to adopt MSW v2 with infrastructure pattern
- Five alternatives considered with rejection rationales
- Consequences (positive, negative, neutral) and known gotchas
- Migration patterns with before/after examples

### Test me

This is a documentation-only change (ADR). Review the ADR at architecture/0013-msw-for-http-mocking.md.

### Checklist

- [x] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist) - adr only

### DIFF
--- a/architecture/0013-msw-for-http-mocking.md
+++ b/architecture/0013-msw-for-http-mocking.md
@@ -1,0 +1,334 @@
+# 13. Mock Service Worker (MSW) for HTTP Mocking in Tests
+
+Date: 2026-02-17
+
+## Status
+
+Proposed
+
+## Context
+
+TerriaJS is a large geospatial web application with extensive HTTP-based testing requirements. The codebase currently uses multiple approaches for HTTP mocking in tests:
+
+1. **jasmine-ajax**: Intercepts XHR requests by monkey-patching the XMLHttpRequest constructor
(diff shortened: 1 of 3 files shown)


architecture/0013-msw-for-http-mocking.md
```diff
@@ -0,0 +1,334 @@
+# 13. Mock Service Worker (MSW) for HTTP Mocking in Tests
```
meiqimichelle: This uses the same ADR number as https://github.com/TerriaJS/terriajs/pull/7764/changes

ljowen (approved):

PR #7786, opened 2026-02-25
Upgrade eslint and use flat config

zoran995:
### What this PR does

Migrate to eslint flat config as eslint removed support for legacy configs in v10.

### Test me

### Checklist


eslint.config.mjs
```diff
@@ -0,0 +1,230 @@
...
+      },
+      globals: {
+        ...globals.browser,
+        ...globals.commonjs,
+        process: true
+      }
+    },
+    settings: {
+      react: {
+        version: "18.3"
```
pjonsson, line 51: Can't this use `"detect"` as version?

eslint.config.mjs
```diff
@@ -0,0 +1,230 @@
...
+          caughtErrorsIgnorePattern: "^_",
+          destructuredArrayIgnorePattern: "^_",
+          ignoreRestSiblings: true
+        }
+      ],
+
+      "no-useless-constructor": "off",
+      "@typescript-eslint/no-useless-constructor": "error",
+      // @TODO: revise these rules
+      "@typescript-eslint/consistent-type-assertions": "error"
```
pjonsson, line 191: Can `"react/prop-types": "warn"` that was disabled in #7382 be re-instated?

gulpfile.js
```diff
@@ -1,7 +1,3 @@
-/*eslint-env node*/
-/*eslint no-sync: 0*/
-/* eslint-disable @typescript-eslint/no-require-imports */
-
```
pjonsson, line 4: ESLint 9 automatically warns for `--report-unused-disable-directives`, so that parameter could possibly be dropped from the lint function below.

eslint.config.mjs
```diff
@@ -0,0 +1,230 @@
...
+
+  js.configs.recommended,
+  tseslint.configs.recommended,
+  react.configs.flat.recommended,
+  react.configs.flat["jsx-runtime"],
+  reactHooks.configs.flat.recommended,
+  {
+    name: "terriajs/include-files",
+    files: [
+      "**/*.jsx",
```
pjonsson, line 23: I think I measured runtime when I did the ESLint 9 config, and "**/*.{jsx,tsx,js,ts,cjs,mjs}" (no spaces after the commas) is faster than a list of multiple globs.

eslint.config.mjs
```diff
@@ -0,0 +1,230 @@
...
+
+import tsParser from "@typescript-eslint/parser";
+import js from "@eslint/js";
+import prettier from "eslint-config-prettier";
+import react from "eslint-plugin-react";
+import reactHooks from "eslint-plugin-react-hooks";
+import globals from "globals";
+import tseslint from "typescript-eslint";
+
+export default defineConfig([
```
pjonsson, line 12: I'm not sure where I found it when I was working on the ESLint 9 upgrade, but if you add `/** @type {import('eslint').Linter.Config[]} */` before this line, I think you'll get type checking on the validity of the config directives you use. (`defineConfig()` seems to be a version of the now deprecated `tselint.config()` that accepts an array of configs instead of using varargs.)

zoran995: It is already type checked by defineConfig

eslint.config.mjs
zoran995: No, eslint v10 is released recently and eslint-plugin-react throws error when detect is used

eslint.config.mjs
zoran995: It is error by default

ljowen (approved):

PR #7796, opened 2026-03-04
migrate to msw - 3rd batch

zoran995:
### What this PR does

Third batch of migrating tests to msw

### Test me

### Checklist

### DIFF
--- a/test/Models/Catalog/CatalogItems/SensorObservationServiceCatalogItemSpec.ts
+++ b/test/Models/Catalog/CatalogItems/SensorObservationServiceCatalogItemSpec.ts
@@ -1,3 +1,4 @@
+import { http, HttpResponse } from "msw";
 import { runInAction } from "mobx";
 import CommonStrata from "../../../../lib/Models/Definition/CommonStrata";
 import SensorObservationServiceCatalogItem from "../../../../lib/Models/Catalog/Ows/SensorObservationServiceCatalogItem";
@@ -5,6 +6,7 @@ import Terria from "../../../../lib/Models/Terria";
 import SimpleCatalogItem from "../../../Helpers/SimpleCatalogItem";
 import TableAutomaticStylesStratum from "../../../../lib/Table/TableAutomaticStylesStratum";
 import { isEnum } from "../../../../lib/Models/SelectableDimensions/SelectableDimensions";
+import { worker } from "../../../mocks/browser";
 
 import GetFeatureOfInterestResponse from "../../../../wwwroot/test/sos/GetFeatureOfInterestResponse.xml";
 import EmptyGetFeatureOfInterestResponse from "../../../../wwwroot/test/sos/GetFeatureOfInterestResponse_NoMembers.xml";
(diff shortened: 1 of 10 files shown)


test/Models/Catalog/CatalogItems/SensorObservationServiceCatalogItemSpec.ts
```diff
@@ -251,49 +330,55 @@ describe("SensorObservationServiceCatalogItem", function () {
...
-        expect(req.data()).toContain(
-          "<sos:featureOfInterest>feature1</sos:featureOfInterest>"
-        );
-        expect(req.data()).toContain(
-          "<gml:beginPosition>2020-01-26T03:56:15.025Z</gml:beginPosition>"
-        );
-        expect(req.data()).toContain(
-          "<gml:endPosition>2020-03-26T03:56:15.025Z</gml:endPosition>"
-        );
+        const result = await runInAction(() => item.loadMapItems());
```
ljowen, line 331: Should there be a mock here erroring unless the request i.e. contains "sos:GetObservation"? This new test doesn;t do much

test/Models/Catalog/Ows/WebProcessingServiceCatalogGroupSpec.ts
```diff
@@ -33,36 +26,47 @@ describe("WebProcessingServiceCatalogGroup", function () {
...
+          ) {
+            return HttpResponse.error();
+          }
+          return new HttpResponse(GetCapabilitiesXml, {
+            headers: { "Content-Type": "text/xml" }
+          });
+        })
       );
-      expect(request.method).toEqual("GET");
+      await wpsGroup.loadMembers();
```
ljowen:
```suggestion
```

test/Models/Catalog/CatalogItems/SensorObservationServiceCatalogItemSpec.ts
```diff
@@ -89,55 +83,111 @@ describe("SensorObservationServiceCatalogItem", function () {
...
+              !body.includes(
+                "<sos:observedProperty>Ph</sos:observedProperty>"
+              ) ||
+              !body.includes("<sos:procedure>Daily Mean</sos:procedure>") ||
+              !body.includes("<sos:procedure>Yearly Mean</sos:procedure>")
+            ) {
+              return HttpResponse.error();
+            }
+
+            if (/<sos:GetFeatureOfInterest/.test(body)) {
```
ljowen: this will always be false with L:92 above

ljowen (approved):
Few small comments, otherwise ran tests locally and looks good :)

test/Models/Catalog/CatalogItems/SensorObservationServiceCatalogItemSpec.ts
zoran995: L92 tests if it doesn't include `sos:GetFeatureOfInterest`, so this will always be true

test/Models/Catalog/CatalogItems/SensorObservationServiceCatalogItemSpec.ts
zoran995: With MSW, we should avoid asserting requests, but the worker will only respond if the request parameters are correct. We could move the worker definition to be per test, so it is more explicit

PR #7807, opened 2026-03-18
Jasmine upgrade

zoran995:
### What this PR does

Upgrade Jasmine to version 6. The biggest change is that the test suite is no longer served on port 3002 but is using actual Jasmine server, so we don't need to maintain version of jasmine files locally. When you run `yarn gulp dev` it will build the terriajs tests and serve the Jasmine at `http://localhost:9876/`.

Changes:
- upgrade jasmine to the latest version
- update test syntax to Jasmine's new syntax, mainly around async stubs. We might need a follow-up to update deprecated methods. 
- removes karma from dependencies
- updates tests that missed await on async methods

### Test me

### Checklist


lib/Map/Cesium/CesiumRenderLoopPauser.ts
```diff
@@ -133,7 +133,9 @@ export default class CesiumRenderLoopPauser {
       timeout: any,
       returnType: any
     ) => {
-      deferred.promise.finally(this._boundNotifyRepaintRequired);
+      deferred.promise
+        .finally(this._boundNotifyRepaintRequired)
+        .catch(() => {});
```
zoran995, line 138: This caused a bunch of issues with tests, as some requests were failing after the test suite is completed, which then triggers an unhandled rejection, which is reported as a test failure in new versions of Jasmine. Catching the error here was the only way to fix those tests, and it seems fine as this only trigger repaint event

test/ReactViews/DimensionSelectorSectionSpec.tsx
```diff
@@ -204,67 +204,67 @@ describe("DimensionSelectorSection", function () {
     ).toBeVisible();
   });
 
-  describe("when given a SelectableDimensionCheckboxGroup", function () {
```
ljowen, line 207: Could you add a comment why this has been disabled?

zoran995: It fails as there are no assertions in the test suite, and to me, it looks like it was never completely implemented

lib/Map/Cesium/CesiumRenderLoopPauser.ts
na9da: That looks ok to me.

test/jasmine-browser.mjs
na9da, line 1: Suggestion: Could this be under `test/support` or `buildprocess` just to avoid creating another top-level dir?
Is there any other reason to have it here?

test/jasmine-browser.mjs
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+  srcDir: "wwwroot",
+  srcFiles: [],
+  specDir: "wwwroot",
+  specFiles: ["build/TerriaJS-specs.js"],
+  helpers: [],
+
+  env: {
+    random: false
```
na9da, line 16: Is `random` triggering errors? Ideally we should enable it, but that could be another PR.

test/ModelMixins/TimeFilterMixinSpec.ts
```diff
@@ -88,7 +88,7 @@ describe("TimeFilterMixin", function () {
         fakeImageryFeature.properties = {
           availabilityAtLocation: ["2023-01-02", "2023-01-03"]
         };
-        spyOn(imageryProvider, "pickFeatures").and.returnValue(
+        spyOn(imageryProvider, "pickFeatures").and.callFake(() =>
```
na9da, line 91: doubt: I see this pattern of change from `returnValue()` to `callFake()` - why is it needed?

na9da:
@zoran995 - looks great! Just one suggestion.

na9da (changes requested):

na9da (approved):

test/jasmine-browser.mjs
zoran995: Yeah, tests start failing with random enabled, it will take more work to make them truly random

PR #7817, opened 2026-03-31
Small refactor of analytics

zoran995:
### What this PR does

Part of simplifying the terria initialisation process, refactor the analytics code
- require the analytics to be supplied as we already do from TerriaMap
- implement NoopAnalytics so analytics is always defined 
- In the future, we should require analytics to be fully configured from TerriaMap

### Test me

### Checklist


lib/Models/Terria.ts
```diff
@@ -761,14 +745,7 @@ export default class Terria {
...
     (buildModuleUrl as any).setBaseUrl(this.cesiumBaseUrl);
 
-    this.analytics = options.analytics;
-    if (!defined(this.analytics)) {
-      if (typeof window !== "undefined" && defined((window as any).ga)) {
-        this.analytics = new GoogleAnalytics();
-      } else {
-        this.analytics = new ConsoleAnalytics();
-      }
-    }
```
zoran995, line 771: I don't see how the window.ga can be defined if it wasn't loaded and configured already as react-ga4 shouldn't attach it to window until intialize is called

lib/Core/Analytics/types.ts
```diff
@@ -0,0 +1,15 @@
+export interface Analytics {
```
zoran995, line 1: Move analytics types to their directory instead of importing from `Models/Terria.ts`. Although it is a type-only import, it is still a cross-module import

lib/Models/Terria.ts
na9da: True. Should we default to `GoogleAnalytics` instead of `NoopAnalytics` if `googleAnalyticsKey` is set? At least from [this doc comment](https://github.com/TerriaJS/terriajs/blob/bc7494fb514b373420fa1bc3d6281f578f7704fe/lib/Models/Terria.ts#L262) it reads like it should.

na9da (changes requested):
Looks good @zoran995.

Couple of comments:
- Should we auto-default to GA if `googleAnalyticsKey` is set? At least the doc comment reads like that to me.
- Can we keep to capitalized directory names? `lib/Core/Analytics` instead of `lib/Core/analytics`?

lib/Models/Terria.ts
zoran995: In my understanding, this wasn't happening in the past, and I would avoid changing defaults, as the idea is to request a fully configured analytics instance from TerriaMap when we finish with the loaders refactor.

na9da (approved):
LGTM. I guess error is from Terriamap?

zoran995:
Yeah, it was failing due to not updated imports on the TerriaMap side after changing the folder name casing

PR #7819, opened 2026-04-03
extract terria config to observable store

### LINKED ISSUE #7816: Decouple config and catalogue loading from TerriaJS core
TerriaJS currently assumes that config and catalogue are loaded from URLs via internal fetching logic. This creates several problems
- Loader is not extensible - apps consuming TerriaJS cannot control how config/catalogue is loaded or pass additional parameters to those requests.
- Magda coupling - Support for loading catalogue definition from Magda was added several years ago, significantly complicating initialisation logic. This made the loading logic much harder to reason about and maintain. We are not aware of any active deployment using this feature.
- URL and routing dependency - TerriaJS has the ability to load config and catalogue definition from URL, along with the share URLs, tightly coupling it to the app's routing mechanism. This makes it harder to integrate with external apps.

Loading those resources needs to happen independently of how they will be used, so we are planning to make several changes:
- Accept preloaded data at initialisation. Change TerriaJS to accept config and catalogue definitions as resolved data, moving the responsibility for loading into the host application (e.g., TerriaMap).
- Implement default config and catalogue loader. Expose the current URL-based loading logic as default loader functions, so TerriaMap and similar hosts remain lean, while custom implementations can provide their own.
- Deprecate loading of Magda catalogue aspects. Remove the Magda aspect-loading code path from core. If community demand exists, this can be reintroduced as a custom loader.
- Decouple share URL handling from URL structure. Rework the share URL resolution so TerriaJS does not depend on the host application's URL conventions.

This is the groundwork for having TerriaJS behave more as a library and less as a self-contained app.

zoran995:
### What this PR does

Part of: #7816 

Extract config parameters from the Terria class and create an observable data store.

### Test me

Ensure that all existing parameters are available in the new store.

### Checklist

### DIFF
--- a/doc/customizing/client-side-config.md
+++ b/doc/customizing/client-side-config.md
@@ -56,7 +56,7 @@ It is also possible to add version 7 init files using the `v7initializationUrls`
 
 ## Parameters
 
-**The best reference for now is [`interface ConfigParameters`](https://github.com/TerriaJS/terriajs/blob/main/lib/Models/Terria.ts#L101) (you may have to search for `interface ConfigParameters` on that page to find it if future code changes change line numbers).**
+**The best reference is [`interface ConfigParameters`](../../lib/Models/TerriaConfig.ts#L14).**
 
 Specifies various options for configuring TerriaJS:
 
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -111,8 +111,6 @@ import { isViewerMode, setViewerMode } from "./ViewerMode";
 import Workbench from "./Workbench";
 import SelectableDimensionWorkflow from "./Workflows/SelectableDimensionWorkflow";
 
-export type { ConfigParameters } from "./TerriaConfig";
-
 interface StartOptions {
   configUrl: string;
   configUrlHeaders?: {
--- a/lib/Models/Terria.ts
+++ b/lib/Models/Terria.ts
@@ -671,7 +671,7 @@ export default class Terria {
       }
       runInAction(() => {
         if (isJsonObject(config) && isJsonObject(config.parameters)) {
-          this.applyConfig(config.parameters);
+          this.updateConfig(config.parameters);
         }
         this.setupInitializationUrls(baseUri, config);
       });
@@ -974,7 +974,7 @@ export default class Terria {
   }
 
   @action
-  applyConfig(config: Partial<ConfigParameters>): void {
+  updateConfig(config: Partial<ConfigParameters>): void {
     this.configParameters.apply(config);
   }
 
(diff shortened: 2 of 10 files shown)


doc/customizing/client-side-config.md
```diff
@@ -56,7 +56,7 @@ It is also possible to add version 7 init files using the `v7initializationUrls`
 
 ## Parameters
 
-**The best reference for now is [`interface ConfigParameters`](https://github.com/TerriaJS/terriajs/blob/main/lib/Models/Terria.ts#L101) (you may have to search for `interface ConfigParameters` on that page to find it if future code changes change line numbers).**
+**The best reference for now is [`interface ConfigParameters`](../../lib/Models/TerriaConfig.ts#L14).**
```
ljowen: Can we remove the "For now" part? ideally this isn't going to change in the near term

lib/Models/Terria.ts
```diff
@@ -1324,15 +974,8 @@ export default class Terria {
...
-  updateParameters(parameters: ConfigParameters | JsonObject): void {
-    Object.entries(parameters).forEach(([key, value]) => {
-      if (Object.hasOwnProperty.call(this.configParameters, key)) {
-        (this.configParameters as any)[key] = value;
-      }
-    });
-
-    this.appName = this.configParameters.appName ?? this.appName;
-    this.supportEmail = this.configParameters.supportEmail ?? this.supportEmail;
+  applyConfig(config: Partial<ConfigParameters>): void {
```
ljowen: Given this is a potentially breaking change (i.e. Maps could be calling `terria.updateParameters()` directly should this be part of a major release? I couldn't find examples in our own map repos though

lib/Models/TerriaConfig.ts
```diff
@@ -0,0 +1,449 @@
...
+ * Apply config from any source in any order — last write wins per field,
+ * `undefined` values do not overwrite existing values.
+ *
+ * @example
+ * const config = new TerriaConfig();
+ * config.apply(await loadConfig("config.json"));
+ * config.apply(parseHashParams(location.hash));
+ * config.apply(readLocalStorage(localStorage));
+ */
+export class TerriaConfig {
```
na9da: Should this do an `implements ConfigParameters`?

lib/Models/Terria.ts
na9da: Suggestion: naming is subjective, just my 2 cents: I find `update` more common/guessable than `apply`. So `updateConfig` might be more suitable?

na9da (changes requested):
Hi @zoran995 - looks great. Just a couple of suggestions.

I have an upcoming PR that adds some cesium options to the config, some of the options will also be stored in local storage - so i'm planning to use zod to make parsing simpler and make sure values are sane. We could use zod if we want to actually parse the config but maybe we keep it simple for now.

lib/Models/TerriaConfig.ts
zoran995: It was better to merge everything into the class and infer the types

zoran995:
superseeded by https://github.com/TerriaJS/terriajs/pull/7876

PR #7820, opened 2026-04-08
Publish config

zoran995:
### What this PR does

The last few releases had issues publishing to npm, and we had to do multiple releases because the version-detection script detected a change only in the last 2 commits. It is more error-prone to check if the version is actually published to npm and, if not, try publishing it (if publishing fails, every merge to main after that will trigger the publish).
Also, we now detect the NPM_TAG from the actual version, so we can publish prereleases by just bumping the version in `package.json`. 

### Test me

### Checklist

- [ ] There are unit tests to verify my changes are correct or unit tests aren't applicable

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -23,24 +23,26 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
-        with:
-          fetch-depth: 2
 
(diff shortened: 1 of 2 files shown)


ljowen (approved):

.github/workflows/npm-publish.yml
ljowen, line 1: should we update `RELEASE_GUIDE.md` to mention how the published version detection now works?

PR #7824, opened 2026-04-14
Fix TypeScript errors from terriajs-cesium workspace package 

zoran995:
### What this PR does

Cesium started adding `@ts-check` to their JS source files. Because terriajs-cesium can be a yarn workspace package symlinked into node_modules, TypeScript resolves the symlink to the real path outside node_modules and reports errors from those files. In the previous release of cesium there were only 3 errors, but in `v1.140`, there is over 150.

`preserveSymlinks: true` keeps the symlink path intact, treats the cesium source as library code, and suppresses error reporting.

### Test me

- Try linking the terriajs-cesium package in the TerriaMap repo and observe that no type errors are reported. 
- Confirm that terriajs type errors are still reported.

### Checklist

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,6 +26,7 @@
       "../../../node_modules/",
       "../../../node_modules/@types/"
     ],
+    "preserveSymlinks": true,
     "types": ["terriajs-cesium", "@testing-library/jasmine-dom"]
   },
   "include": ["./lib/**/*", "./test/**/*", "./buildprocess/**/*"]


na9da:
@zoran995 

How does this affect terriajs checkouts in a terriamap workspace?

One alternate approach might be to instruct fork-ts-checker to ignore issues from cesium, eg:
```js
config.plugins.push(
    new ForkTsCheckerWebpackPlugin({
      typescript: {
        memoryLimit: 4096,
        configFile: path.resolve(__dirname, "..", "tsconfig.json"),
        diagnosticOptions: {
          semantic: true,
          syntactic: true
        }
      },
      issue: {
        exclude: [
          // FIXME: temporary fix to ignore type errors from use @ts-check in Cesium.
          (issue) => issue.file.includes(path.join("terriajs-cesium", "packages", "engine"))
        ]
      }
    })
  )
```

zoran995:
Typechecking for terriajs works as is because we explicitly include those files in the TypeScript config. The difference to your suggestion is that this solution will work across all possible ways to run the TypeScript compiler, and it will actually skip checking those files, whereas with `ForkTsCheckerWebpackPlugin`, it will just stop outputting those errors. I am open to discussing it further, as the main goal is to have a temporary fix until Cesium resolves those issues

soyarsauce (approved):
`preserveSymlinks` is an easy tweak if it needs to change; as you say both work, Nanda's idea is a bit more config, let's simply go with `preserveSymlinks`

PR #7825, opened 2026-04-15
Fix search debounce:

ljowen:
Wrap search functions in useCallback to prevent debounce flushing every render

### What this PR does

Currently search isn't actually debounced because a new function is created every render, wrap the `onDoSearch` props in `useCallback` to prevent this.

### Test me
**Before:** http://ci.terria.io/main  Location or Catalog, search happens on every keystroke (i.e. view network tab)
**After:** http://ci.terria.io/search-debounce-fix Search is properly debounced

### Checklist


zoran995:
just for reference I fixed this also as a part of #7682 and #7715

zoran995:
this should be resolved as part of https://github.com/TerriaJS/terriajs/pull/7682

PR #7834, opened 2026-04-23
Catalog search provider

zoran995:
### What this PR does

Move the catalogue search provider from `SearchBarModel` to the `Catalog` class. I always felt that the catalogue search provider doesn't belong in SearchBarModel, but at the time I was writing it, I wasn't aware of `Catalog`, so it stayed in the model.

### Test me

### Checklist

### DIFF
--- a/lib/Models/Catalog/Catalog.ts
+++ b/lib/Models/Catalog/Catalog.ts
@@ -7,18 +7,33 @@ import Terria from "../Terria";
 import Group from "./Group";
 import { BaseModel } from "../Definition/Model";
 import isDefined from "../../Core/isDefined";
+import CatalogSearchProviderMixin from "../../ModelMixins/SearchProviders/CatalogSearchProviderMixin";
+import CatalogSearchProvider from "../SearchProviders/CatalogSearchProvider";
 
 export default class Catalog {
   @observable
   group: Group & BaseModel;
 
+  @observable
+  searchProvider: CatalogSearchProviderMixin.Instance | undefined;
+
   readonly terria: Terria;
 
   private _disposeCreateUserAddedGroup: () => void;
 
-  constructor(terria: Terria) {
+  constructor(
+    terria: Terria,
+    searchProvider?: CatalogSearchProviderMixin.Instance
+  ) {
     makeObservable(this);
     this.terria = terria;
+    this.searchProvider =
+      searchProvider ??
+      new CatalogSearchProvider(
+        "catalog-search-provider",
+        terria,
+        terria.searchBarModel.minCharacters
+      );
     this.group = new CatalogGroup("/", this.terria);
     this.terria.addModel(this.group);
 
--- a/lib/Models/Catalog/Catalog.ts
+++ b/lib/Models/Catalog/Catalog.ts
@@ -23,17 +23,20 @@ export default class Catalog {
 
   constructor(
     terria: Terria,
-    searchProvider?: CatalogSearchProviderMixin.Instance
+    options: { searchProvider?: CatalogSearchProviderMixin.Instance } = {}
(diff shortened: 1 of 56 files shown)


lib/Models/Catalog/Catalog.ts
```diff
@@ -7,18 +7,33 @@ import Terria from "../Terria";
...
+  ) {
     makeObservable(this);
     this.terria = terria;
+    this.searchProvider =
+      searchProvider ??
+      new CatalogSearchProvider(
+        "catalog-search-provider",
+        terria,
+        terria.searchBarModel.minCharacters
+      );
```
na9da: Suggestion: I know we have done this elsewhere in the past but it might be a good idea to stick with [standard Constructor](https://github.com/TerriaJS/terriajs/blob/2d6577eb740e76adfbe8321f7e46e3062ced64ee/lib/Models/Definition/Model.ts#L20-L22) parameters for models? Using custom constructor parameters is fine if the model is always initialised manually but if at some point we do automatic initialisation from some config, it could create problems. 

One other pattern to make creating the instance easier would be to define a static method like `CatalogSearchProvider.fromOptions({minCharacters})`

lib/Models/Catalog/Catalog.ts
```diff
@@ -7,18 +7,33 @@ import Terria from "../Terria";
...
+import CatalogSearchProviderMixin from "../../ModelMixins/SearchProviders/CatalogSearchProviderMixin";
+import CatalogSearchProvider from "../SearchProviders/CatalogSearchProvider";
 
 export default class Catalog {
   @observable
   group: Group & BaseModel;
 
+  @observable
+  searchProvider: CatalogSearchProviderMixin.Instance | undefined;
+
```
na9da, line 19: Can this be `undefined` if always intialised in constructor?

na9da (changes requested):
Hi @zoran995 - looks great.

One small thing I noticed is we do a cast here to access a property not part of the interface: 

https://github.com/TerriaJS/terriajs/blob/2d6577eb740e76adfbe8321f7e46e3062ced64ee/lib/ReactViews/ExplorerWindow/Tabs/DataCatalogTab.tsx#L57-L62

We should maybe make it part of the interface or traits instead.

lib/Models/Catalog/Catalog.ts
zoran995: changed it to detect differently if the catalogue search provider should be initialised, so we can give the possibility to make it undefined

lib/Models/Catalog/Catalog.ts
zoran995: ended up implementing a getter override that will fallback to the searchBar param value

lib/Models/SearchProviders/CatalogSearchProvider.ts
```diff
@@ -107,24 +106,23 @@ export default class CatalogSearchProvider extends CatalogSearchProviderMixin(
...
 
   get type() {
     return CatalogSearchProvider.type;
   }
 
+  @override
+  get minCharacters(): number | undefined {
+    return super.minCharacters ?? this.terria.searchBarModel.minCharacters;
+  }
+
```
na9da: AFAICT, `.minCharacters` is used only from `SearchProviderMixin`. So, potentially, we could move this logic there instead of the implementer having to override it?

na9da (approved):
LGTM. I've left one additional comment.

lib/Models/SearchProviders/CatalogSearchProvider.ts
zoran995: moved to catalog search provider mixin

PR #7855, opened 2026-06-06
refactor: move catalog index from terria.ts to catalog instance

zoran995:
### What this PR does

Move catalog index from the terria class to catalog as it is closely related to catalog instance

### Test me

### Checklist

### DIFF
--- a/lib/ModelMixins/SearchProviders/CatalogSearchProviderMixin.ts
+++ b/lib/ModelMixins/SearchProviders/CatalogSearchProviderMixin.ts
@@ -19,8 +19,9 @@ function CatalogSearchProviderMixin<
 
     @computed get resultsAreReferences() {
       return (
-        isDefined(this.terria.catalogIndex?.loadPromise) &&
-        fromPromise(this.terria.catalogIndex!.loadPromise).state === "fulfilled"
+        isDefined(this.terria.catalog.index?.loadPromise) &&
+        fromPromise(this.terria.catalog.index!.loadPromise).state ===
+          "fulfilled"
       );
     }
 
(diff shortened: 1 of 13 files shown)


na9da (changes requested):
One thing to consider is that the static index is part of our default [CatalogSearchProvider](https://github.com/TerriaJS/terriajs/blob/ab9fac46444ce5a884c5dcc0d5f3f533eedc100f/lib/Models/SearchProviders/CatalogSearchProvider.ts) flow. It is most likely not going to be used by other catalog search providers (eg SomeServerCatalogSearchProvider). So, would it be more appropriate to move it to `CatalogSearchProvider` instead? `Catalog` itself doesn't have to be aware whether the search provider is using an index or not.

zoran995:
> One thing to consider is that the static index is part of our default [CatalogSearchProvider](https://github.com/TerriaJS/terriajs/blob/ab9fac46444ce5a884c5dcc0d5f3f533eedc100f/lib/Models/SearchProviders/CatalogSearchProvider.ts) flow. It is most likely not going to be used by other catalog search providers (eg SomeServerCatalogSearchProvider). So, would it be more appropriate to move it to `CatalogSearchProvider` instead? `Catalog` itself doesn't have to be aware whether the search provider is using an index or not.

I had the same idea when I started moving it, but we also use the `Catalog` index to find shared items. I assume the initial idea was to implement a catalogue index only for searching. Still, we extended to find items by share key when an item's location in the catalogue changes. Relevant PR https://github.com/TerriaJS/terriajs/pull/6052
method it is used in is getModelByIdShareKeyOrCatalogIndex, https://github.com/TerriaJS/terriajs/blob/09a574a6e2053373c9501584fc5b0fabf1ec4c53/lib/Models/Terria.ts#L978-L1006

na9da (approved):

PR #7858, opened 2026-06-09
feat: simplify initialization of user added data group

zoran995:
### What this PR does

### Test me

### Checklist

### DIFF
--- a/test/Models/TerriaSpec.ts
+++ b/test/Models/TerriaSpec.ts
@@ -125,7 +125,6 @@ describe("TerriaSpec", function () {
       function verifyGroups(groupAspect: any, groupNum: number) {
         const ids = groupAspect.members.map((member: any) => member.id);
         expect(terria.catalog.group.uniqueId).toEqual("/");
-        // ensure user added data co-exists with dereferenced magda members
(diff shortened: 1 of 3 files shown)


test/Models/TerriaSpec.ts
```diff
@@ -141,12 +140,12 @@ describe("TerriaSpec", function () {
         configUrl: "test/Magda/map-config-dereferenced.json",
         i18nOptions
       });
-      verifyGroups(mapConfigDereferencedJson.aspects["group"], 3);
+      verifyGroups(mapConfigDereferencedJson.aspects["group"], 2);
```
soyarsauce, line 144: @zoran995 why do these drop for magda groups as a result

soyarsauce (approved):
@zoran995 can you add a changelog entry before landing

test/Models/TerriaSpec.ts
zoran995: verifyGroup is a helper method defined above, and it will check the number value against the length of `terria.catalog.group.members`  and the user-added data group was removed from catalog group to separate property.

```
expect(terria.catalog.group.members.length).toEqual(groupNum);
```

PR #7859, opened 2026-06-09
feat: drop magda support, user can reimplement the support as part of load config

zoran995:
### What this PR does

Now that we support injecting the load config method, we should be able to drop support for loading Magda config without disrupting users too much. The user can easily load the Magda config as part of the config load method passed to the Terria start method. I have deprecated the MagdaReference catalogue item and removed it from the default catalog items registry. Users can re-add it to their own implementation if needed. If MagdaReference is needed, we can expose it as a plugin if there is a community interest.

Reasoning: We are not aware of any implementation using terriajs with Magda. Also, the learning curve for setting up Magda is pretty steep. 

### Test me

### Checklist


lib/Models/Terria.ts
```diff
@@ -449,11 +446,13 @@ interface StartOptions {
    * than a URL (e.g. a non-browser/SSR environment). Use the exported
    * {@link defaultLoadConfig} helper for the standard behaviour. When omitted,
    * `configUrl`/`configUrlHeaders` are loaded via {@link defaultLoadConfig}.
+   *
+   * @experimental
```
soyarsauce: Thanks for the additional annotation; could mirror the emphasis in `lib/Models/defaultLoadConfig.ts`

lib/Models/Catalog/CatalogReferences/MagdaReference.ts
soyarsauce, line 1: Very brief look - nothing in open repos
Image

soyarsauce (approved):

PR #7860, opened 2026-06-10
Fix minmap marker in pedestrian mode

na9da:
### What this PR does

Main change is to use CSS transform to rotate billboard images in Leaflet view.

Currently, we update billboard rotation by generating a new image every frame. 

This is:
   a. less performant
   b. isn't responsive enough because we draw the new image asynchronously in [onload](https://github.com/TerriaJS/terriajs/blob/fa2a741bb71cb2ac8d8958c38409204bcc0dd105/lib/Map/Leaflet/LeafletVisualizer.ts#L493-L495) callbacks which maybe deferred when there is continuous user input.

### Test me

Before:
- Enter pedestrian mode in main branch - http://ci.terria.io/main
- Simultaneously move forward and rotate the view by pressing `W` key and dragging the mouse to a side
- Note how the marker in the minimap does not rotate until you release `W` key

After:
- Repeat the same steps from this branch - http://ci.terria.io/fix-minmap-marker
- See how the minmap marker updates smoothly

### Checklist

- [X] There are unit tests to verify my changes are correct or unit tests aren't applicable (if so, write quick reason why unit tests don't exist)
- [X] I've updated CHANGES.md with what I changed.

### DIFF
--- a/lib/Map/Leaflet/LeafletVisualizer.ts
+++ b/lib/Map/Leaflet/LeafletVisualizer.ts
@@ -1287,7 +1287,7 @@ function recolorBillboard(
  * Use CSS transform to rotate the billboard icon
  */
 function rotateBillboardIcon(marker: L.Marker, rotation: number) {
-  const el: HTMLElement | undefined = (marker as any)._icon;
+  const el: HTMLElement | undefined = marker.getElement();
   if (!isDefined(el)) {
     return;
   }
--- a/lib/Map/Leaflet/LeafletVisualizer.ts
+++ b/lib/Map/Leaflet/LeafletVisualizer.ts
@@ -500,6 +500,16 @@ class LeafletGeomVisualizer {
         }
       }
     }
+
+    const rotation = getValueOrDefault(
+      markerGraphics.rotation,
+      time,
+      undefined
+    );
+
+    if (rotation !== undefined) {
+      rotateBillboardIcon(marker, rotation);
+    }
   }
 
   private _updateLabel(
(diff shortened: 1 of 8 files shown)


lib/Map/Leaflet/LeafletVisualizer.ts
```diff
@@ -1273,6 +1283,25 @@ function recolorBillboard(
   return canvas.toDataURL();
 }
 
+/**
+ * Use CSS transform to rotate the billboard icon
+ */
+function rotateBillboardIcon(marker: L.Marker, rotation: number) {
+  const el: HTMLElement | undefined = (marker as any)._icon;
```
zoran995: Can we use getElement here instead of the private API?  
https://github.com/Leaflet/Leaflet/blob/ff38fe55d31c05aabb6a2bd3cb948748d4e4b793/src/layer/marker/Marker.js#L203-L205

lib/ReactViews/Tools/PedestrianMode/MiniMap.tsx
```diff
@@ -37,12 +38,9 @@ const MiniMap: FC<MiniMapProps> = (props) => {
   /* eslint-disable-next-line react-hooks/exhaustive-deps */
```
zoran995: Is this still needed?

lib/Map/Leaflet/LeafletVisualizer.ts
```diff
@@ -500,6 +500,16 @@ class LeafletGeomVisualizer {
...
     }
+
+    const rotation = getValueOrDefault(
+      markerGraphics.rotation,
+      time,
+      undefined
+    );
+
+    if (rotation !== undefined) {
+      rotateBillboardIcon(marker, rotation);
```
zoran995, line 511: Is this the only place where we need to apply rotation to leaflet markers? Leaflet applies its own transform when the position or similar properties change, which will effectively remove our rotation and might cause unexpected side effects that we might need to consider.

soyarsauce:
@na9da can we add an entry to `changes.md` as well

soyarsauce:
Thanks for [the loom as well](https://www.loom.com/share/1aa814b9a0694080ac08b8933ab1507a)

lib/Map/Leaflet/LeafletVisualizer.ts
zoran995: @na9da we can merge this without changing this, but a generic approach from the leaflet side would involve extending the marker renderer to include rotation as part of position setting
Similar to what https://github.com/bbecquet/Leaflet.RotatedMarker is doing

zoran995 (approved):

PR #7862, opened 2026-06-12
experimental components DI to support easy components override

zoran995:
### What this PR does

This is a **proof of concept for a "components dependency injection container"** for the Explorer Window area of TerriaJS.

The idea: instead of importing UI components directly, the components that make up the Explorer Window (Data Catalog, Data Catalog search box, and the Explorer Window itself) are exposed through a single object — `ExplorerWindowComponents` — that can have its entries replaced from outside TerriaJS.

This gives us a way to swap out individual components quickly without rebuilding the entire interface, or having to polute terriajs interface with functionality that won't be actually used by the community. The motivation is to support faster iteration on the SaaS side: when we need a custom version of one of these components for a SaaS feature, we can override it in place rather than maintaining a parallel copy of the surrounding code. If a customisation proves generally useful, it's also much easier to backport into the OSS version later, since the override sits alongside the original component rather than diverging from it.

#### Example

```ts
import { ExplorerWindowComponents } from "terriajs/lib/ReactViews/ExplorerWindow/ExplorerWindowComponents";
import MyCustomSearch from "./MyCustomSearch";

// Replace TerriaJS's default Data Catalog with our own component
ExplorerWindowComponents.DataCatalogSearch = MyCustomSearch;
```

**note**: The team needed this to implement custom functionality on the SaaS side, and this is an experiment to make it possible. If it proves useful, we may extend it to other parts of the codebase.

### Test me

### Checklist

### DIFF
--- a/packages/terriajs/lib/ReactViewModels/SearchState.ts
+++ b/packages/terriajs/lib/ReactViewModels/SearchState.ts
@@ -19,9 +19,6 @@ export default class SearchState {
 
   @observable private _locationSearchText: string = "";
 
-  @observable unifiedSearchText: string = "";
-  @observable isWaitingToStartUnifiedSearch: boolean = false;
-
   @observable showLocationSearchResults: boolean = false;
   @observable showMobileLocationSearch: boolean = false;
   @observable showMobileCatalogSearch: boolean = false;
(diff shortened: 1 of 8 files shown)


soyarsauce:
If you intend on merging, can you update the desc a little @zoran995

soyarsauce:
Then we can get it in for use quickly

soyarsauce (approved):

PR #7867, opened 2026-06-17
Fix compass.

na9da:
### What this PR does

Previously:
- Camera changes were rendered only on mouse up event (or a network event) that woke up the [CesiumRendererLoopPauser](https://github.com/TerriaJS/terriajs/blob/main/packages/terriajs/lib/Map/Cesium/CesiumRenderLoopPauser.ts).
- There were some hardcoded magic values that were sensitive to CSS changes and got out of sync when styles changed. Have replaced them with values derived from the style.

### Test me

Before:
- Open CI from main branch - http://ci.terria.io/main
- Interact with the compass
- Notice that the updates are not instant - sometimes the updates render only once you have lifted or moved the mouse

After:
- Try this branch - http://ci.terria.io/fix-compass
- It should react to the changes instantly

### Checklist

### DIFF
--- a/packages/terriajs/lib/ReactViews/Map/MapNavigation/Items/Compass/Compass.tsx
+++ b/packages/terriajs/lib/ReactViews/Map/MapNavigation/Items/Compass/Compass.tsx
@@ -70,6 +70,17 @@ const getCompassScaleRatio = (compassWidth: string) =>
   (Number(compassWidth) + 10) / Number(compassWidth);
 
 /**
+ * Width of the inner side of the ring. This encircles the arrows icon.
+ */
+const gyroWidth = ({
+  compassWidth,
+  ringWidth
+}: {
+  compassWidth: string;
+  ringWidth: string;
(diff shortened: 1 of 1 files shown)


soyarsauce (approved):
You mention it's a bandage fix; what else could we do to improve this?

zoran995 (approved):

na9da:
@soyarsauce 

Not exactly sure, but main thing I would check is whether we can separate the orbit and rotate controls into two separate functional react components. Currently the trigger for orbit and rotate is handled in this [one callback](https://github.com/TerriaJS/terriajs/blob/cc10c8742bac60e5f4992d98aeb4b3d0ce83d615/packages/terriajs/lib/ReactViews/Map/MapNavigation/Items/Compass/Compass.tsx#L239) and we have to adjust those params to make it trigger the right function.

Also, ideally the UX should provide some hint on which function will be triggered when user hovers the mouse - orbit or rotate.

PR #7869, opened 2026-06-17
Refactor sendFeedback into FeedbackService class

zoran995:
### What this PR does

Replaces the standalone sendFeedback function with a FeedbackService class that is instantiated on the Terria model when feedbackUrl is configured, making it easier to inject and test feedback behavior.

### Test me

### Checklist


packages/terriajs/lib/ReactViews/Feedback/FeedbackForm.tsx
```diff
@@ -151,27 +150,36 @@ class FeedbackForm extends Component<IProps, IState> {
...
-        }
-      });
+      this.props.viewState.terria.feedbackService
+        ?.sendFeedback({
+          name: this.state.name,
+          email: this.state.email,
+          sendShareURL: this.state.sendShareURL,
+          comment: this.state.comment
+        })!
+        .then((response) => {
```
na9da, line 160: nitpick: Should we handle `.catch()` now that FeedbackService can be customised and may throw other unhandled errors?

packages/terriajs/lib/Models/Terria.ts
```diff
@@ -729,6 +730,7 @@ export default class Terria {
   );
 
   @observable serverConfig: any; // TODO
+  @observable feedbackService: IFeedbackService | undefined;
```
na9da, line 733: nitpick: If we rename `FeedbackService` to `DefaultFeedbackService` we can use `FeedbackService` as interface name instead of `IFeedbackService`

na9da (approved):
LGTM @zoran995. Just a couple of suggestions.

packages/terriajs/lib/ReactViews/Feedback/FeedbackForm.tsx
zoran995: Response format is typed with `result: 'SUCCESS' | 'FAILED'. If someone writes code according to the interface, it should be clear enough that the code should throw an error

PR #7870, opened 2026-06-18
chore: move dev dependencies to devDependencies

zoran995:
### What this PR does

Due to the previous build process in CI, we had to keep those as dependencies instead of dev dependencies, although they are used only as dev dependencies.

### Test me

### Checklist


soyarsauce (approved):
✅

soyarsauce:
@zoran995 I just did a release of terriajs 8.12.4 with these changes, this potentially breaks consumer TerriaMap pulling in new versions, can you check and revert this + release 8.12.5 if a TerriaMap's gulp scripts are no longer working?
e.g. go clone `TerriaMap` as it stands now, consume `terriajs` 8.12.4, try a `gulp lint`

Below is first symptom, but unravels as you try and re-add all the deps to TerriaMap
  
image

PR #7873, opened 2026-06-20
Revert "chore: move dev dependencies to devDependencies (#7870)"

zoran995:
### What this PR does

This reverts commit d697e29051821c5f775844a688855f6f790d605b. Wing discovered that we broke the environment when using terriajs outside of monorepo, so we need to revert this https://github.com/TerriaJS/terriajs/pull/7870#issuecomment-4749366162
p.s. I also explored the possibility of putting some of the terriajs dev dependencies into the dev dependencies of terriamap, but until we manage to publish the terriajs as a prebuilt package, it won't work. 

### Test me

### Checklist


soyarsauce (approved):
This really just surfaces how not consumable terriajs is on its own

PR #7880, opened 2026-06-23
chore: replace 8398a7/action-slack with slackapi/slack-github-action

zoran995:
### What this PR does

Migrates to the official Slack-maintained action (v3.0.3, SHA-pinned). Uses Block Kit instead of deprecated attachments. Reproduces repo, commit, action, and ref context with @here mention on both success and failure/cancellation notifications.
The replacement was generated by Claude.

Visual comparison of old and new messages, setting line with colour is possible only with attachments, which are deprecated
image
image

### Test me

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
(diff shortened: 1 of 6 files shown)


soyarsauce:
@zoran995 where are these being echo'd to today?

zoran995:
Not sure, they were posted to some Slack channel, but not sure which one, @peterhassall might know more

I tested it with a dummy repo and temp Slack workspace

PR #7882, opened 2026-06-23
Change i18next to use selector pattern

zoran995:
### What this PR does

Follow up on #7881 
Previously, our translation keys were written as plain strings with no compile-time validation and without IntelliSense - a renamed or deleted key would silently produce an untranslated string at runtime. This PR migrates all translation calls across the codebase to the i18next v26 selector pattern (`t(($) => $.namespace.key)`), which TypeScript validates against the translation files at build time, turning missing keys into compile errors.

To support this, the PR adds automated generation of TypeScript types from the English translation JSON files, wired into both the build and watch tasks so types stay in sync during development. The generated file is committed and enforced in CI - editing a translation file without regenerating the types fails the build.

applyTranslationIfExists is simplified to use i18next's built-in defaultValue fallback instead of a manual existence check. Built-in data type entries are updated to use typed selector keys, while the public API still accepts plain strings for third-party callers who provide their own display names.

It also improves the developer experience, as translation keys are autocompleted.

Migration was mainly done using codemod https://github.com/ahrjarrett/i18next-selector, followed by regexes to fix remaining issues. It actually wasn't too complex to update all of them once I got the build issue to properly throw in development mode. 

### Test me

### Checklist


.github/workflows/ci.yml
```diff
@@ -45,6 +45,9 @@ jobs:
         env:
           NODE_OPTIONS: --max_old_space_size=4096
 
+      - name: Check i18n types are up to date
```
soyarsauce: Is there an auto fix for this if it fails?

soyarsauce (approved):

.github/workflows/ci.yml
zoran995: Forgot to remove that, it was needed when I generated the types from the JSON file, but importing JSON directly gives us better DX as go to definition leads directly to `translation.json` file so this is not needed anymore

PR #7883, opened 2026-06-24
gaussian splat fix in order to visualize them properly

Oscarbralo:
### What this PR does

Before this change, gaussian splat loaded but was not able to visualize inside. 

You could see some data but it looked corrupted. 
But because how the splat renderer is treated different, the splat builds an ENU frame on top of the modelMatrix, and cesium treat it as modelMatrix x ENU, so it was taking it in account a second time, so the splat looked like corrupted or zoomed in.

So the fix is: for splats, don't put the big transform in modelMatrix, so Cesium's ENU frame lands where it's supposed to.

I added the pickTranslucentDepth in order to move the splat easily.

All the other parts should work as previously.

### Test me

Just apply the fix and add any gaussian splat you want, you should be able to visualize it correctly. 

Before the change the GS loads but was not able to visualize it.

### Checklist

gaussianAfter
GSBefore

### DIFF
--- a/packages/terriajs/lib/ModelMixins/Cesium3dTilesMixin.ts
+++ b/packages/terriajs/lib/ModelMixins/Cesium3dTilesMixin.ts
@@ -47,6 +47,58 @@ import ClippingMixin from "./ClippingMixin";
 import MappableMixin from "./MappableMixin";
 import ShadowMixin from "./ShadowMixin";
 
+/**
+ * terriajs-cesium 23.0.2 references `Cesium3DTileset#isGltfExtensionRequired` in its
+ * Gaussian splat content routing (see Cesium's Cesium3DTileContentFactory +
+ * GaussianSplat3DTileContent.tilesetRequiresGaussianSplattingExt) but never defines the
+ * method. The routing guard `tileset.isGltfExtensionRequired instanceof Function` is
+ * therefore always false, so every Gaussian splat (.glb) tile falls back to the regular
+ * model loader: the tileset loads but renders nothing. Upstream CesiumJS defines these
+ * methods, which is why the same Cesium ion asset renders in stock Cesium but is invisible
+ * in Terria.
+ *
+ * We define the missing methods here, reading the glTF extension lists that Cesium ion
+ * stores under the tileset's `3DTILES_content_gltf` extension. This runs once when the
+ * module is imported (before any tileset instance is created) and is a no-op if a future
+ * terriajs-cesium version provides the methods itself.
+ */
+function ensureGltfExtensionDetection(): void {
+  const proto = Cesium3DTileset.prototype as unknown as {
+    isGltfExtensionUsed?: (extensionName: string) => boolean;
+    isGltfExtensionRequired?: (extensionName: string) => boolean;
+  };
+
+  const hasGltfExtension = (
+    tileset: { extensions?: Record<string, any> },
+    list: "extensionsUsed" | "extensionsRequired",
+    extensionName: string
+  ): boolean => {
+    const contentGltf = tileset.extensions?.["3DTILES_content_gltf"];
+    const extensions = contentGltf?.[list];
+    return Array.isArray(extensions) && extensions.indexOf(extensionName) > -1;
+  };
+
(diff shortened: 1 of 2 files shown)


packages/terriajs/lib/ModelMixins/Cesium3dTilesMixin.ts
```diff
@@ -47,6 +47,58 @@ import ClippingMixin from "./ClippingMixin";
...
 
+/**
+ * terriajs-cesium 23.0.2 references `Cesium3DTileset#isGltfExtensionRequired` in its
+ * Gaussian splat content routing (see Cesium's Cesium3DTileContentFactory +
+ * GaussianSplat3DTileContent.tilesetRequiresGaussianSplattingExt) but never defines the
+ * method. The routing guard `tileset.isGltfExtensionRequired instanceof Function` is
+ * therefore always false, so every Gaussian splat (.glb) tile falls back to the regular
+ * model loader: the tileset loads but renders nothing. Upstream CesiumJS defines these
+ * methods, which is why the same Cesium ion asset renders in stock Cesium but is invisible
+ * in Terria.
```
zoran995: I find this strange as cesium is defining those functions on Cesium3DTileset.prototype https://github.com/TerriaJS/cesium/blob/upgrade-to-1.142/packages/engine/Source/Scene/Cesium3DTileset.js#L3915-L3959

zoran995 (commented):
Do you maybe have an exaple dataset that can be used to reproduce the issue you are experiencing? I tried using the one provided in cesium PR but to me it looks like everything loads and render correctly, but I might be missing something

packages/terriajs/lib/ModelMixins/Cesium3dTilesMixin.ts
Oscarbralo: @zoran995 Oh, I think that was the first problem I had. Maybe because I was using first an old version. I will modify that off part, the imortant one is the other. Thanks for pointing it out

Oscarbralo: @zoran995 removed the old code. Sorry about it.

Oscarbralo: [cactus_splat3_30kSteps_719k_splats.compressed.zip](https://github.com/user-attachments/files/29314840/cactus_splat3_30kSteps_719k_splats.compressed.zip)
@zoran995  As for the example, here it is. In the normal version of Terriamap it doesnt render correctly.

zoran995: Hi @Oscarbralo, how are you loading it in the app? Which version of terriajs are you using? We recently upgraded cesium to the latest version. 
I tried converting to non-compressed ply (using supersplat) and then converting to 3d tiles using cesium ion, and then rendered 3d tiles in the latest terriajs (at least to me, it looks the same for all options i tried)
image

Oscarbralo: I am loading it from Cesium Ion. @zoran995 

Pretty weird. I tried in the last one you guys released, and couldnt make it work. Will try again

I might be doing something wrong.

Oscarbralo: @zoran995 I made several tests and I cannot make it work with the last version just as it is. 

As for me, even using the test that I sent you, still not work. 

Pretty interesting. Any idea what could be the problem on my side?

Oscarbralo: @zoran995 
https://github.com/user-attachments/assets/7a588292-f078-4966-ba95-7740a10a8afe

For this video, I am using the last version of TerriaMap. 

This terriamap is pointing to the terriajs#main so they should be both in the latest version, is that correct?

Just cloned TerriaMap#main branch, yarn install and run it. 

load the 3dgs from cesium ion, and that is the result.

Oscarbralo:
looks like its working in the last version

PR #7888, opened 2026-07-13
Fix zoom to mappable rectangle.

na9da:
### What this PR does

When zooming to a mappable's rectangle, use the same code path we have for `Rectangle` which is smarter and sets a minimum viewing distance instead of simply fitting the camera view to the rectangle. This is a better behaviour when zooming to simple shapes like a single polygon.

### Test me

Before:
- Open [this CI link](http://ci.terria.io/main/#share=s-xDiGD11AMLSrPulD4V5HYpDONzb) for `main` branch and zoom to the item in workbench
- Observe that the zoom is too close and possibly confusing to the user

After:
- Open [the CI link](http://ci.terria.io/fix-zoom/#share=s-xDiGD11AMLSrPulD4V5HYpDONzb) for this branch and repeat above steps
- Observe that the full layer is correctly in view

### Checklist


zoran995:
@na9da 2 tests are failing 
- TerriaSpec initial zoom behaviour of `initialCamera.focusWorkbenchItems` zooms the map to focus on the workbench items
- TerriaSpec initial zoom behaviour of `initialCamera.focusWorkbenchItems` works correctly even when there is a delay in a Cesium/Leaflet viewer becoming available
Others are not relevant and should pass after those 2 are fixed (they are flaky

Other than that, zoom looks much better now

na9da:
Thanks @zoran995. Fixed. The spec was relying on the zoom happening immediately which was no longer happening because of the nested call.

zoran995 (approved):

PR #7891, opened 2026-07-13
Fix `viewCatalogMember` to correctly switch to the parent tab.

na9da:
### What this PR does

Ensure that the parent tab is focused when `viewCatalogMember` is called for a catalog item. Example by clicking the `About data` button.

### Test me

This is again hard to repro in CI, happens only when outputting job execution results to a group.

### Checklist


zoran995 (approved):
it would be good to cover the behaviour with tests

na9da:
Thanks @zoran! I used Claude to generate a few test cases.

PR #7901, opened 2026-07-15
fix(security): scope custom-markdown sanitizer, drop ALLOW_UNKNOWN_PROTOCOLS

zoran995:
### What this PR does

Replace the blanket ALLOW_UNKNOWN_PROTOCOLS DOMPurify option, which preserved OS/app-handler schemes (vscode:, ms-its:, slack:, ...) on `<a href> / `, with a scoped uponSanitizeAttribute hook added/removed around our own sanitise call. The hook force-keeps custom components' free-text attributes (so values may contain ':') and scheme-validates their URL attributes per URL, closing the comma-separated sources/downloads list gap.

Adds lib/Core/sanitizeUrl.ts and CustomComponent url/free-text classification.

### Test me

Before: http://ci.terria.io/main/#https://gist.githubusercontent.com/zoran995/d85d5bfc6b9e6b366220c10052eaac36/raw/726f58e47cad60942dadfb95bc6a198a84f67584/7901.json
After:
http://ci.terria.io/feat-dompurify-custom-tags/#https://gist.githubusercontent.com/zoran995/d85d5bfc6b9e6b366220c10052eaac36/raw/726f58e47cad60942dadfb95bc6a198a84f67584/7901.json
Verify that description of catalog group `Sanitizer + tjs- Test` is properly sanitized 

### Checklist


na9da:
Instead of allowing `:` in attributes, should we just remove support for it entirely? 

I think in [this PR](https://github.com/TerriaJS/terriajs/pull/4866) we added an alternative to using the key/value style `column-titles="columnName:title1"` with just a simple list `column-titles="title1,title2"`. Same for `column-units`. 

The [DEA map](https://raw.githubusercontent.com/GeoscienceAustralia/dea-config/refs/heads/master/prod/terria/dea-maps-v8.json) for example uses the colon-less syntax for `column-units`.

na9da:
in both `main` and this PR branch, the chart is not showing:

image

na9da:
Also here's [the commit](https://github.com/TerriaJS/terriajs/commit/ed00e9cd0a85cd6e28609481be4d9c67a6346694) that started using `:`. It used to be `columName#title` before.

zoran995:
> in both main and this PR branch, the chart is not showing:


It is getting blocked by `example.com`. Here is the working example for charts
before
http://ci.terria.io/main/#https://gist.githubusercontent.com/zoran995/d85d5bfc6b9e6b366220c10052eaac36/raw/13608cb647632ba47bc68a0fa73ebceb448115ea/7901-2.json
after
http://ci.terria.io/feat-dompurify-custom-tags/#https://gist.githubusercontent.com/zoran995/d85d5bfc6b9e6b366220c10052eaac36/raw/13608cb647632ba47bc68a0fa73ebceb448115ea/7901-2.json

na9da:
Thanks @zoran995,

I'm more inclined to reverting to `#` instead of `:` like it used to be in v7. I don't think the syntax is currently being used by anyone.

Reasoning for keeping support for `column#title` pairs: If we only support a list like `title1,title2` it will be hard to change the title of some column `n`. Also, it becomes position dependent.

zoran995:
@na9da Changing from `:` to `#` doesn't simplify things a lot, as our URL props allow comma-separated values and dompurify validates only the first value. 
i.e. this will pass default dompurify validation `"sources='https://example.com/ok.csv,vscode://evil'`, and it will be executed by us as a data source. 
_edit: We could change to `#` and then apply our hook to all props without needing to know if it is url or not_

na9da (approved):
Sounds right! Thanks @zoran995

PR #7902, opened 2026-07-16
fix(security): guard CSV exports against formula injection

zoran995:
### What this PR does

Prefix cells that a spreadsheet would treat as a formula (=, +, -, @, tab, CR) with a single quote on export, without turning legitimate numbers (e.g. -5) into text. Covers the feature-info CSV download, TableMixin table export, and CsvCatalogItem's inline csvString export. TableMixin exports now also use papaparse for structural escaping (previously joined with ',' unescaped). Remote-url downloads are unchanged (original file served directly).

Also upgrade papaparse to 5.5.4 and adapt loadCsv/Csv to its new types (ParseConfig split into local/worker configs; comments is now string|false).

### Test me

### Checklist


packages/terriajs/lib/Core/sanitizeCsvString.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Parses a CSV string and re-serialises it with every cell passed through
+ * {@link sanitizeCsvValue}, neutralising CSV/formula injection (CWE-1236) while
+ * preserving cell values. The CSV is re-serialised (via papaparse), so exact
+ * byte-for-byte formatting of the source is not preserved.
+ */
+export default function sanitizeCsvString(csv: string): string {
+  const rows = Papa.parse<string[]>(csv).data;
+  return Papa.unparse(rows.map((row) => row.map(sanitizeCsvValue)));
```
na9da: nitpick: Could something like the following be slightly better from a performance perspective? 

```
Papa.unparse(Papa.parse(csvText, {
  transform: sanitizeCsvValue,
}));
```

na9da (approved):
LGTM @zoran995 

Just a nitpicks + suggestion to consolidate the separate `sanitize*.ts` files into one maybe.

PR #7904, opened 2026-07-20
Update styled-components to 6.

na9da:
### What this PR does

- Updates styled-components to version 6 and and fixes resulting type errors

### Checklist


zoran995 (approved):

na9da:
Thanks @zoran!

I've made one more fix to the `Select` component to remove `css={...}` prop to `react-select`. 
* This actually didn't have any effect (?)
* In the new version of styled-components, the `theme` props gets captured by styled-components and is not passed through to `react-select` which breaks the theming. 

Previously:
image

Now:
image

PR #7919, opened 2026-08-20
ci: notify monorepo of terriajs changes (Copybara sync)

zoran995:
Part of adopting [Copybara](https://github.com/google/copybara) to keep `terriajs` in sync (both directions) with the `oss/` subtree of the private monorepo. The sync workflows themselves live in the monorepo; this is the only terriajs-side piece.

**What it does:** on every push to `main`, fire a `repository_dispatch (terriajs-merged)` event, so the monorepo pulls the change into `oss/` promptly instead of waiting for the scheduled inbound run.

**Prerequisite:** the shared GitHub App must be installed on this repo too, with repo variable `APP_CLIENT_ID` and secret `APP_PRIVATE_KEY`. Until then the workflow is inert (no-op on token mint).

**Note for maintainers:** automated `copybara/from-monorepo` PRs must be merged with a **normal merge commit** (no squash/rebase). Full runbook lives in the monorepo at `docs/copybara-sync.md`.

Follow-up (optional): a required-status **merge gate** that blocks merging a PR until the monorepo has synced out, to prevent the mirror's silent-reversion race.

### DIFF
--- a/.github/workflows/copybara-notify.yml
+++ b/.github/workflows/copybara-notify.yml
@@ -1,0 +1,39 @@
+name: Notify monorepo (Copybara inbound)
+
+# terriajs is kept in sync (both ways) with a subtree of TerriaJS/saas-settings via
+# Copybara. When terriajs main changes, ping the monorepo so it pulls the change into
+# oss/ promptly (instead of waiting for the scheduled inbound run).
+#
+# Requires the shared GitHub App (installed on both terriajs and saas-settings):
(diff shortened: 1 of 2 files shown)


soyarsauce (approved):

.github/workflows/copybara-notify.yml
soyarsauce, line 1: Add a merge gate as well

